PR #1687, opened 2022-09-09
use the new ESP Framework for our older boards too

caveman99:


src/mesh/http/WiFiAPClient.cpp
```diff
@@ -238,10 +238,8 @@ bool initWifi(bool forceSoftAP)
                 WiFi.softAPConfig(apIP, apIP, IPAddress(255, 255, 255, 0));
                 DEBUG_MSG("MY IP AP ADDRESS: %s\n", WiFi.softAPIP().toString().c_str());
 
-#if !CONFIG_IDF_TARGET_ESP32S3
                 // This is needed to improve performance.
-                esp_wifi_set_ps(WIFI_PS_NONE); // Disable radio power saving
```
sachaw, line 243: Are these no longer needed?

src/mesh/http/WiFiAPClient.cpp
```diff
@@ -254,35 +252,24 @@ bool initWifi(bool forceSoftAP)
...
                         */
-                        wifiDisconnectReason = info.disconnected.reason;
-                        #endif
-
+                        wifiDisconnectReason = info.wifi_sta_disconnected.reason;
                     },
-#if CONFIG_IDF_TARGET_ESP32S3
                     WiFiEvent_t::ARDUINO_EVENT_WIFI_STA_DISCONNECTED);
-#else
-                    WiFiEvent_t::SYSTEM_EVENT_STA_DISCONNECTED);
```
sachaw, line 284: should this not be `WiFiEvent_t::ARDUINO_EVENT_WIFI_STA_DISCONNECTED);` ?

sachaw (commented):
LGTM, just 2 points I'm not sure about

src/mesh/http/WiFiAPClient.cpp
caveman99: They were removed from the ESP32-S3 target by Macroguard. I have not looked into the new API if there's a function equivalent for ESP32(vanilla) or if the new API just ditches this call, but the S3 doesn't seem to need it any more.

src/mesh/http/WiFiAPClient.cpp
caveman99: it is - the else branch is the old API and is removed by this PR :-)

caveman99:
Just something to note - i modified the HTTP Server we forked to not differentiate between ESP32-S3 and standard target any more, there's one include in there that was moved for SHA Hashes. This PR uses my fork for now (POC) but we should include this mod in our meshtastic fork once this PR is merged.

sachaw:
LGTM, I can merge when your ready

caveman99:
I'm busy over the weekend, will do some more tests monday, not 100% satisfied with this yet.

PR #1689, opened 2022-09-11
More default inits.

sachaw:

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -157,7 +157,7 @@ void setup()
 #endif
 
 #ifdef DEBUG_PORT
-    if (!config.device.serial_disabled) {
+    if (config.device.serial_enabled) {
         consoleInit(); // Set serial baud rate and init our mesh console
     }
 #endif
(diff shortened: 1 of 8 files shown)


sachaw:
Protos need merging before merging this.

src/main.cpp
```diff
@@ -157,7 +157,7 @@ void setup()
 #endif
 
 #ifdef DEBUG_PORT
-    if (!config.device.serial_disabled) {
+    if (config.device.serial_enabled) {
```
thebentern, line 160: This one may be a problem since I believe it happens before nodedb sets up all of the defaults

sachaw: Can we put in a condition where it's accessed before initialised to handle unset as enabled?

sachaw:
currently hanging on boot:
```
0x400d78cd:0x3ffd5750 0x400e0c95:0x3ffd57a0 0x4012f156:0x3ffd57c0
```

PR #1712, opened 2022-09-21
Compiler warnings

caveman99:


src/mesh/NodeDB.cpp
```diff
@@ -493,7 +492,7 @@ void NodeDB::saveToDisk()
         moduleConfig.has_serial = true;
         moduleConfig.has_store_forward = true;
         moduleConfig.has_telemetry = true;
-        saveProto(moduleConfigFileName, LocalModuleConfig_size, sizeof(LocalModuleConfig), LocalModuleConfig_fields, &moduleConfig);
+        saveProto(moduleConfigFileName, LocalModuleConfig_size, sizeof(moduleConfig), LocalModuleConfig_fields, &moduleConfig);
```
thebentern, line 495: Nice. Looks like we were over-allocating for this one?

caveman99: no, that parameter is actually not in use. Ongoing investigation :-)

thebentern: Well that's really weird

PR #1734, opened 2022-09-29
Build image

caveman99:
This will build ONE image for flashing that contains bootloader (the correct one for each board), partition table, OTA setup and the main firmware. OTA firmware and littlefs still go in seaparate bins.

Affects ESP32 only so may need additinal conditions for CI.


thebentern:
Built fine for me on a T-Beam via pio. I dig it! What do to change our device-install scripts now to account for the combined file?

caveman99:
i'll check in changes for device-install and update, we have to adjust CI too...

caveman99:
CI now needs to skip partitions and system info, instead collect and rename firmware.factory.bin

Also the Meshtastic-GUI installer needs to be modified for the new offset and missing parts.

thebentern (approved):

PR #1737, opened 2022-10-01
Let SimRadio communicate via TCP

GUVWAF:
If this PR gets accepted, it should be done in combination with my PRs in Meshtastic-python, Meshtastic-protobufs and framework-portduino. 
This enhances the SimRadio for the Linux native application, in order to let multiple instances communicate using TCP as if they did via their LoRa chip. The procedure on how to use it can be found [here](https://github.com/GUVWAF/Meshtasticator/blob/master/INTERACTIVE_SIM.md).


src/mesh/SimRadio.cpp
```diff
@@ -0,0 +1,250 @@
+#include "SimRadio.h"
```
caveman99, line 1: Since these (h/cpp) are only for Portduino, please move them to src/platform/portduino. Else all other targets will compile (but not link) them

src/mesh/wifi/WiFiServerAPI.h
```diff
@@ -44,7 +44,7 @@ class WiFiServerPort : public WiFiServer, private concurrency::OSThread
     WiFiServerAPI *openAPI = NULL;
 
   public:
-    WiFiServerPort();
+    WiFiServerPort(int port);
```
caveman99, line 47: src/mesh/wifi/WiFiServerAPI.h:47: [low:style] Class 'WiFiServerPort' has a constructor with 1 argument that is not explicit. [noExplicitConstructor]

caveman99 (changes requested):

PR #1758, opened 2022-10-06
update PCF8563_Library to 1.0.0

andrekir:
per https://github.com/meshtastic/Meshtastic-device/commit/dc097c72303acef70d142b617e213f50096349c9#commitcomment-85964186

### DIFF
--- a/variants/m5stack_coreink/platformio.ini
+++ b/variants/m5stack_coreink/platformio.ini
@@ -12,7 +12,7 @@ build_flags =
 lib_deps = 
   ${esp32_base.lib_deps}
   zinggjm/GxEPD2@^1.4.9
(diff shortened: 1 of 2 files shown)


caveman99:
On https://registry.platformio.org/libraries/lewisxhe/PCF8563_Library/versions there is the (oldest) 1.0.0, 0.0.1 and 0.0.2 - 0.0.2 is the most recent one, seems the registry did not pick up (or ignore) the new 1.0.0 and fetches the 3 year old one that throws errors. Let's go with release hash for now. until there is a 1.0.1

caveman99:
1.0.1 was pushed to github by @lewisxhe, waiting for platformio to pick it up. Thanks for the quick action.

PR #1760, opened 2022-10-06
Use small font for freetext messages

caveman99:


caveman99:
Todo: 
- [ ] Wordwrap on freetext input 
- [ ] show 'Sending' Screen every time
- [ ] Select target for message. Channel/Node/Broadcast

PR #1764, opened 2022-10-07
Add t-beam-s3-core onboard device support.

lewisxhe:
Add t-beam-s3-core onboard device support. For details, see the commit description . 

For the QMC6310 and QMI8658 sensors, I will continue to submit PR after completing the library of these two sensors

I have commented the relevant changes in the code. Please check

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -214,6 +214,10 @@ void setup()
     // router = new DSRRouter();
     router = new ReliableRouter();
 
+#ifdef I2C_SDA1
+    Wire1.begin(I2C_SDA1, I2C_SCL1);
+#endif
+
 #ifdef I2C_SDA
     Wire.begin(I2C_SDA, I2C_SCL);
 #elif HAS_WIRE
@@ -229,7 +233,6 @@ void setup()
     delay(1);
 #endif
 
-    scanI2Cdevice();
 #ifdef RAK4630
     // scanEInkDevice();
 #endif
@@ -269,6 +272,12 @@ void setup()
     powerStatus->observe(&power->newStatus);
     power->setup(); // Must be after status handler is installed, so that handler gets notified of the initial configuration
 
+    /*
+    * Move the scanning I2C device to the back of power initialization. 
+    * Some boards need to be powered on to correctly scan to the device address, such as t-beam-s3-core
+    */
+    scanI2Cdevice();
+    
     // Init our SPI controller (must be before screen and lora)
     initSPI();
 #ifndef ARCH_ESP32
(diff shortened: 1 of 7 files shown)


lewisxhe:
@caveman99 Please check

src/main.cpp
```diff
@@ -269,6 +272,12 @@ void setup()
     powerStatus->observe(&power->newStatus);
     power->setup(); // Must be after status handler is installed, so that handler gets notified of the initial configuration
 
+    /*
+    * Move the scanning I2C device to the back of power initialization. 
+    * Some boards need to be powered on to correctly scan to the device address, such as t-beam-s3-core
+    */
+    scanI2Cdevice();
+
```
caveman99, line 280: will this work with the AXP chips? Do they rely on this scan? Everything else comes in afterwards.

variants/tbeam-s3-core/variant.h
```diff
@@ -1,12 +1,18 @@
...
+#define I2C_SDA          17         //For sensors and screens
+#define I2C_SCL          18         //For sensors and screens
 
 #define BUTTON_PIN      0     // The middle button GPIO on the T-Beam S3
 //#define BUTTON_PIN_ALT 13 // Alternate GPIO for an external button if needed. Does anyone use this? It is not documented anywhere.
 // #define EXT_NOTIFY_OUT 13 // Default pin to use for Ext Notify Module.
 
+#define BUTTON1_PIN     47      //Additional keys
+
+
```
caveman99, line 15: What will the button be used for? If there's no default action it may be good to leave it undefined and map it through the inputbroker.

caveman99 (changes requested):
Please review my comments, everything else looks good to me ...

src/main.cpp
lewisxhe: Although the sensor uses Wire, while the PMU uses Wire1, their bus is not shared. If the PMU is not initialized before the sensor voltage is turned on, the scanI2Cdevice will be blocked because the wire cannot scan the device correctly, similar to T-beam's inability to initialize the PMU after turning off DC3

variants/tbeam-s3-core/variant.h
lewisxhe: At present, this button is not used. I just want to highlight that there is an additional button in the t-beam-s3-core. Can I add the function in the PR later

src/main.cpp
caveman99: I was considering the existing T-Beam with AXP192 too - but from the code the i2c detect for them is just cosmetic. So this is good to go!

variants/tbeam-s3-core/variant.h
caveman99: maybe leave the define in and comment it out? We can map additional pins/buttons through a run time configuration, e.g. for a rotary encoder or external sensors.

lewisxhe: This cannot be done because on hardware it has been defined as a key (already physically connected as a key) and it cannot be used as another function

caveman99 (approved):

PR #1768, opened 2022-10-07
Nrf52 rollup

caveman99:


caveman99:
I was hoping this would not happen on Linux. There seems to be a bug with escaping linker include paths that contain spaces on this new version of the toolkit. 

`/home/runner/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/7.2.1/../../../../arm-none-eabi/bin/ld: cannot find -lnrf_cc310_0.9.13-no-interrupts` is not found because `-L"path/to/Adafruit_nRFCrypto"` is wrongly escaped as `"-Lpath/to/Adafruit_nRFCrypto"`

PR #1770, opened 2022-10-08
Switch to OTA firmware

caveman99:
this is pending the protobuf pull...


caveman99:
force pushed cause of CI issues

PR #1774, opened 2022-10-08
erase oldest entry from NodeDB if it is full

caveman99:
plus some wiphone font size change that crept in, sorry for that.


thebentern (approved):
Looks good

Issue #1763, opened 2022-10-07
[Bug]: MQTT JSON Bug


arduionoGP:
### Category

Other

### Hardware

T-Beam, Heltec v2.1

### Firmware Version

1.3.42.9bd9252

### Description

**Environments and settings:**
1.3.42.9bd9252 firmware, heltec and t-beam devices, android 1.3.41, windows chrome web, node-red mqtt broker, using default channel for this example. One device is connected to wifi with Settings for default channel uplink and downlink ENABLED. Settings module, MQTT config, encryption off, JSON output enabled ON.

**Issue:**
When a sequence of LORA message packets is sent to a lora module connected to a mqtt broker, the broker responds by repeatedly sending out only the first message received from that particular client. The correct text messages in the sequence correctly display on the receiving OLED.

**Example:**
Device one which is connected by bluetooth via android sends three messages:    “Let’s stay another message”,    “I meant try”,    
 “It doesn’t increment”.
Device two is connected via wifi network to a mqtt broker. The broker responds to each message but the payloads sent all only contain the first message's text. The three messages are published by the broker on topic:
 “msh/1/json/LongFast/!7efec08c”
and all look like:

{“channel”: 0, “from”: 2130621128, “id”: -1175740138, “payload”: {“text”: “Let’s stay another message”}, “sender”: “!7efec08c”, “timestamp”: 0, “to”: -1, “type”: “text”}

Reset via button on the receiving lora device doesn't fix it. Pulling power on that device to reset allows a fresh message to be sent via mqtt. Two different sending devices will send two different packets to the receiving device connected via mqtt. Those two different messages will then continue to repeat depending on the sending machine.

caveman99:
the MQTT stuff is quite verbose in debug logging. Can you attach a serial debug log of the device (meshtastic --noproto if you got CLI installed) so we can have a look what is actually transmitted? There should be lines like 'publish json message to...' in there

arduionoGP:
Sure. While doing this I noticed another issue.
When I am connected directly via CLI or via the web client the messages published by the MQQT Broker are either missing the "text" field in the JSON payload or are not even being sent to the broker.  This only resolves by disconnecting all serial and a reset of the Heltec. The broker I was using is Aedes running in nod-red on Win10 local network. I just duplicated this behavior running a mosquito broker on a Pi so it's not a broker issue.

When NOT connected via CLI the payload of the packets on topic msh/1/json/LongFast/!7efec08c look like:
[...]

PR #1815, opened 2022-10-17
Don't make a nested json object payload

thebentern:


thebentern:
At least solves part of #1763

PR #1823, opened 2022-10-19
Json cleanup

caveman99:
Only allow 'sendtext' type to be forwarded on. The rest will be discarded


caveman99:
This will prevent 'double looping' of JSON and Proto payloads but still retains the possibility to send out text via JSON.

PR #1842, opened 2022-10-22
Rak13800

caveman99:


src/main.cpp
```diff
@@ -438,11 +438,13 @@ void setup()
     mqttInit();
 #endif
 
+#ifndef ARCH_PORTDUINO
```
thebentern, line 441: It's a thorn that pokes us when we do anything fun

caveman99: plus i don't really get the linker error. network-stubs.cpp should take care of that. That ifnded should not be neccessary. but maybe portduino is 'different' :-)

caveman99:
At least it doesn't blow up in my face any more ...
```
??:??:?? 7 SX1262 Radio init succeeded, using SX1262 radio
??:??:?? 7 Not using Ethernet
??:??:?? 7 (bw=250, sf=11, cr=4/8) packet symLen=8 ms, payloadSize=237, time 3188 ms
??:??:?? 7 myNodeInfo.bitrate = 74.341286 bytes / sec
??:??:?? 7 PowerFSM init, USB power=1
??:??:?? 7 Enter state: BOOT
```

PR #1843, opened 2022-10-22
When we init default channels, set use_preset to true

thebentern:
I don't believe this setting is honored yet, but we should at least initialize it correctly.


caveman99:
The code assumes bandwidth 0 as use_preset .. would be a simple swap.

thebentern:
> The code assumes bandwidth 0 as use_preset .. would be a simple swap.

I think so. Should just be a matter of adding that as a guard to the use-preset switch in the radio interface unless I’m missing something. 
I’m going to be out for a couple of hours. I’ll make that change as well. @garth and I ran across this while doing some channel qr testing across platforms.

PR #1847, opened 2022-10-22
[rak4630] enable 3.3v periphery before scan for i2c devices

### LINKED ISSUE #1846: [Bug]: i2c does not work without gps on rak4631
### Category

Hardware Compatibility

### Hardware

Rak4631

### Firmware Version

1.3.48.2c1bbf1a-d

### Description

If GPS is not connected I2C cannot be initialized, PR is coming.

### Relevant log output

_No response_

lesykm:
3.3 Enable pin in RAK4631 not only enables GPS but any other 3.3V periphery including I2c.
This PR is intended to enable 3.3V periphery from the start.
fixes #1846

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -242,6 +242,12 @@ void setup()
     delay(1);
 #endif
 
+#ifdef RAK4630
+    // We need to enable 3.3V periphery in order to scan it
+    pinMode(PIN_3V3_EN, OUTPUT);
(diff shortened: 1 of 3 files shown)


thebentern:
This is a super good catch. I'm going to take a bit of a closer look before merging because there's some subtleties to the eink RAK target I want to make sure we're good on first.

thebentern:
I think we're good. Merging!

PR #1856, opened 2022-10-25
just remove the proto calls, leave button-force-AP in place

caveman99:
Alternative to the other PR, less invasive.


caveman99:
Killing Force-AP too

PR #1858, opened 2022-10-25
[modules][canned messages] add predefined destination per message

lesykm:
Part 2 ( after Part 1: https://github.com/meshtastic/Meshtastic-protobufs/pull/222 ) for implementing of https://github.com/meshtastic/Meshtastic-device/issues/1837

Algo is following:
1) For all messages find correct address in destinations field.
2) If address is empty - use broadcast
3) If address is invalid - use broadcast
4) If number of addresses < number of messages - pad them with broadcasts.


lesykm:
If this approach seems OK, then will add admin/python code to CRUD these.

lesykm:
Looks like this PR will not be accepted, closing it.

PR #1867, opened 2022-10-27
Powers off eink backlights before shutdown

t413:
- t-echo boards in particular need this
- follows the existing defines structure for backlight control

thanks!

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -234,7 +234,7 @@ bool Power::setup()
 void Power::shutdown()
 {
 
-#if defined(USE_EINK)
+#if defined(USE_EINK) && defined(PIN_EINK_EN)
     digitalWrite(PIN_EINK_EN, LOW); //power off backlight first
 #endif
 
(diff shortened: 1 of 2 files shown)


t413:
Tested on-device and it works as expected. Backlight tuns off when shut down.

src/Power.cpp
```diff
@@ -234,6 +234,9 @@ bool Power::setup()
 void Power::shutdown()
 {
 
+#if defined(USE_EINK)
+    digitalWrite(PIN_EINK_EN, LOW); //power off backlight first
+#endif
```
caveman99, line 239: Can you please add an additional macro guard to make sure PIN_EINK_EN is defined before you use it? Asking for a hypothetical new device with EINK but without that PIN Definition.

caveman99 (changes requested):

t413:
Also addressed the same #define guard in src/ButtonThread.h:162 since that's what I based it on.

caveman99 (approved):

Issue #1647, opened 2022-08-24
Bug: Device power-down leaves text on screen after shutdown


mc-hamster:
When the unset screen is displayed, holding down the middle button on tbeam will not power down the node.

This is on 1.3.39.

thebentern:
Hmmm, holding down the PWR button for 10 seconds is working on my node with the region unset screen.

garthvh:
Same, I have two devices where I can pull the screen and both are behaving properly, is your AXP dead on the board in question?

garthvh:
I updated the title, my device shuts down but the text of the last screen displayed stays on the screen

garthvh:
04:22:38 730 [Button] Long press start!
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.
Core 1 register dump:
PC      : 0x401f5e28  PS      : 0x00060530  A0      : 0x800f8386  A1      : 0x3ffd56f0  
A2      : 0x3ffd59b8  A3      : 0x00000180  A4      : 0x00000002  A5      : 0x00000004  
A6      : 0x3ffc2fe4  A7      : 0x3ffc2f70  A8      : 0x000000c0  A9      : 0x00000001  
A10     : 0xffffffff  A11     : 0x3ffc2fec  A12     : 0x400e01e0  A13     : 0x00000005  
A14     : 0x00000000  A15     : 0x00000001  SAR     : 0x00000002  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000180  LBEG    : 0x400014fd  LEND    : 0x4000150d  LCOUNT  : 0xffffffff  

ELF file SHA256: 0000000000000000

Backtrace: 0x401f5e28:0x3ffd56f0 0x400f8383:0x3ffd5710 0x400df0d7:0x3ffd5730 0x400e02fd:0x3ffd5760 0x400f9776:0x3ffd5780 0x400f97e6:0x3ffd57a0 0x400e01f1:0x3ffd57c0 0x400d82ea:0x3ffd57e0 0x400fa621:0x3ffd5800 0x400e0544:0x3ffd5830 0x4012f169:0x3ffd5850

thebentern:
Output:

 ```
   0x400e01e0: recordCriticalError(_CriticalErrorCode, unsigned int, char const*) at /home/runner/work/Meshtastic-device/Meshtastic-device/src/mesh/NodeDB.cpp:669
[...]

PR #1876, opened 2022-10-30
Fixing a number of T-Beam poweroff display issues

thebentern:
This fixes some of the issues with the 1.3" OLED T-Beam add on screen:
1) Not turning off the screen during shutdown both PWR button long-press
2) Not turning off the screen during soft shutdown 
3) Not displaying the Shutting down screen upon receiving the shutdown admin message from an app

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -233,14 +233,14 @@ bool Power::setup()
 
 void Power::shutdown()
 {
-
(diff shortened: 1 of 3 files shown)


thebentern:
Fixes #1647. May be a bit flaky for AXP shutoff but we can adjust the runOnce timing for the power thread later.

PR #1878, opened 2022-10-31
Making meshtastic-dr-dev E22-900M22S/M30S work alternately

code8buster:
Reorganized my variant.h file so pins shared in common are at the top of the file, and the module-specific pins are in their own blocks.
Currently the M22S is initialized and can transmit.


code8buster:
It is failing a check relating to the Screen.h constructor in src/graphics, may be due to the inclusion of HAS_SCREEN 0 in the variant file. Not sure how to fix that, but I'd like to get this merged so I no longer have to deal with developing on my personal repo's master branch.

caveman99:
These stub constructors are a shim to make the compiler happy. the declaration is 'not clean' cause it doesn't matter anyway. it's indeed the NO_SCREEN setting. Trying to make cppcheck not compain any more.

PR #1884, opened 2022-11-02
Experimental DIY nrf52840 feather support

thebentern:


variants/feather_diy/platformio.ini
```diff
@@ -0,0 +1,12 @@
+; The very slick RAK wireless RAK 4631 / 4630 board - Unified firmware for 5005/19003, with or without OLED RAK 1921
+[env:feather_diy]
+extends = nrf52840_base
+board = adafruit_feather_nrf52840
+build_flags = ${nrf52840_base.build_flags} -Ivariants/feather_diy -D feather_diy
+build_src_filter = ${nrf52_base.build_src_filter} +<../variants/feather_diy>
+lib_deps = 
+  ${nrf52840_base.lib_deps}
+  melopero/Melopero RV3028@^1.1.0
```
caveman99, line 9: Just a pointer - that RV3028 (RTC) lib is RAK specific. Probably not needed as dependency here.

thebentern: Good point!

PR #1886, opened 2022-11-02
Don't allow arbitrary channel name for boundChannel

GUVWAF:
I could send an admin packet over the primary channel after I set a different name to that channel. 

This basically reverts [this commit](https://github.com/meshtastic/firmware/commit/c8aec324f527ec2367c091657f442a613bec3d50#diff-2c96c0d7355013e87dff13cd11e84479bff756bb1a4a2391780b563f6aa4293aR115). It was used as a workaround for getting the ExternalNotification module to work, but that should still work if setting the channel name to 'gpio' or if this checking is disabled after PR #1882 is merged.


thebentern:
lol `strlen(ch->settings.name) > 0`... that was not really too locked down

PR #1887, opened 2022-11-02
Portduino only: don't continue to try rebooting

GUVWAF:
This prevents the Linux application from trying to reboot while it can't.


thebentern:
Great catch!

PR #1899, opened 2022-11-04
fix: use RF95_IRQ for wakeup source in doLightSleep

weber-software:
This PR fixes the wakeup of RF95 based boards from light sleep by the RF95.

The configuration name of the IRQ from `RF95Configuration.h` pin is `RF95_IRQ` and not `RF95_IRQ_GPIO`.

Without the PR my device is missing most of the messages while in light sleep mode.

With this PR my device is able to receive the messages and successfully acknowledges them.


caveman99:
just out of curiousity - which device is this?

weber-software:
@caveman99  Its a T-Beam V1.1

PR #1935, opened 2022-11-12
Update NZ_865 to 36dBm

puzzled-pancake:
Updated NZ_865 to 36dBm as per:
https://rrf.rsm.govt.nz/smart-web/smart/page/-smart/domain/licence/LicenceSummary.wdk?id=219752

6dBW = 36dBm/4watt


caveman99 (commented):
13. Transmitters using e.i.r.p. greater than 0 dBW (1 W) must employ frequency hopping or digital modulation techniques.

We don't do frequency hopping, will LoRa qualify as digial modulation technique?

puzzled-pancake:
As I understand it LoRa does qualify as a digital modulation technique.

PR #1936, opened 2022-11-12
Update tbeam-s3 variant

lewisxhe:
Make the following updates according to the update of the board
1) Update the final t-beam-s3-core pin mapping
2) The initialization of SD card is added to FSCommon. Because there is no function definition for SD card in 'meshtastic', only initialization is added. HSPI is used
3) Change PCF8563 RTC from Wire to Wire1, PCF8563 shares I2C bus with PMU
4. Add ESP32S3 low-speed crystal initialization, which can effectively reduce RTC clock drift during deep sleep


src/main.cpp
```diff
@@ -246,8 +246,24 @@ void setup()
...
 
+#ifdef LILYGO_TBEAM_S3_CORE
+    // In T-Beam-S3-core, the I2C device cannot be scanned before power initialization, otherwise the device will be stuck
+    // PCF8563 RTC in tbeam-s3 uses Wire1 to share I2C bus
+    Wire1.beginTransmission(PCF8563_RTC);
+    if (Wire1.endTransmission() == 0){
+        rtc_found = PCF8563_RTC;
+        DEBUG_MSG("PCF8563 RTC found\n");
+    }
+#else
```
caveman99, line 257: The problem here is, we need to scan for a screen before the nodedb init. In the init we set the bluetooth pin to a fixed number if we don't have a working screen. Will the OLED be attached to a bus we can scan here already?

caveman99 (changes requested):
The PR looks good to me, can we find a solution for the screen scanning? If not, we need to alter the logic for the PIN determination.

src/main.cpp
lewisxhe: If this is the case, the PMU initialization needs to be placed before scanI2C device. Because the default power supply of tbeams3 display screen is off, the display screen power needs to be turned on first to scan the I2C device correctly

caveman99: as a PMU is usually soldered to the baseboard i don't see a problem there - the scan really is for finding peripheral devices like screens or sensors that can be attached or detached.

caveman99 (approved):

src/main.cpp
lewisxhe: No, if the I2C device is scanned when the power supply is not turned on, the esp32s3 will stay in the scanning method for a long time, which will take a very long time. It looks like the board is dead. Because the SDA and SCL are at a low level at this time, just like the bus is occupied. To ensure correct scanning, the power supply of the sensor must be turned on first

caveman99: I revised the entire I2c bus scanning mechanism in the develop branch. Now it will first turn on PMU if defined in the Board variant (the simple GPIO enable on T-Echo and the Chips on both t-beams) without looking for it actually being there. It will log the critical error if it can not be initialized, as it did before.

THEN the firmware will scan "Wire" and if defined "Wire1" for peripherals and remember on which bus it found it. Subsequent access will use the correct bus.

PR #1942, opened 2022-11-13
Develop to Master

thebentern:
@caveman99 are we good to merge upstream and cut a new alpha?


caveman99:
@thebentern a lot of wip stuff in develop, but nothing breaking. I'm not sure portduino works ATM. Had to leave after the last commit .. ;-)

PR #1947, opened 2022-11-15
testing issues with RAK11200/13300

caveman99:


thebentern:
@djjayballz just committed a pin change that might fix your issue. Please try it when you get a chance

thebentern:
@garthvh you have this hardware as well. Can you test these artifacts?

caveman99:
Problems with some boards still not fully fixed, but this board definition works better than the previous one. Also add a specific debug when it crashes/asserts.

Issue #252, opened 2020-07-04
Request for the introduction of a new function - Select radio frequency via GPS


frankenstein91:
I would like to use the function of an Internet independent messaging service while on vacation. But I don't want to have to install a new firmware every time, so I think it would be cool to introduce an automatism. But as I understand it, the hardware is bound to a country, nevertheless I try to find a solution here.

geeksville:
I agree.  At the very least I think we should stop putting differnet builds out for different frequency ranges.  It should be one build and you should be able to change frequency region from the device or the phone.

Though keep in mind: the 400ish mhz radios are physically different parts than the 900ish mhz radios.  So just changing the option won't work in that case.

frankenstein91:
sounds good to me

android606:
I'm looking at this one right now.
I'm thinking I can define a polygons with a list of GPS coordinates for each region. Then, with a little geometry, determine if the user's current GPS position is inside one of those polygons.

I'm not quite sure how to define the regions, but I have some ideas. The ITU specifies three regulatory regions:
![image](https://user-images.githubusercontent.com/797099/106187582-ac43af80-615a-11eb-8126-a1186daaec84.png)
(wikipedia)

The code currently has seven regions listed, but I think this map gives more specific information:
![image](https://user-images.githubusercontent.com/797099/106189337-047bb100-615d-11eb-9dcd-32694a8135b6.png)
(http://iotfactory.eu/lorawan-world-frequency-map-lora-sigfox-nbiot-telecom-operators-per-country/)

It really looks like seven regions is not quite sufficient. I think I'm going to have to specify more regions. Maybe the list we're using is outdated?

So, I could define some large regions and some small country-sized regions. I thought about using the ITU regions as the "fallback", but Russia/Kazakhstan/China sort of make that not work. I think the "Americas" region 2 is fine, but Regions 1 and 3 won't work as-is.

Then, when checking the user's location, check against the regions in order and stop if there's a match. We check the small regions first. If there is no match, I fall back on the larger ITU-ish world-size regions.

I was thinking of mapping regions like this:

// Americas
RegionCode_FG -> French Guiana (EU868 / Same rules as France) 
RegionCode_US -> ITU Region 2 (US915 / Americas)
// Europe / Africa
RegionCode_EU433 -> Sweden (EU433)
RegionCode_KZ -> Kazakhstan (EU433)
[...]

PR #1949, opened 2022-11-16
[t-beam-s3]Add an 8M partition table, modify the partition format to FAT, add usb msc initialization

lewisxhe:
The following improvements have been made:
1. Change `LitteFS` to  `FAT` to make it easier to view internal files when using mass storage
2. Add UF2 boot support, which is provided by @lyusupov. [#252](https://github.com/adafruit/tinyuf2/pull/252)
3. Added UF2 support. You can use the command to generate the firmware into UF2. Like t-echo, you can drag files to upgrade the internal firmware
    ```
    python uf2conv. py ..\.pio\build\tbeam-s3-core\firmware. bin -c -f 0xc47e5767 -b 0 -o firmware.uf2
    ```
4. Modify the t-beam-s3-core partition table to 8M


partition-table-8M.csv
```diff
@@ -0,0 +1,8 @@
+# ESP32S3 TBeamS3Core 8M Partition Table
+# Name,   Type, SubType, Offset,  Size, Flags
+nvs,      data, nvs,    0x009000,  20K,
+otadata,  data, ota,    0x00e000,  8K,
+ota_0,    0,    ota_0,  0x010000,  2048K,
+ota_1,    0,    ota_1,  0x210000,  2048K,
+uf2,      app,  factory,0x410000,  256K,
+ffat,     data, fat,    0x450000,  3776K,
```
caveman99, line 8: The OTA Partitions need to be spaced and named like on the smaller partition table. 
```
app,      app,  ota_0,   0x010000, 0x250000,
flashApp, app,  ota_1,   0x260000, 0x0A0000,
```
The uf2 parttition can then follow at 0x300000

and i am not sure about the change to FAT. When changing from 1.2 to 2.0 development we deliberately chose littlefs to minimize flash wear. We are discussing alternatie means to access file content on this partition through protobuf and HTTP interfaces. In addition, FAT would be the completey opposite to ALL other boards we support and would create a bunch of side issues and strain support.

src/FSCommon.h
```diff
@@ -23,13 +23,24 @@
...
+
+#if CONFIG_IDF_TARGET_ESP32S3
+// ESP32S3 version
+#include "FFat.h"
+#define FSCom FFat
+#define FSBegin() FSCom.begin(true)
+#define FILE_O_WRITE "w"
+#define FILE_O_READ  "r"
+#else
 // ESP32 version
```
caveman99, line 35: Again, i'd rather stick to littlefs for uniform layouts across supported platforms.

caveman99 (changes requested):

lyusupov:
@caveman99 
One of the key ideas of the Lewis proposal is to unify firmware and data management procedures on this (coming soon)  LilyGO T-Beam 'Supreme' (ESP32-S3) board. So different firmware files could operate on the product with little or no harm to each other and reduce hassles for the LilyGO tech. support & customer service.

It is based on the success of the LilyGO T-Echo product where users can 'drag-and-drop' different UF2 firmware files and they typically notice that previously installed firmware did not apply any breaking changes that prevent the next one to operate properly.

LilyGO have used the Adafruit technology (bootloader) on the nRF52840 based T-Echo to achieve that.
Adafruit is one of the leaders of the DIY electronic market that we can relay on.

Similar approach have been chosen for the new T-Beam as well. The partitions layout and the bootloader were inherited from Adafruit ones, they are all equal:
1) [partitions-8MB.csv in Adafruit TinyUF2](https://github.com/adafruit/tinyuf2/blob/master/ports/espressif/partitions-8MB.csv)
2) [partitions-8MB-tinyuf2 for Adafruit board in Arduino Core for ESP32](https://github.com/espressif/arduino-esp32/blob/master/variants/adafruit_feather_esp32s3_nopsram/partitions-8MB-tinyuf2.csv) 

Let's proceed with details of the layout:

1) Wi-Fi OTA regions 
```
ota_0,    0,    ota_0,  0x010000,  2048K,
ota_1,    0,    ota_1,  0x210000,  2048K,
```
Wi-Fi OTA regions are pretty ,much similar across most of the Espressif targets. Their names, types and equal size are prerequisite for Wi-Fi OTA over HTTP (WebUI) or LAN (espota.py) firmware update methods to operate properly ;

2) TinyUF2 second stage bootloader
```
uf2,      app,  factory,0x410000,  256K,
```
this region contains TinyUS2 bootloader binary. It is a prerequisite for the bootloader to operate properly.

3) FAT partition
```
ffat,     data, fat,    0x450000,  3776K,
```
This Adafruit partition is typically in use to store numerous files, such as CircuitPython programs, libraries and configuration files. 
The main advantage of the FAT FS use - when we connect the ESP32-S3 based device to a PC by means of a USB cable - this partition becomes mounted and accessible by Windows, Linux or macOS 'File Manager' application automatically. This is very handy for the user side.

lyusupov:
@caveman99 

> In addition, FAT would be the completely opposite to ALL other boards we support and would create a bunch of side issues and strain support. I'd rather stick to littlefs for uniform layouts across supported platforms.

ESP32-S3 is new platform. There are no users of Meshtastic on this platform at this time and (almost) no developers yet, right?
IMHO, there are little or no risks to give a way for the Lewis proposal (which stands on Adafruit experience), then see if it is good or not. Give it a chance for a half of a year or more. You can always get back to LittleFS afterwards.

lewisxhe:
Thank you @lyusupov for explaining this.
@caveman99 , we can try to follow the style of  `t-echo` and use the style of drag and drop update
 As @ lyusupov said, at least we can try. If the effect is not good, we can switch to littlefs, right.

garthvh:
We can use the adafruit UF2 without having a FAT filesystem.  It has been a lot of work to get the littlefs working on both platforms, there is quite a bit of risk of adding FAT to the mix with no testing.

lewisxhe:
> We can use the adafruit UF2 without having a FAT filesystem. It has been a lot of work to get the littlefs working on both platforms, there is quite a bit of risk of adding FAT to the mix with no testing.

OK. Now hang up the PR and decide whether to merge the PR according to the situation after the board reaches Meshtastic

thebentern:
I'm in full support of the uf2 flashing, but as @caveman99 mentioned, a huge part of the transtition to 2.0 was converting to littlefs across platforms for consistency and decreasing wear. This is neat but without a device to thoroughly test it, I'm not ready to integrate.

PR #1964, opened 2022-11-20
Init default mqtt configurations

thebentern:


sachaw:
Do people ever run MQTT on other ports? If so what do you think about adding it as a config field?

thebentern:
> Do people ever run MQTT on other ports? If so what do you think about adding it as a config field?

Currently if you use hostname:port for the mqtt.address, we parse it out of that.

PR #1974, opened 2022-11-23
variants\tbeam\variant.h: Use LORA_CS instead of RF95_NSS

IhorNehrutsa:
Tested Ok with module LILYGO Model:T-BEAM-V1.1


caveman99:
What difference does this PR make? The TLORA boards use three different chips now, so it doesn't really matter how you name the macros.

caveman99:
closing, no feedback regarding purpose.

PR #1987, opened 2022-11-25
Add t-echo to the operation after the flash operation fails

lewisxhe:
At present, it is only found on NRF52840 that this is a bug that has existed for a period of time. I found this problem a few months ago. Today, I again applied my t-echo to Meshtastic, and still found that the problem still exists. Here are my solutions

Although `FSCom.begin()` returns true. In some cases, the file cannot be opened. I try to open the file in `FSCom` After `begin()`, test the reading and writing of the file. Although the existence and writing of the file can be judged, when you call saveProto, you still respond to **Can't write prefs**,
At present, my method is that when the firmware is written for the first time, meshtastic will call saveProto twice to save the default configuration settings. If both writes fail, format should be called to format the flash memory

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -481,7 +481,7 @@ bool saveProto(const char *filename, size_t protoSize, size_t objSize, const pb_
         if(failedCounter >= 2){
             FSCom.format();
             //After formatting, the device needs to be restarted
-            nodeDB.resetRadioConfig();
+            nodeDB.resetRadioConfig(true);
         }
 #endif
     }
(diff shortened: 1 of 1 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -475,6 +475,15 @@ bool saveProto(const char *filename, size_t protoSize, size_t objSize, const pb_
...
     } else {
         DEBUG_MSG("Can't write prefs\n");
+#ifdef ARCH_NRF52
+        static uint8_t failedCounter = 0;
+        failedCounter++;
+        if(failedCounter >= 2){
+            FSCom.format();
+            //After formatting, the device needs to be restarted
+            nodeDB.resetRadioConfig();
+        }
```
caveman99, line 485: resetRadioConfig will only reboot the device if called as resetRadioConfig(true) inducing a factory reset. You can just use `rebootAtMsec = millis() + (5 * 1000);` instead since on a freshly formatted filesystem the default config will be written on next boot anyway.

caveman99 (changes requested):
Good Catch, the other versions of Littlefs we use will auto-format the partition if no valid filesystem is detected.

src/mesh/NodeDB.cpp
lewisxhe: Changes have been made

PR #1990, opened 2022-11-25
Send 'ACK' response for handled admin message

GUVWAF:
With this, it would suffice to set only wantResponse for an admin packet and not wantAck. This in order to avoid confusing ACKs sent by the Routing Module even when the device could not handle the admin packet. 

Tested with a modified CLI (setting only wantResponse) for both the local node and a remote node. Upon a response, the device internally generates an ACK and informs the client. In this way, if you don't get an ACK or it is a NAK with e.g. 'NOT_AUTHORIZED' error, you know the admin packet could not be handled (or did not even arrive).

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -176,6 +176,12 @@ bool AdminModule::handleReceivedProtobuf(const MeshPacket &mp, AdminMessage *r)
         }
         break;
     }
+    
+    // If asked for a response and it is not yet set, generate an 'ACK' response
+    if (mp.decoded.want_response && !myReply) {
+        myReply = allocErrorResponse(Routing_Error_NONE, &mp);
+    }
+
     return handled;
 }
 
--- a/src/modules/AdminModule.h
+++ b/src/modules/AdminModule.h
@@ -2,7 +2,7 @@
 #include "ProtobufModule.h"
 
 /**
- * Routing module for router control messages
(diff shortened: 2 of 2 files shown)


garthvh:
I am getting a continuous stream of this when I flash the PR artifact 

rst:0x10 (RTCWDT_RTC_RESET),boot:0x17 (SPI_FAST_FLASH_BOOT)
invalid header: 0x42410058
invalid header: 0x42410058
invalid header: 0x42410058
invalid header: 0x42410058
invalid header: 0x42410058
ets Jul 29 2019 12:21:46

GUVWAF:
I can flash a T-Beam and RAK successfully with this artifact. Seems like something went wrong with flashing on your side. If you used device-update, maybe it needs a device-install. Or in case of NRF you need to do a factory reset first?

caveman99:
@garth how did you flash it? we changed generation of firmware files a few days back.

for installs flash the firmware*.bin to offset 0x00 and for udates the firmware*-update.bin to offset 0x10000 - the batch/sh scripts were updated with the new targets but you must remember to specify the -update.bin file for the device_update instead of the main firmware.

garthvh:
@caveman99 using the install.sh script in the PR artifact, just tried again same thing

garthvh (approved):
Works fine with the iOS admin messages that currently wantack, was able to change to wantresponse and get an ack, no double acks in messaging

PR #1992, opened 2022-11-26
Fix the format of t-echo Bluetooth pairing display page

lewisxhe:
The effect of the repair page is as shown in the figure below, and tbeam and techo have been tested

image


thebentern:
Very nice! This one has been bothering me for a bit, but I never got around to fixing it.

PR #1998, opened 2022-11-26
Completely refrain from transmitting if duty cycle limit exceeded

GUVWAF:
With this, the device will completely refrain from transmitting if it is violating the hourly duty cycle. If it happens, it informs the client by generating a NAK with DUTY_CYCLE_LIMIT error (needs updated protobufs). Did some stress testing with the interactive simulator and then it works fine :) 

Still have some doubts. I added a LoRaConfig bool to override it, but am not sure if we should always set this to true if you set the user to is_licensed (for HAMs). Furthermore, do we need to inform the client how many minutes we have to be silent before we can transmit again? Then we need an admin message for that I think.

### DIFF
--- a/src/airtime.cpp
+++ b/src/airtime.cpp
@@ -117,6 +117,20 @@ float AirTime::utilizationTXPercent()
     return (float(sum) / float(MS_IN_HOUR)) * 100;
 }
 
+// Get the amount of minutes we have to be silent before we can send again
+uint8_t AirTime::getSilentMinutes(float txPercent, float dutyCycle) 
+{  
+  float newTxPercent = txPercent;
+  for (int8_t i = MINUTES_IN_HOUR-1; i >= 0; --i) {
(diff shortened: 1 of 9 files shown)


GUVWAF:
Duty cycle limitation will now automatically be overridden when user is set to is_licensed. 
For now I think getting the DUTY_CYCLE_LIMIT error will suffice. Tested it with a T-Beam and the Android app by setting the duty cycle to 0.01%. Got a cloud with strike when it got exceeded. A feature request for the apps would then be to show some pop-up telling what happened (and maybe the option to override it).

caveman99 (approved):
LGTM - the generated files will be overwritten on proto regen... So we need to update that first.

sachaw:
There will likely need to be scope for being able to bypass the duty cycle restrictions and have encryption enabled at the same time.

GUVWAF:
There is a separate LoRaConfig field 'override_duty_cycle'. It will be automatically set to true if you set the user to is_licensed, but it can also be manually set.

PR #2019, opened 2022-12-04
Adds a flag to turn the GPS power rail off entirely on tbeam

code8buster:
I realized I could probably put the flag into ButtonThread and do away with the GPS #include, but I figure at some point we might want to make use of it in the GPS class.


thebentern:
I think before we integrate this, we need a specific frame in screen as a display queue that it's happening. Otherwise it could cause confusion if triggered inadvertantly when a user attempts to quickly toggle to other frames on the screen

code8buster:
Makes sense to have some acknowledgement of it, in the app and the display. Garth also suggested that the position setting might perform the same function. If I get to it, should that be in a separate pull request?

As is, this PR's changes are gated behind a commented line in the tbeam/variant.h file. Someone who knows what they're doing will have to compile their own, rather than it being included in automated builds.

code8buster:
This is simply for switching during operation, with a new event in the ButtonThread. The function added to sleep.cpp could be reused by the config proto. I believe I would need to change the initializer if the status of the GPS chip was variable on-boot with respect to that config; currently it just assumes the GPS is always powered on after boot.

caveman99:
Doing that would cater for both use cases. remote unattended operation and handheld device. When pulling the power line for the chaip it doesn't matter if it is inited or not, we just have to keep record of the state it should be in.

src/gps/GPS.h
```diff
@@ -128,14 +131,14 @@ class GPS : private concurrency::OSThread
...
 
-  private:
     /// Prepare the GPS for the cpu entering deep or light sleep, expect to be gone for at least 100s of msecs
     /// always returns 0 to indicate okay to sleep
     int prepareSleep(void *unused);
 
     /// Prepare the GPS for the cpu entering deep sleep, expect to be gone for at least 100s of msecs
     /// always returns 0 to indicate okay to sleep
     int prepareDeepSleep(void *unused);
+  private:
```
caveman99, line 141: Hesitant here to move them out of private, can you somehow do that via friend class? That will expose these routines only where they're needed.

variants/tbeam/platformio.ini
```diff
@@ -6,4 +6,5 @@ lib_deps =
   ${esp32_base.lib_deps}
 build_flags = 
   ${esp32_base.build_flags} -D TBEAM_V10  -I variants/tbeam
+  ;-DGPS_POWER_TOGGLE ;uncomment this line to allow a double press on the user button to turn off gps entirely.
```
caveman99, line 9: I think we can uncomment this for the commit. It's a good thing to have and its tbeam only.

src/ButtonThread.h
```diff
@@ -159,9 +159,21 @@ class ButtonThread : public concurrency::OSThread
...
+    #endif
+    #if defined(GPS_POWER_TOGGLE)
+        if(config.position.gps_enabled)
+        {
+        DEBUG_MSG("Flag set to false for gps power\n");
+        } 
+        else 
+        {
+        DEBUG_MSG("Flag set to true to restore power\n");
+        }
```
caveman99, line 173: can you add a new string/branch at https://github.com/meshtastic/firmware/blob/de6b752db8874270cd1e82a7ed905e0361afc2ba/src/graphics/Screen.cpp#L533 and https://github.com/meshtastic/firmware/blob/de6b752db8874270cd1e82a7ed905e0361afc2ba/src/graphics/Screen.cpp#L471 to indicate that condition visually? 'disabled by button' or the like.

caveman99 (changes requested):

code8buster:
I think what I've done won't break anything, or change things too drastically. I believe the change in click timing mostly prevents one from accidentally activating the feature while cycling the screen. It can still happen, but you'd have to be clicking right on the ragged edge of the (perceptible) screen update time.

caveman99 (approved):

garthvh (approved):
The faster speed seems better on both the tbeam and rak with rotary, rotary still advances the screen (which also happens already) I do think we should still add and enhancement that takes you to the screen that tells you what happened but that is also a separate issue.

caveman99:
@garthvh the rotary issue is because when the inputbroker is active it overrides the button long/double press behaviour. ideally we should combine the inpoutbroker and the onebutton.

PR #2031, opened 2022-12-09
Update MQTT.cpp

arduionoGP:
Added a third valid JSON envelope to send via MQTT to device. NodeINFO to send User data that triggers User responses. The new type is "senduser".  A valid envelope looks like:
{
    "sender": "Broker",
    "type": "senduser",
    "payload": {
        "id": "!7efec08c",
        "long_name": "Meshtastic c08c",
        "short_name": "c08c",
    }
}

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -81,6 +81,25 @@ void MQTT::onPublish(char *topic, byte *payload, unsigned int length)
                 } else {
                     DEBUG_MSG("JSON Ignoring downlink message we originally sent.\n");
                 }
+             } else if ((json.find("sender") != json.end()) && (json.find("payload") != json.end()) && (json.find("type") != json.end()) && json["type"]->IsString() && (json["type"]->AsString().compare("senduser") == 0)) {
+                //invent the "senduser" type for a valid envelope
+                if (json["payload"]->IsObject() && json["type"]->IsString() && (json["sender"]->AsString().compare(owner.id) != 0)) {
+                    JSONObject usr;
+                    usr=json["payload"]->AsObject(); //get nested JSON User
+                    User u = User_init_default;
+                    strcpy(u.id, usr["id"]->AsString().c_str());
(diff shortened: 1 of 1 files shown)


garthvh (commented):
There are several fields missing
* macaddr
* hw_model
* is_licensed

caveman99:
While this uses the same sematic as sendtext and sendposition, sending a user record is not the same thing, as it is not content but metadata. This essentially allows to inject 'fake' or 'phantom' devices into the mesh. Can you elaborate a bit on the use-case, i am hesitant to support that.

garthvh:
> While this uses the same sematic as sendtext and sendposition, sending a user record is not the same thing, as it is not content but metadata. This essentially allows to inject 'fake' or 'phantom' devices into the mesh. Can you elaborate a bit on the use-case, i am hesitant to support that.

Good point, there is also never a case where a user exists separate from a node info on the mesh, sending in user packets is likely to break apps

arduionoGP:
I left those fields out on the publish side because you also left them out in the decoding side code (the "Case" statements toward the end) and I wanted the code to look consistent.

For a use case, consider two geographically separate meshes, each with their own gateway to their own mqtt server. (Each MQTT server allows for the store/forward functionality which is built into MQTT.) The two servers can only intermittently communicate because internet is down and sat com is not continuous. Mobile assets in both meshes can come and go depending on range and need to be geo-plotted and able to send messages bridging both areas. For transparent bridging it seemed to me that this would require text, pos, and user portnums to both upload and download.

garthvh:
User never exists without a node, why would mqtt need to send one that is not part of a real node?

garthvh:
This will not do anything, there is no routing for a user only packet

caveman99:
Just a few musings on this. the JSON facility is meant to get data from the mesh easily, and provide a few rudimentary functions to send data to it. Bridging was never the scope of JSON/MQTT. The other MQTT-Topic transports all neccessary protobufs to do a bridging setup. If you ant to do more fancy stuff than feed data into NodeRed, for the love of god use the protobufs!

PR #2039, opened 2022-12-13
Enable GPS capabilities on heltec V3

ghostop14:
The heltec V3 variant had HAS_GPS set to disable GPS capabilities.  I've tested with this enabled and using GPIO 2 and 3 via config params.

### DIFF
--- a/variants/heltec_v3/variant.h
+++ b/variants/heltec_v3/variant.h
@@ -1,7 +1,5 @@
 #define LED_PIN LED
 
-#define HAS_GPS 0
-
 #define RESET_OLED RST_OLED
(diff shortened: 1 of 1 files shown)


caveman99 (approved):
Looks good to me

@thebentern on the next alpha we should do that for all devices that don't have a GPS chip soldered on. Remove the RX/TX defines from the device file and document the proto settings instead. nRF52 does not apply as that uses the hw serial anyway and ignores these.

geeksville:

https://meshtastic.discourse.group/t/heltec-esp32-lora-v3-gps/7024/2

Issue #2022, opened 2022-12-05
[Bug]: Recording critical error 11 at src/mesh/RadioLibInterface.cpp:389


D4rk4:
### Category

Other

### Hardware

Rak4631

### Firmware Version

2.0.6

### Description

Same conditions like in https://github.com/meshtastic/firmware/issues/2021

How to repeat: Node connected to a power source (solar,USB), batteries not fully charged (90-95%). Chatty LORA.

HW setup: RAK4631 on RAK19007|RAK19003, RAK1906, w/wo GPS module

### Relevant log output

```Shell
10:15:25 1202 [RadioIf] (bw=250, sf=11, cr=4/8) packet symLen=8 ms, payloadSize=57, time 1042 ms
10:15:25 1202 [RadioIf] Lora RX (id=0x0fc1b46a fr=0xd4 to=0xff, WantAck=0, HopLim=3 Ch=0x8 encrypted rxSNR=6.5 rxRSSI=1.53492e-154)
10:15:25 1202 [RadioIf] AirTime - Packet received : 1042ms
10:15:25 1202 [Router] Found existing packet record for fr=0x25053ad4,to=0xffffffff,id=0xfc1b46a
10:15:25 1202 [Router] Resending implicit ack for a repeated floodmsg
10:15:25 1202 [Router] enqueuing for send (id=0x0fc1b46a fr=0xd4 to=0xff, WantAck=0, HopLim=2 Ch=0x8 encrypted rxSNR=6.5 rxRSSI=1.53492e-154)
10:15:25 1202 [Router] txGood=178,rxGood=359,rxBad=8
10:15:25 1202 [Router] rx_snr found. hop_limit:2 rx_snr:6.500000
10:15:25 1202 [Router] rx_snr found in packet. Setting tx delay:1428
10:15:25 1202 [Router] Found existing packet record for fr=0x25053ad4,to=0xffffffff,id=0xfc1b46a
10:15:25 1202 [Router] Add packet record (id=0x0fc1b46a fr=0xd4 to=0xff, WantAck=0, HopLim=3 Ch=0x8 encrypted rxSNR=6.5 rxRSSI=1.53492e-154)
10:15:25 1202 [Router] Ignoring incoming msg, because we've already seen it (id=0x0fc1b46a fr=0xd4 to=0xff, WantAck=0, HopLim=3 Ch=0x8 encrypted rxSNR=6.5 rxRSSI=1.53492e-154)
[...]

PR #2040, opened 2022-12-14
Disable deep sleep for nRF  and force shutdown for RAK463x on discharged battery

D4rk4:
Fix for https://github.com/meshtastic/firmware/issues/2022


caveman99:
I'm curious, is this really triggering for you? it wants 10 samples below min_battery and if one of those is above it will start counting over at zero. If it is triggering there may be some underlying problem in the ADC routines.

D4rk4:
@caveman99 yes it's shit my node =(
Please give me time for sanity validation of this fix in different conditions.

caveman99:
Sure, i converted your PR to a draft, let's get to the bottom of this.

caveman99 (approved):
We need to investigate the measurements further, this will improve user experience for the time being.

PR #2058, opened 2022-12-22
[modules][serial] switch simple mode from ascii to bytes

### LINKED ISSUE #2057: [Feature Request]: Switch Serial Mode simple mode from ASCII to just bytes
### Platform

NRF52, ESP32

### Description

Right now function selected in Serial Module to handle payload are specifically designed for ASCII characters and will not handle correctly any other bytes.

Idea is to switch this mode to pass-through any bytes - whether it be ASCII or binary data. From customer perspective nothing should change as if customer supplied ASCII characters in the past - they will still get them as they will be sent as-is. However with proposed change it will be possible to send any kind of data.

lesykm:
fixes #2057 - copied description:

Right now function selected in Serial Module to handle payload are specifically designed for ASCII characters and will not handle correctly any other bytes.

Idea is to switch this mode to pass-through any bytes - whether it be ASCII or binary data. From customer perspective nothing should change as if customer supplied ASCII characters in the past - they will still get them as they will be sent as-is. However with proposed change it will be possible to send any kind of data.

Also TEXTMSG is untouched - it still can only process ASCII characters.

### DIFF
--- a/src/modules/SerialModule.cpp
+++ b/src/modules/SerialModule.cpp
@@ -49,7 +49,6 @@
 #define RXD2 16
 #define TXD2 17
 #define RX_BUFFER 128
-#define STRING_MAX Constants_DATA_PAYLOAD_LEN
 #define TIMEOUT 250
 #define BAUD 38400
 #define ACK 1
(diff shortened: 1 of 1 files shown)


src/modules/SerialModule.cpp
```diff
@@ -49,7 +49,6 @@
 #define RXD2 16
 #define TXD2 17
 #define RX_BUFFER 128
-#define STRING_MAX Constants_DATA_PAYLOAD_LEN
```
lesykm, line 52: This is just not used anywhere so just deleted it.

garthvh:
Seems like a good plan

caveman99:
Good call, i'll merge that. Code looks good. Can you please file any future PR's against the develop branch? the idea was to use the master only for bugfixing (resp. hotfixing) and add new/changed features to develop to be able to test them without interfering with the 'bugfix' branch.

lesykm:
Will use develop branch further, sorry for the mess!
I have couple of coming PRs - one for improving of serial module and another one for a new platform :)

caveman99:
No sweat, any input appreciated ;-)

PR #2060, opened 2022-12-22
ProtobufAPI in SerialModule

### LINKED ISSUE #1204: 2.0 Serial Module Feature Request: MeshAPI over UART via protos mode
Expose the client API over a (compile time configurable) serial interface. Either HW Serial Number or RX/TX Pins. This way another Microcontroller can access the API and facilitate the creation of a standalone device.

The Combination RAK5005 Base/RAK4631 Core/RAK2305 IO would be a first platform to implement this on, nRF52 core for meshtastic and an ESP32 for the application, doing screen and an I2C keyboard. 

Apart from that, this would be useful for any integration with existing applications, microcontroller or more high-level hardware.

caveman99:
will resolve #1204

### DIFF
--- a/src/SerialConsole.cpp
+++ b/src/SerialConsole.cpp
@@ -25,7 +25,7 @@ void consolePrintf(const char *format, ...)
 #endif
 }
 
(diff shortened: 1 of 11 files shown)


garthvh:
Very nice!

PR #2062, opened 2022-12-24
About TBEAM-S3-CORE OLED

lewisxhe:
TBEAM-S3-CORE OLED cannot accurately use the [oled_probe] method to obtain the exact type. According to the latest batch of OLEDs, its model is SH1106, but the [oled_probe] returned SSD1306. LilyGo has sent you samples, and you can verify them after receiving them. This request is a reminder. If you have a better method, please close this request.


caveman99:
Thank you for the submission. Is the display pre-soldered like on the T-Lora devices or will it be an option? If it is pre-soldered, you can simply `#define USE_SH1106` in the variant.h file for the board to override the auto-detect. If it is an option, the override can be set via our device CLI if it is detected wrong. 

`meshtastic --set display.oled OLED_SH1106`

lewisxhe:
> Thank you for the submission. Is the display pre-soldered like on the T-Lora devices or will it be an option? If it is pre-soldered, you can simply `#define USE_SH1106` in the variant.h file for the board to override the auto-detect. If it is an option, the override can be set via our device CLI if it is detected wrong.
> 
> `meshtastic --set display.oled OLED_SH1106`

yes. It is equipped with 1.3 inch SH1106 driver chip OLED display by default. I will resubmit new changes. Thank you for your suggestions

PR #924, opened 2021-11-23
ACK retried want_ack packets #919

rxt1077:
This PR addresses Issue #919 by adding some code to `ReliableRouter::shouldFilterReceived` that ACKs retried packets that would normally just be filtered with the call to `FloodingRouter::shouldFilterReceived` at the end.

I'll be the first to point out that this isn't great in that it requires decoding the packet to have a channel to respond on meaning the shouldFilterRecieved inherited functions all go from `cost MeshPacket *` to `MeshPacket *`. I'd prefer that a filter function doesn't modify the thing passed to it, but I didn't think it warranted making a copy of the packet and there's already some interesting ACK logic in the function anyway.

For what it's worth, I wholeheartedly agree with [this comment](https://github.com/meshtastic/Meshtastic-device/blob/bc7fc1285ddb11da651bb0d137dc00cb6a19ea9a/src/mesh/Router.h#L88).

PR #2069, opened 2022-12-28
Flood ACKs on DM back to original sender

GUVWAF:
This changes the hopLimit for real ACKs (only sent when receiving a DM) from 0 to the configured one, like normal packets. This way the ACK will be flooded back to the original sender, such that can be guaranteed that the packet arrived at the intended receiver. 
When a client now gets an ACK, if the ‘from’ field is your nodeNum, then it is an implicit ACK (meaning a rebroadcast on the mesh), if it is the nodeNum of the intended receiver, successful reception is confirmed. 

Based on what I saw using the simulator, I now removed resending ACKs for repeated incoming messages (introduced in #924), as this may happen due to flooding and then you will send another ACK each time. Since the ACKs are now flooded back, there is already some redundancy for lost ACKs.


src/mesh/FloodingRouter.cpp
```diff
@@ -34,7 +34,8 @@ void FloodingRouter::sniffReceived(const MeshPacket *p, const Routing *c)
         // do not flood direct message that is ACKed 
         DEBUG_MSG("Receiving an ACK not for me, but don't need to rebroadcast this direct message anymore.\n");
         Router::cancelSending(p->to, p->decoded.request_id);   // cancel rebroadcast for this DM
-    } else if ((p->to != getNodeNum()) && (p->hop_limit > 0) && (getFrom(p) != getNodeNum())) {
+    } 
+    if ((p->to != getNodeNum()) && (p->hop_limit > 0) && (getFrom(p) != getNodeNum())) {
```
GUVWAF, line 38: Needed as otherwise ACKs will not be flooded.

src/mesh/FloodingRouter.cpp
```diff
@@ -17,7 +17,7 @@ ErrorCode FloodingRouter::send(MeshPacket *p)
     return Router::send(p);
 }
 
-bool FloodingRouter::shouldFilterReceived(MeshPacket *p)
+bool FloodingRouter::shouldFilterReceived(const MeshPacket *p)
```
GUVWAF, line 20: 'const' was removed in #924, but this can now be reverted, since I removed ACK-ing repeated incoming messages.

thebentern (approved):

PR #2074, opened 2022-12-29
a lot of thread housekeeping

caveman99:
Switch them off when not needed / disabled.

### DIFF
--- a/src/modules/SerialModule.cpp
+++ b/src/modules/SerialModule.cpp
@@ -221,9 +221,7 @@ int32_t SerialModule::runOnce()
         return (10);
     } else {
         DEBUG_MSG("Serial Module Disabled\n");
-
-        enabled = false;
-        return INT32_MAX;
+        return disable();
     }
 }
 
--- a/src/modules/SerialModule.cpp
+++ b/src/modules/SerialModule.cpp
@@ -220,7 +220,6 @@ int32_t SerialModule::runOnce()
 
         return (10);
     } else {
-        DEBUG_MSG("Serial Module Disabled\n");
         return disable();
     }
 }
--- a/src/modules/SerialModule.cpp
+++ b/src/modules/SerialModule.cpp
@@ -222,6 +222,7 @@ int32_t SerialModule::runOnce()
     } else {
         DEBUG_MSG("Serial Module Disabled\n");
 
+        enabled = false;
         return INT32_MAX;
     }
 }
@@ -303,9 +304,6 @@ ProcessMessage SerialModuleRadio::handleReceived(const MeshPacket &mp)
                 }
             }
         }
-
-    } else {
-        DEBUG_MSG("Serial Module Disabled\n");
     }
     return ProcessMessage::CONTINUE; // Let others look at this message also if they want
 }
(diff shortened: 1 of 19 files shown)


thebentern:
@caveman99 I replaced those instances with disable and I also made it return an int32_t so that we could return it in a runOnce

caveman99:
Every time we call this, same leak, 360 bytes...
`[NodeInfoModule] Thread NodeInfoModule leaked heap 90348 -> 89988 (-360)`

Also this, but only sometimes, still same amount of bytes when it happens
`[RadioIf] Thread RadioIf leaked heap 88224 -> 87900 (-324)`

src/modules/SerialModule.cpp
```diff
@@ -221,8 +221,7 @@ int32_t SerialModule::runOnce()
         return (10);
     } else {
         DEBUG_MSG("Serial Module Disabled\n");
-
-        return INT32_MAX;
+        return disable();
```
caveman99, line 224: I think these "We dont want that" noisy bits can go ...

src/modules/esp32/AudioModule.cpp
```diff
@@ -259,7 +260,7 @@ int32_t AudioModule::runOnce()
         return 100;
     } else {
         DEBUG_MSG("Audio Module Disabled\n");
-        return INT32_MAX;
+        return disable();
```
caveman99, line 263: Same here, this was good in development, but not any more

caveman99:
> Also this, but only sometimes, still same amount of bytes when it happens `[RadioIf] Thread RadioIf leaked heap 88224 -> 87900 (-324)`

the 324 may be harmless... Router frees that up again, seems to be a packet struct...

```
??:??:?? 16 [RadioIf] (bw=250, sf=11, cr=4/8) packet symLen=8 ms, payloadSize=27, time 690 ms
??:??:?? 16 [RadioIf] Lora RX (id=0x7cf711c6 fr=0x24 to=0xff, WantAck=0, HopLim=2 Ch=0x8 encrypted rxSNR=10.5 rxRSSI=1.84221)
??:??:?? 16 [RadioIf] AirTime - Packet received : 690ms
??:??:?? 16 [RadioIf] --- --- Thread RadioIf leaked heap 120040 -> 119716 (-324)
??:??:?? 16 [Router] Rx someone rebroadcasting for us (id=0x7cf711c6 fr=0x24 to=0xff, WantAck=0, HopLim=2 Ch=0x8 encrypted rxSNR=10.5 rxRSSI=1.84221)
??:??:?? 16 [Router] didn't find pending packet
??:??:?? 16 [Router] Found existing packet record for fr=0x25d63124,to=0xffffffff,id=0x7cf711c6
??:??:?? 16 [Router] Found existing packet record for fr=0x25d63124,to=0xffffffff,id=0x7cf711c6
??:??:?? 16 [Router] Add packet record (id=0x7cf711c6 fr=0x24 to=0xff, WantAck=0, HopLim=2 Ch=0x8 encrypted rxSNR=10.5 rxRSSI=1.84221)
??:??:?? 16 [Router] Ignoring incoming msg, because we've already seen it (id=0x7cf711c6 fr=0x24 to=0xff, WantAck=0, HopLim=2 Ch=0x8 encrypted rxSNR=10.5 rxRSSI=1.84221)
??:??:?? 16 [Router] Incoming message was filtered 0x25d63124
??:??:?? 16 [Router] +++ +++ Thread Router freed heap 119716 -> 120040 (324)
```

PR #2077, opened 2022-12-29
add RemoteHardware (and the missing S&F config)

caveman99:
plus make the config boundaries automatic

### DIFF
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -171,7 +171,7 @@ size_t PhoneAPI::getFromRadio(uint8_t *buf)
         // Advance when we have sent all of our Channels
         if (config_state >= MAX_NUM_CHANNELS) {
             state = STATE_SEND_CONFIG;
-            config_state = Config_device_tag;
+            config_state = _AdminMessage_ConfigType_MIN + 1;
         }
         break;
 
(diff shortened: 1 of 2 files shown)


src/mesh/PhoneAPI.cpp
```diff
@@ -253,11 +257,15 @@ size_t PhoneAPI::getFromRadio(uint8_t *buf)
...
+        case ModuleConfig_remote_hardware_tag:
+            fromRadioScratch.moduleConfig.which_payload_variant = ModuleConfig_remote_hardware_tag;
+            fromRadioScratch.moduleConfig.payload_variant.remote_hardware = moduleConfig.remote_hardware;
+            break;
         }
 
         config_state++;
         // Advance when we have sent all of our ModuleConfig objects
-        if (config_state > ModuleConfig_audio_tag) {
+        if (config_state > (_AdminMessage_ModuleConfigType_MAX + 1)) {
```
thebentern, line 268: Oh man, we should have been doing that all along and we missed it!

thebentern (approved):

mc-hamster:
mind blown!

src/mesh/PhoneAPI.cpp
caveman99: Well it breaks if there's a gap in the numbering.

PR #2081, opened 2022-12-30
Created more structured enterprisey logging with levels

thebentern:


caveman99:
Can we add a legacy DEBUG_LOG with a log level of FIXME? :-) will help in the transition...

Issue #2105, opened 2023-01-07
[Meta]: find and triage remaining asserts in the codebase


caveman99:
Asserts are like pulling a handbrake mid corner. Handle them gracefully instead and allow the firmware to continue. This meta issue keeps track of the remaining asserts in the code.

- [x] mesh/Channels.cpp:194:    assert(chIndex < channelFile.channels_count); // This should be equal to MAX_NUM_CHANNELS
- [x] mesh/PhoneAPI.cpp:302:        assert(0); // unexpected state - FIXME, make an error code and reboot
- [x] mesh/PhoneAPI.cpp:372:        assert(0); // unexpected state - FIXME, make an error code and reboot
- [x] mesh/RadioInterface.cpp:167:    assert(p->which_payload_variant == MeshPacket_encrypted_tag); // It should have already been encoded by now
- [x] mesh/RadioInterface.cpp:176:    assert(slotTimeMsec); // Better be non zero
- [x] mesh/RadioInterface.cpp:273:    assert(sizeof(PacketHeader) == 16); // make sure the compiler did what we expected
- [x] mesh/RadioInterface.cpp:403:            assert(0); // Unknown enum
- [x] mesh/RadioInterface.cpp:425:    assert(myRegion); // Should have been found in init
- [x] mesh/RadioInterface.cpp:492:    assert(!sendingPacket);
- [x] mesh/RadioInterface.cpp:495:    assert(p->which_payload_variant == MeshPacket_encrypted_tag); // It should have already been encoded by now
- [x] mesh/RadioInterface.cpp:505:    assert(p->hop_limit <= HOP_MAX);
- [x] mesh/RadioInterface.cpp:509:    assert(h->from);
- [x] mesh/RadioLibInterface.cpp:231:                        assert(txp);
- [x] mesh/RadioLibInterface.cpp:244:            assert(0); // We expected to receive a valid notification from the ISR
- [x] mesh/RadioLibInterface.cpp:362:                assert(HOP_MAX <= PACKET_FLAGS_HOP_MASK); // If hopmax changes, carefully check this code
- [x] mesh/RadioLibInterface.cpp:369:                assert(((uint32_t)payloadLen) <= sizeof(mp->encrypted.bytes));
- [x] mesh/ReliableRouter.cpp:151:        assert(numErased == 1);
- [x] mesh/ReliableRouter.cpp:226:    assert(iface);
[...]

PR #2106, opened 2023-01-07
initial cleanup work

caveman99:
fixes for https://github.com/meshtastic/firmware/issues/2105 go here


src/mesh/Channels.cpp
```diff
@@ -191,9 +191,20 @@ void Channels::onConfigChanged()
...
+    } else {
+        LOG_ERROR("Invalid channel index %d > %d, malformed packet received?\n", chIndex , channelFile.channels_count);
+
+        Channel *ch = (Channel *)malloc(sizeof(Channel));
+        memset(ch, 0, sizeof(Channel));
+        // ch.index -1 means we don't know the channel locally and need to look it up by settings.name
+        // not sure this is handled right everywhere
+        ch->index = -1;
+        return *ch;
+    }
```
caveman99, line 207: Not entirely sure this is a valid fix. this assert is frequently triggered when 1.2 packets find their way into the system. This has DoS Potential, so returning a new empty struct with ch-index = -1 would be a good thing in my mind (and it's documented like that in the proto comments for the channel struct). The function is used all over the place so i'm not sure every spot handles that return gracefully. This definiely needs testing and some more insight.

caveman99:
did anybody tet these artefacts except me? they were running here all night on a test net with a lot of devices at 40%+ util. No crashes recorded...

PR #2112, opened 2023-01-08
Add Chunked File Transfer to API

caveman99:


caveman99:
here's how it should work in theory...

download a file:
Send an XModem_Control_STX with sequence number 0 and the filename as payload. The Node will answer with XModem_Control_ACK (ok) or XModem_Control_NAK (file not found). Now send an XModem_Control_ACK yourself and the node will start sending chunks. Ack each chunk till you geht an XModem_Control_EOF. Each packet has a sequence number and a CRC16 checksum you need to validate. if something goes wrong, send a NAK and the node will send the last packet again, up to 25 times. If you want to cancel a transfer, send XModem_Control_CAN. You shoud abort a transfer when you receive one.

upload a file: 
Send an XModem_Control_SOH with sequence number 0 and the filename as payload. After receiving the XModem_Control_ACK you can send the first chunk of the file, with sequence number 1 and a computed CRC16. If you receive an XModem_Control_ACK, send the next chunk, for a XModem_Control_NAK the same chunk again. If you sent the last chunk and it was acked send a XModem_Control_EOF.

src/xmodem.cpp
```diff
@@ -0,0 +1,203 @@
...
+    if (crc == tcrc)
+        return 1;
+    else
+        return 0;
+}
+
+void XModemAdapter::sendControl(XModem_Control c) {
+    memset(xmodemStore, 0, XModem_size);
+    xmodemStore->control = c;
+}
```
caveman99, line 69: Does anybody have a better idea than just copying stuff into xmodemStore and waiting for the PhoneAPI to find it? ToPhone queue seems to work the same way...

sachaw (approved):

Issue #2109, opened 2023-01-07
[Bug]: MQTT messages sent to secondary channels are always posted in primary channel 


Kollensperger:
### Category

Other

### Hardware

Rak4631

### Firmware Version

2.0.8

### Description

When sending a correctly formatted mqtt message to the primary channel, it is displayed correctly and as expected in the primary channel of the iOS app. When sending a message to any secondary channel with mqtt upload and download enabled, the same message does not arrive in the expected secondary channel on iOS, but always in the primary channel. 
I have confirmed this behaviour for several secondary channels, and observed the messages in the correct topic using MQTTX.

Example: 

Sending a message to mqtt node msh/2/json/Primary/ works as expected: In MQTTX, it appears in msh/2/json/Primary/ and on the iOS app it also appears in the primary channel. So far so good. Now I change the mqtt node to msh/2/json/Secondary/, and as expected, in the MQTTX browser the same messages appear in the channel msh/2/json/Secondary/, but on the iOS app, the messages still appear on msh/2/json/Primary/. In my view since the same message displays fine in the primary channel, and MQTTX is picking up the messages in the correct secondary channel, that would indicate something is wrong in the meshtastic plumbing?

### Relevant log output

```Shell
These are some bits and bobs that may be relevant from the meshtastic node (RAK4631) connected to iOS. This is connected to a t-beam which is connected to the internet. This is a message being sent on the channel "Vær" which lives in the topic msh/2/json/Vær/ but appears in the primary channel. I have also tried secondary channels without the letter "æ", such as "weather"

22:22:26 6130 [Router] Add packet record (id=0x05d1c3de fr=0xdc to=0xff, WantAck=0, HopLim=3 Ch=0x92 encrypted rxSNR=6 rxRSSI=1.53492e-154)
22:22:26 6130 [Router] recentPackets size=73
22:22:26 6130 [Router] recentPackets size=23 (after clearing expired packets)
22:22:26 6130 [Router] Using channel 0 (hash 0x92)
22:22:26 6130 [Router] Using AES256 key!
22:22:26 6130 [Router] Software encrypt fr=d1c801dc, num=5d1c3de, numBytes=184!
22:22:26 6130 [Router] decoded message (id=0x05d1c3de fr=0xdc to=0xff, WantAck=0, HopLim=3 Ch=0x0 Portnum=1 rxtime=1673130146 rxSNR=6 rxRSSI=1.53492e-154)
[...]

PR #2113, opened 2023-01-09
Find the downlink channel to post packet to.

caveman99:
- getByName was not used anywhere in the code so that change in `src/mesh/Channels.cpp` is safe.
- Please TEST this as i don't have a working Json setup in mqtt at the moment.
- potential fix for #2109


src/mqtt/MQTT.cpp
```diff
@@ -43,9 +43,20 @@ void MQTT::onPublish(char *topic, byte *payload, unsigned int length)
...
         if (json_value != NULL) {
             LOG_INFO("JSON Received on MQTT, parsing..\n");
+
             // check if it is a valid envelope
             JSONObject json;
             json = json_value->AsObject();
+
+            // parse the channel name from the topic string
+            char *ptr = strtok(topic, "/");
+            for (int i = 0; i < 3; i++) {
```
caveman99, line 53: strtok alters the topic string by replacing the / with \0 but it's not used after the operation. I hope it will be reseeded by the next call to onPublish.

PR #2116, opened 2023-01-09
RAK-4631 TX power output fix

thebentern:
Until we figure out what's going on


caveman99 (approved):

SR7673:
I confirm, now the output power is 20 dBm when set to 20 dBm. In 2.0.12 it rather was -7 dBm

Issue #2117, opened 2023-01-10
[Bug]: mqtt hostnames can't always be <32 characters


bmidgley:
### Category

Other

### Hardware

T-Beam

### Firmware Version

2.0.6

### Description

Sometimes hostnames we get for stuff in the cloud are really long and not negotiable. But we're pegged at 31 characters currently.

D4rk4:
Some users in the Russian telegram chat mentioned this issue before.
As a result of this limitation, we can't use free MQTT servers with random domain names for example HiveMQ.

garthvh:
This has been bumped to 64

xhci:
@garthvh by default, my ActiveMQ hostname on AWS is 67 characters (not including the port). It requires significantly more cost like a network load balancer to use custom hostnames. Would it be possible to bump this higher?

garthvh:
> @garthvh by default, my ActiveMQ hostname on AWS is 67 characters (not including the port). It requires significantly more cost like a network load balancer to use custom hostnames. Would it be possible to bump this higher?
> 
> I added an MR here that will probably fix the issue: #5604

This can't be merged (and would have to be made in the protobufs repo) There is no space left on the MQTT config packet.

xhci:
Thanks for the feedback. I'll figure out a solution on my end.

bmidgley:
We made a short cname in our case to avoid the long cloud hostname

PR #2118, opened 2023-01-10
Extend mqtt address length

bmidgley:
https://github.com/meshtastic/firmware/issues/2117


caveman99:
Sorry, this change will be overwritten with the next protobuf import. These fields come from the protobuf repository (hence generated) and the PR needs to be placed there. Also please check if any holding vatiables or temporary buffers are big enough for this

joshpirihi:
This will also require a corresponding update to the protobufs (module_config.options) to allow a longer string to be configured and stored

joshpirihi:
the protobuf file in question is module_config.options

PR #2119, opened 2023-01-10
Turn of 3.3v rail on RAK-4631 on shutdown

thebentern:


caveman99 (approved):

charminULTRA:
@thebentern I use the 3.3v rail for LED power so I'm trying to determine if expected behavior will be impacted for my devices without changing my settings. Tried reading docs but under what conditions does the RAK go into deepSleep? 2 hours under what condition?

PR #2130, opened 2023-01-12
Add 868 MHz Ukrainian Band

Krezalis:

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -100,18 +100,24 @@ const RegionInfo regions[] = {
     */
 
     RDEF(LORA_24, 2400.0f, 2483.5f, 100, 0, 10, true, false, true),
-
+    
+    /*
+        433,05-434,7 Mhz 10 mW
+        https://nkrzi.gov.ua/images/upload/256/5810/PDF_UUZ_19_01_2016.pdf
+    */
+    RDEF(UA_433, 433.0f, 434.0f, 100, 0, 12, true, false, false),
+    
+    /*
+        868,0-868,6 Mhz 25 mW
+        https://nkrzi.gov.ua/images/upload/256/5810/PDF_UUZ_19_01_2016.pdf
+    */
+    RDEF(UA_868, 868.1f, 868.5f, 100, 0, 14, true, false, false),
+    
     /*
         This needs to be last. Same as US.
     */
     RDEF(UNSET, 902.0f, 928.0f, 100, 0, 30, true, false, false)
         
-    /*
-        868,0-868,6 Mhz 25 mW
-        https://nkrzi.gov.ua/images/upload/256/5810/PDF_UUZ_19_01_2016.pdf
-    */
-    RDEF(UA868, 868.1f, 868.5f, 100, 0, 20, true, false, false)
-
 };
 
 const RegionInfo *myRegion;
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -100,24 +100,24 @@ const RegionInfo regions[] = {
     */
 
     RDEF(LORA_24, 2400.0f, 2483.5f, 100, 0, 10, true, false, true),
-    
+
     /*
         433,05-434,7 Mhz 10 mW
         https://nkrzi.gov.ua/images/upload/256/5810/PDF_UUZ_19_01_2016.pdf
     */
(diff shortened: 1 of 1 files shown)


src/mesh/RadioInterface.cpp
```diff
@@ -105,6 +105,12 @@ const RegionInfo regions[] = {
         This needs to be last. Same as US.
     */
     RDEF(UNSET, 902.0f, 928.0f, 100, 0, 30, true, false, false)
+        
+    /*
+        868,0-868,6 Mhz 25 mW
+        https://nkrzi.gov.ua/images/upload/256/5810/PDF_UUZ_19_01_2016.pdf
+    */
+    RDEF(UA868, 868.1f, 868.5f, 100, 0, 20, true, false, false)
```
caveman99, line 113: Please add the name of the definition to https://github.com/meshtastic/protobufs/blob/e00b5ba7d06053d820f1123351881fc4fa9270d1/config.proto#L506 or else you won't be able to use it. Also please add an underscore in front of the frequency designator to come in line with the other definitions.

Suggestion: UA_868

caveman99 (changes requested):

caveman99:
Also: if you don't have another lora range in your country (except the worldwide Lora_24) you can omit the 868 and simply call the region UA

src/mesh/RadioInterface.cpp
```diff
@@ -105,6 +105,12 @@ const RegionInfo regions[] = {
         This needs to be last. Same as US.
     */
     RDEF(UNSET, 902.0f, 928.0f, 100, 0, 30, true, false, false)
+
```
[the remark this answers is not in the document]
Krezalis: Current regulation https://nkrzi.gov.ua/images/upload/256/5810/PDF_UUZ_19_01_2016.pdf

Krezalis:
https://github.com/meshtastic/protobufs/pull/271

src/mesh/RadioInterface.cpp
```diff
@@ -100,12 +100,24 @@ const RegionInfo regions[] = {
     */
 
     RDEF(LORA_24, 2400.0f, 2483.5f, 100, 0, 10, true, false, true),
-
+    
+    /*
+        433,05-434,7 Mhz 10 mW
+        https://nkrzi.gov.ua/images/upload/256/5810/PDF_UUZ_19_01_2016.pdf
+    */
+    RDEF(UA_433, 433.0f, 434.0f, 100, 0, 12, true, false, false),
```
caveman99, line 108: Page 121 states for РІ 42-2: "Робочий цикл менше 10 %" (Duty Cacle not more than 10%) This makes it the same definition as EU_433 (ETSI) and can be omitted.

src/mesh/RadioInterface.cpp
```diff
@@ -100,12 +100,24 @@ const RegionInfo regions[] = {
...
+    
+    /*
+        868,0-868,6 Mhz 25 mW
+        https://nkrzi.gov.ua/images/upload/256/5810/PDF_UUZ_19_01_2016.pdf
+    */
+    RDEF(UA_868, 868.1f, 868.5f, 100, 0, 14, true, false, false),
+
```
caveman99: Sorry, but this is not correct. "Робочий цикл менше 1 %. Робочий цикл може бути більше 1% за умови
використання режиму контролю спектра LBT (режим прослуховування перед
включенням передавача) та широкосмуговою модуляцією з розширенням
спектру методом стрибкоподібної зміни частоти" 

We are not employing frequency hopping or spread spectrum techniques, although we use LBT. So this allocation is restricted to 1% duty cycle.

(P.S. you can define the true borders of the spectrum and don't need to keep a guard distance, the firmware takes care of that. so you can really say it's for 868,0f to 868,6f

caveman99: You may be better off using the EU_868 allocation, it's harmonized by ETSI to use 10% instead of 1% duty cycle.

caveman99 (changes requested):

src/mesh/RadioInterface.cpp
Krezalis: extended freq to 434.7

src/mesh/RadioInterface.cpp
Krezalis: Hm, та = & so if it use only LBT its not be 100% its bad. Ok, for official build it set 1%.

caveman99 (approved):

PR #2158, opened 2023-01-17
Turn GPS off on RAK and maybe others

### LINKED ISSUE #2155: [Feature Request]: Toggle 3v3_S power line via IO2 for RAK devices
### Platform

NRF52

### Description

Enable software toggle of 3v3_S power, via IO2 high/low settings to enable/disable power for RAK-based devices.

The intent is to toggle power to the RAK1910 or RAK12500 GPS modules; adding this feature would allow users to get the maximum runtime from installed batteries, when sending real-time GPS updates is not required or desired.

Users would need to be made aware that disabling 3v3_S power cuts off power _**ALL**_ sensors using Slots A-D for RAK19003/19007/5005 and Slots A-F for RAK19001.

code8buster:
Fixes #2155 
Should do what it says on the tin, cleanly shut down the GPS thread/unit (plus everything else on the 3v3 bus, in the case of RAK4361)

Don't have one to test, but setting position_enable=0 ought to work.

### DIFF
--- a/src/sleep.cpp
+++ b/src/sleep.cpp
@@ -29,7 +29,8 @@ Observable<void *> preflightSleep;
 
 /// Called to tell observers we are now entering sleep and you should prepare.  Must return 0
 /// notifySleep will be called for light or deep sleep, notifyDeepSleep is only called for deep sleep
-/// notifyGPSSleep will be called when config.position.gps_enabled is set to 0 or from buttonthread when GPS_POWER_TOGGLE is enabled.
+/// notifyGPSSleep will be called when config.position.gps_enabled is set to 0 or from buttonthread when GPS_POWER_TOGGLE is
+/// enabled.
 Observable<void *> notifySleep, notifyDeepSleep;
 Observable<void *> notifyGPSSleep;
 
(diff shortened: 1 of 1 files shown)


caveman99:
i think you need to check if the macro is not negative. 

`variants/Dongle_nRF52840-pca10059-v1/variant.h:#define PIN_GPS_EN     (-1)`

garthvh:
This seems like a problematic setting for most handset use cases, I don't think users are going to expect that everything other than the core gets powered off, should this be constrained to routers only?

code8buster:
I'd be happy to scrap the PR and go back to the drawing board, or just gate it behind some other conditionals. There will probably be a board at some point which uses PIN_GPS_EN that doesn't also shut off all other peripherals.

garthvh:
> I'd be happy to scrap the PR and go back to the drawing board, or just gate it behind some other conditionals. There will probably be a board at some point which uses PIN_GPS_EN that doesn't also shut off all other peripherals.

I would definitely prefer that it is more targeted, that is too much text to ensure that a user sees it before enabling a setting and then losing the ability to use the one button we have.  I am seeing weird behavior from the tbeam double click as well https://github.com/meshtastic/firmware/issues/2159 and would like to be careful with something that can dramatically alter device functionality so much.  I am not totally convinced that turning off the user button should ever happen as a handset feature.

tropho23:
Thanks for the effort but this causes all sort of trouble: disabled serial module, bootloops, no way to set region due to lack of serial connection...but mostly the bootloop.  I think a hardware switch that shorts IO2 to GND is the best way to go for this.   Thanks again for giving this a try :)

I also recommend we close my issue #2155 and I will comment there as well.

PR #2179, opened 2023-01-19
Fix vulnerability with "h.from == 0"

nukevoid:
altered packed with "from == 0" can do Remote Node Administration without permission

### DIFF
--- a/src/mesh/RadioLibInterface.cpp
+++ b/src/mesh/RadioLibInterface.cpp
@@ -347,7 +347,11 @@ QueueStatus RadioLibInterface::getQueueStatus()
                 airTime->logAirtime(RX_ALL_LOG, xmitMsec);
             } else {
                 const PacketHeader *h = (PacketHeader *)radiobuf;
-
+                // altered packed with "from == 0" can do Remote Node Administration without permission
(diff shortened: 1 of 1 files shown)


src/mesh/RadioLibInterface.cpp
```diff
@@ -347,7 +347,11 @@ QueueStatus RadioLibInterface::getQueueStatus()
                 airTime->logAirtime(RX_ALL_LOG, xmitMsec);
             } else {
                 const PacketHeader *h = (PacketHeader *)radiobuf;
-
+                // altered packed with "from == 0" can do Remote Node Administration without permission
+                if (h->from == 0) {
+                    LOG_WARN("ignoring received packet without sender\n");
+                    return;
+                }
                 rxGood++;
```
GUVWAF, line 355: Although this packet should not be handled, it would be good to put the if statement after `rxGood++`, to keep the statistics of good receptions correct.

GUVWAF: Thanks. Just an additional note that this will normally not go wrong because a device will never pick a NodeNum < 4, because these are reserved. So, I will merge this.

PR #2180, opened 2023-01-20
Automatically disable WIFI if no connection can be made in 2 minutes.

zjs81:
I edited 
the mesh/http/wifiapclient.cpp to make it so it will turn off wifi after 120 seconds of failed connections. This will make it less annoying when you lose or enter a bad wifi password/network.


zjs81 (commented):
.

zjs81 (commented):
Fix Typo in comment

D4rk4:
It's not a good idea to disable wifi in this case, better set this timeout via config.
For example, we got a power outage, power is back, MCU is booted, home access point still not (some enterprise devices or software AP may boot 5+ minutes)

caveman99:
I second that comment, this is not a good idea. My home router takes about 5 minutes for a firmware update and wifi is down in that timeframe. In addition the current Wifi stack is not very reliable with modern hardware, i've seen cases where the link goes down and recovers 20 minutes later. 

On the other hand - what's really annoying about an ESP device trying to conenct to wifi, apart from the debug log?

mc-hamster:
@zjs81 I too am curious about this. Why is this an issue and what problem does this cause? We'd like to get a root cause on it.

zjs81:
Hey Guys, Its really annoying because it disables Bluetooth so the only way to turn off wifi if you set it to the wrong SSD is to either create a hotspot on your phone with the name and password that you set and turn off wifi that way. Or do it through the serial port. Otherwise, there is no way to connect to the device.

mc-hamster:
@zjs81 Good feedback! The root cause here is that we can only have a single instance of the PhoneAPI at any one time and so we tie that to the interface that's being exposed.

@thebentern @caveman99 Let's have a focused discussion on PhoneAPI soon. This will solve other problems too. This is the "Nasty Bug" that Kevin brought up about a year and a half ago.

PR #2185, opened 2023-01-21
GPS acquisiton on boot only

code8buster:
On GPS enabled boards this adds a new feature when fixed position is enabled:
The GPS chip will remain active until ~1.5 minutes after it gets a valid location. After that, the chip is turned off. 
This would be useful for those who want a hands-off experience of setting a node up on a high point for later retrieval.

The timer is based on the thread awake cycle, and a counter that runs each runOnce. Reason for this delay is I noticed the altitude would be very wrong until a couple more satellites came into view, which takes a little more time. I would like to have done this without adding another variable to GPS.h but didn't really think of a good way to work with the timekeeping functions.
This can remain a draft until it is either accepted as is or a better method is proposed. I realized trunk linting has made my changes less readable so I will point out the lines.

GPS.cpp
298, 405, 504-511

GPS.h
172


caveman99:
looks good to me. Just pointing out, we had a recent addition to OSThreads. they have a disable function now. So if you are positively sure you never need to run this thread again, just call disable(); from within.

PR #2198, opened 2023-01-24
show timestamp for received messages

folkertvanheusden:
When a message is received and displayed, include the recevieve-timestamp. So now it functions as a clock and due to the 900s default screen-lock-timeout you still have an idea of when the message was received.

This is just a suggestion, please feel free to ignore.


sachaw:
Device screen

src/graphics/Screen.cpp
```diff
@@ -379,10 +379,15 @@ static void drawTextMessageFrame(OLEDDisplay *display, OLEDDisplayUiState *state
         display->fillRect(0 + x, 0 + y, x + display->getWidth(), y + FONT_HEIGHT_SMALL);
         display->setColor(BLACK);
     }
-    display->drawStringf(0 + x, 0 + y, tempBuf, "From: %s", (node && node->has_user) ? node->user.short_name : "???");
-    if (config.display.heading_bold) {
-        display->drawStringf(1 + x, 0 + y, tempBuf, "From: %s", (node && node->has_user) ? node->user.short_name : "???");
+
+    if(config.display.heading_bold) {
+        display->drawStringf(1 + x, 0 + y, tempBuf, "           From: %s", (node && node->has_user) ? node->user.short_name : "???");
```
caveman99, line 384: Please use the inserstion on both print statements, heading_bold works by printing the same text twice with a 1 pixed offset. The way you set it up the timestamp is not in bold print.

caveman99 (changes requested):

src/graphics/Screen.cpp
folkertvanheusden: Done in commit 65c0b8e.

caveman99 (approved):

tropho23:
Very nice, but adding a timezone offset for screens would be great.  All logs, etc. should be kept at GMT for universal understanding but user-facing times such as screens should ideally allow the user to choose a timezone offset to display local time.

andrekir:
one solution is using relative time, same as "time ago" format in the nodes last heard timestamp

garthvh:
> one solution is using relative time, same as "time ago" format in the nodes last heard timestamp

Relative time is a much better option than managing time zones and DST

PR #2215, opened 2023-01-28
Wio-E5 WIP

GUVWAF:
Uses the new RadioLib module for STM32WLx. This compiles and the radio initializes, but there's no working file system yet. If you hardcode a region (right now I put EU_868 in there), it can communicate with other Meshtastic devices, but other than that it's not really usable.


thebentern:
Closing. I picked out some of these changes onto an updated branch

PR #2224, opened 2023-01-29
Added modes for rebroadcast

thebentern:
Hopefully I got it in the right spot for the local_only cancellation of rebroadcasts

### DIFF
--- a/src/mesh/FloodingRouter.cpp
+++ b/src/mesh/FloodingRouter.cpp
@@ -60,6 +60,9 @@ void FloodingRouter::sniffReceived(const meshtastic_MeshPacket *p, const meshtas
         }
     }
 
+    if (config.device.rebroadcast_mode == meshtastic_Config_DeviceConfig_RebroadcastMode_LOCAL_ONLY) {
+        Router::cancelSending(p->to, p->decoded.request_id);
+    }
     // handle the packet as normal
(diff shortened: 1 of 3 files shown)


src/mesh/FloodingRouter.cpp
```diff
@@ -60,6 +60,10 @@ void FloodingRouter::sniffReceived(const meshtastic_MeshPacket *p, const meshtas
         }
     }
 
+    if (config.device.rebroadcast_mode == meshtastic_Config_DeviceConfig_RebroadcastMode_LOCAL_ONLY) {
```
GUVWAF, line 63: I think `LOCAL_ONLY` isn’t working correctly like this. Firstly, it will only end up at the clause when it didn’t go to `Router::send()` before. Secondly, cancelSending is only doing something when there is already a packet with that ID in the queue (in the first clause at line 34 a rebroadcast is canceled when in the meantime it already heard an ACK). Right now I think you will see 

> “cancelSending id=xxx, removed=0”

meaning nothing is canceled. 

I think we should just set `encryptedOk` in the RoutingModule to false to ensure we don’t rebroadcast encrypted packets. Then, if you configure channels on a repeater with a different PSK than the default, we already also ignore packets from ‘open’ networks, right?

PR #2230, opened 2023-01-30
User configurable center frequency

caveman99:


thebentern (approved):

PhilipLykov:
Could you please specify how to use it? Since "meshtastic_windows.exe --set loraConfig.override_frequency 430.4" doesn't find such atribute.

PhilipLykov:
here also no such attribute:
https://buf.build/meshtastic/protobufs/docs/main:meshtastic#meshtastic.Config.LoRaConfig

PR #2245, opened 2023-02-02
potential fix for Range Test Leak. Poking around in the dark.

caveman99:


karamo:
Does not work.
I have localized it in
```
void RangeTestModuleRadio::sendPayload(NodeNum dest, bool wantReplies)
{
    meshtastic_MeshPacket *rangetestPacket = allocDataPacket();
```

caveman99:
This change does not improve the original bug but cleans up code residue ...

Issue #2242, opened 2023-02-02
[Bug]: RAK 4631 with ethernet crashes on MQTT message reception


Kollensperger:
### Category

Other

### Hardware

Rak4631

### Firmware Version

2.0.14 and 2.0.17

### Description

Rak 4631 (both on 5005 and 19007 baseboards) with ethernet and MQTT crashes after a short while. This is a known good working config on a t-beam, but instead of wifi enabled, ethernet is enabled in the RAK 4631.
Impossible to send more than 1 MQTT message to the board.
Interestingly enough, even though the message (see log output below) "Testing" was clearly received by the ethernet enabled RAK, the device it was sent from shows both this message and any subsequent messages as "Max retransmission Reached" in iOS.

on occasion the final line in the terminal output is 

`WARNING file:stream_interface.py __reader line:179 Meshtastic serial port disconnected, disconnecting... read failed: [Errno 6] Device not configured`

but this is not always the case. 
This is reproducible on both 2.0.14.2baaad8 and 2.0.17.5d1c06b.
On a separate note it seems the ethernet cable cannot be particularly long - 30 cm long cable yields lit amber and green control lights on the RAK, 1 m long cable yields error message `ERROR | ??:??:?? 67 Ethernet cable is not connected`.

EDIT: I used two different routers, a mobile router and a google mesh router. Whilst the mesh router clearly has access to the internet via ethernet, it won't light up the lights and connect to the RAK with what ever length cable, whereas the mobile router will. So NOT a cable length issue

### Relevant log output

```Shell
//\ E S H T /\ S T / C

DEBUG | ??:??:?? 2 Filesystem files:
WARNING file:mesh_interface.py _sendPacket line:389 Not sending packet because protocol use is disabled by noProto
DEBUG | ??:??:?? 2  channels.proto (252 Bytes)
DEBUG | ??:??:?? 2  config.proto (75 Bytes)
DEBUG | ??:??:?? 2  module.proto (63 Bytes)
DEBUG | ??:??:?? 2  db.proto (316 Bytes)
DEBUG | ??:??:?? 2 Using analog input 5 for battery level
INFO  | ??:??:?? 2 No I2C devices found
INFO  | ??:??:?? 2 Meshtastic hwvendor=9, swver=2.0.14.2baaad8-d
DEBUG | ??:??:?? 2 Reset reason: 0x0
DEBUG | ??:??:?? 2 Setting random seed 2993911210
[...]

PR #2246, opened 2023-02-03
tryfix for #2242

caveman99:


caveman99:
This fixes the display issue but not the crash. Still a good PR :-)

PR #2066, opened 2022-12-25
[meshtastic][board] add Happymodel ES900TX board

lesykm:
!!!DO NOT ACCEPT YET, POC, PROTOBUFS PR INCOMING UPON DISCUSSION FINISHED

fixes #2065

PR #2249, opened 2023-02-04
Add Hardware: BetaFPV 900 Nano TX

MisterRager:
Add New Hardware Type: [BetaFPV Nano TX Module (900)](https://betafpv.com/products/elrs-nano-tx-module)

This device is pretty similar to the recent BetaFPV 2.4GHz hardware but is smaller, has no joystick, lacks a screen, and has no battery input. It still has a more than adequate set of radios, though, for Meshtastic, so here it is!

Waiting on some Protobufs:
https://github.com/meshtastic/protobufs/pull/307


caveman99:
we have an open PR https://github.com/meshtastic/firmware/pull/2066 for a similar module with a PA. Maybe these 2 can be combined? They seem rather similar.

MisterRager:
I could see about just adding the whole lot of ELRS transmitter modules if you like - they have well formed descriptions over on the ELRS side:

https://github.com/ExpressLRS/ExpressLRS/blob/master/src/hardware/TX/

markbirss:
> I could see about just adding the whole lot of ELRS transmitter modules if you like - they have well formed descriptions over on the ELRS side:
> 
> https://github.com/ExpressLRS/ExpressLRS/blob/master/src/hardware/TX/
> 
> I just have a limited selection of modules at hand right now - a few BetaFPV nanos (900, 2.4), the Namimno full-size (2.4) and some STM32-based ones (900) that would be a bigger lift and probably a better fit for the "UART modem" idea I've talked about.

The header files here include more detail if the device has any included features (according to ExpressLRS fw)

https://github.com/BETAFPV/ExpressLRS/tree/master/src/include/target

`// Any device features
#define USE_TX_BACKPACK
#define USE_SX1280_DCDC
#define WS2812_IS_GRB`

One of the targets for example has a SPI OLED (meshtastic currently only supports i2c oled as far an i know)

I have just received the a HappyModel_ES24TX_Pro_Series_2400_TX and has LED_WS2812 on a different pin than the BETAFPV_2400_TX

The BETAFPB is more "can i use it for meshtastic friendly" as you just only have to change dip switches and flash, no need to also open as with the HappyModel

https://www.expresslrs.org/assets/images/dipswitch-Pro.png

caveman99:
Let me put it another way: the point of the firmware repository is not to collect "all the" board definitions, but to add value through a new board, e.g. if the design is unique, very economical, very powerful, very inexpensive to build. That's the opposite of 'i can add definitions for their whole product range'.

It boils down to: WHY do we wanna have this device, that is not really a good form factor to work with, included in ou repository? The PA is a USP for sure, but that's about it. But then again i am not the sole gatekeeper of the firmware and just voicing my opinion. Would be nice if others did so too.

MisterRager:
My motivation is that it's a pretty easy to obtain piece of kit with a power amplifier that comes with a mildly directional antenna that I've got redundancies on. I think they'd make good router nodes where the screen isn't as useful, and I'm looking at putting together something with one of the [micro-nano adapters](https://betafpv.com/products/micro-nano-module-adapter) to provide power and a possible GPS.

It would be nice, though, to be able to just add support for more hardware without flooding the repository with board defs we need to keep up with.

MisterRager:
Also, radiolib doesn't really support the rfm95 amplifier settings well.

garthvh:
How is this board actually working? Does it stay up a day? I have only seen people state that it boots, which is not really enough to justify merging new hardware in.

MisterRager:
It is fully stable - no issues experienced with uptime like with my t-beam

PR #2253, opened 2023-02-04
Set ham mode admin message

thebentern:
For convenience for clients setting up licensed usage and consistency in behavior.

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -106,6 +106,10 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
         else
             handleSetChannel(r->set_channel);
         break;
+    case meshtastic_AdminMessage_set_ham_mode_tag:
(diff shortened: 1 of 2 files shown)


src/modules/AdminModule.cpp
```diff
@@ -598,6 +602,18 @@ void AdminModule::saveChanges(int saveWhat, bool shouldReboot)
...
 }
 
+void AdminModule::handleSetHamMode(const meshtastic_HamParameters &p)
+{
+    strncpy(owner.long_name, p.call_sign, sizeof(owner.long_name));
+    owner.is_licensed = true;
+    config.lora.override_duty_cycle = true;
+    config.lora.tx_power = p.tx_power;
+    config.lora.override_frequency = p.frequency;
+
```
caveman99, line 612: I assume the unset fields will result in unset config settings? e.g. if i don't want to specify a frequency or tx_power.

caveman99 (approved):

PR #2261, opened 2023-02-07
Make Portduino build again

GUVWAF:
Not sure what the implications of using -fPIC are.


GUVWAF:
Never mind, still failing in the Docker container.

PR #2262, opened 2023-02-07
Next try fix Portduino

GUVWAF:


caveman99:
very bad net connection today, but from recherching this, i think you need to add --no-pic rather than PIC...

GUVWAF:
It shows that the build succeeds, but it the Dockerfile pulls master and then it still fails.

caveman99:
yea cause the change is not in master yet

PR #2266, opened 2023-02-08
Ham mode should broadcast in plaintext and nodeinfo every 10 minutes

thebentern:


m7xedhoward:
I believe (for amateurs in the UK) it doesn't need to be plain text as along as the encoding is publicly available. Is that correct?  Does that apply all other regions?

caveman99:
> I believe (for amateurs in the UK) it doesn't need to be plain text as along as the encoding is publicly available. Is that correct? Does that apply all other regions?

This is not about encoding, the payload is encrypted with AES128 by default. We need to switch that functionality off, even for the UK.

src/modules/AdminModule.cpp
```diff
@@ -609,9 +609,20 @@ void AdminModule::handleSetHamMode(const meshtastic_HamParameters &p)
     config.lora.override_duty_cycle = true;
     config.lora.tx_power = p.tx_power;
     config.lora.override_frequency = p.frequency;
+    // Set node info broadcast interval to 10 minutes
+    // For FCC minimum call-sign announcement
+    config.device.node_info_broadcast_secs = 600;
+
+    // Remove PSK of primary channel for plaintext amateur usage
+    auto primaryChannel = channels.getByIndex(channels.getPrimaryIndex());
```
caveman99, line 617: grafik

caveman99 (changes requested):

caveman99:
What issue? These settings have been around for a long time, all this PR does is make it easier under the hood to enable them. For Ham operation the entire package content sent out needs to be plaintext readable. Even encryption with a default key is not permitted, let alone a custom key. Only way to use this legal as a ham is to turn encryption completely off, and this PR does exactly that.

From the UK i know that digital modes like Free-DV or DVB-S have been a point of discussion for a long time, as some saw the content as encrypted. I think this is what you are referring to, our Protobufs are public and the Transmissions are standardized, BUT there's additional real encryption which is not permitted and which we disable.

src/modules/AdminModule.cpp
m7xedhoward: If each option is available as a individual config option, encryption (on/default/off), tx power,  frequency,  duty. I don't to understand why a ham mode is required..

m7xedhoward:
Indeed i was thinking of such examples.  after futher thought you are correct. Best to avoid all encryption, public or not for ham use.

src/modules/AdminModule.cpp
mc-hamster: It shows that we are putting an effort to comply with regional requirements for licensed use.

PR #2272, opened 2023-02-11
Enable DeviceTelemetry on Portduino

GUVWAF:
I added a `HAS_SENSOR` flag to separate DeviceTelemetry from EnvironmentTelemetry such that we can enable it on Portduino. 

Furthermore, I had to put `SimRadio` in a separate thread such that I can use `notifyLater` when transmitting, to fix issues with the packetPool when it was using `ccToPhone` for DeviceTelemetry.

I also guarded the simulator stuff with `HAS_RADIO` such that we can distinguish between a Portduino target with radio and the simulator.

Then I found syslog causes Portduino to crash because it doesn't compile _mesh/http_, so I disabled that with an extra `!defined(ARCH_PORTDUINO)`.

### DIFF
--- a/arch/portduino/portduino.ini
+++ b/arch/portduino/portduino.ini
@@ -12,7 +12,9 @@ build_src_filter =
   -<mesh/http/>
   -<mesh/eth/>
   -<modules/esp32>
-  -<modules/Telemetry>
+  -<modules/Telemetry/EnvironmentTelemetry.cpp>
+  -<modules/Telemetry/AirQualityTelemetry.cpp>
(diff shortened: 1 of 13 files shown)


caveman99 (approved):

caveman99:
Good stuff, the syslog problem may be related to the PIC PIE error message. Since portduino has IP by default, it would be a shame to have to exclude that. Let's keep in mind we have tofix that underlying problem eventually.

caveman99 (approved):

thebentern (approved):

PR #2279, opened 2023-02-16
Preamble change

thebentern:
Worth a try?


src/mesh/RF95Interface.cpp
```diff
@@ -67,19 +67,12 @@ bool RF95Interface::init()
 #endif
     setTransmitEnable(false);
 
-    int res = lora->begin(getFreq(), bw, sf, cr, syncWord, power, currentLimit, preambleLength);
+    int res = lora->begin(getFreq(), bw, sf, cr, syncWord, power, preambleLength);
```
thebentern, line 70: Ooooooh. That was inisidious! 
This is why I love named parameters and hate C/C++ for not having them XD

thebentern:
Tested between SX1276 and 1262 devices and I got reliable transmissions, acks etc. Will throw to the alphanauts for more comprehensive testing

D4rk4:
@thebentern does its break compatibility with earlier firmware? What is the motivation for this change?

thebentern:
> @thebentern does its break compatibility with earlier firmware? What is the motivation for this change?

It should not break backwards compatibility. I tested it on my devices with older versions and they behaved fine. 
The reason for the preamble length change is to decrease the airtime utilization. The preamble's purpose is to alert lora transceivers in range that a mesh packet transmission will be following. We want to keep that as short as possible for efficiency and speed.

PR #2287, opened 2023-02-17
Nano G1 Explorer

neilhao:
Known Issues: 
1. Finally we will use SH1107 128x64 resolution driver, because SH1106 will shift the screen by 2 lines.

The Nano G1 Explorer features a new internal wideband Lora antenna, which supports frequencies from 815 Mhz to 940 Mhz. The wideband antenna, combined with an optimized wideband Lora RF frontend circuit, enables the Nano G1 Explorer to work with the majority of regions' LoRa frequency bands around the world without the need for antenna changes. The design of the device also takes into account the potential effect of the human body on its antenna performance, ensuring optimal RF performance even when carried in a pocket.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,8 +2,9 @@
     // See http://go.microsoft.com/fwlink/?LinkId=827846
     // for the documentation about the extensions.json format
     "recommendations": [
-        "ms-vscode.cpptools",
-        "platformio.platformio-ide",
-        "trunk.io"
+        "platformio.platformio-ide"
     ],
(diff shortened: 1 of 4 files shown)


caveman99 (approved):

neilhao:
This version is already very stable, firmware version 2.0.21. 

Only need to replace the screen driver in the future when sh1107 128x64 driver available for meshtastic, however SH1106 does not affect the use for now. The SH1106's snow screen problem also disappeared, I think some other code update happened to give the screen enough time to initialize, but that can be just luck.

PR #2312, opened 2023-02-27
Trying to debug transient "disconnects" in iOS

thebentern:
* Added back lines from original Adafruit example sketch


thebentern:
Did not affect transient disconnect issues, but still integrating this because it's part of the reference code from Adafruit and doesn't seem to hurt anything

PR #2333, opened 2023-03-06
SX126x/8x: Add HEADER_VALID IRQ flag for actively receiving check

### LINKED ISSUE #2307: [Bug]: SX1262 radios drop packets when doing CAD while actively receiving
### Category

Other

### Hardware

Rak4631

### Firmware Version

2.0.17

### Description

My devices don't always show traffic in serial debugging, after a non-determinate amount of reboots it works as expected. Troubleshooting and logs in #2291 

### Relevant log output

_No response_

GUVWAF:
Fixes #2307.

This updates the RadioLib dependency to the latest release which includes user-configurable IRQ flags for SX126x and SX128x radios. Now the `isActivelyReceiving()` function actually works, which gives way better reliability.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -67,7 +67,7 @@ lib_deps =
   nanopb/Nanopb@^0.4.6
   erriez/ErriezCRC32@^1.0.1
 ;  jgromes/RadioLib@^5.5.1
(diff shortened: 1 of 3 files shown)


thebentern (approved):

GUVWAF:
@mrx23dot Interesting. Although now we also check for `PREAMBLE_DETECT` (#2349), so it's way more reliable already.

PR #2341, opened 2023-03-09
Report special battery_level of > 100 instead of zero to indicate USB

thebentern:


caveman99 (approved):
Assumiong getHasUSB() does the right thing, that looks fine.

PR #2342, opened 2023-03-09
Use startReceive() instead of startReceiveDutyCycleAuto()

GUVWAF:
@andrekir found that especially between **two** SX1262 radios, performance is still sub-optimal. Looks like it is missing some packets under high load, but when using a SX1276 for one of the two, it doesn't really create problems. 
Now the function `startReceive()` works again and seems to perform better. Leaving it as a draft for now, as I will ask him to test as well.

### DIFF
--- a/src/mesh/SX126xInterface.cpp
+++ b/src/mesh/SX126xInterface.cpp
@@ -212,10 +212,9 @@ template <typename T> void SX126xInterface<T>::startReceive()
 
     setStandby();
 
-    // We use a 16 bit preamble so this should save some power by letting radio sit in standby mostly.
     // Furthermore, we need the HEADER_VALID IRQ flag to detect whether we are actively receiving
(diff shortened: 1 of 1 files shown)


thebentern:
@andrekir working well for you?

GUVWAF:
Unfortunately the results are inconsistent. Using a T-Echo and RAK it was not better for him, while using a T-Beam with SX1262 and RAK it works pretty well, but that is regardless of using this PR or 2.1.0. 
I'm having perfect results (10 messages sent as fast as possible are successful without any retransmissions) using two (SX1276) T-Beams, while if one of them is a RAK, there are retransmissions needed. I'm suspecting it is related to the MCU then, maybe the different way of interrupt handling between nRF52 and ESP32 has anything to do with it. I will close this now and see if there's anything we can do about that.

Issue #2241, opened 2023-02-02
Eink Screen not working with 2.0.14.2baaad8


feh123:
I have three rak4631's with eink screens. I use them to communicate so they were using 2.0.3 to get the best signal range. I thought I would try 2.0.14 as the range issue was solved. I have updated pip and meshtastic to the latest versions. I tried the update using nrfutil with the ota-zip file from terminal. The updates were successfully installed (as seen in meshtastic --info) but eink screen did not update and still displayed 2.0.3 as the FW. I tried this on two rak4631's with the same result. I have not tried any other versions of the software. I went back to 2.0.3 on both and the screens are working well.

jfirwin:
Thanks @garthvh!

PR #2346, opened 2023-03-11
fix rak4631_epaper PIN_EINK_EN

### LINKED ISSUE #2101: [Bug]: Rak5005/4631 with eink screen and FW 2.0.10 to 2.0.12
### Category

Hardware Compatibility

### Hardware

Rak4631

### Firmware Version

2.0.10, 2.0.11 and 2.0.12

### Description

I was using 2.0.6 on my rak5005/4631/14000 (eink screen) devices. I have tried updating via terminal and nrfutil dfu serial to FW 2.0.10, 2.0.11 and 2.0.12 but there are some problems. With these versions I see no meshtastic logo splash screen on update completion. The updating seems to work - no errors are reported. When I connect the devices via the latest iOS app (from Testflight) I get no rebooting screen when changes parameters. Again the changes are accepted. I have a TBeam running 2.0.12 in the mesh as well. That reboots correctly displaying the splash screen and the rebooting message on parameter change. Oddly however in the iOS app it still says the TBeam is on 2.0.6! So nothing too serious but I thought I would report it - maybe not many users are using a 14000 screen.   

### Relevant log output

_No response_

andrekir:
`firmware-rak4631_eink` works again on stock RAK4631 + RAK14000 combo reverting 2.0.8 `PIN_EINK_EN` change back to (0 + 2). tested and confirmed by @feh123

fixes #2101 #2241

### DIFF
--- a/variants/rak4631_epaper/variant.h
+++ b/variants/rak4631_epaper/variant.h
@@ -126,7 +126,7 @@ static const uint8_t SCK = PIN_SPI_SCK;
  * eink display pins
  */
 
-#define PIN_EINK_EN (32 + 2) // (0 + 2) Note: this is really just backlight power
+#define PIN_EINK_EN (0 + 2) // (0 + 2) Note: this is really just backlight power
 #define PIN_EINK_CS (0 + 26)
 #define PIN_EINK_BUSY (0 + 4)
 #define PIN_EINK_DC (0 + 17)
--- a/variants/rak4631_epaper/variant.h
+++ b/variants/rak4631_epaper/variant.h
@@ -126,7 +126,6 @@ static const uint8_t SCK = PIN_SPI_SCK;
  * eink display pins
  */
 
-#define PIN_EINK_EN (0 + 2) // (0 + 2) Note: this is really just backlight power
 #define PIN_EINK_CS (0 + 26)
 #define PIN_EINK_BUSY (0 + 4)
 #define PIN_EINK_DC (0 + 17)
--- a/variants/rak4631/variant.h
+++ b/variants/rak4631/variant.h
@@ -126,7 +126,6 @@ static const uint8_t SCK = PIN_SPI_SCK;
  * eink display pins
  */
 
-#define PIN_EINK_EN (32 + 2) // (0 + 2) Note: this is really just backlight power
 #define PIN_EINK_CS (0 + 26)
 #define PIN_EINK_BUSY (0 + 4)
 #define PIN_EINK_DC (0 + 17)


caveman99:
Good it's working again. Can anybody explain WHY? The Power enable is clearly WB_IO2 and every bit of sample code provided by RAK says Eink_Power_En is WB_IO2 which is GPIO 34 (P1.02)

GPIO2 (P0.02) is QSPI_D3 which is part of the SPI Flash interface. When you trace that line from Core through Baseboard to the I/O connector (alternating named LED3 or IO8) it's a Non-Connect on the Epaper Module.

caveman99:
After further investigation we have ButtonThread.h userButtonDoublePressed() which sends a network ping AND switches the backlight on T-Echo. On RAK (Eink firmware only) the double click will pull IO2 and thus power to all peripherals if PIN_EINK_EN stays at 34. We should reconsider making the T-Echo send out a netowrk ping when you toggle the backlight.

The real problem is in EinkDisplay2.cpp Linke 173. On minit it switches the perceived backlight pin to LOW disabling the peripheral power for the EINK Firmware on RAK. So my suspicion that this should best be UNdefined on the 4631-EINK is confirmed. The pin is not named very clever since it only controls the backlight. RAK epaper does NOT use backlight at all :-)

variants/rak4631_epaper/variant.h
```diff
@@ -126,7 +126,7 @@ static const uint8_t SCK = PIN_SPI_SCK;
  * eink display pins
  */
 
-#define PIN_EINK_EN (32 + 2) // (0 + 2) Note: this is really just backlight power
+#define PIN_EINK_EN (0 + 2) // (0 + 2) Note: this is really just backlight power
```
caveman99, line 129: I propose to remove that define alltogether. Should still work. 0+2 just points somehwhere random. see my findings in the main comments.

feh123:
As I recall and maybe this is obvious but from --noproto output from 2.0.8 onwards there was always a statement about the screen power on. There was no indication that the screen was an eink.

variants/rak4631_epaper/variant.h
caveman99, line 129: I propose to remove that define alltogether. Should still work. 0+2 just points somehwhere random. see my findings in the main comments.

caveman99 (changes requested):

andrekir:
@feh123 can you please confirm if this change works https://github.com/meshtastic/firmware/suites/11504984242/artifacts/594672483

feh123:
Hi @andrekir this looks fine too.

andrekir:
@feh123 thank you

caveman99 (approved):

PR #2349, opened 2023-03-11
SX126x/8x: Also use PREAMBLE_DETECTED IRQ flag for actively receiving check

GUVWAF:
This also uses the `PREAMBLE_DETECTED` IRQ for determining whether we are actively receiving with SX126x and SX128x radios. Before, sometimes if you only received the preamble and not yet the header, it would perform CAD and then detect there is LoRa activity, but then you lose the packet you are currently receiving, because to go into CAD mode it goes into standby first.

Downside of this is that it may lead to false detections. To overcome this, I added a timing check to ignore the preamble detection flag when we should have received a header. I set it to two times the preamble time to be safe (and backwards compatible). In the worst case we start transmitting a bit later than necessary with this. 
It also ignores the `HEADER_VALID` flag when we should have gotten the `RX_DONE` IRQ, for which the timeout is set to the airtime of a packet with maximum payload. I’ve never seen a false header detection and this is also very unlikely to happen I think, but I added it just in case.

@andrekir and I confirmed that this does work even better under high load, also with three devices.


geeksville:

https://meshtastic.discourse.group/t/traceroute-not-working/7097/30

Issue #2358, opened 2023-03-14
[Bug]: Heap Leak of 324 bytes every 60s when DeviceTelemetryModule is doing a local update


mverch67:
### Category

Memory leak

### Hardware

T-Beam

### Firmware Version

2.1.2

### Description

Compile 2.1.2 with -DDEBUG_HEAP in platform.ini.
Flash firmware to device and connect via bluetooth with (Android)APP.

With firmware 2.1.2 the free heap is constantly decreasing every 60 seconds by 324 bytes.

### Relevant log output

```Shell
DEBUG | ??:??:?? 22 [Power] Heap status: 120452/205824 bytes free (-21280), running 11/24 threads
DEBUG | ??:??:?? 42 [Power] Heap status: 120388/205792 bytes free (-64), running 11/24 threads
DEBUG | ??:??:?? 62 [Power] Heap status: 120348/205776 bytes free (-40), running 11/24 threads
DEBUG | ??:??:?? 82 [Power] Heap status: 120028/205696 bytes free (-320), running 11/24 threads
DEBUG | ??:??:?? 122 [Power] Heap status: 119664/205664 bytes free (-364), running 11/24 threads
DEBUG | ??:??:?? 182 [Power] Heap status: 119340/205648 bytes free (-324), running 11/24 threads
DEBUG | ??:??:?? 242 [Power] Heap status: 119016/205632 bytes free (-324), running 11/24 threads
DEBUG | ??:??:?? 302 [Power] Heap status: 118692/205616 bytes free (-324), running 11/24 threads
DEBUG | ??:??:?? 362 [Power] Heap status: 118368/205600 bytes free (-324), running 11/24 threads
DEBUG | ??:??:?? 422 [Power] Heap status: 118044/205584 bytes free (-324), running 11/24 threads
DEBUG | ??:??:?? 482 [Power] Heap status: 117720/205568 bytes free (-324), running 11/24 threads
DEBUG | ??:??:?? 542 [Power] Heap status: 117396/205552 bytes free (-324), running 11/24 threads
DEBUG | ??:??:?? 602 [Power] Heap status: 117072/205536 bytes free (-324), running 11/24 threads
DEBUG | ??:??:?? 662 [Power] Heap status: 116748/205520 bytes free (-324), running 11/24 threads
DEBUG | ??:??:?? 722 [Power] Heap status: 116424/205504 bytes free (-324), running 11/24 threads
[...]

PR #2359, opened 2023-03-14
Fix heap leak mentioned in #2358

mverch67:
This is a fix for #2358. 

In MeshService::sendToPhone() a copy of a meshPacket is created. This copy is sent into the phoneQueue where it eventually gets deleted when fetched by the phone. The original packet is only released when it is sent over the mesh, which will not happen if DeviceTelemetryModule enters the upper branch in method EnvironmentTelemetrieModule::sendTelemetry():

```
    lastMeasurementPacket = packetPool.allocCopy(*p); // first copy
    if (phoneOnly) {
        LOG_INFO("Sending packet to phone\n");
        service.sendToPhone(p);                       // another copy
    } else {
        LOG_INFO("Sending packet to mesh\n");
        service.sendToMesh(p, RX_SRC_LOCAL, true);
    }
```
The fix avoids the copying in sendToPhone(). The copying is better done in sendToMesh(), just before invoking sendToPhone():

```
    if (ccToPhone) {
        sendToPhone(packetPool.allocCopy(*p));
    }

```

Note: the relevant code in AirQualityTelemetryModule looks exactly the same as in EnvironmentTelemetryModule, so there should be no issue. However, I was not able to test it due to lack of such equipment.

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -82,7 +82,7 @@ int MeshService::handleFromRadio(const meshtastic_MeshPacket *mp)
     }
 
     printPacket("Forwarding to phone", mp);
-    sendToPhone((meshtastic_MeshPacket *)mp);
+    sendToPhone(packetPool.allocCopy(*mp));
 
     return 0;
 }
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -231,7 +231,7 @@ void MeshService::sendToMesh(meshtastic_MeshPacket *p, RxSource src, bool ccToPh
     }
 
     if (ccToPhone) {
-        sendToPhone(p);
+        sendToPhone(packetPool.allocCopy(*p));
     }
 }
 
@@ -262,9 +262,8 @@ void MeshService::sendToPhone(meshtastic_MeshPacket *p)
             releaseToPool(d);
     }
 
-    meshtastic_MeshPacket *copied = packetPool.allocCopy(*p);
-    perhapsDecode(copied);
-    assert(toPhoneQueue.enqueue(copied, 0));
+    perhapsDecode(p);
+    assert(toPhoneQueue.enqueue(p, 0));
     fromNum++;
 }
 


thebentern:
Great catch! AirQualityTelemetry was based on this module as well, so it will need to be patched

thebentern (approved):

mverch67:
Unfortunately I overlooked one invocation of sendToPhone() in line 85 where it should also read
    sendToPhone(packetPool.allocCopy(*mp));

How to proceed?

mverch67:
> 

The is no need to change AirQualityTelemetry because I decided to fix this issue in sentToPhone() which leaked the memory. Otherwise, if a fix had been done only in DeviceTelemetry (and AirQualityTelemetry) the interface of MeshService would look very awkward because sendToMesh() would have to be called with a copied package (because it deletes it) while sendToPhone() had to be called without a copy (because it did do a copy for itself). Now, with the change in MeshService the interfaces are more clean. It may have other side-effects, though (e.g. SimRadio which also calls sendToPhone() directly).

thebentern:
> There is no need to change AirQualityTelemetry because I decided to fix this issue in sentToPhone() which leaked the memory. Otherwise, if a fix had been done only in DeviceTelemetry (and AirQualityTelemetry) the interface of MeshService would look very awkward because sendToMesh() would have to be called with a copied package (because it deletes it) while sendToPhone() had to be called without a copy (because it did do a copy for itself). Now, with the change in MeshService the interfaces are more clean. It may have other side-effects, though (e.g. SimRadio which also calls sendToPhone() directly).

I wasn't paying enough attention to your original change. Makes sense. Looks good to me

GUVWAF:
Great catch indeed!

>  It may have other side-effects, though (e.g. SimRadio which also calls sendToPhone() directly).

I wouldn’t worry too much about SimRadio. This is only used for simulating the radio on a native Linux system, so it will not be used for long consecutive periods and only on more powerful machines. I quickly checked it with the interactive simulator and this at least doesn’t break it :)

PR #2361, opened 2023-03-16
fix channel_num log

andrekir:
the value shown in the log for `channel_num` is actually (channel_num - 1) because that is the value used in the formula to calc frequency center:

https://github.com/meshtastic/firmware/blob/b398f31b64de5162549cedbb5e7d3ae9964e6f61/src/mesh/RadioInterface.cpp#L457
https://github.com/meshtastic/firmware/blob/b398f31b64de5162549cedbb5e7d3ae9964e6f61/src/mesh/RadioInterface.cpp#L463

below should be preferred for clarity:
```
int channel_num = (loraConfig.channel_num ? loraConfig.channel_num - 1 : hash(channelName)) % numChannels + 1;
...
float freq = myRegion->freqStart + (bw / 2000) + ((channel_num - 1) * (bw / 1000));
```

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -454,13 +454,13 @@ void RadioInterface::applyModemConfig()
 
     // If user has manually specified a channel num, then use that, otherwise generate one by hashing the name
     const char *channelName = channels.getName(channels.getPrimaryIndex());
-    int channel_num = (loraConfig.channel_num ? loraConfig.channel_num - 1 : hash(channelName)) % numChannels;
+    int channel_num = (loraConfig.channel_num ? loraConfig.channel_num : hash(channelName)) % numChannels;
 
     // Old frequency selection formula
     // float freq = myRegion->freqStart + ((((myRegion->freqEnd - myRegion->freqStart) / numChannels) / 2) * channel_num);
 
     // New frequency selection formula
-    float freq = myRegion->freqStart + (bw / 2000) + (channel_num * (bw / 1000));
+    float freq = myRegion->freqStart + (bw / 2000) + ((channel_num - 1) * (bw / 1000));
 
     // override if we have a verbatim frequency
     if (loraConfig.override_frequency) {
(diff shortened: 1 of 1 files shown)


src/mesh/RadioInterface.cpp
```diff
@@ -454,13 +454,13 @@ void RadioInterface::applyModemConfig()
 
     // If user has manually specified a channel num, then use that, otherwise generate one by hashing the name
     const char *channelName = channels.getName(channels.getPrimaryIndex());
-    int channel_num = (loraConfig.channel_num ? loraConfig.channel_num - 1 : hash(channelName)) % numChannels;
+    int channel_num = (loraConfig.channel_num ? loraConfig.channel_num : hash(channelName)) % numChannels;
```
caveman99, line 457: That would be a breaking change for everyone defining a fixed channel_num.

src/mesh/RadioInterface.cpp
```diff
@@ -454,13 +454,13 @@ void RadioInterface::applyModemConfig()
...
 
     // Old frequency selection formula
     // float freq = myRegion->freqStart + ((((myRegion->freqEnd - myRegion->freqStart) / numChannels) / 2) * channel_num);
 
     // New frequency selection formula
-    float freq = myRegion->freqStart + (bw / 2000) + (channel_num * (bw / 1000));
+    float freq = myRegion->freqStart + (bw / 2000) + ((channel_num - 1) * (bw / 1000));
```
caveman99, line 463: This will shift the actual frequency even if we use the hash formula. Old behaviour: 1 is the lowest number, cause 0 in the protobuf means 'calculate it yourself'. Now: lowest number is 0. So freq will be different after the patch.

caveman99 (changes requested):
Valid observation, but the proposed fix will break things badly for all users. Can't we just fix the log display?

src/mesh/RadioInterface.cpp
andrekir: ah, good you mention. that modulus operation threw me off. `channel_num` variable actually returns between 0 and (numChannels - 1).

Issue #2365, opened 2023-03-17
[Bug]: Sporadig crashes with reboot


mverch67:
### Category

Firmware crash

### Hardware

TLora T3S3

### Firmware Version

2.1.2

### Description

The device reboots sporadically. This happens more frequently if more traffic is generated (e.g. range test set to 15s). 

### Relevant log output

```Shell
//\ E S H T /\ S T / C

INFO  | ??:??:?? 1 Booted, wake cause 0 (boot count 1), reset_reason=reset
DEBUG | ??:??:?? 1 Filesystem files (479232/1048576 Bytes):
DEBUG | ??:??:?? 1  /prefs/channels.proto (178 Bytes)
DEBUG | ??:??:?? 2  /prefs/config.proto (95 Bytes)
DEBUG | ??:??:?? 2  /prefs/db.proto (142 Bytes)
DEBUG | ??:??:?? 2  /prefs/module.proto (71 Bytes)
DEBUG | ??:??:?? 2  /static/.gitkeep (0 Bytes)
DEBUG | ??:??:?? 2  /static/Logo_Black.svg.gz (592 Bytes)
DEBUG | ??:??:?? 2  /static/Logo_White.svg.gz (600 Bytes)
DEBUG | ??:??:?? 2  /static/apple-touch-icon.png.gz (3162 Bytes)
DEBUG | ??:??:?? 2  /static/favicon.ico.gz (2221 Bytes)
DEBUG | ??:??:?? 2  /static/icon.svg.gz (842 Bytes)
DEBUG | ??:??:?? 2  /static/index-0e816946.js.gz (399863 Bytes)
DEBUG | ??:??:?? 2  /static/index-9602d760.css.gz (17494 Bytes)
DEBUG | ??:??:?? 2  /static/index-fc2e9253.js.gz (347 Bytes)
DEBUG | ??:??:?? 2  /static/index.html.gz (547 Bytes)
DEBUG | ??:??:?? 2  /static/robots.txt.gz (42 Bytes)
DEBUG | ??:??:?? 2  /static/site.webmanifest.gz (196 Bytes)
DEBUG | ??:??:?? 2 Using analog input 1 for battery level
INFO  | ??:??:?? 2 Scanning for i2c devices...
DEBUG | ??:??:?? 2 Scanning for i2c devices on port 2
DEBUG | ??:??:?? 2 Scanning for i2c devices on port 1
DEBUG | ??:??:?? 2 I2C device found at address 0x3c
[...]

PR #2366, opened 2023-03-17
Fix: Sporadig crashes and reboot

mverch67:
This is a fix for [Bug]: Sporadig crashes with reboot #2365.

After some investigations it is clear that the cause is a dangling pointer which is used after being erased from a hashmap:

```
bool ReliableRouter::stopRetransmission(GlobalPacketId key)
{
    auto old = findPendingPacket(key);          // returns PendingPacket pointer 
    if (old) {
        auto numErased = pending.erase(key);    // PendingPacket is deleted from hashmap
        assert(numErased == 1);
        cancelSending(getFrom(old->packet), old->packet->id);  // old is now a dangling pointer!
        if (old->packet)
            packetPool.release(old->packet);
        return true;
    } else
        return false;
}
```

The fix will save the MeshPacket pointer within the PendingPacket envelope before deletion.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,8 +2,9 @@
     // See http://go.microsoft.com/fwlink/?LinkId=827846
     // for the documentation about the extensions.json format
     "recommendations": [
-        "ms-vscode.cpptools",
-        "platformio.platformio-ide",
-        "trunk.io"
+        "platformio.platformio-ide"
     ],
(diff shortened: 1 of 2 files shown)


thebentern:
Hate the dangling pointer! Nice catch

caveman99:
That was the spot where I said 'lets try it that way, if it crashes there we know it didn't work' 3 weeks back ;-)

PR #2371, opened 2023-03-21
Use accelerometer to wake up screen

thebentern:


imkebe (commented):
Hi. Did you managed to get it working?  I was thinking of using MPU6050 in my TTGO Beam.

PR #2384, opened 2023-03-25
Add nodedb channel handling

sbias:
Store channel of a node to nodedb and use it to send our packets to.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -165,6 +165,8 @@ void NodeDB::installDefaultConfig()
     config.lora.hop_limit = HOP_RELIABLE;
     config.position.gps_enabled = true;
     config.position.position_broadcast_smart_enabled = true;
+    config.position.broadcast_smart_minimum_distance = 100;
+    config.position.broadcast_smart_minimum_interval_secs = 30;
     if (config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER)
         config.device.node_info_broadcast_secs = 3 * 60 * 60;
     config.device.serial_enabled = true;
@@ -231,9 +233,7 @@ void NodeDB::installRoleDefaults(meshtastic_Config_DeviceConfig_Role role)
     } else if (role == meshtastic_Config_DeviceConfig_Role_REPEATER) {
         config.display.screen_on_secs = 1;
     } else if (role == meshtastic_Config_DeviceConfig_Role_TRACKER) {
-        config.position.position_broadcast_smart_enabled = false;
-        config.position.position_broadcast_secs = 120;
-        config.position.gps_update_interval = 60;
+        config.position.gps_update_interval = 30;
     } else if (role == meshtastic_Config_DeviceConfig_Role_SENSOR) {
         moduleConfig.telemetry.environment_measurement_enabled = true;
         moduleConfig.telemetry.environment_update_interval = 300;
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -58,12 +58,6 @@ meshtastic_User &owner = devicestate.owner;
 
 static uint8_t ourMacAddr[6];
 
-/**
- * The node number the user is currently looking at
- * 0 if none
- */
-NodeNum displayedNodeNum;
-
 NodeDB::NodeDB() : nodes(devicestate.node_db), numNodes(&devicestate.node_db_count) {}
 
 /**
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -744,6 +744,15 @@ void NodeDB::updateFrom(const meshtastic_MeshPacket &mp)
     }
 }
 
+uint8_t NodeDB::getNodeChannel(NodeNum n)
+{
+    meshtastic_NodeInfo *info = getNode(n);
+    if (!info) {
+        return 0; // defaults to PRIMARY
+    }
+    return info->channel;
+}
+
 /// Find a node in our DB, return null for missing
 /// NOTE: This function might be called from an ISR
 meshtastic_NodeInfo *NodeDB::getNode(NodeNum n)
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -734,7 +734,9 @@ void NodeDB::updateFrom(const meshtastic_MeshPacket &mp)
         if (mp.rx_snr)
             info->snr = mp.rx_snr; // keep the most recent SNR we received for this node.
 
-        info->channel = mp.channel;
+        if (mp.decoded.portnum == meshtastic_PortNum_NODEINFO_APP) {
+            info->channel = mp.channel;
+        }
     }
 }
 
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -727,7 +727,7 @@ void NodeDB::updateUser(uint32_t nodeId, const meshtastic_User &p)
 void NodeDB::updateFrom(const meshtastic_MeshPacket &mp)
 {
     if (mp.which_payload_variant == meshtastic_MeshPacket_decoded_tag && mp.from) {
-        LOG_DEBUG("Update DB node 0x%x, rx_time=%u\n", mp.from, mp.rx_time);
+        LOG_DEBUG("Update DB node 0x%x, rx_time=%u, channel=%d\n", mp.from, mp.rx_time, mp.channel);
 
         meshtastic_NodeInfo *info = getOrCreateNode(getFrom(&mp));
         if (!info) {
@@ -739,6 +739,8 @@ void NodeDB::updateFrom(const meshtastic_MeshPacket &mp)
 
         if (mp.rx_snr)
             info->snr = mp.rx_snr; // keep the most recent SNR we received for this node.
+
+        info->channel = mp.channel;
     }
 }
 
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -173,7 +173,8 @@ ErrorCode Router::sendLocal(meshtastic_MeshPacket *p, RxSource src)
             handleReceived(p, src);
         }
 
-        p->channel = nodeDB.getNodeChannel(p->to);
+        if (p->channel) // don't override if a channel was requested
+            p->channel = nodeDB.getNodeChannel(p->to);
         LOG_DEBUG("localSend to channel %d\n", p->channel);
 
         return send(p);
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -173,6 +173,9 @@ ErrorCode Router::sendLocal(meshtastic_MeshPacket *p, RxSource src)
             handleReceived(p, src);
         }
 
+        p->channel = nodeDB.getNodeChannel(p->to);
+        LOG_DEBUG("localSend to channel %d\n", p->channel);
+
         return send(p);
     }
 }
--- a/bin/regen-protos.sh
+++ b/bin/regen-protos.sh
@@ -8,7 +8,7 @@ echo "prebuilt binaries for your computer into nanopb-0.4.7"
 
 # the nanopb tool seems to require that the .options file be in the current directory!
 cd protobufs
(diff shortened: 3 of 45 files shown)


GUVWAF:
While the code looks good to me, the only thing that changes here is the initial NodeInfo (when you don’t know the node yet) and the function `sendNetworkPing`, which is sent on a double button press and regularly when the screen is currently displaying this node.
All the broadcast intervals you set only apply to the primary channel. If I’m right, the goal of this is to share NodeInfo and position only to specific nodes. With this, it works only for these cases I mentioned before. I’m not sure whether that solves the use case and if it’s not going to create more confusion, since you need to have different primary channels, but the one node needs the primary channel of the other as a secondary channel. 
I'm wondering if it would be better to make the channel on which you sent position updates configurable, such that you can share a secondary channel only to the nodes that are allowed to receive them.

With this you can indeed also _transmit_ DMs on a secondary channel, but you can still only _receive_ DMs on your primary channel (because that is where you sent your NodeInfo on and will be saved as your channel by the other node). Meaning that still anyone that has the same primary channel can receive the DMs. If you want to make a real private conversation, you would have to do a 'handshake' telling which channel to use for a DM to a specific node.

caveman99:
@GUVWAF the idea behind this PR was to lay the groundworks for sending DM not only on the primary channel. by remembering where you heard the node (and only responding to requests, not actively broadcasting info on a non-primary channel *to you* - meaning you can use setups where you have several mesh channels on the same device and talk/dm even to nodes you don't receive on your primary channel (which may very well be THEIR primary channel. So broadcasting/announcing should only happen on the primary (for me) but i should respond to nodes not on the primary, with nodeinfo and positions requested.

garthvh:
> @GUVWAF the idea behind this PR was to lay the groundworks for sending DM not only on the primary channel. by remembering where you heard the node (and only responding to requests, not actively broadcasting info on a non-primary channel _to you_) - meaning you can use setups where you have several mesh channels on the same device and talk/dm even to nodes you don't receive on your primary channel (which may very well be THEIR primary channel. So broadcasting/announcing should only happen on the primary (for me) but i should respond to nodes not on the primary, with nodeinfo and positions requested. This way (in theory) we don't put unneccessary packets on the air, but building up a database to remeber how to reach a certain node by DM.

Not actively broadcasting seems like the right choice here.  This has zero client impact since we already get this data for every packet correct?

sbias:
Well, that is right, both of it. The first goal was to implement direct communication between 2 devices where one has the primary channel as the secondary but don't share the primary.
As a next step it might be possible to make channels configurable if broadcasts should be sent on, but that means sending my nodeinfo 8 times for one broadcast if i have configured 8 channels for this. I'm not sure if this worth the airtime.
In this PR the nodeinfo is sent to any unknown nodes once heard, so, it might take some minutes but that info will propagate over time. And if a posInfo was sent by a tracker on his primary which was received on a secondary, that posInfo will be updated in nodeDb as well.

GUVWAF:
I now see that for responses already the same channel as the request came in on was used, so position requests also work indeed:
```
void setReplyTo(meshtastic_MeshPacket *p, const meshtastic_MeshPacket &to)
{
    ...
    p->channel = to.channel; // Use the same channel that the request came in on
```

> As a next step it might be possible to make channels configurable if broadcasts should be sent on, but that means sending my nodeinfo 8 times for one broadcast if i have configured 8 channels for this. I'm not sure if this worth the airtime.

I agree that you should not actively broadcast on every channel, I was more thinking of setting it to one specific channel (like it is the primary channel now), such that you only need one channel for that in a mesh. Now if you want to use separate channels for DMs, you need a different primary channel. If you then still want to receive the position/device telemetry broadcasts, you have to add all primary channels of the other nodes as secondary channels.

GUVWAF:
Basically this means per node you can only have 1 private chat for DMs and _only_ the node of that chat can also receive your position/device telemetry packets. 

If you share your primary channel to multiple other nodes such that they can also receive your position/device telemetry, they all can receive DMs sent to you.

caveman99:
therte's no privacy concern i think. dm's can be read by the passing nodes already, the difference here is we send it out to another network we may have just as sewcondary. this is not to improvve privacy, but to do more flexible setups, in line with QR-sharing only a subset or even one of your channels. imagine you having your private mesh but you can add your public city net as a secondary and rech people on it through DM.

GUVWAF:
If you have a private mesh, you can indeed send a DM on a secondary channel to someone on a public net. But if the one on the public net wants to send a DM back, it needs to have the primary channel of the private mesh as a secondary channel (so it's not really private anymore for that node).

But if this is not to improve privacy, I'm fine with it :)

GUVWAF (approved):

caveman99:
If i want to answer the DM in your scenario, won't te sending node receive the DM through the same channel? Cause the answering node ideally also has on file on which channel it heard the target node, and sends it down that way. Or do we need to add DM receive (in contrast to sending) capability on secondary channels?

GUVWAF:
Oh, I was wrong there, a node can receive a DM on a secondary channel, there's nothing we need to add for that, because it will just be able to decode it. 

In this scenario, the one on the private net would receive the NodeInfo on the channel of the public net and send a response on that channel. Therefore, the one on the public net would just use its primary channel to send the DM. So, the one on the public net doesn't need to have the key of the private net as secondary channel. 

It's just that the user of the private net doesn't know whether the channel of the private or public net will be used (if you don't know whether the other has the private net's channel as primary or not), so if you're using secondary channels, you can't assume that DMs can only be read by nodes on your primary channel. But, yeah, this PR is not about improving privacy.

TL/DR: This stuff is confusing, but it looks like it is going to work just fine with this PR :)

thebentern (approved):

thebentern:
Great work!
We'll have to have our alphanauts test this one thoroughly

garthvh:
I am interested in thoughts from @andrekir he know these details inside and out.

src/mesh/Router.cpp
```diff
@@ -173,6 +173,9 @@ ErrorCode Router::sendLocal(meshtastic_MeshPacket *p, RxSource src)
             handleReceived(p, src);
         }
 
+        p->channel = nodeDB.getNodeChannel(p->to);
```
andrekir, line 176: not too familiar with the code, does this override the packet's channel? (ie. client sends packet to node_123 on channel 8 but this changes to NodeInfo.channel)

if so, it is not recommended and packet channel should always be honored. otherwise it can break things (like admin channels).

src/mesh/NodeDB.cpp
```diff
@@ -739,7 +739,18 @@ void NodeDB::updateFrom(const meshtastic_MeshPacket &mp)
 
         if (mp.rx_snr)
             info->snr = mp.rx_snr; // keep the most recent SNR we received for this node.
+
+        info->channel = mp.channel;
```
andrekir, line 743: consider saving NodeInfo.channel for NodeInfo(?) packets only, so that we only tag the remote nodes primary channel. 

otherwise it will save different channel values every time packets are received in any shared secondary channels.

andrekir:
added a few remarks but overall really like the proposal. 👍 

> This has zero client impact since we already get this data for every packet correct?

the advantage here is clients get NodeInfo.channel upfront in wantConfig, so we don't need to wait for a packet from the node to know their primary channel and can just point DMs to use NodeInfo.channel.

src/mesh/Router.cpp
sbias: this is a good point. if a channel was requested it should be respected. but, any idea how to differ between not set and primary (0) ?

src/mesh/NodeDB.cpp
sbias: This is an improvement of cause, but it does not resolve channel-hopping entirely. what if 2 nodes are on 2 shared channels and a 3. node requests nodeinfo on a secondary. that will change the channel too since it hears a nodeinfo on a secondary channel. i was thinking about a channel change if a node is heard on a primary channel, or channel-index is higher in NodeDb than new packet. But there is my main problem again, how to differ between, heard on primary and unset.

caveman99:
regarding the channel 0 vs. unset problem: The field is new, so nobody uses it yet. Also there were requests in the past to allow more than 8 channels. i propose using a magic value of 31 (for reasons) to signal 'channel zero/primary'. this is never sent over the air, but only stored on a node.

src/mesh/Router.cpp
caveman99: the only packets this is intended for are DM packets, which are up to now on channel 0 - secondary channels will always be non-zero value. So if the channel (from the phone) is zero AND the destination is not a broadcast we can safely alter it. If not, leave it alone.

src/mesh/NodeDB.cpp
caveman99: every node has only one primary channel. So if you trigger the update only on received nodeinfo packets you will always have that node's primary channel from your perspective. If this value is zero, you share a common primary channel. If the local value here is already non-zero, don't update it, cause you obviously hear the node across several paths. This assumes a fairly static setup.

sbias: sorry, but i think that might not work if we send our nodeInfo on a secondary channel to reach out new unknown nodes while a third node is listening on that channel as well. I added that here: https://github.com/meshtastic/firmware/pull/2384/commits/85b18a4387383ad4e62f9e9a10393371872c53bd

src/mesh/Router.cpp
sbias: if we want to set this only for DM it might be an idea to not set it in this router but in the client itself? Or somewhere between?

sbias:
Ok, an idea: I fix that toxic behavior of my change of ignoring requested channels @andrekir mentioned here (https://github.com/meshtastic/firmware/pull/2384#discussion_r1148523839) by if-ing channel > 0 .

I add a check for updating channel in NodeDB to do this only if it is not a well known channel.

If done, i don't see any problems caused by this pr, so we can test it.
What do you think?

src/mesh/Router.cpp
andrekir: NodeInfo.channel set vs. unset is not the issue. overriding the channel will break the special admin/gpio channels and agree clients should pick NodeInfo.channel if that's the intended target. this is best left out.

andrekir:
> edit: we can set channel-id to 255 in node-db as default and store only a first-seen channel. that should prevent channel-hoping. there is no problem of using channel 31 for default, but, in my opinion this should be changed everywhere than. translating ids in code might cause some confusion later on.

if-ing channel > 0 is not viable, so I believe using any non-0 value is not necessary. yet. 😃

src/mesh/NodeDB.cpp
andrekir: one option when receiving NodeInfo on a secondary channel --> `sendOurNodeInfo` with channel = our secondary channel (or any value for the matter, non-0 to indicate this is not our primary channel).

- nodes receiving a NodeInfo packet with a tagged NodeInfo.channel should consider lower priority 
- if we already seen that node on our primary don't replace (unless 2-3x `node_info_broadcast_secs` ago?)

thoughts?

src/mesh/Router.cpp
sbias: my plan was to check p.channel if a channel was request to not override it :)

src/mesh/Router.cpp
```diff
@@ -173,7 +173,8 @@ ErrorCode Router::sendLocal(meshtastic_MeshPacket *p, RxSource src)
             handleReceived(p, src);
         }
 
-        p->channel = nodeDB.getNodeChannel(p->to);
+        if (p->channel) // don't override if a channel was requested
```
GUVWAF, line 176: Shouldn't this be if (!p->channel) then?

src/mesh/NodeDB.cpp
caveman99: That's a good point - sholdn't we consider nodeinfo / nodedb stale at some point and clean it up? That's another issue/PR but it would take care of this situation. My suggestion to prioritize reception on a secondary channel over the primary was that it is easier to do in code. Here again 0 can mean 2 things...

geeksville:

https://meshtastic.discourse.group/t/question-about-private-messages-with-multiple-channels/11092/4

PR #2387, opened 2023-03-26
Remove sending network ping to displayed node on interval

### LINKED ISSUE #2360: [Bug]: Only send NodeInfo in response to a position request, not on the position schedule
### Category

Other

### Hardware

Not Applicable

### Firmware Version

2.1.2

### Description

When a position is sent directly to another node and a response is requested we send a NodeInfo if the device does not have a position.  This works pretty well and at least lets the other user know that the device responded.

We also seem to send an empty NodeInfo on the position schedule if the device has no GPS, and no fixed or phone provided positions which seems unnecessary.

### Relevant log output

_No response_

GUVWAF:
Fixes #2360.
This removes sending a network ping on an interval to the node that is currently displayed on a screen. This feature silently overrules the position broadcast setting, or sends its NodeInfo over and over again, which seems unnecessary. It will also create a lot of packets when you set the screen to carousel.
The ‘last seen’ time already gets updated with any other packet received from the node. Besides, this feature is not documented, so I don’t think anyone will miss it.

### DIFF
--- a/src/PowerFSM.cpp
+++ b/src/PowerFSM.cpp
@@ -194,17 +194,6 @@ static void onEnter()
     LOG_INFO("Enter state: ON\n");
     screen->setOn(true);
     setBluetoothEnable(true);
-
-    static uint32_t lastPingMs;
-
-    uint32_t now = millis();
-
-    if ((now - lastPingMs) >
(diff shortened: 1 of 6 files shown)


caveman99 (approved):

garthvh:
Amazing find!

PR #2391, opened 2023-03-28
DIY Targets nRF52/ESP32-S3/ESP32-C3

markbirss:
MakerPython nRF52 SX1280 OLED/e-Ink
x2

ESP32-S3 SX1280 OLED/e-Ink
x2

ESP32-C3 M5Stack Stamp C3 RFM95W/SX1262 OLED
x1

Rak6431 e-Ink directly connected to 5005-O base i2c/serial pins
x1


markbirss:
This is just a collection aimed to help others wanting to add DIY targets

PR #2393, opened 2023-03-29
Support double tap as button press for supported accelerometers

thebentern:


caveman99 (approved):
Looking ace. MPU6050 doesn't work?

thebentern:
> Looking ace. MPU6050 doesn't work?

I believe MPU6050 hardware supports tap / click interrupts with use some register, but the Adafruit lib sadly excludes it and it would seem they aren't even looking at PRs folks have sent. So I suspect it's going to remain statel. 

I think there are some alternative libraries we should look at. I was hoping to remain consistent with using Adafruit libs for the common peripherals, but it seems like some of them have like the MPU6050 have decent capability gaps.

caveman99:
The gap issue is also apparent with the BME680.

PR #2398, opened 2023-03-31
DIY external power pin

osmanovv:
This fixes various issues such as unexpected node sleep, WiFi disconnection.

On some boards (DIY boards with ESP32 DevKit and LILYGO® TTGO T-Energy T18) we don't have a power management chip (like AXPxxxx) so we use GPIO pin `EXT_PWR_DETECT` to detect the external power supply.

To detect an external power supply, we need to pull up the `EXT_PWR_DETECT` pin on the board.

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -165,11 +165,11 @@ class AnalogBatteryLevel : public HasBatteryLevel
     virtual bool isVbusIn() override
     {
 #ifdef EXT_PWR_DETECT
-    // if external powered that pin will be pulled up
-    if (digitalRead(EXT_PWR_DETECT) == HIGH) {
-        return true;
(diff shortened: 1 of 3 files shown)


caveman99 (approved):

osmanovv:
@thebentern, thank you for fixing formatting. I biuld on Windows machine where trunk extension is not available. Is there other way to check code style without this Trunk extension?

caveman99:
@osmanovv i use windows too, i have a WSL2 window running ubuntu and use trunk in there. the windows filesystems are available inside WSL under /mnt/<driveletter>

osmanovv:
Thank you! I have just managed how to run it in WSL:

image

Issue #2402, opened 2023-04-02
[Bug]: Screen issues on device display


mverch67:
### Category

Other

### Hardware

T-Beam, TLora

### Firmware Version

2.1.6

### Description

There are a couple of issues regarding the behavior and the information on the device display :- 

1- the screen timeout does not work properly; the display won't go blank (only in mode ROUTER it does)

2- in case the display was blank; it won't go blank anymore after it went on (because of a message that was displayed)

3- after reboot it takes a while until all the nodes are shown in the various frames; time differs from device to device

4- on button press the display shows "powered..." or "unpowered...". This is not critical but annoying and should only be shown when the USB is plugged-in or pulled out.

5- when in Router mode, screen is blank, then a short button press does nothing; screen stays dark, no BLE

6- when in Router mode, screen is blank, then a long button press leads to crash with reboot

PR #2403, opened 2023-04-02
Fix for device display issues

mverch67:
This is a fix for #2402 Screen issues on device display.

Regarding the issues mentioned in #2402 :- 
 1- the missing screen blanking is fixed by adding a new state transition from POWER -> DARK
 2- the missing screen blanking while in DARK (when displaying a message) is fixed by adding a new state transition DARK -> DARK
 3- to show all nodes right from the beginning the node number is taken from nodeDB 
 4- a couple of state checks fix the unwanted display of "powered... ", "unpowered...". Tested well on Tbeam with insertion/removal of USB cable.
 5+6- the button GPIO sleep issue and crash is fixed
 7- FREE GIFT: Router IRQ wakeup for SX126x/SX128x is fixed!!! Saw this when I corrected the button GPIO sleep issue.

### DIFF
--- a/src/PowerFSM.cpp
+++ b/src/PowerFSM.cpp
@@ -69,11 +69,11 @@ static void lsIdle()
 
     // Do we have more sleeping to do?
     if (secsSlept < config.power.ls_secs) {
-        // Briefly come out of sleep long enough to blink the led once every few seconds
-        uint32_t sleepTime = 30;
-
         // If some other service would stall sleep, don't let sleep happen yet
         if (doPreflightSleep()) {
+            // Briefly come out of sleep long enough to blink the led once every few seconds
+            uint32_t sleepTime = 30;
+
             setLed(false); // Never leave led on while in light sleep
             esp_sleep_source_t wakeCause2 = doLightSleep(sleepTime * 1000LL);
 
--- a/src/PowerFSM.cpp
+++ b/src/PowerFSM.cpp
@@ -172,7 +172,13 @@ static void powerEnter()
     } else {
         screen->setOn(true);
         setBluetoothEnable(true);
-        screen->print("Powered...\n");
+        // within enter() the function getState() returns the state we came from
+        if (strcmp(powerFSM.getState()->name, "BOOT") != 0 &&
+            strcmp(powerFSM.getState()->name, "POWER") != 0 &&
+            strcmp(powerFSM.getState()->name, "DARK") != 0)
+        {
+            screen->print("Powered...\n");
+        }
     }
 }
 
@@ -189,7 +195,8 @@ static void powerExit()
 {
     screen->setOn(true);
     setBluetoothEnable(true);
-    screen->print("Unpowered...\n");
+    if (!isPowered())
+        screen->print("Unpowered...\n");
 }
 
 static void onEnter()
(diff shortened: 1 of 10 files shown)


thebentern (approved):
Great work!
I've noticed the powered/unpowered issue for a long time and never dug into it.

mverch67:
Hold on... I need to (re-)test the router mode...

caveman99:
> Hold on... I need to (re-)test the router mode...

marked as draft, give us a ring when it's ready ...

src/PowerFSM.cpp
```diff
@@ -249,7 +254,7 @@ void PowerFSM_setup()
     // Handle press events - note: we ignore button presses when in API mode
     powerFSM.add_transition(&stateLS, &stateON, EVENT_PRESS, NULL, "Press");
     powerFSM.add_transition(&stateNB, &stateON, EVENT_PRESS, NULL, "Press");
-    powerFSM.add_transition(&stateDARK, &stateON, EVENT_PRESS, NULL, "Press");
+    powerFSM.add_transition(&stateDARK, isPowered() ? &statePOWER : &stateON, EVENT_PRESS, NULL, "Press");
```
caveman99, line 257: not sure if this is evaluated every time it is called or only once at startup. If the latter is the case, we need to move the conditional.

src/graphics/Screen.cpp
```diff
@@ -1206,7 +1206,7 @@ void Screen::setFrames()
     LOG_DEBUG("Total frame count: %d\n", totalFrameCount);
 
     // We don't show the node info our our node (if we have it yet - we should)
-    size_t numnodes = nodeStatus->getNumTotal();
+    size_t numnodes = nodeDB.getNumNodes();
```
caveman99, line 1209: This is coming from the same source of truth. What is the reason you move the call one step down? This can also be made current with a forceUpdate call iirc...

caveman99 (changes requested):

src/PowerFSM.cpp
mverch67: The transition is only processed when event and "from state" match, in this case a press button event, which happens very rarely. The code behind isPowered() looks simple enough to be optimized by the compiler to a few assembly instructions. I don't see here much gain for any kind of optimisations. I guess there are many other low-hanging fruits at other places in the code.

src/graphics/Screen.cpp
mverch67: The reason for the problem is the order of initialisation in main.cpp and the interdependencies between the two modules nodeDB and screen.
nodeDB is initialized first, creates the nodes and notifies observers. At that time screen is not subscribed because it's not initialized. When the observers in screen are initialized, it takes time to update the node number until the first notifiy arrives (typically within one minute, sometimes much longer). The change in the code will immediately display the right number of known nodes (due to PR #2404).

The forceUpdate works in the other direction, from nodeDB to screen. But how could screen ask for an update via that mechanism?
As screen is already directly(!) accessing attributes of nodeDB (updateGUI), I think for the sake of simplicity and user-friendlyness that is gained screen may also request the node number directly.

caveman99: yes, i was thinking backwards here, good to go ... :-)

Issue #2404, opened 2023-04-02
[Bug]: nodes are not always stored persistently


mverch67:
### Category

Other

### Hardware

T-Beam

### Firmware Version

2.1.6

### Description

After the device reboots most nodes are gone from NodeDB and it takes a while until all the nodes show up again.
This is especially annoying during a conversation with a remote partner, as after every reboot (sometimes sudden especially on esp32-S3 when sending text messages) you'll have a long coffee break.

The nodes should be saved persistently so that known nodes are available immediately after reboot.

PR #2405, opened 2023-04-02
fix: store NodeDB persistently

mverch67:
This one-liner is a fix for #2404.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -710,9 +710,8 @@ void NodeDB::updateUser(uint32_t nodeId, const meshtastic_User &p)
         powerFSM.trigger(EVENT_NODEDB_UPDATED);
         notifyObservers(true); // Force an update whether or not our node counts have changed
 
-        // Not really needed - we will save anyways when we go to sleep
         // We just changed something important about the user, store our DB
-        // saveToDisk();
+        saveToDisk(SEGMENT_DEVICESTATE);
     }
 }
 
--- a/src/mesh/Channels.cpp
+++ b/src/mesh/Channels.cpp
@@ -260,6 +260,9 @@ const char *Channels::getName(size_t chIndex)
             case meshtastic_Config_LoRaConfig_ModemPreset_LONG_FAST:
                 channelName = "LongFast";
                 break;
+            case meshtastic_Config_LoRaConfig_ModemPreset_LONG_MODERATE:
+                channelName = "LongMod";
+                break;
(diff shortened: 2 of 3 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -710,9 +710,8 @@ void NodeDB::updateUser(uint32_t nodeId, const meshtastic_User &p)
         powerFSM.trigger(EVENT_NODEDB_UPDATED);
         notifyObservers(true); // Force an update whether or not our node counts have changed
 
-        // Not really needed - we will save anyways when we go to sleep
         // We just changed something important about the user, store our DB
-        // saveToDisk();
+        saveToDisk(SEGMENT_DEVICESTATE);
```
caveman99, line 714: The old call was in doDeepSleep (sleep.cpp 204) and was saving everything. Is devicestate enough here?

mverch67: I guessed so, because this function is triggered (only) when a node changes. And the DEVICESTATE segment contains the node info. Other segments are handled during configurations already where we don't have issues.

thebentern: I did some more digging into this and noticed another (albeit minor) side effect:
MeshService::reloadOwner calls this NodeDB::updateUser method and it has it's own saveToDisk(SEGMENT_DEVICESTATE) call. This would result in an unnecessary write to disk under certain circumstances

caveman99: > I did some more digging into this and noticed another (albeit minor) side effect: MeshService::reloadOwner calls this NodeDB::updateUser method and it has it's own saveToDisk(SEGMENT_DEVICESTATE) call. This would result in an unnecessary write to disk under certain circumstances

In turn that would mean removing the call in reloadOwner. Since every call we make to reloadOwner saves devicestate immediately afterwards anyway.

caveman99: ref: AdminModule.cpp 234 and 612,

mverch67: > > I did some more digging into this and noticed another (albeit minor) side effect: MeshService::reloadOwner calls this NodeDB::updateUser method and it has it's own saveToDisk(SEGMENT_DEVICESTATE) call. This would result in an unnecessary write to disk under certain circumstances
> 
> In turn that would mean removing the call in reloadOwner. Since every call we make to reloadOwner saves devicestate immediately afterwards anyway.

yes, that makes sense... we don't need three saveToDisk when the owner name changes;-)

PR #2426, opened 2023-04-12
use BSEC lib for BME680

caveman99:


src/modules/Telemetry/Sensor/BME680Sensor.cpp
```diff
@@ -1,18 +1,26 @@
...
 
 int32_t BME680Sensor::runOnce()
 {
-    LOG_INFO("Init sensor: %s\n", sensorName);
+    LOG_INFO("Init sensor: %s with the BSEC Library\n", sensorName);
     if (!hasSensor()) {
         return DEFAULT_SENSOR_MINIMUM_WAIT_TIME_BETWEEN_READS;
     }
-    status = bme680.begin(nodeTelemetrySensorsMap[sensorType]);
+    bme680.begin(nodeTelemetrySensorsMap[sensorType], Wire);
```
caveman99, line 15: How does the new I2C scan work if this is detected on the secondary bus? Right now all sensors assume 'Wire'

PR #2456, opened 2023-05-02
Disable TX/RX EN in favor of power EN over TX_EN

thebentern:


caveman99 (approved):

ReiHof (commented):
Why this change ?  Diy Nano_VHF with Ebyte E22 900M30S  works with RXEN/TXEN on IO 14/13
With firmware V2.1.11 a continous current of 600mA will result, 
the modul is heading over 50 degrees.

thebentern:
> Why this change ? Diy Nano_VHF with Ebyte E22 900M30S works with RXEN/TXEN on IO 14/13 With firmware V2.1.11 a continous current of 600mA will result, the modul is heading over 50 degrees.

That's not good.  I'll take a look into this. 
The reason for the change was that the 900M30S was not outputting the full 30dBm due to the RF ant switch not being properly triggered by the TX/RXEN pins. This was presumably due to the same scheme as some other SX1262 based targets such as the RAK-4631, wherein this pin is used to power dio2 which automatically switches modes internally. It would appear this is not so cut and dry. I need to investigate and see after restoring the TX/RXEN pin definitions if we cannot figure out why the switch is not getting triggered by RadioLib as it should. Perhaps it's a configuration issue or perhaps a bug.

thebentern:
@ReiHof, please try the PR build artifacts here to test the fix:
https://github.com/meshtastic/firmware/pull/2484

PR #2467, opened 2023-05-05
use BSEC2

caveman99:
moving over for other archs as well. Step by step.


variants/ai-c3/platformio.ini
mverch67, line 1: in file variant.h: also remove the defines for telemetry and sensors

PR #2476, opened 2023-05-10
update portduino to radiolib6

caveman99:


caveman99:
The build script is pulling master ... biting the bullet here.

PR #2480, opened 2023-05-10
Up OneButton library version to 2.1.0

IhorNehrutsa:


caveman99:
While i am all for this change, can we wait a couple of days more till this version shows up in the PIO registry rather than pulling it from github?

IhorNehrutsa:
As you wish.

PR #2485, opened 2023-05-12
Fix LOG_DEBUG messages when no DEBUG_PORT.

IhorNehrutsa:
Fix messages like
image
image

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -464,8 +464,9 @@ void setup()
 
     if (gps)
         gpsStatus->observe(&gps->newStatus);
-    else
+    else {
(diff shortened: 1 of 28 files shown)


src/main.cpp
```diff
@@ -464,8 +464,9 @@ void setup()
 
     if (gps)
         gpsStatus->observe(&gps->newStatus);
-    else
+    else {
```
caveman99, line 467: If you put parentheses here, do it on both legs of the statement :-)

caveman99 (changes requested):

caveman99 (approved):

thebentern (approved):

PR #2495, opened 2023-05-16
Remote hardware overhaul (WIP)

thebentern:


src/mesh/NodeDB.cpp
```diff
@@ -349,6 +349,10 @@ void NodeDB::init()
     if (channelFileCRC != crc32Buffer(&channelFile, sizeof(channelFile)))
         saveWhat |= SEGMENT_CHANNELS;
 
+    if (!devicestate.node_remote_hardware_pins) {
+        memset(devicestate.node_remote_hardware_pins, 0, sizeof(devicestate.node_remote_hardware_pins));
+    }
```
caveman99, line 354: you can also directly assign meshtastic_ModuleConfig_RemoteHardwareConfig_init_zero

caveman99: or [...]_init_default for that matter

caveman99 (approved):

src/mesh/NodeDB.cpp
thebentern: Can't reinitialize arrays in C :-(

Issue #2497, opened 2023-05-17
[Bug]: T-Echo "Shutting Down" text is cropped


pdxlocations:
### Category

Hardware Compatibility

### Hardware

T-Echo

### Firmware Version

2.1.11

### Description

![20230517_144327](https://github.com/meshtastic/firmware/assets/117498748/960a6526-c77c-497e-ac89-5720f2161955)
When shutting down the T-Echo, the "Shutting Down..." text crops the first two letters.  I have the same issue on two T-Echo's

markbirss:
https://github.com/meshtastic/firmware/pull/2499

PR #2499, opened 2023-05-18
Fix https://github.com/meshtastic/firmware/issues/2497

markbirss:


thebentern (approved):

thebentern:
Looks good. Just need to make trunk happy :-)

mverch67:
Doesn't look that good to me... the code should use SCREEN_WIDTH instead of hard coded numbers.

PR #2510, opened 2023-05-23
Adding new DYI variant

clownjoy:
v1_2  -  DevKit C4 / Lolin32 / Lolin32 lite

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -2,7 +2,7 @@
 ; https://docs.platformio.org/page/projectconf.html
 
 [platformio]
-default_envs = tbeam
+;default_envs = tbeam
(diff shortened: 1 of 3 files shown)


caveman99 (changes requested):
Thanks for adding this one. Maybe we should pick a different name, the versioning implies some kind of progress between them. I suggest devkit-c4 to stay in line with the dr-dev board definition. The existing v1/v1_1 naming scheme is indeed an evolution of the identical board design.

clownjoy:
I'm closing this pull, request because I just realized that dyi/v1_1 actually  fits the needs  for both  boards (C4 and LoLin32).  I'm sorry for confusion.

PR #2529, opened 2023-05-29
Fix t-beam-s3-core BME280 wrong recognition as BMP280

lewisxhe:
In `t-beam-s3-core`, the IC type is obtained by reading the BMX register, and the sensor returns as BMP280. Although the method of reading the ID is correct, the actual use in `t-beam-s3-core` is BME280. Add compile macro to force change to BME280


Tom9876:
Thanks for taking care. Your fix works for me.

PR #2530, opened 2023-05-30
Account for AO7801/3401 mosfet

code8buster:
The Heltec v3, Wireless Stick v3, and Wireless Stick Lite v3 all have some version of this P channel mosfet tied to GPIO37.

This pull request makes sure that when we init the ADC code, the pin is not in high impedance and is LOW.


prokrypt:
Works great on Heltec WSnotL! Battery reading is actually useful now, and no more premature low-voltage shutdowns.

PR #2533, opened 2023-05-31
Remove macro guards for T-Echo

thebentern:


caveman99 (approved):
LGTM, might wanna add to the module docs that the t-echo only works in override mode.

PR #2534, opened 2023-05-31
Enable range test module (sending only) on NRF

### LINKED ISSUE #1451: 1.3 [Feature Request]: nRF52 support for Range Test module sender
I would like to see nRF52 support for the Range Test module, especially as these devices are becoming more popular. Additionally, since the stock PCB antenna for the RAK 4631 is a PCB antenna with only uFL connector, typical VSWR measurement tools cannot be used without special adapters.

thebentern:
Closes #1451


src/modules/Modules.cpp
```diff
@@ -82,9 +82,11 @@ void setupModules()
 
         storeForwardModule = new StoreForwardModule();
 
-        new RangeTestModule();
 #elif defined(ARCH_NRF52)
         externalNotificationModule = new ExternalNotificationModule();
+#endif
+#if defined(ARCH_ESP32) || defined(ARCH_NRF52)
+        new RangeTestModule();
```
caveman99, line 89: Not scope, but you can shove the externalNotification init there as well for clarity, removing the dual definition and elif above. :-)

caveman99 (changes requested):

src/modules/Modules.cpp
thebentern: Good call. We've got enough intense macro gaurding in there

PR #2535, opened 2023-05-31
Broadcast neighbor info

caveman99:
also update trunk


GUVWAF:
Logic seems okay now. (Node 0 has NodeNum 16, Node 1 has 17, Node 2 has 18).

image

Still needs some more work, e.g. handling the AdminMessages for configuring the module and sending only the `n` most relevant edges.

@ajmcquilkin There is a `MAX_NEIGHBOR_AGE` of 10 minutes defined, but neighbors are not yet removed. You mentioned "Clients will need to individually set the timeout duration", but that would require another setting. I suggest we set this to 2x the broadcast interval (which is configurable), do you agree?

ajmcquilkin:
@GUVWAF I think 2x the `MAX_NEIGHBOR_AGE` would be perfect, although as long as this is a client-side duration all we would need to do is document this expectation somewhere. I'm assuming that this would need to be 2x the broadcast interval of the node that the client is connected to.

This also makes me wonder whether 2x should be the default for the client, but where the user can override this default duration in the application settings. I'm envisioning an application override being necessary since the user may know that the broadcast interval of the connected node is much slower or faster than most nodes on the network.

A random idea would be for nodes to somehow advertise their timeout when they enter the network, but 1) that sounds out of scope of this PR and 2) there still wouldn't be a guarantee that all nodes would receive this advertisement.

GUVWAF:
@ajmcquilkin If you want it to be configurable in the client, we can also just not remove any neighbors in the firmware (like it is now, so easiest to implement) except when one does the NodeDB reset. Either way, I think then we still need to store the last Rx time and send this to the client. Then you can decide in the client whether a link should be removed. You could also more easily warn a user not to set it lower than the broadcast interval of any node, because then you might remove it too soon. 

I know I voted against transmitting the Rx time over the mesh and I think that's also not necessary, but storing it will be needed if you want to remove any links after a certain time.

Unfortunately I can't attend the meeting on Thursday, but maybe you can discuss this there.

D4rk4:
Any chance to get this feature in next release?

caveman99:
This will go into 2.2.x - rebasing

PR #2536, opened 2023-05-31
Use INA for device battery level

### LINKED ISSUE #2514: Allow INA-2XX sensors to be authoritative source for device battery level
### Platform

NRF52

### Description

This functionality will use a new module config protobuf `telemetry.device_battery_ina_address`, which will specify an i2c address of a connected INA_2XX sensor to use as the authoritative source of device battery voltage instead of merely as an environmental telemetry voltage source.
The feature will be piloted in KeithMon's bespoke solar LTO charging system.

thebentern:
Closes #2514


code8buster (approved):
Does it make sense to remove the macro repetition in Power::getINAVoltage() with a call to hasINA() ? Either way looks good to me.

tobiascrackz:
I’m facing the same problem although using current firmware. Seems like the modifications have been undone?

PR #2544, opened 2023-06-06
Fix static ethernet config

### LINKED ISSUE #2543: [Bug]: Static IP configuration doesn't work
### Category

Other

### Hardware

Rak4631

### Firmware Version

2.1.11

### Description

Connect a RAK with an ethernet module and DHCP enabled, it will successfully get an IP address (e. g. 192.168.1.223) and can be pinged.

Switch to the static IP configuration, set the exact same IP address (192.168.1.223) and appropriate subnet mask (255.255.255.0) and you can no longer ping the device.

### Relevant log output

_No response_

micheljung:
With static ethernet config, `status` stayed `0` which let the function return without setting `ethEvent`. Therefore, `reconnectETH` was never called and network services were never started.

Fixes #2543

### DIFF
--- a/src/mesh/eth/ethClient.cpp
+++ b/src/mesh/eth/ethClient.cpp
@@ -98,6 +98,11 @@ static int32_t reconnectETH()
     return 5000; // every 5 seconds
 }
 
+static uint32_t bigToLittleEndian(uint32_t value)
+{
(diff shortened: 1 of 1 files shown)


micheljung:
Seems like there are more issues, I'll report back when I successfully tested it.

thebentern (approved):

thebentern (approved):

PR #2552, opened 2023-06-07
Remove deprecated MyNodeInfo fields

thebentern:


src/airtime.cpp
```diff
@@ -6,20 +6,24 @@ AirTime *airTime = NULL;
 
 // Don't read out of this directly. Use the helper functions.
 
+#define MAX_PERIODS 8
```
GUVWAF, line 9: Maybe better to use `PERIODS_TO_LOG` here, which is already defined in the header file.

GUVWAF (approved):

thebentern:
Smoke tested on T-Beam with update and pairing to app. Seems to be in good shape

PR #2571, opened 2023-06-25
RAK14001 LED - Turn on to 50% at boot

charminULTRA:
Enables the RAK14001 to turn on to a static 50% white upon boot, and creates a central location (RAKled.h) for defining "rgb". 

I could use help in one aspect of this - I don't know C++ fundamentals, so I suspect the organization of code between this and the ExternalNotificationModule could be better organized. I'm not 100% sure if I used best practices, but I tried moving the declaration of "rgb" into the RAKled.h file to give it a more centralized location rather than the Module's file.

In the future, I would like to add the ability to control the LED levels/colors/cycles via the CLI and apps.

### DIFF
--- a/src/graphics/RAKled.h
+++ b/src/graphics/RAKled.h
@@ -1,0 +1,7 @@
+#include "main.h"
+
+#ifdef RAK4630
+#include <NCP5623.h>
+extern NCP5623 rgb;
+
+#endif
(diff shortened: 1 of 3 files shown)


charminULTRA:
I'll also note that I live tested with on a RAK 4631

PR #2573, opened 2023-06-25
fixes #2572 remove bad assignment, fix random ranges

### LINKED ISSUE #2572: [Bug]: NodeDB::pickNewNodeNum() Returns bad values
### Category

Hardware Compatibility

### Hardware

T-Lora v2 1.6

### Firmware Version

v2.1.16.a2c5b92

### Description

It seems that I was the unlucky one and kept hitting the `FIXME` issue in the node selection on my board. The `pickNewNodeNum()` looks like it might have had a bad merge for this area so I'll submit a PR for you.

### Relevant log output

```Shell
The repeated behavior is that you get a lot of logs that are just in a loop of:

`NOTE! Our desired nodenum 0x4 is in use, so trying for 0x4\n`
```

wyattearp:
Addresses the following
- removes useless assignment that is immediately overwritten
- moves to past the `NUM_RESERVED` so that we're actually not picking exactly what everyone else would be picking (low likelihood of happening right?)
- Updates the selection of `random()` to address the `FIXME` by using a number with the same width for a max.

NOTE: this creates an artificial limit on the max selected node number to be within half the range; however, it seems safer than continuing to send in a -1 and failing the function.

There should also likely be a "fail and reboot" at this level because after a certain amount of time, we'll have drained all the randomness from the system and should just reboot and try again.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -376,18 +376,17 @@ void NodeDB::init()
  */
 void NodeDB::pickNewNodeNum()
 {
-    NodeNum r = myNodeInfo.my_node_num;
-
     // Pick an initial nodenum based on the macaddr
-    r = (ourMacAddr[2] << 24) | (ourMacAddr[3] << 16) | (ourMacAddr[4] << 8) | ourMacAddr[5];
+    NodeNum r = (ourMacAddr[2] << 24) | (ourMacAddr[3] << 16) | (ourMacAddr[4] << 8) | ourMacAddr[5];
 
-    if (r == NODENUM_BROADCAST || r < NUM_RESERVED)
-        r = NUM_RESERVED; // don't pick a reserved node number
+    if (r == NODENUM_BROADCAST || r < NUM_RESERVED) {
+        r += NUM_RESERVED; // don't pick a reserved node number
+    }
 
     meshtastic_NodeInfoLite *found;
     while ((found = getMeshNode(r)) && memcmp(found->user.macaddr, owner.macaddr, sizeof(owner.macaddr))) {
-        // FIXME: input for random() is int, so NODENUM_BROADCAST becomes -1
-        NodeNum n = random(NUM_RESERVED, NODENUM_BROADCAST); // try a new random choice
+        // NOTE: input for random() is int, so NODENUM_BROADCAST becomes -1, use INT32_MAX as it is the same width
+        NodeNum n = random(NUM_RESERVED, INT32_MAX); // try a new random choice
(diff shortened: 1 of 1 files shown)


thebentern (approved):

GUVWAF:
Thanks for the contribution. However, multiple users have reported this issue on Discord since version 2.1.12. The chance that you are actually picking the same NodeNum is very low since it's a 32-bit number based on the MAC address.  

I deliberately put the FIXME there last week instead of fixing it, because if we fix only this, we still have a bug, but it will be less noticeable, as a node will pick a new random NodeNum. Others will then think there is a new node while there actually isn’t. 

Somehow the device ends up in this clause because it mixes `my_node_num` with that of another node. If we find the cause of that, we can merge this I think. To fix the artificial limit, we could call `random(INT32_MIN, IN32_MAX)` and then check afterwards if we didn't pick `r == NODENUM_BROADCAST || r < NUM_RESERVED`.

wyattearp:
That's fair - I was dropping a quicker fix, but I'll take a look at it again to see if I can understand the underlying issue.

wyattearp:
After some further digging, I believe the reason that the original defect of dropping into the `while()` loop and then just getting `4` repeatedly is likely failing because `pickNewNodeNum()` can be called twice, but is likely only called once. The second time, which might be a weird scenarios that the `ourMacAddr` isn't initialized, which results in `0` that then gets pushed to `4` and the loop goes and does it's thing.

`NodeDB::init()`, which will first call `loadFromDisk()` which then calls `loadProto()` which can fail and then call `installDefaultDeviceState()`. If it *normally* fails, this calls the `getMeshNode()` which returns a `NULL` and fails the loop, then assigns the node number. Once the load returns, there's a subsequent call to `pickNewNodeNum()`, which might result in a change, but that doesn't really matter because call to store the data in `getOrCreateMeshNode(getNodeNum())` does the add to the database.

Where it gets weird is if `loadProto()` succeeds in some weird fashion, `ourMacAddr` is left uninitialized because `installDefaultDeviceState()` is never called, which would result in the error of always driving everything thing to `4` because if you just keep shifting `0` in, you're going to get `0`.

Reviewing the NodeDB code though, I'm not sure there's a good way to address it without cleaning up a lot more of the code. A quick "hack" that would probably remove it would be to always force the call to `getMacAddr(ourMacAddr)` inside of the `pickNewNodeNum()`, removing the other ... but that also feels like it's not addressing the real challenge of the initialization path being overloaded for fresh start and then all the other subsequent starts (where it tries to recover parts of the configuration). That feels like a series of design decision I'm not qualified to make :)

wyattearp:
See #2576

PR #2574, opened 2023-06-25
GPS: Performance improvment for U-Blox hardware

D4rk4:
In this Pull Request, I propose improvements to our devices' Global Navigation Satellite System (GNSS) performance by adapting the navigation engine's configuration.

Here are the key changes:

- Switching to GPS + SBAS (Satellite-Based Augmentation System) + GLONASS for enhanced stability. By leveraging multiple satellite systems, we can ensure more robust satellite coverage and mitigate the effects of signal blockages or interferences that might occur in certain environments.
- Enabling the AssistNow Autonomous feature. This allows the GNSS module to predict satellite positions based on previously collected ephemeris data, which in turn helps to reduce the Time-To-First-Fix (TTFF). It offers better performance in weak signal conditions and helps to achieve a faster startup.
- Turning on the anti-jamming feature. This adds an extra layer of protection to our GNSS module by making it more resilient to intentional or unintentional radio-frequency interferences, hence ensuring a consistent GNSS performance.
- Making sure the new configuration persists across reboots.

By integrating these improvements, I am confident that the GNSS performance of our devices will be significantly enhanced, leading to better stability, faster TTFF, and more reliable functionality under varying signal conditions.

Please review these changes and provide your feedback. These adjustments will greatly contribute to the system's performance and reliability.


charminULTRA:
Did you make sure to use Trunk? That was causing my checks to fail because of formatting issues.

src/gps/GPS.cpp
```diff
@@ -192,6 +198,147 @@ bool GPS::setupGPS()
...
+            // Calculate the checksum, starting from the CLASS field (which is _message_GNSS[2])
+            for (size_t i = 2; i < sizeof(_message_GNSS) - 2; i++) {
+                CK_A = (CK_A + _message_GNSS[i]) & 0xFF;
+                CK_B = (CK_B + CK_A) & 0xFF;
+            }
+
+            // Place the calculated checksum values in the message
+            _message_GNSS[sizeof(_message_GNSS) - 2] = CK_A;
+            _message_GNSS[sizeof(_message_GNSS) - 1] = CK_B;
+
```
caveman99, line 235: My suggestion would be to make a function to calculate the checksum of a buffer and send to the chip. This block is repeated during init

caveman99 (changes requested):
Great PR, if we can clean it up a bit it can go in!

src/gps/GPS.cpp
code8buster: Added in 5b75af1

thebentern (approved):

kokroo:
@D4rk4 Which Ublox chips does it work for exactly? Were you able to notice any performance improvements in the real world?

PR #2582, opened 2023-07-02
Initial support for MonteOps's fixed hardware platform

D4rk4:
This merge request provides the initial implementation for the MonteOps M.Node fixed hardware design. The specific hardware features we're addressing in this configuration include:

- The RAK4630 core MCU
- An independent hardware watchdog, which, though not firmware-dependent, is a critical feature we're excited to share
- An onboard BME280 sensor for the remote monitoring of the hardware state
- The W5100SL embedded Ethernet, combined with a USB bridge to the LAN7850 module
- The U-Blox EVA-M8E GPS module

Due to the unique hardware configurations required for the W5100SL and EVA-M8E, these changes have prompted us to create a distinct firmware variant. As we continue to develop and refine this firmware, we plan to add support for external flash handling for the RAK4630, which will also be managed separately.


mc-hamster:
wow

PR #2584, opened 2023-07-02
Ambient Lighting Module

charminULTRA:
Creates the "Ambient Lighting Module", a new module dedicated to having a place for controlling ambient LED lights (rather than status lights and activity lights) being powered by the board. An example is the RAK14001, which is the primary use for this module right now.

Primary controls of the Module will include:
1) Brightness (current) control
2) Individual RGB level controls

DRAFT WORK:
1) Waiting on Protobuf PR merge - then can fully implement the firmware-side code
2) Make sure I use the new NCP references rather than RAK4630


charminULTRA:
RAKled.h contents were moved to the Ambient Lighting Module.

PR #2588, opened 2023-07-06
Temporary band-aid to address mesh [un]reliability after queue "fix"

prokrypt:
This lets routers/repeaters rebroadcast a packet that is in queue even if another node has already rebroadcasted, ensuring that a router/repeater in a good location does not get its chance to transmit taken away by another node in a worse location.

Positive things: much better mesh reliability in cases when there are two or more routers in range of each other and one of them is at a strategic location (linking two groups of nodes together, etc).

Negative things: possibly more collisions, unless router/repeater nodes are positioned well and not all clumped together.

Am I doin' it right?


GUVWAF (approved):
Looks good, makes sense to not cancel for a router/repeater as you're not supposed to have many of them close together.

PR #2592, opened 2023-07-07
PIN_GPS_EN power toggling

### LINKED ISSUE #2591: [Feature Request]: Add ability to toggle power to GPS via GPIO high/low setting
### Platform

NRF52, ESP32

### Description

Request ability to toggle external GPS module (e.g. GT-U7) by setting GPIO to high or low state, so connected GPS modules can be powered on and off as needed.  I have successfully proven this to be possible using a Heltec v3, a GT-U7 GPS module, and the External Nofitications module to set GPIO36 to HIGH.  A 2N2222 transistor was used as a trigger/gate to provide power to the GPS module from a 3V3 pin on the Heltec v3 board.  Such a transistor is required for this to work.

I think using the same 'triple user button press' action already used for toggling GPS for T-Beam, would be best so we have the same action for multiple target devices.

*Note:* The Heltec boards (v2.1, v3, and WSLv3) include a Vext pin that is supposed to be able to be set to high/low if GPIO36 is toggled, but if an OLED screen is present (as it is with the Heltec v3) then Vext will still maintain a residual voltage (1.44v) even when set to low where it should have no voltage at all.  If the OLED is disabled by shorting GPIO21 (OLED_RST) to GND then Vext reports 0v but this is not feasible is the user wants to see the device screen.

The above use of Vext *would* be possible when using the screenless WSLv3 board, but it would be better to implement a universal GPIO trigger/toggle capability that any device can use, rather than tailoring a specific use case (toggle Vext pin) for a specific board (Heltec series).

thebentern:
@tropho23 make sure this doesn't affect RAK setups negatively. It looks like we are using this for peripherals on that platform

Fixes #2591

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -423,7 +423,7 @@ bool GPS::setup()
 {
     // Master power for the GPS
 
-#ifdef HAS_PMU || PIN_GPS_EN
+#if defined(HAS_PMU) || defined(PIN_GPS_EN)
     if (config.position.gps_enabled) {
 #ifdef PIN_GPS_EN
         pinMode(PIN_GPS_EN, OUTPUT);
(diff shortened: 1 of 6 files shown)


code8buster (approved):
Nice, minimal changes for the desired result. Approved!

PR #2594, opened 2023-07-08
Partial Heltec Wireless Paper and Wireless Tracker support

thebentern:
Screens still aren't working


mverch67:
> Screens still aren't working

I think the cause may be the TFT_eSPI library. If in main.cpp I place the tft.init() above the I2C scanning the display works well, but LoRa doesn't. If the tft.init() is placed below the I2C scanning LoRa work but not the display. Looks like an SPI issue handling both SPI devices at the same time.

markbirss:
Further clues as to which e-ink display is actually used might be visible on the flex cable or on the back of the display itself (will need to carefully detach is, likely stuck on with double sided scotch tape)

image

Found these links 
https://resource.heltec.cn/download/e-ink/213/2.13b%26w/HTEW0213T5_V2

The DEPG0213BN is the GxEPD2_213_BN

I also tried to no avail in the past ST7735/ST7789/ILI9xxx display in the past and could only get a black/grey picture even tried on nrf52 also

variants/heltec_wireless_tracker/variant.h
```diff
@@ -0,0 +1,53 @@
...
+#define BUTTON_PIN 0
+
+#define BATTERY_PIN 1 // A battery voltage measurement pin, voltage divider connected here to measure battery voltage
+#define ADC_CHANNEL ADC1_GPIO1_CHANNEL
+#define ADC_ATTENUATION ADC_ATTEN_DB_2_5 // lower dB for high resistance voltage divider
+#define ADC_MULTIPLIER 4.9
+
+#undef GPS_RX_PIN
+#undef GPS_TX_PIN
+#define GPS_RX_PIN 33
```
mverch67, line 28: Are these RX/TX pins correct? In the heltec-tracker pinout it is TX=33, RX=34

markbirss: According to this pins_arduino.h - https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/master/esp32/variants/wireless_paper/pins_arduino.h, you could leave them under and test

thebentern:
> Further clues as to which e-ink display is actually used might be visible on the flex cable or on the back of the display itself (will need to carefully detach is, likely stuck on with double sided scotch tape)

FPC-7528B on the flex cable

variants/heltec_wireless_tracker/variant.h
mverch67: This would be the proper pins_arduino.h for the tracker;
https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/master/esp32/variants/wireless_track/pins_arduino.h

And there is and TX=33 RX=34 definition.

markbirss:
> FPC-7528B

Ok, according to these to sources that is the GxEPD2_213_BN display (DEPG0213BN) same as RAK 2.13 b/w e-ink

https://forum.arduino.cc/t/solved-waveshare-pico-epaper-2-13-on-arduino-pro-mini-3-3v-with-gxepd2-library/1120833/4

https://www.aliexpress.com/i/1005005199511879.html

@thebentern can you share a current boot log for the wireless tracker with #define TECHO_DISPLAY_MODEL GxEPD2_213_BN ?

mverch67:
I got the tracker display working with the lovyanGFX library:

image

markbirss:
@mverch67 

Well done

mverch67:
I found the issue with the Heltec GPS: 

PIN_GPS_RESET must be set to HIGH for a functional serial communication.

variants/heltec_wireless_tracker/variant.h
mverch67: #define GPS_RX_PIN 33
#define GPS_TX_PIN 34
 is correct

PR #2607, opened 2023-07-12
Fix various typos

luzpaz:
Found via `codespell -q 3 -L acount,clen,dout`


code8buster (approved):

thebentern:
@luzpaz thanks for the fixes. Just as an FYI, those generated files come from the protobuf comments, so those fixes will disappear the next time we have a change there and regenerate the classes.

PR #2608, opened 2023-07-12
Add Hydra specific target to define GPS EN pin and limit tx power

thebentern:


variants/diy/hydra/variant.h
```diff
@@ -0,0 +1,41 @@
...
+#define RF95_NSS 18
+
+#define USE_SX1262
+
+#define SX126X_CS 18 // NSS for SX126X
+#define SX126X_DIO1 LORA_DIO1
+#define SX126X_BUSY LORA_DIO2
+#define SX126X_RESET LORA_RESET
+
+// Set lora.tx_power to 13 for Hydra or other E22 900M30S target due to PA
```
kokroo, line 39: @thebentern Why is the power for Ebyte chips limited to 22 here, even when they can go to 30? In the comment, it mentions it's because of the PA, but I am not sure how that's affecting this.

PR #2617, opened 2023-07-17
Added triple-press GPS toggle button changes for select ESP32 devices

tropho23:
Added triple-button press to disable GPS power for the below targets:

1. T3S3 v1: uses pin 42 for trigger
2. Tlora 2.1-1.6: uses pin 19 for trigger
3. Heltec v2.1: uses pin 37 for trigger

I have tested all of these changes on real hardware with no issues noted.

We could extend this to older ESP32 devices, but I don't have really old Tlora boards (v1/v2) to test.  The above list, plus the existing triple-button press already implemented for T-Beam v1.1 should cover the vast majority of users.

This does not affect RAK boards, where toggling 3v3_S (by shorting IO2 to GND) disables power to ALL modules, not just GPS.  I may submit another PR for consideration to address that.


mverch67:
extensions.json should probably not be included in this commit as it removes trunk extension recommendation.

However I have a similar issue in vscode... it permanently changes extensions.json and add it's own stuff.

.vscode/extensions.json
thebentern, line 1: Lets undo this file's changes. Otherwise, looks good.

tropho23:
Is it possible to disable/prevent this from occuring?

Tony

PR #2618, opened 2023-07-17
Fix: increase BT NIMBLE task stack size by 1k

### LINKED ISSUE #2416: [Bug]: sporadic crashes when sending text messages via phone
### Category

Stability

### Hardware

T-Lora T3S3

### Firmware Version

2.1.7

### Description

When sending text messages via phone APP the device crashes and reboots. This happens quite often!

I suspect that the channel crypto methods / CryptoEngine are not reentrant and cannot handle two send or receive requests at the same time (getKey/setKey/setCrypto), which leads to possible memory corruption and eventually a crash (w/o stacktrace on esp32-S3).

The channel crypto methods should be guarded with a concurrency lock or the send/receive logic that uses this CryptoEngine.

### Relevant log output

_No response_

mverch67:
Stacksize for task nimble_host is too small and leads to crashes when sending text messages from phone:

```
DEBUG | 21:27:08 12 [Screen] Add packet record (id=0xf0f7443c fr=0x00 to=0xff, WantAck=1, HopLim=3 Ch=0x0 Portnum=1 rxtime=1689542828)
DEBUG | 21:27:08 12 Original length - 2 
DEBUG | 21:27:08 12 [GPS] Compressed length - 2 
DEBUG | 21:27:08 12 [GPS] Original message - Hi 
DEBUG | 21:27:08 12 [GPS] Not using compressing message.
DEBUG | 21:27:08 12 [GPS] Using AES256 key!

***ERROR*** A stack overflow in task nimble_host has been detected.
```

This PR increases the NIMBLE task stacksize for all esp32 devices by 1k which  will finally fix #2416.


thebentern (approved):

thebentern:
Great fix! This sheds some light on why I experienced multiple consecutive crashes when trying to send an arbitrarily large text message from the iOS app recently.

mverch67:
> Great fix! This sheds some light on why I experienced multiple consecutive crashes when trying to send an arbitrarily large text message from the iOS app recently.

Thanks, finally this annoying bug is solved. I had this problem with the heltec-tracker, where *every* message sent from phone lead to the crash. But luckily I spotted this error message about nimble_host task.

thebentern:
> Thanks, finally this annoying bug is solved. I had this problem with the heltec-tracker, where _every_ message sent from phone lead to the crash. But luckily I spotted this error message about nimble_host task.

Annecdotal... but it _seemed_ to occur more frequently on the new S3 based devices much more frequently

PR #2624, opened 2023-07-18
fix BLE PIN screen for not so large screens

mverch67:
Screens with a display height = 80 do not properly display the BLE pin screen.

image

This PR tweaks the y_offset calculation, now it's like this:

image

### DIFF
--- a/boards/heltec_wireless_tracker.json
+++ b/boards/heltec_wireless_tracker.json
@@ -1,0 +1,38 @@
+{
+  "build": {
+    "arduino": {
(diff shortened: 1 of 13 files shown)


mverch67:
BTW: is it a bug in github that it displays all my commits of the last days under this PR? I did create a new branch for just this change.

thebentern:
> BTW: is it a bug in github that it displays all my commits of the last days under this PR? I did create a new branch just for this change.

If you created a new branch off of master HEAD, then it seems that way

PR #2632, opened 2023-07-19
T-Watch S3 Support

thebentern:


variants/t-watch-s3/variant.h
```diff
@@ -0,0 +1,90 @@
...
+#define ST7789_CS 12
+#define ST7789_RS 38  // DC
+#define ST7789_SDA 13 // MOSI
+#define ST7789_SCK 18
+#define ST7789_RESET -1
+#define ST7789_MISO -1
+#define ST7789_BUSY -1
+#define ST7789_BL 45
+#define ST7789_SPI_HOST SPI2_HOST
+//#define ST7789_BACKLIGHT_EN 15
```
mverch67, line 11: The #define ST7789_BACKLIGHT_EN 45 is probably needed so that after display.timeout the TFT is turned off

mverch67:
> @mverch67 have you had issues with PlatformIO not flashing the device with the esp-builtin toolchain? It worked great the first time, but not subsequent upload attempts failed.
> 
Yeah, I know that. There are to ways to flash successfully:
- upload_protocol = esp-builtin ; requires boot mode (press boot-button while powering on), serial terminal could stay connected
- upload_protocol = esptool: does not require boot mode, but serial terminal must be disconnected; often fails though ... then must switch back to esp-builtin

you may also try to use the "use_1200bps_touch": true ... maybe it helps

boards/t-watch-s3.json
```diff
@@ -0,0 +1,37 @@
...
+      "-DARDUINO_USB_MODE=0",
+      "-DARDUINO_RUNNING_CORE=1",
+      "-DARDUINO_EVENT_RUNNING_CORE=1"
+    ],
+    "f_cpu": "240000000L",
+    "f_flash": "80000000L",
+    "flash_mode": "dio",
+    "hwids": [["0X303A", "0x1001"]],
+    "mcu": "esp32s3",
+    "variant": "t-watch"
```
mverch67, line 20: t-watch-s3 ?!

thebentern: Good call

thebentern:
@lewisxhe are the GPIO pin references in the graphic at https://www.lilygo.cc/products/t-watch-s3 correct? I'm crashing on bootup because it's failing to init the SX1262, which 99% of the time is due to incorrect SPI pin numbers.

mverch67:
> @lewisxhe are the GPIO pin references in the graphic at https://www.lilygo.cc/products/t-watch-s3 correct? I'm crashing on bootup because it's failing to init the SX1262, which 99% of the time is due to incorrect SPI pin numbers.

you could also try to set ST7789_SPI_HOST SPI**3**_HOST as LoRa uses another SPI than TFT

thebentern:
> you could also try to set ST7789_SPI_HOST SPI**3**_HOST as TFT uses another SPI than LoRa
Ah good point. You think it might be SPI bus contention?

lewisxhe:
Screen and Radio use separate buses.

thebentern:
> Screen and Radio use separate buses.

Thanks. We got past the radio issue and made some progress on other peripherals. I'm currently running into issues getting the tft.getTouch(x,y) to return something valid

lewisxhe:
You are using Bodmer's TFT_ ESPI built-in getTouch? TFT_ ESPI does not support capacitive screens and only supports xpt2046. The watch-s3 uses ft62xx

thebentern:
> You are using Bodmer's TFT_ ESPI built-in getTouch? TFT_ ESPI does not support capacitive screens and only supports xpt2046. The watch-s3 uses ft62xx
> 
> https://github.com/Xinyuan-LilyGO/TTGO_TWatch_Library/blob/t-watch-s3/src/LilyGoLib.cpp#L123-L123

Using LovyanGFX. That tip about the touch driver was the ticket. Thanks!
We were assuming a GT911 by default.

thebentern:
@lewisxhe last question. Trying to get the DRV2605 operational... I'm able to begin and issue commands but nothing is happening. I noticed that you have BLDO2 on for that in watch library. Is there anything else we need to be doing to allow the driver to actuate on drv.go()?

// DRV2605 power channel
PMU->setPowerChannelVoltage(XPOWERS_BLDO2, 3300);
PMU->enablePowerOutput(XPOWERS_BLDO2);

lewisxhe:
> @lewisxhe last question. Trying to get the DRV2605 operational... I'm able to begin and issue commands but nothing is happening. I noticed that you have BLDO2 on for that in watch library. Is there anything else we need to be doing to allow the driver to actuate on drv.go()?
> 
> // DRV2605 power channel PMU->setPowerChannelVoltage(XPOWERS_BLDO2, 3300); PMU->enablePowerOutput(XPOWERS_BLDO2);

setWaveform(26) feels very slight and may not be felt, please try to run 75, or cycle test the most suitable value.

anemes:
Follow-up on T-watch support - Is there any likelihood canned replies on watch or keyboard for responding without phone connection is doable on the current hardware?

PR #2633, opened 2023-07-20
xiao-new: add initial support for the Xiao BLE + Ebyte E22-900M30S

andrew-moroz:
This adds initial support for the Xiao BLE + Ebyte E22-900M30S.

In `variants/xiao_ble`, I have included some additional resources in hopes that everything needed could be found under one roof:

- `README.md` is a short how-to guide.
- `xiao_ble.sh` is a script that pulls in some needed dependencies from Seed Studio's Arduino core. (Guessing there may be a better way to do this?)
- `xiao_nrf52840_ble_bootloader-0.7.0-22-g277a0c8_s140_7.3.0.zip` is the Adafruit UF2 bootloader compiled for the Xiao BLE. I included it here since the compiled version is not included with Adafruit's releases (but rather [here](https://github.com/adafruit/Adafruit_nRF52_Bootloader/tree/master/src/boards)) and the compilation process is somewhat arcane.
- `xiao-ble-internal-format.uf2` is identical to `Meshtastic_nRF52_factory_erase.uf2` included in Meshtastic releases, but compiled for softdevice 7.3.0.

Not sure if all of this really belongs in the variant folder. I like the idea of making it easy for first-time DIYers by keeping it all in one place, but happy to move stuff if the feeling is that this is too much bloat 😀

Beyond the files in the `variants/xiao_ble` folder, I also made some small changes to SX126xInterface.cpp to support using the E22 in 'automatic antenna switching' mode, which is nice since it frees up an additional GPIO. These should not affect any existing variants, but a careful review here is definitely in order since the logic is common to all SX1262 builds, most of which I don't have on hand to test.

Thanks in advance for giving this a look!

** EDIT: Working on those trunk checks. I'm a bit behind the times and didn't have the VS Code extension installed until just now.


variants/xiao_ble/eagle-project/xiao-ble-e22-schematic.png
thebentern, line 1: Let's remove the schematics from the project

thebentern: You may consider hosting them on a Github repo and linking them

src/mesh/SX126xInterface.cpp
```diff
@@ -70,12 +70,25 @@ template <typename T> bool SX126xInterface<T>::init()
...
+        LOG_DEBUG("SX126X_TXEN pin defined. Setting RF Switch: RXEN=%i, TXEN=%i\n", SX126X_RXEN, SX126X_TXEN);
         res = lora.setDio2AsRfSwitch(false);
         lora.setRfSwitchPins(SX126X_RXEN, SX126X_TXEN);
     }
+#elif defined(SX126X_RXEN) && (SX126X_RXEN != RADIOLIB_NC && defined(E22_TXEN_CONNECTED_TO_DIO2))
+    // Otherwise, if SX126X_RXEN is connected to the MCU, and E22_TXEN_CONNECTED_TO_DIO2 is defined, we are letting the
+    // E22 control RX and TX via DIO2. In this configuration, the E22's TXEN and DIO2 pins are connected to each other,
+    // but not to the MCU.
+    // However, we must still connect the E22's RXEN pin to the MCU, define SX126X_RXEN accordingly, and then call
+    // setRfSwitchPins, otherwise RX sensitivity (observed via RSSI) is greatly diminished.
```
thebentern, line 85: Great find. I need to incorporate this into the Hydra variant. I recently converted the official variant to use DIO2 / TXEN bridged. I need to make this accomanying change for usings RXEN only switching

.vscode/extensions.json
thebentern, line 1: Let's undo this

.trunk/trunk.yaml
thebentern, line 1: Let's undo this one as well

thebentern (changes requested):

andrew-moroz:
All requested changes made. Looks like gather-artifacts failed, not sure what that's about.

markbirss:
> * `xiao-ble-internal-format.uf2` is identical to `Meshtastic_nRF52_factory_erase.uf2` included in Meshtastic releases, but compiled for softdevice 7.3.0.

Can you confirm "Meshtastic_nRF52_factory_erase.uf2" wont work for softdevice 7.3.0?  
as it is just erasing the internal littlefs region and should be softdevice version independent

https://github.com/markbirss/nRF52_Internal_Format/blob/main/Meshtastic_nRF52_factory_erase/Meshtastic_nRF52_factory_erase.ino

I am in the process of adding Adafruit UF2 support for Makerdiary nRF82540 Connect Kit also, as it comes with out-of-the-box features and a desktop app but with SoftDevice S132 5.1.0

https://wiki.makerdiary.com/nrf52840-connectkit/out-of-box-experience/#installing-the-nrf-connect-bluetooth-low-energy
https://github.com/NordicPlayground/pc-nrfconnect-ble-standalone

Im also aware RAK Wireless still maintain their own fork UF2 bootloader

thebentern (approved):

andrew-moroz:
> Can you confirm "Meshtastic_nRF52_factory_erase.uf2" wont work for softdevice 7.3.0? as it is just erasing the internal littlefs region and should be softdevice version independent
> 

Just circled back to test this again and can confirm Meshtastic_nRF52_factory_erase.uf2 doesn't work. But the same exact Arduino sketch (Meshtastic_nRF52_factory_erase.ino) does work when compiled for the Xiao BLE. I do see differences in the sizes of the binaries exported by Arduino IDE, and their offsets when converted to UF2... when compiled for the Adafruit Feather nRF52840 Express, the start address is 0x26000:

```
~/meshtastic/firmware/bin/uf2conv.py Meshtastic_nRF52_factory_erase.ino.hex -c -o Meshtastic_nRF52_factory_erase.ino.uf2 -f 0xADA52840
Converting to uf2, output size: 123904, start address: 0x26000
Wrote 123904 bytes to Meshtastic_nRF52_factory_erase.ino.uf2
```

But when compiled for the Xiao BLE, it's 0x27000:

```
~/meshtastic/firmware/bin/uf2conv.py Meshtastic_nRF52_factory_erase.ino.hex -c -o Meshtastic_nRF52_factory_erase.ino.uf2 -f 0xADA52840
Converting to uf2, output size: 122880, start address: 0x27000
Wrote 122880 bytes to Meshtastic_nRF52_factory_erase.ino.uf2
```

I think SoftDevice 7.x is bigger than 6.x, which may explain the difference in offset? (Can't remember where I first read that, but searching just now I found [this thread](https://devzone.nordicsemi.com/f/nordic-q-a/56374/how-can-bootloader-know-softdevice-size) which mentions it.) So my thinking was that if Meshtastic_nRF52_factor_erase.uf2 were compiled for the Adafruit board and SD 6.x, flashing it to the Xiao was probably overwriting the tail end of its SD 7.x and causing it not to boot at all.

markbirss:
@andrew-moroz thank you for confirmation

PR #2638, opened 2023-07-23
Check Position Request for Primary Channel

jp-bennett:
Prevents leaking location data to secondary channels.

### DIFF
--- a/src/modules/PositionModule.cpp
+++ b/src/modules/PositionModule.cpp
@@ -52,11 +52,22 @@ bool PositionModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mes
 
     nodeDB.updatePosition(getFrom(&mp), p);
 
+    // Only respond to location requests on the channel where we broadcast location.
+    if (channels.getByIndex(mp.channel).role == meshtastic_Channel_Role_PRIMARY) {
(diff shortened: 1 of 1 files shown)


thebentern (approved):

GUVWAF (changes requested):
I now remember that there is a way to ignore requests, which seems like a better approach than overriding `sendResponse` with largely duplicated logic and exposing `currentReply` to all modules. This should be as simple as doing the if-condition within PositionModule's `allocReply`, set `ignoreRequest` to true and return NULL (see https://github.com/meshtastic/firmware/blob/master/src/modules/NodeInfoModule.cpp#L62). Can you do it in that way?

GUVWAF:
Ah, you're right and you mentioned that already, sorry. Your other approach sounds good and would be cleaner indeed.

GUVWAF:
You could set `ignoreRequest` to true in `handleReceivedProtobuf` and return NULL in `allocReply` if this is the case.

jp-bennett:
@GUVWAF Seems to work as expected. Agreed, this is a much better approach.

GUVWAF (approved):
@jp-bennett Great! Looks good to me.

Issue #2639, opened 2023-07-24
[Bug]: Install Script Failing for T-Deck (likely other recently added S3 devices too)


rcarteraz:
### Category

Other

### Hardware

Other

### Firmware Version

2.1.20

### Description

I've seen two reports of the install script failing when attempting to flash a T-Deck. [Here's one example.](https://meshtastic.discourse.group/t/t-deck-firmware-t-deck-2-1-20-470363d-bin-issues/7909)

One thing I noticed in their log output was this line:
```
Unexpected chip id in image. Expected 9 but value was 0. Is this image for a different chip model?

A fatal error occurred: bleota.bin is not an ESP32-S3 image. Use --force to flash anyway.
```

Even though the device is an S3 and the esptool.py detects it as such. 
```
Connecting...
Detecting chip type... ESP32-S3
Chip is ESP32-S3 (revision v0.1)
```

Taking a look at the install script though, it appears to account for the S3 chip based on the file name, or at least that's how I'm reading it. I could be wrong. But if that is the case, since neither device has s3 or or v3 in the file name, I would assume it would fail as it's attempting to use bleota.bin instead of bleota-s3.bin. Right? 

```bash
	if [ ! -z "${FILENAME##*"s3"*}" ] && [ ! -z "${FILENAME##*"-v3"*}" ]; then
	    "$PYTHON" -m esptool  write_flash 0x260000 bleota.bin
	else
	    "$PYTHON" -m esptool  write_flash 0x260000 bleota-s3.bin
```

PR #2644, opened 2023-07-24
Add T-Deck to S3 ota logical branch

thebentern:
Potential fix for #2639
@rcarteraz would you mind testing this?


thebentern:
> > Potential fix for #2639
> 
> > @rcarteraz would you mind testing this?
> 
> 
> 
> I don't have one to test but I can post on the discourse thread and ask them to. Also, this addresses the t-deck but wouldn't an update need to also be made for the T-Watch? Since it has neither v3 or s3 in the file name... Which now that I think about it, have we heard of any issues with the Wireless Tracker or Wireless paper since both are also S3 devices that do not have S3 or V3 in the name? 

T-Watch S3 is good to go already. Good call on the new Heltec hardware though.

PR #2648, opened 2023-07-26
Ambient Lighting Module

charminULTRA:
Creates the "Ambient Lighting Module", a new module dedicated to having a place for controlling ambient LED lights (rather than status lights and activity lights) being powered by the board. An example is the RAK14001, which is the primary use for this module right now.

Primary controls of the Module will include:

    Brightness (current) control
    Individual RGB level controls


charminULTRA:
I'm struggling with the code inside of src/modules/AmbientLightingModule.cpp. I keep getting "namespace "meshtastic" has no member "ModuleConfig" and 'ModuleConfig' in namespace 'meshtastic' does not name a type" and I have no idea how to fix these. ChatGPT is failing me hard on this one.

charminULTRA:
I'm completely stuck on this, I keep getting the errors mentioned above, and I can't find a solution.

PR #2659, opened 2023-07-27
Convert protobuf values that are unsigned properly to uint in JSON

### LINKED ISSUE #2565: [Bug]: MQTT JSON returns signed "FROM" field
### Category

Other

### Hardware

T-Beam, Rak4631, Other

### Firmware Version

2.1.14

### Description

When enabling the MQTT output and JSON formatting, the FROM field is represented as a signed integer, resulting in negative values for some radios. In most places, like teh debug output in the Android app, these values are parsed as unsigned values. The JSON output should probably send these on as unsigned values as well.

### Relevant log output

```Shell
{"channel":0,"from":-631445440,blah blah
```

GUVWAF:
Note: this PR is going to the 2.2-working-changes branch. Needs to be in the release notes of 2.2, as it will break set-ups relying on this JSON value.

Should fix #2565. 

Untested yet with an actual MQTT set-up.

### DIFF
--- a/src/mqtt/JSONValue.cpp
+++ b/src/mqtt/JSONValue.cpp
@@ -364,6 +364,19 @@ JSONValue::JSONValue(int m_integer_value)
 }
 
 /**
+ * Basic constructor for creating a JSON Value of type Number
+ *
+ * @access public
+ *
+ * @param uint m_integer_value The number to use as the value
+ */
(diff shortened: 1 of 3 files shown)


garthvh:
Should also fix quite a bit of JSON crashing, heltec and tloras often have node ids that overflow a int32.

thebentern (approved):

PR #2665, opened 2023-07-28
T deck: support keyboard, trackball and touchscreen

mverch67:
- the touchscreen base class implements the events swipe up/down/left/right tap/double-tap/long press 
- the touchscreen impl class hooks the getTouch callback from the TFT display driver into and translates the touch events to cardkb input events (long press maps to CANCEL)
- the trackballInterruptBase class is copied from the up/downInterrupt class and extended by the left/right INTs of the trackball
- the keyboard does not require special routines, just hook the I2C address into the various places in the code and let cardkb/cannedMessageModule do the work
- the screen class has been extended to accept inputEvents and handlers for showPrev/NextFrame 

Note: as usual all commits from last week (which are already merged into firmware:master) are included in this PR (again and again and again). This is not intended and seems to be a github bug!?


caveman99:
Looks excellent to me.

You can ignore the log lines, it fill find what is already available on merge and skip the rest. Merging to Master is also cleaner logged with a rebase, so unless it tells you it can not rebase, use this one for merges ;-)

thebentern (approved):

mverch67:
I just did some fine-tuning regarding sensitivity and accuracy and now it detects the touch actions really well without accidental taps when swiping or single tap when double-tapping. Also the swipe detection is now configurable (via variants.h) for individual devices.

mverch67:
And during my testing it happened 100 times that I accidentally send out a canned message with an accidental tap while swiping. Now I chose to use double-tap for sending and now it's impossible to accidentally send out a wrong message. 
As a consequence the screen does not react to single tap (e.g. when it's black / power off) but only to swipe and double-tap.

PR #2670, opened 2023-07-31
SerialModule SIMPLE mode: use `write()` instead of `printf()`

### LINKED ISSUE #2669: [Bug]: Serial data output to Console doesn't allow binary data
### Category

Serial

### Hardware

T-Beam

### Firmware Version

1.2.19 beta

### Description

Line 252 in src/modules/SerialModule.cpp does

    serialPrint->printf("%s", p.payload.bytes);

but should be

    serialPrint->write(p.payload.bytes, p.payload.size);

so that binary data is output completely.  At present a value of 0 in the payload will stop output.

I'm outputting serial data to the serial console after setting device.serial_enabled to false and override console_serial_port to true.  But since my serial messages are binary, the output was interrupted if a 0x00 was embedded in the serial message.

### Relevant log output

_No response_

GUVWAF:
Fixes #2669.


geeksville:

https://meshtastic.discourse.group/t/how-to-build-firmware-for-t-beam/7940/2

thebentern (approved):

caveman99 (approved):

PR #2680, opened 2023-08-06
Initialize the L76K Chip, use GPS + GLONASS + BEIDOU

neilhao:
Enabled BEIDOU for L76K Chip

Reference: https://www.waveshare.com/wiki/L76K_GPS_Module

Please don't merge until the user who owns the T-Echo device confirms that there is no problem. The Nano series passed the test.


thebentern:
Works great on my T-Echo

PR #2694, opened 2023-08-08
T-Watch S3 fixes

### LINKED ISSUE #2692: [Bug]: T-deck - Freetext last character
### Category

Other

### Hardware

T-Deck

### Firmware Version

2.1.21.97d7a89

### Description

In Freetext mode, the last character is not sent. 

### Relevant log output

_No response_

mverch67:
This PR fixes a couple of issues for T-Watch S3 :-
- corrected swipe/touch rotation
- default screen rotation set for right-handed users (wearing the watch on the left arm)
- implemented flip screen (for left-handed users)
- haptic feedback events reduced to actual no. of touch events
- avoid screen blanking during touch operations when using USB power
- when dark a single tap turns screens on
- VScode automatic port detection fixed
- T-Deck: CannedMessageModule did replace last character with bell character (fix #2692)


src/modules/CannedMessageModule.cpp
```diff
@@ -225,8 +225,8 @@ void CannedMessageModule::sendText(NodeNum dest, const char *message, bool wantR
     p->decoded.payload.size = strlen(message);
     memcpy(p->decoded.payload.bytes, message, p->decoded.payload.size);
     if (moduleConfig.canned_message.send_bell) {
-        p->decoded.payload.bytes[p->decoded.payload.size - 1] = 7; // Bell character
-        p->decoded.payload.bytes[p->decoded.payload.size] = '\0';  // Bell character
+        p->decoded.payload.bytes[p->decoded.payload.size] = 7;        // Bell character
+        p->decoded.payload.bytes[p->decoded.payload.size + 1] = '\0'; // Bell character
         p->decoded.payload.size++;
```
caveman99, line 230: I don't think you can do that - cause +1 does not belong to you - need a bit of string math here :-)

caveman99 (changes requested):

src/modules/CannedMessageModule.cpp
mverch67: Added checks for max. payload size (depending on bell switch).

thebentern (approved):

caveman99 (approved):

PR #2696, opened 2023-08-09
[breaking] fix EU433 LoRa definitions

### LINKED ISSUE #3371: [Bug]: EU 433 channels centers incorrect
### Category

Other

### Hardware

Not Applicable

### Firmware Version

2.2.24

### Description

Accorting to lora alliance the band plan starts at 433.05Mhz.
Not 433.00. It ends at 434.79. The very first channel used by meshtastic is centered at 433.125, but it should be 433.175. 
ETSI has defined 433 to 434 MHz frequency band for LoRa application. It uses 433.175 MHz, 433.375 MHz and 433.575 MHz frequency channels.
Having a bandwidth of 250 kHz makes it out of band and illegal to be used without a licence.
Please consider shifting all the channels.
IMG_20240311_131545_169

### Relevant log output

_No response_

andrekir:
[EN300220 ETSI V3.2.1 [Table B.1, Item H, p. 21]](https://www.etsi.org/deliver/etsi_en/300200_300299/30022002/03.02.01_60/en_30022002v030201p.pdf)

EN300220 ETSI V3 2 1  Table B 1, Item H, p  21

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -32,9 +32,10 @@ const RegionInfo regions[] = {
     RDEF(US, 902.0f, 928.0f, 100, 0, 30, true, false, false),
 
     /*
-        https://lora-alliance.org/wp-content/uploads/2020/11/lorawan_regional_parameters_v1.0.3reva_0.pdf
-     */
-    RDEF(EU_433, 433.0f, 434.0f, 10, 0, 12, true, false, false),
+        EN300220 ETSI V3.2.1 [Table B.1, Item H, p. 21]
+        https://www.etsi.org/deliver/etsi_en/300200_300299/30022002/03.02.01_60/en_30022002v030201p.pdf
+    */
+    RDEF(EU_433, 433.05f, 434.79f, 10, 0, 10, true, false, false),
 
     /*
        https://www.thethingsnetwork.org/docs/lorawan/duty-cycle/


andrekir:
marking as draft since changing the start frequency will likely break backwards compatibility. other regions using 433.0f should also be reviewed.

caveman99:
I am not sure this is necccessary, at least i recall similar discussions a while ago. Seems like ETSI specifies center frequencies with a bit of guard to allow for the bandwidth. The matching ITU allocations start and end at the band boundaries and Meshtastic takes that into account when calculating the center frequencies. Let's follow this up on #frequency-bands in discord.

andrekir:
no rush, leaving as a reminder for the next breaking cycle.

considering convention and even EU_868, it's pretty safe to say 433,050 MHz to 434,790 MHz is the proper frequency range for EU_433.

mc-hamster:
Good eye @andrekir !

fifieldt:
Q) How about offering EU_433_OLD and EU_433_NEW in parallel for a while? That way communities could choose to move independently of our firmware timing?

andrekir:
on second look, the added 50 Hz represents about 0.0115% in frequency shift, which I suspect is well within tolerance for radios to interoperate using either center frequency. I don’t have any 433 MHz radios to test that theory, though.

caveman99:
> on second look, the added 50 KHz represents about 0.0115% in frequency shift, which I suspect is well within tolerance for radios to interoperate using either center frequency. I don’t have any 433 MHz radios to test that theory, though.

Right now we have 433.0-434.0 which is 1 MHz or 4 default channels. New definition is 433.050-434.790 which is 1.74 MHz being (almost) 7 default channels. I propose to change it to 433.0-434.8 with a margin of error ... :-) - the 50khz spacing is really a guard against the next band user below and the gauss tx distribution is weaker near the edge of the band. Also we're talking a few milliwatts here, it's not like this is the Chinese Firedragon Shortwave jammer ...

Nevermore-42:
> Q) How about offering EU_433_OLD and EU_433_NEW in parallel for a while? That way communities could choose to move independently of our firmware timing?

I like this idea if it means we can have a corrected preset sooner. Eventually we can deprecate the EU_433_OLD after some time :)

GUVWAF:
If we add the part of 434,04-434,79MHz to the new version, I would advise that at least the default frequency slot does not fall within there. This part can be used without duty cycle with 10mW when using a narrow bandwidth (25kHz), hence it could lead to severe interference from other systems.

kholia:
@andrekir Hi - Can we please rebase this PR and make it ready for review?

It seems this old PR is blocking the following similar work:

- https://github.com/meshtastic/protobufs/pull/998
- https://github.com/meshtastic/firmware/pull/11268

Ideally, IN_433 should be a separate entity or perhaps an alias to this new EU_433 region.

CC @caveman99 An old PR from 2023 (!) should NOT stop us from making progress on a new region (namely `IN_433`) perhaps? IN is NOT in EU and both have independent multiple regulatory authorities I guess.  

Thank you for the help.

PR #2701, opened 2023-08-10
My branch mfrc522

DerFedor:


thebentern:
Hi, 

A few items of feedback here:

1. Any changes to the protobuf definitions need to happen in the protobufs repository before using them in device firmware.
2. I don't think trying to put NFC sensors in the telemetry is a good idea. It's a pretty curated list of environment monitor sensors. If we're adding NFC functionality, it really should be in a new module.
3. Any new contributors must first sign the CLA in order to have changes accepted.

Issue #2311, opened 2023-02-26
[Feature Request]: Add Support for LilyGo Q10 Keyboard Driver Module


rcarteraz:
### Platform

NRF52, ESP32

### Description

LilyGo offers a [BB Q10 driver module](https://www.aliexpress.us/item/3256804521965718.html) (Note: This product page has multiple options, the one I'm referring to is the first one which is a small pcb that is $9), that allows you to utilize a Q10 keyboard via i2c. I've found this [LilyGo Github repo](https://github.com/Xinyuan-LilyGO/TTGO_TWatch_Library/tree/master/src) which has the Q10 Keyboard info, if that helps but I'm not sure if that's the exact one or not.

If support is added, it would be a great option for compact standalone devices as the size is much smaller than the cardkb. One thing to note, unlike cardkb, these keyboards do not have arrow keys. However, we could possibly utilize the alt button as a modifier and then use wasd for up, left, down, right when modified? 

<img width="479" alt="Screenshot 2023-07-28 at 12 50 44 PM" src="https://github.com/meshtastic/firmware/assets/6912600/453a5f4a-2c20-4587-8856-afd1b78f7ad4">

Californian:
I'd like this for exactly the use case described (standalone communication device) -- if anyone can point me in the right direction I'd be happy to look into contributing!

mverch67:
@rcarteraz I created a draft PR, please install the firmware from the .zip and check if it works for you!

caveman99:
This will be in 2.2.3

PR #2703, opened 2023-08-10
support BB Q10 keyboard

mverch67:
LILYGO BB Q10 keyboard support, see #2311

Arrow key simulation:
```
UP: SYM SHIFT+2
DOWN: SYM SHIFT+8
LEFT: SYM SHIFT+4
RIGHT: SYM SHIFT+6
```
(first press SYM and release it, then SHIFT plus the number key at the same time)


mverch67:
@rcarteraz I created this PR, please install the firmware from the .zip and check if it works for you!

caveman99 (approved):
LGTM :-) But see comment.

caveman99:
Still lookin good 👍

mverch67:
> @mverch67 Detected but not working.


Did you enabled the CannedMessageModule???

mverch67:
> Yes to both

Do you see anything in the debug log when typing on the keyboard?

mverch67:
> Checked for that too, and nothing that indicates a button has been pressed. Just the typical stuff.

Maybe the BBQ10 requires an additional I2C driver, e.g. https://github.com/arturo182/arduino_bbq10kbd
It could be that something like this is implemented in the esp32-C3 of the T-Deck...

caveman99:
Looks like the older BBQ10 Lilygo Firmware is based on this code ... https://github.com/solderparty/bbq10kbd_i2c_sw - it's queried differently than the T-Deck firmware or the cardKB. Best bet would really to include the access library from the T-Watch firmware and modify the kb driver to use that. if 1F is detected.

caveman99:
@rcarteraz that is to be expected, it's a rough approximation of what 'should' work. It seems with the older code the firmware needs to do the debouncing. I have found sample code for these and will try to improve them ... in the meantime can you check the keymapping is correct? e.g. no matter how often, but it should be the right key.

src/input/kbI2cBase.cpp
```diff
@@ -66,47 +66,50 @@ int32_t KbI2cBase::runOnce()
...
-                break;
-            case 0x16: // Right (sym shift+6)
-                e.inputEvent = meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar_RIGHT;
-                e.kbchar = 0x00; // tweak for destSelect
-                break;
-            case 0x0d: // Enter
-                e.inputEvent = meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar_SELECT;
-                break;
-            case 0x00: // nopress
+            if ((key.key != 0x00) && (key.state == BBQ10Keyboard::StateRelease)) {
```
mverch67, line 69: I think this will never be true :)

mverch67:
I was thinking about something like this:

```
BBQ10Keyboard::KeyEvent BBQ10Keyboard::keyEvent() const
{
    KeyEvent event = {.key = '\0', .state = StateIdle};

    if (keyCount() == 0)
        return event;

    const uint16_t buf = readRegister16(_REG_FIF);
    char key = buf >> 8;
    KeyState state = KeyState(buf & 0xFF);

    if (key == KEY_MOD_SYM && state == StateRelease) {
        sym_pressed = true;
        return event;
    }

    if (sym_pressed && key > 0x20) {
        sym_pressed = false;
        event.key = key - 0x20;
    }
    else {
        event.key = key;
    }
    event.state = state;

    return event;
}
```

PR #2713, opened 2023-08-12
Extend node max to 100 and remove mesh_sds_timeout_secs

thebentern:


src/mesh/http/ContentHandler.cpp
```diff
@@ -467,7 +465,7 @@ void handleFormUpload(HTTPRequest *req, HTTPResponse *res)
     // first semicolon, if one exists:
     size_t semicolonPos = contentType.find(";");
     if (semicolonPos != std::string::npos) {
-        contentType = contentType.substr(0, semicolonPos);
+        contentType = contentType.substr(0, semicolonPos); // cpp-check-suppress
```
mverch67, line 468: contentType.resize(semicolonPos)  would be much more efficient as it does not construct a new string (and alloc temp memory on the heap)

mverch67 (approved):

PR #2719, opened 2023-08-13
Update T-Deck variant.h to remove HAS_GPS

rcarteraz:
This removes the HAS_GPS 0 definition from the variant.h file so users can add gps without needing to build custom firmware.


jp-bennett:
I approve of this in principle, but don't have a t-deck to actually smoke test.

jp-bennett (approved):

code8buster (approved):

PR #2722, opened 2023-08-14
Initial Detection sensor module feature

thebentern:
Tested with RAK PIR module as a pilot

### DIFF
--- a/src/mesh/generated/meshtastic/module_config.pb.h
+++ b/src/mesh/generated/meshtastic/module_config.pb.h
@@ -125,6 +125,33 @@ typedef struct _meshtastic_ModuleConfig_NeighborInfoConfig {
     uint32_t update_interval;
 } meshtastic_ModuleConfig_NeighborInfoConfig;
 
+/* Detection Sensor Module Config */
+typedef struct _meshtastic_ModuleConfig_DetectionSensorConfig {
+    /* Whether the Module is enabled */
+    bool enabled;
+    /* Interval in seconds of how often we can send a message to the mesh when a state change is detected */
+    uint32_t minimum_broadcast_secs;
+    /* Interval in seconds of how often we should send a message to the mesh with the current state regardless of changes
+ When set to 0, only state changes will be broadcasted
+ Works as a sort of status heartbeat for peace of mind */
+    uint32_t state_broadcast_secs;
+    /* Send ASCII bell with alert message
+ Useful for triggering ext. notification on bell */
+    bool send_bell;
+    /* Friendly name used to format message sent to mesh
+ Example: A name "Motion" would result in a message "Motion detected"
+ Maximum length of 20 characters */
+    char name[20];
+    /* GPIO pin to monitor for state changes */
+    uint8_t monitor_pin;
+    /* Whether or not the GPIO pin state detection is triggered on HIGH (1)
+ Otherwise LOW (0) */
+    bool detection_triggered_high;
+    /* Whether or not use INPUT_PULLUP mode for GPIO pin
+ Only applicable if the board uses pull-up resistors on the pin */
+    bool use_pullup;
+} meshtastic_ModuleConfig_DetectionSensorConfig;
+
 /* Audio Config for codec2 voice */
 typedef struct _meshtastic_ModuleConfig_AudioConfig {
     /* Whether Audio is enabled */
@@ -342,6 +369,8 @@ typedef struct _meshtastic_ModuleConfig {
         meshtastic_ModuleConfig_NeighborInfoConfig neighbor_info;
         /* TODO: REPLACE */
         meshtastic_ModuleConfig_AmbientLightingConfig ambient_lighting;
+        /* TODO: REPLACE */
+        meshtastic_ModuleConfig_DetectionSensorConfig detection_sensor;
     } payload_variant;
 } meshtastic_ModuleConfig;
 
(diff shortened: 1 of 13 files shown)


src/modules/DetectionSensorModule.cpp
```diff
@@ -0,0 +1,94 @@
...
+        sendCurrentStateMessage();
+        return DELAYED_INTERVAL;
+    }
+    return GPIO_POLLING_INTERVAL;
+}
+
+void DetectionSensorModule::sendDetectionMessage()
+{
+    LOG_DEBUG("Detected state change. Sending message\n");
+    char *message = new char[40];
```
mverch67, line 59: When are these 40 bytes released? Memory leak?!

src/modules/DetectionSensorModule.cpp
```diff
@@ -0,0 +1,94 @@
...
+        p->decoded.payload.size++;
+    }
+    LOG_INFO("Sending message id=%d, dest=%x, msg=%.*s\n", p->id, p->to, p->decoded.payload.size, p->decoded.payload.bytes);
+    lastSentToMesh = millis();
+    service.sendToMesh(p);
+}
+
+void DetectionSensorModule::sendCurrentStateMessage()
+{
+    char *message = new char[40];
```
mverch67, line 77: When are these 40 bytes released? Memory leak?!

src/modules/DetectionSensorModule.cpp
thebentern: Doh!

src/modules/DetectionSensorModule.cpp
```diff
@@ -0,0 +1,94 @@
...
+    meshtastic_MeshPacket *p = allocDataPacket();
+    p->want_ack = false;
+    p->decoded.payload.size = strlen(message);
+    memcpy(p->decoded.payload.bytes, message, p->decoded.payload.size);
+    LOG_INFO("Sending message id=%d, dest=%x, msg=%.*s\n", p->id, p->to, p->decoded.payload.size, p->decoded.payload.bytes);
+    lastSentToMesh = millis();
+    service.sendToMesh(p);
+}
+
+bool DetectionSensorModule::hasStateChanged()
```
mverch67, line 91: Just a question if I understood correctly, because the term "has changed" seems confusing to me. To determine if sth. has changed you normally would need to compare it with the **old state**, but in this function the actual state is compared against the preset high or low. For motion this may make sense, but what about other types of sensors (e.g. water level). Should the message only be send if a "change" has been detected or periodically as long as the actual value differs from the preset?

src/modules/Modules.cpp
```diff
@@ -31,6 +31,7 @@
 #include "modules/ExternalNotificationModule.h"
 #include "modules/RangeTestModule.h"
 #if (defined(ARCH_ESP32) || defined(ARCH_NRF52)) && !defined(CONFIG_IDF_TARGET_ESP32S2)
+#include "modules/DetectionSensorModule.h"
```
mverch67: excluded for RP2040?

thebentern: Moved it up. I didn't even notice it was macro guarded until portduino target failed 😅

src/modules/DetectionSensorModule.cpp
```diff
@@ -0,0 +1,94 @@
+#include "DetectionSensorModule.h"
+#include "MeshService.h"
+#include "NodeDB.h"
+#include "PowerFSM.h"
+#include "configuration.h"
+#include "main.h"
+
+DetectionSensorModule *detectionSensorModule;
+
+#define GPIO_POLLING_INTERVAL 80
```
mverch67, line 10: Is it really required to poll the GPIO state 1000 / 80 = 12.5 times per second?

src/modules/DetectionSensorModule.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // moduleConfig.detection_sensor.detection_triggered_high = false;
+    // strcpy(moduleConfig.detection_sensor.name, "Motion");
+
+    if (moduleConfig.detection_sensor.enabled == false)
+        return disable();
+
+    if (firstTime) {
+        // This is the first time the OSThread library has called this function, so do some setup
+        firstTime = false;
+        if (moduleConfig.detection_sensor.monitor_pin > 0) {
```
mverch67, line 32: What if the monitor_pin is kept 0 and the module is enabled? Looks like hasStateChanged() tries to read from GPIO0.

src/mesh/generated/meshtastic/module_config.pb.h
```diff
@@ -125,6 +125,33 @@ typedef struct _meshtastic_ModuleConfig_NeighborInfoConfig {
     uint32_t update_interval;
 } meshtastic_ModuleConfig_NeighborInfoConfig;
 
+/* Detection Sensor Module Config */
+typedef struct _meshtastic_ModuleConfig_DetectionSensorConfig {
+    /* Whether the Module is enabled */
+    bool enabled;
+    /* Interval in seconds of how often we can send a message to the mesh when a state change is detected */
+    uint32_t minimum_broadcast_secs;
```
mverch67, line 133: Is there a minimum value (or a recommendation)?

src/modules/DetectionSensorModule.cpp
thebentern: I agree. I think hasDetectionEvent() is a better description of what I'm looking for, because I don't actually want to be in the business of tracking previous states.

caveman99 (approved):

PR #2724, opened 2023-08-15
Move RangeTestModule to its own port number

caveman99:


thebentern (approved):

caveman99:
TEsting partially successful. I have a T-Beam sending rangetest packets every 60 seconds. I see them in the other nodes and if rangetest is not enabled they're discarded. If rangetest is enabled they're processed but the rewriting to display and show them on a connected app does not seem to work. At least it doesn't crash the RX node... :-)

thebentern:
Closing after discussion about less invasive method implemented in #2732. It will be addressed there.

PR #2728, opened 2023-08-16
Permanently Enable Canned Messages on T-Deck and Picomputer

caveman99:
picomputer has a really dark TFT; switch color to white for better UX.


caveman99:
Tested with picomputer s3. kb is enabled after factory reset. RAK4631 does not enable it by default.

thebentern (approved):

PR #2732, opened 2023-08-18
Portnum promiscuity for text messages from other modules

thebentern:


src/modules/ExternalNotificationModule.cpp
```diff
@@ -128,6 +128,11 @@ int32_t ExternalNotificationModule::runOnce()
     }
 }
 
+bool ExternalNotificationModule::wantPacket(const meshtastic_MeshPacket *p)
+{
+    return MeshService::isTextPayload(p);
+}
+
 /**
```
caveman99, line 136: A similar change needs to go into the payload gatherer for the Store&Forward App.

caveman99 (changes requested):

caveman99 (approved):

PR #2737, opened 2023-08-22
Bbq10 kb fix

caveman99:


src/input/kbI2cBase.cpp
```diff
@@ -78,21 +78,49 @@ int32_t KbI2cBase::runOnce()
                     e.inputEvent = meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar_BACK;
                     e.kbchar = key.key;
                     break;
-                case 0x12: // sym shift+2
-                    e.inputEvent = meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar_UP;
-                    e.kbchar = 0xb5;
+                case 'e': // sym e
+                case '2': // sym shift+2
```
mverch67, line 82: should probably be 'E'

src/input/kbI2cBase.cpp
```diff
@@ -78,21 +78,49 @@ int32_t KbI2cBase::runOnce()
...
+                        e.inputEvent = ANYKEY;
+                        e.kbchar = key.key;
+                    }
                     break;
-                case 0x18: // sym shift+8
-                    e.inputEvent = meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar_DOWN;
-                    e.kbchar = 0xb6;
+                case 'x': // sym x
+                case '8': // sym shift+8
+                    if (is_sym) {
```
mverch67, line 93: should probably be 'X'

caveman99:
Right now the sym state will not reset itself after the 2nd keypress, the device will stay in SYM mode till you press the button again. Much like Capslock. I will add a reset when i do the final cleanup.

caveman99:
Do we have a final verdict if this can go in?

PR #2739, opened 2023-08-22
UI/UX: Display delivered message on incoming ACK.

caveman99:
Needs more work, see comments.


src/modules/CannedMessageModule.h
```diff
@@ -37,15 +38,29 @@ class CannedMessageModule : public SinglePortModule, public Observable<const UIF
     const char *getMessageByIndex(int index);
     const char *getNodeName(NodeNum node);
     bool shouldDraw();
-    void eventUp();
-    void eventDown();
-    void eventSelect();
+    // void eventUp();
+    // void eventDown();
+    // void eventSelect();
```
caveman99, line 44: This is some leftover, testing if they are still needed

src/modules/CannedMessageModule.cpp
```diff
@@ -546,6 +552,24 @@ void CannedMessageModule::drawFrame(OLEDDisplay *display, OLEDDisplayUiState *st
...
 
+ProcessMessage CannedMessageModule::handleReceived(const meshtastic_MeshPacket &mp)
+{
+    if (mp.decoded.portnum == meshtastic_PortNum_ROUTING_APP) {
+        // look for a request_id
+        if (mp.decoded.request_id != 0) {
+            UIFrameEvent e = {false, true};
+            e.frameChanged = true;
+            this->runState = CANNED_MESSAGE_RUN_STATE_ACK_RECEIVED;
+            this->incoming = mp.decoded.request_id;
```
caveman99, line 563: This is not actually the node number but the message id. We need to dig into the outgoing message store if we find the 'dest' field in it. Right now it always displays 'unknown'

GUVWAF: About this: we could for now just display "Last message delivered" or "Last delivery failed" until we have a way of finding the `dest` field?

src/mesh/MemoryPool.h
jp-bennett, line 1: Is this intended to be part of this PR?

caveman99: Yes this is intended. While researching alternative ways to find the originating message, I found out this is dead code. We can put it into a separate pr but I don't think it warrants the overhead.

GUVWAF (approved):
Looks good to me. 
I think we can eventually get rid of this special FreeRTOS queue for the `ToPhoneQueue` (seems this was actually [the plan](https://github.com/meshtastic/firmware/blob/master/src/concurrency/OSThread.h#L29)) to handle persistent storage for message history on standalone devices. If we can replace it by a `MeshPacketQueue`, which is actually a `std::vector`, we can more easily cycle through it and reorder packets.

caveman99:
i pushed the current state to master, let's keep this branch open to refine this further. I'll push canned messaging to secondary channels next.

PR #2744, opened 2023-08-24
Use fixed position regardless of gps lock

thebentern:


thebentern:
After fix, even though we receive a new position, we retain the set one:
DEBUG | 12:00:08 100 [GPS] New GPS pos@64e74647:3 lat=3***, lon=-9**, alt=90, pdop=2.53, track=0.00, speed=0.00, sats=5
WARN  | 12:00:08 100 [GPS] Using fixed position
DEBUG | 12:00:08 100 [GPS] onGPSChanged() pos@0, time=1692878408, lat=3*****, lon=-9*****, alt=102
INFO  | 12:00:08 100 [GPS] updatePosition LOCAL pos@0, time=1692878408, latI=3****, lonI=-9******, alt=102

PR #2745, opened 2023-08-24
Don't turn off bluetooth when serial stream api is connected

thebentern:


thebentern:
Observed odd behavior when connected to NRF52 via bluetooth / iOS app and serial / MNMC. Seems like the serial takes precedence. Going to run some more experiments.

caveman99:
What about a semaphore/singleton approach? There can only be one API connection. So we don't have the need to physically disable bluettoth (or eth/wifi) with all its side effects?

PR #2765, opened 2023-09-04
Fix possible memory leak in NeighborInfo Module

GUVWAF:
Looks like this neighborInfo packet was never freed. This should actually not be dynamically allocated, it will be copied in the MeshPacket which is dynamically allocated with `allocDataProtobuf()`.


thebentern (approved):

jp-bennett (approved):
~~Seems to fix the crash on my T-beam.~~
Spoke too soon. After an hour and 10 minutes, crashed again. It might be less often, or might be a coincidence.

PR #2772, opened 2023-09-05
UBX-RXM-PMREQ soft-off implemented

code8buster:
Rather than be mad at git for my misuse of its rebase function, I just made a new branch and copied my simple changes.
Enjoy all!

Original PR:
This should allow anyone without dedicated gpio to put their Ublox GNSS receivers to sleep.
Either using triple user button press or simply setting position.gps_enable to zero.

We can also start constructing packets for use elsewhere in the code base. Previously all ublox messages were defined and sent within functions.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -28,11 +28,19 @@ GPS *gps;
 /// Multiple GPS instances might use the same serial port (in sequence), but we can
 /// only init that port once.
 static bool didSerialInit;
+uint8_t UBXscratch[250] = {0};
 
 struct uBloxGnssModelInfo info;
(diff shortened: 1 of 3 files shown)


src/gps/GPS.cpp
```diff
@@ -28,11 +28,19 @@ GPS *gps;
 /// Multiple GPS instances might use the same serial port (in sequence), but we can
 /// only init that port once.
 static bool didSerialInit;
+uint8_t UBXscratch[250] = {0};
```
thebentern, line 31: Let's move this into the header

thebentern (approved):

PR #2775, opened 2023-09-05
Only update neighbors when module is enabled

GUVWAF:
If the NeighborInfoModule is not enabled, we shouldn't do any bookkeeping upon receiving a packet as we don't do anything with it and we also don't clean-up any neighbors, thus only consuming processing power/writes to flash (and it even caused a bug before #2767). It will still update the `last_sent_by_id` when passing through it in the FloodingRouter.


jp-bennett:
Can remove line 102, as it's no longer necessary with this change.

GUVWAF:
That's right, done.

code8buster (approved):

jp-bennett (approved):

PR #2777, opened 2023-09-06
Add ubx-cfg-rxm for ublock 6 powersave

jp-bennett:


code8buster:
The closer approximation for what CFG-PMS does would be eco-mode. For that, you'd set the non-reserved payload byte to 0x04
If you stick with 0x01 you'll also have to implement CFG-PM2 which is probably more granular than most users would care to configure

jp-bennett:
@code8buster Does this CFG-PM2 config look right? Generated using u-center, doing a 1 second update period, and 1 second ontime. Since we're also set to do one update per second, I think that ends up being about like a 1hz powersave mode. Power usage is really quite impressive on my test setup.

thebentern (approved):

code8buster (approved):
This looks good.
As written, the receiver will turn off for the duration of the search period when it can't get a fix, and wake up again at that regular interval to search again.

If you make that second byte to 0x83, the receiver doesn't stop looking for a fix when it can't find one. That flips the doNotEnterOff bit

PR #2778, opened 2023-09-07
Update variant.h

RicInNewMexico:
Add second i2c channel on external connector for Station G1 on pins 14/15


thebentern:
Looks good. If you run trunk fmt, we can get it merged

PR #2780, opened 2023-09-07
Experiment with moving gps init

thebentern:
Seems to behave fine in my local testing of a device w/ a GPS (Heltec Wireless Tracker) and a device with no GPS (bare RAK-4631)


mverch67:
Yeah, move it to the end. Looks good to me.

thebentern:
> Yeah, move it to the end. Looks good to me.

Yep. The only thing ahead of it is the setCPUFast(false). I figured it would be good to keep the CPU at full speed while we're interrogating GPS. 😄

jp-bennett (approved):
I have some existing problems with dropped serial messages, that makes this a bit difficult to review, but it seems to be working as expected.

PR #2789, opened 2023-09-12
Fix time=0 bug for fixed_position nodes

thebentern:
Should fix bug reported by @b8b8 and @cmh in Discord, wherein fixed position nodes send out problematic time=0 with position packets


garthvh (approved):
Awesome

Issue #2754, opened 2023-08-29
[Bug]: Node set name and device info from other node in network


mrekin:
### Category

Other

### Hardware

Heltec V3

### Firmware Version

2.2.3.282cc0b

### Description

This is very strange story, I don't have full info and can't reproduce it right now. I don't know if that firmare bug or android app.

Short version: my node at one moment get short\long name and device info from other node in network. The mate node get name\device info from my node (and may be gps position was switched

Long story:
My node: heltec wsl, latest firmware ([2.2.3.282cc0b]), latest android app (2.2.3), node name: narm_ws_solar
Mate node: DIYv1, latest firmware ([2.2.3.282cc0b]), nodename: spap25e
Nodes has no direct connection (2 nodes between at the way), both uses fixed gps and wifi connection.
At sunday we try to switch from LF preset to LM preset for testing (7 nodes at all) - do some test and return back to LongFast  preset at the end of day.
At monday I found that my node has name "spap25e" and device info "diy_v1" in android app (in nodes tab and in  device list tab). My second node also show incorrect node name for the first node.
 I open Radio configuration, user info - there was correct name. 
After changing name in user info to other value and saving - node name was correct in android app for aprx. ~30min and then switched back to  "spap25e".
I set name again and tried to drop node DB - the situation repeated.
I try to set name again and node stops loading.
In serial i get 
`INFO  | ??:??:?? 1 Loading /prefs/channels.proto
INFO  | ??:??:?? 1 Loaded saved channelFile version 22
[  1016][E][vfs_api.cpp:105] open(): /littlefs/oem/oem.proto does not exist, no permits for creation
INFO  | ??:??:?? 1 No /oem/oem.proto preferences found
WARN  | ??:??:?? 1 NOTE! Our desired nodenum 0x75cec040 is in use, so trying for 0x4
WARN  | ??:??:?? 1 NOTE! Our desired nodenum 0x4 is in use, so trying for 0x4
WARN  | ??:??:?? 1 NOTE! Our desired nodenum 0x4 is in use, so trying for 0x4
WARN  | ??:??:?? 1 NOTE! Our desired nodenum 0x4 is in use, so trying for 0x4`

After this I flash node with full erase (latest fw again) and now all is correct.

Later mate says that his node get the same behavior - his node get my node name and info.
All other 5 nodes has no problems, most of them at older [firmwares.]

Here the screen video from my node and some screenshots
(url)

[...]

PR #2798, opened 2023-09-15
fix: nodenum 4

mverch67:
fix for #2754.

Root cause:
When receiving the admin message to reset the NodeDB an empty db is saved and then waited for 7s for reboot. If in the meantime a new node msg was received then it is saved in db.proto (before the reboot), however the own mac address is 0 in this list (as the own node was already removed). This will lead to the discrepancy on next reboot. 

This fix does the following: when the node starts and loads the db.proto then the meshnode array is checked for entries with has_user==false and removes these as they are the problematic ones (and not shown at the client anyway).

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -81,13 +81,11 @@ bool NodeDB::resetRadioConfig(bool factory_reset)
 
     radioGeneration++;
 
-    if (factory_reset)
-    {
+    if (factory_reset) {
         didFactoryReset = factoryReset();
     }
 
-    if (channelFile.channels_count != MAX_NUM_CHANNELS)
-    {
+    if (channelFile.channels_count != MAX_NUM_CHANNELS) {
         LOG_INFO("Setting default channel and radio preferences!\n");
 
         channels.initDefaults();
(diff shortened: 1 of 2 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -388,25 +390,21 @@ void NodeDB::init()
...
-    while ((found = getMeshNode(r)) && memcmp(found->user.macaddr, owner.macaddr, sizeof(owner.macaddr))) {
-        // FIXME: input for random() is int, so NODENUM_BROADCAST becomes -1
-        NodeNum n = random(NUM_RESERVED, NODENUM_BROADCAST); // try a new random choice
-        LOG_WARN("NOTE! Our desired nodenum 0x%x is in use, so trying for 0x%x\n", r, n);
-        r = n;
+    while ((n == NODENUM_BROADCAST || n < NUM_RESERVED) ||
+           ((found = getMeshNode(n)) && memcmp(found->user.macaddr, owner.macaddr, sizeof(owner.macaddr) == 0))) {
+        NodeNum r = random(NUM_RESERVED, LONG_MAX); // try a new random choice
+        LOG_WARN("NOTE! Our desired nodenum 0x%x is invalid or in use, so trying for 0x%x\n", n, r);
+        n = r;
```
thebentern, line 403: While we're in here, can we rename n to **nodeNum** and r to something like **candidate**? The single character vars have had a good run 😅

thebentern (approved):

thebentern (approved):

caveman99 (approved):

PR #2802, opened 2023-09-17
Simplify SX126x configuration & Add support for "E22-900M30S with ESP32-S3-WROOM-1-N4" board

S5NC:
Will require updating all variant files which use SX126x!
Will require adding board to [src/platform/esp32/architecture.h](https://github.com/meshtastic/firmware/blob/master/src/platform/esp32/architecture.h) and [src/mesh/generated/meshtastic/mesh.pb.h](https://github.com/meshtastic/firmware/blob/master/src/mesh/generated/meshtastic/mesh.pb.h)

Board name may need to be changed to something a little shorter, I trust you can find a name which meets current conventions, though please note that this is for the ESP32-**S3** in particular, the regular ESP32 would require slightly different files

`// FIXME`s added where relevant for future changes or improvements

Unsure whether the "-DARDUINO_ESP32S3_DEV" flag is needed, nor of its purpose, and the effect of the other extra flags in ESP32-S3-WROOM-1-N4.json

### DIFF
--- a/variants/E22-900M30S-with-ESP32-S3-WROOM-1-N4/platformio.ini
+++ b/variants/E22-900M30S-with-ESP32-S3-WROOM-1-N4/platformio.ini
@@ -3,7 +3,6 @@ extends = esp32s3_base
 ; board assumes: 4 MB (Quad SPI) Flash, No PSRAM
 board = ESP32-S3-WROOM-1-N4
 board_level = extra
-upload_speed = 2000000
 ; build_unflag assumes aforecommented lack of PSRAM
 build_unflags = -DBOARD_HAS_PSRAM=1
 build_flags =


jp-bennett:
Is this actually a board, or is it just the CPU? What is the specific hardware being targeted?

jp-bennett:
I would suggest picking a name for the board, rather than just using the CPU name, to avoid confusion.

src/mesh/SX126xInterface.cpp
```diff
@@ -26,20 +26,25 @@ template <typename T> bool SX126xInterface<T>::init()
...
+    LOG_DEBUG("DIO3_AS_TCXO_AT_1V8 mode false. Not using DIO3 to power TCXO\n");
+    float tcxoVoltage = 0; // None - we use an XTAL
 #endif
     bool useRegulatorLDO = false; // Seems to depend on the connection to pin 9/DCC_SW - if an inductor DCDC?
 
     RadioLibInterface::init();
 
-    if (power == 0)
-        power = SX126X_MAX_POWER;
+    // Do not correct 0dBm to 22dBm as may be using external amplifier! (removed code that did this)
```
thebentern, line 44: Don't make this change... we already have precdence for this. Look at the SX126X_MAX_POWER definition in the Station G1 variant for an example. Default limit for SX1262/8 is 22dbm. If targets have amplification, they must set their own limits in the variant

thebentern:
> Will require updating all variant files which use SX126x! 
This doesn't sound too appealing, especially as we're trying to stabilize 2.2.x for a beta release.

src/mesh/SX126xInterface.cpp
caveman99: That code was for setting sensible defaults when we install a new configuration cause of version changes. Not sure it's still neccessary, it sure was transitioning from 1.2 to 2.0

PR #2803, opened 2023-09-17
Always turn on TCXO for SX126X chips

kokroo:
Always power on the TCXO on SX126X devices, since all of them have a TCXO onboard which needs to be activated by setting DIO3 to high. Reference: https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/schematic/T-SX1262.pdf. We don't need to have an "if" check here.

### DIFF
--- a/src/mesh/SX126xInterface.cpp
+++ b/src/mesh/SX126xInterface.cpp
@@ -26,7 +26,9 @@ template <typename T> bool SX126xInterface<T>::init()
     pinMode(SX126X_POWER_EN, OUTPUT);
 #endif
     // Always power on the TCXO on SX126X devices, since all of them have a TCXO onboard
-    // which needs to be activated by setting DIO3 to high. Reference: https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/schematic/T-SX1262.pdf
+    // which needs to be activated by setting DIO3 to high. Reference:
+    // https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/schematic/T-SX1262.pdf
+
     // Use DIO3 to power tcxo per https://github.com/jgromes/RadioLib/issues/12#issuecomment-520695575
     float tcxoVoltage = 1.8;
 
--- a/src/mesh/SX126xInterface.cpp
+++ b/src/mesh/SX126xInterface.cpp
@@ -25,13 +25,11 @@ template <typename T> bool SX126xInterface<T>::init()
     digitalWrite(SX126X_POWER_EN, HIGH);
     pinMode(SX126X_POWER_EN, OUTPUT);
 #endif
-
-#ifndef SX126X_E22
-    float tcxoVoltage = 0; // None - we use an XTAL
-#else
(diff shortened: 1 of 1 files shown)


caveman99:
So you can guarantee this is always true? For all boards that use the Sx1262? Lilygo, Heltec, Wisblock, DIY, Ebyte?

Also for the Rp2040 based cores and the STM32 targets? What about portduino and raspberry shields? Heltec sufficient iot gateway?

Lemme put it another way. There's a reason radiolib exposes that setting. If it was true for every Sx126x we wouldn't need it.

kokroo:
@caveman99  I believe S5NC answered your question for both of us.

@S5NC This confuses me. Why is the E22 definition present on boards that don't even have it?
And according to the information you just presented, am I correct to assume that the T-Echo and the T-Beam Supreme devices have the TCXO correctly turned on, within the latest Mesthastic beta?

src/mesh/SX126xInterface.cpp
```diff
@@ -25,13 +25,13 @@ template <typename T> bool SX126xInterface<T>::init()
...
 #endif
+    // Always power on the TCXO on SX126X devices, since all of them have a TCXO onboard
+    // which needs to be activated by setting DIO3 to high. Reference:
+    // https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/schematic/T-SX1262.pdf
 
-#ifndef SX126X_E22
-    float tcxoVoltage = 0; // None - we use an XTAL
-#else
     // Use DIO3 to power tcxo per https://github.com/jgromes/RadioLib/issues/12#issuecomment-520695575
     float tcxoVoltage = 1.8;
```
mverch67, line 33: If this place in the code is touched I suggest to make the tcxoVoltage configurable via variant files, as some boards may benefit from a different value than this default. E.g.

#ifndef TCXO_VOLTAGE
    float tcxoVoltage = 1.8;
#else
    float tcxoVoltage = TCXO_VOLTAGE;
#endif

mc-hamster:
TXCO only provides a measurable benefit with bw<=32k.

We can be conservative and say <=64k.

Advice not always turning it on unless there's a need for it.

@kokroo What's the reason for turning this on for your use case?

kokroo:
@mc-hamster I am using "Very Long Slow" inside a city, to achieve better range. It makes a huge difference and Meshtastic is impractical without it in a dense city like mine.

Even at BW ~ 62, it is unstable. BW 125 works flawlessly but is less reliable. A TCXO is a must-have in these conditions, especially with winter approaching. Some nodes are freezing outside while others are warm indoors.

As far as my armchair research over the last 1.5 years goes, a TCXO is not necessary above 62.5 KHz but it definitely improves message delivery rates even for higher bandwidths.

mc-hamster:
How can we get confidence this won't have any unintended effects?

What's the additional power consumption?

lyusupov:
majority of STM32WLE5 based modules/boards on the market contain **SX1262 with NO TCXO**:
https://github.com/lyusupov/SoftRF/blob/master/software/firmware/source/SoftRF/src/platform/STM32.h#L84-L88

PR #2807, opened 2023-09-18
Gps cleanup and powersave

jp-bennett:


src/gps/GPS.h
```diff
@@ -154,27 +156,14 @@ class GPS : private concurrency::OSThread
...
     static GPS *createGps();
 
   protected:
-    /// If possible force the GPS into sleep/low power mode
-    virtual void sleep();
-
-    /// wake the GPS into normal operation mode
-    virtual void wake();
-
     /** Subclasses should look for serial rx characters here and feed it to their GPS parser
```
mverch67, line 166: comment is not terminated... got compiler warnings

src/gps/GPS.h:168:5: warning: "/*" within comment [-Wcomment]

mverch67:
GPS.h line #185: void setNumSatellites(uint8_t n);

This method is never used (dead code?).

src/gps/GPS.cpp
```diff
@@ -563,24 +589,25 @@ int32_t GPS::runOnce()
...
         }
     }
     if (config.position.gps_enabled == false)
-        return 0;
+        return disable();
 
     // Repeaters have no need for GPS
-    if (config.device.role == meshtastic_Config_DeviceConfig_Role_REPEATER)
+    if (config.device.role == meshtastic_Config_DeviceConfig_Role_REPEATER) {
         disable();
```
mverch67, line 600: shouldn't it read *return* disable() here as well??

src/gps/GPS.cpp
```diff
@@ -650,21 +672,6 @@ int32_t GPS::runOnce()
     return isAwake ? GPS_THREAD_INTERVAL : 5000;
```
mverch67, line 672: The value GPS_THREAD_INTERVAL does not match the one in the comment above

code8buster:
Is there any way you can break out into a couple smaller sets of changes?
I'm leery of how many things are being packed into this; it can easily exceed my capacity in both time and hardware to test the corner cases of everything it touches.

src/gps/GPS.cpp
```diff
@@ -563,24 +589,25 @@ int32_t GPS::runOnce()
         }
         GPSInitFinished = true;
         if (config.position.gps_enabled == false) {
```
mverch67, line 591: Shouldn't this code be at the *beginning* of runOnce(). If GPS is disabled there shouldn't be any probing.

My tests: current behavior with GPS disabled:
DEBUG | ??:??:?? 4 [GPS] Probing for GPS at 9600 
INFO  | ??:??:?? 4 [GPS] Found a UBlox Module using baudrate 9600
DEBUG | ??:??:?? 4 [GPS] Module Info : 
DEBUG | ??:??:?? 4 [GPS] Soft version: 7.03 (45969)
DEBUG | ??:??:?? 5 [GPS] Hard version: 00040007
DEBUG | ??:??:?? 5 [GPS] Extensions:0
WARN  | ??:??:?? 6 [GPS] Unable to save GNSS module configuration.
DEBUG | ??:??:?? 6 [GPS] WANT GPS=0
INFO  | ??:??:?? 6 [GPS] Setting GPS power=0

After moving to the beginning:
DEBUG | ??:??:?? 4 [GPS] WANT GPS=0
INFO  | ??:??:?? 4 [GPS] Setting GPS power=0

src/gps/GPS.cpp
```diff
@@ -563,24 +589,25 @@ int32_t GPS::runOnce()
...
+            return disable();
         }
     }
     if (config.position.gps_enabled == false)
-        return 0;
+        return disable();
 
     // Repeaters have no need for GPS
-    if (config.device.role == meshtastic_Config_DeviceConfig_Role_REPEATER)
+    if (config.device.role == meshtastic_Config_DeviceConfig_Role_REPEATER) {
```
mverch67, line 599: Shouldn't this code be at the *beginning* of runOnce(). If GPS is disabled there shouldn't be any probing.

src/gps/GPS.cpp
jp-bennett: No, because GPS disabled does not mean there is no GPS. GPS disabled implies we want the GPS powered down, if it's present. And in some cases, to do a power down, we have to detect the GPS.

src/gps/GPS.cpp
jp-bennett: It probably should. Fixing.

src/gps/GPS.cpp
jp-bennett: That's... odd. Been that way for a while, too. We can try setting it to 200.

src/gps/GPS.cpp
mverch67: This is really really bad :(

It means that *all* devices that do not have GPS run into this lengthy 3x GPS probing with all the delays in it. Please try to find a solution to resolve this....

jp-bennett: We already have a couple solutions in place to deal with this. First, if the GPS pins aren't defined, the code knows to skip it. Which is now default on just about every device without a built-in GPS. And second, it does it lazy, so the rest of the boot can complete anyways.

mverch67: Or need a triple switch for gps_enabled: enabled/disabled/ignored

jp-bennett: GPS is going to be active on boot by default. We don't send any special code to tell it to turn on. On devices with a PMU or GPS enable line, maybe default off, I suppose. But take the Station G1 for instance. It has a port intended to add an external GPS, but it's just power and serial lines. On boot, that GPS chip is on and looking for satellites. To disable it, we first have to know that it's there.

mverch67: > We already have a couple solutions in place to deal with this. First, if the GPS pins aren't defined, the code knows to skip it. Which is now default on just about every device without a built-in GPS. And second, it does it lazy, so the rest of the boot can complete anyways.

I need to try this one... maybe because the old firmware forcibly defined the GPS pins they are still stored in the filesystem. Need to try a flash erase and check if it resolves this issue. Thanks.

mverch67: With my heltec_v3 it worked!

DEBUG | ??:??:?? 1 Read RTC time as 1695057317
DEBUG | ??:??:?? 1 Running without GPS.

jp-bennett: That unit doesn't have a GPS by default? If so, we need to remove the GPS PIN define from its variant.ini. Let the user set it if a GPS is installed.

mverch67: Yes, all the TLora boards don't have GPS built-in.

jp-bennett: I'll queue a separate PR to fix that, later today.

jp-bennett:
20 minutes with a GPS set to 120 second update times on a T-beam:
2.2.0: 187 mWh
This patch: 81 mWh

src/gps/GPS.cpp
```diff
@@ -433,30 +433,69 @@ bool GPS::setup()
...
+    }
+    isInPowersave = !on;
+    if (!standbyOnly && en_gpio != 0 &&
+        !(HW_VENDOR == meshtastic_HardwareModel_RAK4631 && (rotaryEncoderInterruptImpl1 || upDownInterruptImpl1))) {
+        LOG_DEBUG("GPS powerdown using GPS_EN_ACTIVE\n");
+        digitalWrite(en_gpio, on ? GPS_EN_ACTIVE : !GPS_EN_ACTIVE);
+        return;
+    }
+#ifdef HAS_PMU
+    if (pmu_found && PMU) { // Powers down GPU
```
mverch67, line 456: GPU or GPS?

src/modules/PositionModule.cpp
```diff
@@ -230,4 +230,42 @@ int32_t PositionModule::runOnce()
...
     return 5000; // to save power only wake for our callback occasionally
+}
+
+void PositionModule::handleNewPosition()
+{
+    meshtastic_NodeInfoLite *node = nodeDB.getMeshNode(nodeDB.getNodeNum());
+    const meshtastic_NodeInfoLite *node2 = service.refreshLocalMeshNode(); // should guarantee there is now a position
+    // We limit our GPS broadcasts to a max rate
+    uint32_t now = millis();
+    uint32_t intervalMs = getConfiguredOrDefaultMs(config.position.position_broadcast_secs, default_broadcast_interval_secs);
```
mverch67, line 241: src/modules/PositionModule.cpp: In member function 'void PositionModule::handleNewPosition()':
src/modules/PositionModule.cpp:241:14: warning: unused variable 'intervalMs' [-Wunused-variable]
     uint32_t intervalMs = getConfiguredOrDefaultMs(config.position.position_broadcast_secs, default_broadcast_interval_secs);

src/gps/GPS.cpp
jp-bennett: I blame too much coding after midnight. ><

Issue #2805, opened 2023-09-18
Set hop_limit to default when set by user to a bad value


RicInNewMexico:
Conversation with garth led to the thinking that, rather than setting hop_limit to the max of 7 when a bad (>7 hop) value is configured, we should maybe just cause invalid values to revert back to the default hop_limit of 3.

Thinking is that anyone setting their device to a value over the maximum of 7 is likely doing so arbitrarily and without doing sufficient research into the effect that hop count has on the overall network. As the network they are operating on may not be their own, especially in large event settings like Burning Man and Defcon, reverting back to the default on an invalid entry would help limit network flooding / congestion.

jp-bennett:
Ideally with a message, to avoid a stream of bug reports?

PR #2808, opened 2023-09-18
Deny maxhops to anyone who sets >7

code8buster:
#2805 

Is this even the place to make this change? I wonder if it shouldn't be handled in the saving of the lora config whenever a config change is processed.


caveman99 (approved):
I think it's just the right spot - this way we can document non-conformist behaviour in a post-mortem on the proto files ;-)

jp-bennett (approved):

G2Dolphin:
What is the reason for the new commit of setting explicitly 3 hop_limit rather than node's current setting?

Issue #2793, opened 2023-09-13
[Bug]: Bluetooth pairing always fails


mverch67:
### Category

BLE

### Hardware

heltec_v3, t3s3, t-deck, t-watch s3

### Firmware Version

2.2.3 - 2.2.6

### Description

Since firmware 2.2.3 the pairing with Android APP fails for devices when installed using flash erase. The APP constantly tries to re-connect but pairing fails when entering the proper 6 digit pin code.

Workaround:
Install firmware version 2.2.2, then connect with APP via bluetooth. After successful connect an update (but not install with flash erase) to firmware 2.2.3 -2.2.6 is possible which will then allow to connect with the APP via BT.

### Relevant log output

```
INFO  | ??:??:?? 4 [PowerFSM] Initialise the NimBLE bluetooth module
I NimBLEDevice: BLE Host Task Started
I NimBLEDevice: NimBle host synced.
D NimBLEDevice: >> setPower: 11 (type: 11)
D NimBLEDevice: << setPower
D NimBLEDevice: Setting bonding: 1, mitm: 1, sc: 1
D NimBLEServer: >> createService - 6ba1b218-15a8-461f-9fa8-5dcae273eafd
D NimBLEServer: << createService
D NimBLEService: >> start(): Starting service: UUID: 6ba1b218-15a8-461f-9fa8-5dcae273eafd, handle: 0xffff
D NimBLEService: Adding 3 characteristics for service UUID: 6ba1b218-15a8-461f-9fa8-5dcae273eafd, handle: 0xffff
D NimBLEService: << start()
D NimBLEAdvertising: >> Advertising start: customAdvData: 0, customScanResponseData: 0
primary service
           uuid 0x1800
         handle 1
     end_handle 5
characteristic
           uuid 0x2a00
     def_handle 2
     val_handle 3
   min_key_size 0
          flags [READ]
characteristic
           uuid 0x2a01
     def_handle 4
     val_handle 5
   min_key_size 0
          flags [READ]
primary service
           uuid 0x1801
         handle 6
     end_handle 9
characteristic
           uuid 0x2a05
     def_handle 7
     val_handle 8
   min_key_size 0
          flags [INDICATE]
ccc descriptor
           uuid 0x2902
         handle 9
   min_key_size 0
          flags [READ|WRITE]
primary service
           uuid 6ba1b218-15a8-461f-9fa8-5dcae273eafd
         handle 10
     end_handle 17
characteristic
[...]

PR #2809, opened 2023-09-18
Temp fix for S3 bluetooth

RicInNewMexico:
Need to roll back espressif to v6.3.2 to avoid the issues in #2793

### DIFF
--- a/arch/esp32/esp32.ini
+++ b/arch/esp32/esp32.ini
@@ -1,7 +1,7 @@
 ; Common settings for ESP targes, mixin with extends = esp32_base
 [esp32_base]
 extends = arduino_base
-platform = platformio/espressif32@^6.4.0
+platform = platformio/espressif32@6.3.2
 
 build_src_filter = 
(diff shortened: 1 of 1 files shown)


arch/esp32/esp32.ini
```diff
@@ -1,7 +1,7 @@
 ; Common settings for ESP targes, mixin with extends = esp32_base
 [esp32_base]
 extends = arduino_base
-platform = platformio/espressif32@^6.4.0
+platform = platformio/espressif32@6.3.2
```
caveman99, line 4: Change is fine, i'd recommend adding a comment to look for the root cause

caveman99 (changes requested):

arch/esp32/esp32.ini
RicInNewMexico: Done

thebentern (approved):

me639276639936:
Thanks!

PR #2813, opened 2023-09-18
Simplify SX126x variant configuration

S5NC:
All board variant files should function exactly the same as before this pull request. This means boards which worked before will continue to work, and also those which didn't will continue not working. Separate pull requests will be made to address issues with variant files for boards that are (and were) **(potentially)** misconfigured.

FIXME comments highlight future improvements

This PR is the simplification part of https://github.com/meshtastic/firmware/pull/2802


caveman99:
Please don't get me wrong, i am all for simplyfying stuff. BUT this looks to me like a rename spree. Plus it doesn't compile.

Simple question? Why is this neccessary and what do we gain from it except the hassle of checking every build manually?

thebentern:
Also please sign the CLA. We require this for any potential contributions.

thebentern:
Reviewed and everything looks good. 
I'm going to test w/ some of the key devices and put them on my TinySA to make sure.

variants/station-g1/variant.h
```diff
@@ -27,7 +27,9 @@
 #define SX126X_DIO1 LORA_DIO1
 #define SX126X_BUSY LORA_DIO2
 #define SX126X_RESET LORA_RESET
-// #define SX126X_E22 // Not really an E22
+// Not really an E22
+// #define SX126X_DIO2_AS_RF_SWITCH
```
neilhao, line 31: The line may need to be uncommented, because station g1 do use DIO2 to control RF switch. 
The next line is correct, Station G1 does not have a TCXO so there is no need to set the TCXO voltage.

thebentern (approved):

PR #2817, opened 2023-09-20
Fix for Pi Pico hang

jp-bennett:
The pico startup code was leaving _serial_gps set to null, but then failing to properly catch that state and short-circuit GPS probing. Fix by disabling the GPS thread when null.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -252,7 +252,7 @@ bool GPS::setup()
 {
     int msglen = 0;
 
-    if (_serial_gps && !didSerialInit) {
+    if (!didSerialInit) {
 #if !defined(GPS_UC6580)
         if (tx_gpio) {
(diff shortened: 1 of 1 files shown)


thebentern (approved):

jp-bennett:
Half a fix. Hang on.

NickDolfin:
2.2.8 does not work on the rp20 pico

jp-bennett:
> 2.2.8 does not work on the rp2040 pico

The first generated .zip does not have the second half of the fix in it.

jp-bennett:
@NickDolfin Can you try with the image generated here: https://github.com/meshtastic/firmware/actions/runs/6255294648#artifacts

NickDolfin:
@jp-bennett Yes, you are right. I was to fast and downloaded the wrong one.

Now 2.2.8 is working.

Many thanks for the fast fix.

PR #2819, opened 2023-09-22
Fix compression (#2806)

### LINKED ISSUE #2806: [Bug]: Compression does not work at all
### Category

Other

### Hardware

T-Beam

### Firmware Version

2.1.23.04bbdc6

### Description

Compression does not work on Meshtastic as of firmware version 2.1.23.04bbdc6.

I have been using VeryLongSlow to send messages. While doing some experiments I noticed that the logs report an airtime equal to non-compressed messages, even if the message being sent can be compressed.

The message is compressed, but the uncompressed version is transmitted and we have no savings on airtime. I can confirm this is the case because I compiled a custom firmware to manually set the portnum to 7, which transmitted the compressed payload with reduced airtime but unfortunately, it doesn't get decoded on the other end.

When I send this "compressed" message, it takes 27 seconds, even though the compressed size is 21 bytes.

Screenshot_2023-09-18_at_02 59 53

When I send something that is actually around ~21 bytes uncompressed, it only takes 6 seconds.

Screenshot_2023-09-18_at_03 02 39

You may use this test message without quotes :

"Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello"

If someone could fix this by Monday evening, I would be really grateful since I have a 7-day mountain hike coming up on Tuesday :D

### Relevant log output

_No response_

GUVWAF:
Fixes #2806.

The packet was encoded to protobufs before the `decoded` part was modified, so in protobufs encoding, still the old payload and portnum was used. Furthermore, while decompressing (which actually didn't happen because the portnum was not `TEXT_MESSAGE_COMPRESSED_APP`), the payload size wasn't modified.

Not sure if this has ever worked before, since there were no real changes in this part of the code since it was implemented.

### DIFF
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -382,9 +382,7 @@ meshtastic_Routing_Error perhapsEncode(meshtastic_MeshPacket *p)
 
     // If the packet is not yet encrypted, do so now
     if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
-        size_t numbytes = pb_encode_to_bytes(bytes, sizeof(bytes), &meshtastic_Data_msg, &p->decoded);
-
-        // Only allow encryption on the text message app.
+        // Only allow compression on the text message app.
         //  TODO: Allow modules to opt into compression.
         if (p->decoded.portnum == meshtastic_PortNum_TEXT_MESSAGE_APP) {
 
@@ -396,17 +394,12 @@ meshtastic_Routing_Error perhapsEncode(meshtastic_MeshPacket *p)
             int compressed_len;
             compressed_len = unishox2_compress_simple(original_payload, p->decoded.payload.size, compressed_out);
 
-            LOG_DEBUG("Original length - %d \n", p->decoded.payload.size);
-            LOG_DEBUG("Compressed length - %d \n", compressed_len);
+            LOG_DEBUG("Length - %d, compressed length - %d \n", p->decoded.payload.size, compressed_len);
             LOG_DEBUG("Original message - %s \n", p->decoded.payload.bytes);
 
             // If the compressed length is greater than or equal to the original size, don't use the compressed form
             if (compressed_len >= p->decoded.payload.size) {
-
                 LOG_DEBUG("Not using compressing message.\n");
-                // Set the uncompressed payload variant anyway. Shouldn't hurt?
-                // p->decoded.which_payloadVariant = Data_payload_tag;
-
                 // Otherwise we use the compressor
             } else {
                 LOG_DEBUG("Using compressed message.\n");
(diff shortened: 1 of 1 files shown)


thebentern (approved):

code8buster (approved):

kokroo:
@GUVWAF Mc-hamster did a great job integrating Unishox 2 but I guess all of us got mislead into thinking that compression is working, especially since Longfast has really short airtime. This bug became apparent only after I used VeryLongSlow.

Big thanks to you and @mc-hamster for working on compression!

I hope you were able to verify that compression actually works now? My long sample message which contained dozens of 'hello's should go across on VeryLongSlow in less than 8 seconds instead of 27 seconds. I'm currently far away from my laptop and civilisation so can't check it out myself unfortunately.

GUVWAF:
Yes, confirmed on VeryLongSlow: 

> DEBUG | ??:??:?? 66 [SerialModule] Add packet record (id=0x00453f8b fr=0x00 to=0xff, WantAck=1, HopLim=3 Ch=0x0 Portnum=1)

DEBUG | ??:??:?? 66 [SerialModule] Length - 227, compressed length - 20 
DEBUG | ??:??:?? 66 [SerialModule] Original message - Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello Hello H
DEBUG | ??:??:?? 66 [SerialModule] Using compressed message.
DEBUG | ??:??:?? 66 [SerialModule] Expanding short PSK #1
DEBUG | ??:??:?? 66 [SerialModule] Using AES128 key!
DEBUG | ??:??:?? 66 [SerialModule] enqueuing for send (id=0x00453f8b fr=0x39 to=0xff, WantAck=0, HopLim=3 Ch=0x59 encrypted)
DEBUG | ??:??:?? 66 [SerialModule] txGood=2,rxGood=0,rxBad=0
DEBUG | ??:??:?? 66 [RadioIf] Starting low level send (id=0x00453f8b fr=0x39 to=0xff, WantAck=0, HopLim=3 Ch=0x59 encrypted priority=64)
DEBUG | ??:??:?? 66 [RadioIf] (bw=62, sf=12, cr=4/8) packet symLen=65 ms, payloadSize=40, time 6045 ms
DEBUG | ??:??:?? 66 [RadioIf] AirTime - Packet transmitted : 6045ms

GUVWAF:
@kokroo FYI this has been reverted since it caused a breaking change with previous firmware versions (https://github.com/meshtastic/firmware/issues/2833). It will be added for the 3.0 release. It can be used though if you only use devices on 2.2.8.

PR #2824, opened 2023-09-24
Run handleNewPosition() for position packets from phone

thebentern:
Need to test it, but I think this should do the trick

### DIFF
--- a/src/modules/PositionModule.cpp
+++ b/src/modules/PositionModule.cpp
@@ -22,17 +22,6 @@ bool PositionModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mes
 {
     auto p = *pptr;
 
-    // If inbound message is a replay (or spoof!) of our own messages, we shouldn't process
-    // (why use second-hand sources for our own data?)
-
-    // FIXME this can in fact happen with packets sent from EUD (src=RX_SRC_USER)
-    // to set fixed location, EUD-GPS location or just the time (see also issue #900)
-    if (nodeDB.getNodeNum() == getFrom(&mp)) {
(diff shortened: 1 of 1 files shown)


thebentern:
On second review, it appears that these would get sweeped up in the runOnce for positionModule... however maybe we should explore consolidating the logic between the runOnce / handleNewPosition with regard to the smart broadcast mechanism and look at tweaking the runOnce interval. Seems like the on-demand model opens us up to the possibility that we could make the thread duty cycles less frequent? 
@jp-bennett thoughts?

thebentern:
Also, observed behavior that I'm looking into. GPS is disabled on this device and yet it gets probed.

jp-bennett:
> Seems like the on-demand model opens us up to the possibility that we could make the thread duty cycles less frequent?

I think it would be worthwhile to move in that direction. Maybe also call the new function from the Bluetooth thread?

And yes, the probe still runs with GPS disabled. On purpose. First, so we know what GPS we have to power save. And second, so we know what GPS we have for a triple-click.

You can disable probe by setting the GPS tx and rx pins back to 0, on most devices.

jp-bennett:
> Maybe also call the new function from the Bluetooth thread?

Maybe I should have looked at the patch before commenting. What does come to mind is how often does the phone send a location update? Would driving down the highway spam location packets too quickly? Other than that, this seems like a reasonable continuation of the GPS refactor.

jp-bennett (approved):

garthvh:
I push new locations from the phone as fast as 5 seconds

jp-bennett:
> I push new locations from the phone as fast as 5 seconds

Should we re-use the GPS interval setting to also rate limit when the phone is the source?

PR #2830, opened 2023-09-25
RP2040: Add SerialModule support 

### LINKED ISSUE #2825: [Bug]: Pico rp2040 serial output not working
### Category

Serial

### Hardware

Raspberry Pi Pico (W)

### Firmware Version

2.2.8

### Description

Dear programmer,

I am trying to get a serial output port to work.
I tried pins gp4+gp5 (uart1). There is some activity on reboot, but looks like gibberisch on a (serial) terminal like putty or 'feasycom serial port' program
In addition, when I use different pins eg. go0/gp1 (uart0) or gp8/gp9 (uart1) no output is seen.

I used e.g:

meshtastic --set serial.enabled true --set serial.echo true --set serial.mode TEXTMSG --set serial.baud BAUD_115200 --set serial.rxd 9 --set serial.txd 8
Connected to radio
Set serial.enabled to true
Set serial.echo to true
Set serial.mode to TEXTMSG
Set serial.baud to BAUD_115200
Set serial.rxd to 9
Set serial.txd to 8

Question? Is the serial output available in the rp2040?
If yes, how do I get it to work?

Thanks

### Relevant log output

_No response_

GUVWAF:
Fixes #2825. Tested with `serial.txd 8` and `serial.rxd 9`. 

Had to increase the `RX_BUFFER` to 256, otherwise long packets would occasionally fail. I saw that for ESP32 it even [throws an error]( https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/HardwareSerial.cpp#L580.) if you have it set to 128 or lower.

### DIFF
--- a/src/modules/SerialModule.cpp
+++ b/src/modules/SerialModule.cpp
@@ -47,7 +47,7 @@
 #if (defined(ARCH_ESP32) || defined(ARCH_NRF52) || defined(ARCH_RP2040)) && !defined(CONFIG_IDF_TARGET_ESP32S2) &&               \
     !defined(CONFIG_IDF_TARGET_ESP32C3)
 
-#define RX_BUFFER 128
+#define RX_BUFFER 256
 #define TIMEOUT 250
 #define BAUD 38400
 #define ACK 1
--- a/src/modules/SerialModule.cpp
+++ b/src/modules/SerialModule.cpp
@@ -188,7 +188,7 @@ int32_t SerialModule::runOnce()
                     }
                 }
             }
-#if !defined(TTGO_T_ECHO) && !defined(ARCH_RP2040)
+#if !defined(TTGO_T_ECHO)
             else {
                 while (Serial2.available()) {
                     serialPayloadSize = Serial2.readBytes(serialBytes, meshtastic_Constants_DATA_PAYLOAD_LEN);
--- a/src/modules/SerialModule.cpp
+++ b/src/modules/SerialModule.cpp
@@ -44,7 +44,8 @@
 
 */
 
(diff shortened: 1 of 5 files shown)


src/modules/SerialModule.cpp
```diff
@@ -44,9 +44,10 @@
 
 */
 
-#if (defined(ARCH_ESP32) || defined(ARCH_NRF52)) && !defined(CONFIG_IDF_TARGET_ESP32S2) && !defined(CONFIG_IDF_TARGET_ESP32C3)
+#if (defined(ARCH_ESP32) || defined(ARCH_NRF52) || defined(ARCH_RP2040)) && !defined(CONFIG_IDF_TARGET_ESP32S2) &&               \
+    !defined(CONFIG_IDF_TARGET_ESP32C3)
 
-#define RX_BUFFER 128
```
thebentern, line 49: I'll have to go back and check, but didn't we run into an issue with larger serial buffers on ESP32s and had to crank it down?

thebentern (approved):

src/modules/SerialModule.cpp
mc-hamster: @thebentern See:

https://www.esp32.com/viewtopic.php?t=8858

mc-hamster: Setting to rx buffer to 256 may work now, but that suggests there was previously a problem.

mc-hamster: Setting the buffer to a large value means that the serial module isn't running frequently enough. Maybe the processor is being taxed and the can't get the data in time?

thebentern: > @thebentern See:
> 
> https://www.esp32.com/viewtopic.php?t=8858

Ah yes! I think it was bellyaching in the logs if I'm remembering correctly.

Issue #2832, opened 2023-09-26
[Bug]: GPS Module not found on Nano G1 Explorer


Eric-Sparks:
### Category

Hardware Compatibility

### Hardware

Other

### Firmware Version

2.2.7.e8970ad.alpha

### Description

When upgrading from 2.2.4 alpha, and before, to 2.2.7 alpha the GPS module on the Nano G1 Explorer isn't detected.  I've been able to recreate this problem on two of four devices I have.  Upgrading to the latest firmware (2.2.8 alpha) does not fix the problem.

jp-bennett:
Can you check the position configuration, whether GPS_RX_PIN and GPS_TX_PIN are set differently in the units that are failing to detect GPS? The boot log output of the working and non-working devices would be helpful as well. Thanks!

jp-bennett:
I believe this should be fixed in the 2.2.10 Alpha release. Let us know if there are further problems.

PR #2834, opened 2023-09-26
Add GPS pin definitions missed in revamp

jp-bennett:


jp-bennett:
Probably closes #2832

code8buster (approved):

thebentern (approved):

PR #2843, opened 2023-09-28
Maybe fix TBEAM 1.2 GPS

jp-bennett:
Wild, blind thrashing about, that just maybe fixes a GPS problem.


src/Power.cpp
```diff
@@ -562,7 +562,7 @@ int32_t Power::runOnce()
     }
 #endif
     // Only read once every 20 seconds once the power status for the app has been initialized
```
mverch67, line 564: update comment to 60 seconds as well

PR #2851, opened 2023-10-01
Add timeout to ublox PMREQ command

jp-bennett:
Some U-blox GPS chips don't seem to properly wake up from the PMREQ command on receiving data on the serial port. That command does allow specifying a timeout, so calculate and do so, to work around that issue.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -433,7 +433,7 @@ GPS::~GPS()
     notifyGPSSleepObserver.observe(&notifyGPSSleep);
 }
 
-void GPS::setGPSPower(bool on, bool standbyOnly)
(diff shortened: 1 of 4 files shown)


thebentern (approved):

mverch67:
> Some U-blox GPS chips don't seem to properly wake up from the PMREQ command on receiving data on the serial port. That command does allow specifying a timeout, so calculate and do so, to work around that issue.

I read that when the UBLOX sleeps it cannot communicate over the serial, but it requires a toggle on the RX pin. Did you try that?

jp-bennett:
> it requires a toggle on the RX pin. Did you try that?

https://github.com/meshtastic/firmware/blob/master/src/gps/GPS.cpp#L491 is where we do that. It works on most models, but at least one user has a GPS that does not wake this way. So setting the timer as well.

PR #2854, opened 2023-10-02
make esp32 deepsleep button wakeup functional again

mverch67:
- fixed missing brackets in original deepsleep code
- added wakeup reason log info
- added resume screen
- tested with tbeam and t-deck (tracker mode)

```
//\ E S H T /\ S T / C

INFO  | ??:??:?? 0 Booted, wake cause 3 (boot count 2), reset_reason=ext1 RTC_CNTL
```


mverch67:
I simply added the wakeCause for the button trigger, but I don't know if the original suppress logic in main.cpp makes sense. For the tracker you don't see any effects when you're outside, because it immediately falls asleep before even trying to turn on the screen. Inside the screen also stays black which is probably not intended.

PR #2856, opened 2023-10-02
Next Hop-based routing with fallback to managed flooding

GUVWAF:
## Description

This adds a NextHopRouter for direct messages, which only relays if it is the next hop for a packet. The next hop is set by the current relayer of a packet, which bases this on information from  a previous successful delivery to the destination via flooding. 
Namely, in the PacketHistory, we keep track of (up to 3) relayers of a packet. When the ACK is delivered back to us via a node that also relayed the original packet, we use that node as next hop for the destination from then on. This makes sure that only when there’s a two-way connection, we assign a next hop. Both the ReliableRouter and NextHopRouter will do retransmissions (the NextHopRouter only 1 time). For the final retry, if no one actually relayed the packet, it will reset the next hop in order to fall back to the FloodingRouter again. Note that thus also intermediate hops will do a single retransmission if the intended next-hop didn’t relay, in order to fix changes in the middle of the route.

It is backwards compatible with all 2.x versions of Meshtastic, because for those nodes it will always fall back to flooding.

### Implementation details
The `next_hop` and `relay_node` are added to the unencrypted PacketHeader. Since we only used 14 bytes out of the 16-byte header (it got this size due to memory alignment), the two unused bytes are used to save the last byte of the NodeNum of the next hop and current relayer. When `hop_start` was added (version 2.3), these bytes were set to 0, so we can use these bytes safely when `hop_start` is set. 
The re-use of header space means that there is no additional over-the-air overhead, except that up to 2 retransmissions are needed when a change of next hop occurs. So, the benefit will be most pronounce for rather static meshes, but since the next hop is only set after a successful two-way connection is set up, and it falls back to flooding rather quickly, even for dynamic meshes there is likely a benefit to using this.

In terms of memory overhead, the `next_hop` has to be stored in the NodeDB (only 1 byte), and there are 4 additional bytes required per packet in the PacketHistory to store the `next_hop` and three relayers. 

Using only 1 byte for the `next_hop` means that there is a chance that the last byte of two nodes match, and then they will both try to relay. However, that’s not a big issue as that would be similar to flooding. The chance that only the intended next hop relays depends on the amount of nodes that can hear the packet. If that are 10 nodes (that would be a lot), the chance that only the next hop relays is 83.7% (`(255/256*254/256*253/256 ... 247/256*)*100`), for 5 nodes the chance is 96.1% (`(255/256*254/256*253/256*252/256)*100`).

## Examples 
With this, you get rid of the unnecessary rebroadcasts caused by flooding like the one from node 0 as shown below. (Note that an arrow to a node means that it received it (so there might be multiple arrows for one packet), but not necessarily that it was addressed.)
image
With the NextHopRouter, node 0 doesn't try to relay:
image

Furthermore, it solves this issue of the current implementation where the wrong node (1, because it has the lowest SNR) is relaying:
image
With the NextHopRouter, due to randomness in the order of rebroadcasting, at some point the route will succeed and 2 will be set as next hop from then on.
image

## Notes for reviewers 

NextHopRouter inherits from the FloodingRouter. Since it also requires retransmissions, this logic is now moved from the ReliableRouter to the NextHopRouter, and the ReliableRouter inherits from the NextHopRouter. 
Also, since the Router and NextHopRouter need to have access to the PacketHistory as well, the Router now inherits from it instead of only the FloodingRouter.


loodydo:
I have started looking. May take me some to to finish. I am not a statistician so I asked ChatGPT about the probability that the last byte of two nodes match. It came up with the result of 16.31%. Considering that this is a non breaking change I would think that the risk of two nodes broadcasting is still worth it.

GUVWAF:
>  I am not a statistician so I asked ChatGPT about the probability that the last byte of two nodes match. It came up with the result of 16.31%. Considering that this is a non breaking change I would think that the risk of two nodes broadcasting is still worth it.

It depends on the amount of nodes that can hear the packet (so the amount of immediate neighbors). But indeed, the chance that then *only* the intended next hop relays is a bit different than I presented before. I've updated the description with two calculations.

src/mesh/RadioInterface.h
```diff
@@ -31,6 +31,12 @@ typedef struct {
 
     /** The channel hash - used as a hint for the decoder to limit which channels we consider */
     uint8_t channel;
+
+    // Last byte of the NodeNum of the next-hop for this packet
+    uint8_t next_hop;
+
+    // Last byte of the NodeNum of the current relayer of this packet
+    uint8_t current_relayer;
```
loodydo, line 39: If I am fallowing the code correctly "current_relayer" indicates where the packet just came from. For me it took me a moment to understand this since "current" often implies something that is active or happening at the present moment. I would suggest something like "last_relay_node" for the variable name.

GUVWAF: I understand why this can be confusing, but it depends on whether you look at it from the perspective of the transmitter or the receiver. For the transmitter, it would then need to set the `last_relay_node` of a packet it is about to sent to itself, which is also a bit strange. I took 'current' because a node it currently preparing/transmitting/receiving/processing this packet. Or would it be better to just call it `relay_node`?

loodydo: I think I understand but I don't know that I have a full picture yet of the code. My suggestion would be to error on the side of how the variable is used in logic. My impression is that it mostly gets used from the perspective of what was the last relay node. For clarity, I would then suggest adding a comment when transmitting "We are/will be the last_relay_node".

loodydo: Maybe the part I don't understand is "a packet it is about to sent to itself".

GUVWAF: I'm sorry, I mean that it has to set the `last_relay_node` to its own NodeNum before transmitting that packet.

src/mesh/NextHopRouter.cpp
```diff
@@ -0,0 +1,106 @@
...
+                    if (traceRouteModule && traceRouteModule->wantPacket(tosend))
+                        traceRouteModule->updateRoute(tosend);
+                    // If it is a neighborInfo packet, update last_sent_by_id
+                    if (neighborInfoModule && neighborInfoModule->wantPacket(tosend))
+                        neighborInfoModule->updateLastSentById(tosend);
+                }
+
+                tosend->hop_limit--; // bump down the hop count
+                NextHopRouter::send(tosend);
+            } else if (p->next_hop == 0) {
```
loodydo, line 69: Suggest avoiding using the magic number 0 by defining a constant. For example:

constexpr NodeNum NO_PREFERENCE = 0;

src/mesh/NextHopRouter.cpp
```diff
@@ -0,0 +1,106 @@
...
+                    LOG_DEBUG("Update next hop of %x to %x based on received ACK.\n", p->from, p->current_relayer);
+                    sentTo->next_hop = p->current_relayer;
+                }
+            }
+        }
+    }
+
+    if (config.device.role != meshtastic_Config_DeviceConfig_Role_CLIENT_MUTE) {
+        if ((p->to != getNodeNum()) && (getFrom(p) != getNodeNum())) {
+            if (p->next_hop == getNodeNum()) {
```
loodydo, line 54: If a function is added (part of NeighborInfoModule?) that determines if we only have one neighbor that could help us avoid "tip of tree" rebroadcasts. For example:

if (p->next_hop == getNodeNum() && !hasOnlyOneNeighbor()) {

src/mesh/NextHopRouter.cpp
```diff
@@ -0,0 +1,106 @@
...
+                        neighborInfoModule->updateLastSentById(tosend);
+                }
+
+                tosend->hop_limit--; // bump down the hop count
+                NextHopRouter::send(tosend);
+            } else if (p->next_hop == 0) {
+                // No preference for next hop, use FloodingRouter
+                LOG_DEBUG("No preference for next hop, using FloodingRouter\n");
+                FloodingRouter::sniffReceived(p, c);
+            } else if (p->to == NODENUM_BROADCAST) {
```
loodydo, line 73: This would also change if we add "tip of tree" detection.

else if (p->to == NODENUM_BROADCAST && !hasOnlyOneNeighbor()) {

src/mesh/NextHopRouter.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+void NextHopRouter::sniffReceived(const meshtastic_MeshPacket *p, const meshtastic_Routing *c)
+{
+    bool isAck =
+        ((c && c->error_reason == meshtastic_Routing_Error_NONE)); // consider only ROUTING_APP message without error as ACK
+    if (isAck) {
+        // Update next-hop of this successful transmission to current relayer, but ONLY if "from" is not 0 or ourselves (means
+        // implicit ACK or someone is relaying our ACK)
+        if (p->from != 0 && p->from != getNodeNum()) {
+            if (p->current_relayer) {
```
loodydo, line 42: Seeing this logic here puts me on board with changing "current_relayer" to "relay_node". This code would then read 'was there a relay node?' rather than 'was there a current relayer?' or even as I had suggested 'was there a last relay node'. Since there doesn't seem to be multiple relayers or relay steps tracked within a single packet transmission, I don't think it needs the additional designator.

src/mesh/NextHopRouter.cpp
loodydo: It sound questionable if "tip of tree" detection is desirable so I may end up deleting these "tip of tree" related comments.

GUVWAF:
I pushed a new commit to resolve @loodydo’s comments. Also added the hop limit setting of the original transmitter to the header flags (using 3 of the 4 currently unused bits), such that we can determine the amount of hops the packet has already traveled. This comes in handy to set the next hop for immediate neighbors, and I think it’s also useful for displaying in the apps. Currently it will always show the SNR/RSSI of nodes in the list, but this is actually the SNR/RSSI to the last relayer. In cases that a node is not an immediate neighbor, we could display the hops towards that node.

I also changed the way how the `next_hop` and `relay_node` are stored in the MeshPacket and NodeDB. Now only the last byte is stored, since only that byte is sent over the air. Besides, it mitigates the case where you might assign the wrong NodeNum if you happen to have multiple nodes with the same last byte in the NodeDB.

GUVWAF:
I think this PR is ready to merge, except that it’s a breaking change unfortunately. 
I did a test and it seems that the two unused bytes are not set to zero with current firmware. Meaning with this PR it would think someone set the next hop for a specific node and no one will relay the message. 
We’ll have to wait for 3.0 with this.

src/mesh/ReliableRouter.cpp
```diff
@@ -227,9 +228,23 @@ int32_t ReliableRouter::doRetransmissions()
...
+                if (config.lora.next_hop_routing && p.packet->to != NODENUM_BROADCAST) {
+                    if (p.numRetransmissions == 1) {
+                        // Last retransmission, reset next_hop (fallback to FloodingRouter)
+                        p.packet->next_hop = NO_NEXT_HOP_PREFERENCE;
+                        // Also reset it in the nodeDB
+                        meshtastic_NodeInfoLite *sentTo = nodeDB->getMeshNode(p.packet->to);
+                        if (sentTo) {
+                            LOG_DEBUG("Resetting next hop for packet with dest 0x%x\n", p.packet->to);
+                            sentTo->next_hop = NO_NEXT_HOP_PREFERENCE;
+                        }
```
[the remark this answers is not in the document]
GUVWAF: Yes, `getMeshNode()` returns a pointer to the node, so it will modify its `next_hop` in the DB.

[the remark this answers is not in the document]
mcumings: Been a while since I was working with mutable shared data structures.  thx.

roha-github:
If you use 1 byte, then I suggest modulo 251 of the whole name represented by UINT_64. The probability is then 0.4%

GUVWAF:
@roha-github What do you mean with "whole name represented by UINT_64" and which probability are you referring to?

roha-github:
@GUVWAF 

"I used the 2 unused bytes to save the last byte of the NodeNum of the next hop and current relayer" ... 

don't "last byte of the NodeNum", use NodeNum modulo 251 (max 1 byte prim number) as filter

GUVWAF:
The NodeNum is only 32 bits and based on the MAC address of the device.
Assuming that that number is uniformly distributed I don't get why using modulo of the largest prime number is a better approach.

Where does your 0.4% probability refer to?

roha-github:
UINT32 modulo 251 = 0..250
1/251 = 0.4%
Using prim numbers for checksum is a common pattern

GUVWAF:
I'm sorry but I don't get why that is better than using the last byte, which is 0..255.

It's not a checksum, it's to uniquely identify a node. If the NodeNum is already uniformly distributed, taking the last byte should not have any bias.

roha-github:
About the routing itself ...

Each node could regularly publish a broadcast of its direct neighbors and communicate the reception quality of RSSI and SNR.

The list is sorted by reception quality and only the best 32 nodes are published.

1 byte or 8 bits are used per node:
Bit 1 = RSSI quality > -115 dB
Bit 2 = SNR quality > -11 dB
Bit 3..8 = NodeNum mod 64

Together with the node DB and the GPS coordinates, either the node itself or the app (smartphone, web browser, CLI) can then create a network topology.

Only 7 bytes would be required for a route with 7 hops.

Translated with DeepL.com (free version)

caveman99:
That's what the neighborinfo module does.

roha-github:
> That's what the neighborinfo module does.

Ok, now I understand the benefits of the Neighbor Info Module.

Should the node do the routing itself later? Does the node DB then store all the neighbors of each node? How much memory would this require? I once read somewhere that the node DB would only store up to 80 nodes.

The documentation on the Neighbor Info Module does not really encourage activation. The module would be needed for routing and ever larger mesh networks.

tve:
> I think this PR is ready to merge, except that it’s a breaking change unfortunately.
> I did a test and it seems that the two unused bytes are not set to zero with current firmware. Meaning with this PR it would think someone set the next hop for a specific node and no one will relay the message.
> We’ll have to wait for 3.0 with this.

I do not understand the "wait for 3.0", most likely I don't understand what deploying 3.0 means... I would assume that nodes in a mesh will be upgraded slowly, especially big meshes, like SoCalMesh with hundreds of nodes, some in pretty inaccessible locations. This means that whether it's called 3.0 or not the next hop routing will have to co-exist with 2.x. Unless you count on all the 2.X having been upgraded to a version that sets next_hop to zero by then...

I'm wondering whether it would be possible to use the NeighborInfo to gate next-hop routing. If the NeighborInfo can have a flag that says "I do next-hop routing" then on the outgoing side a relayer knows whether populating next-hop makes sense or not.

On the incoming side it's a bit trickier in that a relayer would look up the NeighborInfo based on relay_node. If relay_node matches a NeighborInfo that says it does next-hop routing then the relayer would obey next_hop, although of course the match may be by chance. As a special case, if hop_start is set and equal to hop_limit then IIUC the message is received from the original sender and thus NeighborInfo can be looked up using the from address.

In any case, it seems worth it to figure out some heuristics to make 3.0 interoperable with pre-2.3, which introduced the explicit next_hop field as far as I can tell and not rely on "it's going to take so long for 3.0 to become available that everyone will have upgraded past 2.3 by then"...

Apologies if I'm completely off-base and don't understand some key fact here.

tve:
Shouldn't the NeighborInfo module be enabled by default?

GUVWAF:
Version 3.0 will most likely have a different default modem preset (people have advocated for MediumFast) or another sync word to deliberately make nodes incompatible. 

I'm not really sure half-working interoperable solutions are the way to go here, especially if it requires the NeighborInfo module to be installed enabled.
It's not that this PR has a huge impact anyway, as it only applies to DMs while I think the majority of packets are still broadcasts.

NeighborInfo creates quite a lot of additional traffic and is not necessary for NextHop routing, so I don't think it should be enabled by default.

KyleMaas:
> Version 3.0 will most likely have a different default modem preset (people have advocated for MediumFast) or another sync word to deliberately make nodes incompatible.

@GUVWAF I'd like to chime in here and say that I think this part has some major downsides.  This threatens to split deployed infrastructure between old and new and drop coverage levels for both old and new until everything, including all client devices, have been changed over.  For anyone who has gone further with deployment than just tinkering, this could be a major undertaking just due to the number of devices.  And for anyone with infrastructure on mountaintops or other places which are difficult to access, these devices can be a pain in the neck to try to update.  The more people using the same settings, the better it is for everyone's coverage, and the experience for new users would be better if they already have an existing network to connect to without having to build it all out themselves.  Changing the default modem preset should not be done lightly.

And I agree with @tve's comments about version interoperability.  A thoughtful and gradual upgrade path would make it far more likely for people to actually upgrade.  Ideally you want people to upgrade to the new version so that you can continue to innovate moving forward.  If the changes are too drastic, you risk having people stuck with "good enough, works for me" instead, or even potentially having someone maintain an old compatible version as an LTS release.  Or looking at the project/protocol as so unstable that it's not worth continuing to deploy.

I love the idea of this, but a gradual rollout would make it a much easier pill to swallow.

GUVWAF:
Ofcourse the decision will not be taken lightly and people have been saying it will be happing at least a year from now.

I'm not entirely following why you're advocating for gradual changes, as precisely that leads to "good enough, works for me", instead of doing all breaking changes at once, where you **have** to upgrade to continue to talk to others that have.

garthvh:
> I love the idea of this, but a gradual rollout would make it a much easier pill to swallow.

This pull is not merged because it is a breaking change that needs to wait until 3.0, and there is not enough other stuff to make a breaking change worth it.  There is no way to do this gradually.  Meshtastic is just not set up for multiple versions, when version 3 gets promoted version 2 is going to be deprecated.

tve:
> There is no way to do this gradually.

I'm not convinced, that's why I made suggestions above. Yes, there's no way to make the changes without compromises, but I don't know that the compromises have been explored or discussed. Quite possibly just my ignorance.

> Meshtastic is just not set up for multiple versions, when version 3 gets promoted version 2 is going to be deprecated.

This is one of the biggest issues I have with the proposal: it uses up all the unused bits for its purposes and doesn't fix the versioning issue which causes the incompatibility nightmare. This kind'a guarantees that no more non-breaking changes can be made going forward which I'd rate as "not good".

At a high level I'm wondering what the plan is for wide-area meshes, which is where the routing improvements have the most impact I believe. I'm part of the southern california mesh with hundreds of nodes spread over hundreds of square km, many mountain ranges dividing the geography, and many nodes in difficult to access/update places. My overall assessment is that the mesh as a whole is pretty unusable other than to toy around and experiment (which is cool). If the goal is to make this type of mesh usable over longer distances (like sending messages from Santa Barbara to San Diego -- approx 350km) some serious thought needs to go into routing. Maybe the answer is that this should be done via MQTT (internet), maybe the answer is that there should be an overlay backbone network to bridge long distances, maybe the answer lies in just improving the current type of routing. I don't know and it also depends a lot on what the project goals are.

I'm not sure what the best place for this type of discussion is, but I believe it's relevant here because if there's going to be a breaking change then at least make the most out of it. I'm also not trying to piss on this PR: I actually think it's great, I'm just wondering about the bigger picture.

GUVWAF:
> This is one of the biggest issues I have with the proposal: it uses up all the unused bits for its purposes and doesn't fix the versioning issue which causes the incompatibility nightmare.

But this is really beyond the scope of this PR. Currently these two bytes are useless anyway, because before 2.3 they could take any value depending on what was in that memory location. This is just one of the breaking changes currently proposed for 3.0 (fixing text compression is another one), I believe a change to the header to make it future-proof deserves its own PR.

> My overall assessment is that the mesh as a whole is pretty unusable other than to toy around and experiment (which is cool).

What kind of issues are you having and what makes you believe this has to do with routing? For broadcasts the current managed flooding approach is really not too bad. I'm not convinced there even exists a protocol that's guaranteed to be better in all the different use cases for Meshtastic (especially fixed or mobile nodes, or a mix), that doesn't require an internet connection and can run on microprocessors with limited RAM. For "smarter" routing you need periodic control messages, but those eat up the same bandwidth as regular traffic. 

Other problems with wireless mesh networks, e.g. the hidden node problem or unstable/changing RF conditions are not solved with better routing.

KyleMaas:
@GUVWAF Yeah, starting that discussion to ask about forward compatibility is what led me into joining the discussion on this issue.  I have several places in mind to add nodes that would vastly increase coverage in my area, but they are inaccessible enough that once they're up there, I'm probably realistically only going to be able to update the firmware maybe every 12 months or so.  So I need to wait for a version of the firmware that has enough interoperability, or have all of the proposed protocol changes interoperable enough with older versions, that I can actually make these nodes available to the community.  All of the current nodes I run are in areas which are at least reasonably accessible.  But there's no sense in me taking on the time, effort, and expense of building out the expansions if it's going to be completely obsolete in a month - it makes more sense to wait for stability and not recommend Meshtastic to anyone who isn't already in our local coverage zones.  So protocol stability and interoperability is extremely important to me.

garthvh:
3.0 is like 18 months away best case. Honestly not really enough there yet to even consider it.

roha-github:
@GUVWAF "Version 3.0 will most likely have a different default modem preset (people have advocated for MediumFast)"

Where are such far-reaching decisions discussed and decided? What exactly is the background? The minimal increase in bandwidth is "bought" with a loss of range and all 2.x routers & repeaters.

Alternatively, 3.x clients switch to Long / Fast or both networks run in parallel and apart.

GUVWAF:
> Where are such far-reaching decisions discussed and decided?

This was discussed by admins on Discord, but is also just an idea right now. When 3.0 comes closer, there will likely be an [RFC](https://github.com/meshtastic/rfcs) for this. 
Let's now continue this discussion in #3932.

GUVWAF:
After looking at this again and realized I made a big assumption that’s likely often not valid. I set the `next_hop` of a node to the one that was forwarding the ACK/response of a DM for that node. However, with asymmetric links, it might well be that that node is not the best `next_hop`, and it might even be you can’t reach it, while it can reach you. In my simulator I don’t simulate asymmetric links, but judging from how often people mention they receive “beacons” very well, but can’t send a message to someone, I think this is quite common.

Another thing I didn’t consider is that only the original transmitter will fall back to flooding when it doesn’t hear anyone rebroadcast, but it might also be that intermediate hops loose their `next_hop` link. In my fork I’ve fixed this by letting the intermediate hops also retransmit once with fall back to flooding, but I’m unsure about the possible overhead this creates. 

At least before the first issue is resolved, I’m not confident on merging this yet. Happy to hear any suggestions on this.

daramos:
> I think this PR is ready to merge, except that it’s a breaking change unfortunately. I did a test and it seems that the two unused bytes are not set to zero with current firmware. Meaning with this PR it would think someone set the next hop for a specific node and no one will relay the message. We’ll have to wait for 3.0 with this.

@GUVWAF  I got curious about this and dug around the git history a bit. I think maybe a way to determine if you can depend on these bytes being initialized is to see if `mp->hop_start` is non-zero? 
Looks like both #3479 and #3321 were both first publicly released in [v2.3.2.63](https://github.com/meshtastic/firmware/releases/tag/v2.3.2.63df972).

Since they were both publicly released in the same official release, I think you could use `mp->hop_start` being set as a condition to trust the last two bytes being zero-initialized. If `hop_start` is non-zero then the transmitting node is at least running a fw version that also initialized the unused bytes. 

I think the only way this doesn't hold true is if someone compiled a firmware themselves between those two commits.

Unfortunately these really are the last free bits in the header. I dug around a bit see if there was a backwards-compatible way to add a trailing struct at the end of the packet for expansion but I don't see a way that would be possible with the way the rest of the packet is immediately memcpy'ed into a encrypted meshpacket variant.

But - I wonder if it would be possible to use the bytes to indicate protocol version? The benefit would be a way to accomplish mostly backwards compatible extensibility in the transport layer. 

For instance, the first byte could indicate what version of the packet header is used (`packet_version`). 
The second byte could be used to record the size of an extended header struct (`extended_header_size`). Firmware versions that are not up to date can use this field to skip bytes that are not in it's own extended header struct and get the payload data.

for instance:
 `packet_version`: 0, `extended_header_size`: 0 - describes the current situation we're in. Both bytes are initialized to 0. 

`packet_version`: 1, `extended_header_size`: 1 - We added a new byte field to the header (maybe a flags field), we'll increment the version number as well so that nodes running the latest firmware knows we are compatible with the new features. Older firmware nodes won't have the new feature but it knows to skip a byte after the header to start reading the payload.

`packet_version`: 2, `extended_header_size`: 1 - We've added a new flag or changed some functionality but didn't grow the header.

This of course would only be possible for non-broadcast messages currently but eventually if most clients upgrade to a version that support the extended header - we could deprecate the old handling of packet headers and use the extended header broadcast messages as well.

I think this also allows room for V3 without totally breaking interoperability with V2. 

 I don't know enough about Lora's MTU to know if it's even possible to increase the total packet size that way and this is probably the wrong place to discuss it anyway but I figured I'd throw that out here. 
Thank you to all the developers of this project; you guys are awesome!

GUVWAF:
@daramos Indeed using hopStart to determine whether the bytes are valid is something that was discussed on Discord, and would be doable. 

While I’m open to other ideas for those bytes, I’m not sure if a protocol version and extended header is worthwhile. 

First, you will still need a breaking change to introduce this the first time, because old firmware wouldn’t know about an extended header. 
Second, everything you add cannot be required; it can only be optional. If you keep on working with nodes that do not use a new functionality you introduced, those might break it. Sometimes a breaking change is for the better. Adding optional things can also be done in the encrypted part of the payload, which would still work for all nodes on a common channel.
Third, with this you can only add things to the header, and not change the existing one. It would thus only lead to more and more overhead, which is not insignificant for LoRa. We are already hitting almost the packet size limit for certain types.
Lastly, there are other things beside the header that might break compatibility. For example, a new default LoRa modem preset, new rules for who rebroadcasts a packet first, breaking changes to the protobuf encoding in the encrypted part, etc.

daramos:
Makes sense, thanks!

KodinLanewave:
So I was thinking... Is there any reason NeighborInfo packets _must_ honor the same number of hops as other packet types?  For NextHopRouting this could default to zero hops (only transmitted to immediately-connected nodes) as an example, severely reducing overhead on a high-saturation mesh.  Any thoughts?  I figure if you want telemetry from a node you can manually query NeighborInfo with a DM to get that data "phoned home" if you will, and otherwise neighbors are only sent to their adjacent nodes for supporting NextHop operation.

GUVWAF:
@KodinLanewave By default, if you don't have the NeighborInfo Module enabled, you'll also not rebroadcast those packets, so essentially they are already 0-hop if others are not interested in them. 

I agree it would be nice to have different intervals for LoRa, PhoneAPI and MQTT (or disable one of them), but it also complicates things quite a lot. Maybe something for Meshtastic 3.0.

thebentern (approved):

todd2982:
I like the concept of this a lot.  The only thing I'd like to mention that I don't see so far is nodes on the move.  If a user receives a message while in a car/plane/rowboat/etc., the best route back could have changed rapidly between receiving the message and typing a reply.  This is even more of an issue if the sender, or nodes along the original route are also moving.  I think some checks against GPS data for movement speed should go into deciding if failing back to flood routing is required.

This may not be feasible with the default GPS polling intervals, but moving through the landscape should be taken into account somehow.  I also understand that flood routing is an automatic fallback, but if we can smartly choose the best option we should.

todd2982:
I understand.  I just wanted to bring it up.  Since I first read this on the project board that scenario was stuck in my head.  GPS is what made sense to me, but I mainly just wanted to have the people smarter than me (I?) considering the same scenario.

I look forward to seeing how this works out!

fifieldt:
@GUVWAF , do you have any suggestions on how one may test this?

GUVWAF:
@fifieldt You could flash it on a couple of devices and send some DMs or traceroutes. If all is well, these should be at least as reliable as before, and if so, it should also be more efficient, but that would be most pronounced if all nodes are updated. 
Especially going from e.g. a direct connection to one with a hop in between would be interesting to test. In my testing it switched seamlessly. 

If you want to see it's really assigning the next-hops, you will have to look at the serial logs.

medentem:
@GUVWAF now that moving nodes and asymmetric links are modeled in the simulator, maybe we should test how this performs under those conditions...

GUVWAF:
@medentem I did some crude simulations using the interactive simulator by letting it assign next-hops and then remove the nodes (using `remove <id>`) to ensure it falls back to flooding after 1 retransmission. I’ve also hacked in asymmetric links (by enlarging the Tx power for some nodes) to make sure it only assigns a next-hop when you can hear *and* transmit towards the node.

Indeed I’ve not done large-scale simulations with moving nodes, which might be worth doing if I (or someone) finds the time. However, the benefit of using dedicated relayers for DMs is much more obvious. Even if you only use the next-hop once and the next time you have to fall back to flooding again because the node moved, the penalty is just one retransmission, whereas by not using flooding you’ll have likely saved multiple transmissions.

wlockwood:
> I have started looking. May take me some to to finish. I am not a statistician so I asked ChatGPT about the probability that the last byte of two nodes match. It came up with the result of 16.31%. Considering that this is a non breaking change I would think that the risk of two nodes broadcasting is still worth it.

I don't think that's accurate. If we assume that the node IDs are completely random - they're probably not! - it would be a 1 out of 256, which is a 0.4% chance. Given that these are hardware values assigned by the manufacturer, there's a good chance that they are not random and instead incrementing, but since this is the final byte and thus should change the most frequently if it's the least significant byte, it should be close enough to random for the purposes of this PR. ... I wanted to validate that, but I don't have a convenient data source to do so.

fifieldt (approved):
Read the code thoroughly, nice and clean. Logic makes sense and couldn't find any obvious flaws.

fifieldt:
@GUVWAF , I took some NextHopRouter nodes into the wild for some testing - mountaintops and streets, and everything worked at least as well as before :)

GUVWAF:
Yes, that's fine.

Thanks everyone for the comments and reviews. Looking forward to see it in action in version 2.6 soon 😀

PR #2861, opened 2023-10-03
UBLOX- Don't stop searching before acquiring fix

code8buster:
the ubx GNSS receiver with this flag flipped will not enter the "inactive for search" state. It will enter the inactive state until the update period ticks once the fix is acquired.

The update period remains the same.


jp-bennett:
Does it seem to make cold-start fixes quicker?

jp-bennett (approved):

code8buster:
> Does it seem to make cold-start fixes quicker?
I desoldered the battery to make sure the battery backup couldn't keep the RTC or old positions.

Tbeam cold start times:
PR2861 = 80 seconds with stock ceramic or 18mm antenna

Firmware 2.2.9 alpha:
150s with stock antenna
100s with 18mm

There may have been an almanac/ephemeris update mid-test, and 2.2.9 didn't generate a position for 1200+ seconds.

jp-bennett:
I like it. Let's do it.

PR #2873, opened 2023-10-07
[WIP] find channelIndex for decoded packets by using channels.decryptForHash

### LINKED ISSUE #2466: [Bug]: New Channel on NodeInfo is usually not correct.
### Category

Other

### Hardware

Not Applicable

### Firmware Version

2.1.xx

### Description

The new channel index provided on node info is not proving to be very accurate.

I would think the channel number should be different than the default of 0 is the two nodes share a secondary but not primary channel on the network where the node info's channel > 0

Instead what happens is that usually a node is misidentified as having come from another network when it is local.

On iOS the channel displays on a node if it is anything other than 0

### Relevant log output

_No response_

sbias:
fixes #2466

### DIFF
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -299,8 +299,8 @@ bool perhapsDecode(meshtastic_MeshPacket *p)
         config.device.rebroadcast_mode == meshtastic_Config_DeviceConfig_RebroadcastMode_ALL_SKIP_DECODING)
         return false;
 
-    if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag)
-        return true; // If packet was already decoded just return
+    //if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag)
+    //    return true; // If packet was already decoded just return
 
     // assert(p->which_payloadVariant == MeshPacket_encrypted_tag);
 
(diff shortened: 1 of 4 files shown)


GUVWAF (approved):

src/mesh/Router.cpp
```diff
@@ -316,15 +316,18 @@ bool perhapsDecode(meshtastic_MeshPacket *p)
...
     // Try to find a channel that works with this hash
     for (ChannelIndex chIndex = 0; chIndex < channels.getNumChannels(); chIndex++) {
         // Try to use this hash/channel pair
         if (channels.decryptForHash(chIndex, p->channel)) {
+            if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
+                // packet is not encoded, just set channel and return
+                p->channel = chIndex; 
+                return true;
+            }
+
```
andrekir, line 330: for decoded MeshPackets `p->channel` is already channel index, so `decryptForHash()` won't work since it can't match parameters (p->channel isn't channelHash):

[bool Channels::decryptForHash(ChannelIndex chIndex, ChannelHash channelHash)](https://github.com/meshtastic/firmware/blob/master/src/mesh/Channels.cpp#L294)

src/mesh/Router.cpp
```diff
@@ -317,6 +318,7 @@ bool perhapsDecode(meshtastic_MeshPacket *p)
             memcpy(bytes, p->encrypted.bytes,
                    rawSize); // we have to copy into a scratch buffer, because these bytes are a union with the decoded protobuf
             crypto->decrypt(p->from, p->id, rawSize, bytes);
+            p->channel = chIndex;
```
sbias, line 321: in encrypted packets the channel index is set by the sending node which have to set to local channel index AFTER decoding (unwrapping)

caveman99:
@sbias is this ready ? (apart from the formatting...)

PR #2884, opened 2023-10-11
Update variant.h

loodydo:
Swapping I2C ports to support sensors module.

It would seem that sensors don't like to work on I2C ports other than 0.
This change makes the external IO pins used for I2C set to port 0 rather than 1.
This then requires the internal I2C port used for the OLED to change from port 0 to port 1.

### DIFF
--- a/variants/station-g1/variant.h
+++ b/variants/station-g1/variant.h
@@ -1,10 +1,12 @@
 // #define BUTTON_NEED_PULLUP // if set we need to turn on the internal CPU pullup during sleep
 
-#define I2C_SDA 21
-#define I2C_SCL 22
+// i2c channel on external IO connector. NOTE sensors have had trouble on I2C ports other than 0
+#define I2C_SDA 14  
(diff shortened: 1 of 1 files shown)


mverch67:
There is a more elegant way to solve this in a generic way which will work for all devices, not just a single one.

loodydo:
@mverch67 How would you know which pins to use for I2C port 0 and which pins to use for I2C port 1 for all devices in a generic way? Or I guess fixing the sensors needing to use port 0 would be a very nice way to solve the problem.

mverch67:
I published an enhancement in #2891 which obsoletes this PR.

mverch67:
> Or I guess fixing the sensors needing to use port 0 would be a very nice way to solve the problem.

You can try this artifact if it works for you:
https://github.com/meshtastic/artifacts/blob/device/pr/pr2891-firmware-2.2.12.4721059.zip?raw=true

PR #2888, opened 2023-10-13
refactor: always use channel hash in `Router::send()`

andrekir:
This PR fixes channel routing for decoded MeshPackets, ensuring they are received in the proper message channels by using the channel `hash` for all MeshPackets sent over the wire.


garthvh (approved):
Beautiful

src/mesh/Router.cpp
```diff
@@ -316,15 +323,18 @@ bool perhapsDecode(meshtastic_MeshPacket *p)
         config.device.rebroadcast_mode == meshtastic_Config_DeviceConfig_RebroadcastMode_ALL_SKIP_DECODING)
         return false;
 
-    if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag)
```
thebentern, line 319: Is there any case in which we might have an already decoded a packet before we call perhapsDecode and this short circuit would still be desired?

andrekir: good catch. indeed, this can only run once.

thebentern (approved):

src/mesh/Router.cpp
```diff
@@ -248,6 +248,13 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
     // If the packet is not yet encrypted, do so now
     if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
         ChannelIndex chIndex = p->channel; // keep as a local because we are about to change it
+        auto hash = channels.setActiveByIndex(chIndex);
```
mverch67, line 251: The methods perhapsEncode() and perhapsDecode() are protected by a concurrency lockGuard, that prevents that the crypto lib (which is not reentrant) is run by two threads in parralel and possibly crashes. Now that you introduced encryption in another method (send) this possibly also requires a lockGuard for encryption here.

mverch67 (changes requested):
This PR touches some delicate spots in the code where we faced concurrency issues in the past. I recommend extensive testing of these changes with several channels and excessive text messaging, preferably on an esp-S3 which is easier to crash.

src/mesh/Router.cpp
andrekir: good points. will use a different method here and not change perhapsEncode().

andrekir:
for a bit more context, the goal here is to have the same `channel` field [logic](https://github.com/meshtastic/protobufs/blob/master/meshtastic/mesh.proto#L855-L858) for both `encrypted` and `decoded` MeshPackets.

this change would make it easier to handle `decoded`packets (from over the wire) into the appropriate local channels.

but first I need to confirm where `decoded` can be handled from outside ("over the wire") so this approach has value. so far seems MQTT only.

src/mesh/Router.cpp
GUVWAF: I believe the `LockGuard` can be safely removed since we moved to ArduinoThread, because with that the device will never be simultaneously in two function calls. Seems to be a TODO since then: https://github.com/meshtastic/firmware/blob/4a867c81c05583a5f36259d60d38309eb8e936fe/src/concurrency/OSThread.h#L26

caveman99: > I believe the `LockGuard` can be safely removed since we moved to ArduinoThread, because with that the device will never be simultaneously in two function calls. Seems to be a TODO since then:
> 
> https://github.com/meshtastic/firmware/blob/4a867c81c05583a5f36259d60d38309eb8e936fe/src/concurrency/OSThread.h#L26

The lockguard was added later than that comment, i.e. fairly recently. The comment by @mverch67 about the BT Stacksize is valid too, we never really found out where the S3 crashes came from and suspected concurrency. 

My 2 cents: leave it there, it doesn't hurt and just eats a few cpu cycles.

GUVWAF: Ah, yes, it was added to the crypto function recently. Yes, we can keep it, but I don't think the change here will lead to crashes again.

andrekir:
closing for now. this would require a significant refactor to make sure decoding only runs once, and a fix for the MQTT issue was addressed in #2892.

PR #2891, opened 2023-10-14
fix/enhance: allow telemetry sensors connected to second I2C port

mverch67:
When connecting a telemetry sensor to the I2C port 1 the sensor is detected during the I2C scan but cannot be used. Only sensors connected to I2C port 0 are functional.

```
INFO  | ??:??:?? 15 [EnvironmentTelemetryModule] Environment Telemetry: Initializing
INFO  | ??:??:?? 15 [EnvironmentTelemetryModule] Init sensor: BMP280
WARN  | ??:??:?? 15 [EnvironmentTelemetryModule] Could not connect to detected BMP280 sensor.
```

The telemetry sensor classes use only the global variable "Wire" even the sensor was detected on "Wire1".

This PR fixes this behavior by passing the wire pointer to the sensor class using an enhanced version of the nodeTelemetrySensorsMap. 

```
INFO  | ??:??:?? 14 [EnvironmentTelemetryModule] Environment Telemetry: Initializing
INFO  | ??:??:?? 14 [EnvironmentTelemetryModule] Init sensor: BMP280
INFO  | ??:??:?? 14 [EnvironmentTelemetryModule] Opened BMP280 sensor on i2c bus
DEBUG | ??:??:?? 15 [EnvironmentTelemetryModule] BMP280Sensor::getMetrics
INFO  | ??:??:?? 15 [EnvironmentTelemetryModule] (Sending): barometric_pressure=949.862549, current=0.000000, gas_resistance=0.000000, relative_humidity=0.000000, temperature=23.020000, voltage=0.000000
```

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -29,6 +29,7 @@
 #include "target_specific.h"
 #include <Wire.h>
 #include <memory>
+#include <utility>
 // #include <driver/rtc_io.h>
 
 #include "mesh/eth/ethClient.h"
@@ -122,9 +123,8 @@ uint32_t serialSinceMsec;
 
 bool pmu_found;
 
-// Array map of sensor types (as array index) and i2c address as value we'll find in the i2c scan
-uint8_t nodeTelemetrySensorsMap[_meshtastic_TelemetrySensorType_MAX + 1] = {
-    0}; // one is enough, missing elements will be initialized to 0 anyway.
+// Array map of sensor types with i2c address and wire as we'll find in the i2c scan
+std::pair<uint8_t, TwoWire *> nodeTelemetrySensorsMap[_meshtastic_TelemetrySensorType_MAX + 1] = {};
 
 Router *router = NULL; // Users of router don't care what sort of subclass implements that API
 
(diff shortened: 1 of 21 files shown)


src/main.cpp
```diff
@@ -122,9 +123,8 @@ uint32_t serialSinceMsec;
 
 bool pmu_found;
 
-// Array map of sensor types (as array index) and i2c address as value we'll find in the i2c scan
-uint8_t nodeTelemetrySensorsMap[_meshtastic_TelemetrySensorType_MAX + 1] = {
-    0}; // one is enough, missing elements will be initialized to 0 anyway.
+// Array map of sensor types with i2c address and wire as we'll find in the i2c scan
+std::pair<uint8_t, TwoWire *> nodeTelemetrySensorsMap[_meshtastic_TelemetrySensorType_MAX + 1] = {};
```
thebentern, line 127: Neato! So this is a Tuple?

thebentern (approved):

src/main.cpp
mverch67: This is a "pair" from the C++ STL. Tuples also exist in STL but are two or more types while pair is exactly two types.

PR #2903, opened 2023-10-22
Hydra variant rectification

S5NC:
Removes unecessary definitions, makes RF switching configuration match actual PCB and schematic, and corrects SX1262 power output configuration for actual 1W output.


kkwestt:
please add next line
 **#define USE_SX1268**

PR #2909, opened 2023-10-26
INA3221 Sensor Implementation

RicInNewMexico:
This PR, and the associated protobufs PR, enable use of the TI INA3221 3 Channel Voltage / Current sensor.


RicInNewMexico:
I'm aware this is probably horribly formatted code and there may be better ways to accomplish some things.. but for now.. it compiles without any warnings or errors for me and functionality has been tested on 3 different devices.

PR #2917, opened 2023-10-28
RTTTL ringtones on T-Deck / T-Watch S3 and potentially more I2S audio enabled devices

thebentern:


variants/t-watch-s3/variant.h
```diff
@@ -30,6 +30,10 @@
 
 #define TFT_BL ST7789_BACKLIGHT_EN
 
+#define DAC_I2S_BCK 48
```
mverch67, line 33: T-Deck defines a _HAS_I2S_ ... is it missing here?

src/AudioThread.h
```diff
@@ -0,0 +1,67 @@
...
+class AudioThread : public concurrency::OSThread
+{
+  public:
+    AudioThread() : OSThread("AudioThread") { initOutput(); }
+
+    void beginRttl(const void *data, uint32_t len)
+    {
+        setCPUFast(true);
+        rtttlFile = new AudioFileSourcePROGMEM(data, len);
+        i2sRtttl = new AudioGeneratorRTTTL();
```
mverch67, line 25: where is this allocated memory released?

src/AudioThread.h
```diff
@@ -0,0 +1,67 @@
...
+#define AUDIO_THREAD_INTERVAL_MS 25
+
+class AudioThread : public concurrency::OSThread
+{
+  public:
+    AudioThread() : OSThread("AudioThread") { initOutput(); }
+
+    void beginRttl(const void *data, uint32_t len)
+    {
+        setCPUFast(true);
```
mverch67, line 23: shouldn't the CPU clock be restored afterwards? Otherwise it consumes more (battery) power.

danumbaman:
Not sure if this is the correct place but I would love for a way to not only have audible pings for messages. I would like a option for device is charging and device is fully charged.

PR #2919, opened 2023-11-01
WIP LoRAType

thebentern:


markbirss:
I have had a look at what possible changes are needed but just cant get the device i have to flash at all



Have referenced the product pages, but just cant flash
the 2 buttons are broken on top even tried soldering wires directly to esp32 chippu (en) and G0

https://github.com/AutomationArt/LoraType/tree/main/GettingStarted
https://raw.githubusercontent.com/AutomationArt/LoraType/main/Image/LoraType_ForDIY.png

@thebentern 

i could share the bin file if you want to try it

PR #2920, opened 2023-11-01
Change default Coding Rate to 4/5 for <=LongFast

GUVWAF:
Increases throughput by about 1.5x at the cost of a little bit of sensitivity (according to "_Experimental Evaluation of the Packet Reception Performance of LoRa_" by Guo, Q. et al. it's less than 1 dB). LoRaWAN is also using only 4/5 (see [RP002-1.0.1](https://lora-alliance.org/wp-content/uploads/2020/11/rp_2-1.0.1.pdf) Section 3.1.2). 
This is non-breaking as the CR is sent in the LoRa header. I tested that two devices with CR of 4/5 and 4/8 can talk to each other.

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -384,27 +384,27 @@ void RadioInterface::applyModemConfig()
         switch (loraConfig.modem_preset) {
         case meshtastic_Config_LoRaConfig_ModemPreset_SHORT_FAST:
             bw = (myRegion->wideLora) ? 812.5 : 250;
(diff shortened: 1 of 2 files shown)


GUVWAF:
@rcarteraz I think only this table would need to be changed: https://meshtastic.org/docs/overview/radio-settings#predefined-channels
The data rate there also needs to be recalculated (or retrieved from the initial logs when a node boots).

thebentern (approved):

caveman99 (approved):

PR #2930, opened 2023-11-06
Add E22-400M30S support for Hydra

S5NC:
Add support for E22-400M30S, based on @kkwestt's [request](https://github.com/meshtastic/firmware/pull/2903#issuecomment-1794513391). Please test on your board, we have not tested SX1268 configured output power vs E22-**400**M30S output power. If you have an SA I strongly recommend that you set the output power to a low value, such as +5 dBm and measure the output power, increasing up to 22 if necessary for you to reach 30 dBm output power from your module.


thebentern (approved):

sniporbob:
If someone in the US has one of these and wants to ship it to me for testing, I'll put it on the spectrum analyzer to verify the proper setting before shipping it back.

PR #2933, opened 2023-11-07
Add support for the rak10701 board, no touch

HookdomPonix:
Add support for the RAK10701 Field Tester device.  You must flash from the RUI3 firmware to Arduino before use.  

The TFT and the button on the side work, but the touch screen will require more work as its not the same libs as the ESP32 TFT/touch devices.

### DIFF
--- a/src/graphics/TFTDisplay.cpp
+++ b/src/graphics/TFTDisplay.cpp
@@ -105,6 +105,10 @@ class LGFX : public lgfx::LGFX_Device
 
 static LGFX tft;
 
+#elif defined(RAK14014)
+#include <TFT_eSPI.h>
+TFT_eSPI tft = TFT_eSPI();
+
 #elif defined(ST7789_CS)
 #include <LovyanGFX.hpp> // Graphics and font library for ST7735 driver chip
 
@@ -327,7 +331,7 @@ static TFT_eSPI tft = TFT_eSPI(); // Invoke library, pins defined in User_Setup.
 
 #endif
 
-#if defined(ST7735_CS) || defined(ST7789_CS) || defined(ILI9341_DRIVER)
+#if defined(ST7735_CS) || defined(ST7789_CS) || defined(ILI9341_DRIVER) || defined(RAK14014)
 #include "SPILock.h"
 #include "TFTDisplay.h"
 #include <SPI.h>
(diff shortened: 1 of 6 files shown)


src/graphics/TFTDisplay.cpp
```diff
@@ -484,16 +496,21 @@ bool TFTDisplay::connect()
...
 #elif defined(T_DECK) || defined(PICOMPUTER_S3)
     tft.setRotation(1); // T-Deck has the TFT in landscape
 #elif defined(T_WATCH_S3)
     tft.setRotation(2); // T-Watch S3 left-handed orientation
 #else
     tft.setRotation(3); // Orient horizontal and wide underneath the silkscreen name label
-#endif
     tft.fillScreen(TFT_BLACK);
+#endif
     return true;
```
caveman99, line 514: That change eliminates the fillScreen call for all other variants which is a break. What happens if you put the call back after the endif and remove it in the RAK14014 condition?

caveman99 (changes requested):
See comments inline please.

variants/rak10701/platformio.ini
```diff
@@ -0,0 +1,18 @@
+; The very slick RAK wireless RAK10701 Field Tester device.  Note you will have to flash to Arduino bootloader to use this firmware.  Be aware touch is not currently working.
+[env:rak10701]
+extends = nrf52840_base
+board = wiscore_rak4631
+build_flags = ${nrf52840_base.build_flags} -Ivariants/rak10701 -D RAK_4631
+  -L "${platformio.libdeps_dir}/${this.__env__}/BSEC2 Software Library/src/cortex-m4/fpv4-sp-d16-hard"
+  -DGPS_POWER_TOGGLE ; comment this line to disable triple press function on the user button to turn off gps entirely.
+build_src_filter = ${nrf52_base.build_src_filter} +<../variants/rak4631> +<mesh/eth/> +<mesh/api/> +<mqtt/>
+lib_deps =
```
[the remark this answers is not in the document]
HookdomPonix: Fixed in the latest commit.  Thanks for the heads up.

src/graphics/TFTDisplay.cpp
HookdomPonix: Thanks for the catch.   Fixed and recompiled/works on the 10701.

caveman99 (approved):

thebentern (approved):

geeksville:

https://meshtastic.discourse.group/t/rak10701-field-tester-as-a-meshtastic-device/8406/7

PR #2935, opened 2023-11-08
Add missing defaulting for HAS_SENSOR to configuration.h

S5NC:
Add missing `HAS_SENSOR` defaulting. The value is used twice in `/src/modules/Modules.cpp` ([source](https://github.com/search?q=repo%3Ameshtastic%2Ffirmware%20HAS_SENSOR&type=code))


GUVWAF:
Actually this is not necessary (https://stackoverflow.com/questions/5085392/what-is-the-value-of-an-undefined-constant-used-in-if), but it's good for consistency.

PR #2941, opened 2023-11-15
Added Remove node by nodenum admin message

### LINKED ISSUE #1761: [Feature request] Add ability to delete/remove individual nodes from nodedb
Request ability to delete/remove individual nodes from nodedb, rather than clearing the entire database. It is useful to keep some historical nodes in the database, such as those for lost devices, old devices that may come online again (such as fixed nodes with dead batteries).

Since `--set reset-nodedb` is currently used for clearing nodedb, recommend syntax `--set remove-nodedb <node name>` or similar.

I do not recommend allowing manually _adding_ nodes as that could be considered tainting or 'poisoning' the node database with false or incorrect node information.

thebentern:
Closes #1761

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -327,7 +327,7 @@ void NodeDB::removeNodeByNum(uint nodeNum)
 {
     int newPos = 0, removed = 0;
     for (int i = 0; i < *numMeshNodes; i++) {
-        if (meshNodes[i].num == nodeNum)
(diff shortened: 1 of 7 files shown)


caveman99 (approved):

caveman99:
There's similar code here... https://github.com/meshtastic/firmware/blob/6f09cead4c57a628c7be733f9dd88d83dd47d4e2/src/mesh/NodeDB.cpp#L847C1-L872C6 - bonus points for folding this into the change :-)

PR #2942, opened 2023-11-16
Standardise order for setting GPIO pin default values

S5NC:
In cases where a digitalWrite is (clearly) called on a pin, and then pinMode is set for that same pin, the two lines are flipped such that the pinMode is set before calling the digitalWrite.


GUVWAF:
Yes, I think it makes sense to call `pinMode` before `digitalWrite`. 
Can you run `trunk fmt` to resolve the CI errors?

PR #2943, opened 2023-11-16
YAML based config for PI / Portduino

thebentern:


src/platform/portduino/PortduinoGlue.cpp
```diff
@@ -94,6 +100,39 @@ void portduinoSetup()
     printf("Setting up Meshtastic on Portduino...\n");
 
 #ifdef ARCH_RASPBERRY_PI
+    YAML::Node yamlConfig;
+    try {
+        yamlConfig = YAML::LoadFile("config.yaml");
```
mverch67, line 105: This is not a good (but inefficient and expensive) way in C++ to check if a file exist. Pls. don't misuse the YAML class for this and run purposely into exceptions. Also, there are several types of YAML exceptions, e.g. it could be that there is a syntax error in the file. 
It's better (and the code structure becomes easier) to use the standard C++ way to find the file and if it is found then use YAML to parse the file.

jp-bennett: That's fair. Something like access() a better option?

src/main.cpp
```diff
@@ -690,15 +691,31 @@ void setup()
...
-        rIf = new SX1262Interface((LockingArduinoHal *)RadioLibHAL, 21, 16, 18, 20);
-        if (!rIf->init()) {
-            LOG_WARN("Failed to find SX1262 radio\n");
-            delete rIf;
-            rIf = NULL;
-        } else {
-            LOG_INFO("SX1262 Radio init succeeded, using SX1262 radio\n");
+    if (settingsMap[use_sx1262]) {
+        if (!rIf) {
+            PiHal *RadioLibHAL = new PiHal(1);
```
mverch67, line 696: where is this memory released if the radio is not sx1262?

src/main.cpp
```diff
@@ -690,15 +691,31 @@ void setup()
...
+                LOG_WARN("Failed to find SX1262 radio\n");
+                delete rIf;
+                rIf = NULL;
+            } else {
+                LOG_INFO("SX1262 Radio init succeeded, using SX1262 radio\n");
+            }
+        }
+    } else if (settingsMap[use_rf95]) {
+        if (!rIf) {
+            PiHal *RadioLibHAL = new PiHal(1);
```
mverch67, line 709: Variable PiHal goes out of scope at the end of the if-block. Where is the memory released?

src/main.cpp
jp-bennett: Honestly, if we can't init the radio as requested by the config, we should exit().

PR #2951, opened 2023-11-19
Retain device nodeinfo during reset-nodedb

### LINKED ISSUE #2939: [Feature Request]: When reseting NodeDB, ignore the current node
### Platform

ESP32

### Description

We should delete all _other_ nodes in the NodeDB. This is unnecessary because it will get restored and it is unhelpful because it removes the last position, effectively making fixed_position nodes lose their lat / long.

RicInNewMexico:
Did the thing.. 

Closes #2939

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -317,7 +317,6 @@ void NodeDB::installDefaultChannels()
 void NodeDB::resetNodes()
 {
     devicestate.node_db_lite_count = 1;
-    // memset(devicestate.node_db_lite, 0, sizeof(devicestate.node_db_lite));
     std::fill(&devicestate.node_db_lite[1], &devicestate.node_db_lite[MAX_NUM_NODES - 1], meshtastic_NodeInfoLite());
     saveDeviceStateToDisk();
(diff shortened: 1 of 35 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -316,8 +316,8 @@ void NodeDB::installDefaultChannels()
 
 void NodeDB::resetNodes()
 {
-    devicestate.node_db_lite_count = 0;
-    memset(devicestate.node_db_lite, 0, sizeof(devicestate.node_db_lite));
+    devicestate.node_db_lite_count = 1;
+    std::fill(&devicestate.node_db_lite[1], &devicestate.node_db_lite[MAX_NUM_NODES - 1], meshtastic_NodeInfoLite());
```
thebentern, line 320: Is the current device's nodeinfo guaranteed to be in this index?

RicInNewMexico: It's the first entry generated after the nodedb is initialized and I don't believe there are any functions or processes that would cause the ordering of nodes in the database to be re-sorted or re-arranged. To the best of my knowledge it should remain as the first entry in the nodedb.

PR #2955, opened 2023-11-21
Modify hydra config

Fexiven:
Since theres also the option for 433 Mhz i think this config should be included too

### DIFF
--- a/variants/diy/hydra/variant.h
+++ b/variants/diy/hydra/variant.h
@@ -17,8 +17,8 @@
 #define LED_PIN 2           // add status LED (compatible with core-pcb and DIY targets)
 
 // Radio
-//#define USE_SX1262 // E22-900M30S uses SX1262 - 868Mhz
-//#define USE_SX1262 // E22-300M30S uses SX1268 - 433Mhz
+#define USE_SX1262 // E22-900M30S uses SX1262 - 868Mhz
+#define USE_SX1262 // E22-300M30S uses SX1268 - 433Mhz
 
 #define SX126X_MAX_POWER                                                                                                         \
     22 // Outputting 22dBm from SX1262 results in ~30dBm E22-900M30S output (module only uses last stage of the YP2233W PA)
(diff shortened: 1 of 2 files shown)


variants/diy/hydra/variant.h
```diff
@@ -17,7 +17,9 @@
 #define LED_PIN 2           // add status LED (compatible with core-pcb and DIY targets)
 
 // Radio
-#define USE_SX1262 // E22-900M30S uses SX1262
+//#define USE_SX1262 // E22-900M30S uses SX1262 - 868Mhz
+//#define USE_SX1262 // E22-300M30S uses SX1268 - 433Mhz
+
 #define SX126X_MAX_POWER                                                                                                         \
```
caveman99, line 23: Sorry, this change will stop the hydra target precompiled binaries from working. there's no problem defining both chips, as you can see in the t-beam variant.h

caveman99 (changes requested):

variants/diy/hydra/variant.h
Fexiven: I defined both modules then, i wasn't sure about it

caveman99:
That should work fine now, please sign the CLA

caveman99 (approved):

jp-bennett (commented):
Looks good to me. No hardware here to actually test the change.

Fexiven:
I have an tinyVNA but I think that's not suitable for testing so high power modules?

I got an flipper zero, that's not very scientific, but I might see how much dBi it's emitting

I will try later

Fexiven:
Sadly i cant get the module to work outside of meshtastic with minimal config.... Can anyone help? Meshtastic is sadly the only firmware that gets it magically running

Fexiven:
I tried with setting different TX levels... but my Flipper shows -36dbm no matter what i set (1-30)... so i guess my Flipper is incorrect (which is very very likely its not made to be a spectrum analyzer)

i have two of these neat boards, but i don't know how i can get the second board to be a power logger, i don't get the RadioLib to work by myself...

PR #2956, opened 2023-11-22
Add USB detection to RAK4631 based boards.

HookdomPonix:
This PR will add support for detecting if the USB cable is plugged in on the RAK4631-derived boards.   It may also work for other NRF52 systems, but I don't have any I can test.  

If you wish to test on an NRF board, simply add the define of NRF_APM to your variant.h and let me know how it works out.  

The check runs on the power status update thread, so it can take 20 seconds to detect USB or no USB.  

We assume the system is charging if there is a battery and USB is plugged in.

### DIFF
--- a/.github/workflows/build_raspbian.yml
+++ b/.github/workflows/build_raspbian.yml
@@ -1,0 +1,45 @@
+name: Build Raspbian
+
+on: workflow_call
+
+permissions:
+  contents: write
+  packages: write
+
+jobs:
+  build-raspbian:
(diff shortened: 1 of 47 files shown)


caveman99 (approved):

thebentern (approved):

HookdomPonix:
Seems functional.  Some systems will report a batterymv even when there is none - thats the charge current sharing a line with the batterymv detect line.

PR #2960, opened 2023-11-23
Make STM compile again and update toolchain.

caveman99:
The binary is too big for the flash. WIP


geeksville:

https://meshtastic.discourse.group/t/linking-failed-to-generate-firmware-elf-cannot-find-larm-cortexm4lf-math-for-wio-e5-variant-lora-e5-branch/11050/2

jp-bennett:
Set MAX_NUM_NODES super low in your variant.h, and see if it gives you enough headroom.

caveman99:
> Set MAX_NUM_NODES super low in your variant.h, and see if it gives you enough headroom. I merged locally and attempted this. Getting errors
> 
> ```
> /home/jbennett/.platformio/packages/framework-arduinoststm32/system/Drivers/CMSIS/Device/ST/STM32WLxx/Include/stm32wle5xx.h:993:47: error: expected ')' before '*' token
>   993 | #define RNG                     ((RNG_TypeDef *) RNG_BASE)
>       |                                  ~            ^
> .pio/libdeps/rak3172/Crypto/RNG.cpp:159:10: note: in expansion of macro 'RNG'
>   159 | RNGClass RNG;
> ```

The RNG stuff is from a double macro definition in the used crypto lib. i might swap that out for another one. locally it builds with a modified copy. Maxnodes is a good pointer...

jp-bennett:
Another flash saving potential is the unishox library. It's only actually used for ATAK packets. #ifdef it out of existence, to save a few more bytes.

goshawk22:
Another idea for saving flash is to remove all the I2C stuff. It wouldn't be needed for a minimal build but is still included as far as I can tell.

thebentern (approved):

PR #2963, opened 2023-11-26
L76 soft standby

thebentern:
Will test it tomorrow

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -492,6 +492,7 @@ void GPS::setGPSPower(bool on, bool standbyOnly, uint32_t sleepTime)
         pinMode(PIN_GPS_STANDBY, OUTPUT);
         return;
     }
+
 #endif
     if (!on) {
         if (gnssModel == GNSS_MODEL_UBLOX) {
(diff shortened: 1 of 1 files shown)


caveman99 (approved):

src/gps/GPS.cpp
```diff
@@ -502,11 +503,22 @@ void GPS::setGPSPower(bool on, bool standbyOnly, uint32_t sleepTime)
             }
             msglen = gps->makeUBXPacket(0x02, 0x41, 0x08, gps->_message_PMREQ);
             gps->_serial_gps->write(gps->UBXscratch, msglen);
+        } else if (gnssModel == GNSS_MODEL_MTK) {
+            // Enter soft standby mode
+            LOG_DEBUG("L76K entering soft standby mode\n");
+            gps->_serial_gps->write("$PMTK161,0*28\r\n");
```
thebentern, line 509: @neilhao this command seems correct based on the MTK datasheets and I see that SoftRF does the same thing. However, I'm not seeing any power savings with my ATGM336H module. Do you happen to know if they implement the standby command? I can't find any docs specific to the module.

code8buster: Give "$PCAS12,65535*1E\r\n" a try on your personal branch @thebentern @neilhao if this command doesn't work for you. It's listed in the CASIC protocol manual as a standby control msg. Should make the receiver standby for 65535 seconds, if the module supports it.

src/gps/GPS.cpp
```diff
@@ -502,11 +503,22 @@ void GPS::setGPSPower(bool on, bool standbyOnly, uint32_t sleepTime)
...
     } else {
         if (gnssModel == GNSS_MODEL_UBLOX) {
             gps->_serial_gps->write(0xFF);
             clearBuffer(); // This often returns old data, so drop it
+        } else if (gnssModel == GNSS_MODEL_MTK) {
+            LOG_DEBUG("L76K exit soft standby mode\n");
+            // Start vehicle mode (to exit standby mode)
+            gps->_serial_gps->write("$PCAS11,3*1E\r\n");
+            delay(250);
         }
```
caveman99, line 522: if the flush() call resolves the problem it's needed here too

thebentern: I added that for good measure when I was troubleshooting why the power consumption wasn't lowering. I can probably yank it honestly.

PR #2966, opened 2023-11-26
Move to portduino GPIO, adding user button support

jp-bennett:
This moves away from using pigpio for hardware support, getting us partway to being able to do away with a raspbian-specific variant. It also adds support for general GPIO like a user button.

Depends on https://github.com/meshtastic/framework-portduino/pull/8


src/gps/GPS.cpp
```diff
@@ -18,8 +18,8 @@
 #if defined(NRF52840_XXAA) || defined(NRF52833_XXAA) || defined(ARCH_ESP32)
 HardwareSerial *GPS::_serial_gps = &Serial1;
 #elif defined(ARCH_RASPBERRY_PI)
-// need a translation layer to make _serial_gps work with pigpio https://abyz.me.uk/rpi/pigpio/cif.html#serOpen
-HardwareSerial *GPS::_serial_gps = NULL;
+
+HardwareSerial *GPS::_serial_gps = &Serial1;
 #else
```
caveman99, line 23: duplicate line, can merge into one, no need for the elif

jp-bennett: Agreed, fixed.

src/platform/portduino/architecture.h
```diff
@@ -17,3 +17,5 @@
 #ifndef HAS_TELEMETRY
 #define HAS_TELEMETRY 1
 #endif
+#define HAS_WIRE 1
+#define HAS_SCREEN 1
```
mverch67, line 21: Shouln"t it read #ifndef HAS_WIRE ... #endif to be in line with all other definitions in architecture.h and configuration.h

jp-bennett: The last commit is my in-progress work, hacks included, trying to get a screen to light up.

src/ButtonThread.h
```diff
@@ -45,8 +48,14 @@ class ButtonThread : public concurrency::OSThread
     // callback returns the period for the next callback invocation (or 0 if we should no longer be called)
     ButtonThread() : OSThread("Button")
     {
-#ifdef BUTTON_PIN
+#if defined(ARCH_RASPBERRY_PI) || defined(BUTTON_PIN)
```
mverch67, line 51: This line looks superfluous, as both conditions are alread checked below.

PR #2976, opened 2023-12-01
Remove old SX126x and SX128x boosted gain commented-out code

S5NC:
Old, now another method is used. [Old usage](https://github.com/meshtastic/RadioLib/blob/23002909f79f836cee68bcd582031f321da1df05/src/modules/SX126x/SX126x.cpp#L713C5-L713C5). Perhaps the boosted gain mode should be set again during reconfigure, or is this done elsewhere?


caveman99 (approved):
probably good to add it to the re-init too...

PR #2980, opened 2023-12-02
Pico W: Initial Wi-Fi support

GUVWAF:
This enables the Wi-Fi capability for the Pico W. Tested with MNMC, Android app and MQTT. Probably needs further tuning for reconnects etc.

Had to move some calls for the Wi-Fi stack that affect ESP32 as well, but I confirmed it to still be working on a T-Beam.


src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -44,6 +47,61 @@ Syslog syslog(syslogClient);
 
 Periodic *wifiReconnect;
 
+static void onNetworkConnected()
```
GUVWAF, line 50: Only moved this up here such that it is defined before usage on line 155. 
Including it in the header file lead to errors.

src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -109,66 +174,17 @@ void deinitWifi()
     LOG_INFO("WiFi deinit\n");
 
     if (isWifiAvailable()) {
+#ifdef ARCH_ESP32
+        WiFi.disconnect(true, false);
+#else
         WiFi.disconnect(true);
-        WiFi.mode(WIFI_MODE_NULL);
+#endif
+        WiFi.mode(WIFI_OFF);
```
GUVWAF, line 182: Same as `WIFI_MODE_NULL`, but that's only defined for ESP32 (https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiType.h#L32). Same holds for `WIFI_MODE_STA` vs `WIFI_STA`.

thebentern (approved):

Issue #2983, opened 2023-12-03
[Bug]:Wireless Tracker v1.0 or v1.1 the GNSS chip is initialized to only recieve GPS & BDS satellites.


GPSFan:
### Category

Other

### Hardware

Heltec Wireless Tracker

### Firmware Version

2.2.15.31c4693

### Description

The protocol spec provided on the Heltec download site is for an older version of the Firebird series. the correct protocol spec is available at:
https://en.unicorecomm.com/products/detail/34
where 1.4.2.5 CFGSYS: Configure Satellite System
describes the much expanded command for the Firebird II chips of which the UC6580 is one.
The initialization code in GPS.cpp is:

       } else if (gnssModel == GNSS_MODEL_UC6850) {

            // use GPS + GLONASS
            _serial_gps->write("$CFGSYS,h15\r\n");
the h15 will init the chip for GPS L1 & L5 and BDS B1I
IMHO it should bi initialized for GPS L1 & L5, BDS B1I & B2a, GLONASS  L1, GALILEO E1 & E5a, and SBAS.
That would make the command _serial_gps->write("$CFGSYS,h25155\r\n");

### Relevant log output

```Shell
N/A
```

garthvh:
Looks like you have everything here to open a PR

PR #2988, opened 2023-12-04
Fix typo in GNSS_MODEL defination and usages for the UC6580

GPSFan:
Correct the $CFGSYS init string for the UC6580 to init the receiver for: GPS L1 & L5 + BDS B1I & B2a + GLONASS L1 + GALILEO E1 & E5a + SBAS

Fixes Bug #2983

The protocol spec provided on the Heltec download site is for an older version of the Firebird series. the correct protocol spec is available at:
https://en.unicorecomm.com/products/detail/34
where 1.4.2.5 CFGSYS: Configure Satellite System
describes the much expanded command for the Firebird II chips of which the UC6580 is one.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -293,7 +293,7 @@ bool GPS::setup()
             gnssModel = GNSS_MODEL_UNKNOWN;
         }
 #else
-        gnssModel = GNSS_MODEL_UC6850;
+        gnssModel = GNSS_MODEL_UC6580;
 #endif
 
         if (gnssModel == GNSS_MODEL_MTK) {
@@ -311,10 +311,10 @@ bool GPS::setup()
             // Switch to Vehicle Mode, since SoftRF enables Aviation < 2g
             _serial_gps->write("$PCAS11,3*1E\r\n");
             delay(250);
-        } else if (gnssModel == GNSS_MODEL_UC6850) {
+        } else if (gnssModel == GNSS_MODEL_UC6580) {
 
-            // use GPS + GLONASS
-            _serial_gps->write("$CFGSYS,h15\r\n");
+            // use GPS L1 & L5 + BDS B1I & B2a + GLONASS L1 + GALILEO E1 & E5a + SBAS
+            _serial_gps->write("$CFGSYS,h25155\r\n");
             delay(250);
         } else if (gnssModel == GNSS_MODEL_UBLOX) {
             // Configure GNSS system to GPS+SBAS+GLONASS (Module may restart after this command)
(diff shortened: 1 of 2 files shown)


thebentern (approved):

caveman99 (approved):

thebentern:
@GPSFan given your username, maybe you can help with another GPS model issue. I've been unsuccessful with attempts to put an ATGM336H module into standby mode for power savings via UART. I assumed that it would follow the MTK  style commands ala the L76K. That has not proved to be correct in my experience though. If you have any insights on this or know where to dig up a more comprehensive datasheet, that would be swell.

GPSFan:
I'll take a look, I have no experience with the L76 or the ATGM336H modules, so any pointers to the relevant docs would be helpful.

GPSFan:
Looking at the datasheet for the AGTM336H:
https://www.icofchina.com/d/file/xiazai/2016-12-05/b5c57074f4b1fcc62ba8c7868548d18a.pdf
I found a reference to the protocol spec:
CASIC multi mode satellite navigation receiver protocol specification
https://espruino.microcosm.app/api/v1/files/68b597874e0a617692ebebc6a878032f76b34272.pdf
There it describes the $CAS12 command to put the receiver into standby, BUT they also imply that only the -5L low power variant supports that command.
I saw your code in the l76-soft-standby branch and the command you send to the L76 is very different.
I haven't looked for the L76 docs yet so don't have a clue as to how similar or different the hardware is.

code8buster:
https://github.com/meshtastic/firmware/pull/2963
I also suggested that. We discussed it on the Discord, and the CAS12 command had no effect; likely for the reason you stated. The PMTK reference comes from the Quectel L76 protocol document.
https://forums.quectel.com/uploads/short-url/7THMsVAcMkh8pFZM87JHj1b5zXa.pdf

GPSFan:
Thanks for that spec link, it seems the L76 and the  AGTM336H are very different beasts software wise. 
Using the CAS06 command might return some definitive product info on whether the individual module is a low power one or not, then again it might not...

code8buster:
The only ones I've seen available to consumers is the ATGM336H-5N module, and not the 336H-5L

GPSFan:
Yeah, the ones that filter out to us outside of CN are never the ones you want. ;>(
The ATGM336H-5N is a fairly old module, I'd be looking at something M10 based for new designs, THe T-Beam supreme has an M10 option, I have an M10 I got off Amazon that works quite well with a very small antenna. Hopefully, the CN modules will inspire the folks like u-blox to do an L1/L2/L5 that is affordable, well documented and, well, uhh .....

geeksville:

https://meshtastic.discourse.group/t/no-gps-module-t-echo/8824/2

PR #3001, opened 2023-12-08
Portduino - shutdown when reboot is requested (handle reboot external)

wnagele:
Align Portduino reboot logic with Raspberry Pi. Restart can be handled outside (Docker, systemd, etc.).


GUVWAF:
Hmm, if we do this, we'd probably at least want a different exit code for reboot versus shutdown.
https://github.com/mate-dev/mmrelaynode and https://github.com/GUVWAF/Meshtasticator are using Portduino already and at least the latter should be changed if this is applied.
Would be nice if there is a way to do a proper "reboot" from within the program itself.

mverch67:
Actually under Linux a program can restart itself by executing

`execv(argv[0], argv);`

So if you can get hold of the argv string the meshtasticd would automatically restart and not exit as it is now.

jp-bennett:
> Actually under Linux a program can restart itself by executing
> 
> `execv(argv[0], argv);`
> 
> So if you can get hold of the argv string the meshtasticd would automatically restart and not exit as it does now.

We should just add a reboot() function to Portduino to handle this.

jp-bennett:
> We should just add a reboot() function to Portduino to handle this.

Made a first attempt at this, and the second boot fails, as it cannot get the TCP port. This approach doesn't let go of ports or hardware, so would be a bit more involved to do cleanly. I'm currently inclined to use exit codes to indicate why the binary is exiting, and let systemd handle restarts.

mverch67:
I think this can be closed due to https://github.com/meshtastic/framework-portduino/pull/10

thebentern:
Closing since @jp-bennett addressed this in the framework

PR #3004, opened 2023-12-08
Reduce Serial Traffic on Heltec Wireless Trackers GNSS port

GPSFan:
While investigating a "Bogus course.value" message in the log I found that there was a lot on unnecessary message traffic from the UC6580 chip on the Heltec Wireless Tracker v1.1.
Most of this traffic was GSV sentences which provide azimuth, elevation, and signal strength for the currently visible satellites. While interesting, this information is not needed by Meshtastic (yet). In addition there are also some __TXT messages that are for use by Unicore, and should not be of interest to Meshtastic.
This PR turns these messages off and adds a bit of documentation.

### DIFF
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -212,6 +212,9 @@ jobs:
           repository: ${{github.event.pull_request.head.repo.full_name}}
 
   gather-artifacts:
+    permissions:
+      contents: write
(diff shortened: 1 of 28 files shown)


jp-bennett (approved):
Looks great!

code8buster (approved):

thebentern (approved):

PR #3017, opened 2023-12-15
Add fallthrough option to avoid a GPS stuck off.

jp-bennett:


geeksville:

https://meshtastic.discourse.group/t/no-gps-module-t-echo/8824/2

PR #3033, opened 2023-12-24
Portduino reboot

wnagele:
Follow-up for #3001. Portduino has the necessary implementation now but we didn't execute it outside of the Raspi arch.

### DIFF
--- a/src/shutdown.h
+++ b/src/shutdown.h
@@ -19,15 +19,18 @@ void powerCommandsCheck()
         NVIC_SystemReset();
 #elif defined(ARCH_RP2040)
         rp2040.reboot();
-#elif defined(ARCH_RASPBERRY_PI) || defined(ARCH_PORTDUINO)
+#elif defined(ARCH_RASPBERRY_PI)
         deInitApiServer();
-#if defined(ARCH_RASPBERRY_PI)
         if (aLinuxInputImpl)
             aLinuxInputImpl->deInit();
-#endif
         SPI.end();
         Wire.end();
         Serial1.end();
+        reboot();
+#elif defined(ARCH_PORTDUINO)
+        deInitApiServer();
+        SPI.end();
+        Wire.end();
         reboot();
 #else
         rebootAtMsec = -1;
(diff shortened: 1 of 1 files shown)


GUVWAF:
This fails for me after the "reboot" with: 

> WiFiClient WiFiServer::available(byte*): Assertion `psock' failed. 

I'm trying to figure out how to deal with this.

GUVWAF:
Seems like the socket is always giving me a file descriptor with value 0 after rebooting, but actually that works fine. Probably the assert should be changed to `psock >= 0`.

src/shutdown.h
```diff
@@ -19,10 +19,12 @@ void powerCommandsCheck()
...
-#elif defined(ARCH_RASPBERRY_PI)
+#elif defined(ARCH_RASPBERRY_PI) || defined(ARCH_PORTDUINO)
         deInitApiServer();
+#if defined(ARCH_RASPBERRY_PI)
         if (aLinuxInputImpl)
             aLinuxInputImpl->deInit();
+#endif
         SPI.end();
         Wire.end();
         Serial1.end();
```
GUVWAF, line 30: After digging further into my issue, I found that it's caused by the call to `Serial1.end()`. Since this is only used for GPS on `ARCH_RASPBERRY_PI` and not for `ARCH_PORTDUINO`, its file descriptor is still 0, which is normally used for `stdin`. Thus when rebooting, `stdin` is closed and after the reboot, Linux sees 0 as free and assigns the socket to it. 

@wnagele can you therefore move `Serial1.end()` into the `#if defined(ARCH_RASPBERRY_PI)`? This fixes my issue.

GUVWAF (changes requested):

src/shutdown.h
wnagele: I've given the `ARCH_PORTDUINO` its own block now. Otherwise the macros become unwieldy and hard to read.

wnagele:
@jp-bennett this should be good to merge.

PR #3036, opened 2023-12-24
Allow override of HWID using environment variable

wnagele:
When testing sometimes it is useful to be able to set this explicitly.


GUVWAF (approved):
Seems there was just a hiccup with terrascan. I re-ran the failed jobs and it's fine now.

wnagele:
@jp-bennett this should be good to merge.

PR #3069, opened 2024-01-08
Heltec paper

thebentern:
Created from @ixt's fork

### DIFF
--- a/src/graphics/EInkDisplay2.cpp
+++ b/src/graphics/EInkDisplay2.cpp
@@ -234,8 +234,10 @@ bool EInkDisplay::connect()
         auto lowLevel = new TECHO_DISPLAY_MODEL(PIN_EINK_CS, PIN_EINK_DC, -1, PIN_EINK_BUSY);
         adafruitDisplay = new GxEPD2_BW<TECHO_DISPLAY_MODEL, TECHO_DISPLAY_MODEL::HEIGHT>(*lowLevel);
         hspi = new SPIClass(HSPI);
+        // Bogus MISO because espressif complains
         hspi->begin(PIN_EINK_SCLK, 46, PIN_EINK_MOSI, PIN_EINK_CS); // SCLK, MISO, MOSI, SS
-        adafruitDisplay->init(115200, true, 10, false, SPI, SPISettings(4000000, MSBFIRST, SPI_MODE0));
+        delay(10);
+        adafruitDisplay->init(115200, true, 10, false, SPI, SPISettings(6000000, MSBFIRST, SPI_MODE0));
         adafruitDisplay->setRotation(3);
         adafruitDisplay->setPartialWindow(0, 0, displayWidth, displayHeight);
     }
--- a/src/graphics/EInkDisplay2.cpp
+++ b/src/graphics/EInkDisplay2.cpp
@@ -47,7 +47,7 @@ SPIClass *hspi = NULL;
 
 #elif defined(HELTEC_WIRELESS_PAPER)
 // #define TECHO_DISPLAY_MODEL GxEPD2_213_T5D
-#define TECHO_DISPLAY_MODEL GxEPD2_213_BN
+#define TECHO_DISPLAY_MODEL GxEPD2_213_FC1
 #endif
 
 GxEPD2_BW<TECHO_DISPLAY_MODEL, TECHO_DISPLAY_MODEL::HEIGHT> *adafruitDisplay;
@@ -119,7 +119,6 @@ bool EInkDisplay::forceDisplay(uint32_t msecLimit)
         // tft.drawBitmap(0, 0, buffer, 128, 64, TFT_YELLOW, TFT_BLACK);
         for (uint32_t y = 0; y < displayHeight; y++) {
             for (uint32_t x = 0; x < displayWidth; x++) {
-
                 // get src pixel in the page based ordering the OLED lib uses FIXME, super inefficient
                 auto b = buffer[x + (y / 8) * displayWidth];
                 auto isset = b & (1 << (y & 7));
@@ -148,7 +147,8 @@ bool EInkDisplay::forceDisplay(uint32_t msecLimit)
 
 #elif defined(PCA10059) || defined(M5_COREINK)
         adafruitDisplay->nextPage();
-
+#elif defined(HELTEC_WIRELESS_PAPER)
+        adafruitDisplay->nextPage();
 #elif defined(PRIVATE_HW) || defined(my)
         adafruitDisplay->nextPage();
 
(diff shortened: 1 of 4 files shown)


variants/heltec_wireless_paper/variant.h
markbirss, line 1: This seems not to be needed ? The MISO and MOSI are swapped also

 * SPI interfaces
 */
#define SPI_INTERFACES_COUNT 2

#define PIN_SPI_MISO 10 // MISO      P0.17
#define PIN_SPI_MOSI 11 // MOSI      P0.15
#define PIN_SPI_SCK 9   // SCK       P0.13

#define LORA_SCK 9
#define LORA_MISO 11
#define LORA_MOSI 10
#define LORA_CS 8

src/graphics/EInkDisplay2.cpp
markbirss, line 1: #elif defined(HELTEC_WIRELESS_PAPER)
        adafruitDisplay->nextPage();

this should be same as t-echo ?
adafruitDisplay->display(false);

as they dont support partial updates

also is this e-ink a b/w/r or b/w ?

variants/heltec_wireless_paper/variant.h
ixt: Yeah SPI_INTERFACES_COUNT doesnt work with that core anyway, I must of left that in! Apologies! 

I might revisit the fork and bring it up to date if its an issue that its old? I don't see it as being much work as most of the changes I saw between the two we're pretty minor for the majority of pathways I looked at

src/graphics/EInkDisplay2.cpp
ixt: It's just B/W, It's unclear if it supports partial updates or not (the driver basically doesn't). We can switch it over to the same as t-echo for now for clarity but if a datasheet/more info does surface I'll update all relevant.

ixt:
ribboncablewirelesspaper
here is an image of my flex cable, HINK-E0213A162-FPC-AO, DEN 2242

variants/heltec_wireless_paper/platformio.ini
```diff
@@ -5,4 +5,7 @@ build_flags =
   ${esp32s3_base.build_flags} -D HELTEC_WIRELESS_PAPER -I variants/heltec_wireless_paper
 lib_deps =
   ${esp32s3_base.lib_deps}
-  zinggjm/GxEPD2@^1.5.2
+  https://github.com/ixt/GxEPD2#39f325b677713eb04dfcc83b8e402e77523fb8bf
```
ixt, line 8: This should probably be changed to latest commit on meshtastic's fork now, although it's no issue for it to track mine.

markbirss:
@ixt is there no print on the other side (outside) of the e-ink flex cable ?



this was just before i tried to move up those dots long ago (rak14000 e-ink)

thebentern:
> @ixt is there no print on the other side (outside) of the e-ink flex cable ?

Mine has FPC-7528B on the outside

markbirss:
@thebentern and the changes work ok on your version ?

ixt:
@markbirss nope, there isnt, that photo is of the screen of the v1 board (red tab) -> https://docs.heltec.org/en/node/esp32/hardware_update_log.html

markbirss:
@ixt 

Ok, it  leads to sample code of a b/w/r e-ink.....

QYEG0213RWS800_BWR
[Sample code V1.1](https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/tree/master/esp32/libraries/Heltec-Example/examples/E-INK/QYEG0213RWS800_BWR) - https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/tree/master/esp32/libraries/Heltec-Example/examples/E-INK/QYEG0213RWS800_BWR

ixt:
@markbirss that isnt the sample code for v1.1, its linked wrong. this is for v1.1 https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/master/esp32/libraries/Heltec-Example/examples/E-INK/HT_lCMEN2R13EFC1/HT_lCMEN2R13EFC1.ino

For the v1 I believe the QYE... code is correct, BUT its also not BWR... (To my understanding). The sample code, documentation in general is all over the place for wireless paper :/

ixt:
@markbirss if you see on this page there is a tip that is the most correct about which example is which https://docs.heltec.org/en/node/esp32/wireless_paper/via_program_refresh.html#summary

thebentern:
> @thebentern and the changes work ok on your version ?

Just tested. Sadly it does not. I suspect mine is the original version and they made changes similar to the Wireless Tracker v1.1 vs 1.0

ixt:
@thebentern I'm guessing that the v1 may work with one of the other display's drivers as it looks much more like what I've been seeing in other displays here: https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/master/esp32/libraries/DISPLAY/src/HT_QYEG0213RWS800_BWR.h 

Pulling Vext low might still be needed tho since if the driver was all that was needed it would have been fixed by now

StevenCellist:
https://forum.arduino.cc/t/new-e-ink-display-not-in-gxepd2/1196494 here is a discussion about this new panel - Heltec has not responded in any way other than pointing to their stupid incomplete HT_ICMEN driver.

markbirss:
> https://forum.arduino.cc/t/new-e-ink-display-not-in-gxepd2/1196494 here is a discussion about this new panel - Heltec has not responded in any way other than pointing to their stupid incomplete HT_ICMEN driver.

Thank you for pointing us to the already existing discussion in understanding some of the issues

geoffwatts:
I have tried this (on my v1.1 stickered device) and it's working great.

thebentern:
I'm going to go ahead and merge this and then try to get my v1.0 working with the a different driver I guess

ixt:
@thebentern I'm guessing if you clone the driver I wrote, swap out the 0x13 commands for 0x24s, update full and part's 0x12s for 
```
		sendCommand(0x22); 
		sendData(0xF7);
		sendCommand(0x20); 
```
it might just work??

https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/master/esp32/libraries/DISPLAY/src/HT_QYEG0213RWS800_BWR.h is heltechs driver for the v1 display, you can compare that with the v1.1 https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/master/esp32/libraries/DISPLAY/src/HT_lCMEN2R13EFC1.h and the one I smashed together. It shouldn't be too difficult. 

Other than the driver the .ino examples look pretty much the same. The GPIO's look different but I'd be surprised if they bothered, and just copied whatever vendor's example they got....

I found eerily similar examples on gooddisplay's website when i was looking it up. for prosperity here is a link also to the correct .ino example for the v1 display, https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/master/esp32/libraries/Heltec-Example/examples/E-INK/QYEG0213RWS800_BWR/QYEG0213RWS800_BWR.ino

good luck! I messaged a few sellers to see if i can get a hold of a v1 device but no replies yet. Glad to see this merged though!

ixt:
@markbirss I think I saw this in passing, but it didnt really help much for the v1.1s EPD, the controller isn't entirely UC8151D like, the resolution setting stuff (0x61) always crashed the display for me. And the example code included commands post-0x90 that weren't documented in any screen I looked at. I think either there are parts that are left in from an OLED driver or we are looking at a hypothetical UC815E/UC815 Variant in the v1.1 case. But almost certainly it's built on the UC8151D set. 

Though, having said all that, it is hopeful that these displays can do partial update with how its explicitly labeled there for both versions

PR #3073, opened 2024-01-09
fix: `ipv4_config` byte order already little endian

andrekir:
removes `bigToLittleEndian()` since `ipv4_config` byte order is already little endian


andrekir:
confirmed this is necessary to get `network.ipv4_config` working with static IPs on RAK 4631 + 13800 Ethernet combo.

ESP32 not affected and working as well after this change.

thebentern (approved):

PR #3081, opened 2024-01-11
[BOARD] Adds Waveshare ESP32-S3-PICO

markbirss:
Waveshare ESP32-S3-Pico (16 MB FLASH, 2 MB PSRAM)
https://www.waveshare.com/esp32-s3-pico.htm

Requires

SX1262 LoRa Node Module for Raspberry Pi Pico, LoRaWAN, Choice Of Frequency Band
https://www.waveshare.com/pico-lora-sx1262-868m.htm

SX1262 LoRa Node Module for Raspberry Pi Pico, LoRaWAN, Choice Of Frequency Band
https://www.waveshare.com/pico-lora-sx1262-868m.htm

2.9inch E-Paper E-Ink Display Module for Raspberry Pi Pico, 296×128, Black / White, SPI
https://www.waveshare.com/pico-epaper-2.9.htm
Wired by external cable since GPIO conflict with MISO pin of LoRa HAT

https://www.waveshare.com/wiki/Pico-UPS-A
https://www.waveshare.com/wiki/Pico-UPS-B


markbirss:
I think it should be good now, i fixed the correct board/target name

geeksville:

https://meshtastic.discourse.group/t/boards-new-esp32-s3-pico-2-9-e-ink-waveshare-pico-lora-sx1262-hat/9256/1

src/graphics/EInkDisplay2.cpp
```diff
@@ -93,6 +93,12 @@ EInkDisplay::EInkDisplay(uint8_t address, int sda, int scl, OLEDDISPLAY_GEOMETRY
     setGeometry(GEOMETRY_RAWMODE, 296, 128);
     LOG_DEBUG("GEOMETRY_RAWMODE, 296, 128\n");
 
+#elif defined(ESP32_S3_PICO)
```
mverch67, line 96: The E-Ink display could also be used with the pico, so I suggest not to create a dependency to S3 pico here. Just use a new define for the E-ink display and use it here. This way it is more generic. Also it allows to use S3 pico with other displays.

src/graphics/EInkDisplay2.cpp
```diff
@@ -256,7 +264,7 @@ bool EInkDisplay::connect()
     adafruitDisplay->init(115200, true, 40, false, SPI, SPISettings(4000000, MSBFIRST, SPI_MODE0));
     adafruitDisplay->setRotation(0);
     adafruitDisplay->setPartialWindow(0, 0, EPD_WIDTH, EPD_HEIGHT);
-#elif defined(my)
+#elif defined(my) || defined(ESP32_S3_PICO)
```
mverch67, line 267: same as my previous comment

src/graphics/EInkDisplay2.cpp
markbirss: e-ink support has not been added to pico at all yet

src/graphics/EInkDisplay2.cpp
markbirss: only i2c displays at the moment

src/graphics/EInkDisplay2.cpp
mverch67: The S3 pico and pico share the same pinout. So with your extensions (and some platform.ini enhancements) the same E-ink display would also work with the pico I guess.

mverch67: My main point here is not to create artificial dependencies in the code. Keep the display separated from the board.

boards/esp32-s3-pico.json
```diff
@@ -0,0 +1,40 @@
...
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_ESP32S3_DEV",
+      "-DARDUINO_USB_MODE=1",
+      "-DARDUINO_RUNNING_CORE=1",
+      "-DARDUINO_EVENT_RUNNING_CORE=1"
+    ],
+    "f_cpu": "240000000L",
+    "f_flash": "80000000L",
+    "flash_mode": "qio",
```
mverch67, line 16: Does the PSRAM work with this setting? How much PSRAM is reported in the debug log for your board?

boards/esp32-s3-pico.json
```diff
@@ -0,0 +1,40 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "esp32s3_out.ld",
+      "partitions": "default_16MB.csv"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DARDUINO_ESP32S3_DEV",
```
mverch67, line 9: add -DBOARD_HAS_PSRAM

boards/esp32-s3-pico.json
```diff
@@ -0,0 +1,40 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "esp32s3_out.ld",
+      "partitions": "default_16MB.csv"
```
mverch67, line 5: try to add "memory_type": "qio_opi" if no PSRAM is reported in debug log

src/platform/esp32/architecture.h
```diff
@@ -119,8 +119,8 @@
 #define HW_VENDOR meshtastic_HardwareModel_BETAFPV_900_NANO_TX
 #elif defined(PICOMPUTER_S3)
 #define HW_VENDOR meshtastic_HardwareModel_PICOMPUTER_S3
-#elif defined(HELTEC_HT62)
-#define HW_VENDOR meshtastic_HardwareModel_HELTEC_HT62
+#elif defined(ESP32_S3_PICO)
```
thebentern, line 122: Looks like HELTEC_HT62 accidentally got removed here

PR #3084, opened 2024-01-11
INA3221 library branch to support negative values

KodinLanewave:
Original INA3221 library does not handle negative values properly due to mishandling of signed bits in sensor reading and processing.  I have branched the library, changed the code, and tested with several deployed nodes to confirm functionality.

(I first tried to reach out to the library maintainer and filed a bug report over a month ago, then got impatient and fixed it myself in a new branch)


platformio.ini
```diff
@@ -115,7 +115,7 @@ lib_deps =
   https://github.com/boschsensortec/Bosch-BSEC2-Library#v1.5.2400
   boschsensortec/BME68x Sensor Library@^1.1.40407
   adafruit/Adafruit MCP9808 Library@^2.0.0
-  https://github.com/Tinyu-Zhao/INA3221@^0.0.1
+  https://github.com/KodinLanewave/INA3221#main
```
thebentern, line 118: Can we do a version tag instead? Pegging it to a branch makes for non-idempotent builds potentially.
We may actually want to fork this one into the meshtastic org at some point.

KodinLanewave: @thebentern sure, I'm just way less familiar with tags and couldn't get my build environment to utilize them when I set that up.  I'll see what I can do though.

KodinLanewave (commented):
@thebentern Version tag set up and tested

PR #3088, opened 2024-01-12
Prioritize lowest channel for each remote node in NodeDB

### LINKED ISSUE #3086: [Feature Request]: NodeDB "Node Channel" should prefer channels based on user preference
### Platform

NRF52, ESP32

### Description

Hello, I am creating this issue to discuss something I may make a PR for, but want to get alignment from the maintainers on first.

### Problem
In the following scenario:
- Three node mesh: NodeA, NodeB, NodeD
- NodeA and NodeB have a private channel with default LongFast as secondary, NodeD has a default channel configuration
- NodeA and NodeB use the Direct Message feature often to communicate, and expect privacy
- NodeD has a setup that allows them to log all non-broadcast messages on LongFast, such as by using MQTT

In the above scenario, the following negative situation may occur:
1. NodeA receives a NodeInfo packet from NodeB on the LongFast channel
2. NodeA updates it's NodeDB entry for NodeB to note that NodeB is now on the LongFast channel
3. NodeA sends a direct message to NodeB, which is now sent over LongFast
4. NodeD intercepts and reads this message

If any assumptions made above are incorrect, and this situation would in fact not occur, please feel free to close this issue.

### Proposed Solution

I suggest implementing a priority of channels, where the channel closest to 0 will be chosen for each remote node and kept consistent.

### Proposed simple implementation
Add a channel number minimization check to:
https://github.com/meshtastic/firmware/blob/master/src/mesh/NodeDB.cpp#L851 
```c++
if (mp.decoded.portnum == meshtastic_PortNum_NODEINFO_APP && mp.channel < info->channel) {
    info->channel = mp.channel;
}
``` 

### Shortcomings with above implementation
a. ~~Channel 0 is not always the primary channel, this simple implementation could thus prioritize communicating on a non-user intended channel~~
b. Remote nodes that remove the preferred channel will not be communicable until removed and re-entered in the NodeDB
c. The user changing their channels from a default implementation to one with a PSK locked primary channel will continue attempting to communicate on the new locked primary channel until NodeDB is reset.

What do you think?

Evanito:
- Simple implementation that would close #3086 
- Every time we update NodeDB with info from a remote node's packet, we only update the channel if it is "closer" to our primary channel 0. This results in not downgrading remote nodes away from the primary channel in certain situations.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -844,7 +844,7 @@ void NodeDB::updateFrom(const meshtastic_MeshPacket &mp)
         if (mp.rx_snr)
             info->snr = mp.rx_snr; // keep the most recent SNR we received for this node.
 
-        if (mp.decoded.portnum == meshtastic_PortNum_NODEINFO_APP) {
+        if (mp.decoded.portnum == meshtastic_PortNum_NODEINFO_APP && mp.channel < info->channel) {
             info->channel = mp.channel;
         }
     }


GUVWAF:
While I agree the situation you describe is not ideal, it makes multi-channel handling even more complicated than it already is (see #2466). 

This PR will not work, because when a new node on a channel different than 0 comes in, the correct channel will never be set since `info.channel` is initialized at 0. So for new nodes, it should always allow it.

Evanito:
Very true, didn't account for the channel defaulting to 0

Issue #3075, opened 2024-01-10
[Bug]: E-paper partial updates on T-Echo and Wireless Paper


ixt:
### Category

Hardware Compatibility

### Hardware

T-Echo

### Firmware Version

2.2.16

### Description

Both of these epaper displays look to support partial updates, I would have opened this issue at our [GxEPD2 repo](https://github.com/meshtastic/GxEPD2) but cant. 
Looking at this implementation of T-Echo's display driver https://github.com/cfr34k/t-echo-lora-aprs/blob/main/src/epaper.c#L377 it's not impossible to drive the display in this way. 
This may also help mitigate #2356 .
The controller for T-Echo's display is SSD1681, its commands can be found here: https://v4.cecdn.yun300.cn/100001_1909185148/SSD1681.pdf 

Wireless paper v1.1's display has a UC8151D-like controller, either there is some misunderstanding on my part of how partial updates are prepared for these controllers or we just don't have the right commands yet. Regardless, it's likely that this display also supports partial updates.
https://v4.cecdn.yun300.cn/100001_1909185148/UC8151D.pdf

markbirss:
@ixt 

These are the specific  changes that were made initially to support a e-ink on 2nd spi bus on the t-echo as GxEPD2 don't/wont support any back light or touch displays, you would need different e-paper driver boards like the  DESPI-C03 anyway

T-Echo
https://github.com/ZinggJM/GxEPD2/compare/master...geeksville:GxEPD2:master

DESPI-C03
https://www.good-display.com/companyfile/32.html

The other fast display support i added including the RAK14000 eink i moved to using the recent/current GxEPD2 library as i was able to modify/specify SPI settings 

I suggested recently on the discord group that it might help to take discussion with LilyGo regarding possible hardware changes/enhancements to possibly fix the highlighted issues already known

Im aware the actual e-ink display in the t-echo can support partial updates as i used the same display with my first diy nrf52 board

Here the other PR's for reference
https://github.com/meshtastic/firmware/pull/1240
https://github.com/meshtastic/firmware/commit/f474953b513ed8ec3b4cef5e9af76186c746359b
https://github.com/meshtastic/firmware/commit/8a1bf8cd86ab58e9ace8cd0634f6624077453c81

I agree with the possibility for partial updates with the Wireless paper v1.1 also.

The other issue to include in discussion  is the need for a full update to be includes after so many partial updates - e-ink displays are not meant to be partially updates for ever

[...]

PR #3090, opened 2024-01-13
Initial Partial Updates on t-echo

ixt:
#3075 

Partial Updates for t-echo, The boot screen is very faint. And we don't (to my knowledge) do a full refresh at some point to help with ghosting but this should be good for now. I can't believe it was basically working. I'll find some time to write something for full updates (maybe on manual loop of screens and after n-partial updates)


markbirss:
@ixt it normal for some e-inks to have faint boot screen

yes, a counter but it must goto the tros tasks i think

src/graphics/EInkDisplay2.cpp
```diff
@@ -55,7 +55,7 @@ GxEPD2_BW<TECHO_DISPLAY_MODEL, TECHO_DISPLAY_MODEL::HEIGHT> *adafruitDisplay;
 EInkDisplay::EInkDisplay(uint8_t address, int sda, int scl, OLEDDISPLAY_GEOMETRY geometry, HW_I2C i2cBus)
 {
 #if defined(TTGO_T_ECHO)
-    setGeometry(GEOMETRY_RAWMODE, TECHO_DISPLAY_MODEL::WIDTH, TECHO_DISPLAY_MODEL::HEIGHT);
```
thebentern, line 58: Is this change necessary?

markbirss: I was looking at all the e-ink support as raised by the previous ESP32-S3-PICO PR. I wanted to look at supply these geometry values in the platformio.ini variant files which would reduce the checks so it ok for now

ixt: This was the change actually that made it all work for me (aside from the explicitly partial update parts). The why is an open question to me though

PR #3094, opened 2024-01-14
Update channel of node in `updateUser` and write to flash if needed

### LINKED ISSUE #2466: [Bug]: New Channel on NodeInfo is usually not correct.

GUVWAF:
Potentially fixes #2466, see my comment there.


andrekir (approved):

thebentern (approved):
This solution makes a ton of sense.

PR #3111, opened 2024-01-18
Default mqtt root to msh/region from unset

thebentern:
This means that going forward, the root mqtt topic would be partitioned by lora region when setting up a blank device. 
image

Users can always set the root topic back to "msh" if desired, but the impetus of this change is to start decreasing growth and congestion of the main public mqtt traffic.


caveman99 (approved):

geeksville:

https://meshtastic.discourse.group/t/where-is-the-world-map/6882/24

PR #3117, opened 2024-01-20
Mark packets received via MQTT and add option to ignore them

GUVWAF:
This will use one bit in the unencrypted Meshtastic header flags (out of 4 currently unused) to mark that a packet passed via MQTT somewhere on the path towards the node. Using the LoRa config `ignore_mqtt`, one can then ignore packets that came via MQTT. The `via_mqtt` tag in the MeshPacket can also be used to show that a Traceroute did not go purely via LoRa.

Tested with three devices, 2 of which where connected to MQTT, of which one had LoRa Tx disabled. Packets received via MQTT were marked accordingly and ignored correctly. Note that the tag is only set when *receiving* a downlink on MQTT and **not** when an uplink is sent via MQTT, as it would mark the packet in memory and then also send it out marked when sending via LoRa.

Needs protobuf PR https://github.com/meshtastic/protobufs/pull/423.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -7,6 +7,7 @@
 #include "mesh/Router.h"
 #include "mesh/generated/meshtastic/mqtt.pb.h"
 #include "mesh/generated/meshtastic/telemetry.pb.h"
+#include "modules/RoutingModule.h"
 #if defined(ARCH_ESP32)
 #include "../mesh/generated/meshtastic/paxcount.pb.h"
 #endif
@@ -464,6 +465,9 @@ void MQTT::publishQueuedMessages()
 
 void MQTT::onSend(const meshtastic_MeshPacket &mp, ChannelIndex chIndex)
 {
+    if (mp.via_mqtt)
+        return; // Don't send messages that came from MQTT back into MQTT
+
     auto &ch = channels.getByIndex(chIndex);
 
     if (&mp.decoded && strcmp(moduleConfig.mqtt.address, default_mqtt_address) == 0 &&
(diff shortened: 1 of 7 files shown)


src/mqtt/MQTT.cpp
```diff
@@ -142,6 +142,7 @@ void MQTT::onReceive(char *topic, byte *payload, size_t length)
                 if (strcmp(e.channel_id, channels.getGlobalId(ch.index)) == 0 && e.packet && ch.settings.downlink_enabled) {
                     LOG_INFO("Received MQTT topic %s, len=%u\n", topic, length);
                     meshtastic_MeshPacket *p = packetPool.allocCopy(*e.packet);
+                    p->via_mqtt = true; // Mark that the packet was received via MQTT
```
andrekir, line 145: do we also need to mark `via_mqtt` in the _if_ block above (for `mqtt.json_enabled`)?

GUVWAF: I thought this could be considered as a client <-> device connection. However, when I now look at it again, apparently it checks whether the "sender" is **not** equal to its own owner ID:
https://github.com/meshtastic/firmware/blob/f0b8430c6e3ffae33daca5375c0e037c5c3f5d8f/src/mqtt/MQTT.cpp#L99

Isn't this feature to allow people to inject a packet on their mesh via JSON? Now this looks very odd to me as it seems you could spoof a text message as coming from another node if you send this on an MQTT server.

andrekir:
looks good!

in [MQTT::onSend](https://github.com/meshtastic/firmware/blob/master/src/mqtt/MQTT.cpp#L464) can we add a check for `via_mqtt` and return when it is tagged? this will prevent a feedback loop (causing duplicates) for packets that we now know came from MQTT.

GUVWAF:
> in [MQTT::onSend](https://github.com/meshtastic/firmware/blob/master/src/mqtt/MQTT.cpp#L464) can we add a check for `via_mqtt` and return when it is tagged? this will prevent a feedback loop (causing duplicates) for packets that we now know came from MQTT.

Yes, we can do that, but we would need to create an implicit ACK in another way then, otherwise it tries to retransmit every time. Shall we just create an implicit ACK whenever it is connected to the MQTT broker?

andrekir:
> Yes, we can do that, but we would need to create an implicit ACK in another way then, otherwise it tries to retransmit every time. Shall we just create an implicit ACK whenever it is connected to the MQTT broker?

yes, I believe preventing duplicates would be the priority. we're at over 85% duplicates in the public MQTT LongFast, if this is the main source it will be a big improvement.

src/mqtt/MQTT.cpp
andrekir: very odd indeed. I wonder if this `sender` spoofing is also a source of duplicates. but let's leave this alone for now.

GUVWAF:
Implemented in [5363ad8](https://github.com/meshtastic/firmware/pull/3117/commits/5363ad8bbaea62ab50f03c76ee4fc4f0f39fdb64).

thebentern:
2.2.19 protos yanked and reprovisioned with the new config property

thebentern (approved):

PR #3122, opened 2024-01-23
Don't ever delete own node from DB

jp-bennett:


thebentern (approved):

thebentern:
Now with less segfaults!

mverch67:
> Now with less segfaults!

The good thing about segfaults is that it clearly indicates an issue in the software and it is easy to pinpoint on a linux system and also to fix. Much better than if issues are silently caught and ignored in the software without any notice.

src/modules/PositionModule.cpp
```diff
@@ -204,6 +204,8 @@ int32_t PositionModule::runOnce()
     }
 
     meshtastic_NodeInfoLite *node = nodeDB.getMeshNode(nodeDB.getNodeNum());
+    if (node == nullptr)
+        return RUNONCE_INTERVAL;
```
mverch67, line 208: Would be way better to also log an error here (instead of silently ignoring the fact that this get function betrayed us).
==> LOG_ERROR("getMeshNode() returned nullptr!");

Imagine we'd have dozens of this kind of code lines in the software where severe problems are simply ignored, how you're gonna pinpoint any bug in the system?

thebentern: Good point

Issue #3080, opened 2024-01-11
[Board]: Is it possible to add the Lilygo T3 LoRa ESP 32 TCXO?


Burny08:
### SOC

ESP32

### Lora IC

LILYGO 868/915Mhz Modell: LORA32 TCXO

### Product Link

https://de.aliexpress.com/item/32872078587.html?spm=a2g0o.order_list.order_list_main.9.36e05c5fprjabY&gatewayAdapt=glo2deu

### Description

I have been informed by the manufacturer that the TCXO variant is not supported. 

The Modell is: T3 V1.6.1 with SX1276

Unfortunately I am not a programmer myself, but he told me that this board can probably be added. 
Here is the message from the manufacturer: 

> Hello, the meshtastic firmware is not adapted to the TXCO version yet.
> You can add the below sentences in the source code to make it work.
> https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/examples/RadioLibExamples/SX1276/SX1276_Receive_Interrupt/boards.h
> 
> Or to test hardware function, you can upload the below firmware to  check :
> https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/tree/master/firmware/hardware_test/T3_V1.6.1_TXCO
> https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/tree/master/firmware#to-write-firmware-please-check-the-corresponding-link-below
![Se2e11434816643559b22b149995f467cP png](https://github.com/meshtastic/firmware/assets/100558141/10c39e33-c266-42f5-ba7c-1cb6a902881c)

marclura:
Hi, I managed to make the LilyGo T3 V1.6.1 TXCO version work, by modifying the existing _tlora_v2_1_16_  variant (I couldn't figure out how to create a new variant and make it work as a separate board).

Burny08:
> Hi, I managed to make the LilyGo T3 V1.6.1 TXCO version work, by modifying the existing _tlora_v2_1_16_ variant (I couldn't figure out how to create a new variant and make it work as a separate board).

That sounds great! Can you tell me how I can get this version onto my board?
I can use the Espressif Flasher, but then my knowledge slowly stops :D

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/t3-v1-6-tcxo-variant/9240/2

Burny08:
It works!
[...]

PR #3124, opened 2024-01-23
feat(variants): Add support for TXCO on TLORA_V2_1_6 devices

Scarg:
## Description

Add support for LILYGO®LoRa V2.1_1.6 devices that come with TCXO following LILYGO's examples (see [here](https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/examples/RadioLibExamples/SX1276/SX1276_Receive_Interrupt/boards.h#L338)).

Before these changes, the device didn't work properly and instead reported **`Critical fault #3`**.

## Notes

Creating a new variant seemed like the correct solution as it was the easiest way to handle the removal of pin 33 from LORA_DIO1.
Instead of creating an entirely new `variant.h`, something that seemed overkill as they are otherwise the same device, I simply created a new platform and added an extra build flag while still basing the environment on the default V2.1_1.6 variant definition.

I would have preferred to not touch LORA_DIO1, especially since I'm not sure what should be the correct handling in `sleep.cpp` , but it's used by `RF95_DIO1` (although the RF95 setup fails _gracefully_) so you can't leave it undefined.

---
Solves #3080


code8buster:
Nice and minimal.
As a point of curiosity, what effect does setting LORA_DIO1 = RADIOLIB_NC have on your variant? Does the rf95 setup still fail gracefully, or does it succeed?
RadioLib I'm wondering if that failure is RadioLib trying to attach an interrupt to that pin.

Scarg:
Hello @code8buster!

> Nice and minimal. As a point of curiosity, what effect does setting LORA_DIO1 = RADIOLIB_NC have on your variant? Does the rf95 setup still fail gracefully, or does it succeed? RadioLib I'm wondering if that failure is RadioLib trying to attach an interrupt to that pin.

So, apparently, I got a severe case of brainfart and didn't commit the change on tlora_v2_1_16/variant.h.
My idea was to do the following:
```cpp
#ifdef TLORA_V2_1_6_TXCO
#define LORA_DIO1 32
#else
#define LORA_DIO1 33 // https://www.thethingsnetwork.org/forum/t/big-esp32-sx127x-topic-part-3/18436
#endif
````

But this is what I get in the logs:
 - if I use the stock firmware (without the TCXO support, so the one that gets Critical Fault #3)
 ```
INFO  | ??:??:?? 1 RF95 init result -16
INFO  | ??:??:?? 1 Frequency set to 906.875000
INFO  | ??:??:?? 1 Bandwidth set to 250.000000
INFO  | ??:??:?? 1 Power output set to 20
WARN  | ??:??:?? 1 Failed to find RF95 radio
```
- if I use my firmware with `#define LORA_DIO1 32` or `#define LORA_DIO1 RADIOLIB_NC` (same result)
```
INFO  | ??:??:?? 0 RF95 init result 0
INFO  | ??:??:?? 0 Frequency set to 906.875000
INFO  | ??:??:?? 0 Bandwidth set to 250.000000
INFO  | ??:??:?? 0 Power output set to 20
INFO  | ??:??:?? 0 RF95 Radio init succeeded, using RF95 radio
```
- if I use the stock firmware (without TCXO support) on a device that **does not** have TCXO
```
INFO  | ??:??:?? 1 RF95 init result 0
INFO  | ??:??:?? 1 Frequency set to 869.462524
INFO  | ??:??:?? 1 Bandwidth set to 125.000000
INFO  | ??:??:?? 1 Power output set to 20
INFO  | ??:??:?? 1 RF95 Radio init succeeded, using RF95 radio
```


Let me know what's your opinion on this.

code8buster:
It would be better to set it as RADIOLIB_NC since 32 is not the DIO1 pin of the radio on the tcxo variant, it is the busy pin.

Scarg:
> It would be better to set it as RADIOLIB_NC since 32 is not the DIO1 pin of the radio on the tcxo variant, it is the busy pin.

I've updated the PR with the variant.h changes using RADIOLIB_NC as suggested.

code8buster (approved):

code8buster:
Added generic naming scheme per mverch on #3080 
I think it's ready to go in for 2.2.20

Burny08:
I was about to install version 2.3.2 via the Meshtastic Web Flasher on my T-Lora 1.6 2.1, but I still get a Critical Fault #3 error. I thought the TCXO chip was supported now?

I don't have much experience, but maybe someone can check it for me please :)

Thanks!

PR #3126, opened 2024-01-24
fix: allow MQTT `encryption_enabled` with `json_enabled`

andrekir:
this change decouples JSON publishing from MeshPacket encryption and should allow MQTT using both `encryption_enabled` and `json_enabled` simultaneously.

### DIFF
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -248,10 +248,11 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
     // If the packet is not yet encrypted, do so now
     if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
         ChannelIndex chIndex = p->channel; // keep as a local because we are about to change it
-        meshtastic_MeshPacket *p_decoded = new meshtastic_MeshPacket(*p);
+        meshtastic_MeshPacket *p_decoded = packetPool.allocCopy(*p);
 
         auto encodeResult = perhapsEncode(p);
         if (encodeResult != meshtastic_Routing_Error_NONE) {
+            packetPool.release(p_decoded);
             abortSendAndNak(encodeResult, p);
             return encodeResult; // FIXME - this isn't a valid ErrorCode
         }
@@ -261,7 +262,7 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
             if (mqtt)
                 mqtt->onSend(*p, *p_decoded, chIndex);
         }
-        delete p_decoded;
+        packetPool.release(p_decoded);
     }
 
     assert(iface); // This should have been detected already in sendLocal (or we just received a packet from outside)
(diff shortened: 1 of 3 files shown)


caveman99 (approved):

thebentern (approved):

src/mesh/Router.cpp
```diff
@@ -248,15 +248,7 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
...
-
-        if (moduleConfig.mqtt.enabled) {
-
-            LOG_INFO("Should encrypt MQTT?: %d\n", moduleConfig.mqtt.encryption_enabled);
-
-            // the packet is currently in a decrypted state.  send it now if they want decrypted packets
-            if (mqtt && !moduleConfig.mqtt.encryption_enabled)
-                mqtt->onSend(*p, chIndex);
-        }
+        meshtastic_MeshPacket *p_decoded = p;
```
GUVWAF, line 251: Hmm, you're letting `p_decoded` point to the same memory location as `p` here. When `perhapsEncode()` modifies the memory pointed to by `p`, it also modifies where `p_decoded` points to. 

We'd need to make a copy, but then we also need to be careful that the memory gets released afterwards.

GUVWAF (changes requested):

src/mesh/Router.cpp
andrekir: thanks for catching that. fix incoming.

src/mesh/Router.cpp
```diff
@@ -248,15 +248,7 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
...
-
-        if (moduleConfig.mqtt.enabled) {
-
-            LOG_INFO("Should encrypt MQTT?: %d\n", moduleConfig.mqtt.encryption_enabled);
-
-            // the packet is currently in a decrypted state.  send it now if they want decrypted packets
-            if (mqtt && !moduleConfig.mqtt.encryption_enabled)
-                mqtt->onSend(*p, chIndex);
-        }
+        meshtastic_MeshPacket *p_decoded = new meshtastic_MeshPacket(*p);
```
GUVWAF, line 251: For MeshPackets, in the rest of the code we use the `packetPool`. E.g. `packetPool.allocCopy()` and `packetPool.release()` for freeing.

src/mesh/Router.cpp
```diff
@@ -248,15 +248,7 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
...
 
         auto encodeResult = perhapsEncode(p);
         if (encodeResult != meshtastic_Routing_Error_NONE) {
```
GUVWAF, line 254: The copy should be freed before `return` here as well.

GUVWAF (changes requested):

GUVWAF (approved):
Looks good now, thanks!

PR #3142, opened 2024-01-28
GPS updates

GPSFan:
This is the beginning of the GPS updates as a result of my audit of the initialization message flow for the u-blox series of receivers.
The first step was correction several of the init messsages sent by Meshtastic to configure the GPS receiver(s), There are many models and some require different commands to do the same thing. That will be the focus of the next update. This PR only addresses the Neo-6, and adds some groundwork for the future changes.
Testing has been done on my Chatter 2 platform with a Neo-6 module and on a t-beam 1.2 with a Neo-6 as well.


jp-bennett:
We've found some really weird details around the PM2 config message. Comparing the u-center output with https://content.u-blox.com/sites/default/files/products/documents/u-blox6_ReceiverDescrProtSpec_(GPS.G6-SW-10018)_Public.pdf reveals some weird things. The most notable thing is that the PDF says the internal flag must be 000, while u-center sets it to 111.

So the bytes we set as 0x0E814301 should actually be 0x00814301. According to the U-blox spec.

There are many other fields that are marked reserved in the PDF, that u-center sets values to.

src/gps/ubx.h
```diff
@@ -10,19 +10,19 @@ const uint8_t GPS::_message_CFG_RXM_PSM[] PROGMEM = {
...
-    0x00, // Reserved 2
-    0x00, // Reserved 1
-    //    0x0E, 0x81, 0x43, 0x01, // flags-> cyclic mode, wait for normal fix ok, do not wake to update RTC or EPH, doNotEnterOff,
-    //    LimitPeakCurrent
-    0x00, 0x01, 0x03, 0x00, // flags
+    0x01,                   // version
+    0x00,                   // Reserved 1, set to 0x06 by u-Center
+    0x00,                   // Reserved 2
+    0x00,                   // Reserved 1
+    0x00, 0x11, 0x03, 0x00, // flags-> cyclic mode, wait for normal fix ok, do not wake to update RTC or EPH, doNotEnterOff,
```
code8buster, line 24: The flags here are provided by U-Center. Are you basing these changes on only flipping the bits that u-center flips when you toggle the flags in the Message View hex previewer?

GPSFan: My testing conditions are with an external antenna (u-blox ANN-MB-00-00) on the roof with a clear view of the sky. Connected to a Neo-6 module connected to my Chatter 2 esp32 based board. I have a Nordic Power Profiler Kit 2 connected between the Chatters 3.3v supply and the Neo-6. For testing with u-center I have the same module connected to a serial-> USB adapter connected to a loptop running u-center 23.08.

PR #3147, opened 2024-01-30
Added LilyGo T3 V1.6.1 TCXO  Version Support

lewisxhe:
The LilyGo T3 V1.6.1 TCXO version is different from the normal version and the original variant cannot be used.
https://www.lilygo.cc/products/lora3?variant=43365007425717

protobufs PR:

https://github.com/meshtastic/protobufs/pull/426

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -15,6 +15,7 @@ default_envs = tbeam
 ;default_envs = tlora_v1_3
 ;default_envs = tlora-v2
 ;default_envs = tlora-v2-1-1_6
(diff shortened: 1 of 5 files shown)


code8buster:
https://github.com/meshtastic/firmware/pull/3124

This is probably more to your (and our!) liking; it achieves the same result while not disabling any future TCXO-via-SPI configurations.

lewisxhe:
Ohh. I didn't see that someone has already submitted the update, thank you, I will close it

Issue #3129, opened 2024-01-24
[Bug]: Node crashes when establishing WLAN connection with the Android APP


mverch67:
### Category

WiFi

### Hardware

T-Deck

### Firmware Version

since 2.2.16

### Description

T-Deck boots in a circle when WLAN is enabled and the Android APP is waiting for connection. If the APP node connection is set to _None_, then T-Deck continues to run. At the moment Android App is set to connect via IP the node crashes and reboots. 

Android APP 2.2.19

mverch67:
This issue was introduced in 2.2.16. In 2.2.15 the WLAN connection still works.

mverch67:
Summary of investigation result:
- either commenting out the GPS auto detection -> does not show the wifi reboot anymore
- or commenting out the I2S stuff(audio) -> does not show the wifi reboot anymore
- both code included -> crash/reboot (even when gps detection finished after a minute)

PR #3157, opened 2024-01-31
Triple stack

thebentern:


src/ButtonThread.h
```diff
@@ -205,6 +205,16 @@ class ButtonThread : public concurrency::OSThread
         }
     }
 
+    static meshtastic_Config_PositionConfig_GpsMode toggleGpsMode(meshtastic_Config_PositionConfig_GpsMode currentMode)
```
mverch67, line 208: I'd rather not put this GPS related method into this ButtonThread class. Doesn't it belong to NodeDB for determining the default/migrated state?

src/mesh/NodeDB.cpp
```diff
@@ -182,7 +182,11 @@ void NodeDB::installDefaultConfig()
 #else
     config.device.disable_triple_click = true;
 #endif
-    config.position.gps_enabled = true;
+#if HAS_GPS == 0
+    config.position.gps_mode = meshtastic_Config_PositionConfig_GpsMode_NOT_PRESENT;
+#else
+    config.position.gps_mode = meshtastic_Config_PositionConfig_GpsMode_ENABLED;
```
mverch67, line 188: Here I think we need to check for GPS_RX_PIN and config.position.rx_gpio. If neither is set then NOT_PRESENT.
Also #ifdef T_DECK then not_present.

src/gps/GPS.cpp
```diff
@@ -252,6 +252,11 @@ bool GPS::setup()
 {
     int msglen = 0;
 
+    if (config.position.gps_mode == meshtastic_Config_PositionConfig_GpsMode_NOT_PRESENT) {
```
mverch67, line 255: I think this could be checked already in runOnce() just before setup() is called, so that disable() is executed.

src/gps/GPS.cpp
```diff
@@ -625,659 +630,678 @@ void GPS::publishUpdate()
...
+        setAwake(false);
+
+        return INT32_MAX;
+    }
 
-    return INT32_MAX;
-}
\ No newline at end of file
+    void GPS::toggleGpsMode()
+    {
```
mverch67, line 1293: Wouldn't it be simpler (and more logical) to just toggle between enabled/disabled and if not present it does nothing?

thebentern:
Looks good to me

mverch67:
T-Deck WLAN still rebooting, workaround didn't help

mverch67:
This is a workaround for #3129

PR #3163, opened 2024-02-02
Fixes #3074 and some other minor issues, improves battery level

### LINKED ISSUE #3074: Heltec Tracker Screen Timeout[Bug]:
### Category

Hardware Compatibility

### Hardware

Heltec Wireless Tracker v1.1

### Firmware Version

2.2.17.dbac2b1 and prior

### Description

I have two Heltec Wireless Tracker devices. The screen timeout is non-functional and the display remains powered up all the time no matter what value I enter in for the time delay.

### Relevant log output

_No response_

Gabrielerusso:
Fixes #3074  of heltec tracker, i found other issues such as the I2C pin not specified and the stock 21/22 were used, but pin 22 is not available, now using 45/46. Also the screen dimension and offsets now are correct.

A major difference was to split the two hardware version, to reduce overhead code as all the new boards are V1.1 and to me it made little sense to have such differentiations at runtime. The two hardware version have the same "external" name, just two different binaries.
new env names: 
1. heltec-wireless-tracker-V1_0
2. heltec-wireless-tracker-V1_1

NOTE: i2c pin fix was never tried, original i2c pin are reported as 41/42 but that was false as those are used for the TFT SPI


Gabrielerusso:
Added lookup table for a better battery level estimation, using a 11 value table and interpolating between them, for the moment shoould be good enough, if you wan't i could add more, but 11 values in my test (during BMS development) seemed good enough.
Still the automatic heltec tracker build fails as the new envs are:
- heltec-wireless-tracker-V1_0
- heltec-wireless-tracker-V1_1

thebentern:
@Gabrielerusso couple of things before we move forward. Can you sign the CLA and make the protobuf changes in a PR in the protobufs repo first?

src/Power.cpp
```diff
@@ -212,19 +229,16 @@ class AnalogBatteryLevel : public HasBatteryLevel
...
-            digitalWrite(ADC_CTRL, HIGH);
-            delay(10);
-        }
+        pinMode(ADC_CTRL, OUTPUT);
+        digitalWrite(ADC_CTRL, HIGH);
+        delay(10);
 #endif
         for (int i = 0; i < BATTERY_SENSE_SAMPLES; i++) {
             raw += adc1_get_raw(adc_channel);
+            delayMicroseconds(100);
```
code8buster, line 238: What problem is this solving? IDK that we should be delaying in something that runs as often as this task runs.

Gabrielerusso: none, this could be removed, i just gave a look at adc1_get_raw and it's waiting on it's own for the sampling. a delay may help with noise mitigation if the board has a LPF but i just wrote it as an habit.
I also think 10ms for the ADC_CTRL it's way too much but i hadn't time to give a peek with the oscilloscope.

Gabrielerusso:
> @Gabrielerusso couple of things before we move forward. Can you sign the CLA and make the protobuf changes in a PR in the protobufs repo first?

CLA signed, I made a request for the updated protobuf and I hope it was correct. Pretty busy in this period and I'm reading/writing when I've a minute so I may have done something wrong with protobuf that I'm unfamiliar.

Gabrielerusso:
Quick note, sometimes the Heltec Tracker says "USB" when it's on battery, after a restart it shows back the battery voltage. This does happen also with the 2.2.19 so I don't think is related with my changes.

Somehow it doesn't recognize the battery presence.

Gabrielerusso:
I know those are starting to be a lot of commit and file changed for a single PR, maybe it would've been better to do so in multiple PR.

src/Power.cpp
```diff
@@ -226,19 +224,24 @@ class AnalogBatteryLevel : public HasBatteryLevel
...
+#ifdef ADC_CTRL //enable adc voltage divider when we need to read
         pinMode(ADC_CTRL, OUTPUT);
-        digitalWrite(ADC_CTRL, HIGH);
+        digitalWrite(ADC_CTRL, ADC_CTRL_ENABLED);
         delay(10);
 #endif
         for (int i = 0; i < BATTERY_SENSE_SAMPLES; i++) {
-            raw += adc1_get_raw(adc_channel);
+            int val_ = adc1_get_raw(adc_channel);
+            if(val_ >= 0){ //save only valid readings
```
Gabrielerusso: Check for valid readings, invalid should give -1 but could be usefull to use strictly greather than 0 instead of >=.

src/Power.cpp
```diff
@@ -226,19 +224,24 @@ class AnalogBatteryLevel : public HasBatteryLevel
...
         uint32_t raw = 0;
+        uint8_t raw_c = 0; //raw reading counter
 
 #ifndef BAT_MEASURE_ADC_UNIT // ADC1
-#ifdef ADC_CTRL
+#ifdef ADC_CTRL //enable adc voltage divider when we need to read
         pinMode(ADC_CTRL, OUTPUT);
-        digitalWrite(ADC_CTRL, HIGH);
+        digitalWrite(ADC_CTRL, ADC_CTRL_ENABLED);
         delay(10);
```
Gabrielerusso, line 234: if there aren't filter caps with the resistor value i've seen  the fet should saturate in less than 3ms

code8buster:
> I know those are starting to be a lot of commit and file changed for a single PR, maybe it would've been better to do so in multiple PR.

Generally yes lol it does make it easier to merge when it's not all wrapped up. I'll spend some time this evening reviewing the changes.

Gabrielerusso:
> > I know those are starting to be a lot of commit and file changed for a single PR, maybe it would've been better to do so in multiple PR.
> 
> Generally yes lol it does make it easier to merge when it's not all wrapped up. I'll spend some time this evening reviewing the changes.

I'm step-by step trying to divide the changes in different branches to do multiple PR

Gabrielerusso:
i'm keeping this as draft for a personal reference, i did the first PR, once that get reviewed i'll make another one, also i'm removing all the "Merge remote-tracking branch 'upstream/master'" by doing a rebase from time to time. This should make the commits much easier to understand.

PR #3175, opened 2024-02-07
Update link to docs from webserver when file not found

tekstrand:
## What
1. Updates link to documentation shown when the webserver is unable to find the requested file

## Why
Current link 404s


src/mesh/http/ContentHandler.cpp
```diff
@@ -854,4 +854,4 @@ void handleScanNetworks(HTTPRequest *req, HTTPResponse *res)
     JSONValue *value = new JSONValue(jsonObjOuter);
     res->print(value->Stringify().c_str());
     delete value;
-}
\ No newline at end of file
+}
```
tekstrand, line 857: How y'all feel about EOF here? Let me know if I need to change.

thebentern (approved):

PR #3176, opened 2024-02-07
Update variant.h heltec v3 adc multiplier to match actual readings

fuutott:
As per message adc multiplier 5.05 results in correct battery readings for heltec v3


Gabrielerusso:
are you sure this is valid for all V3 chips? according to schematic divider should be 4.9 (390k and 100k) but there is a fet, so sligth variations from board to board will occour and aren't the same for all batches.

thebentern:
I recommend you use the power.adc_multiplier_override to compensate for the small differences in your setup.

PR #3183, opened 2024-02-08
MQTT JSON downlink fixes

### LINKED ISSUE #3168: [Bug]: New MQTT subtopic breaks the logic for parsing the channel when sending JSON.
### Category

Other

### Hardware

Not Applicable

### Firmware Version

2.2.21.7f7c5cb

### Description

The code on the MQTT module that checks if the MQTT message was sent to a channel called mqtt:

```
// parse the channel name from the topic string
            char *ptr = strtok(topic, "/");
            for (int i = 0; i < 3; i++) {
                ptr = strtok(NULL, "/");
            }
```
is fine when the topic is **msh/2/json/mqtt** but with the updated scheme breaking up mqtt into regions **msh/EU_868/2/json/mqtt** that now thinks the channel name is "json/mqtt" instead of "mqtt" and the message is rejected because it isn't a known channel called mqtt.

I have verified with a private MQTT server that the topic root must only be one level deep, any more than that and you can't send a JSON MQTT message for publishing to the mesh.

### Relevant log output

_No response_

GUVWAF:
- Fixes #3168 by looking for the last occurrence of "/" in the topic, instead of the third.
- Allow specifying channel index to send on via LoRa in new JSON field "channel" while still requiring the message to be published to a channel called "mqtt".

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -67,7 +67,9 @@ void MQTT::onReceive(char *topic, byte *payload, size_t length)
                         // construct protobuf data packet using TEXT_MESSAGE, send it to the mesh
                         meshtastic_MeshPacket *p = router->allocForSending();
                         p->decoded.portnum = meshtastic_PortNum_TEXT_MESSAGE_APP;
-                        p->channel = sendChannel.index;
+                        if (json.find("channel") != json.end() && json["channel"]->IsNumber() &&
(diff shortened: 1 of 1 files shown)


thebentern (approved):

GUVWAF:
Latest commit makes it non-breaking: apparently in some places in the documentation another "/" was added after the channel name.

thebentern (approved):

PR #3184, opened 2024-02-08
Fix init resolution for all architectures

code8buster:
RP2040 was initializing to 10bit adc resolution b/c of the analogReadResolution call being cordoned off by a preprocessor ifdef.
Now fixed, hopefully for all archs

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -389,11 +389,9 @@ bool Power::analogInit()
 #else
     analogReference(AR_INTERNAL); // 3.6V
 #endif
-    analogReadResolution(BATTERY_SENSE_RESOLUTION_BITS); // Default of 12 is not very linear. Recommended to use 10 or 11
-                                                         // depending on needed resolution.
-
 #endif // ARCH_NRF52
-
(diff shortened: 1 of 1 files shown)


src/Power.cpp
```diff
@@ -389,10 +389,8 @@ bool Power::analogInit()
 #else
     analogReference(AR_INTERNAL); // 3.6V
 #endif
-    analogReadResolution(BATTERY_SENSE_RESOLUTION_BITS); // Default of 12 is not very linear. Recommended to use 10 or 11
-                                                         // depending on needed resolution.
-
 #endif // ARCH_NRF52
+    analogReadResolution(BATTERY_SENSE_RESOLUTION_BITS);
```
Gabrielerusso, line 393: This breaks ESP32 architecture adc readings as it will set the ADC to 10bit when the default value for esp32 is set to 12bit. Fixed in [11d78be](https://github.com/meshtastic/firmware/pull/3163/commits/11d78be541809a2ebba8ea9d58b03619c0336e68)

PR #3189, opened 2024-02-09
ATAK plugin

thebentern:
Handle TAK Packets with transparent compression / decompression of any contained strings with unishox2.

Added some plumbing for altering MeshPackets flowing through the protobuf module pipeline as well.


src/modules/Modules.cpp
```diff
@@ -61,7 +62,7 @@ void setupModules()
         traceRouteModule = new TraceRouteModule();
         neighborInfoModule = new NeighborInfoModule();
         detectionSensorModule = new DetectionSensorModule();
-
+        atakPluginModule = new AtakPluginModule();
```
GUVWAF, line 65: Maybe we can only enable it for the `TAK` role to enforce people using that for even more optimized settings?

thebentern: I had wondered if we might do that as well. I'm somewhat on the because I'm not sure what value it holds to gate it off with the role, since the module should ignore traffic outside of its portnum otherwise. anyway

GUVWAF: I see, that's true.

PR #3193, opened 2024-02-10
E-Ink "Dynamic Partial"

todd-herbert:
> The other issue to include in discussion is the need for a full update to be includes after so many partial updates - e-ink displays are not meant to be partially updates for ever.

https://github.com/meshtastic/firmware/issues/3075#issuecomment-1885420501

## What it is
Selects between full refresh, partial refresh, and skipped updates, balancing urgency and display health. Non-intrusive, disabled by default. Implementable for individual boards.

#### 1. Configure in `variant.h`
```cpp
#define USE_EINK
#define USE_EINK_DYNAMIC_PARTIAL
#define EINK_LOWPRIORITY_LIMIT_SECONDS 30
#define EINK_HIGHPRIORITY_LIMIT_SECONDS 1
#define EINK_PARTIAL_REPEAT_LIMIT 5
```
  * `EINK_LOWPRIORITY_LIMIT_SECONDS`: interval between "low priority" updates
  * `EINK_HIGHPRIORITY_LIMIT_SECONDS`: min. interval between "high priority" updates
  * `EINK_PARTIAL_REPEAT_LIMIT`: max. consecutive partial refreshes, before a full refresh is forced

#### 2. Adjust `configForPartialRefresh()` and `configForFullRefresh()`, if necessary 

These methods are used internally to execute any code required to switch the display between partial and full refresh configurations. The *should* work out-of-the-box, but any "display specific" code can go here.

### In GFX code

Graphics code should suggest that an update is either "high priority" or "low priority". This is considered by the *Dynamic Partial* code, when determining whether to use partial or full refresh.
* `EInkDisplay::highPriority()` - subsequent updates are important: show ASAP
* `EInkDisplay::lowPriority()` - no rush with these ones

For the EInkDisplay class as it currently exists,  **this pull request has already implemented the required changes.**
This is hooked into the existing update `display()` and `forceDisplay()` methods:
* updates called via `display()` are considered *low priority* - receives one per `loop()` (?)
* updates called via `forceDisplay()` are considered *high priority*

## What does it change?

The main *selection* process is handled by this block, in `EInkDisplay::forceDisplay()`
```cpp
#if defined(USE_EINK_DYNAMIC_PARTIAL)
    // Decide if update is partial or full
    bool continueUpdate = determineRefreshMode();
    if (!continueUpdate)
        return false; // Abort forceDisplay()
#else
```

For the most part, the changes are additions, selected by the preprocessor. One exception is the minor refactoring of the "rate-limiting" code in `forceDisplay()`:
https://github.com/meshtastic/firmware/blob/1085b54069e426b592f7d458548de6e485f72c8b/src/graphics/EInkDisplay2.cpp#L131

Structure has changed from:
```cpp
if (condition_met)
  doStuff();
```

to:
```cpp
if (!condition_met)
  return;

doStuff();
```

## What will I notice?
Probably, nothing. This pull request only enables *Dynamic Partial* mode for `HELTEC_WIRELESS_PAPER_V1_0`. Ideally, there should be no impact on other boards.

The hope is that devs will test *Dynamic Partial* mode on their own hardware, and if the result is good, implement it for the relevant boards; even if only as an interim measure, pending a rewrite of the EInkDisplay class I had seen [mentioned in passing](https://github.com/meshtastic/firmware/issues/3075#issuecomment-1902231040).

## How is the refresh mode selected?
I will admit, the logic is slightly convoluted. I do believe though that the resulting behavior feels intuitive.

A rough outline:

flowchart of dynamic partial modes' selection logic

### DIFF
--- a/variants/heltec_wireless_paper_v1/variant.h
+++ b/variants/heltec_wireless_paper_v1/variant.h
@@ -5,6 +5,14 @@
 #define I2C_SCL SCL
 
 #define USE_EINK
+
+// Settings for Dynamic Partial mode
+// Change between partial and full refresh config, or skip update, balancing urgency and display health.
+#define USE_EINK_DYNAMIC_PARTIAL
+#define EINK_LOWPRIORITY_LIMIT_SECONDS 30
+#define EINK_HIGHPRIORITY_LIMIT_SECONDS 1
+#define EINK_PARTIAL_REPEAT_LIMIT 5
+
 /*
  * eink display pins
  */
--- a/src/graphics/EInkDisplay2.cpp
+++ b/src/graphics/EInkDisplay2.cpp
@@ -125,61 +125,68 @@ bool EInkDisplay::forceDisplay(uint32_t msecLimit)
     // No need to grab this lock because we are on our own SPI bus
     // concurrency::LockGuard g(spiLock);
 
+#if defined(USE_EINK_DYNAMIC_PARTIAL)
+    // Decide if update is partial or full
+    bool continueUpdate = determineRefreshMode();
+    if (!continueUpdate)
+        return false;
+#else
+
     uint32_t now = millis();
     uint32_t sinceLast = now - lastDrawMsec;
 
-    if (adafruitDisplay && (sinceLast > msecLimit || lastDrawMsec == 0)) {
+    if (adafruitDisplay && (sinceLast > msecLimit || lastDrawMsec == 0))
         lastDrawMsec = now;
-
-        // FIXME - only draw bits have changed (use backbuf similar to the other displays)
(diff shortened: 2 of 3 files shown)


variants/heltec_wireless_paper_v1/variant.h
markbirss, line 1: I see value in adding the ability to perform dynamic partial updates and we should certainly give it some test time with the heltec wireless paper

markbirss (approved):

markbirss:
@todd-herbert nice to see the display working without line at the bottom like the RAK14000

@thebentern sent me a v1 board and i already flashed your wip tree, just wish it had a i2c port.

thebentern:
Great work. This is a good approach that we can slowly test out on other eink boards.

thebentern (approved):

PR #3194, opened 2024-02-10
StoreForward updates

GUVWAF:
- Send history in "text" variant (needs protobuf PR https://github.com/meshtastic/protobufs/pull/437).
- Don't send history the client already got when `last_request` is given.
- Check if PSRAM is full.
- More sensible defaults.

Already does something using the iOS app, but it currently puts a `*` before each message:
image


GUVWAF:
@garthvh @andrekir With the latest commit, when the Router sends its history, it will set the `RequestResponse` tag to `ROUTER_TEXT_DIRECT` for direct text messages and to `ROUTER_TEXT_BROADCAST` for broadcasts.

andrekir:
> @garthvh @andrekir With the latest commit, when the Router sends its history, it will set the `RequestResponse` tag to `ROUTER_TEXT_DIRECT` for direct text messages and to `ROUTER_TEXT_BROADCAST` for broadcasts.

so we would need to request broadcasts or DMs separately?

GUVWAF:
@andrekir No, you can request the history of all packets with the `CLIENT_HISTORY` tag with `history` variant. The router then first responds with one `ROUTER_HISTORY` packet containing the amount of packets it is going to send in the field `history_messages` and the index of the last packet in `last_request` (which you can save and set in a subsequent request to avoid getting the same message twice even if the time window overlaps). Then it will send all the packets, which have either the `TEXT_DIRECT` or `TEXT_BROADCAST` tag.

andrekir:
> @andrekir No, you can request the history of all packets with the `CLIENT_HISTORY` tag with `history` variant. The router then first responds with one `ROUTER_HISTORY` packet containing the amount of packets it is going to send in the field `history_messages` and the index of the last packet in `last_request` (which you can save and set in a subsequent request to avoid getting the same message twice even if the time window overlaps). Then it will send all the packets, which have either the `TEXT_DIRECT` or `TEXT_BROADCAST` tag.

sounds really good!

garthvh:
What does the router history do now? That is what I was sending before.

GUVWAF:
@garthvh I believe you were sending the `CLIENT_HISTORY`: 
https://github.com/meshtastic/Meshtastic-Apple/blob/cd2d70c0563ba65a0d48b9d8c9ed1ff49427f0cb/Meshtastic/Helpers/BLEManager.swift#L2384
Which is good, that's the one you use to request the history, nothing changed there.

You'll only receive the `ROUTER_HISTORY` that has some informative fields like the number of messages the router is going to send.

thebentern (approved):

garthvh (approved):
This is all working on my end

thebentern (approved):

PR #3195, opened 2024-02-10
Added RP2040-LoRA target

thebentern:
Based on @markbirss's work


markbirss:
Im my tests i did get some noise and bad packets

thebentern:
> Im my tests i did get some noise and bad packets

Perhaps there is something misconfigured with the SX1262 pins in the variant. The pin map indicates a SX1262_ANT_SW on GPIO17 which sounds similar to how the xiao-ble variant uses the RXEN pin only in conjunction with DIO2. I wonder if the noise and bad packets could be that the switch is still in transmit mode as we try to receive message.

Issue #3131, opened 2024-01-25
[Bug]: Wireless Paper (ESP32s3) does not show battery percentage


ixt:
### Category

Hardware Compatibility

### Hardware

Heltec Wireless Paper

### Firmware Version

2.2.17+

### Description

Currently Wireless Paper (Assuming both versions, although only tested v1.1) does not display battery power, this is due to the battery sensing pin being on ADC2 and incorrectly stated. 

ADC2 doesn't by default like being read from ESP32 or ESP32s3, and has to be read via the RTC due to Wifi-related issues in IDF (If i'm understanding correctly). 

The [schematic](https://resource.heltec.cn/download/Wireless_Paper/Wireless_Paper_V0.4_Schematic_Diagram.pdf) doesn't make obvious which pin we should be reading from, but its either 19 or 20. In the Heltech v2 boards we use a workaround that sets up the registers (SENS_SAR_READ_CTRL2_REG). But on ESP32s3 they are renamed to SENS_SAR_READER2_CTRL_REG. Making those updates however doesn't get me a working solution. I dont quite understand the esp_adc_cal code yet, so not sure which knobs to turn to help it along, but I think I've found the right tree to bark up...

I imagine this would be an issue on any other boards that happen to place the battery sensing on the ADC2 on ESP32/s3/c3. So any Heltec folks listening (if they are here), for v1.2 move the battery sensing to a pin with ADC1, I know thats a ridiculous ask but it does seem like even if we get this working it'll probably be unreliable. 

I tried finding other examples of people using ADC2 on a ESP32s3, but I don't feel convinced anyone has managed to do it from scratch as most of the instances I could find where just ESP32 examples with commented out parts that would have given compile issues.

It's possible I missed some so it's worth searching anyway. Attached is a log of the assert fail I get from the loop. 

Again, not my area at all so might have just missed something obvious that someone in the area might know. Don't mistake my depth for subject knowledge lol

I also changed the battery pin stuff to this 

```
#define ADC_MULTIPLIER 0.02
#define BATTERY_PIN 20 // A battery voltage measurement pin, voltage divider connected here to measure battery voltage
#define ADC_ATTENUATION ADC_ATTEN_DB_11 // lower dB for high resistance voltage divider
#define ADC_CTRL 19
#define ADC_CHANNEL ADC2_GPIO20_CHANNEL
#define BAT_MEASURE_ADC_UNIT 2
```

[...]

PR #3208, opened 2024-02-11
fix: Wireless Paper (v1.0 & v1.1) not showing battery percentage

todd-herbert:
Addresses https://github.com/meshtastic/firmware/issues/3131

### Affects:
Boards which are both:
* ESP32S3, and
* Measure battery with ADC2

As far as I am aware, this is currently only Wireless Paper V1.0 and V1.1. 
Presumably any new boards meeting the criteria will suffer from the same issue?

### Caveats:
`usbPower` and `isCharging` are not working correctly.
Looking at the [schematic](https://resource.heltec.cn/download/Wireless_Paper/Wireless_Paper_V0.4_Schematic_Diagram.pdf), it might not be possible to get the necessary information (?)
* With no battery connected, `VBAT` hovers around 3.9V
* The `C̅H̅R̅G̅` pin of the TP4054 Li-lon Battery Charger IC does not connect to the MCU, only to an orange LED.

### Code changes:
* ADC config in `variant.h` (both boards)
* Use gate `ADC_CTRL` to connect / disconnect voltage divider
* For ESP32S3: avoid "ADC2 wifi bug workaround". Not required for S3, and breaks compile: https://github.com/meshtastic/firmware/blob/1d448157811b3fe77cd3f063d072359f16f528c0/src/Power.cpp#L271-L273
  Instead, check the `esp_err_t` returned by `adc2_get_raw()` to confirm that read was successful.

* For ESP32S3: handle new enum value in logging: https://github.com/meshtastic/firmware/blob/1d448157811b3fe77cd3f063d072359f16f528c0/src/Power.cpp#L428-L433


src/Power.cpp
```diff
@@ -226,7 +226,46 @@ class AnalogBatteryLevel : public HasBatteryLevel
...
+                read_result = adc2_get_raw(adc_channel, ADC_WIDTH_BIT_12, &adc_buf);
+                if (read_result == ESP_OK)
+                    break;
+            }
+
+            // Handle the result
+            if (read_result == ESP_OK)
+                raw += adc_buf;
+            else {
+                // How to handle read failure?
```
todd-herbert, line 262: Just a heads up: if ten read attempts fail, the current behavior is to record 0V for the sample. 
Not sure if this should be handled differently.

For what it's worth, during brief testing (with WiFi connected), I don't remember seeing a single failed reading, let alone ten, so it may be an uncommon scenario.

todd-herbert:
@code8buster I do realize the suggestion was that the deprecated interface should be avoided. https://github.com/meshtastic/firmware/issues/3131#issuecomment-1909274163

I did look into the new API, but as far as I could tell (and I'm no expert), the required methods haven't yet been implemented in the [Heltec Enviornment](https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/), despite recent releases.

Gabrielerusso:
> Caveats:
usbPower and isCharging are not working correctly.
Looking at the [schematic](https://resource.heltec.cn/download/Wireless_Paper/Wireless_Paper_V0.4_Schematic_Diagram.pdf), it might not be possible to get the necessary information (?)
With no battery connected, VBAT hovers around 3.9V
The C̅H̅R̅G̅ pin of the TP4054 Li-lon Battery Charger IC does not connect to the MCU, only to an orange LED.

same as Heltec Tracker [schematic](https://resource.heltec.cn/download/Wireless_Tracker/Wireless_Tacker1.1/HTIT-Tracker_V0.5.pdf), charge voltage is around 4.15V (average), isCharging in my case is working only with full battery as the lookup table i use measures anything above 4150mV as fully charged, so by going over that it recognizes it's connected.

For the usbPower also there is the same problem.

I don't understand why the charge output of the TP4057 isn't connected even if would be pretty simple to do so afterwards (DIY).

src/Power.cpp
Gabrielerusso: check my PR, i added a valid reading counter so the division is made only by the valid readings, this will solve the invalid readign issue even if not probable.

todd-herbert: Good idea! https://github.com/meshtastic/firmware/pull/3208/commits/05dd680dfa54473f0bbbda1c957c12d956d1802c

src/Power.cpp
```diff
@@ -226,7 +227,37 @@ class AnalogBatteryLevel : public HasBatteryLevel
             digitalWrite(ADC_CTRL, LOW);
         }
 #endif
-#else  // ADC2
+#else // ADC2
+#ifdef ADC_CTRL
+#if defined(HELTEC_WIRELESS_PAPER) || defined(HELTEC_WIRELESS_PAPER_V1_0)
+        pinMode(ADC_CTRL, OUTPUT);
+        digitalWrite(ADC_CTRL, LOW); // ACTIVE LOW
+        delay(10);
```
todd-herbert, line 235: Could (should) this delay be shorter? 10ms was originally taken from https://github.com/meshtastic/firmware/blob/05dd680dfa54473f0bbbda1c957c12d956d1802c/src/Power.cpp#L215-L220

Mentioned here: https://github.com/meshtastic/firmware/pull/3163/commits/58acd2809eb2e5148bee2170baef122e6a43605b#r1485661627

Datasheet for the FET: [AO3401](https://aosmd.com/sites/default/files/res/datasheets/AO3401.pdf)

Gabrielerusso: I Need to do some measurements with the oscilloscope to evaluate rise/fall time with those resistors more precisely. Unfortunately now I don't have the time to do that, I'll do it in 3/4 weeks.

todd-herbert:
@Gabrielerusso Sorry, I bumped the button there. I know you've already reviewed it. Thanks

code8buster (approved):
I don't really have hardware to test this with, but it does pass the smell test.
It's possible that your code for reading adc2 is actually better overall. 

I never found anyone to test the "workaround" with a heltec v2/2.1 that was also connected to wifi.

todd-herbert:
It might have been unnecessary at this point, but here is at least *some* measure of proof.

<a href="https://github.com/meshtastic/firmware/assets/24772776/599f90b6-c1e3-4929-90d1-ae4a37b06634">

</a>

Issue #3187, opened 2024-02-09
[Bug]: Heltec Tracker display reacts with a delay and the display does not go out


spacepc-de:
### Category

Other

### Hardware

Heltec Wireless Tracker

### Firmware Version

2.2.19

### Description

I have 7 Heltec trackers with the current firmware 2.2.19. On all of them, the display does not go out after a defined timeout, no matter what I set. In addition, when I press the button to change the views, it reacts with a long delay or not at all. Since the display is always on, the battery life is greatly reduced.

Gabrielerusso:
> The button accuracy should be improved by this PR: #3188. The screen is (re)printed one time/sec, so it takes one second to advance to the next.
> 
> The newer heltec tracker model v1.1 has a different power logic. AFAIK it's not possible to switch off the screen without also switching off the LoRa chip.

It is possible to switch the screen off, it was the 1.0 that has a problem. My PR #3163 fixes it.

BobsBlueNorth:
I have been seeing the same issue on my v1.1 boards with all newer firmware since 2.11. By rolling back to 2.11 they are useable. I just tried 2.22 today but no dice. The screen "locks" up at the timeout.

I didn't realize there was a 2.23 to try.

dchauran:
Came here to report thr.timeout issue, only to find that it had not only been reported but fixed and was in Alpha. I installed alpha and it worked. Screen refresh is faster and the display does indeed turn off now.

PR #3213, opened 2024-02-12
Fixes [3074] Heltec Tracker Screen issues + minor fixes

Gabrielerusso:
Fixes [3074](https://github.com/meshtastic/firmware/issues/3074) and [3187](https://github.com/meshtastic/firmware/issues/3187) of heltec tracker, i found other issues such as the I2C pin not specified and the stock 21/22 were used, but pin 22 is not available, now using 45/46. Also the screen dimension and offsets now are correct.

A major difference was to split the two hardware version, to reduce overhead code as all the new boards are V1.1 and to me it made little sense to have such differentiations at runtime. The two hardware version have the same "external" name, just two different binaries.
new env names:

heltec-wireless-tracker-V1_0
heltec-wireless-tracker-V1_1
NOTE: i2c pin fix was never tried, original i2c pin are reported as 41/42 but that was false as those are used for the TFT SPI.

This is related to PR [3163](https://github.com/meshtastic/firmware/pull/3163), i'm just dividing that in multiple PR for easier management


variants/heltec_wireless_tracker_V1_1/platformio.ini
```diff
@@ -0,0 +1,12 @@
+[env:heltec-wireless-tracker-V1-1]
```
thebentern, line 1: Let's make this platformio variant name remain 'heltec-wireless-tracker' for v1.1? Otherwise, this breaks all of the previous CI/CD work as well as the meshtastic API for the web flasher.

Gabrielerusso: yeah we can do that, i just hope they won't make another version.

code8buster (approved):

thebentern:
@Gabrielerusso FYI, we had to revert the HW_MODEL name change of HELTEC_WIRELESS_TRACKER_V1_1 back to HELTEC_WIRELESS_TRACKER because it was a breaking change for the Android platform.

variants/heltec_wireless_tracker_V1_1/platformio.ini
```diff
@@ -0,0 +1,14 @@
+[env:heltec-wireless-tracker]
+extends = esp32s3_base
+board = heltec_wireless_tracker
+upload_protocol = esp-builtin
+
+build_flags = 
+  ${esp32s3_base.build_flags} -I variants/heltec_wireless_tracker_V1_1
+  -D HELTEC_TRACKER_V1_1
```
thebentern, line 8: I believe this needs to be changed to have the correct HW_MODEL selection again in architecture.h

Gabrielerusso: should i change heltec_wireless_tracker_V1_1 to heltec_wireless_tracker and the relative folder name?

code8buster: That means the main_matrix.yml would need changed back as well if we've reverted Gabriel's protobuf commit right?

Gabrielerusso: Variants adds only the 1.0 version, should I just change the folder name of the 1.1?

thebentern: Yes, I think folder name, variant, and the -D HELTEC_TRACKER_V1_1

Gabrielerusso: So we should completely drop the support to V1.0 PCB and leave only the names as before?

thebentern: I realize I wasn't very clear. 
The 1.0 variant is fine, but we want the platformio assets for the 1.1 variant to remain (heltec_wireless_tracker) for backwards compatibility of the firmware flashing API collecting those assets. v1.1 is also the most common version folks have. Splitting this target into two discrete ones will surely involve some support cleanup effort, but this will minimize it a bit.

thebentern (approved):

thebentern:
@Gabrielerusso if you can fix the merge conflicts, I'll just go ahead and merge and fix the variant.

Gabrielerusso:
> @Gabrielerusso if you can fix the merge conflicts, I'll just go ahead and merge and fix the variant.

i'm sorry, due to imminent exams i'm too busy rigth now.

Gabrielerusso:
> @Gabrielerusso if you can fix the merge conflicts, I'll just go ahead and merge and fix the variant.

should be all ok, i'll try to compile and flash it rigth now and see if everything is working.

thebentern (approved):

Issue #3172, opened 2024-02-05
[Bug]: Heltec Tracker can't turn off GPS location.


Edgs:
### Category

Hardware Compatibility

### Hardware

Heltec Wireless Tracker

### Firmware Version

2.2.19

### Description

Tracker transmits live GPS location, even though it appears disabled in settings.
'Smart Position' = disabled
'Use fixed position' = disabled
'GPS enabled' = disabled
All the update and broadcast intervals are at their default (non zero) values.

Gabrielerusso:
i don't remember having this problem, both disabling it via settings and via triple button push.

Edgs:
It does appear to have a 'last known location' stored in the fixed position values, but it continues to transmit this as a current location, even though 'Use fixed position' is disabled.
I would expect it to transmit no location with the above settings.

Gabrielerusso:
> It does appear to have a 'last known location' stored in the fixed position values, but it continues to transmit this as a current location, even though 'Use fixed position' is disabled. I would expect it to transmit no location with the above settings.

"Use fixed position" it's for the fixed position value you set, that you see it's the last known one, this is very usefull for a tracker as not every time the gps fix is available. This is the same for all meshtastic devices, if you want to reset every past GPS data you can simply press the reset button and this will erase the last known GPS location.

It's possible to edit the firmware to reset the location data when the GPS is disabled, but i don't know if this would be an issue for some users that activate the GPS one time to get a position and then deactivate it.

Example: i'm using a Heltec Tracker to track my dog, the dog goes inside an house where the location data isn't available, the tracker keeps giving the last known position (outside the house) .

Edgs:
I agree with all the posibilities you list there, except:
> ...if you want to reset every past GPS data you can simply press the reset button and this will erase the last known GPS location.
This does not happen - the 'last location' is always there, and always transmitted, even though both Location options are disabled.

The user would want the last location to transmit if GPS was enabled, but unable to fix it's location.
The user would want the 'fixed location' to be transmitted if 'Transmit fixed location' was enabled.
But if GPS is disabled, and 'Transmit fixed location' is disabled, then it should not transmit any location.

I'm setting these options using the Android app. incidentally.
The location in 'Fixed location' came from a GPS position on initial power on after installation - I didn't enter anything.
If I do try to set the Location to 0.0, 0.0, after reboot the original Location is still set, and still being transmitted.

Gabrielerusso:
[...]

PR #3215, opened 2024-02-12
#3172 Fixes possible position leak when location is disabled

Gabrielerusso:
Related to #3172 fixes location leaks when location is disabled and we don't want our position to be given.
basically we send our position only if the location is enabled, and if it get disabled we reset the local data to avoid any unwanted data sharing.

### DIFF
--- a/src/GPSStatus.h
+++ b/src/GPSStatus.h
@@ -57,8 +57,10 @@ class GPSStatus : public Status
 #endif
             meshtastic_NodeInfoLite *node = nodeDB.getMeshNode(nodeDB.getNodeNum());
             return node->position.latitude_i;
-        } else {
+        } else if (config.position.gps_mode == meshtastic_Config_PositionConfig_GpsMode_ENABLED) {
             return p.latitude_i;
+        } else {
+            return 0;
         }
     }
 
@@ -70,8 +72,10 @@ class GPSStatus : public Status
 #endif
             meshtastic_NodeInfoLite *node = nodeDB.getMeshNode(nodeDB.getNodeNum());
             return node->position.longitude_i;
-        } else {
+        } else if (config.position.gps_mode == meshtastic_Config_PositionConfig_GpsMode_ENABLED) {
             return p.longitude_i;
+        } else {
+            return 0;
         }
     }
 
@@ -83,25 +87,18 @@ class GPSStatus : public Status
 #endif
             meshtastic_NodeInfoLite *node = nodeDB.getMeshNode(nodeDB.getNodeNum());
             return node->position.altitude;
-        } else {
+        } else if (config.position.gps_mode == meshtastic_Config_PositionConfig_GpsMode_ENABLED) {
             return p.altitude;
+        } else {
+            return 0;
         }
     }
 
-    uint32_t getDOP() const
-    {
-        return p.PDOP;
-    }
+    uint32_t getDOP() const { return p.PDOP; }
 
-    uint32_t getHeading() const
-    {
-        return p.ground_track;
-    }
+    uint32_t getHeading() const { return p.ground_track; }
 
-    uint32_t getNumSatellites() const
-    {
-        return p.sats_in_view;
-    }
+    uint32_t getNumSatellites() const { return p.sats_in_view; }
 
     bool matches(const GPSStatus *newStatus) const
     {
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -1301,6 +1301,7 @@ bool GPS::whileIdle()
     }
     return isValid;
 }
+
 void GPS::enable()
 {
     enabled = true;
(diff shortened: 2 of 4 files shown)


garthvh:
This changes the meaning of a bunch of settings, there is no leak, someone sent a position to the node.

Gabrielerusso:
> This changes the meaning of a bunch of settings, there is no leak, someone sent a position to the node.

What does change? Everything works the same as before, just when the gps is manually disabled no more data is streamed. If a fixed position is set then nothing changes.

Gabrielerusso:
> there is no leak, someone sent a position to the node
The local node position is never deleted when location is disabled, unless you restart the board after disabling the location it will keep that.

Gabrielerusso:
> > > there is no leak, someone sent a position to the node
> > 
> > 
> > The local node position is never deleted when location is disabled, unless you restart the board after disabling the location it will keep that.
> 
> There is no disable location setting, that turns off the hardware GPS.

It's called "meshtastic_Config_PositionConfig_GpsMode_DISABLED" this does enable or disable location (position). GPS was just used to refer for the location, doesn't actually exist a disable option for every single GNSS.

garthvh:
> > > > there is no leak, someone sent a position to the node
> > > 
> > > 
> > > The local node position is never deleted when location is disabled, unless you restart the board after disabling the location it will keep that.
> > 
> > 
> > There is no disable location setting, that turns off the hardware GPS.
> 
> It's called "meshtastic_Config_PositionConfig_GpsMode_DISABLED" this does enable or disable location (position). GPS was just used to refer for the location, doesn't actually exist a disable option for every single GNSS.

It turns off the GPS module itself, you can still get positions from the phone, or broadcast the position already stored in the nodedb.  There is no "I sent a position to the device but I don't ever want it to ever be used" option currently, if we have a position it gets used.  It only every gets broadcast over the mesh if you have fixed position on.

thebentern:
Yes, GPS != Position module is where the confusion comes from

Gabrielerusso:
> > > > > there is no leak, someone sent a position to the node
> > > > 
> > > > 
> > > > The local node position is never deleted when location is disabled, unless you restart the board after disabling the location it will keep that.
> > > 
> > > 
> > > There is no disable location setting, that turns off the hardware GPS.
> > 
> > 
> > It's called "meshtastic_Config_PositionConfig_GpsMode_DISABLED" this does enable or disable location (position). GPS was just used to refer for the location, doesn't actually exist a disable option for every single GNSS.
> 
> It turns off the GPS module itself, you can still get positions from the phone, or broadcast the position already stored in the nodedb. There is no "I sent a position to the device but I don't ever want it to ever be used" option currently, if we have a position it gets used. It only every gets broadcast over the mesh if you have fixed position on.

If it was a setting to disable one by one every GNSS systems then i would agree that this would change the dynamics, but in our case the "GPS" name is just used to refer to the location, if there is an hardware capable of GLONASS or whatever it's enabled and disabled with the same flag.
I uderstand what do you mean, but i think this is too general to  be used in that way, most people call GPS the location services.

src/gps/GPS.cpp
```diff
@@ -1313,7 +1314,10 @@ int32_t GPS::disable()
     enabled = false;
     setInterval(INT32_MAX);
     setAwake(false);
-
+    // reset position data if location is disabled and not fixed
+    if (config.position.gps_mode == meshtastic_Config_PositionConfig_GpsMode_DISABLED && !config.position.fixed_position) {
+        positionModule->clearLastPosition();
```
Gabrielerusso, line 1319: I think this one is still valid even if the "gpsmode" is used for the hardware GNSS module, as when it's decativated there is no use for the previous data if "fixed position" is disabled.

src/GPSStatus.h
Gabrielerusso, line 1: all the changes on this file on the other hand could be omitted until there is a location option.

src/gps/GPS.cpp
garthvh: I would prefer a clean pull, probably doing this on nodedb reset.  @jp-bennett I think this could be related to the preventing a node from deleting itself from the nodedb, should we clear the last position from the current node when we reset the nodedb?

jp-bennett:
I'm working on a per-channel GPS setting that might be a better way to do this. It's going to be a setting that explicitly controls the sending of locations.

src/gps/GPS.cpp
Gabrielerusso: > Does that change apply when doing a NodeDB reset?

The change i did applies only when GPS::disable() is called.

Gabrielerusso:
> I'm working on a per-channel GPS setting that might be a better way to do this. It's going to be a setting that explicitly controls the sending of locations.



> I'm working on a per-channel GPS setting that might be a better way to do this. It's going to be a setting that explicitly controls the sending of locations.

I think it's better to start calling things by their names, otherwise a lot of confusion is generated, by "GPS settings" you mean the general position settings or the hardware GNSS module settings? 
If we use "GPS" to refer to the Hardware GNSS module and then we use "GPS" for the location service and then we use GPS for the signle GNSS mode of Hardware GNSS this becomes a nigthmare :D

PR #3216, opened 2024-02-12
Battery level with proportional filter and lookup table

Gabrielerusso:
Add a more precise battery level estimation based on a Open Circuit Voltage table.

A P controller acts as a low pass filter to smooth out the voltage readings and on that is applied a OCV table to estimate more precisely different kind of cells chemistry.

Now supported: 
1. LiFePO4 
2. Lead Acid
3. LiIon/LiPo
With a variable number of cells.

A valid voltage reading counter was added to avoid doing an ADC average on incorrect values.

Heltec V2.1 voltage multiplier was corrected.


code8buster (approved):
Voltage ranges look good. Could you modify any other variant.h files that might be using the existing BAT_FULLVOLT or other similar macros? I think variants/station-g1 might be using it

Gabrielerusso:
> Voltage ranges look good. Could you modify any other variant.h files that might be using the existing BAT_FULLVOLT or other similar macros? I think variants/station-g1 might be using it

Ok, added also Nimh/Alkaline cells that were used by another variant

> Run `trunk fmt` as well, to make the CI checks happy.

Done but some checks still don't pass, if i run trunk fmt won't it reformat every file?

code8buster:
Nicely done, I'll take a look at the trunk stuff tonight to see if I can't roundhouse kick it into working.

code8buster:
@thebentern OK should be ready for your discerning eye.

TBH, I'm not sure we'll ever see chargers built into future boards for any chemistry other than lithium ion but it's pretty cool to have some compile-time modularity for DIYers.

thebentern (approved):

shalberd:
Thank you, @thebentern and @Gabrielerusso 
Can we ever expect battery type from power.h, e.g. LTO, being user-configurable without having to compile own firmware by adding to variants/rak4631/variant.h:

`#define CELL_TYPE_LTO`

https://uart.cz/en/2557/lto-bms-development-notes/

https://uart.cz/en/2534/solar-mppt-charger-for-meshtastic/

PR #3227, opened 2024-02-15
Enable NMEA Messages on USB port for u-blox receivers

GPSFan:
Normally, it is a good idea to disable messages on unused ports.
Native Linux needs to be able to use GNSS modules connected via
via either serial or USB.
In the future I2C connections may be required, but are not enabled for now.


thebentern:
Would it be worthwhile to gate this to only native targets? I'm mostly wondering if there's a power usage penalty for enabling it.

GPSFan:
It was enabled when I started the rework project, I didn't consider the native linux requirement till last night when Jonathan Bennett asked about it, frankly I had forgotten I turned it off. On the Neo-6/7 and M8 modules there is a separate VDD_USB pin which is supposed to be grounded, and is whne usb is not used, like on the t-beams. I can do some tests on an M8 module with and without the NMEA messages coming out the USB and see what difference it makes.

GPSFan:
Done some power tests and I can't see any differences with data going out USB as well as serial. The Neo-6/7/M8 have a seperate VDD_USB pin which powers the USB guts, so on things like t-beams where that pin is grounded (as suggested by u-blox) the only additional power consumption is the additional cpu work to fill the usb buffer. In any case, I saw no differences.

thebentern (approved):

PR #3243, opened 2024-02-17
feat: implement StoreAndForward `lastRequest` map handling

andrekir:
simplified Store and Forward server side handling of each client node's 'last_request' using an unordered_map to store the last request for each nodeNum (`to` field).


andrekir:
closed in favor of #3246

Issue #3138, opened 2024-01-28
[Feature Request]: MQTT downlink without LoRa Tx and uplink local node only


milkpirate:
### Category

Other

### Hardware

Heltec V3

### Firmware Version

v2.2.17.dbac2b1

### Description

My minimal setup looks like so:

```mermaid
graph LR
    PC <-->|Serial| NodeA
    <--> |LORA| NodeB
    <--> |Bluetooth| Android

    NodeA <--> |MQTT| mosquitto
```

the `mosquitto` log looks like so (converted to YAML for better readability):

```yaml
---
# lora/2/json/home-prim/!NodeA
channel: 0
from: 145556388
id: 1057405219
payload:
  air_util_tx: 0.125111117959023
  battery_level: 101
  channel_utilization: 0
  voltage: 4.35400009155273
sender: '!NodeA'
timestamp: 1706448997
to: 4294967295
type: telemetry
```
Messages sent from `Android` to `NodeB` in the prim. channel, look like so:
```yaml
---
# lora/2/json/home-secd/!NodeA
channel: 1
from: 3806671628
id: 1421562879
payload:
  text: test-message
rssi: -60
sender: '!NodeA'
snr: 13.5
timestamp: 1706449977
to: 4294967295
type: text
```
**1. Problem**: The message sent from `NodeB` does not appear as beeing sent from `NodeB` instead from `NodeA`.

When I send a message in the "private" channel from `NodeB` to `NodeA`, the message appears as:
```yaml
---
# lora/2/json/home-prim/!NodeA
channel: 0
from: 145556388
id: 1057405208
sender: '!NodeA'
timestamp: 1706448349
to: 3806671628
type: ""
```
**2. Problem**:
- The message type is empty
- There is no text at all
- And the source (sender) is not `NodeB` 

Is that as intended? Since to me it does not seem to be desirable.

GUVWAF:
The "sender" field is always the node connected to MQTT (gateway). You'll need to look at the "from" field, which is the unique node number, which you can map to a name received in NodeInfo messages, see also: https://meshtastic.org/docs/software/mqtt/#json-topic

For 2: do you have encryption enabled? I believe if you disable that you should see the type and text.

milkpirate:
@GUVWAF do you mean encryption at all or the MQTT encryption:
![image](https://github.com/meshtastic/firmware/assets/9251021/26429616-b744-4fd3-b9a3-6241ba0ba364)
since this one is off.

[...]

PR #3245, opened 2024-02-18
Only publish MQTT on LoRa Tx if packet is from us and on Rx if not

GUVWAF:
Currently we're only publishing to MQTT when we're *transmitting* a packet via LoRa. Nevertheless, we still see *received* packets on MQTT as well, but this is only because we forced a node to always rebroadcast when MQTT is enabled:
https://github.com/meshtastic/firmware/blob/143ee9cdf61773f0d719a9162affd96427f8b3ad/src/mesh/FloodingRouter.cpp#L24 

However, as observed in #3138, direct messages towards the MQTT node will not be published, because we don't rebroadcast if we're the destination.

With this, we're publishing when transmitting via LoRa only when we're the original transmitter, and we publish when receiving when we're not the original transmitter. This way we can get rid of the "hack" where we're always rebroadcasting via LoRa when MQTT is enabled, and we'll see direct messages to the MQTT node as well. 

Since we already don't publish packets to MQTT if they came via MQTT, this will not create any extra traffic when bridging networks.

Also fixed a possible issue where `env->packet` contains the encrypted version of the packet and we're trying to print `decoded.portnum`.


src/mesh/Router.cpp
```diff
@@ -257,10 +257,9 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
             return encodeResult; // FIXME - this isn't a valid ErrorCode
         }
 
-        if (moduleConfig.mqtt.enabled) {
-            LOG_INFO("Should encrypt MQTT?: %d\n", moduleConfig.mqtt.encryption_enabled);
-            if (mqtt)
-                mqtt->onSend(*p, *p_decoded, chIndex);
+        // Only publish to MQTT if we're the original transmitter of the packet
+        if (moduleConfig.mqtt.enabled && p->from == nodeDB.getNodeNum() && mqtt) {
```
thebentern, line 261: Just confirming, we would still "rebroadcast" a message to MQTT received from someone else transmitting on lora?

GUVWAF: Yes, in this case it's handled upon receiving the packet via LoRa: 
https://github.com/meshtastic/firmware/blob/834490175d312dd7c4a2d6049185f030fffe85a3/src/mesh/Router.cpp#L454-L456

GUVWAF:
Converting to draft for now. I just realized we should filter received packets with `wasSeenRecently()` to avoid getting duplicates. Need to find a clean way to do this.

GUVWAF:
> I just realized we should filter received packets with `wasSeenRecently()` to avoid getting duplicates.

After further testing I found that this is not needed because it is already done in `shouldFilterReceived()` which is called in `Router::perhapsHandleReceived()` before `handleReceived()`, where we send to MQTT. 

This should be good to go then :)

PR #3248, opened 2024-02-19
Adds support for position precision settings

jp-bennett:


src/mesh/NodeDB.cpp
```diff
@@ -195,6 +195,13 @@ void NodeDB::installDefaultConfig()
     config.position.position_broadcast_smart_enabled = true;
     config.position.broadcast_smart_minimum_distance = 100;
     config.position.broadcast_smart_minimum_interval_secs = 30;
+
+    config.position.channel_precision_count = 8;
+    config.position.channel_precision[0] = 32;
```
thebentern, line 200: Technically, we shipped the protobufs themselves on the last release. Should we create a migration of sorts to init this for the first time?

jp-bennett: Ironically we've come to a bit different conclusion on how this should work. Will be dropping some of this work in favor of that approach.

caveman99:
@jp-bennett Can this be closed?

src/mesh/NodeDB.cpp
aaaidan: > Will be dropping some of this work in favor of that approach.

@jp-bennett can you say more about this? Has a precision masking feature already been implemented or rolled out?

jp-bennett: Yes, it's been in firmware for quite a while. We set the precision per channel.

aaaidan: Oh, that’s fantastic. Is this setting in all apps? I can’t find this in iOS, for example, but don’t know where to look.

I would love to share my general location I would just like to make sure my exact position is never transmitted without my intent.

PR #3256, opened 2024-02-21
E-Ink: additional conditions for "Dynamic Partial" mode

todd-herbert:
## Purpose
Offer more tools for "dynamic partial" code to find the best balance of full-refresh and partial-refresh. 
Lay groundwork for imminent pull-requests (Wireless Paper v1.1, sleep screens).

## Why
When using partial-refresh, some E-Ink displays struggle to "erase" black pixels *- to change an existing black pixel back to white*. Rather than returning to white, these "erased" pixels can remain a light shade of gray. This is seen as "ghosting" - a faint version of the previous image remains visible.

The prevalence of this effect varies from display to display.
While working to implement partial-refresh for Heltec Wireless Paper V1.1, it has become clear that "ghosting" is somewhat more of an issue than with the V1.0 board.

## What the pull-request implements

Two new "dynamic partial" controls are proposed here:

* #### `EINK_PARTIAL_ERASURE_LIMIT`
  Only active if defined in `variant.h`. Attempts to track the amount of "erased pixels" since last full-refresh. When this amount eventually exceeds `EINK_PARTIAL_ERASURE_LIMIT`, ghosting is considered likely to be noticeable, and a full-refresh is forced.

  The tracking of "erased pixels" is an approximation, by comparing total black pixel count from image to image. Theoretically, an accurate number could be achieved by comparing image buffers *pixel-to-pixel*, but this would be more intrusive to implement. 

* #### `demandFullRefresh()`
  Public method, suitable to call from a `FrameCallback`. Single shot, specifies that the next update is cosmetic (splash screens?), and *must* be drawn with full-refresh, to maximize image quality. **No calls to `demandFullRefresh()` are proposed in this specific pull-request**

## Performance
Overhead for entire "dynamic partial" code (incl. proposed changes):
  * typically <10µS per loop
  * 11ms for all checks to pass, approve partial refresh. (Including `LOG_DEBUG()` calls)

*\*Benchmarked on Heltec Wireless Paper*

### Demonstration of `EINK_PARTIAL_ERASURE_LIMIT`
  Highlights a scenario where "erasure limit" applies
 
https://github.com/meshtastic/firmware/assets/24772776/e9203bb4-8b64-4e08-866e-972615764706

## Flow Diagram for `determineRefreshMode()`
*This section edited to correct error in diagram.*
flow diagram for determineRefreshMode()

### DIFF
--- a/src/graphics/EInkDisplay2.cpp
+++ b/src/graphics/EInkDisplay2.cpp
@@ -359,6 +359,12 @@ void EInkDisplay::lowPriority()
     isHighPriority = false;
 }
 
+// Full-refresh is explicitly requested for next one update - no skipping please
+void EInkDisplay::demandFullRefresh()
+{
+    demandingFull = true;
+}
+
 // configure display for partial-refresh
 void EInkDisplay::configForPartialRefresh()
 {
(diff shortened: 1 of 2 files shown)


thebentern (approved):

markbirss (approved):

src/graphics/EInkDisplay2.cpp
markbirss, line 1: quite neat to count the amount of black pixels

PR #3259, opened 2024-02-21
fancy T-Deck / SenseCAP Indicator / unPhone / PICOmputer-S3 TFT screen

mverch67:
nothing to see here, yet. Work in progress


bobricius:
Is possible to test on PICOmputer S3 ? I can send sample hardware.

mverch67:
> Is possible to test on PICOmputer S3 ? I can send sample hardware.

To Germany? Good. I can support it when I have the actual hardware available.
Currently (as of right now) it won't function as there is no cardKB-like keyboard input driver implemented which would be the precondition for any interaction, as the PICOmputer does not have a touch screen.

bobricius:
> > Is possible to test on PICOmputer S3 ? I can send sample hardware.
> 
> To Germany? Good. I can support it when I have the actual hardware available. Currently (as of right now) it won't function as there is no cardKB-like keyboard input driver implemented which would be the precondition for any interaction, as the PICOmputer does not have a touch screen.

worldwide, Germany is almost a neighbor, I'm from Slovakia. Picomputer use direct gpio matrix keyboard implemented by @caveman99  cpu is bpi-PicoW-S3 + RFM95 send my your postage address to bobricius@gmail.com

caveman99:
I am on the road right now but iirc the picomputer does only have 4MB flash. May be a tight fit. The TFT tracksenger  may be a better fit. 8 MB is the min for an integrated device I would think.

@mverch I have a picomputer prototype I can send you. Assuming @bobricius doesn't have a newer revision :-)

bobricius:
I ma not sure but on product page is: 
ESP32-S3R2
2MB Quad SPI PSRAM 
8MB Quad SPI Flash
Also I have Picomputer based on Wroom module which is more flexible with flash/psram options

esputer

mverch67:
> I ma not sure but on product page is: ESP32-S3R2 2MB Quad SPI PSRAM 8MB Quad SPI Flash Also I have Picomputer based on Wroom module which is more flexible with flash/psram options

That's also what I read on the ALI-Express page for the Banana-Pi. Should be fine.

mverch67:
PicomputerS3b

variants/picomputer-s3/platformio.ini
```diff
@@ -27,6 +27,7 @@ build_flags = ${esp32_base.build_flags}
 	-D PICOMPUTER_S3
   -D MESHTASTIC_EXCLUDE_CANNEDMESSAGES=1
   -D MESHTASTIC_EXCLUDE_INPUTBROKER=1
+  -D MESHTASTIC_EXCLUDE_BLUETOOTH=1
```
ChrDel, line 30: Will there be an option on device to turn on/off the Bluetooth? 
The screen is hard to impossible to read at times and the only way to be able to do anything in that case is to connect to a phone with a better screen...
🙏

mverch67: Currently, you can connect a phone using WLAN. That should work.

ChrDel: Wifi is not supported on iOS :(

mverch67: It works with Android. Another option is to use the admin channel for configuration. If you can't read the screen because of some eye related issues I think this device is not the right one for you. Instead get a LoRa device without screen (e.g. a RAK19003+RAK4631, they have huge discount at the moment) and use your phone to operate it.

ChrDel: Yeah it's always what Android can do... I'm using iOS though... 
This device is great, especially when the 3 days of darkness (no electricity, not internet...) will be upon us...
All I was asking is whether you would add an option so ppl can chose to use BLE or not since I see you are completely disabling it and it was working before... 
I have a battery with my T-Deck that can get me 2 weeks with BLE on...

mverch67: There is a limitation in the firmware that let's either use BLE app or the TFT app but not both at the same time, and that is not an easy fix.

Kangie:
This is looking great. I've been running it on hardware on and off for a few weeks and even in the current state I'd say it's made the T-Deck far more usable (as a standalone device) than it was before. I'm very impressed with the progress so far and look forward to seeing the improvements.

Willing volunteer if you need anything tested or want specific feedback.

martinbogo:
Downlaoded and installed latest CI artifacts.  Will report back with any issues/bugs I find.

ulrichard:
Great stuff!!!   thanks for all the work. 
A while ago I installed the firmware from https://github.com/meshtastic/firmware/actions/runs/10327954038 on one of my T-Decks.  I loved it!
Then I wanted to upgrade to a newer version, as the branch seems to be active and maintained. But none of the recent firmwares from this branch work on my device. Which one is the last known good revision?

martinbogo:
SAME question!  I tried installing the artifact from the continuous
integration ... no go on T-Deck.  Would love to track progress by "driving"
the firmware from time to time.

caveman99:
> Latest CI run - no t-deck artifact/zip was created...

CI will not build these images any more. You need to compile yourself. Circulation of these Tech Previews has caused a considerable amount of support requests. There will be an announcement if wider testing is available and required.

ndoo:
Hi there, is https://github.com/meshtastic/firmware/pull/3259/commits/f184ba53b76852523f5d67b4411a599016d4ebc6 a commit that should be cherry-picked to the main branch?

It seems like settings being lost on reboot is still occasionally a problem and not just on the TFT GUI branch.

src/mesh/NodeDB.cpp
```diff
@@ -846,6 +855,12 @@ void NodeDB::loadFromDisk()
         }
     }
 
+    state = loadProto(uiconfigFileName, meshtastic_DeviceUIConfig_size, sizeof(meshtastic_DeviceUIConfig),
```
fifieldt, line 858: Does this need to be wrapped in some kind of exclude, for nodes not using the new UI?

mverch67: No. We don't know at compile time if the node has connected a UI or not.

fifieldt: Thanks!

zpatten:
@mverch67 Thanks for all your work on this.  Was wondering if there could be an update on the most recent build artifacts?  I have several t-decks and while I realize this is still very alpha and technically not supported I would very much like to experiment with this unfortunately trying to dig out newer build artifacts I have been unable to successfully flash this on any device and get anything other than a black screen.  

The build artifacts I could find don't seem to have a matching release as I understand you need to flash a matching release first then this into the other slot; that being said I'm only going off what I have read elsewhere.  Is it possible to get an update with some links to some recent artifacts that can be flashed onto the t-deck?  Alternatively I do have a development environment setup for meshtastic so should I just attempt to compile this from there?  

Mainly looking for some direction on how to get the most recent version of this onto one of my t-decks to experiment with as mentioned.  Again appreciate all your hard work here.

martinbogo:
There are no artifacts on purpose.

Just download the build environment and check out the latest commits to
test.  That is The Way(tm) and it prevents people from overwhelming GitHub
with issues before there is a real Beta that’s ready.

zpatten:
@martinbogo Awesome thanks, I had a feeling; definitely don't want to be creating issues here around this since it's so early.  I already have a meshtastic dev environment setup so shouldn't be an issue.  Was my understanding correct that you need two copies of the firmware (a base image and the fancy ui) to flash based on the other instructions I've read or is that incorrect?

martinbogo:
It will flash directly using the build tools in VSCode … but the
ui/interface is far from ready for daily driving.

It also disables Bluetooth, as the mechanism the UI uses requires the same
resource to render the interface.

zpatten:
> It will flash directly using the build tools in VSCode … but the ui/interface is far from ready for daily driving. It also disables Bluetooth, as the mechanism the UI uses requires the same resource to render the interface.

Awesome, thanks for the clarification!  I'm aware of the Bluetooth and some of the other issues.  Just wanting to experiment.  I have other devices dedicated to daily driving.

mariotti:
@mverch67 , by the way, I am flashing this PR to a t-deck weekly.
The last commit about few hours ago, it did not brick it :)

Please let me know (weekly ;) ) if I can support, sadly it is a pretty common device.

mverch67:
> Please let me know (weekly ;) ) if I can support, sadly it is a pretty common device.

Hi Mariotti, your support is very welcome, thanks a lot ;-)

Hope, you enjoy the latest updates.
Cheers

ndoo:
Feature request: Show distance in the node list environment telemetry.

If distance is >0. Thank you for your consideration :)

I had a look at the code for 320x480 but didn't know if it was straightforward as to add a _tm3 variable.

mverch67:
> Feature request: Show distance in the node list environment telemetry.
The distance to other nodes is already shown in the node list once the own position is known and updated.

ndoo:
> > Feature request: Show distance in the node list environment telemetry.
> 
> The distance to other nodes is already shown in the node list once the own position is known and updated.

Hi there,

This is referring to the distance value in Environment telemetry protobuf, not distance between nodes - i.e. the distance measured by an ultrasonic sensor

mariotti:
Hoi Manuel,

So I fleshed successfully on few t-deck. And tested a bit. Here my notes.

 - First we need to pull always the submodule device-ui master, its not an issue, but we must remember it. As device-ui is basically master driven, I was checking if a submodule can alway point to master. There are solutions but I could not find any that is very clean. (this note is probably for others that wants to jump in).

 - Then there is a little bug when the device is "remotely" administered.
For example here I change the "role" in the UI for a remotely administered device, i.e. it should not be possible:

Notification_Center

We see the rebooting... overlay but the device is not rebooting [expected], The UI is ignoring the settings changes [expected], but the overlay cannot be removed:

Photo_-_Google_Photos

The device is working as expected, ignoring admin changes, as it is remotely administered, but the [rebooting...] is still on screen.

At present only on/off cycle removed the banner.

Then,

Is this the kind of thinks you expects from me? ;)

I am getting more and more into the code, but it will take me a bit more time to be able to address these issues to code lines. Working on it ;)

ahh a last note. Which I am not sure about.

The UI usability seems heavily dependent on the power. If you set this:

Meshtastic_Web

The usability seems laggy, odd, slow. I found it out by flashing on a t-deck plus, which was using a lot of power, but the UI was working incredibly better then my previous bare t-deck.

I guess it is a known issue. But I would put this ahead as we (you/meshtastic) will get a lot of UI usability complains if users run on very low power.

btw, I am almost done with my first project with this UI :)

Cheers
A+
F

mverch67:
Hi, thanks for your tests. I don't know what version you are using as there are constant updates but here is a quick answer to your findings:

_> We see the rebooting... overlay but the device is not rebooting [expected], The UI is ignoring the settings changes [expected], but the overlay cannot be removed:_

This is probably not a UI issue. The firmware must be sending a reboot message to the UI otherwise it would not be shown.

_> At present only on/off cycle removed the banner._
Long press the Home button would do a quick resync and remove it, too.

_> The UI usability seems heavily dependent on the power. If you set this:_
I don't quite understand what you mean. If you mean the power saving options then these should not be modified as it is set to its best values already (on initial factory reset with the UI). If you mean the battery power level then this also should also not impact the device. 
The only thing that impacts the UI are CPU heavy tasks or things blocking the UI such as a high number of nodes (>200) or enabled Wifi or constant GPS auto scans when no GPS is present but the switch is turned on.

Cheers,
Manuel

mariotti:
Hoi,

Quick answer: If I write here, the version will be the last commit. Unless we are very unlucky in timing :) .

I'll double check for the [rebooting ...], but it happens only if the device is "remotely administered", it does not happen in other case. Just to be sure.

I'll try the long press tip.

A+
F

mariotti:
Hoi, yes, the rebooting might be a firmware issue.

But in the sense that the UI gets no feedback. I will check first in the old UI how this is handled.

In practice this UI assumes, for all events, after reboot notification, that the device will actually reboot.
If the device is remotely administered the reboot is not done.

I am really poor in C++ code but for example:

``
                THIS->controller->sendConfig(meshtastic_Config_DeviceConfig{device}, THIS->ownNode);
                THIS->notifyReboot(true);
``

Assumes that a reboot will happen. Or?

.gitmodules
```diff
@@ -1,6 +1,9 @@
 [submodule "protobufs"]
 	path = protobufs
 	url = https://github.com/meshtastic/protobufs.git
+[submodule "lib/device-ui"]
+	path = lib/device-ui
+	url = https://github.com/meshtastic/device-ui.git
```
j6k4m8, line 6:
```suggestion
	url = https://github.com/meshtastic/device-ui.git
    branch = master
```

j6k4m8: @mariotti:
> As device-ui is basically master driven, I was checking if a submodule can alway point to master. [[comment](https://github.com/meshtastic/firmware/pull/3259#issuecomment-2526297841)]

I believe this solves this particular complaint! But as far as I know you cannot specify a commit hash or tag, so this is good for keeping the HEAD of master but not going deeper. 
@mverch67 I will make a PR — you are welcome to use or reject!

mverch67:
> Assumes that a reboot will happen. Or?

Yes, exactly. This is to notify the user that the device will restart in a few seconds. UI and firmware have implemented the same kind of knowledge which command leads to a restart.

The reboot notification that is issued by the firmware indicates to the UI/App that the device is going to reboot. If it send this message but doesn't reboot then this is clearly a bug in the firmware. Some remote administration will lead to a reboot, some does not. There is no general rule, but only a case-by-case decision.

.gitmodules
mverch67: I don't think this is a good idea in general as it won't be possible to ever re-build old versions of the software. The branch tag is currently not maintened as there are no automitic builds but will be as soon as the last needed feature is running.

mariotti: Nice! But sadly it will work for a short time :(

If tomorrow we will start versioning the device-ui, which will be the idea, then we have an issue.

I think it is best to leave this to the developers until it is merged.

mariotti:
Got what you mean, then it is a FW issue.

mariotti:
Ok, then I will test the behaviour for the latest alpha in the standard UI, and start a ticket/issue on that.
cheers

mariotti:
@mverch67 Hoi, I kept flashing t-deck's :) But I have some issues with the touchscreen that got very laggy. I am not sure it actually depends on this PR, yet it seems unresponsive only with the device-ui. It can even be hardware, but I got 2 t-deck with the issue, so I am really confused. I flashed also old commits.

Do you mind if I open a discussion under: https://github.com/meshtastic/firmware/discussions for my issue and maybe generally for other flashers? Or: do you prefer less noise while developing? Or the opposite, I should report here?

Cheers
A+
F

mariotti:
Hoi Manuel,

Indeed about the "rebooting...", [https://github.com/meshtastic/firmware/pull/3259#issuecomment-2526297841] I had a look at the code and it looks like that all is passing by the AdminModule.

In the admin module you have something like this:

```
    if (messageIsResponse(r)) {
        LOG_DEBUG("Allow admin response message");
    } else if (mp.from == 0) {
        if (config.security.is_managed) {
            LOG_INFO("Ignore local admin payload because is_managed");
            return handled;
        }
```

Now at: https://github.com/meshtastic/firmware/blob/master/src/modules/AdminModule.cpp#L79

So it bails you out immediately if the node is managed. "if (config.security.is_managed)"

I assume that the UI goes also via the AdminModule too, which is auspicabile.

In that case a reboot is not going to take place.

We might need to import some security settings and apply the same behaviour.

If I got it right, of course.

A+
F

ndoo:
> @mverch67 Hoi, I kept flashing t-deck's :) But I have some issues with the touchscreen that got very laggy. I am not sure it actually depends on this PR, yet it seems unresponsive only with the device-ui. It can even be hardware, but I got 2 t-deck with the issue, so I am really confused. I flashed also old commits.
> 
> Do you mind if I open a discussion under: https://github.com/meshtastic/firmware/discussions for my issue and maybe generally for other flashers? Or: do you prefer less noise while developing? Or the opposite, I should report here?
> 
> Cheers A+ F

Early T-Deck Plus units have [a known issue of an almost unresponsive touchscreen](https://github.com/Xinyuan-LilyGO/T-Deck/issues/42) due to incorrect programming of the GT911 touchscreen IC at the factory. Flashing it with the [UnitTest firmware](https://github.com/Xinyuan-LilyGO/T-Deck/blob/master/firmware/T-Deck_UnitTest_240913.bin) and then back to the TFT GUI firmware will fix it.

mariotti:
Thanks both, the t-Deck-plus is back in track :)

JIC, It was not that obvious, as I knew that repo, but not the actual issue so initially I did not try too hard.
In addition from a Mac a simple esptool.py was not working :( 
What worked was flashing from https://espressif.github.io/esptool-js/ and using this .bin https://github.com/Xinyuan-LilyGO/T-Deck/blob/master/firmware/T-Deck-Plus-TouchFix_241025.bin (not sure the UnitTest, as building it did not work).
I will try next week to reproduce it on my second t-Deck-Plus and create a discussion to describe the steps.

Flashed successfully, basic usage working. This on the last commit with both:  current pointer to device-ui and device-ui updated to the last current master commit.

Going to flash the other t-decks :)

Thanks again
A+
F

mariotti:
[t-deck]
Hoi, JIC, I just flashed commit 5554114
https://github.com/meshtastic/firmware/pull/3259/commits/5554114d11c938cda352b1e72d6a964ccded7a32
with current link to device-ui (just the usual array resize fix) and with device-ui master. 
Both fine with basic navigation working.

[t-deck-plus]
At commit: 66101bc
https://github.com/meshtastic/firmware/pull/3259/commits/66101bc9f758e462b64d05b84100d51fa3c7403b
(with device-ui at master)
I gave 2 t-deck-plus to my daughter and her friend, both 9 yo. They used it for a week (few hundred meters away).
I set up 1 single encrypted channel. They had no issue. I just told them never click on settings and on the home button.
This UI is kids friendly and kids proof!

In the next year I will have a bit of time for more specific tests. For the moment: Happy New Year!!

arch/nrf52/nrf52.ini
```diff
@@ -4,7 +4,7 @@ platform = platformio/nordicnrf52@^10.7.0
 extends = arduino_base
 platform_packages =
   ; our custom Git version until they merge our PR
-  framework-arduinoadafruitnrf52 @ https://github.com/geeksville/Adafruit_nRF52_Arduino.git
+  framework-arduinoadafruitnrf52 @ https://github.com/geeksville/Adafruit_nRF52_Arduino.git#4f591d0f71f75e5128fab9dc42ac72f1696cf89f
   toolchain-gccarmnoneeabi@~1.90301.0
```
mariotti, line 8: I think this can go to master pretty soon. So some clean up.
Do we still need this reference? Or can we go with the latest?

platformio.ini
```diff
@@ -123,7 +123,7 @@ lib_deps =
 
 [radiolib_base]
 lib_deps =
-	jgromes/RadioLib@7.1.0
+	https://github.com/jgromes/RadioLib.git#b2b4c9e0f73bd9d5a435b291ee27574ee9c2f69f
```
mariotti, line 126: As above: do we need the specific reference?

src/main.cpp
```diff
@@ -730,11 +742,92 @@ void setup()
...
+    }
+#else
+    deviceScreen = &DeviceScreen::create();
+    PacketAPI::create(PacketServer::init());
+    deviceScreen->init(new PacketClient);
+#endif
 #endif
 
     // Initialize the screen first so we can show the logo while we start up everything else.
+#if HAS_SCREEN
```
mariotti, line 828: Can this be pushed to master right away?

src/mesh/eth/ethClient.cpp
```diff
@@ -66,8 +66,9 @@ static int32_t reconnectETH()
                 syslog.enable();
             }
 
-            // initWebServer();
+#if !MESHTASTIC_EXCLUDE_SOCKETAPI
```
mariotti, line 69: This is also a change that is just master dependent. Can we push this?

src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -106,7 +106,9 @@ static void onNetworkConnected()
 #if defined(ARCH_ESP32) && !MESHTASTIC_EXCLUDE_WEBSERVER
         initWebServer();
 #endif
+#if !MESHTASTIC_EXCLUDE_SOCKETAPI
```
mariotti, line 109: One more master change that has little to do with the UI.

src/main.cpp
fifieldt: Yeah, I wasn't sure about this one. Wanted to test if it broke any non-TFT devices.

platformio.ini
fifieldt: Yes, there's a need for a fix that isn't released yet

src/mesh/eth/ethClient.cpp
fifieldt: What about devices that run a web server on Ethernet right now?

mverch67: @caveman99

fifieldt (approved):

platformio.ini
mariotti: mmmm, they released a new version (7.1.1) but this dependency seems not yet there. It looks like an important change for devices waking up just an external feature. This might indeed affect all devices, not only the once with a UI.

mverch67: This radioLib commit fixed the annoying crash you probably saw on the T-Deck. Need to wait for a new RadioLib release (7.1.2 or 7.2) to get rid of the commit reference.

src/mesh/eth/ethClient.cpp
fifieldt: This is fine -- I misread, the commented webserver line was already in master.

src/mesh/wifi/WiFiAPClient.cpp
fifieldt: https://github.com/meshtastic/firmware/pull/5729

src/mesh/eth/ethClient.cpp
fifieldt: https://github.com/meshtastic/firmware/pull/5729

platformio.ini
mariotti: We have it :)

https://github.com/jgromes/RadioLib/releases/tag/7.1.2

See also:

https://github.com/jgromes/RadioLib/discussions/1364

As I am not an expert here I let you doing the updates :(

mariotti:
Sorry I was a bit absent. But:

I flashed this PR to t-deck and t-echo and all fine.
(only with current device-ui master: i.e. I checked out master from the submodule)

What I tested (basic):
 - it flashes with no errors
 - it receives messages
 - I can send messages (*)
 

The procedure for both:
 - flash the latest alpha release from web, to check it works with latest master (tests above)
 - flash with this PR (no -tft), same above tests
 
Procedure for t-deck:
 - flash with env t-deck-tft and same tests above.
 - little playing with the UI, i.e. verify only the main menus information. DID NOT change settings from the UI.

[I tested previous commits a bit more about the UI, so I kept this short]

My opinion (but see also questions below):

We should merge asap.

This PR is a firmware support for the device-ui lib, and currently it seems more like checking master.
To reduce work, I would push to master and allow others to test different hardwares.
This will also open to others some nice API. This also needs to be tested on master for devices
and use cases.

From the pure technical point we solved the issue with RadioLibs. Not Tested, indeed.

There is the comment from @fifieldt on main.cpp which we might need to resolve,
I am not sure if the t-echo with e-paper solves the issue. :) 

And the not in sync process for remote administered devices.

My questions:

If we merge this, how do we handle dev "tft" device-ui lib dependencies?

The UI is going to get out resources: for example Bluetooth, or anyway something else.
How are we going to handle it? Two versions of firmware for each device?
But most important: 2 UIs? t-deck and t-deck-tft?

I will be happy with one version. I do not get a device with TFT if I am not able to pay the costs.

I really hope we can merge this asap.

mariotti (approved):
Fantastic!

mariotti:
Sorry the (*) above.
For sending messages, which is indeed important.
t-deck tested only with the TFT and UI.
t-echo tested only via bluetooth from a phone

mariotti:
@ChrDel @j6k4m8 What do you think? I'd push to master asap, otherwise it is not possible to support every device. Or better this PR might diverge or go in parallel to master, wasting time of good coders.

j6k4m8:
Don't know how much my opinion matters here — I've only done a little work here and have only flashed/run on a T-Deck! But I'm definitely very supportive of merging to master to help coordinate dev efforts, I imagine that we can all be more helpful to @mverch67 who is doing some extraordinary work here!

mariotti:
@j6k4m8 but you can approve the PR to show some love ;)

j6k4m8:
can't argue with that :)

mariotti:
@fifieldt I can test some other devices but at this point I'd like to do it from master.

This PR did not break any of my devices. All lilygo based, but apparently @mverch67
was developing on some more basic processors.

fifieldt:
Thanks @mariotti , @j6k4m8 . I've been trying to keep master up to date by cherry-picking anything from this branch that can go into master cleanly. Where I think we're at is device-ui (or at least parts of it, like the PacketAPI) need to go in before we'd be able to merge this. I'm not sure what the plan is for that, to be honest!

mariotti:
@mverch67 you are invited to push the button and merge!

I think everybody will thank you for this work!

But do not get sad, you will see me in the UI repo :)

mariotti:
Hoi @fifieldt the only issue I really have is how to handle the versions with UI or without.

But I think this cannot be decided in this PR anyway. This is a requirement so that we can ask the right question ;)

caveman99:
@mariotti please stop spamming the comment section of this PR. The core devs pretty much know how and when to merge this. Everytime you post here or 'approve this' all devs get an email.

arch/nrf52/nrf52.ini
mverch67: I think this commit reference is to guarantee that the latest commit from that fork is pulled in. Otherwise platformio would just use the version it finds in its cache.
What I can't answer is why Adafruit littlefs version is 1.7.2 while the ARMmbed littlefs github version is at 2.10.1

fifieldt: looks like master just updated to use our own fork: https://github.com/meshtastic/firmware/pull/5924/files . let's try switching to that :)

kabili207:
Just a small heads up, the commit referenced by the `device-ui` submodule is extremely out of date (currently pointing to meshtastic/device-ui@6b760f5373b91de1d8234922a446f7232d5247d0) and this PR will not build unless you manually checkout the submodule's master branch

PR #3263, opened 2024-02-22
Adds position precision support

jp-bennett:


GUVWAF:
Since the firmware will just send a packet coming from the client in `MeshService::handleToRadio()` without modifying the payload, clients are still allowed to let the device send a position packet with arbitrary position which they populate themselves. Do we want to enforce this in the firmware or is it the client's responsibility to use the correct precision for each channel?

jp-bennett:
> Do we want to enforce this in the firmware

Hmmm. I'm inclined to think we let a client send what it wants. Mainly because implementing this sort of firewall on firmware would be a pain.

PR #3272, opened 2024-02-24
Update UNSET Region Screen

rcarteraz:
Removes flasher as it's no longer an option, and replaces with web. Also pluralizes client.


GUVWAF:
Looks good.
While we're at it, maybe we can also replace the `warning: node_db_lite full! erasing oldest entry\n` [message](https://github.com/meshtastic/firmware/blob/b2ea1e23be49ddc63b7032531572c656d559bcb0/src/mesh/NodeDB.cpp#L906) to something like `Warn: node database full!\nErasing oldest entry\n`, which makes more sense to the average user and fits on smaller screens.

PR #3276, opened 2024-02-24
update node db lite log message

rcarteraz:
Replaces the `warning: node_db_lite full! erasing oldest entry\n`  with `Warn: node database full!\nErasing oldest entry\n` per the request from @GUVWAF here https://github.com/meshtastic/firmware/pull/3272#issuecomment-1962302749


GUVWAF (approved):
Great!

PR #3277, opened 2024-02-24
Ignore MQTT by default if region has a duty cycle limit

GUVWAF:
As per a request of Discord user drewzh, by default avoid relaying packets sent via MQTT in duty cycle limited regions.


TomBayne:
Makes a lot of sense for regions like the UK where the mesh is getting crowded, and the change to the code is very simple - has my vote.

caveman99 (approved):

Issue #3284, opened 2024-02-25
[Feature Request]: Add UIFrame for PaxCounter module


thoherr:
### Platform

ESP32

### Description

Currently the PaxCounter module added by @caveman99 only sends the gathered information via the mesh.
A display of the current counter (WiFi/BLE/uptime) via an UI frame for the module would be useful.

I already implemented a prototype for this by adding the implementation of method drawFrame() and enabling wantUIFrame(), tested on my LilyGo LoRa32 V2.1_1.6.

caveman99:
is this a local display (the sending node) or remote, like with the environment telemetry?

Either way, please open up a PR, i'd gladly review.

thoherr:
> is this a local display (the sending node) or remote, like with the environment telemetry?

currently only local display, i'm just getting into this stuff.
a remote display would certainly make sense as well, sould be the next step.

i'll open a PR for my changes, thanks for considering it.

PR #3285, opened 2024-02-25
add UI frame for PaxCounter module

thoherr:
## add UI frame for PaxCounter module

- implements #3284 
- additional non related change: had to limit upload speed to 11200 for my board

### DIFF
--- a/variants/tlora_v2_1_16/platformio.ini
+++ b/variants/tlora_v2_1_16/platformio.ini
@@ -4,4 +4,3 @@ board = ttgo-lora32-v21
 build_flags = 
   ${esp32_base.build_flags} -D TLORA_V2_1_16 -I variants/tlora_v2_1_16
   -DGPS_POWER_TOGGLE ; comment this line to disable triple press function on the user button to turn off gps entirely.
-upload_speed = 115200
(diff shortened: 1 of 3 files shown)


variants/tlora_v2_1_16/platformio.ini
```diff
@@ -3,4 +3,5 @@ extends = esp32_base
 board = ttgo-lora32-v21
 build_flags = 
   ${esp32_base.build_flags} -D TLORA_V2_1_16 -I variants/tlora_v2_1_16
-  -DGPS_POWER_TOGGLE ; comment this line to disable triple press function on the user button to turn off gps entirely.
\ No newline at end of file
+  -DGPS_POWER_TOGGLE ; comment this line to disable triple press function on the user button to turn off gps entirely.
+upload_speed = 115200
```
caveman99, line 7: This seems unrelated to functionality?

caveman99 (approved):

variants/tlora_v2_1_16/platformio.ini
thoherr: Yes, see my second bulletpoint in PR description. That's only neccessary to upload firmware to the board. since some other variants have this settings in their platformio.ini, i included it in the PR

thebentern: This is not true of every setup though. Some of us have CP2102 based TLoRA v2 1.6 which flash fine at 921600 baud

thoherr: Reverted the commit with this change, since it is completely unrelated for the feature and as @thebentern mentioned not valid for all setups

thebentern (approved):

PR #3304, opened 2024-02-27
Revert "Fix LED pinout for T-Echo board marked v1.0"

thebentern:
Try-fix T-Echo boot loop


andrekir:
LiliGo's GitHub mentions two revisions for the T-Echo: https://github.com/Xinyuan-LilyGO/T-Echo/blob/57dc0e69c45bf35950ce1d7898783c7b31d04295/examples/GPS/utilities.h#L81-L89

```c
#if defined(VERSION_1)
#define GreenLed_Pin        _PINNUM(0,13)
#define RedLed_Pin          _PINNUM(0,14)
#define BlueLed_Pin         _PINNUM(0,15)
#else
#define GreenLed_Pin        _PINNUM(1,1)
#define RedLed_Pin          _PINNUM(1,3)
#define BlueLed_Pin         _PINNUM(0,14)
#endif
```

thebentern:
> LiliGo's GitHub mentions two revisions for the T-Echo: https://github.com/Xinyuan-LilyGO/T-Echo/blob/57dc0e69c45bf35950ce1d7898783c7b31d04295/examples/GPS/utilities.h#L81-L89
> 
> ```c
> #if defined(VERSION_1)
> #define GreenLed_Pin        _PINNUM(0,13)
> #define RedLed_Pin          _PINNUM(0,14)
> #define BlueLed_Pin         _PINNUM(0,15)
> #else
> #define GreenLed_Pin        _PINNUM(1,1)
> #define RedLed_Pin          _PINNUM(1,3)
> #define BlueLed_Pin         _PINNUM(0,14)
> #endif
> ```

Interesting. Think we should continue to revert these changes, since the revision pin macros are causing boot-loop in a sub-set of T-Echos? This started in 2.2.18.

PR #3308, opened 2024-02-28
Don't send anybody to null island ever

### LINKED ISSUE #3228: [Bug]: When GPS lock is lost, position is still reported over MQTT with lat and long of 0,0.
### Category

Other

### Hardware

T-Beam

### Firmware Version

2.2.21.7f7c5cb

### Description

When a device acting as an MQTT gateway sends a position report for a node with a satellite fix, it looks as follows (JSON):

```
{
  PDOP: 191,
  altitude: 187,
  ground_speed: 1,
  latitude_i: REDACTED
  longitude_i: REDACTED
  sats_in_view: 8,
  time: 1708024383
}
```

However, if the device loses it's fix, reports are still periodically sent which look like this:

```
{
  latitude_i: 0,
  longitude_i: 0,
  time: 1708021680
}
```

I have not investigated protobuf output, only JSON at present, but the MQTT.cpp source file looks like it's doing the right thing as far as not including fields that aren't present in the protobuf.

This is causing some quality-of-life issues on the development end, as there isn't an immediate indication that the position is invalid, and in fact, 0, 0 certainly is a valid position, though a remote one unlikely but not impossible to be visited by mesh nodes.

My immediate thought for a solution would be to either suppress reports that do not have a GPS fix, or add a field to the payload to indicate that there is no GPS fix.  The device knows there is no fix, as it indicates "No Sats" on screen 3 and "No GPS Lock" on screen 4.

I don't want to simply filter out results of 0, 0, or those without the `sats_in_view` key, as both of those could be legitimate reports from some current or future device.  The issue is that the GPS fix status, which the device does know, is not being provided -- at least to MQTT in the JSON package.

### Relevant log output

_No response_

thebentern:
Closes #3228 and fixes the sending of zero lat/lon positions in general

### DIFF
--- a/src/modules/PositionModule.cpp
+++ b/src/modules/PositionModule.cpp
@@ -176,6 +176,11 @@ meshtastic_MeshPacket *PositionModule::allocReply()
         LOG_INFO("Providing time to mesh %u\n", p.time);
     }
 
+    if (p.latitude_i == 0 && p.longitude_i == 0) {
+        LOG_WARN("Skipping position send because lat/lon are zero!\n");
+        return nullptr;
+    }
+
     LOG_INFO("Position reply: time=%i, latI=%i, lonI=%i\n", p.time, p.latitude_i, p.longitude_i);
 
     // TAK Tracker devices should send their position in a TAK packet over the ATAK port


jp-bennett (approved):
Yeah, this makes sense. Send it!

PR #3317, opened 2024-03-02
Remove problematic IO2 3V3 toggling (again) as GPS EN pin

### LINKED ISSUE #3128: [Bug]: GPS power saving causing unintentional Ethernet module shutdown
### Category

Other

### Hardware

Rak4631

### Firmware Version

2.2.19.8f6a283

### Description

When setting "Device GPS Enabled" in the iOS app settings to FALSE, the firmware will attempt to probe for a GPS module on startup, eventually fail, and then decide to shut down power to the IO rail, causing ALL, not just the GPS device to become inactive. In my case, this is the active Ethernet module.

An observation (and my current workaround as I don't have or want a GPS module for my base station) is setting "Device GPS Enabled" to TRUE, which results in the same probing attempts (trying different baud rates before giving up), but the logic decides to not go into power saving.

I understand from speaking to devs on Discord that the power rail shutdown is a solution to conserving power by hard-cutting the power to the GPS module, but I think this is a bit of a short-sighted approach without first checking the presence of anything else requiring power.

It would seem like a reasonable solution to either check first that nothing else is using the power rail before switching off the power or somehow regulate the power usage of the GPS module directly (I heard some chatter that this may be already happening and that the rail cut-off might be more of a legacy thing that may need re-addressing).

### Relevant log output

_No response_

thebentern:
Also closes #3128

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -255,19 +255,6 @@ bool GPS::setup()
     if (!didSerialInit) {
 #if !defined(GPS_UC6580)
 
-#if defined(RAK4630) && defined(PIN_3V3_EN)
-        // If we are using the RAK4630 and we have no other peripherals on the I2C bus or module interest in 3V3_S,
-        // then we can safely set en_gpio turn off power to 3V3 (IO2) to hard sleep the GPS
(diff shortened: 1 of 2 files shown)


tropho23:
Looks good to me!  Thanks

PR #3321, opened 2024-03-03
Add original hop limit to header to determine hops used

GUVWAF:
This adds the original hop limit of a packet (`hopStart`) to the unencrypted header using the three remaining bits such that we can derive with how many hops each packet was received. This will be saved in the NodeInfo as `hopsAway`. Also `viaMQTT` is now saved in the NodeInfo.

Using this, we can adapt the `hopLimit` of a response and real ACK to more or less hops if needed. This is implemented with `RoutingModule::getHopLimitForResponse()`. If the amount of hops used for the request was more than our setting, we use the amount of hops that was used for the request. When it used less than our setting, I chose a margin of 2 more hops for the response in case the path back is different.

This also allows us to determine neighbors even if others don’t have the NeighborInfo Module enabled, which is implemented as well.


thebentern (approved):

src/mesh/NodeDB.cpp
```diff
@@ -875,6 +875,12 @@ void NodeDB::updateFrom(const meshtastic_MeshPacket &mp)
 
         if (mp.rx_snr)
             info->snr = mp.rx_snr; // keep the most recent SNR we received for this node.
+
+        info->via_mqtt = mp.via_mqtt; // Store if we received this packet via MQTT
```
thebentern, line 879: This is good. I was wondering about how we could keep the information up to date with the latest packet delivery in the odd case of an mqtt node we know about that become available via lora range at some port, or vice-versa. Based on following this logic, updateFrom should keep us in sync with whatever state we have from the last packet, unless I'm mistaken.

thebentern (approved):

src/mesh/NodeDB.cpp
GUVWAF: Indeed, it will get updated based on the value from each received packet.

Issue #3310, opened 2024-03-01
[Bug]: GPS Checksum errors, HGLRC M100_MINI


jigpu:
### Category

Hardware Compatibility

### Hardware

Heltec V3

### Firmware Version

2.2.23.5672e68

### Description

I've just purchased an HGLRC M100_MINI GPS module from Amazon and wired it up to my Heltec V3. After sorting out the RX/TX pins in the configuration, I've been able to get Meshtastic to read an initial position fix. Unfortunately, after this initial fix things stop working. The GPS goes to sleep, and when awoken 2 minutes later it reports an endless series of "GPS checksum failures" messages. This continues for 15 minutes, after which the system gives up and puts the GPS to sleep for another 2 minutes. When the GPS wakes up again, we run into more checksum failures.

Highlights from the serial log:

System probes for and finds the GPS module:
~~~
DEBUG | ??:??:?? 1 [GPS] Probing for GPS at 9600 
WARN  | ??:??:?? 2 [GPS] Failed to find UBlox & MTK GNSS Module using baudrate 9600
DEBUG | ??:??:?? 4 [GPS] Probing for GPS at 4800 
DEBUG | ??:??:?? 4 [GPS] Setting Baud to 4800
WARN  | ??:??:?? 6 [GPS] Failed to find UBlox & MTK GNSS Module using baudrate 4800
DEBUG | ??:??:?? 8 [GPS] Probing for GPS at 38400 
DEBUG | ??:??:?? 8 [GPS] Setting Baud to 38400
INFO  | ??:??:?? 8 [GPS] BLE authentication complete
INFO  | ??:??:?? 8 [GPS] Found a UBlox Module using baudrate 38400
INFO  | ??:??:?? 9 [GPS] From Radio onread
DEBUG | ??:??:?? 9 [GPS] Module Info : 
DEBUG | ??:??:?? 9 [GPS] Soft version: ROM SPG 5.10 (7b202e)
DEBUG | ??:??:?? 9 [GPS] Hard version: 000A0000
DEBUG | ??:??:?? 9 [GPS] Extensions:4
DEBUG | ??:??:?? 9 [GPS]   FWVER=SPG 5.10
DEBUG | ??:??:?? 9 [GPS]   PROTVER=34.10
DEBUG | ??:??:?? 9 [GPS]   GPS;GLO;GAL;BDS
DEBUG | ??:??:?? 9 [GPS]   SBAS;QZSS
[...]

PR #3327, opened 2024-03-03
Fix for issue #3310

GPSFan:
I was able to duplicate the issue and have traced it back to my forgetting to save the baud rate settings for the M10. There are several flavors of M10 based modules about. The one I have with an SMA has a default baud of 9600 baud. The one I have with 38400 baud has no external connector so I had to switch back and forth to do testing on all the VALSET commands I was developing for the GPS rework. As luck would have it most of the testing was done with the module with the SMA antenna connector so I could get a good signal. That worked just find during the awake -->>sleep -->>awake sequence as all but the baud rate was saved in BBR, and its default baud rate was 9600.
This PR should fix that issue.


thebentern (approved):

thebentern:
Closes #3310

PR #3334, opened 2024-03-05
Add delay so GPS and Radio have time to power up

lolsborn:
- Added the boot-loader HW ID so you can flash from PlatformIO after inital boot loader setup
- Added a delay on boot which seems to have fixed some race-conditions with both the radio and GPS startup.  Canary has a separate power rail that was being brought up in initVariant() and sometimes either the GPS or Radio hardware would not be present after bootup.


src/main.cpp
```diff
@@ -529,6 +529,7 @@ void setup()
     digitalWrite(LED_PIN, 1 ^ LED_INVERTED); // turn on for now
 #endif
 
+
```
caveman99, line 532: Please don't check in whitespace changes. i recommend using `trunk fmt` to reformat your changed files.

src/main.cpp
```diff
@@ -679,6 +681,12 @@ void setup()
     digitalWrite(SX126X_ANT_SW, 1);
 #endif
 
+#ifdef CANARYONE
+    // This is required to give the GPS time to start up as this board has a separate power supply
```
caveman99, line 685: Can you define something more generic? Other boards may benefit from this delay too, and it's tied in with the `PIN_PWR_EN` so you can name it appropriately. (perhaps `PIN_PWR_DELAY_MS`?)

caveman99 (changes requested):

thebentern (approved):

PR #3337, opened 2024-03-05
Turn off certain modules not wanted in custom builds

caveman99:


caveman99:
Turns out i forgot the inputbroker in the 'nuke all' macro... will revisit tomorrow. It's not a real saving, gives you 2% flash on a t-beam.

I'd like to keep both options available, arch and module macros. this way the tuning is more granular, and you can make 'feature firmwares' for memory restraint devices. Apollo-Router, Apollo-Tracker, Apollo-Sensor and so on.

jimwhitelaw:
> or an ESP32 build without webserver (hint, hint)

I've done most of the work for this in my project already and would be happy to contribute it. Question: with no webserver, is there still a use case for enabling wifi? My code currently assumes there is not and disables both.

jp-bennett:
> with no webserver, is there still a use case for enabling wifi?

Technically there is still the TCP API, usable through the CLI and a couple apps.

jimwhitelaw:
OK, thx.  My initial work started with disabling wifi and then anything that depended on that, but now that I know it makes sense to have the TCP/wifi and webserver independently enabled, I can re-write to support that.

garthvh:
This is likely to crash when connected to a client app.

jimwhitelaw:
@garthvh Can you elaborate? I've been testing and haven't found any crash scenario, just that CLI and web apps fail to connect when webserver is disabled. BT connection with iOS app still works fine.

PR #3265, opened 2024-02-22
Add web server to portdunio linux version so the web gui can be served. Configuration doesn't need the CLI Interface anymore for the Linux Version

marelab:
Added a Webserver and WebService Interface to the Linux Portdunio Version. It now can deliver the web meshtastice project dist content also in the native linux version. It interfaces to the phoneapi the same way the ESP32 version does but only with Linux libs to get rid of the hardware dependencys of the ESP32 libs. For building under Linux it needs some additional libs these can be installed with bin/prepare-native.sh. To configure the web server the config.yaml got additional settings, that can be changed as needed. 

Webserver:
  Port: 9001 # Port for Webserver & Webservices
  RootPath: /home/marc/web # Root Dir of WebServer
  
The webserver RootPath has to point to the web/dist diretory of the https://github.com/meshtastic/web project, after it has been build.

PR #3343, opened 2024-03-06
Native Webserver

jp-bennett:


jp-bennett:
This history is messy enough, we should squash it when pulling.

thebentern (approved):

thebentern:
Close #3265 after merging this, or is there something additional?

jp-bennett:
> Close #3265 after merging this, or is there something additional?

We'll need to actually build and package the web project as part of the .deb to really make it plug-and-play.

Issue #3342, opened 2024-03-06
[Feature Request]: use callback of libpax for PaxcounterModule


thoherr:
### Platform

ESP32

### Description

The PaxcounterModule currently polls data from [libpax](https://github.com/dbinfrago/libpax) and sends it to the mesh (with configurable update interval). The underlying library is therefore initialized with countermode 1, which means that it continuously counts BLE and WiFi devices since start of the node, without ever resetting the data. Depending on the location and the uptime of the node, this accumulated count after a while looses its semantics.

Taking the standalone [PAXcounter Firmware](https://github.com/cyberman54/ESP32-Paxcounter) as blueprint, the idea there is to have counting intervals for the devices, after which the data is sent to a callback function and the values are reset to 0 (which allows to have time series of counts).

I therefore suggest to use the callback and the pax_report_interval_sec of the libpax library to implement the PaxcounterModule and set the countermode to another value, in order to have a time sliced count. This behaviour (countermode) may also be made configureable for the PaxcounterModule, which would be another (easy to implement) feature.

I already implemented the necessary changes and tested them successfully on my TTGO LoRa32 V2.1_1.6.

What do you thing about this change, @caveman99 ?

caveman99:
This sounds like a sensible approach, can you make a PR i can review?

PR #3351, opened 2024-03-08
Refactor paxcounter

thoherr:
Implements #3342 (using timesliced count, not yet configurable)

### DIFF
--- a/src/modules/esp32/PaxcounterModule.cpp
+++ b/src/modules/esp32/PaxcounterModule.cpp
@@ -7,8 +7,6 @@
 
 PaxcounterModule *paxcounterModule;
 
-void NullFunc(){};
-
 // paxcounterModule->sendInfo(NODENUM_BROADCAST);
 
 PaxcounterModule::PaxcounterModule()
@@ -19,9 +17,13 @@ PaxcounterModule::PaxcounterModule()
 
 bool PaxcounterModule::sendInfo(NodeNum dest)
 {
-    libpax_counter_count(&count_from_libpax);
-    LOG_INFO("(Sending): pax: wifi=%d; ble=%d; uptime=%d\n", count_from_libpax.wifi_count, count_from_libpax.ble_count,
+    if (paxcounterModule->reportedDataSent)
+        return false;
+
+    LOG_INFO("(Sending): pax: wifi=%d; ble=%d; uptime=%lu\n", count_from_libpax.wifi_count, count_from_libpax.ble_count,
              millis() / 1000);
+
+    paxcounterModule->reportedDataSent = true;
 
     meshtastic_Paxcount pl = meshtastic_Paxcount_init_default;
     pl.wifi = count_from_libpax.wifi_count;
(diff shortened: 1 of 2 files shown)


src/modules/esp32/PaxcounterModule.cpp
```diff
@@ -31,9 +49,12 @@ bool PaxcounterModule::sendInfo(NodeNum dest)
     meshtastic_MeshPacket *p = allocDataProtobuf(pl);
     p->to = dest;
     p->decoded.want_response = false;
-    p->priority = meshtastic_MeshPacket_Priority_MIN;
+    p->priority = meshtastic_MeshPacket_Priority_DEFAULT;
```
caveman99, line 52: Please leave this as min, regular message traffic should have priority.

src/modules/esp32/PaxcounterModule.h
```diff
@@ -8,11 +8,15 @@
 #include <libpax_api.h>
 
 /**
- * A simple example module that just replies with "Message received" to any message it receives.
+ * Wrapper module for the estimate passenger (PAX) count library (https://github.com/dbinfrago/libpax) which
+ * implements the core functionality of the ESP32 Paxcounter project (https://github.com/cyberman54/ESP32-Paxcounter)
  */
```
caveman99, line 13: Woops, good catch :-)

src/modules/esp32/PaxcounterModule.cpp
```diff
@@ -76,12 +97,14 @@ int32_t PaxcounterModule::runOnce()
...
-            libpax_counter_init(NullFunc, &count_from_libpax, UINT16_MAX, 1);
+            libpax_counter_init(handlePaxCounterReportRequest, &count_from_libpax,
+                                moduleConfig.paxcounter.paxcounter_update_interval, 0);
             libpax_counter_start();
         } else {
             sendInfo(NODENUM_BROADCAST);
         }
-        return getConfiguredOrDefaultMs(moduleConfig.paxcounter.paxcounter_update_interval, default_broadcast_interval_secs);
+        // we check every second if the counter had new data to send
+        return 1000;
```
caveman99, line 107: this is rather heavy on the threading system, like polling for a flag. My suggestion is to keep the original interval (no change here) and trigger the osthread from the outside when new data arrived. See comment above

src/modules/esp32/PaxcounterModule.cpp
```diff
@@ -7,21 +7,39 @@
...
+ * We only clear our sent flag here, since this function is called from another thread, so we
+ * cannot send to the mesh directly.
+ */
+void PaxcounterModule::handlePaxCounterReportRequest()
+{
+    // The libpax library already updated our data structure, just before invoking this callback.
+    LOG_INFO("PaxcounterModule: libpax reported new data: wifi=%d; ble=%d; uptime=%lu\n",
+             paxcounterModule->count_from_libpax.wifi_count, paxcounterModule->count_from_libpax.ble_count, millis() / 1000);
+    paxcounterModule->reportedDataSent = false;
+}
```
caveman99, line 21: you can trigger the osthread processing here by calling paxcounterModule->setIntervalFromNow(0);

caveman99 (changes requested):

src/modules/esp32/PaxcounterModule.cpp
thoherr: ok, no problem.
Just a question: When would BACKGROUND be appropriate?

caveman99: Pretty well explained in https://github.com/meshtastic/protobufs/blob/00332412b238fe559175a6e83fdf8d31fa5e209a/meshtastic/mesh.proto#L819-L878 ... I think BACKGROUND may be more appropriate than MIN here, and in other places that use it. (Telemetry and so on...) For now i'd stick to min cause the other plugins use min too. Let's address this in a different PR.

caveman99 (approved):

PR #3363, opened 2024-03-10
always send range tests with zero hops

andrekir:
range testing is typically measured point-to-point (sender --> receiver), adding hops can result in inaccurate data and unnecessarily extend packet air time.


garthvh (approved):
Great idea!

caveman99 (approved):

Issue #3336, opened 2024-03-05
[Feature Request]: MQTT Location only mode


nukemrules:
### Platform

NRF52, ESP32

### Description

While the MQTT platform works great if you know what you are doing, lots of the new Meshtastic users are using the MQTT uplink only to get their location on one of the map sites. Unfortunately this resolves in many loops and multiple connections in a mesh to the official MQTT server. With the duty cycle of only 10% in Europe, this makes that the airtime is reached quite fast if you connect to the big topics on the official MQTT server too.

A way to move new users to a more secure way of publishing their or mesh locations is to have a separate topic on the MQTT server and a setting in the node to only publish its location. So its a one way publish. The node itself never reads any information from that topic. The Node map/list on the node should only be filled with nodes directly from the mesh. 
Device and/or Environment metrics could also be a big plus to publish in this or separate topic, but it should be optional. This way the node can be publicly monitored over the internet.

With this feature the mesh can do its thing without all of the extra traffic, but the rest of the world can see where the nodes/mesh networks are available.

GUVWAF:
Just to be sure: you know you can already only enable uplink (and not downlink) to avoid transmitting packets from MQTT over LoRa while still publishing your position on the map? 

I don't see why you need separate topics for position and telemetry to accomplish what you want.

nukemrules:
> Just to be sure: you know you can already only enable uplink (and not downlink) in the channel settings to avoid transmitting packets from MQTT over LoRa while still publishing your position on the map?
> 
> I don't see why you need separate topics for position and telemetry to accomplish what you want.

Sure I know, that is a full one way publish and means that everything from the mesh is published to the MQTT server too. Unfortunately when someone subscribes to that topic and uses up- and downlink, messages can't ever reach the nodes via the node that has only uplink enabled. With a location only mode, you protect the expectation of reaching nodes that can't be reached. And with a separate topic, you protect the mesh from flooding too.

GUVWAF:
> You can't hide messages in your public primary channel. Sure you can create a "private" channel, but you want a public one too for people from outside your mesh.

You can create a private primary channel to publish your position and not use it for anything else and have the "public" channel as secondary.
If you want to talk in the clear over LoRa on the primary channel but not publish it to MQTT, while still publishing your location, it sounds like a solution would be to allow periodic position publishing to MQTT on a different channel than the primary.

nukemrules:
[...]

PR #3365, opened 2024-03-10
Periodic reporting of device information to a map via MQTT

GUVWAF:
This adds an easy way to periodically report all kinds of information about the device (long/short name, position with precision, firmware version, LoRa settings, number of local online nodes, etc.) to a map via MQTT only, regardless of any channel and is thus unencrypted. It will be published to a new `/2/map/` topic as a protobuf in a `ServiceEnvelope` with `MeshPacket` portnum `MAP_REPORT_APP` and can be decoded using the new `MapReport` message. 
If map reporting is enabled, the publish interval (default is 15 min.) and position precision (default is full precision) can be configured.

Implements #3336.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -77,8 +77,7 @@ void MQTT::onReceive(char *topic, byte *payload, size_t length)
                         if (jsonPayloadStr.length() <= sizeof(p->decoded.payload.bytes)) {
                             memcpy(p->decoded.payload.bytes, jsonPayloadStr.c_str(), jsonPayloadStr.length());
                             p->decoded.payload.size = jsonPayloadStr.length();
-                            meshtastic_MeshPacket *packet = packetPool.allocCopy(*p);
-                            service.sendToMesh(packet, RX_SRC_LOCAL);
+                            service.sendToMesh(p, RX_SRC_LOCAL);
                         } else {
                             LOG_WARN("Received MQTT json payload too long, dropping\n");
                         }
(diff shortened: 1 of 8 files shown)


src/mqtt/MQTT.cpp
```diff
@@ -77,8 +77,7 @@ void MQTT::onReceive(char *topic, byte *payload, size_t length)
                         if (jsonPayloadStr.length() <= sizeof(p->decoded.payload.bytes)) {
                             memcpy(p->decoded.payload.bytes, jsonPayloadStr.c_str(), jsonPayloadStr.length());
                             p->decoded.payload.size = jsonPayloadStr.length();
-                            meshtastic_MeshPacket *packet = packetPool.allocCopy(*p);
-                            service.sendToMesh(packet, RX_SRC_LOCAL);
+                            service.sendToMesh(p, RX_SRC_LOCAL);
```
caveman99, line 80: is there a reason why you removed the alloc?

caveman99 (changes requested):
I updated protos for your convenience :-) Please see comment.

src/mqtt/MQTT.cpp
GUVWAF: Oh, yeah that's actually unrelated to this pull, but I just stumbled upon it. I believe this leads to a memory leak as `p` wouldn't be freed. I believe the copy is unnecessary here; the "sendtext" part also doesn't use it.

caveman99: We can try, sounds plausible ... (how does telemetry handle this? there was a leak in there too...)

GUVWAF: At least currently it only allocates a `MeshPacket` once and uses that in `sendToMesh()`, as it should be:
https://github.com/meshtastic/firmware/blob/d1a25947e302cff9c0f1cb9b636569b67960fa9f/src/modules/Telemetry/DeviceTelemetry.cpp#L91

caveman99: sounds about right, telemetry was one of the sources of memory leaks... one more down :-)

caveman99 (approved):

PR #3369, opened 2024-03-11
Adds configuration option to exclude the webserver on esp32.

jimwhitelaw:
Followup to the [suggestion](https://github.com/meshtastic/firmware/pull/3337#issuecomment-1979652448) by @caveman99 to make including the webserver optional on esp32. Excluding the webserver saves about 8kb of flash  on the device and disables the web app and the TCP API. TCP (via wifi or ethernet) remains enabled for use by NTP or MQTT if needed. Tested with tbeam and tlora-v2-1-1_6 devices.


src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -92,11 +96,13 @@ static void onNetworkConnected()
...
         }
 
+#if !MESHTASTIC_EXCLUDE_WEBSERVER
+
 #ifdef ARCH_ESP32
         initWebServer();
 #endif
         initApiServer();
-
+#endif
```
caveman99, line 105: That technically works, but it excludes the API Socket Server which may be available on other platforms (nRF52 with ETH) if that macro is set there errorneously. For clarity, can you add an ELSE branch to the outer macro making sure the apiserver is initialized even when the macro is set?

caveman99 (changes requested):

src/mesh/wifi/WiFiAPClient.cpp
jimwhitelaw: Hi, thank you for the comments. I had done as you suggest initially, but in testing I realized that the API server relies on the http(s) server anyway and is disabled by this change. There might be some utility in separating the two services, but it won't create any substantial space savings. For platforms with ETH, I think that initAPIServer() will be called from ethClient.ccp (and will probably fail when webserver is disabled, now that I think of it).

jimwhitelaw: OK, I see what I was missing before. There is an API server running on a separate port, independent of the web/http servers. Adding an update to the PR.

jimwhitelaw:
@caveman99, I'm not sure what I'm to do about these failing CI checks on files I haven't changed. Can you offer any insight? I've got some additional changes lined up to facilitate minimal builds using the pattern you've initiated (build w/o BT, Wifi, GPS) but I'm struggling to understand what mistake I'm making here. thx

caveman99 (approved):

thebentern (approved):

PR #3383, opened 2024-03-12
Fix for incorrect mapTopic mqtt path

### LINKED ISSUE #3382: [Bug]: mapTopic mqtt path is built incorrectly
When building the `mapTopic` variable with a mqtt root set, it appends the `jsonTopic` variable instead of the `mapTopic` variable resulting in incorrect mqtt path.

https://github.com/meshtastic/firmware/blob/5f47ca1f32dd283739c4d78a4e08f20a9bd60fa1/src/mqtt/MQTT.cpp#L188

Example console output:
```
INFO | 02:00:33 900 [mqtt] MQTT Publish map report to msh/EU_868msh/EU_868/2/json/
```

Expected console output:
```
INFO | 02:00:33 900 [mqtt] MQTT Publish map report to msh/EU_868/2/map/
```

AeroXuk:
fixes #3382


GUVWAF (approved):
Good catch! Stupid copy-paste error and I didn't test with custom root topic.

PR #3385, opened 2024-03-12
Fix LTO discharge curve

wnagele:
Been testing LTO cells recently and noticed that the SoC values didn't make much sense.

Upon a closer look the values for the discharge curve are far off from what a typical LTO discharge curve looks like. Here I put in the current curve (red dots) to visualise this for myself:
lto_dis_curve
This PR changes the curve to the values from this graph (1C) which in turn result in reasonable SoC values.

I did a cursory check on the other chemistries and they don't seem to suffer the same issue. So I guess some wrong discharge curve was used to deduct those values from maybe?


Gabrielerusso (commented):
Those are not code of line and aren't put there randomly, for example there are three different curve for LiIon to chose from, and are there for reference for who wants to edit them to their custom usage.

Gabrielerusso:
> Maybe @Gabrielerusso can also give some feedback. As the original implementation is from him in #3216. I think the LTO curve was from @code8buster.

Never took a look on what the LTO curve was, the other one I took them from BMS projects I was working so should be correct.

wnagele:
On second thought - this is probably too many options.
If people already have to go re-configure, they can of course also adjust those curves whichever way they see fit.
So I suggest we keep this to the already long list of chemistries that are supported now.

@thebentern this should be fine to merge.

PR #3387, opened 2024-03-12
Check AQ_SET_PIN instead of EINK dependency

wnagele:
We should really check if the `AQ_SET_PIN` is defined here instead of depending on some other configuration element that has nothing directly to do with it. Checked against current RAK4630 based configs and should not break anything.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -342,7 +342,7 @@ void setup()
     pinMode(PIN_3V3_EN, OUTPUT);
     digitalWrite(PIN_3V3_EN, HIGH);
 #endif
-#ifndef USE_EINK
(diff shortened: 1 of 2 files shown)


caveman99 (commented):
I think the tie-in with EINK in this case is because it also uses the same GPIO line. Need to reconfirm that change.

wnagele:
> I think the tie-in with EINK in this case is because it also uses the same GPIO line. Need to reconfirm that change.
Yes it is - but this should be clear for the ones using the eInk stuff in their configs as they have that pin assigned there. Otherwise we have tons of such checks for possible collisions in the code at some point.

PR #3404, opened 2024-03-13
use decoded packets in public MQTT range test/detection sensor filter

andrekir:
always use decoded packets (`mp_decoded`) to filter range test or detection sensor messages out of the public MQTT server, as we can't read the portnum from encrypted packets.


thebentern:
Good catch!

thebentern (approved):

PR #3410, opened 2024-03-14
(1/3) Support L76B GNSS chip found on pico waveshare shield.

caveman99:
Original work by @Mictronics for slow CPU clock support on Pico


src/gps/GPS.cpp
```diff
@@ -1168,6 +1211,9 @@ bool GPS::factoryReset()
         // byte _message_CFG_RST_COLDSTART[] = {0xB5, 0x62, 0x06, 0x04, 0x04, 0x00, 0xFF, 0xB9, 0x00, 0x00, 0xC6, 0x8B};
         // _serial_gps->write(_message_CFG_RST_COLDSTART, sizeof(_message_CFG_RST_COLDSTART));
         // delay(1000);
+    } else if (HW_VENDOR == meshtastic_HardwareModel_RPI_PICO) {
```
thebentern, line 1214: Should this only be based on one hardware variant?

caveman99: no, but let's address this after i pushed 3/3 :-)

thebentern (approved):

PR #3414, opened 2024-03-14
More comprehensive client proxy queue guards

thebentern:
From a call with @mc-hamster ;-)

### DIFF
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -105,7 +105,8 @@ bool PhoneAPI::handleToRadio(const uint8_t *buf, size_t bufLength)
             break;
         case meshtastic_ToRadio_mqttClientProxyMessage_tag:
             LOG_INFO("Got MqttClientProxy message\n");
-            if (mqtt && moduleConfig.mqtt.proxy_to_client_enabled && moduleConfig.mqtt.enabled && channels.anyMqttEnabled()) {
+            if (mqtt && moduleConfig.mqtt.proxy_to_client_enabled && moduleConfig.mqtt.enabled &&
+                (channels.anyMqttEnabled() || moduleConfig.mqtt.map_reporting_enabled)) {
                 mqtt->onClientProxyReceive(toRadioScratch.mqttClientProxyMessage);
             } else {
                 LOG_WARN("MqttClientProxy message received but it is not enabled or no channels have up/downlink\n");
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -105,8 +105,10 @@ bool PhoneAPI::handleToRadio(const uint8_t *buf, size_t bufLength)
             break;
         case meshtastic_ToRadio_mqttClientProxyMessage_tag:
             LOG_INFO("Got MqttClientProxy message\n");
-            if (mqtt && moduleConfig.mqtt.proxy_to_client_enabled) {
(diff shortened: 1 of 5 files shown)


src/mqtt/MQTT.cpp
```diff
@@ -401,7 +401,7 @@ bool MQTT::wantsLink() const
 
 int32_t MQTT::runOnce()
 {
-    if (!moduleConfig.mqtt.enabled)
+    if (!moduleConfig.mqtt.enabled || !channels.anyMqttEnabled())
```
GUVWAF, line 404: Makes sense indeed to not run the thread in this case, but most of this is already captured by [`MQTT:wantsLink()`](https://github.com/meshtastic/firmware/blob/9c37e57e750a72324ea58f6acda53168c4efae50/src/mqtt/MQTT.cpp#L372), which now also checks whether map reporting is enabled.

thebentern: Just consolidated that channel check with my new method.

thebentern: Should we just use if (!wantsLink()) return disable() perhaps?

GUVWAF: Changing channel settings does not necessarily cause a reboot, right? If not, it would never re-enter the `runOnce()` if you enable uplink/downlink after enabling MQTT.

thebentern: This is true. We probably need to leave that guard out to be safe.
So this PR will still have some value as moving some of that logic back onto Channel and protecting the client proxy queue from getting spammed for no reason (which I witnessed during testing.

I'll make the changes.

GUVWAF: Indeed, that part looks good.

src/mesh/PhoneAPI.cpp
```diff
@@ -105,8 +105,10 @@ bool PhoneAPI::handleToRadio(const uint8_t *buf, size_t bufLength)
             break;
         case meshtastic_ToRadio_mqttClientProxyMessage_tag:
             LOG_INFO("Got MqttClientProxy message\n");
-            if (mqtt && moduleConfig.mqtt.proxy_to_client_enabled) {
+            if (mqtt && moduleConfig.mqtt.proxy_to_client_enabled && moduleConfig.mqtt.enabled && channels.anyMqttEnabled()) {
```
thebentern, line 108: @GUVWAF this guard probably needs to consider the new map report enabled too right?

GUVWAF (approved):

src/mqtt/MQTT.cpp
caveman99: > Changing channel settings does not necessarily cause a reboot, right? If not, it would never re-enter the `runOnce()` if you enable uplink/downlink after enabling MQTT.

you can always resurrrect a thread by calling `setIntervalFromNow(0)` to get it back out of the MAX_UINT trap. So if we just do that on channel changes we'll be fine.

caveman99: > This is true. We probably need to leave that guard out to be safe. So this PR will still have some value as moving some of that logic back onto Channel and protecting the client proxy queue from getting spammed for no reason (which I witnessed during testing.

See my comment to @GUVWAF :-) we can do this...

caveman99 (changes requested):
See comments in discussion

thebentern:
> See comments in discussion

Tested on my device. MQTT starts up like a champ after channels change to have uplink/downlink

PR #3415, opened 2024-03-15
NodeInfo broadcast ensure default on 0 and enforce 1 hour minimum

thebentern:
As discussed in #transport-layer room.

Also removed some of the no_save cruft that's been an eye-sore.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -101,16 +101,7 @@ bool NodeDB::resetRadioConfig(bool factory_reset)
     // devicestate.no_save = true;
     if (devicestate.no_save) {
         LOG_DEBUG("***** DEVELOPMENT MODE - DO NOT RELEASE *****\n");
-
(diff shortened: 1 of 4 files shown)


GUVWAF (approved):
Looks good. We should document this and probably recommend enabling smart position for "wardriving".

caveman99:
> Also removed some of the no_save cruft that's been an eye-sore.

Does anybody actually use this? if it was for me we can throw it out completely. Also the module presets in source comments...

thebentern:
I do not. I've maybe used it once in the past 3 years. I suspect Kevin used it a lot in earlier development of the project.
Should we get rid of the no_save on device state protos? I don't believe anybody else uses it.

> > Also removed some of the no_save cruft that's been an eye-sore.
> 
> Does anybody actually use this? if it was for me we can throw it out completely. Also the module presets in source comments...

geeksville:

https://meshtastic.discourse.group/t/settings-for-driving-around-to-find-other-nodes/7525/10

PR #3423, opened 2024-03-16
Bump minimum NodeInfo send to 5 minutes

thebentern:
What do we think about this change to combat some of the excessive chattiness of NodeInfo on larger meshes?

### DIFF
--- a/src/modules/NodeInfoModule.cpp
+++ b/src/modules/NodeInfoModule.cpp
@@ -58,8 +58,8 @@ void NodeInfoModule::sendOurNodeInfo(NodeNum dest, bool wantReplies, uint8_t cha
 meshtastic_MeshPacket *NodeInfoModule::allocReply()
 {
     uint32_t now = millis();
-    // If we sent our NodeInfo less than 1 min. ago, don't send it again as it may be still underway.
-    if (lastSentToMesh && (now - lastSentToMesh) < 60 * 1000) {
+    // If we sent our NodeInfo less than 3 min. ago, don't send it again as it may be still underway.
+    if (lastSentToMesh && (now - lastSentToMesh) < (3 * 60 * 1000)) {
         LOG_DEBUG("Sending NodeInfo will be ignored since we just sent it.\n");
         ignoreRequest = true; // Mark it as ignored for MeshModule
         return NULL;
--- a/src/modules/NodeInfoModule.cpp
+++ b/src/modules/NodeInfoModule.cpp
@@ -58,8 +58,8 @@ void NodeInfoModule::sendOurNodeInfo(NodeNum dest, bool wantReplies, uint8_t cha
 meshtastic_MeshPacket *NodeInfoModule::allocReply()
 {
     uint32_t now = millis();
-    // If we sent our NodeInfo less than 3 min. ago, don't send it again as it may be still underway.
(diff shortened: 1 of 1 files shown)


andrekir:
considering the [flood expire time](https://github.com/meshtastic/firmware/blob/54a2a4bcc67793fc8e72621d487153d4eb801c3d/src/mesh/PacketHistory.h#L7) of 10 minutes, 3 min is very reasonable. maybe could even bump it up to 5.

garthvh:
Why not the whole 10?

GUVWAF:
I don't think we should set it too high, it's a bit annoying for first-time users if you have to wait a while before a node shows up in your node list if you start up one after the first sent out its initial NodeInfo.

thebentern:
Survey time? 😄 
Thumbs up this comment for 3 minutes
Thumbs down this comment for 5 minutes

andrekir:
https://github.com/meshtastic/firmware/blob/13cc1b0252b5fdae9a195412039ea46b0288d634/src/mesh/MeshService.cpp#L92-L93

something else we should consider is always sending NodeInfo to `NODENUM_BROADCAST`, instead of DMs.

garthvh:
> I'm fine with either 3 or 5 minutes. no real preference.
> 
> https://github.com/meshtastic/firmware/blob/13cc1b0252b5fdae9a195412039ea46b0288d634/src/mesh/MeshService.cpp#L92-L93
> 
> something else we should consider is always sending NodeInfo to `NODENUM_BROADCAST`, instead of DMs.

Yeah I fixed this the other day because it was making it seem like some nodes were on the wrong channel

GUVWAF:
I'm also fine with either 3 or 5 min. 

In the specific case Andre linked I think it makes sense to send the NodeInfo as DM, because we request a response from the unknown node, we don't want everyone to respond.

andrekir:
> In the specific case Andre linked I think it makes sense to send the NodeInfo as DM, because we request a response from the unknown node, we don't want everyone to respond.

for sure, we would send to `NODENUM_BROADCAST` and wantReplies `false`. unless we really need a response. otherwise as a DM we would only reply to one single node every 3-5 minutes.

GUVWAF:
Yeah, makes sense to only set `lastSentToMesh` when it's a broadcast.

thebentern:
Could we make NodeInfo DMs (or maybe they already are and I missed it) read promiscuously as long as we can decode? That way, if I send a DM to interrogate NodeInfo only for a specific destination node that I don't know yet, any observers still would also get the benefit of observing that (but not rebroadcasting).

garthvh:
> Could we make NodeInfo DMs (or maybe they already are and I missed it) read promiscuously as long as we can decode? That way, if I send a DM to interrogate NodeInfo only for a specific destination node that I don't know yet, any observers still would also get the benefit of observing that (but not rebroadcasting).

Yeah I was already able to decode them, I just was not expecting a dm

andrekir:
> Could we make NodeInfo DMs (or maybe they already are and I missed it) read promiscuously as long as we can decode? 

I like that. we could finally use `sniffReceived()`.

https://github.com/meshtastic/firmware/blob/e27f029d09424698f6e8b00ac64efe73ee85749a/src/mesh/Router.cpp#L282-L289

PR #3427, opened 2024-03-16
Add MaxNodes to Native config

jp-bennett:


thebentern (approved):

bin/regen-protos.sh
```diff
@@ -10,7 +10,9 @@ echo "prebuilt binaries for your computer into nanopb-0.4.7"
 cd protobufs
 ../nanopb-0.4.7/generator-bin/protoc --nanopb_out=-v:../src/mesh/generated/ -I=../protobufs meshtastic/*.proto --experimental_allow_proto3_optional
 
-# cd ../src/mesh/generated/meshtastic
+cd ../src/mesh/generated/meshtastic
+rename .c .cpp *.c
```
mverch67, line 14: I think it would be better to use the nanopb option to change the default extension to .cpp:
```
  -S EXTENSION, --source-extension=EXTENSION
                        Set extension to use for generated source files.
                        [default: .c]
```

src/modules/esp32/AudioModule.cpp
```diff
@@ -1,12 +1,11 @@
-
 #include "configuration.h"
 #if defined(ARCH_ESP32) && defined(USE_SX1280)
 #include "AudioModule.h"
 #include "FSCommon.h"
 #include "MeshService.h"
-#include "NodeDB.h"
 #include "RTC.h"
 #include "Router.h"
+#include "nodeDB.h"
```
caveman99, line 8: That include is not found here... needs a capital N

caveman99 (changes requested):

bin/regen-protos.sh
caveman99: Also valid concern by trunk : `Use ./*glob* or -- *glob* so names with dashes won't become options.  shellcheck/SC2035` - better ditch that line and go with the other suggestion

caveman99 (changes requested):

bin/regen-protos.sh
jp-bennett: This isn't an option in protoc, that I see. 
```
jbennett@Celeborn:~/src/firmware$ ./nanopb-0.4.7/generator-bin/protoc --source-extension=.cpp
Unknown flag: --source-extension
```

src/modules/esp32/AudioModule.cpp
jp-bennett: Doh! Fixed.

bin/regen-protos.sh
jp-bennett: That is not straightforward. But got it.

jp-bennett:
I think this is in generally good shape, if we're willing to squash it into one commit for merge. I've been running it on a native client connected to MQTT, with no issues for nearly 5 hours, and a RAK also seems to be behaving well with it.

src/mesh/NodeDB.cpp
```diff
@@ -925,6 +925,8 @@ meshtastic_NodeInfoLite *NodeDB::getOrCreateMeshNode(NodeNum n)
     meshtastic_NodeInfoLite *lite = getMeshNode(n);
 
     if (!lite) {
+        if ((numMeshNodes >= MAX_NUM_NODES) || (memGet.getFreeHeap() < meshtastic_NodeInfoLite_size * 3))
+            cleanupMeshDB();
```
mverch67, line 929: I think this should go into a separate PR. Increasing MAX_NUM_NODES or making it configurable is unrelated to the way we handle purging old entries. 
Also, this function should be tested sufficiently to guarantee correct and desired behavior (e.g. sending two successive msgs from new nodes while max_num is reached), so that e.g. ACKs are always sent. Has such a test been done (e.g. with MAX_NUM_NODES set to a very low number like 1 or 2 and check what happens)?

GUVWAF: Sending ACKs doesn’t require a node to be in the NodeDB. The destination (`to`) will be set according to the `from` field of the received packet.

jp-bennett: Seems reasonable to pull this, and make it a dedicated PR.

PR #3459, opened 2024-03-21
Add sanity check for map report interval and position precision

### LINKED ISSUE #3458: [Bug]: MapInfo being sent multiple times per second
### Firmware Version

2.3.1.4fa7f5a0

### Description

Using MQTT Explorer I can see that some nodes are sending their mapInfo multiple times per second.

I do not know the reason for this at the moment, it could be misconfigured nodes (in which case maybe we need to enforce a minimum refresh rate), or it could be a bug.

Bad Node Examples:
image

image

Topic Summary:
image

"Bad" nodes are sending hundereds to thousands of messages per minute. Where as "good" nodes are only sending every 15 mins.

GUVWAF:
Should fix #3458.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -195,8 +195,10 @@ MQTT::MQTT() : concurrency::OSThread("mqtt"), mqttQueue(MAX_MQTT_QUEUE)
         }
 
         if (moduleConfig.mqtt.map_reporting_enabled && moduleConfig.mqtt.has_map_report_settings) {
-            map_position_precision = moduleConfig.mqtt.map_report_settings.position_precision;
-            map_publish_interval_secs = moduleConfig.mqtt.map_report_settings.publish_interval_secs;
+            if (moduleConfig.mqtt.map_report_settings.position_precision > 0)
+                map_position_precision = moduleConfig.mqtt.map_report_settings.position_precision;
+            if (moduleConfig.mqtt.map_report_settings.publish_interval_secs > 0)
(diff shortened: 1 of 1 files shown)


src/mqtt/MQTT.cpp
```diff
@@ -195,8 +195,10 @@ MQTT::MQTT() : concurrency::OSThread("mqtt"), mqttQueue(MAX_MQTT_QUEUE)
         }
 
         if (moduleConfig.mqtt.map_reporting_enabled && moduleConfig.mqtt.has_map_report_settings) {
-            map_position_precision = moduleConfig.mqtt.map_report_settings.position_precision;
-            map_publish_interval_secs = moduleConfig.mqtt.map_report_settings.publish_interval_secs;
+            if (moduleConfig.mqtt.map_report_settings.position_precision > 0)
+                map_position_precision = moduleConfig.mqtt.map_report_settings.position_precision;
+            if (moduleConfig.mqtt.map_report_settings.publish_interval_secs > 0)
+                map_publish_interval_secs = moduleConfig.mqtt.map_report_settings.publish_interval_secs;
```
thebentern, line 201: Perhaps this is a good candidate for our Default::getIntervalOrDefaultMs?

GUVWAF: Ah, good one. I'll wait until #3456 finishes CI so I don't have to resolve conflicts :)

GUVWAF: Done.

thebentern (approved):

PR #3469, opened 2024-03-22
Add multiple configuration options for a minimized build (GPS,WiFi,BT,…,MQTT,Screen)

jimwhitelaw:
This change follows up on the work to exclude modules and webserver from a minimized build. It provides options to exclude Wifi, Bluetooth, GPS, Screen, and MQTT from hardware that would otherwise support those options. The options can be enabled individually or globally with -DMESHTASTIC_MINIMIZE_BUILD. Tested for unchanged mesh/messaging behaviour on TBeam and Lora32 v2.1-1.6 using CLI test and with other nodes sending messages through the system. I'm unable to test with other hardware currently.


src/modules/AdminModule.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 #include "ProtobufModule.h"
-#if HAS_WIFI
+#if HAS_WIFI && !MESHTASTIC_EXCLUDE_WIFI
```
thebentern, line 3: Is this needed if MESHTASTIC_EXCLUDE_WIFI defines HAS_WIFI as 0?

thebentern:
@jimwhitelaw out of pure curiosity, what is your use case for shedding many of the peripherals for a minimized build? I know we're definitely looking at it as a way to easily cut down on things included in lower resource platforms like STM32.

src/modules/AdminModule.h
jimwhitelaw: Oh, yes you’re right. That is redundant. Not at the computer now, but I can push the change later.

jimwhitelaw:
@thebentern My project using MT is a remote flag signalling system for motorsports. A network of nodes placed around a racetrack connected to large LED panels that display electronic “flags” to racers. The motivation to minimize is two-fold: 1) to keep the system as closed as possible to prevent any outside intervention with the system (can’t hack WiFi or BT if they aren’t available) and 2) to potentially reduce cpu utilization and flash space to allow for running other code. 

I have a working prototype using a custom module that can send and process signal messages and update individual LED panels on a small network of 5 nodes. Configuration is done via serial and uploading pre-built config files for each node. All the nodes will be stationary at known locations so no need for GPS. 

I don’t know much about STM32 development currently, but I’m happy to contribute what I can. 

With all that being said, since I started working with the project, I’ve become interested in the comms side and will be setting up a couple “permanent” nodes on the MT network and experimenting with that too.

src/modules/AdminModule.h
jimwhitelaw: Looking at this again, I think the original is correct. If MESHTASTIC_EXCLUDE_WIFI is not defined, HAS_WIFI can still be 0 or 1 (hardware dependent). Re-ordering the test might make the logic a bit more clear though.

PR #3474, opened 2024-03-23
Fix #3452: only alter received packet if port number matches

### LINKED ISSUE #3452: Messages often come through garbled as "�떗" since recent update
Since updating to Android client 2.3.0, messages on LongFast sometimes come through garbled. In one particularly bad case, 8 out of 9 messages received over the course of an hour were garbled. The displayed text is always the same, regardless of the original message content, or the sender.

`�떗`

This was also a few days after applying firmware update 2.2.24 to my Heltec V3, so it is possible the issue resides there, however no iOS users I've spoken with thus far have noted any issues since the firmware update, and in the days between updating the firmware and the Android client this issue did not crop up. So, my first thought is that it could be a client issue (but this needs more investigation).

Attached are three screenshots from the Android app, one showing LongFast with two garbled messages, and the other two screenshots are from the debug screen showing the details of those two messages.

IMG_20240314_230748
IMG_20240314_230619
IMG_20240314_230642

A few other users have discussed the issue in the Meshtastic Discord, they might be able to provide more information, and if there's any more info I can provide just let me know. Thank you for all your hard work, it's greatly appreciated <3

GUVWAF:
Fixes #3452. 

If enabled, the NeighborInfo module was altering a packet even if it wasn’t a NeighborInfo packet. This was introduced when `wantPacket()` was changed in PR #3321, because with `hopStart` we can also determine who are neighbors for any other packet.

To avoid this, I let the NeighborInfo module (and also Traceroute) now use `alterReceivedProtobuf()` instead of calling the methods in the FloodingRouter. To make sure `alterReceived()` of the Protobuf module does not call this with `decoded` set to NULL, I moved that into the clause which checks if the port number matches. 
Tested that this fixes the decoding issues and it still works correctly for both the NeighborInfo module and Traceroute (also with hop in between).

Also removed some extensive logging from the NeighborInfo module since it would print your full NodeDB when sending a packet. Now it only prints your neighbors.

### DIFF
--- a/src/mesh/FloodingRouter.cpp
+++ b/src/mesh/FloodingRouter.cpp
@@ -49,15 +49,6 @@ void FloodingRouter::sniffReceived(const meshtastic_MeshPacket *p, const meshtas
 
                 tosend->hop_limit--; // bump down the hop count
 
-                if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
-                    // If it is a traceRoute request, update the route that it went via me
-                    if (traceRouteModule && traceRouteModule->wantPacket(p))
-                        traceRouteModule->updateRoute(tosend);
-                    // If it is a neighborInfo packet, update last_sent_by_id
-                    if (neighborInfoModule && neighborInfoModule->wantPacket(p))
-                        neighborInfoModule->updateLastSentById(tosend);
-                }
-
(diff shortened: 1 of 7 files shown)


thebentern (approved):

thebentern:
Very nice! It cleans up the code quite a bit to use that new alter method.

Issue #3226, opened 2024-02-15
[Feature Request]: Special case in `Screen::handleSetOff()` for E-Ink displays


todd-herbert:
### Platform

ESP32

### Description

After several minutes at idle, log gets message `[PowerFSM] Turning off screen`. 
After this point, the E-Ink display does not receive any updates until user presses a button to wake the device.

Because the E-Ink display image persists, even after `Screen::handleSetOff()` runs, it is not immediately clear that display updates have suspended.

It may be worth considering either:

  * Allowing the E-Ink display to remain active
    Unlike traditional displays, there is no considerable standby power consumption.

  * Moving to a standby screen before "turning off screen"
    This would make it clear to the user that updates are suspended, and that a button press is required to display current info.

thebentern:
Will be closed by #3477

PR #3477, opened 2024-03-23
E-Ink Screensaver

todd-herbert:
On E-Ink devices, when screen-timeout is reached, a "screensaver frame" is drawn before updates are paused.
This prevents misleading, out-of-date information remaining on the display.
Addresses https://github.com/meshtastic/firmware/issues/3226

Tested on Heltec Wireless Paper V1.0 and V1.1. Would be good to confirm that everything is working as expecting on other E-Ink devices too.

(It's not toasters, *I swear*)

https://github.com/meshtastic/firmware/assets/24772776/017199b9-b47c-4895-8841-a1287045d7c8


todd-herbert:
Finger slipped onto the enter key..

GMart:
I think it's a great idea, but I don't like that all the useful information other than the version goes away. Maybe, the screen could stay there, but just have text in the middle of the screen over everything that says "Screen paused"? So the info on the screen stays.
Or, since it's e-ink, when screen timeout happens, continue running the screen at half the regular update rate? Or maybe only when there is a position packet/direct message/something else?

todd-herbert:
> I don't like that all the useful information other than the version goes away.

Is there any information in particular that you feel would be useful to remain on screen?

> Or maybe only [update] when there is a position packet/direct message/something else?

I think what you're talking about might already be implemented with https://github.com/meshtastic/firmware/pull/3316
It's currently enabled for Heltec Wireless Paper and T-Echo, but hasn't been configured yet for any of the other E-Ink devices. It should be pretty straight-forward if you'd like to have a go.

___

At this point I'm thinking that the most diplomatic solution might be to make the screensaver behavior customizable by end-users via the app / cli. I've got this working tonight as a bit of proof of concept, but I'll have to talk to someone more knowledgable about whether it's an appropriate option, and about would it would take to expose the setting in the mobile apps.

jimwhitelaw:
I think it would be useful to have the node id remain visible on the screensaver. If you have multiple devices sleeping, you can determine at glance which is which.

todd-herbert:
> I think it would be useful to have the node id remain visible on the screensaver. If you have multiple devices sleeping, you can determine at glance which is which.

Good call! I'll look at getting that onto the icon screen, maybe replacing the version number?

___

The new plan is: offer users a choice in the display preferences.
* LOGO
  Similar to screen shown in original post

* OVERLAY
  Some sort of minimal "heads up", drawn over the existing (pre-pause) screen image. Proof-of-concept working, but haven't thought about a design yet.

* BLANK
  Just wipe the display
  
 * NONE
  No action taken (current behavior)

* CUSTOM
  A custom GFX method, only built if method is set in variant's platformio.ini build_flags.
  Supports vendors and DIYers to make their own aesthetic choices.
  ```cpp
  #define EINK_CUSTOM_SCREENSAVER helloWorld

  #ifdef EINK_CUSTOM_SCREENSAVER
  namespace einkCustomScreensavers
  {
    void helloWorld(OLEDDisplay *display)
    {
      display->setTextAlignment(TEXT_ALIGN_LEFT);
      display->drawString(30, 30, "Hello world!");
    }
  }
  #endif
  ```

___

Not confirmed yet though; will have to wait and see how the dev-team feel about it.

todd-herbert:
@jimwhitelaw Node ID, short name, or both?

jimwhitelaw:
I think any id that fits nicely on the display. Last 4 is fine, short name is possibly friendlier. As long a sleeping units can be identified, you’re golden.

todd-herbert:
Following feedback on https://github.com/meshtastic/protobufs/pull/474, scope of this pull-request in reduced. A single screensaver behavior will be implemented. Abandoned work-in-progress is mothballed as [eink-screensaver-configurable](https://github.com/todd-herbert/meshtastic-firmware/tree/eink-screensaver-configurable) branch.

After some time spent yesterday developing the different screensaver options, my personal opinion now is that the subtle overlay suggestion is best https://github.com/meshtastic/firmware/pull/3477#issuecomment-2016678721. This also meets the criteria that nodes should be identifiable with the screen off https://github.com/meshtastic/firmware/pull/3477#issuecomment-2018234455.

overlay

todd-herbert:
Tested Heltec Wireless Paper V1.0 and V1.1
No issues expected, but be aware that no testing has taken place with differentl resolutions / platforms.

GUVWAF:
I like the simple overlay, although depending on which screen it was showing, it may or may not show the ID of the current node. 

Formatting on the T-Echo looks good as well:
image

GUVWAF:
One more thing: if you turn it off while it was showing the screensaver, it will not show the "Sleeping..." screen.

todd-herbert:
> One more thing: if you turn it off while it was showing the screensaver, it will not show the "Sleeping..." screen.

Hmm, I was aware of that issue but thought I had caught it.. I'll take another look and see if I can figure a more robust way to solve it.

When you're testing, are you rebooting and then shutting down within roughly the first 30 seconds?

GUVWAF:
> When you're testing, are you rebooting and then shutting down within roughly the first 30 seconds?

No, it was more than that. I set the screen timeout on 60 seconds, so I waited at least that. I just tested with several minutes in between and it still happens.

> Do you have any suggestions here from a design point of view about where and how to display this info? Do you prefer "last 4" of node id, or short name?

To be honest for me displaying it on the screen is not so important, you can also mark the device itself with a sticker. I'm fine with the current approach.

todd-herbert:
> No, it was more than that. I set the screen timeout on 60 seconds, so I waited at least that. I just tested with several minutes in between and it still happens.

Thanks for that info. I *think* I might have a fix, but not sure just yet. I should actually have a T-Echo turning up in the next day or two, so I can check then whether I've solved it or not.

> To be honest for me displaying it on the screen is not so important, you can also mark the device itself with a sticker. I'm fine with the current approach.

I've modified the design slightly now to include short name, and I'm fairly please with how it looks. I'll push that in the near-future too.

___

There's a slight hold up on those final tweaks: I found a low-down bug in the system. When the screen class turns off, there are no more calls to `display()`, which is where the async-refresh code polls for completion. This means that when the screensaver is displayed, the panel's voltage generating hardware remains active.

A quick fix for this would be to use a blocking fast-refresh here, but I'd rather try tackle the problem at its source. I'm looking into inheriting the OSThread class, to allow polling after the screen is "turned off".

caveman99:
There is a variation of osthteads called the notified worker. It will not loop but execute a task when externally notified. Maybe this one is better suited.

todd-herbert:
@caveman99 
> There is a variation of osthteads called the notified worker.

Thank you for pointing me in that direction, that class feels very tidy! I did get stuck for a while trying to figure out why my enumeration 0 wasn't triggering `onNotify()`, but other than that all good.

todd-herbert:
overlay version two
Short name is displayed on overlay.

todd-herbert:
(that's it this time)
(I hope)

todd-herbert:
And for the third time tonight, I'm going to claim that everything is final now. Except that this time I'm going to bed so I *have* to leave it alone..

thebentern (approved):

todd-herbert:
Reference doc, for future maintenance:
Async Refresh (updated) | Refresh Selection (updated) | Screensaver
---|---|---
flowchart for async refresh | flowchart for the determine mode method | flowchart for E-Ink screensaver

PR #3486, opened 2024-03-24
Native: fail-safes for simulated node without config file

GUVWAF:
When trying to run multiple simulated nodes without config file on my Ubuntu system, I experienced two crashes. The first one was because `settingsMap[maxnodes]` was not defined. Second one was because LinuxInput was closing file descriptor 0 when rebooting and after boot `psock` in WiFiServer got assigned to 0, which lead to an assert.

I also set the default log level to debug.


jp-bennett (approved):
Yes, good catches!

PR #3487, opened 2024-03-24
Add Shutdown and reboot to CardKB and friends

caveman99:


tropho23:
Tested, works perfectly!

Tony Good

PR #3488, opened 2024-03-24
boards: remove -DSTM32WLxx from generic_wle5

Jorropo:
This define is no longer referenced anywhere.


caveman99:
Thank you for your contribution. We have a dedicated branch for developing the stm32wl target. Would you mind referencing this PR to that branch instead of master? Looks like by now that boards file is not even in the branch any more ...

Jorropo:
@caveman99 thx I guess you meant #2960, I was trying to compile for stm32l, I guess the stm32wl branch would be a better starting point.

PR #3490, opened 2024-03-25
Fixed position admin messages

thebentern:
Bespoke admin messages for setting or clearing fixed position in one operation.


thebentern:
Needs testing. I'll integrate it into the C# client soon and make sure it behaves as billed.

caveman99 (approved):

CamFlyerCH:
I am not sure if related, bit with the newest release 2.3.3.8187fa7 I am not able change the coordinates or height of my fixed position anymore.

thebentern:
Not related. This actually addresses that concern more cleanly for clients. In the previous version, we actually started enforcing fixed positions being static. You will need to turn off fixed position to change your node's position, and then turn it back on to have it remain "fixed".

PR #3500, opened 2024-03-27
Fixes to allow Docker build on ARM architectures

### LINKED ISSUE #3499: [Bug]: Linux native docker build fails on ARM
### Category

Other

### Hardware

Other

### Firmware Version

2.3.2

### Description

The alpine image in use has no ARM version causing ARM docker builds for Linux native to fail.

I have fixed locally and am raising a PR.

### Relevant log output

_No response_

jaredquinn:
Update both build image and final image to use debian bookworm (slim).

Final image had previously used a alpine-glibc build without ARM support and issues with libgpio binary builds for bullseye were encounted.

Additionally updated hard coded architecture references to calls to $(uname -m).

Fixes #3499


thebentern (approved):

jp-bennett:
I'll try to actually run it later, but it looks good to me.

jaredquinn:
I do also note that the docker build *always* pulls the same branch; regardless of which branch you're operating in.

Perhaps it should use the current checked out files to allow easier building for development?

thebentern:
> Perhaps it should use the current checked out files to allow easier building for development?

Yes. This has been discussed, and I think that is a better strategy than pulling from git.

Dockerfile
```diff
@@ -27,15 +27,23 @@ RUN wget https://raw.githubusercontent.com/platformio/platformio-core-installer/
...
-	groupadd -g 1000 mesh && useradd -ml -u 1000 -g 1000 mesh
+##### PRODUCTION BUILD #############
 
-COPY --from=builder /tmp/firmware/release/meshtasticd_linux_x86_64 /home/mesh/
+FROM debian:bookworm-slim
+ENV DEBIAN_FRONTEND=noninteractive
+ENV TZ=Etc/UTC
+
+RUN apt-get update && \
+    apt-get -y install libc-bin libc6 libgpiod2 libyaml-cpp0.7
```
jp-bennett, line 39: We might want to add libulfius-dev liborcania-dev and libssl-dev to enable the web interface. Not sure if that belongs in this PR, though.

jp-bennett (approved):

Dockerfile
```diff
@@ -27,15 +27,23 @@ RUN wget https://raw.githubusercontent.com/platformio/platformio-core-installer/
...
+RUN groupadd -g 1000 mesh && useradd -ml -u 1000 -g 1000 mesh
 
 USER mesh
 WORKDIR /home/mesh
-CMD sh -cx "./meshtasticd_linux_x86_64 --hwid '${HWID:-$RANDOM}'"
+COPY --from=builder /tmp/firmware/release/meshtasticd /home/mesh/
+
+CMD sh -cx "./meshtasticd -d /home/mesh/data --hwid '${HWID:-$RANDOM}'"
```
wnagele, line 46: I wouldn't hard-code a directory here. Instead leave that up to being changed through the configuration file for the ones that need it.

jaredquinn: This is pretty common practice within a container;  with the intention that a user mounts a docker volume on that path on container start.
 
```
docker run -d \
    -p 4403:4403 \
    --name meshtastic-node \
    -v /opt/meshtastic/data:/home/mesh/data \
    --log-opt mode=non-blocking --log-opt max-buffer-size=4m  \
    meshtastic:v2.3.2
```

The particular path was chosen because the application is ran as the mesh user in the container; so it made sense that the data directory was under the home.

If the volume isn't mounted to that path at start it will be empty;  used and discarded on container remove.

caveman99:
@jaredquinn can you address the errors from the trunk checks please?

jaredquinn:
> @jaredquinn can you address the errors from the trunk checks please?

@jp-bennett  suggested we also build webserver support by default into the docker image.   I've done this over in #3506;   which was built ontop of this initial PR - perhaps I should kill this one in favour of that one?

thebentern:
Sure. Sounds good

PR #3505, opened 2024-03-28
Don't reboot for non-radio lora config changes

### LINKED ISSUE #3504: [Feature Request]: Admin Message for ignore incoming that does not require a reboot
### Platform

NRF52, ESP32

### Description

I have been getting requests to add ignore incoming to the long press menu for a node / user, I think having functions that require a reboot in a context menu is a bad pattern, I would add this function if the update can be made without a reboot.

thebentern:
Closes #3504

### DIFF
--- a/src/PowerFSM.cpp
+++ b/src/PowerFSM.cpp
@@ -246,7 +246,6 @@ Fsm powerFSM(&stateBOOT);
 void PowerFSM_setup()
 {
     bool isRouter = (config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER ? 1 : 0);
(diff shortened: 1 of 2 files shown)


jp-bennett:
I like it. But I suspect we'll find edge cases for a while to come, where parts of the firmware is written with the assumption that settings never change.

thebentern:
> I like it. But I suspect we'll find edge cases for a while to come, where parts of the firmware is written with the assumption that settings never change.

I did a deep dive. Everything right now just reads from the updated values in lora config, so we should be fine. The only thing that _needs_ the reboot within this section are settings which affect the RadioLib init of the radio.

PR #3506, opened 2024-03-29
Native Linux Build (ARM support and webserver deps)

### LINKED ISSUE #3499: [Bug]: Linux native docker build fails on ARM

jaredquinn:
Changes to support Docker on ARM

* Use bookworm-slim rather than previous base image which was only supported on x64
* Copy resulting target architecture binary to meshtasticd
 
Changes to include webserver dependencies in default Linux Native builds.

* Update apt-install calls to include -dev and binary versions of the libraries in the appropriate builds
* Use current working directory for build rather than pulling down the current master branch from github and only building that.

Updated Dockerfile tested against v2.2.9 and current master.
Additionally fixes #3499 

As discussed with @jp-bennett  on discord.

### DIFF
--- a/arch/portduino/portduino.ini
+++ b/arch/portduino/portduino.ini
@@ -1,6 +1,6 @@
 ; The Portduino based sim environment on top of any host OS, all hardware will be simulated
 [portduino_base]
-platform = https://github.com/meshtastic/platform-native.git#1b8a32c60ab7495026033858d53c737f7d1cb34a
+platform = https://github.com/meshtastic/platform-native.git#117acc5e7fcc2047e9ba1dc11789daea26fc36d2
 framework = arduino
 
 build_src_filter = 
--- a/arch/portduino/portduino.ini
+++ b/arch/portduino/portduino.ini
@@ -35,8 +35,3 @@ build_flags =
   -lbluetooth
   -lgpiod
   -lyaml-cpp
-  -lssl
-  -lorcania
-  -lulfius
-
-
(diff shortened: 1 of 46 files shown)


arch/portduino/portduino.ini
```diff
@@ -34,4 +34,9 @@ build_flags =
   -DPORTDUINO_LINUX_HARDWARE
   -lbluetooth
   -lgpiod
-  -lyaml-cpp
\ No newline at end of file
+  -lyaml-cpp
+  -lssl
+  -lorcania
```
jp-bennett, line 39: The problem with this approach is that we would like to keep this buildable for distros that don't package orcania/ulfius. See https://github.com/meshtastic/firmware/blob/master/variants/portduino/platformio.ini#L6C4-L6C31

I'm surprised the inclusion of those lines in the variants/portduino/platformio.ini isn't sufficient.

jp-bennett (changes requested):

arch/portduino/portduino.ini
jaredquinn: This has been fixed in the PR now and those lines removed from portuino.ini
Confirmed my builds work successfully with webserver now.

jp-bennett (approved):

jaredquinn:
Added trunk-check ignores for terrascan/AC_DOCKER_0002; it currently seems to provide false-positives on RUN apt-get:

https://github.com/tenable/terrascan/discussions/977

Added ignores for hadolint/DL3008 - ideally we want the container to stay up to date with platform packages; I see no reason to pin build dependencies here.

PR #3509, opened 2024-03-29
Make native honor HAS_SCREEN 0

jp-bennett:


Jorropo (commented):
Works for me, thx.
I havn't checked the implications this have.

PR #3510, opened 2024-03-29
linux-native: only install linux native deps

Jorropo:
This is a couple times faster because platformio checks all environment sequentially.


jp-bennett:
While this looks sane to me, I'll leave the final decision to someone who knows pio's quirks better.

PR #3518, opened 2024-03-31
Handle edge cases for E-Ink screensaver

todd-herbert:
Improved handling of "special screens" (shutdown, reboot, bluetooth pairing)


todd-herbert:
My version of clang-format seems to disagree about the ordering of headers
trunk console output

Issue #3517, opened 2024-03-31
[Bug]: PowerFSM timed transition preventing light sleep


todd-herbert:
### Category

Other

### Hardware

~~Heltec Wireless Paper~~ All ESP32(?)

### Firmware Version

2.3.4.fd704728

### Description

Unable to enter light-sleep.

**Added some additional debug output**, which seems to show that device is kept awake by timed state transition at [PowerFSM.cpp L351-L353](https://github.com/meshtastic/firmware/blob/279464f96d5139920b017d437501233737daf407/src/PowerFSM.cpp#L351-L353)

As a test, disabling these lines did allow the device to enter stateLS as expected.

Not sure what the intended purpose of the transition is; catch-all for any missed state changes maybe?
Didn't notice any immediate consequences of disabling the dark-to-dark transition. Will do some more testing later so try and identify where, if anywhere, it has an effect.

Light sleep duration was set at 35 seconds during the testing.

### Relevant log output

```Shell
DEBUG | ??:??:?? 67 [Power] Battery: usbPower=0, isCharging=0, batMv=3867, batPct=67
DEBUG | ??:??:?? 71 [PowerFSM] Timed transition: Dark to Dark
DEBUG | ??:??:?? 71 [PowerFSM] darkEnter()
DEBUG | ??:??:?? 87 [Power] Battery: usbPower=0, isCharging=0, batMv=3905, batPct=71
DEBUG | ??:??:?? 101 [PowerFSM] Timed transition: Dark to Dark
DEBUG | ??:??:?? 101 [PowerFSM] darkEnter()
DEBUG | ??:??:?? 107 [Power] Battery: usbPower=0, isCharging=0, batMv=3913, batPct=72
DEBUG | ??:??:?? 127 [Power] Battery: usbPower=0, isCharging=0, batMv=3931, batPct=74
DEBUG | ??:??:?? 131 [PowerFSM] Timed transition: Dark to Dark
DEBUG | ??:??:?? 131 [PowerFSM] darkEnter()
DEBUG | ??:??:?? 147 [Power] Battery: usbPower=0, isCharging=0, batMv=3917, batPct=72
DEBUG | ??:??:?? 161 [PowerFSM] Timed transition: Dark to Dark
DEBUG | ??:??:?? 161 [PowerFSM] darkEnter()
DEBUG | ??:??:?? 167 [Power] Battery: usbPower=0, isCharging=0, batMv=3885, batPct=69
```

Issue #3512, opened 2024-03-30
[Bug]:Power Save mode no longer letting screen to wake


HarukiToreda:
### Category

Other

### Hardware

Heltec V3

### Firmware Version

2.3.2

### Description

I have a few standalone Heltec V3s with CardKBs that were recently upgraded to firmware 2.3.2. I use them on power save mode. When the screen goes to sleep at 10min I can use the program button to wake it with no issues, but after 15min it no longer wakes the screen when i press the Program button, I can confirm it is still on because the screen wakes up when I send a message to it. It seems something was changed on recent firmwares. I can confirm with another node that has 2.2.22 where it does wake up normally regardless how long the screen is off. 
I've tried to factory reset the node and only turned on Powersave mode and cardkb and it still happens on all nodes with this firmware.

garthvh:
The heltec devices may not be the best choice for sleep features

5kamisama5:
> The heltec devices may not be the best choice for sleep features

So you suggest i never upgrade past  firmware 2.2.22? Cause it works just fine there.

todd-herbert:
Is the device sleeping (LED heartbeat stopped), or is the screen just off?

5kamisama5:
> Is the device sleeping (LED heartbeat stopped), or is the screen just off?

LED heartbeat is still working, just the screen is off.

garthvh:
> > The heltec devices may not be the best choice for sleep features
> 
> So you suggest i never upgrade past firmware 2.2.22? Cause it works just fine there.

A board with battery detection that is finicky is not going to be a great fit for roles that sleep, sleep requires that the power / battery data be reasonably accurate.

5kamisama5:
> battery detection that is finicky is not going to be a great fit for roles that sleep, sleep requires that the power / battery data be reasonably accurate.

it is true it's not the best board but people have it and are using it. A feature that worked before and doesn't work now, means something was changed to break it. I've just taken the time to upgrade to each firmware to pinpoint when it broke. It's not popular among the PROs but people still have this board. 

2.2.22 - Still Works
2.2.23 - Still Works
2.2.24 - Broke Here (Screen not turning on after 15min when button pressed)
2.3.2  - Screen not turning on after 15min
2.3.3  - Screen not turning on after 15min

[...]

PR #3521, opened 2024-03-31
Fix Light-sleep for ESP32

todd-herbert:
Hopefully addresses https://github.com/meshtastic/firmware/issues/3512

Main culprit seems to have been the addition of [`enableLoRaInterrupt()`](https://github.com/meshtastic/firmware/blame/a4c22321fca6fc8da7bab157c3812055603512ba/src/sleep.cpp#L323), which is competing with the hardware button for use of the EXT0 wake source. Using GPIO wake for the button works around this issue.

Secondary issues with ISRs on wake, and PRESS event detection have also been addressed (kludged).

Tested with Heltec Wireless Paper, however this does currently require a hotfix for https://github.com/meshtastic/firmware/issues/3517 to first be applied. Will be separately address soon. Briefly tested with T-Echo. All seems to be running fine, at least at first glance.

I would expect this bug to affect all ESP32 devices(?). It would be good to have the fix independently tested, if anyone has the time / energy to spare.

### DIFF
--- a/src/PowerFSM.cpp
+++ b/src/PowerFSM.cpp
@@ -102,23 +102,18 @@ static void lsIdle()
                 powerFSM.trigger(EVENT_SERIAL_CONNECTED);
                 break;
 
+            case ESP_SLEEP_WAKEUP_GPIO:
+                // GPIO wakeup is now used for all ESP32 devices during light sleep
+                powerFSM.trigger(EVENT_PRESS);
+                break;
+
             default:
-                // We woke for some other reason (button press, device interrupt)
-                // uint64_t status = esp_sleep_get_ext1_wakeup_status();
+                // We woke for some other reason (device interrupt?)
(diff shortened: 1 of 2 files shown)


todd-herbert:
Seems likely that https://github.com/meshtastic/firmware/issues/3517 also applies to all ESP32 devices, not just Heltec Wireless Paper.

5kamisama5:
I can confirm that the program button on Heltec V3 now wakes the device from PowerSave mode as it used to. Thank you.
I plugged a multimeter to see when the different states happen:
When powered on device runs at 108mA
Screen goes off after 10min and runs at 104mA
After 11min it goes to PowerSave mode at 11mA
I tested the program button once on 11mA and it wakes up back to 108mA
I restested after 15, 20, 30min and it worked everytime.

I believe you fixed it.

thebentern (approved):

PR #3530, opened 2024-04-01
Draft PR to support new device unPhone

garethhcoleman:
Hello

I've been encouraged on Discord to put in this early PR, so I can get feedback without going too far wrong!

I have helped develop the unPhone ([website](https://unphone.net) [shop link](https://shop.pimoroni.com/products/unphone)). It's got a 480x320 touchscreen, RFM95W 868MHz LoRa, battery, accelerometer/compass, sd card, RGB and IR LED's, & vibration feedback. Now I am porting the rather fantastic Meshtastic firmware to our device.

I've got the screen, radio and sd card (seemingly) working in this PR. I had to make very minor changes to files in /graphics to use large fonts, a small bit of setup in main.cpp because our device has the backlight on a I2C expander. The main additions in this PR are the creation of a new variant.h and platformio.ini file for the new device, and a new driver configuration in graphics/TFTDisplay.cpp (using the existing LovyanGFX pattern).

I am requesting advice and guidance to make integrating these changes in the best way possible. For example, I believe the patterrn I have used for implementing HX8357 support for our device could be re-used by others.

Please accept my apologies if I've missed some guidance in preparing this PR - I am very keen to do this 'right' so just let me know if I should re-work anything etc!

We can supply hardware (gratis no obligations) to developers if that would be useful.


variants/unphone/platformio.ini
```diff
@@ -0,0 +1,17 @@
...
+extends = esp32s3_base
+board_level = extra
+board = unphone9
+upload_speed = 921600
+monitor_speed = 115200
+monitor_filters = esp32_exception_decoder
+
+build_flags = ${esp32_base.build_flags} 
+  -D PRIVATE_HW
+  -D UNPHONE
```
caveman99, line 12: Just a technicality - once this is finished, please register a proper device identifier in our protobufs [here](https://github.com/meshtastic/protobufs/blob/e6b4c590e7c489306c9c44e3ad1fcf62a3efd288/meshtastic/mesh.proto#L219). The PRIVATE_HW is only for development or unpublished ports. You can probably lose the UNPHONE define this way too.

src/main.cpp
```diff
@@ -588,7 +588,19 @@ void setup()
...
+    Wire.write(0x06);
+    Wire.write(0x7A);
+    Wire.write(0xDD);
+    Wire.endTransmission();
+    Wire.beginTransmission(0x26);
+    Wire.write(0x02);
+    Wire.write(0x04); // Backlight on
+    Wire.write(0x22); // G&B LEDs off
+    Wire.endTransmission();
+#endif
```
caveman99, line 603: I'd like to propose a wider scope for the use of port expanders in meshtastic. Which chip do you use exactly?
Suggestion would be love leave this as-is and repalce later with proper handling of port expanders.

Issues.txt
```diff
@@ -0,0 +1,10 @@
+Issues:                                     File to start looking at?
+Does Lora Radio actually work?!?
+Fix need for serial monitor                 ??
+Get touchscreen working                     graphics/TFTDisplay.cpp
+Use power switch                            Power.cpp
+Battery level measurement                   Power.cpp
+Utilise RGB led for notifications           modules/ExternalNotificationModule.cpp AmbientLightingThread.h
+Utilise Vibration for notifications         modules/ExternalNotificationModule.cpp
+Use accelerometer/compass                   Configuration.h, main.cpp, add library
+use IR leds                                 new code!
```
caveman99, line 10: Please make sure the Todo list is not in the final commit :-)

jp-bennett:
So far this is looking great. Support for a compass module is something we're already looking at adding, so some great cross-pollination there. It would probably be good to get a unit to a developer or two.

variants/unphone/platformio.ini
garethhcoleman: Ah yes, I wondered if I should do that now or later.... I'll go and put in a PR to that repo now. Thanks!

garethhcoleman: PR against the protobufs submitted [here](https://github.com/meshtastic/protobufs/pull/477)

Issues.txt
garethhcoleman: facepalm!

garethhcoleman:
> So far this is looking great. Support for a compass module is something we're already looking at adding, so some great cross-pollination there. It would probably be good to get a unit to a developer or two.

Thanks, sorry I was mistaken and it's only an accelerometer and gyroscope not a full compass, sadly.

We will email you to setup sending out a couple of units to developers.

garethhcoleman:
Thanks for the advice and suggestions, I will now close this draft and prepare another PR incorporating the changes.

PR #3539, opened 2024-04-03
Disable holds / isolations on RTC IO pads after deep sleep

todd-herbert:
> Anonymoushindeiru
 — 
Yesterday at 9:20 PM
I found a bug: after waking from manual sleep/poweroff on a nano G1 explorer, buzzer no longer works

___

On ESP32, on wake from deep sleep, releases RTC IO pads which were held or isolated to minimize current leakage, freeing them to resume their normal function.

### DIFF
--- a/src/graphics/EInkDisplay2.cpp
+++ b/src/graphics/EInkDisplay2.cpp
@@ -151,31 +151,12 @@ bool EInkDisplay::connect()
 
 #elif defined(HELTEC_WIRELESS_PAPER_V1_0) || defined(HELTEC_WIRELESS_PAPER)
     {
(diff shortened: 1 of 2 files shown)


todd-herbert:
> Talie5in
 — 
Today at 12:12 AM
Yup
Yeah hitting power button quickly does full boot
After long press for sleep
>
> todd-herbert
 — 
Today at 12:13 AM
Power button yes, usr button no?
> 
> Talie5in
 — 
Today at 12:14 AM
Correct

Reported that issue also affects tbeam-s3-core

thebentern (approved):

Issue #2539, opened 2023-06-01
[Board]: STM32WL (Seeed Studio Wio-E5)


GUVWAF:
### SOC

Other

### Lora IC

SX1262 (inside STM32WL)

### Product Link

https://www.seeedstudio.com/LoRa-E5-mini-STM32WLE5JC-p-4869.html

### Description

There is currently a variant of the firmware for this architecture with which the device boots and the radio initializes, but LittleFS is not working. See #2265.

Should be a very low-power device as it uses the same processor as NRF52 (single-core ARM Cortex-M4) running at lower clock frequency, no Bluetooth, and the radio is integrated on the same chip.

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/stm32wl55-mesh-lora/7820/2

Dr-Gandalf:
Is it really worth to develop this HW? It only has 256kB flash memory, if it ever works, it will be really limited, considering the that other ARM devices compiles to ~600kB binary file. It may be better to implement a stm32wb5mmg + sx1262 https://www.st.com/en/microcontrollers-microprocessors/stm32wb5mmg.html this will have Bluetooth and 1MB flash, just an idea.

thebentern:
It will be worthwhile as an ultra low-power repeater

PR #3540, opened 2024-04-03
Add configuration option for LoRa Region Code override for region-locked builds/variants

ndoo:
The main use case for this will be to create a custom Heltec WiFi LoRa 32 V3 SG_923 variant, which will be pre-flashed and sent for regulatory approval for retail sale.


ndoo:
Seems like this will also help with #2539 instead of hardcoding region in RadioInterface.cpp

PR #3541, opened 2024-04-03
Update GeoCoord.cpp

rishranyal:
Increases precision


thebentern:
Hi. We can't just change this in the firmware. It's is a breaking change. It would affect all clients.

Issue #3523, opened 2024-04-01
[Bug]: Heltec Wireless Paper deep-sleep current too high


todd-herbert:
### Category

Hardware Compatibility

### Hardware

Heltec Wireless Paper

### Firmware Version

2.3.4.fd704728

### Description

The deep-sleep current for Heltec Wireless Paper currently measures 2.2mA.
From personal experience, I know that the manufacturer's claims of 18µA are indeed achievable, but this requires a very particular set of pin settings / sleep code. I'll look into implementing these shortly.

PR #3545, opened 2024-04-04
Reduce deep-sleep current for Heltec Wireless Paper

todd-herbert:
Addresses https://github.com/meshtastic/firmware/issues/3523

At present, deep-sleep current is 2.2mA @ 3.8V.
Changes lower the current draw to 18µA @ 3.8V.

These modifications are likely to be applicable to a number of ESP32 devices, but I don't have the hardware to test this.


caveman99 (approved):
Can we test this PR on more devices before we accept it? This sounds like it's applicable to more than the wireless papers.

todd-herbert:
The code changes right now are intentionally limited to Wireless Paper, but I imagine many ESP32 / ESP32S3 devices could benefit here, if anyone with other hardware wants to check?

PR #3554, opened 2024-04-04
This change to the I2C Scan is to distinguish QMI8658 and BQ24295 

garethhcoleman:
Both these devices share the same I2C address (0x6B) the QMI8658 IMU and BQ24295 PMU.

I've got a BQ24295 to test and it seems to work but I am unable to test with a QMI8658. I don't imagine reading from register 0xA would cause any problems, but if would be really good if someone is able to verify this, also to check that the contents of it don't accidently match the version ID from the BQ24295!!

Fingers crossed I got the trunking formatted correctly this time...

### DIFF
--- a/src/detect/ScanI2C.h
+++ b/src/detect/ScanI2C.h
@@ -38,6 +38,7 @@ class ScanI2C
         MPU6050,
         LIS3DH,
         BMA423,
+        BQ24295,
 #ifdef HAS_NCP5623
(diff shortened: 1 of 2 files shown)


thebentern (approved):

caveman99 (approved):
Can't test either but looks straightforward. Can always release as alpha and wait for complaints rolling in :-)

Issue #3533, opened 2024-04-02
[Bug]: T-Echo double press


todd-herbert:
### Category

Other

### Hardware

T-Echo

### Firmware Version

2.3.4.a4c22321

### Description

Double press on T-echo is assigned to two different actions:
* Toggle backlight
* Send ad-hoc ping

thebentern:
Perhaps the capacitive touch button on top would be better suited to the backlight toggle, since the double press ad-hoc ping is available on any platform with a user button

todd-herbert:
That's certainly an idea that I can look into. I do remember hearing that there are some design issues with the touch button: that it is triggered every time a LoRa packet is transmitted, but I haven't confirmed this for myself yet.

todd-herbert:
@Nikto655 Currently it doesn't do very much.

___

After a bit or experimentation, I think the most user friendly option is to use the touch button as a modifier. 
 
Double press user button to send adhoc ping
Hold touch button + double press user button to toggle backlight.

This both aligns with standard meshtastic behavior, and remains somewhat familiar to T-Echo users.

I should have a pull request for this (and a few other related fixes) shortly.

b8b8:
Can we exclude the touch button completely please. the alternate cases i design for the t-echo are not compatible with it and ive pealed it off completely. Needing the touch button over the regular buttons severely limits future designs. Can i ask we add 4x button push for the t-echo light.

PR #3560, opened 2024-04-06
Remap backlight toggle and touch button

todd-herbert:
Addresses https://github.com/meshtastic/firmware/issues/3533

* Ensure E-Ink display updates after "adhoc ping" or "toggle GPS"

* Touch button wakes screen, disables screensaver (Includes TFT / OLED. Should narrow scope?)

* Re-map buttons for T-Echo
  &nbsp; | <nobr>Send Adhoc Ping</nobr> | <nobr>Toggle Backlight</nobr> | <nobr>Exit `stateDARK`</nobr> | <nobr>Force display update</nobr>
  ---|---|---|---|---
  Previous | Double Press (*user button*) | Double Press (*user button*) | -- | Press (*touch button*)
  New | Double Press (*user button*) | Hold (*touch button*) <br /> + Double Press (*user button*) | Press (*touch button*) | Press (*touch button*)

  Intent is to separate "adhoc ping" and "toggle backlight" behaviors.

___

I vaguely remember reading a comment that T-Echo touch has a hardware issue, where LoRa TX consistently triggers the touch button. I did not observe any cases of this during testing, but if would be good to hear from anyone who has more information.

### DIFF
--- a/src/ButtonThread.cpp
+++ b/src/ButtonThread.cpp
@@ -249,7 +249,9 @@ void ButtonThread::userButtonMultiPressed(void *callerThread)
 // Non-static method, runs during callback. Grabs info while still valid
 void ButtonThread::handleMultiPress()
 {
+#ifdef BUTTON_PIN
     multipressClickCount = userButton.getNumberClicks();
+#endif
 }
 
 void ButtonThread::userButtonPressedLongStart()
--- a/src/ButtonThread.cpp
+++ b/src/ButtonThread.cpp
@@ -180,8 +180,14 @@ int32_t ButtonThread::runOnce()
         }
         case BUTTON_EVENT_TOUCH_PRESSED: {
             LOG_BUTTON("Touch press!\n");
-            if (screen)
+            if (screen) {
+                // Wake if asleep
+                if (powerFSM.getState() == &stateDARK)
+                    powerFSM.trigger(EVENT_PRESS);
+
+                // Update display (legacy behaviour)
                 screen->forceDisplay();
+            }
             break;
         }
         default:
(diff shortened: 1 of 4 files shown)


b8b8:
@todd-herbert  Can we exclude the touch button completely please. the alternate cases i design for the t-echo are not compatible with it and ive pealed it off completely since all it did before was refresh the screen. Needing the touch button over the regular user buttons severely limits future designs. Can i ask we add 4x user button push for the t-echo light. The change you are proposing would render the light with alternate cases useless.

todd-herbert:
@b8b8 I wasn't aware that there was any sort of customization going like that going on. I certainly don't want to break existing set-ups. 

I do think there is value in making use of the touch button, but I will see if 4x user button presses is a viable alternative. It's clear that *some* solution has to implement which doesn't require use of the touch button. 

I will also look into using the "Wake screen on tap or motion" user-setting to entirely disable touch button functionality.

b8b8:
@todd-herbert great, thank you for understanding. The main issue is of course the backlight is pretty much exclusive to the t-echo and so is the touch button. 

https://www.thingiverse.com/thing:5028720
here is one of my cases. i have another one that is different but has the same restriction on touch button use.

todd-herbert:
&nbsp; | Send Adhoc Ping | Toggle Backlight | Exit `stateDARK` | Force display update
  ---|---|---|---|---
 <nobr>Previous</nobr> | Double Press (*user button*) | Double Press (*user button*) | -- | Press (*touch button*)
  Initial Suggestion | Double Press (*user button*) | Hold (*touch button*) <br /> + Double Press (*user button*) | Press (*touch button*) | Press (*touch button*)
   __Revised__ | __Double Press (*user button*)__ | __4x Press (*user button*)__ | __Press\* (*touch button*)__ | __Press\* (*touch button*)__

__\*__ only if user sets "Wake screen on tap or motion"

Proposed behavior affects any devices which have an E-Ink backlight, and any devices with a touch button. Touch function is disabled by default, enabled with "Wave screen on tap or motion" option in display settings.

thebentern (approved):

PR #3564, opened 2024-04-07
Detect USB power on Heltec boards

todd-herbert:
Implements a novel method for detecting USB power, on some Heltec devices which currently lack the ability.

These boards use a CP2102 USB to UART bridge. This IC is powered with `VCC_5V`, which is only supplied by USB. When USB is connected, and CP2102 is powered, it holds its TX pin HIGH (3.3V), as expected for UART. When USB is disconnected, and CP2102 is unpowered, its TX pin consistently reads LOW.

When `isVbusIn()` runs, it takes three digital readings of the MCU's UART RX pin, spaced 10ms apart. This is at attempt to avoid false negatives, in case CP2012 *is* powered, but MCU's RX pin is LOW due to a transmission in progress.

These readings do not seem to disrupt UART transmission, but if anyone has specific knowledge of this area and has concerns, please do comment.

Tested with Heltec Wireless Paper.
After reviewing the schematics, I am assuming that this technique is applicable to (and suggesting implementing for) the following variants:
* Heltec Wireless Paper
* Heltec V3
* Heltec Wireless Stick Lite V3

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -330,6 +330,20 @@ class AnalogBatteryLevel : public HasBatteryLevel
             return true;
         }
         // if it's not HIGH - check the battery
+#elif defined(EXT_POWER_DETECT_CP2102)
+        bool rxHeldHigh = false;
+
+        // Check (several times) to see if UART RX is held HIGH
+        // CP2102 is powered by VCC_5V. If unpowered, pin is LOW
+        for (uint8_t i; i < 3; i++) {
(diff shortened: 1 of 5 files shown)


caveman99 (approved):

todd-herbert:
When unpowered, CP2102 TX pin acquires gradual charge. Looking into non-intrusive ways to dissipate this.

todd-herbert:
This one might be a lost cause. I'm going to have to have a rethink.

Issue #2586, opened 2023-07-03
[Bug]:  BME680 - saving state of the IAQ calculation fails


rpsainio:
### Category

Other

### Hardware

T-Beam

### Firmware Version

2.1.17.7ca2e81

### Description

This is a TBeam with the latest FW installed (not updated) with web installer. Once the initialization is over and the status is 3, FW is supposed to write the state of the BSEC calculation on persistent storage so that after next reset/reboot the starting of IAQ calculation goes faster. Unfortunately it seems that this does not succeed. 
To verify this I made a reset afterwards and logs show that state file is not present.

Besides there seems to a typeconversion error in 
BME680 state update every 21600000 minutes

### Relevant log output

```Shell
DEBUG | 16:45:46 159618 [EnvironmentTelemetryModule] BME680 state update IAQ accuracy 3 >= 3
INFO  | 16:45:46 159618 [EnvironmentTelemetryModule] BME680 state write to /prefs/bsec.dat.
[159618722][E][vfs_api.cpp:105] open(): /littlefs/prefs/bsec.dat.tmp does not exist, no permits for creation
[159618723][E][vfs_api.cpp:137] rename(): /prefs/bsec.dat.tmp does not exists
ERROR | 16:45:46 159618 [EnvironmentTelemetryModule] Error: can't rename new state file
.................
DEBUG | 16:55:46 160218 [EnvironmentTelemetryModule] BME680 state update every 21600000 minutes
INFO  | 16:55:46 160218 [EnvironmentTelemetryModule] BME680 state write to /prefs/bsec.dat.
[160218756][E][vfs_api.cpp:105] open(): /littlefs/prefs/bsec.dat.tmp does not exist, no permits for creation
[160218757][E][vfs_api.cpp:137] rename(): /prefs/bsec.dat.tmp does not exists
ERROR | 16:55:46 160218 [EnvironmentTelemetryModule] Error: can't rename new state file

.................reset

INFO  | ??:??:?? 12 [EnvironmentTelemetryModule] Environment Telemetry: Initializing
[ 12687][E][vfs_api.cpp:105] open(): /littlefs/prefs/bsec.dat does not exist, no permits for creation
INFO  | ??:??:?? 12 [EnvironmentTelemetryModule] No BME680 state found (File: /prefs/bsec.dat).
INFO  | ??:??:?? 12 [EnvironmentTelemetryModule] Init sensor: BME680 with the BSEC Library version 2.4.0.0
[...]

PR #3580, opened 2024-04-09
add BME680 IAQ reading

caveman99:
Range is from 0 (clean) - 500 (extremely polluted)


thebentern:
BBBBBBBBBSEC

GUVWAF:
We might want to check if #2586 is still an issue.

PR #3585, opened 2024-04-09
Only set NodeNum based on MAC if it's still zero

GUVWAF:
Seems there is a batch of RAK4631 going around that have (nearly) the same MAC address, causing nodes to choose a random new NodeNum when it detects another in its DB. However, it's currently doing this at every reboot: https://meshtastic.discourse.group/t/android-app-duplicate-nodes/11806 

This reverts this piece of code to the original behaviour before it was removed in a failed attempt (#2567) to fix the "nodeNum = 4" issue.


geeksville:

https://meshtastic.discourse.group/t/android-app-duplicate-nodes/11806/2

PR #3590, opened 2024-04-10
fix: device sleep (part 1)

mverch67:
This PR fixes some issues introduced in #3251 and does a few debug cosmetics:

The function _enableLoraInterrupt()_ assumes that all LORA/DIO1/CS/RESET GPIOs are **RTC GPIOs** which is not true (only RTC GPIOs can be used for ext0).
As a result there are error messages in the log and all devices where the DIO1 pin is not an RTC GPIO will fail to wake up from sleep when a LoRa packet is received (one packet is received though, but processed on the next wakeup).

```
E (221030) RTCIO: rtc_gpio_pulldown_en(129): RTCIO number error
E (72312) RTCIO: rtc_gpio_pullup_en(109): RTCIO number error
E (72313) RTCIO: rtc_gpio_pullup_en(109): RTCIO number error
```

This PR replaces the RTC GPIO functions by generic functions which check for RTC GPIO internally and does a check for the validity of RTC GPIOs for DIO1.

### DIFF
--- a/src/sleep.cpp
+++ b/src/sleep.cpp
@@ -152,7 +152,7 @@ void initDeepSleep()
     // If waking from sleep, release any and all RTC GPIOs
     if (wakeCause != ESP_SLEEP_WAKEUP_UNDEFINED) {
         LOG_DEBUG("Disabling any holds on RTC IO pads\n");
-        for (uint8_t i = 0; i <= 45; i++) {
+        for (uint8_t i = 0; i <= GPIO_NUM_MAX; i++) {
             if (rtc_gpio_is_valid_gpio((gpio_num_t)i))
                 rtc_gpio_hold_dis((gpio_num_t)i);
         }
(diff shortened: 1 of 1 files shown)


thebentern (approved):

mverch67:
As the PR re-introduces GPIO wakeup source for all hardware devices that don't have RTC GPIO DIO1 some code changes from #3521 had to be reverted (to avoid screen turn on with every single packet).

src/sleep.cpp
```diff
@@ -428,20 +447,34 @@ bool shouldLoraWake(uint32_t msecToWake)
...
 #elif defined(LORA_DIO1) && (LORA_DIO1 != RADIOLIB_NC)
-    gpio_wakeup_enable((gpio_num_t)LORA_DIO1, GPIO_INTR_HIGH_LEVEL); // SX126x/SX128x interrupt, active high
+    if (radioType != RF95_RADIO) {
+        LOG_INFO("setup LORA_DIO1 (GPIO%02d) with wakeup by gpio interrupt\n", LORA_DIO1);
+        gpio_wakeup_enable((gpio_num_t)LORA_DIO1, GPIO_INTR_HIGH_LEVEL); // SX126x/SX128x interrupt, active high
+    }
 #endif
-#ifdef RF95_IRQ
-    gpio_wakeup_enable((gpio_num_t)RF95_IRQ, GPIO_INTR_HIGH_LEVEL); // RF95 interrupt, active high
+#if defined(RF95_IRQ) && (RF95_IRQ != RADIOLIB_NC)
```
thebentern, line 473: Good call. Technically we have targets like the T-Beam that have RF95 macros for the SX1276 and also SX126X ones, so runtime discrimination probably heads off some potential issues there hopefully.

mverch67: yes, I've tested that already. Working as expected.

PR #3591, opened 2024-04-10
Ignore MQTT by default and then don't transmit downlink over LoRa

GUVWAF:
This implements what has been discussed up till now in https://github.com/meshtastic/rfcs/pull/9 and thus needs to be approved first.
The "ignore MQTT" LoRa setting will be enabled by default and this now also means if you have MQTT downlink enabled, it doesn't transmit these packets over LoRa. MQTT-only will work as before and if you also want your mesh to receive the downlinks, you have to explicitly disable "ignore MQTT".

Tested both that MQTT downlink works and also ignoring MQTT works as before.

### DIFF
--- a/src/mesh/FloodingRouter.cpp
+++ b/src/mesh/FloodingRouter.cpp
@@ -43,7 +43,8 @@ void FloodingRouter::sniffReceived(const meshtastic_MeshPacket *p, const meshtas
     }
     if ((p->to != getNodeNum()) && (p->hop_limit > 0) && (getFrom(p) != getNodeNum())) {
         if (p->id != 0) {
-            if (config.device.role != meshtastic_Config_DeviceConfig_Role_CLIENT_MUTE) {
+            if (config.device.role != meshtastic_Config_DeviceConfig_Role_CLIENT_MUTE &&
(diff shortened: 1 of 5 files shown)


tropho23:
Wonderful!  I just submitted a feature request with this exact suggestion, and tagged your PR as a probably solution to get it some attention.

thespillmonkey:
with this change, if disabled on the radio connected to the broker, after an mqtt packet makes it's hop from the broker to the radio and gets transmitted to the mesh, will it then continue to hop just like a normal packet, or will it continue relaying on other radios if they have this enabled? Or is it "flagged" in some way as an mqtt packet and wont be relayed?

garthvh (approved):

fifieldt:
Is this still needed after dontMQTTMebro?

GUVWAF:
This is the other way around: packets from MQTT (downlink) can be controlled to not be transmitted over LoRa. Just like the public MQTT server is currently doing with zero-hopping, but then enforced on the firmware, such that it also holds for private MQTT servers.

ProjectMoon:
I've been searching for this exact feature after testing MQTT today. All discussion leads to this PR and then just kinda stops. Why was this closed? What was the end result? Is there a different way to do this?

Issue #3555, opened 2024-04-05
[Feature Request]: Add support for LSM6DS3TR-C 6 axis accelerometer / gyroscope


garethhcoleman:
### Platform

ESP32

### Description

As used on the unPhone ([details](https://unphone/net) [shop](https://shop.pimoroni.com/products/unphone)) I'd like to introduce this sensor to the Meshtastic firmware.

I've started work and feel reasonably confident it will fit in fairly easily but I'd appreciate any feedback / suggestions you can offer.

There are (at least) two libraries that I'm looking at using, the [adafruit one](https://github.com/adafruit/Adafruit_LSM6DS) and one by [chip makers ST](https://github.com/stm32duino/LSM6DS3). I haven't got shaking working yet using the Adafruit one and considering trying the other one, again advice welcomed.

caveman99:
i'd try the adafruit first since many other sensors use the lib stack and this way you avoid code duplication.

jaredquinn:
I believe there's a discussion here to have around where this sensor data should go?

I see arguments for and against both as position related data and as telemetry (though not necessarily Environmental Telemetry).

garethhcoleman:
I would say our primary or at least initial use case would be to use the
accelerometer for tap to wake or wake on pickup etc.

Telemetry wouldn't be the obvious fit with the unPhone as the device is
more targeting hand held than install up a pole.

Even though it might not be a great fit I wanted to find out if it was
considered best practice to include a telemetry aspect to interface with
that module.

caveman99:
raw telemetry is ruled out for any high-rate sensor like these. We only support slow-changing data or data with is averaged over time. Use cases for this type of sensor is heading while stationary (add-on to position module) or the tap-modes already present. Another use-case would be threshold based, so strong deltas would indicate a emergency notification needs to be send.

garethhcoleman:
Thanks for the info, seems to make the way forward more straightforward.

garethhcoleman:
PR [3593](https://github.com/meshtastic/firmware/pull/3593) submitted

PR #3593, opened 2024-04-11
LSM6DS3TR-C support

garethhcoleman:
This PR introduces support for the LSM6DS3TR-C Accelerometer and gyroscope.

Adafruit library is used as this integrates well with support for other sensors also using Adafruit libraries. Thanks Adafruit!!

fixes [3555](https://github.com/meshtastic/firmware/issues/3555)

After a very brief set of tests I've set the shake threshold to 5 as this wakes up the sleeping screen on pickup without actually needing to shake the device. This might need adjustment as others test, so I've exposed that control and set a default of 20 if it isn't defined.

Please let me know any issues mistakes improvements etc


garethhcoleman:
I've just re-read a compiler warning I missed before:

"src/AccelerometerThread.h:119:71: warning: ?: using integer constants in boolean context, the expression will always evaluate to 'true' [-Wint-in-bool-context]
lsm.enableWakeup(config.device.double_tap_as_button_press ? 2 : 1, 1, LSM6DS3_WAKE_THRESH);"

So I think that hold this PR for a few min until I can adjust the line so the config is respected. Thanks!

garethhcoleman:
OK fixed my mistake, good to go now

src/AccelerometerThread.h
```diff
@@ -5,6 +5,7 @@
 #include "power.h"
 
 #include <Adafruit_LIS3DH.h>
+#include <Adafruit_LSM6DS3TRC.h>
```
hamishcunningham, line 8: should there be a guard here, e.g. `#ifdef UNPHONE` so that devices that don't have that model of accel don't get the dependency?

garethhcoleman: Well maybe there should, but seems like the others didn't get one, and I was just tryin' to fit in...

garethhcoleman: In fact seems like all the sensors need to get defined in a platformio.ini somewhere outside of board variants... Building is failing for all the other boards! I'll be able to fix this later today

garethhcoleman: Of course! It needs to be like this so that folks can add a sensor after compilation and it just works...

thebentern: We definitely want this to be a non-gated peripheral option, especially since it's adafruit. Let's make it globallly available

garethhcoleman:
Yup totally get it, I'm just away from my dev environment today and can't
move the lib_deps from our variant to the main platformio.ini.

I'll be able to push a correction in a few hours.

Thanks again!

garethhcoleman:
get back up to date quick before another commit!!

garethhcoleman (commented):
I think it's ready to merge at your convenience

thebentern (approved):

PR #3597, opened 2024-04-11
T-Beam S3 core - enable USB CDC

oseiler2:
The current flag 'ARDUINO_USB_MODE=0'  disables USB CDC, and enables USB OTG (which isn't used as far as I can see).
Changing to 'ARDUINO_USB_MODE=1' allows firmware updates via USB (eg the web flasher) without having to change into download mode first by pressing the BOOT button during power on/reset.

see https://thingpulse.com/usb-settings-for-logging-with-the-esp32-s3-in-platformio/
This is in line with other ESP32-S3 board definitions, eg https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/boards/t-beams3-supreme.json


oseiler2:
image

thebentern (approved):

ndoo:
This change is causing my T-Beam Supreme to not boot until the serial terminal is accessed, when connected to a USB host.

ndoo:
I believe, this is what is blocking. 

https://github.com/espressif/arduino-esp32/issues/7554

oseiler2:
Apologies, I can reproduce this issue now. Will try to find a fix.
@thebentern please revert this change for now if it's causing wider issues.

oseiler2:
@ndoo could you please have a look at my PR #3652 and see if that fixes the problem for you?

mrekin:
> @ndoo could you please have a look at my PR #3652 and see if that fixes the problem for you?

Little bit offtopic, but  about same problem with esp32C3. 
CDC starts working fine with
`ARDUINO_USB_MODE=1
ARDUINO_USB_CDC_ON_BOOT=1`
and the latest (2.0.15) arduino platform. With the current used 2.0.14 not working.
Just tested (but here a lot of new warnings while build)

garethhcoleman:
We ran into this problem very recently. Good comment here helps explain: https://github.com/espressif/arduino-esp32/issues/7554#issuecomment-1969614096

We ended up setting cdc_on_boot to 1 and usb_mode to 0. I saw quite a few other esp32 S3 boards sharing this config.

Side benefit is that our usb vid:pid get passed to the host, so our board gets identified instead of just a generic JTAG debugger etc.

ndoo:
> @ndoo could you please have a look at my PR #3652 and see if that fixes the problem for you?

No longer blocks boot when connected to my PC. However, do look at the various replies above, perhaps there is a better combination of flags to use.

PR #3544, opened 2024-04-03
New device unPhone using HX8357D LCD and XPT2046 touchscreen

garethhcoleman:
The unPhone ([device](https://unphone.net) [shop](https://shop.pimoroni.com/products/unphone)) has a phone form factor, 480x320 touchscreen, battery, accelerometer, RGB and IR leds and a vibration motor for haptic feedback.

These changes are to implement support for HX8357D based devices (using LovyanGFX library) and XPT2046 touchscreens (WIP) as used on the unphone.

I hope I have done everything correctly to allow you to easily review these changes. Just let me know if I've missed anything or misunderstood, thanks!

PR #3600, opened 2024-04-11
fix/enhancement: TFT device powersave (part 3)

mverch67:
This PR aims to reduce the power consumption for TFT devices such as T-Deck and T-Watch-S3 (**but not heltec tracker**).

In powersave mode the following savings are achieved:
**T-Deck: 130mA (screen on, BT on) -> 30mA (screen dark, light sleep)
T-Watch: 130mA (screen on, BT on) -> 12mA (screen dark, light sleep)**

_-> values are measured in idle mode, not during LoRa sending!_

Prior to this PR the idle power consumption has always been 120-130mA, even with dark display. The powersave option wasn't functional.

During light sleep the T-Deck touch and keyboard are disabled. LoRa radio is still active. Wakeup is done using the trackball. For T-Watch the wakeup is done by touch.

To achieve the best power saving results for the user powersave is now default for T-Deck and T-Watch and some other parameters have changed defaults. The powersave mode can easily be disabled if not desired for whatever reason.
Default screen timeout is now 30 seconds and additonal 30 seconds to go into light sleep with disabled bluetooth. To enable bluetooth communication either tab on the screen before the 60 seconds have passed or press the trackball. The Android App automatically re-connects seemlessly with a disconnected device.

To achieve even further powersavings bluetooth can be disabled manually (saves ~40mA during operation and device goes faster into light sleep). In this case the device configuration can still be managed when using the admin channel over a second device. Works best with pre-installed canned messages, especially for T-Watch.

This PR relies on my two other PRs (#3590 part1 and #3596 part2).

Note: _the unPhone support in #3544 uses a proprietary implementation for the power handling and does not make used of lovyanGFX's unified Light_PWM interface (which is the core functionality of this PR). I believe it achieves basically the same, but I don't have such device to test it. For sure, the powersave mode will also work with the unPhone, though it's not specifically mentioned here._

### **NOTE: **
**To benefit from the new optimized default settings it'recommended to do a flash erase during installation**


gordontantdotcom:
This is great. The idea to me sounds perfect. It would also help if a long press of the trackball makes it a forced action ? 

this is my first venture into GitHub. Is there a download to test this ?

mverch67:
> this is my first venture into GitHub. Is there a download to test this ?

There's a bug in github actions. It fails to upload the .zip artifact for installation.

src/graphics/TFTDisplay.cpp
```diff
@@ -556,16 +556,12 @@ void TFTDisplay::sendCommand(uint8_t com)
         display(true);
         if (settingsMap[displayBacklight] > 0)
             digitalWrite(settingsMap[displayBacklight], TFT_BACKLIGHT_ON);
-#elif defined(ST7735_BL_V03)
```
thebentern, line 559: Do we need some testing on the Heltec trackers?

mverch67: That would be good. I'm just compiling for the 1.0 version. Results in a minute.

mverch67: Just noticed that after wake up from light sleep the BT is not turned back on. Not sure if this is a heltec only thing. But maybe unrelated to this PR, looks like problem in PowerFSM.

mverch67:
Heltec Tracker V1.0 with powersave enabled: 
Before: 106mA (BT on, Screen on) -> 112mA (BT on, screen blank) -> not functional
With PR: 106mA (BT on, Screen on) -> 112mA (BT on, screen off) -> 23mA (BT off, light sleep)

src/graphics/TFTDisplay.cpp
mverch67: I think I found the bug... it never worked :)

mverch67: Oh I was wrong... it did work but the change in #3596 / #3545 broke it.

mverch67: No, yes, no.... not working.... setBluetoothEnable() in main-esp32 is doing nothing when re-enabling.

thebentern (approved):

thebentern:
Looks good to me

PR #3615, opened 2024-04-14
Add more useful error output in radio interfaces

jp-bennett:


src/mesh/SX126xInterface.cpp
```diff
@@ -279,7 +287,8 @@ template <typename T> bool SX126xInterface<T>::isChannelActive()
     result = lora.scanChannel();
     if (result == RADIOLIB_LORA_DETECTED)
         return true;
-
+    if (result != RADIOLIB_ERR_WRONG_MODEM)
```
GUVWAF, line 290: Hmm, the result will most likely be `RADIOLIB_CHANNEL_FREE`, so probably not a good idea to always return an error here: 
https://github.com/jgromes/RadioLib/blob/46ef20ebe9c671da35a9a7a02f59d1498eb2bf61/src/modules/SX126x/SX126x.cpp#L775

jp-bennett: I just followed the existing code logic. Feel free to open a PR. =)

jp-bennett: Oh! I got the logic backwards! Whoops

PR #3638, opened 2024-04-16
Add multiple SPI devices for Radio, Display, and Touchscreen

jp-bennett:
This changeset gives us the ability to specify a separate SPI device for the LoRa, Display, and Touchscreen. The changes in Portduino also add support for specifying a new SPI speed for each transaction. All together, this means that we can let the Linux OS manage the CS lines, and also get much faster SPI speeds, leading to better framerates.


thebentern (approved):

mverch67:
Some general comments:
- it would be good to have detailed information about what's going on here. There is no single line of description provided (not only) to this PR
- the entire bus_spi class from lovyanGFX was copied into the firmware folder without providing any insight why this was necessary and what has been changed. Preferably you'd derive from bus_spi instead of IBus  and overwrite virtual methods where necessary (including comments what's changed) rather than doing copy&paste. Furthermore in C++ the filename should be identical with the class name (bus_spi.h vs class Mesh_Bus_SPI).
- when migrating PORTDUINO to device-ui support the bus_spi changes will become superfluous anyway

jp-bennett:
I think this is in a decent spot. I've tested it with 

> Some general comments:
> 
> * it would be good to have detailed information about what's going on here in this PR. There is no single line of description provided (not only) to this PR
That's fair. Will add.

> * the entire bus_spi class from lovyanGFX was copied into the firmware folder without providing any insight why this was necessary and what has been changed. Preferably you'd derive from Bus_SPI instead of IBus  and overwrite virtual methods where necessary (including comments what's changed) rather than doing copy&paste. Furthermore in C++ the filename should be identical with the class name (bus_spi.h vs class Mesh_Bus_SPI).
I don't know that the current upstream Lovyan makes enough of the class available to do this. Most notably, _cfg is private instead of protected. I'll do the rename.

> * when migrating PORTDUINO to device-ui support the bus_spi changes will become superfluous anyway
The old UI isn't completely going away, is it? I'd like to keep it accessible on Native also, as a development tool.

PR #3640, opened 2024-04-16
No more power-state messages printed to screen (?)

todd-herbert:
Suggestion: remove code which prints "Powered.." and "Unpowered.." to device screens.

This pull-request is submitted as "very much open for discussion".

Several days ago, on the discord server, there was a small discussion about the usefulness of these message. The shared opinion was that they may not be particularly useful to users, and that they have a tendency to appear at somewhat arbitrary times.

If anyone has an attachment to this feature, or feels that they have an important use, it would be good to hear from you!


thebentern:
Good riddance!

PR #3645, opened 2024-04-16
Cosmetics: rename remaining plugins → modules and less errors

GUVWAF:
- Found some remaining references to "Plugin", which I renamed to "Modules". 
- An implicit ACK is not an error, so let's not print that (it will print "generating implicit ack" before that).
- Also for SX128x and SX127x, `CHANNEL_FREE` result is not an error.


thebentern (approved):

thebentern:
Glad you got that implicit ack one. It was driving me nuts!

PR #3647, opened 2024-04-16
Support for generic 4pin RGB LEDs (both CC and CA) and NeoPixels, also RGB LED and vibration notification support for unPhone

garethhcoleman:
This PR introduces support for generic 4 wire RGB leds both Common Cathode and Common Anode types. Support is also provided for NeoPixels. Also RGB  and vibration functions in Meshtastic to the unPhone.

Ambient lighting and external notifications are supported.

All the code is macro guarded (#ifdef etc) so should have no impact on firmware size for others etc.

To maximise code-reuse I removed #ifdef guards around the NCP5623 in DeviceType - hope that seems fitting let me know if not. Edit - now reversed! Edit - double reversed, as in reinstated!

Also, perhaps controversially, I put the RGBLED_CA (common anode) in as a DeviceType despite it flagrantly not being an I2C device at all! again just because it made the code a bit simpler elsewhere. Edit - also reversed!

Let me know if I've broken a good rule and I'll rework it accordinly.

By request I went through variants and tried to implement RGB led functions where these had been half completed.


garethhcoleman:
Oh forgot to say, if there is an easy way to make this code more usable by other folks outside of the unPhone let me know

thebentern:
I'd like to keep the I2C device types true to it's name. If we could just use macros to gate the behavior, that would ideal. 

Even better would be to create a generic HAS_RGB macro that could be controlled for any target with an RGB led on pre-defined gpio in the variant. It looks like you have some of that behavior abstracted in the unphone library though, so we would need to pull some of that into Meshtastic to be shared with other consuming variants.

thebentern:
> The wrinkle with the unphone is that two of the RGB led pins are on the i2c io expander, so we can't just use digitalwrite...

Oooooh! That makes more sense :-)

garethhcoleman:
OK so good news is that it was fairly trivial to add support for RGB led's in the code...
Bad news is I don't have my stash of electronics with me right now and so I can't actually test it.

So please mark this as draft until we can get a real world confirmation, I mean it looks kinda right to me and I tried to make sure I didn't miss anything but, there's no test like a real test...

Usage would be something like this in variant.h:

// RGB LED configuration
#define RGBLED_RED 13 // GPIO pin connected to the red pin of the RGB LED
#define RGBLED_GREEN 14 // GPIO pin connected to the green pin of the RGB LED
#define RGBLED_BLUE 15 // GPIO pin connected to the blue pin of the RGB LED
#define RGBLED_CA // comment out this line if you have a common cathode type, as defined use common anode logic

garethhcoleman:
Hmm that's odd, it compiles in the heltec-wireless-paper environment for me... Is this comment enough to restart the CI checks?

thebentern (approved):

garethhcoleman:
Ok I've finally been able to test and generic 4 pin LEDs seem to work as expected.
IMG_20240422_011828
Ready for review and potential merging now.

garethhcoleman:
I've added support for [NeoPixels](https://learn.adafruit.com/adafruit-neopixel-uberguide) and tested them very briefly!

Usuage would be like this:
`// RGB LED configuration
#define HAS_NEOPIXEL                       // Enable the use of neopixels
#define NEOPIXEL_COUNT 1                   // How many neopixels are connected
#define NEOPIXEL_DATA A0                   // gpio pin used to send data to the neopixels
#define NEOPIXEL_TYPE (NEO_GRB+NEO_KHZ800) // type of neopixels in use
//   NEO_KHZ800  800 KHz bitstream (most NeoPixel products w/WS2812 LEDs)
//   NEO_KHZ400  400 KHz (classic 'v1' (not v2) FLORA pixels, WS2811 drivers)
//   NEO_GRB     Pixels are wired for GRB bitstream (most NeoPixel products)
//   NEO_RGB     Pixels are wired for RGB bitstream (v1 FLORA pixels, not v2)
//   NEO_RGBW    Pixels are wired for RGBW bitstream (NeoPixel RGBW products)`

garethhcoleman:
OK so I've gone over various boards that seemed to have RGB leds and incomplete support and tried to get them working with this new approach. I have none of these available to test with.

Boards that should now be supported and **just work** are:
betafpv_2400_tx_micro
Dongle_nRF52840-pca10059-v1
esp32-s3-pico
lora_relay_v1
lora_relay_v2
my_esp32s3_diy_eink
my_esp32s3_diy_oled

caveman99 (approved):

PR #3648, opened 2024-04-16
Added one minute throttling to NodeDB save to disk

thebentern:
This is to prevent the hammering of the file system on large meshes (public mqtt included). I ran a test earlier on the msh/US root and experienced 13 nodedb saves to disk within about a one minute period. It is likely responsible for some of the issues with nodes "loosing settings". 

I have plans for a more uses of the generic throttle methods. A follow up PR to refactor smart position will utilize it.


GUVWAF (approved):
Looks good, I like the modular approach.
Another low-hanging fruit would be to just don't save new nodes anymore when the NodeDB is full.

IMMORTAL0018:
Can the throttling be responsible for some position jumping while doing live tracking using mesh network with LoRa. During my test i found some jumping in the position of nodes while tracking. I have attached the video ..

PR #3666, opened 2024-04-19
Reduce blink interval to 30s in power saving mode, to save power

ndoo:
Picked 30 seconds as it felt like the limits of patience for anyone to stare at their node, while having significantly lower LED power consumption than at 1 second.

Refactored slightly and moved duration definitions to configuration.h.

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -140,6 +140,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 // -----------------------------------------------------------------------------
 // LED
 // -----------------------------------------------------------------------------
+#define LED_BLINK_TIME 1
+#define LED_BLINK_INTERVAL 1000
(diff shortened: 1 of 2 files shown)


garthvh:
Does this actually save any power?  Seems pretty infrequent.

ndoo:
> Does this actually save any power? Seems pretty infrequent.

I don't have the equipment to measure unfortunately, probably need an oscilloscope to measure because of the short blink. But I did this thinking it would be a good intermediate step between desoldering the LED, and adding a user configurable dark mode.

Maybe it's better to abandon this PR and actually work on letting the user disable the blink entirely 😅

garthvh:
It is not really going to save an appreciable amount of power, not having any indication that the device is working is not a great UX.

garthvh:
There are led's outside of the control of the firmware on every device, so dark mode is never going work really.

ndoo:
Closing in favor of #3674.

PR #3667, opened 2024-04-19
Add M5Paper support

fonix232:
## New device: M5Paper

I've had a few of these lying around for a long time, and given the form factor, I thought they'd be perfect for Meshtastic.

Sadly, the official M5Stack LoRaWAN Unit, which uses the ASR6501 radio over UART, can't be supported, so I'm a bit stuck on what kind of LoRa radio to attach to this.

Nonetheless I thought it would be nice to get some initial reviews on the parts I did finish.

## Notes

1. This will obviously not boot as-is due to the hard requirement for a LoRa radio. To work this around, any reviewer will need to [change this line](https://github.com/meshtastic/firmware/blob/master/src/main.cpp#L921) and replace it with a non-critical log (e.g. `LOG_DEBUG("LoRa Radio not found!\n");`). I've [started a thread on the forum](https://meshtastic.discourse.group/t/best-way-to-add-lora-to-the-m5paper/12084) to hunt for a module that could be used.
2. I still haven't gotten the touch screen to work, as the appropriate bits of code are in the TFTDisplay class. I'd like to propose moving all commonly shared display bits to a separate `BaseDisplay` class that both TFTDisplay and EInkDisplay can inherit from, thereby working around this limitation - but that is way out of scope for this PR and needs proper planning.
3. For now, both LORA and GPS definitions are commented out - I will be enabling these once I have a solid approach to them.
4. The display is barely readable, as there's no scaling for physical size in the current GUI framework. This, as well as handling of a partial window (the M5Paper's display runs to the edges, and since Meshtastic draws to the very edge of the screen, this makes some parts hard to read, even at appropriate scaling), would be needed for _proper_ support of the M5Paper.
5. Currently, this is using my fork of GxEPD2 as ZinggMJ refuses to merge any PRs or include any display they don't own. As such, I will be keeping that tag up to date (though I doubt any further changes will be needed as the display works just fine).
6. There's a weird `invalid pin` error during initial boot:
  ```
INFO  | 17:38:13 1 External Notification Module Disabled
INFO  | 17:38:13 1 Doing EInk init
[  1737][E][esp32-hal-gpio.c:102] __pinMode(): Invalid pin selected
E (1732) gpio: gpio_set_level(226): GPIO output gpio_n_IT8951SetVCOM : 430955
set VCOM = -2.30
GetIT8951SystemInfo : 6001
Panel(W,H) = (960, 540)
  ```
which I can't place. It interrupts, and gets interrupted, by the EINK init sequence, so the logs are a bit garbled, making it hard to debug. Any advice regarding this would be appreciated!

### DIFF
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -77,6 +77,7 @@ jobs:
           - board: station-g1
           - board: m5stack-core
           - board: m5stack-coreink
+          - board: m5stack-m5paper
           - board: nano-g1-explorer
           - board: chatter2
     uses: ./.github/workflows/build_esp32.yml
--- a/boards/m5stack-m5paper.json
+++ b/boards/m5stack-m5paper.json
@@ -1,0 +1,33 @@
+{
+    "build": {
+      "arduino": {
+        "ldscript": "esp32_out.ld",
+        "partitions": "default_16MB.csv"
+      },
+      "core": "esp32",
+      "extra_flags": [
+        "-DARDUINO_M5Stack_M5Paper",
+        "-DBOARD_HAS_PSRAM",
+        "-mfix-esp32-psram-cache-issue",
+        "-mfix-esp32-psram-cache-strategy=memw"
+      ],
+      "f_cpu": "240000000L",
+      "f_flash": "80000000L",
+      "flash_mode": "qio",
+      "mcu": "esp32",
+      "variant": "m5stack_m5paper"
+    },
+    "connectivity": ["wifi", "bluetooth"],
+    "frameworks": ["arduino", "espidf"],
+    "name": " M5Stack M5Paper (16 MB FLASH, 8 MB PSRAM)",
+    "upload": {
+      "flash_size": "16MB",
+      "maximum_ram_size": 327680,
+      "maximum_size": 16777216,
+      "require_upload_port": true,
+      "speed": 460800
+    },
+    "url": "https://docs.m5stack.com/en/core/m5paper",
+    "vendor": "M5Stack"
+  }
+  
--- a/boards/m5stack-m5paper.json
+++ b/boards/m5stack-m5paper.json
@@ -1,33 +1,32 @@
 {
-    "build": {
-      "arduino": {
-        "ldscript": "esp32_out.ld",
-        "partitions": "default_16MB.csv"
-      },
-      "core": "esp32",
-      "extra_flags": [
-        "-DARDUINO_M5Stack_M5Paper",
-        "-DBOARD_HAS_PSRAM",
-        "-mfix-esp32-psram-cache-issue",
-        "-mfix-esp32-psram-cache-strategy=memw"
-      ],
-      "f_cpu": "240000000L",
-      "f_flash": "80000000L",
-      "flash_mode": "qio",
-      "mcu": "esp32",
-      "variant": "m5stack_m5paper"
+  "build": {
+    "arduino": {
+      "ldscript": "esp32_out.ld",
+      "partitions": "default_16MB.csv"
     },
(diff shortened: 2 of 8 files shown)


caveman99:
maybe talk to @markbirss he has don work with m5paper before. His idea was to interface a generic lora radio through the sd card slot.

fonix232:
@caveman99 I was thinking of using two of the three Grove ports to provide a second SPI interface (see the forum thread) - an SX126/7/8X radio could work over 4 pins, and that leaves one of the Grove ports free for a GPS module.

markbirss:
> @caveman99 I was thinking of using two of the three Grove ports to provide a second SPI interface (see the forum thread) - an SX126/7/8X radio could work over 4 pins, and that leaves one of the Grove ports free for a GPS module.

Hi, I tried hard before but just could not get both display and lora 

I tried the tf breakout but those pins must have resistors online or something

image

Sadly the new yet Tobe released esp32-s3 version lacks gpio breakout from the initial pictures I have seen

I have been much more successful with the m5stack coreink

Let me have a look at the changes you made if it was similar what I did before as it was tricky to get gxepd2 to use the m5paper e-eink at the time

markbirss:
Oh 8 also had the same garbled uart issue with m5stack coreink but it went better after esp32 bsp update

markbirss:
> maybe talk to @markbirss he has don work with m5paper before. His idea was to interface a generic lora radio through the sd card slot.

Ok I see you your own added m5paper support to GxEPD2 .

At the time I found a pin combination and modified the resulting only for one of the other ITE8951 dusplays

fonix232:
@markbirss thanks for the info! Could you by any chance link to the new gen M5Paper pictures you've seen?

I doubt M5 will add a proper GPIO header, the M5Paper is meant to be a handheld. But if they got rid of the IT8951, that alone will free up resources that could make it more viable (not to mention the improved power profile).

> Oh i also had the same garbled uart issue with m5stack coreink but it went better after esp32 bsp update

What BSP update?

> Ok I see you added your own added m5paper support to GxEPD2 .

Yep, I did many trials and the default IT8951E IT60 approach worked quite well, although needed some tinkering with the timings and the order of commands (the M5Paper's approach is to bring the MCU online, set VCOM and other settings and only then request the IC info - any other order and the chip simply stops responding). But my fork of GxEPD2 works just fine, tested it in depth myself.

markbirss:
https://twitter.com/M5Stack/status/1620371661433475075?t=G4QzN5tZQgbEOs6O7wBq6g&s=19

This is a new picture that maybe shows some pins on the side, previous pictures did  not show

It the espressif board support version upgrade at the time when I added n5stack coreink

At first it not allowed me to use the python cli over serial

If you are a discord member we can chat more there

markbirss:
> @markbirss thanks for the info! Could you by any chance link to the new gen M5Paper pictures you've seen?
> 
> I doubt M5 will add a proper GPIO header, the M5Paper is meant to be a handheld. But if they got rid of the IT8951, that alone will free up resources that could make it more viable (not to mention the improved power profile).
> 
> > Oh i also had the same garbled uart issue with m5stack coreink but it went better after esp32 bsp update
> 
> What BSP update?
> 
> > Ok I see you added your own added m5paper support to GxEPD2 .
> 
> Yep, I did many trials and the default IT8951E IT60 approach worked quite well, although needed some tinkering with the timings and the order of commands (the M5Paper's approach is to bring the MCU online, set VCOM and other settings and only then request the IC info - any other order and the chip simply stops responding). But my fork of GxEPD2 works just fine, tested it in depth myself.

Well I'm not saying it could not work at all, it possible you approach is different so you could try like I tried the broken out tf/sdcard pins again

PR #3673, opened 2024-04-20
Easier re-use of heartbeat LED with External Notification module

todd-herbert:
Applies when:
* `external_notification.active` enabled
* `external_notification.alert_message` or `external_notification.alert_bell` enabled
* `external_notification.output` is either `LED_PIN` or 0

If above conditions are met, the default LED blink behavior is suppressed, and the LED will instead behave as configured by the External Notification module.

If a user enables *External Notification module* and *Alert when receiving a message*, no further configuration is required to enable the alert behavior. This makes the external notification module more accessible to less experienced users.


todd-herbert:
> Would it make sense to only enable it on the LED pin if the new device.status_led_off is set to true? That would effectively "unlock" the default LED pin for using in the external notifications module.

Makes sense to me!

todd-herbert:
Which is preferred?
* Value of GPIO0 for `external_notification.output` selects `LED_PIN` (manual entry also accepted)
* `external_notification.output` must be manually set to correct LED pin by user

thebentern:
Could be either one, but users may not know the GPIO pin value of the status LED off-hand, so I like the 0 coalescing to LED pin

todd-herbert:
Another commit may or may not be required, depending on the future of `config.device.status_led_off`. Might be an idea to let this one sit for the moment.

todd-herbert:
Note that `heartbeat_disabled` needs a fix similar to [a162fb1](https://github.com/meshtastic/firmware/pull/3673/commits/a162fb1aaf6b9bc324df8984bbf5dd049bcc9b26), regardless of whether this PR is merged.

src/main.cpp
```diff
@@ -693,6 +693,34 @@ void setup()
     // Now that the mesh service is created, create any modules
     setupModules();
 
+#ifdef LED_PIN
+    // Turn LED off after boot, if disabled by config
```
todd-herbert, line 697: Note that this is a separate issue relating to `heartbeat_disabled`, independent of this PR's scope.

PR #3674, opened 2024-04-20
Add ability to turn off heartbeat LED blinking

### LINKED ISSUE #3635: [Feature Request]: Option to turn off LED inducators on ESP32 Heltec V3
### Platform

ESP32

### Description

Please can we have an option in the Android app to turn off the blinking indicator LEDS when connected to USB power. It lights up the window and room at night in which the Heltec V3 is in.

ndoo:
Fixes #3635 and depends on [protobufs PR #485](https://github.com/meshtastic/protobufs/pull/485).

This only affects the GPIO-controllable LED and may be desirable behavior for Meshtastic users who sleep in the vicinity of their node and are annoyed by the blinking LED, especially for boards that indicate charging when USB power is connected, resulting in a 1 second LED illumination cycle.

Most boards however, have a non-GPIO controlled power LED (i.e. hardwired to VBUS/VCC/etc.) that can only be turned off by desoldering or otherwise removing that LED. This can't be addressed by firmware changes.

Implemented as inhibiting rather than `status_led` because having `status_led = true` by default will cause status LED to get disabled for users upgrading from previous firmware due to `installDefaultConfig` not being run.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -179,7 +179,8 @@ const char *getDeviceName()
 
 static int32_t ledBlinker()
 {
-    // Still set up the blinking (heartbeat) interval but skip code path below, so LED will blink if config.device.led_heartbeat_disabled is changed
+    // Still set up the blinking (heartbeat) interval but skip code path below, so LED will blink if
+    // config.device.led_heartbeat_disabled is changed
     if (config.device.led_heartbeat_disabled)
         return 1000;
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -179,6 +179,10 @@ const char *getDeviceName()
 
 static int32_t ledBlinker()
 {
+    // Still set up the blinker interval but skip code path below, so LED will blink if config.device.status_led_off is changed
+    if (config.device.status_led_off)
+        return 1000;
+
     static bool ledOn;
     ledOn ^= 1;
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -179,8 +179,8 @@ const char *getDeviceName()
 
 static int32_t ledBlinker()
 {
-    // Still set up the blinker interval but skip code path below, so LED will blink if config.device.status_led_off is changed
-    if (config.device.status_led_off)
+    // Still set up the blinking (heartbeat) interval but skip code path below, so LED will blink if config.device.led_heartbeat_disabled is changed
+    if (config.device.led_heartbeat_disabled)
         return 1000;
 
     static bool ledOn;


thebentern (approved):

thebentern:
Looks good. I hate how bright that stupid white LED is on the heltec!

garthvh (changes requested):
This is not going to provide a dark mode, can we just reduce the brightness of the heltec and not affect every device?

thebentern:
> This is not going to provide a dark mode, can we just reduce the brightness of the heltec and not affect every device?

This just allows the user to disable the blinking white LED. I think it potentially also has some utility if we can repurpose the status LED for use with ext. notifications to control the behavior instead, in #3673

ndoo:
> This is not going to provide a dark mode, can we just reduce the brightness of the heltec and not affect every device?

The default behavior of this change does not affect any device unless the user specifically flips the configuration toggle.

Reducing brightness adds a lot of complexity/refactoring to the blink routine as we have to configure and attach the ESP32 PWM LEDC to the pin, as well as adding the necessary protobuf config to support configuring the brightness to make this feature available to all other boards. Hardcoding the brightness is also possible, but then we become responsible to choose the brightness that most users would find comfortable.

Hardcoding this for just the Heltec will still require changing the main.cpp ledBlinker routine, it seems more straightforward to just implement a flag to turn the LED off entirely for users who are affected by the blinking behavior.

ndoo:
> Looks good but needs a trunk fmt. I hate how bright that stupid white LED is on the heltec!

trunk fmt fixed with latest force push

ndoo:
Thanks for the major refactoring work and glad to see its inclusion.

punkpar:
Thankyou all for following up my initial feature request. I'm still new to GitHub, will this feature be in the next alpha release of meshtastic?

todd-herbert:
> Thankyou all for following up my initial feature request. I'm still new to GitHub, will this feature be in the next alpha release of meshtastic?

The pull request https://github.com/meshtastic/firmware/pull/3674 was included in 2.3.7, but I believe there was a bug in that code, so the feature might not be usable until 2.3.8.

I don't believe either of the phone Apps allow you to change this setting just now. You *can* set it with the python CLI, but I believe it needs a fix such as https://github.com/meshtastic/firmware/pull/3673/commits/a162fb1aaf6b9bc324df8984bbf5dd049bcc9b26 to work correctly.

punkpar:
Thank you todd-herbert, does this need requesting in the Android and iPhone firmware section, or is this something that will naturally be implemented?

todd-herbert:
To be honest, I don't know exactly what will happen with the different apps, but I'm sure the developers behind them are aware of the changes; no need to worry about opening requests.

PR #3678, opened 2024-04-20
Support radar sensor RCWL-9620 on i2c

caveman99:
Needs PR from Protos repo to compile.


caveman99:
Needs PR from Protos repo to compile.

thebentern (approved):

caveman99:
ARM runner still FUBAR

thebentern:
Would be fair to say that the ARM runner needs a HAND?

caveman99:
it needs to be put out of its misery ... :-)

ndoo:
Hi @caveman99, I think /src/serialization/MeshPacketSerializer.cpp and possibly the nRF version of this file would need the meshtastic_PortNum_TELEMETRY_APP section updated to actually put the distance out into transmitted protobufs. 

Let me know if a separate issue should be raised.

fifieldt:
@ndoo thanks for your detailed analysis. Please do raise a separate issue so we don't lose this important information :)

PR #3693, opened 2024-04-21
Updates for esp32s2 builds

titan098:
Minor update to allow builds for `esp32s2` modules to succeed. The esp32s2 doesn't have Bluetooth so exclude various Bluetooth related components to allow for a successful build if the target platform is an `esp32s2`.

### DIFF
--- a/arch/esp32/esp32s2.ini
+++ b/arch/esp32/esp32s2.ini
@@ -2,14 +2,17 @@
 extends = esp32_base
 
 build_src_filter = 
-  ${esp32_base.build_src_filter} -<nimble/> -<mesh/raspihttp>
+ ${esp32_base.build_src_filter} - <libpax/> -<nimble/> -<mesh/raspihttp>
 
 monitor_speed = 115200
 
 build_flags =
   ${esp32_base.build_flags} 
   -DHAS_BLUETOOTH=0
+  -DMESHTASTIC_EXCLUDE_PAXCOUNTER
+  -DMESHTASTIC_EXCLUDE_BLUETOOTH
   
 lib_ignore = 
   ${esp32_base.lib_ignore} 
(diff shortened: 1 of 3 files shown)


src/modules/AdminModule.cpp
```diff
@@ -721,7 +721,7 @@ void AdminModule::handleGetDeviceConnectionStatus(const meshtastic_MeshPacket &r
 #if HAS_BLUETOOTH
     conn.has_bluetooth = true;
     conn.bluetooth.pin = config.bluetooth.fixed_pin;
-#ifdef ARCH_ESP32
+#if defined(ARCH_ESP32) && !defined(CONFIG_IDF_TARGET_ESP32S2)
```
caveman99, line 724: Is this really neccessary? ESP32-S2 should not have HAS_BLUETOOTH defined.

caveman99 (changes requested):
why don't you just set -DMESHTASTIC_EXCLUDE_PAXCOUNTER and -DMESHTASTIC_EXCLUDE_BLUETOOTH in the arch/esp32s2.ini for S2?

src/modules/AdminModule.cpp
titan098: Updated as suggested.

Following updates were needed:
- change to `PaxCounterModule.cpp` is needed as the build will still pull in the cpp file regardless if the define is set
- change to `sleep.cpp` to exclude pulling in nimble if `MESHTASTIC_EXCLUDE_BLUETOOTH` is defined.

caveman99 (approved):

src/modules/AdminModule.cpp
caveman99: Excellent - this way we can dodge the dreaded CONFIG_IDF macros ... :-)

PR #3668, opened 2024-04-19
Added modifier key combination to allow keyboard shortcuts on t-deck

nbaddorf:
The keyboard on the t-deck does not have a 'tab' button or a function button for triggering keyboard shortcuts. I added a 'modifier' key combo that is triggered by pressing shift and the mic (0) button on t-decks keyboard. Then press the letter for the shortcut you want. Currently implemented is: q (quit or cancel), t (tab), and m (mute/unmute).

PR #3695, opened 2024-04-21
Fixed node and channel selection for t-deck

nbaddorf:
To change nodes and channels, press the tab button (use a modifier key) to change nodes. Press it again to toggle to the channel selection. 

(for a tab modifier key for the t-deck checkout my other pull request: https://github.com/meshtastic/firmware/pull/3668)


caveman99 (changes requested):
not 100% sure about that one. I remember them being there for a reason. Maybe talk to @mverch67 about that PR

mverch67:
> not 100% sure about that one. I remember them being there for a reason. Maybe talk to @mverch67 about that PR

Yes, it's ok. These lines were in place to automatically enter the node selection mode when swiping left/right due to a missing tab key. Now that a tab key is simulated on T-Deck these lines can be removed.

nbaddorf:
Great thanks!

thebentern (approved):

PR #3696, opened 2024-04-21
Use correct format specifier and fixed typo.

Mictronics:
Use correct format specifier in timestamp log. Both position time and timestamp are uint32_t.


thebentern:
Good catch!

PR #3701, opened 2024-04-22
Add power limit for TW region

fifieldt:
The TW region had now power limit set, so defaulted to 16dBm.

The relevant regulation is section 5.8.1 of the _Low-power Radio-frequency Devices Technical Regulations_, which notes the limits of 0.5W (27dBM) indoor or coastal, and 1.0W (30dBM) outdoor.

This patch updates the power limit to 27dbM, using the the lower limit specified in the regulations to be conservative. 

**Regulation references**
[latest English version](https://www.ncc.gov.tw/english/files/23070/102_5190_230703_1_doc_C.PDF)
[latest Chinese version, February 2024](https://gazette.nat.gov.tw/egFront/e_detail.do?metaid=147283)

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -76,9 +76,12 @@ const RegionInfo regions[] = {
     RDEF(KR, 920.0f, 923.0f, 100, 0, 0, true, false, false),
 
     /*
-        ???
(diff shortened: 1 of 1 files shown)


SCWhite:
This change is valid, can confirm the regulation is true, and it's the latest version.

PR #3702, opened 2024-04-22
Add SX1268 modules support for linux-native

SannySannyman:
The code is tested with E22 400M33S module, but should work for other SX1268-based modules, as it reuses the existing radio interface code.


mverch67:
bin/config-dist.yaml
 -:-  fmt  Incorrect formatting, autoformat by running '**trunk fmt**'  prettier

src/main.cpp
```diff
@@ -768,6 +768,21 @@ void setup()
                 LOG_INFO("SX1280 Radio init succeeded, using SX1280 radio\n");
             }
         }
+    } else if (settingsMap[use_sx1268]) {
+        if (!rIf) {
+            LOG_DEBUG("Attempting to activate sx1268 radio on SPI port %s\n", settingsStrings[spidev].c_str());
+            LockingArduinoHal *RadioLibHAL = new LockingArduinoHal(SPI, spiSettings);
```
jp-bennett, line 774: I believe this now needs to use *LoraSPI to avoid potential conflicts with the display or touchscreen.

jp-bennett (changes requested):

src/main.cpp
SannySannyman: Fixed.

SannySannyman:
> bin/config-dist.yaml -:- fmt Incorrect formatting, autoformat by running '**trunk fmt**' prettier

Done.

jp-bennett (approved):
Looks good to me

PR #3728, opened 2024-04-26
Reverse directions for swiping left and right on touchscreen

nbaddorf:
Reverse the directions for swiping left and right on touch screen. This fix makes it correct on T-deck.


mverch67:
Actually the swipe direction was intentionally implemented this way: if you swipe a screen from right to left then the screen on the right side reveals. Imagine that the screens contents would be shown while swiping, then you get it.

nbaddorf:
I guess that makes sense but it still feels wrong that swiping right to left on the trackball goes one way and the same swipe on the screen goes the other. 

One other issue with the current direction setting for the screen is when typing out a message if you swipe right to left it moves the cursor to the right.
If the directions are kept the way they are currently, we could implement my change so the trackball and screen act the same, and then swap the showNextFrame and showPrevFrame at the bottom of Screen.cpp so that the frames move the way you like. This way the cursor when typing moves correctly.

PR #3735, opened 2024-04-27
Allow setting hopLimit for MQTT json sendtext and sendposition

mskvortsov:
- Allow setting hopLimit for mqtt sendtext and sendposition
- Fix channel name extraction bug


mskvortsov:
@GUVWAF Take a look, please.

The channel name bug has been spotted on a native 64-bit platform.

And my use case for the hopLimit addition was for an automatic monitoring of how well a target neighbor node hears a source node. (The source transmits a test packet downlinked via sendtext with hopLimit set to 1, waits on MQTT for a copy uplinked by the target, and logs the resulting signal levels.) I think hopLimit might be potentially useful in other cases too, so why not to allow setting it.

GUVWAF (approved):
Looks good to me. 
Checking the channel name is indeed even simpler like this.

Issue #3749, opened 2024-04-30
[Bug]: CardKB not starting on ESP32 when first powered


HarukiToreda:
### Category

Other

### Hardware

Heltec V3

### Firmware Version

v2.3.7.30fbcab

### Description

this has been an issue for a while but i figure I open this now. When you first power an ESP32 device with a CardKB, it always says "Canned Message disabled", to fix it I always have to press the reset button and once I do that it is recognized and fixed. This happens when using the power switch to boot

I believe this is due to the CardKB having it's own Atmega chip that takes a few milisecs longer than the initial i2c check on node boot.

HarukiToreda:
Issue fixed on recent PL
https://github.com/meshtastic/firmware/pull/3933

PR #3750, opened 2024-04-30
Fix for boot error for CardKB on ESP32 Heltec V3

HarukiToreda:
Fix for "[Bug]: CardKB not starting on ESP32 when first powered #3749"

I introduced a slight delay to allow the Atmega chip on the CardKB to boot properly and be ready for when the I2C scan happens on the ESP32. With this, the error of "Canned Message disabled" is fixed.


HarukiToreda:
If anyone could please help with the trunk error, this has been tested on the Heltec V3, WSL and the tracker, I just have no not been able to fix this Trunk error

.trunk/configs/.clang-format
```diff
@@ -3,4 +3,4 @@ IndentWidth: 4
 ColumnLimit: 130
 PointerAlignment: Right
 BreakBeforeBraces: Linux
-AllowShortFunctionsOnASingleLine: Inline
+AllowShortFunctionsOnASingleLine: Inline
```
caveman99, line 6: Please revert the changes to the .trunk and .vscode files - these changes are unrelated to this PR

src/main.cpp
```diff
@@ -391,6 +391,7 @@ void setup()
 
     // We need to scan here to decide if we have a screen for nodeDB.init() and because power has been applied to
     // accessories
+    delay(500);  // Introduce a delay before scanning for I2C devices
```
caveman99, line 394: Alternate suggestion: repeat the bus scan/test for keyboards in the cannedMessages init. This 500msec will go on top the 5s delay we have waiting for the serial port.

caveman99 (changes requested):

src/main.cpp
HarukiToreda: I added a way to rescan right after and it works. Removed the delay

HarukiToreda: nvm, it gave me lots of errors. reverting to delay

caveman99:
My i add a suggestion to your repo? so you can compile&test. That will rescan only the known keyboard addresses and not all of the bus.

HarukiToreda:
I have not been able to figure out a different way, not sure if you'd prefer this closed for someone else to try someday.

PR #3758, opened 2024-05-01
Publish fixed position updates and consider changes in only altitude as an updated point

ndoo:
Enabled GPS thread when fixed position is updated, to let the GPS thread run once and publish the new fixed position.

Also consider changes in only altitude as an updated point, so that running Python `meshtastic` with `--setlat --setlong --setalt` but only changes to altitude will still publish to the mesh.


jp-bennett (approved):
Generally looks OK to me.

Issue #3736, opened 2024-04-27
[Bug]: T-Beam Supreme (S3 core) loses USB connectivity with latest FW 2.3.6.7a3570a Beta


GregEigsti:
### Category

Serial

### Hardware

T-Beam S3

### Firmware Version

2.3.6.7a3570a Beta

### Description

I have a python app that connects to a T-Beam Supreme via serial/USB and noticed that I lose the ability to connect via serial/USB with the latest FW `2.3.6.7a3570a Beta` running. Rolling back to FW `2.3.4.ea61808 Beta` fixes the issue thankfully. The python CLI app `meshtastic` is also unable to connect via serial/USB when this T-Beam Supreme is running FW `2.3.6.7a3570a Beta`. I have rolled the FW from previous to latest several times and this is 100% reproducible; for me anyway ;)

The T-Beam is connected to a Ras Pi 5 via serial/USB where a python script, which uses the Meshtastic python API, interacts with it. The T-Beam is also configured to use WiFi and uses that network connection to forward MQTT. I do notice the same issue when connecting this T-Beam to a MacOS work station.

### Relevant log output

```Shell
greg@meshaprs:~/src/python/meshtastic/meshaprs$ meshtastic --support

If having issues with meshtastic cli or python library
or wish to make feature requests, visit:
https://github.com/meshtastic/python/issues
When adding an issue, be sure to include the following info:
 System: Linux
   Platform: Linux-6.5.0-1015-raspi-aarch64-with-glibc2.38
   Release: 6.5.0-1015-raspi
   Machine: aarch64
   Encoding (stdin): utf-8
   Encoding (stdout): utf-8
 meshtastic: v2.3.7
 Executable: /home/greg/.local/bin/meshtastic
 Python: 3.11.6 CPython GCC 13.2.0

Please add the output from the command: meshtastic --info

Will add "meshtastic --info" upon request and privately.
```

GregEigsti:
I wonder if the T-Beam, with FW `2.3.6.7a3570a Beta` is stuck in `DFU` (or whatever) mode? I did notice during one of the rollbacks to 2.3.4 that I did not have to do the funky dual button press to make it accept new FW from the web flasher - it was ready to accept new FW immediately after booting with no button presses. Maybe this will help...

GUVWAF:
Seems that this #3597 might be the cause.

GUVWAF:
I don't have a T-Beam Supreme, but maybe @oseiler2 can have a look.

[...]

Issue #3650, opened 2024-04-17
[Bug]: T-Beam S3 blocks during boot


oseiler2:
### Category

Hardware Compatibility, Serial

### Hardware

T-Beam S3

### Firmware Version

v2.3.5.2f9b68e

### Description

T-Beam S3 devices can hang at boot until a serial connection is established on the USB CDC interface.
This was likely introduced by https://github.com/meshtastic/firmware/pull/3597
Also see https://github.com/meshtastic/firmware/pull/3597#issuecomment-2058845144

It hangs after the below logs for me. When I connect a serial on the USB CDC port it continues booting.

FWIW, I don't think it is necessarily related to this https://github.com/espressif/arduino-esp32/issues/7554
I have checked all `flush()` calls on the relevant HWCDC and couldn't isolate the issue by doing so.

### Relevant log output

```Shell
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x1 (POWERON),boot:0x8 (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce3808,len:0x3ac
load:0x403c9700,len:0x9b4
load:0x403cc700,len:0x28d8
entry 0x403c98bc
E (414) esp_core_dump_flash: No core dump partition found!
E (414) esp_core_dump_flash: No core dump partition found!
[  1304][I][esp32-hal-psram.c:96] psramInit(): PSRAM enabled
```

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/t-beam-wont-power-on-without-usb/12043/2

PR #3764, opened 2024-05-02
Fix T-Beam S3 USB CLI #3736

oseiler2:
Fix for #3736 

It seems that we do need ARDUINO_USB_CDC_ON_BOOT=1 after all in order for the CLI to work with the ESP32-S3. Otherwise I cannot get the S3 to read from USB Serial.

Issue https://github.com/meshtastic/firmware/issues/3650 that can cause the S3 to block on boot when no Serial is connected (and is triggered by ARDUINO_USB_CDC_ON_BOOT=1 in combination with ARDUINO_USB_MODE=1) has recently been fixed in [arduino-esp32 @ 2.0.15](https://github.com/espressif/arduino-esp32/releases/tag/2.0.15) . I've confirmed that using platform_packages = framework-arduinoespressif32 @ https://github.com/espressif/arduino-esp32.git#2.0.15

### DIFF
--- a/boards/tbeam-s3-core.json
+++ b/boards/tbeam-s3-core.json
@@ -7,6 +7,7 @@
     "extra_flags": [
       "-DBOARD_HAS_PSRAM",
       "-DLILYGO_TBEAM_S3_CORE",
+      "-DARDUINO_USB_CDC_ON_BOOT=1",
       "-DARDUINO_USB_MODE=1",
       "-DARDUINO_RUNNING_CORE=1",
       "-DARDUINO_EVENT_RUNNING_CORE=1"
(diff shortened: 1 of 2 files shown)


oseiler2:
I hope this finally fixes any issues around USB Serial while allowing to flash new FW without having to put the device in download mode first. Apologies for the issues this has been causing in the meantime.
I've tried to test a range of scenarios using my T-Beam S3, but keen to rule out any further issues if someone can think of anything else I can test.

caveman99 (approved):

mverch67:
This probably introduces a new issue: the device cannot work without serial connection...

oseiler2:
It does work without serial, I've tested that. The bug that causes serial to block has been fixed in the espressif arduino framework

oseiler2:
@mverch67 see https://github.com/espressif/arduino-esp32/issues/7554#issuecomment-2067585042

PR #3776, opened 2024-05-03
Fix #MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR

oseiler2:
Ensure that `#define MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR` actually excludes the sources of all environmental sensors, so that the lib dependencies do not need to be pulled in.

Also moved [QMC5883LCompass](https://github.com/meshtastic/firmware/blob/master/platformio.ini#L99) under [[environmental_base]](https://github.com/meshtastic/firmware/blob/master/platformio.ini#L115)

Fixed compile error when MESHTASTIC_EXCLUDE_GPS defined (CannedMessageModule)

see #3734


oseiler2:
I'm new to trunk - it's not raising any issues in my VS Code environment and I can't see anything obvious why some files are failing. Any pointers as to what it's complaining about?

thebentern (approved):

thebentern:
You can manually run `trunk fmt`. Looks like it's just complaining about whitespace / styling issues

caveman99:
> You can manually run `trunk fmt`. Looks like it's just complaining about whitespace / styling issues

trunk is now supported on windows. so the vscode plugin actually works. just it's missing a few formatters compared to the linux version. notably clang-format. so people on windows get the thumbs up from the plugin and CI still complains.

thebentern:
Ah yes. I've been running trunk manually on my windows box inside WSL.

caveman99:
> Ah yes. I've been running trunk manually on my windows box inside WSL.

me too but i see this pop up a lot lately. How feasible is a 'trunk fmt, then push to pr' ci action to run manually?

caveman99 (approved):

oseiler2:
Weird, I thought trunk was happy (screenshot is on the file version that the runner complained about), and I have run `trunk fmt` in terminal without it showing any errors. Appreciate any pointers to get this resolved.
image

PR #3778, opened 2024-05-03
Enable compiling with gccnoneeabi 12.3.1 for nRF52 targets, additional small fixes

tuxphone:
When compiling with gccnoneeabi 12.3.1 (on windows) one runs into the problem that there are some definitions using the type "uint", which is not a standard type. Resulting in a failed compiler run.

Also there are two instances of a missing time.h (error: time_t not declared).

The meshtastic type of nodeNum is "NodeNum", so it should be used instead of uint.

channel_num is defined as uint, but should rather be uint32_t (type of hash() and num_channels)

Typo "Adverising" fixed in nrf/bluetooth

### DIFF
--- a/arch/nrf52/nrf52.ini
+++ b/arch/nrf52/nrf52.ini
@@ -3,7 +3,7 @@
 platform = platformio/nordicnrf52@^10.4.0
 extends = arduino_base
 
-build_type = debug ; I'm debugging with ICE a lot now
+build_type = release
 build_flags = 
   ${arduino_base.build_flags}
   -DSERIAL_BUFFER_SIZE=1024
--- a/src/SerialConsole.cpp
+++ b/src/SerialConsole.cpp
@@ -2,6 +2,7 @@
 #include "NodeDB.h"
 #include "PowerFSM.h"
 #include "configuration.h"
+#include "time.h"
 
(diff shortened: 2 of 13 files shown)


tuxphone:
I'm not sure what went wrong with the tbeam 0.7 check:
Could build it on my machine with no problems.

...
Linking .pio\build\tbeam0_7\firmware.elf
Retrieving maximum program size .pio\build\tbeam0_7\firmware.elf
Checking size .pio\build\tbeam0_7\firmware.elf
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [=         ]   7.3% (used 96040 bytes from 1310720 bytes)
Flash: [========= ]  85.3% (used 2069017 bytes from 2424832 bytes)
Building .pio\build\tbeam0_7\firmware.bin
esptool.py v4.5.1
Creating esp32 image...
Merged 25 ELF sections
Successfully created esp32 image.

caveman99 (approved):

caveman99:
I have the same changes in an uncommitted branch where i wanted to try something out. absolutely spot on ... thanks

caveman99:
I remember now where these changes were needed. ESP32-C6 support requires a newer ESP32 toolchain, and this one complained about the same defines, among other things.

Issue #2590, opened 2023-07-07
[Feature Request]: Updated Device Screen UI modes for all devices


tropho23:
### Platform

NRF52, ESP32

### Description

# **Updated Device Screen UI modes**

## *Overview of this proposal:*
The current Meshtastic device screen user interface (UI) needs to be updated to display additional node and sensor information, enable enhanced user interaction without relying on BLE-paired mobile devices, and support advanced features introduced by new standalone devices integrating keyboards, rotary encoders, and matrix keypads.  Device capabilities and new user interactions have grown beyond the capabilities of the current device screen UI.

Ideally, the current screen UI library should be able to support these proposed navigation flows and nested menus/actions for the new 128x64 screen UI modes; if not a new screen UI library must be considered.  If currently supported e-ink displays can also take advantage of this proposal, that compatibility should also be accommodated if feasible.

For the Advanced Screen UI mode, a new screen UI/display library will need to be integrated to not only take advantage of TFT LCD and controllers, but also offer the graphical performance to support higher screen resolutions, interactive map displays, and possibly even touch interface, if feasible.  The goal of the Advanced Screen UI mode is to minimize, or even eliminate the need to use BLE-paired mobile devices for any non-configuration use.

Selecting these modes will require a CLI/app/web configuration setting to choose Device Screen UI mode.

## **List of proposed Device Screen UI modes:**
- [ ] 1. **Boot/splash screen update**:   (128x64+) See note below
- [ ] 2. **Unset region screen update**:   (128x64+) See note below
- [ ] 3. **Standard Screen UI mode**:  (128x64) Improved single-screen information display for devices without user buttons
- [ ] 4. **Carousel Info mode**:  (128x64) Continually cycle between 'default info' and 'additional info' screens; useful for testing/setup
- [ ] 5. **Enhanced Screen UI mode**:  (128x64) Standard mode plus new menu navigation flows and functionality; includes additional menu navigation for sending canned messages
- [ ] 6. **Standalone Screen UI mode**:  (128x64) For standalone devices with CardKB, rotary encoder, or matrix keypad input methods; supports existing features (DMs and canned messages) and planned features (group messages, access previously received message store)
- [ ] 7. **Advanced Screen UI mode**: (320x240+) Eventually integrate TFT screens; support all features of UI modes 3-6 plus maps, waypoints, etc. and perhaps even touch-enabled input; likely requires different screen/UI library
[...]

Issue #1985, opened 2022-11-24
[Feature request]: Access previously received messages on-device


tropho23:
Request capability for users to access (and delete) previously received messages, stored on-device using the Store and Forward (S&F) module, and displayed on the device screen.  This could be used for any device, not just standalone messengers with keyboard or rotary encoder but that is the intended use case.  The 'one button' library supports button longpress and other features to support this.

I propose UI navigation solutions below to enable users to invoke a UI level to view and scroll through received messages; one for rotary encoder and one for CardKB users.  Ideally a single solution would be implemented for simplicity, assuming there is a way to do so without conflicting with current button assignments.  However, button assignments could be modified to support this.

This feature request is associated with feature request https://github.com/meshtastic/firmware/issues/1983 (Store and Forward module), which would presumably be required for this feature request to work.

Using a Rotary encoder:
1. User longpresses encoder to access 'Received Messages' menu
2. Screen displays 'Received Messages' on top row of screen
3. Rows below top row display received messages in descending order (newest first)
4. Newest received message received at 05:00 is highlighted by default (inverse text)
 4a. CW encoder rotation scrolls to the next oldest message received at 04:58
 4b. CCW encoder rotation scrolls back to the newer received message
5. Single press of encoder presents two choices to user: View and Delete (view highlighted by default)
 5a. CW/CCW encoder rotation highlights View or Delete, single press of encoder selects choice
 5b. If user selects Delete, present user with confirmation "Delete this message?", single press of encoder confirms and the list of received messages is again displayed, minus the deleted message
 5c. If user selects View, display selected message onscreen
 5d. A second single press of encoder dismisses the message, going back to the list of received messages
6. Longpress encoder rotation dismisses the received messages list, returns to main screen
 6a. Longpressing encoder should also exit the received messages list and/or menus at any time

Using a CardKB keyboard (and future supported keyboards):
1. User presses a custom key combination (e.g. fn key, then M key for example)
2. Screen displays 'Received Messages' on top row of screen
3. Rows below top row display received messages in descending order (newest first)
4. Newest received message received at 05:00 is highlighted by default (inverse text)
 4a. Down arrow key scrolls to the next oldest message received at 04:58
 4b. Up arrow key scrolls back to the newer received message
5. Pressing the enter key presents two choices to user: View and Delete (view highlighted by default)
 5a. Up/down arrow keys highlight View or Delete, enter key selects choice
[...]

PR #3784, opened 2024-05-04
Add persistent message storage and message log frame

ndnestor:
## Added
- Persistent message storage
- Message RAM cache
- Message log UI frame
- Message protobuf

## Changed
- Trunk version to latest
- Trunk linters' versions to latest
- Touchscreen tap and double tap input events

## Removed
- Text message frame that only showed last received message

## Video Demonstration
[Here is a video demonstration of changes made](https://drive.google.com/file/d/1BUeJQAR8TjdxFDLwkAr7xPpC10Q1lMMu/view). It is slightly outdated.

## Recommended Future Changes
- Replace current ACK UI frame with a symbol that is inline with the ACK'd message
- Add "default auto-focus state" toggle to settings
- Add "delete all messages" button to settings
- Add setting to change how often messages are saved
- Use `ostream` for more efficient string building in `drawMessageLogFrame`
- Add external storage support (e.g. SD card)
- Only run through message log frame drawing logic when what is drawn needs to be changed
- Warn user when running out of storage
- Delete only the oldest message when running out of storage


src/graphics/Screen.cpp
```diff
@@ -1609,7 +1745,8 @@ void Screen::handleOnPress()
     // If screen was off, just wake it, otherwise advance to next frame
     // If we are in a transition, the press must have bounced, drop it.
     if (ui->getUiState()->frameState == FIXED) {
-        ui->nextFrame();
+        // TODO: Figure out if the rest of this is needed for waking up
+        // ui->nextFrame();
```
todd-herbert, line 1749: Does this need to be disabled? It prevents devices with only a user-button from cycling through the frames.

src/graphics/Screen.cpp
```diff
@@ -402,43 +409,170 @@ static bool shouldDrawMessage(const meshtastic_MeshPacket *packet)
...
+}
+
+// Format message for displaying as text
+std::string createMessageString(OLEDDisplay *display, const meshtastic_Message msg)
+{
+    uint32_t seconds = sinceReceived(&msg);
     uint32_t minutes = seconds / 60;
     uint32_t hours = minutes / 60;
     uint32_t days = hours / 24;
+    std::string timestamp = screen->drawTimeDelta(days, hours, minutes, seconds);
```
todd-herbert, line 470: Rather than "time since received", could this display the local-time that the message was received (where possible)? Otherwise this information quickly becomes invalid on E-Ink displays, which won't refresh this info until a new message arrives.

todd-herbert (changes requested):
Looking good! Do you have any plans to implement some of the features (cycle through channel?) for devices which only have a single user-button?

GUVWAF:
Very nice. 

On which devices/screens has this been tested next to the T-Deck?

src/modules/TextMessageModule.cpp
```diff
@@ -14,9 +17,8 @@ ProcessMessage TextMessageModule::handleReceived(const meshtastic_MeshPacket &mp
 #endif
 
     // We only store/display messages destined for us.
-    // Keep a copy of the most recent text message.
-    devicestate.rx_text_message = mp;
-    devicestate.has_rx_text_message = true;
+
+    nodeDB->saveMessageToDisk(mp);
```
GUVWAF, line 21: I don’t think it’s a good idea to save to flash every time it receives a message, and especially not if the CannedMessageModule is not enabled (or it doesn’t even have a screen). Saving to flash will wear it out, and it can take a long time (especially on nRF52 as @prokrypt recently found), so it even affects packet reception and routing.

Also it uses precious RAM for devices without screen, which should be avoided in my opinion.

thebentern: Throttling this like we're doing for nodedb updates when we have an updated NodeInfo perhaps would be helpful

GUVWAF: I would add it only to the RAM cache when the TextMessageModule processes it and save it to flash at a different place. E.g. in `NodeDB::loadMessage()` (only when there's a new message) where you also potentially remove messages from the cache and when rebooting/shutting down.

src/modules/CannedMessageModule.cpp
```diff
@@ -1,3 +1,6 @@
+#include <algorithm>
```
GUVWAF, line 1: These includes are unnecessary (just like those in `TextMessageModule`).

src/mesh/NodeDB.cpp
```diff
@@ -718,6 +726,258 @@ void NodeDB::saveDeviceStateToDisk()
...
+        if (mp.to == nodeDB->getMeshNodeByIndex(i)->num)
+            return true;
+
+    return false;
+}
+
+void NodeDB::saveMessageToDisk(const meshtastic_MeshPacket &mp)
+{
+    auto *msg = new meshtastic_Message();
+    auto senderNode = nodeDB->getMeshNode(mp.from);
```
GUVWAF, line 800: Better to use `getMeshNode(getFrom(mp))` here, same for line 805. Currently if you're sending a packet from the phone to the device, `from` will still be 0 here, so it shows as "UNK".

Also currently packets sent from the phone will not be shown for DMs because `handleReceived()` is not called for DMs:
https://github.com/meshtastic/firmware/blob/70712d859cdc242b1873fa9b9622cf1c1711fb06/src/mesh/Router.cpp#L175-L179

.trunk/trunk.yaml
```diff
@@ -1,32 +1,32 @@
 version: 0.1
 cli:
-  version: 1.20.1
+  version: 1.22.0
```
caveman99, line 3: Please remove the changes to .trunk and protobufs/generated files. Protobufs need to go into its own repository.

src/mesh/NodeDB.cpp
```diff
@@ -718,6 +726,258 @@ void NodeDB::saveDeviceStateToDisk()
               &devicestate);
 }
 
+void NodeDB::initSavedMessages()
+{
+    FSCom.mkdir("/msgs");
+    for (int category = 0; category < CATEGORY_COUNT; category++)
```
caveman99, line 732: this directory is not erased when doing a factory erase. Please add the erase.

caveman99 (changes requested):

HarukiToreda:
Tested on Heltec V3, wireless Paper, Wireless Tracker. Although the screen on the Tracker is the same size as the V3 i can view less messages per line, this is not an issue since we have the ability to scroll with the arrow keys. However the User button does not work anymore, no matter how many times it is pressed it remains on the same Message screen on all 3 devices. Arrow keys on the CardKB are the only way to go to the next screens.

src/graphics/Screen.cpp
ndnestor: A solution could be enabling user button cycling for devices with only one button. It was disabled because the user button press now has a new function. What do you think?

ndnestor:
> Very nice.
> 
> On which devices/screens has this been tested next to the T-Deck?


@GUVWAF 
I've only tested it on the T-Deck so far since that's all I own.

src/modules/TextMessageModule.cpp
ndnestor: This new feature does not seem practical for use on nRF52 devices if it takes a long time to save messages. On top of that there is no way of checking how much free RAM is left as far as I can tell. I will make it fall back on the old way of displaying messages for nRF52 devices, devices with the router role, devices without a screen, and devices where the canned message module is disabled.

As for the flash wearing out, where can I find information on the amount of reads/writes that can be done? It would seem to me that the same amount of reads and writes are occurring after these changes because every message received was being saved to disk before I changed anything. I don't think it is a good idea to save messages in batches because power loss would cause messages to be lost.

src/graphics/Screen.cpp
ndnestor: Displaying the time and date would take up a lot of screen space which is why I went with this approach. For E-Ink displays, you can manually refresh by doing any action. For now, that'll have to be the workaround.

src/graphics/Screen.cpp
todd-herbert: From a quick look, it seems like next frame has been disabled in `handleOnPress()` for all devices, and is now intercepted in `handleInputEvent()`, as `meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar_SELECT`, right?
Only a select few devices will have the correct inputs to use this new behavior. Both `handleInputEvent()` and `handleOnPress()` will need to be tied to some sort of config option. 

A macro maybe wouldn't a great choice here, because a lot of DIY devices (e.g. Heltec V3) may have a single user button, or may have a more complex keyboard, depending on how the device is built. It wouldn't be fair to inconvenience all these users by forcing them to custom-build firmware from now on.

A lot of this functionality might need to tied to a new user-setting? Possibly there is an existing user-setting which can be used in this particular case (determining how to handle input)?

If you haven't considered it already, I'd also think about how this code interacts with the existing canned-messages module. Also might be an idea to think about how users with canned-message inputs different than the T-Deck would interact with this new feature.

src/graphics/Screen.cpp
todd-herbert: If I'm being honest, I'm particularly satisfied with this idea as a workaround. I think there's definitely a workable solution here somewhere. Seeing how this behavior is really only important to E-Ink devices, it's absolutely possible to implement two different pieces of code here, wrapped with the `USE_EINK` macro.

If it helps, I can have a play around and see if I can come up with code for this?

Having spend a tiny bit more time testing the PR now, I'm not actually sure this threaded-messaging idea is super helpful for devices with smaller screens. Realistically, there's only enough space to display one message at a time anyway. It might actually be an idea to disable this threaded-messaging feature entirely for some variants?

todd-herbert:
> > Very nice.
> > On which devices/screens has this been tested next to the T-Deck?
> 
> @GUVWAF I've only tested it on the T-Deck so far since that's all I own.

What I might actually suggest is: instead of modifying the `Screen` class directly, write a new class which derives from `Screen`, then override what you need. Then, where `Screen` is instantiated, use a macro to instead use your derived class.

That way, you can focus on implementing the new features for just T-Deck right now, and gradually edge your way out, one device at a time.

todd-herbert:
Another general heads up: checked for T-Echo, and I don't think this would fit on the device. 
>  region `FLASH' overflowed by 10988 bytes

src/graphics/Screen.cpp
ndnestor: I should've mentioned that I didn't mean for the E-Ink workaround to be a permanent solution. I'm just trying to get to a minimum viable product for such a large update since I think, even in a not perfect state, it's a big step forwards for users. I don't have an E-Ink device to test on so determining if anything I code even works will not be possible. If you'd like to play around with it, that would be great.

Threaded messaging is still useful on small displays since you can scroll to see older messages. I do think this feature should be disabled for some variants as I mentioned [here](https://github.com/meshtastic/firmware/pull/3784#discussion_r1594862401).

ndnestor:
> > > Very nice.
> > > On which devices/screens has this been tested next to the T-Deck?
> > 
> > 
> > @GUVWAF I've only tested it on the T-Deck so far since that's all I own.
> 
> What I might actually suggest is: instead of modifying the `Screen` class directly, write a new class which derives from `Screen`, then override what you need. Then, where `Screen` is instantiated, use a macro to instead use your derived class.
> 
> That way, you can focus on implementing the new features for just T-Deck right now, and gradually expand to more devices later on.

The issue I see with making a T-Deck-specific device is that expanding would mean making classes for other models which seems impractical in the long run as new devices come out. For maintainability, one class that can generically deal with everything would be nice.

src/graphics/Screen.cpp
ndnestor: > From a quick look, it seems like next frame has been disabled in `handleOnPress()` for all devices, and is now intercepted in `handleInputEvent()`, as `meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar_SELECT`, right?

Yes.

> For the record, pressing the user-button right now doesn't toggle the scrolling either on most devices , as the press event is not generally connected to the canned message select event (I think?)

Okay. Linking those scroll functions to the canned message select input should be changed.

> A macro maybe wouldn't a great choice here, because a lot of DIY devices (e.g. Heltec V3) may have a single user button, or may have a more complex keyboard, depending on how the device is built. It wouldn't be fair to inconvenience all these users by forcing them to custom-build firmware from now on.
>
> A lot of this functionality might need to tied to a new user-setting? Possibly there is an existing user-setting which can be used in this particular case (determining how to handle input)?
>
> If you haven't considered it already, I'd also think about how this code interacts with the existing canned-messages module. Also might be an idea to think about how users with canned-message inputs different than the T-Deck would interact with this new feature.

What do you mean by macro?

todd-herbert:
> > > > Very nice.
> > > > On which devices/screens has this been tested next to the T-Deck?
> > > 
> > > 
> > > @GUVWAF I've only tested it on the T-Deck so far since that's all I own.
> > 
> > 
> > What I might actually suggest is: instead of modifying the `Screen` class directly, write a new class which derives from `Screen`, then override what you need. Then, where `Screen` is instantiated, use a macro to instead use your derived class.
> > That way, you can focus on implementing the new features for just T-Deck right now, and gradually expand to more devices later on.
> 
> The issue I see with making a T-Deck-specific device is that expanding would mean making classes for other models which seems impractical in the long run as new devices come out. For maintainability, one class that can generically deal with everything would be nice.

I'm still suggesting using one class for all devices, but use a macro to select the new class. If a variant is built with the macro USE_PERSISTENT_MSG (or whatever suits), it will build with the new class, otherwise it will build with the existing Screen class.

Once it's confirmed that a new device works correctly with the new class, that generic macro can be added to its specific platformio.ini file

Have a look at how the [`Screen` constructor](https://github.com/meshtastic/firmware/blob/master/src/graphics/Screen.cpp#L969-L998) does it. I think something like this might be useful

ndnestor:
> > > > > Very nice.
> > > > > On which devices/screens has this been tested next to the T-Deck?
> > > > 
> > > > 
> > > > @GUVWAF I've only tested it on the T-Deck so far since that's all I own.
> > > 
> > > 
> > > What I might actually suggest is: instead of modifying the `Screen` class directly, write a new class which derives from `Screen`, then override what you need. Then, where `Screen` is instantiated, use a macro to instead use your derived class.
> > > That way, you can focus on implementing the new features for just T-Deck right now, and gradually expand to more devices later on.
> > 
> > 
> > The issue I see with making a T-Deck-specific device is that expanding would mean making classes for other models which seems impractical in the long run as new devices come out. For maintainability, one class that can generically deal with everything would be nice.
> 
> I'm still suggesting using one class for all devices, but use a macro to select the new class. If a variant is built with the macro USE_PERSISTENT_MSG (or whatever suits), it will build with the new class, otherwise it will build with the existing Screen class.
> 
> Once it's confirmed that a new device works correctly with the new class, that generic macro can be added to its specific platformio.ini file
> 
> Have a look at how the [`Screen` constructor](https://github.com/meshtastic/firmware/blob/master/src/graphics/Screen.cpp#L969-L998) does it. I think something like this might be useful

Oh I see. That makes sense to me. Will do!

src/graphics/Screen.cpp
ndnestor: I didn't realize macro refers to a C++ concept. I looked it up and figured it out. I have never coded in C++ before this so I'm learning as I go.

todd-herbert: I wouldn't worry too much about all these finer details right now to be honest. As long it's set up to only build for the T-Deck at first, all this stuff can wait.

ndnestor: Alright, sounds good.

.trunk/trunk.yaml
ndnestor: I reverted the changes made to the generated files. The pipeline fails because it is now missing those generated files. Just to be clear, you wanted me to revert changes made in the directory `src/mesh/generated/meshtastic`, right?

src/graphics/Screen.cpp
ndnestor: Resolving since now E-Ink devices will use the old message frame.

src/platform/esp32/architecture.h
```diff
@@ -87,6 +87,7 @@
 #define HW_VENDOR meshtastic_HardwareModel_TLORA_V2_1_1P8
 #elif defined(T_DECK)
 #define HW_VENDOR meshtastic_HardwareModel_T_DECK
+#define USE_PERSISTENT_MSG
 #elif defined(T_WATCH_S3)
```
caveman99, line 91: This define is best moved to the variant file if you really want it only for the t-deck.

.trunk/trunk.yaml
caveman99: yes, correct. you can ignore the failing pipeline, as long as your local copy compiles. Once the PR is polished, we'll merge the protobuf changes and the pipeline starts to work.

caveman99:
just a general question, since you seem to go to great lengths to implement this. I am not convinced the heavy filesystem operation this relys on will work on the embedded platforms when the number of files grows. You do know realize we have a new UI for the T-Deck ind evelopment that will mimick a phone app more than the current Oled UI? This seems like double effort.

ndnestor:
> just a general question, since you seem to go to great lengths to implement this. I am not convinced the heavy filesystem operation this relys on will work on the embedded platforms when the number of files grows. You do know realize we have a new UI for the T-Deck ind evelopment that will mimick a phone app more than the current Oled UI? This seems like double effort.

@caveman99 Yes, I've heard about that. When I first asked about this feature on the Discord server, I was told that it would, optimistically, take a year. Based on the progress done so far, I would guess that it's going to take decently longer than that. Given the utility of this feature, I figured it is worth it to have this not-so-fancy version for now and get the fancy version later.

src/platform/esp32/architecture.h
todd-herbert: @caveman99 Would you prefer to try implement it everywhere? I'll withdraw my objections if you do think it'd be better.

HarukiToreda: I second this, the Tdeck UI will be out someday, but all other devices will be left behind, this is the best option we've gotten so far. Maybe have a new module and the ability to give the choice of turning this on or not.

todd-herbert: My idea initially was to maybe get it released for T-Deck first, and add other devices once they're tested. There's definitely a lot of edge case stuff that would need to be handled, and some devices probably shouldn't get this feature at all imo, but if the devs want to go ahead and try, I'll support the effort.

src/modules/TextMessageModule.cpp
GUVWAF: For ESP32 it's around [100k](https://docs.espressif.com/projects/esp-faq/en/latest/software-framework/storage/nvs.html#if-data-needs-to-be-stored-or-updated-to-flash-every-minute-can-esp32-nvs-meet-this-requirement) writes, for nRF52 more like [10k](https://devzone.nordicsemi.com/f/nordic-q-a/15408/nrf52-write-erase-flash-cycles-confusion#:~:text=Flash%20write%2Ferase%20cycles%3A%2010,a%20block%20between%20erase%3A%20181). However, even for ESP32 saving to flash takes a considerable amount of time.

As you can see here, previously it would only set the last received message, but it only gets written to flash when the devicestate is written to (e.g. with `saveDeviceStateToDisk()`). So, yes, if power is lost in between, it might be some messages are missed, but I think that's not a big deal.

I think people would find it a shame if it's completely removed for nRF52, especially for @tropho23's Meshenger, who originally opened feature requests #1985 and #2590.

`MemGet::getFreeHeap()` is already implemented for nRF52 as well:
https://github.com/meshtastic/firmware/blob/859fd7c25110c038bcb41aa7ef01d6b603b8ab9b/src/memGet.cpp#L23-L25

ndnestor: From firsthand experience, with the cache being implemented as it currently is, I haven't noticed the time it takes to save messages.

For nRF52 devices, their future is in external storage devices. There is no support for that but non-nRF52 users will really benefit from this. Especially given that this update will only be enabled for the T-Deck and will be enabled for more devices in the future.

HarukiToreda: I have not seen any impact to the speed either and I am using this on heltec v3, wireless paper and wireless tracker.

todd-herbert:
Will variants that exclude these changes at compile time also avoid the extra wear on the flash that this feature involves?

src/modules/TextMessageModule.cpp
GUVWAF: I'm not really concerned about "speed" in a way that it slows messaging down (indeed you wouldn't really notice an additional 500ms delay). However, the action is not asynchronous, so I'm wondering if you miss packets when you receive several short text messages after each other, just like happened before e-ink updates were implemented asynchronously.

bobricius:
What about reserve some array in RAM for cca 10 messages ?

tropho23:
Of course more than 10 would better 🙂

My ideal is the last 30 messages if available memory allows.

Tony Good

src/modules/TextMessageModule.cpp
GUVWAF: Using a Heltec Wireless Paper, I added `millis()` around `nodeDB→saveMessageToDisk(mp);` and it varied from 187-736ms; it’s mostly around 350ms. Considering a message on LongFast usually takes around 500ms, this is a considerable amount of time. The radio chip has place in the buffer for only 1 packet and it's shared between transmit and receive.

I also did a test by sending 4 times 5 text messages with 1 character quickly after each other. Two of them didn’t get acknowledged, and the packets were often received out of order, meaning that retransmissions were needed (which is only done for the 0-hop, so for further hops it would be lost).

With the same test on current master, all 20 packets got acknowledged.

ndnestor:
We've solved the throttling problem by implementing a database file for each "category" instead of individual message files in a file structure. This allows us to save messages in batches which store in a buffer that gets saved every 30 seconds to a single file, limiting our number of writes. We also now execute saving to disk asynchronously which makes the whole process safe from packet-loss.

HarukiToreda:
I tried building this to test but it's not building. Missing "message.pb.h"
```
In file included from src/DisplayFormatters.h:2,
                 from src/DisplayFormatters.cpp:1:
src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
 #include "mesh/generated/meshtastic/message.pb.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
In file included from src/GPSStatus.h:2,
                 from src/gps/GPS.h:5,
                 from src/ButtonThread.cpp:4:
src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
 #include "mesh/generated/meshtastic/message.pb.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
*** [.pio\build\heltec-v3\src\DisplayFormatters.cpp.o] Error 1
*** [.pio\build\heltec-v3\src\ButtonThread.cpp.o] Error 1
In file included from src/modules/Telemetry/Sensor/TelemetrySensor.h:7,
                 from src/modules/Telemetry/Sensor/INA219Sensor.h:6,
                 from src/power.h:41,
                 from src/power.cpp:13:
src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
 #include "mesh/generated/meshtastic/message.pb.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
*** [.pio\build\heltec-v3\src\Power.cpp.o] Error 1
In file included from src/mesh/Default.h:2,
                 from src/PowerFSM.cpp:11:
src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
 #include "mesh/generated/meshtastic/message.pb.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
*** [.pio\build\heltec-v3\src\PowerFSM.cpp.o] Error 1
In file included from src/RedirectablePrint.cpp:2:
src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
 #include "mesh/generated/meshtastic/message.pb.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
*** [.pio\build\heltec-v3\src\RedirectablePrint.cpp.o] Error 1
In file included from src/airtime.cpp:2:
src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
 #include "mesh/generated/meshtastic/message.pb.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
*** [.pio\build\heltec-v3\src\airtime.cpp.o] Error 1
In file included from src/buzz/buzz.cpp:2:
src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
 #include "mesh/generated/meshtastic/message.pb.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
*** [.pio\build\heltec-v3\src\buzz\buzz.cpp.o] Error 1
In file included from src/SerialConsole.cpp:2:
src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
 #include "mesh/generated/meshtastic/message.pb.h"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
*** [.pio\build\heltec-v3\src\SerialConsole.cpp.o] Error 1
```

ndnestor:
> I tried building this to test but it's not building. Missing "message.pb.h"
> 
> ```
> In file included from src/DisplayFormatters.h:2,
>                  from src/DisplayFormatters.cpp:1:
> src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
>  #include "mesh/generated/meshtastic/message.pb.h"
>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> In file included from src/GPSStatus.h:2,
>                  from src/gps/GPS.h:5,
>                  from src/ButtonThread.cpp:4:
> src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
>  #include "mesh/generated/meshtastic/message.pb.h"
>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> *** [.pio\build\heltec-v3\src\DisplayFormatters.cpp.o] Error 1
> *** [.pio\build\heltec-v3\src\ButtonThread.cpp.o] Error 1
> In file included from src/modules/Telemetry/Sensor/TelemetrySensor.h:7,
>                  from src/modules/Telemetry/Sensor/INA219Sensor.h:6,
>                  from src/power.h:41,
>                  from src/power.cpp:13:
> src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
>  #include "mesh/generated/meshtastic/message.pb.h"
>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> *** [.pio\build\heltec-v3\src\Power.cpp.o] Error 1
> In file included from src/mesh/Default.h:2,
>                  from src/PowerFSM.cpp:11:
> src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
>  #include "mesh/generated/meshtastic/message.pb.h"
>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> *** [.pio\build\heltec-v3\src\PowerFSM.cpp.o] Error 1
> In file included from src/RedirectablePrint.cpp:2:
> src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
>  #include "mesh/generated/meshtastic/message.pb.h"
>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> *** [.pio\build\heltec-v3\src\RedirectablePrint.cpp.o] Error 1
> In file included from src/airtime.cpp:2:
> src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
>  #include "mesh/generated/meshtastic/message.pb.h"
>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> *** [.pio\build\heltec-v3\src\airtime.cpp.o] Error 1
> In file included from src/buzz/buzz.cpp:2:
> src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
>  #include "mesh/generated/meshtastic/message.pb.h"
>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> *** [.pio\build\heltec-v3\src\buzz\buzz.cpp.o] Error 1
> In file included from src/SerialConsole.cpp:2:
> src/mesh/NodeDB.h:13:10: fatal error: mesh/generated/meshtastic/message.pb.h: No such file or directory
>  #include "mesh/generated/meshtastic/message.pb.h"
>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> *** [.pio\build\heltec-v3\src\SerialConsole.cpp.o] Error 1
> ```

@HarukiToreda I was told to remove that file from the PR since it can be generated using `nanopb`. You can manually add `message.pb.c` from [here](https://github.com/meshtastic/firmware/blob/a9c49e883e8f50fde4914d0bb8109621efffaec2/src/mesh/generated/meshtastic/message.pb.c) and `message.pb.h` from [here](https://github.com/meshtastic/firmware/blob/a9c49e883e8f50fde4914d0bb8109621efffaec2/src/mesh/generated/meshtastic/message.pb.h) by copying and pasting the file contents.

ndnestor:
Can somebody review this please?

HarukiToreda:
> Can somebody review this please?

even if I manually add the message files missing, this still won't build, we shouldn't have to manually add these files.

```
Compiling .pio\build\heltec-v3\src\input\kbMatrixBase.cpp.o
src/graphics/Screen.cpp:960:13: error: redefinition of 'void graphics::drawBattery(OLEDDisplay*, int16_t, int16_t, uint8_t*, const meshtastic::PowerStatus*)'
 static void drawBattery(OLEDDisplay *display, int16_t x, int16_t y, uint8_t *imgBuffer, const PowerStatus *powerStatus)
             ^~~~~~~~~~~
src/graphics/Screen.cpp:430:13: note: 'void graphics::drawBattery(OLEDDisplay*, int16_t, int16_t, uint8_t*, const meshtastic::PowerStatus*)' previously defined here
 static void drawBattery(OLEDDisplay *display, int16_t x, int16_t y, uint8_t *imgBuffer, const PowerStatus *powerStatus)
             ^~~~~~~~~~~
src/graphics/Screen.cpp:1420:6: error: redefinition of 'bool graphics::deltaToTimestamp(uint32_t, uint8_t*, uint8_t*, int32_t*)'
 bool deltaToTimestamp(uint32_t secondsAgo, uint8_t *hours, uint8_t *minutes, int32_t *daysAgo)
      ^~~~~~~~~~~~~~~~
src/graphics/Screen.cpp:890:6: note: 'bool graphics::deltaToTimestamp(uint32_t, uint8_t*, uint8_t*, int32_t*)' previously defined here
 bool deltaToTimestamp(uint32_t secondsAgo, uint8_t *hours, uint8_t *minutes, int32_t *daysAgo)
      ^~~~~~~~~~~~~~~~
src/graphics/Screen.cpp:960:13: warning: 'void graphics::drawBattery(OLEDDisplay*, int16_t, int16_t, uint8_t*, const meshtastic::PowerStatus*)' defined but not used [-Wunused-function]
 static void drawBattery(OLEDDisplay *display, int16_t x, int16_t y, uint8_t *imgBuffer, const PowerStatus *powerStatus)
             ^~~~~~~~~~~
In file included from src/graphics/Screen.cpp:44:
src/graphics/images.h:197:22: warning: 'poo' defined but not used [-Wunused-variable]
 static unsigned char poo[] PROGMEM = {
                      ^~~
src/graphics/images.h:186:22: warning: 'heart' defined but not used [-Wunused-variable]
 static unsigned char heart[] PROGMEM = {
                      ^~~~~
src/graphics/images.h:175:22: warning: 'devil' defined but not used [-Wunused-variable]
 static unsigned char devil[] PROGMEM = {
                      ^~~~~
src/graphics/images.h:165:22: warning: 'fog' defined but not used [-Wunused-variable]
 static unsigned char fog[] PROGMEM = {
                      ^~~
src/graphics/images.h:154:22: warning: 'cloud' defined but not used [-Wunused-variable]
 static unsigned char cloud[] PROGMEM = {
                      ^~~~~
src/graphics/images.h:143:22: warning: 'rain' defined but not used [-Wunused-variable]
 static unsigned char rain[] PROGMEM = {
                      ^~~~
src/graphics/images.h:132:22: warning: 'sun' defined but not used [-Wunused-variable]
 static unsigned char sun[] PROGMEM = {
                      ^~~
src/graphics/images.h:115:22: warning: 'deadmau5' defined but not used [-Wunused-variable]
 static unsigned char deadmau5[] PROGMEM = {
                      ^~~~~~~~
src/graphics/images.h:104:22: warning: 'cowboy' defined but not used [-Wunused-variable]
 static unsigned char cowboy[] PROGMEM = {
                      ^~~~~~
src/graphics/images.h:93:22: warning: 'wave_icon' defined but not used [-Wunused-variable]
 static unsigned char wave_icon[] PROGMEM = {
                      ^~~~~~~~~
src/graphics/images.h:82:22: warning: 'haha' defined but not used [-Wunused-variable]
 static unsigned char haha[] PROGMEM = {
                      ^~~~
src/graphics/images.h:71:22: warning: 'bang' defined but not used [-Wunused-variable]
 static unsigned char bang[] PROGMEM = {
                      ^~~~
src/graphics/images.h:61:22: warning: 'question' defined but not used [-Wunused-variable]
 static unsigned char question[] PROGMEM = {
                      ^~~~~~~~
src/graphics/images.h:51:22: warning: 'thumbdown' defined but not used [-Wunused-variable]
 static unsigned char thumbdown[] PROGMEM = {
                      ^~~~~~~~~
src/graphics/images.h:43:22: warning: 'thumbup' defined but not used [-Wunused-variable]
 static unsigned char thumbup[] PROGMEM = {
                      ^~~~~~~
*** [.pio\build\heltec-v3\src\graphics\Screen.cpp.o] Error 1
```

ndnestor:
> > Can somebody review this please?
> 
> even if I manually add the message files missing, this still won't build, we shouldn't have to manually add these files.
> 

I would told to not include them here: https://github.com/meshtastic/firmware/pull/3784#discussion_r1597515177

Those are strange errors. I have never messed with any of those functions. I also do not get them on my end, even if I build for a Heltec v3. I'll look into it

src/mesh/NodeDB.cpp
```diff
@@ -755,6 +821,171 @@ void NodeDB::saveDeviceStateToDisk()
...
+        if (mp.to == nodeDB->getMeshNodeByIndex(i)->num)
+            return true;
+
+    return false;
+}
+
+void NodeDB::addMessage(const meshtastic_MeshPacket &mp)
+{
+    auto *msg = new meshtastic_Message();
+    msg->from = mp.from;
```
GUVWAF, line 856: Please use `getFrom()` again here, as packets from the phone originate with `from == 0`.

robertfisk:
This feature is very cool and also beneficial for some meshtastic hardware I'm developing. @ndnestor Would you mind if I fork this and send you PRs over the next couple of weeks to help get it merged?

ndnestor:
> This feature is very cool and also beneficial for some meshtastic hardware I'm developing. @ndnestor Would you mind if I fork this and send you PRs over the next couple of weeks to help get it merged?

As far as I'm aware, the only thing that needs to be changed to get this merged is that last thing @GUVWAF commented on. I'm about to push that now. But if anything else comes up that needs to be changed, I welcome the help! Thanks!

thatSFguy:
This is exactly what I've been wanting for my t-lora 1.6 kiss messenger.  Looks like the initiative lost steam, but it would be great to have this feature.  
Kiss

mcules:
I assume that this pull request has not been merged yet. What still needs to be done to make that happen, or has it become obsolete because the functionality has already been implemented in the firmware in a different way?

garthvh:
> I assume that this pull request has not been merged yet. What still needs to be done to make that happen, or has it become obsolete because the functionality has already been implemented in the firmware in a different way?

It has not been merged, and the way it was implemented raised some concerns with the core team maintaining the firmware that were not addressed, it is essentially abandoned at this point.

Issue #3751, opened 2024-04-30
Enhancement: Air Quality values getting cut off


HarukiToreda:
When Telemetry is turned on, the Air quality values are getting cut off at the bottom of the screen. I think reducing the header size will give enough room for this to show on smaller screens like on the Heltec V3.

![20240430_000502](https://github.com/meshtastic/firmware/assets/116696711/4aa99a4a-1b2d-4396-8053-04557f1ed022)

todd-herbert:
In EnvironmentTelemetry.cpp, it might be an idea to modify the `drawFrame` method, so that it checks `display->height()`. If the height is below some value, could then use`FONT_SMALL` everywhere `FONT_MEDIUM` is used?

HarukiToreda:
> In EnvironmentTelemetry.cpp, it might be an idea to modify the `drawFrame` method, so that it checks `display->height()`. If the height is below some value, could then use`FONT_SMALL` everywhere `FONT_MEDIUM` is used?

Thank you, I think I have a working rough draft, I will wait for my other pull request to post the fix for this. 

![20240430_011336](https://github.com/meshtastic/firmware/assets/116696711/ebed7847-6d41-4ddd-bd4d-4976f1720f91)

paoloponzano:
Very nice fix, thank you :)

thebentern:
Closed by #3785

PR #3785, opened 2024-05-05
Fix for EnvironmentTelemetry Screen

HarukiToreda:
Fixing issue for issues #3751 for telemetry screen. Made it so that all information fits in the screen properly

Before
After

### DIFF
--- a/src/modules/Telemetry/EnvironmentTelemetry.cpp
+++ b/src/modules/Telemetry/EnvironmentTelemetry.cpp
@@ -172,8 +172,8 @@ void EnvironmentTelemetryModule::drawFrame(OLEDDisplay *display, OLEDDisplayUiSt
         return;
     }
 
-    // Display "From: ..." on its own
-    display->drawString(x, y, "From: " + String(lastSender) + "(" + String(agoSecs) + "s)");
+    // Display "Env. From: ..." on its own
+    display->drawString(x, y, "Env. From: " + String(lastSender) + "(" + String(agoSecs) + "s)");
 
     String last_temp = String(lastMeasurement.variant.environment_metrics.temperature, 0) + "°C";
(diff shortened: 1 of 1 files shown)


src/modules/Telemetry/EnvironmentTelemetry.cpp
```diff
@@ -151,48 +151,49 @@ uint32_t GetTimeSinceMeshPacket(const meshtastic_MeshPacket *mp)
...
+    if (lastMeasurement.variant.environment_metrics.voltage != 0) {
+        display->drawString(x, y += fontHeight(FONT_SMALL), "Volt/Cur: " + String(lastMeasurement.variant.environment_metrics.voltage, 0) + "V / " + String(lastMeasurement.variant.environment_metrics.current, 0) + "mA");
+    }
+
+    if (lastMeasurement.variant.environment_metrics.iaq != 0) {
         display->drawString(x, y += fontHeight(FONT_SMALL), "IAQ: " + String(lastMeasurement.variant.environment_metrics.iaq));
-    if (lastMeasurement.variant.environment_metrics.distance != 0)
-        display->drawString(x, y += fontHeight(FONT_SMALL),
-                            "Water Level: " + String(lastMeasurement.variant.environment_metrics.distance, 0) + "mm");
+    }
```
caveman99, line 196: Please don't drop the distance display from the radar sensor.

caveman99 (changes requested):

src/modules/Telemetry/EnvironmentTelemetry.cpp
HarukiToreda: Sorry didn't notice. I put it back and reorganized lines as they used to keep it neat

PR #3792, opened 2024-05-05
Add Sensirion SHT4X sensors

pr000t:
Add SHT45 sensor.
For the SHT40 and SHT41, the registerValue in ScanI2CTwoWire.cpp need to be added.


platformio.ini
```diff
@@ -133,3 +133,4 @@ lib_deps =
   https://github.com/lewisxhe/SensorLib#27fd0f721e20cd09e1f81383f0ba58a54fe84a17
   adafruit/Adafruit LSM6DS@^4.7.2
   mprograms/QMC5883LCompass@^1.2.0
+  https://github.com/Sensirion/arduino-i2c-sht4x
```
caveman99, line 136: Please add a release tag. Probably https://github.com/Sensirion/arduino-i2c-sht4x.git#1.1.0

src/mesh/generated/meshtastic/telemetry.pb.h
```diff
@@ -45,7 +45,9 @@ typedef enum _meshtastic_TelemetrySensorType {
     /* BMP085/BMP180 High accuracy temperature and pressure (older Version of BMP280) */
     meshtastic_TelemetrySensorType_BMP085 = 15,
     /* RCWL-9620 Doppler Radar Distance Sensor, used for water level detection */
-    meshtastic_TelemetrySensorType_RCWL9620 = 16
+    meshtastic_TelemetrySensorType_RCWL9620 = 16,
```
caveman99, line 48: Please don't check in generated files.

src/detect/ScanI2CTwoWire.cpp
```diff
@@ -292,7 +292,18 @@ void ScanI2CTwoWire::scanPort(I2CPort port)
...
+                if (registerValue == 0x11a2) {
+                    type = SHT4X;
+                    LOG_INFO("SHT4X sensor found\n");
+                } else {
+                    type = SHT31;
+                    LOG_INFO("SHT31 accelerometer found\n");
+                }
+
+                break;
+
```
caveman99, line 306: If this only works for the SHT45 sensor as per your comment, please name the macro accordingly to leave room for 41 and 42

caveman99 (changes requested):

src/detect/ScanI2CTwoWire.cpp
pr000t: SHT31 and all SHT4X share the same ADDR (0x44).
Unfortunately I don't have the other sensors to test for the registerValue but like all SHT4x sensors use the same library, i would go for something like:
`if (registerValue == 0x11a2 || registerValue == <sht40> || registerValue == <sht41>)`

caveman99 (changes requested):
Looks good to me. Just a technicality: can you please allow maintainer edits on your branch? I need to fast-forward it to HEAD to merge it, and i can't do that without maintainer edit permission.

caveman99 (approved):

src/detect/ScanI2CTwoWire.cpp
```diff
@@ -292,7 +292,18 @@ void ScanI2CTwoWire::scanPort(I2CPort port)
...
 
-                SCAN_SIMPLE_CASE(SHT31_ADDR, SHT31, "SHT31 sensor found\n")
+            case SHT31_4x_ADDR:
+                registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0x89), 2);
+                if (registerValue == 0x11a2) {
+                    type = SHT4X;
+                    LOG_INFO("SHT4X sensor found\n");
+                } else {
+                    type = SHT31;
+                    LOG_INFO("SHT31 accelerometer found\n");
```
thebentern, line 302: Copy paste issue? Should say sensor instead of accelerometer

pr000t: indeed, fixed. Thank you

thebentern (approved):

PR #3793, opened 2024-05-05
Wiphone

caveman99:
very basic support. keyboard and sensors still missing


markbirss:
@caveman99 

Looking at the schematics they also wired DIO1 and DIO2 to pins, which would if you were able to swap out the RFM95W with a NiceRF1262, could work quite well as those are the pins needed to work

caveman99:
This is by no means finite. I just made the included Lora backplate work ;-)

markbirss:
@caveman99 understood, the other breakout just looks bit odd

PR #3803, opened 2024-05-06
Send fixed position to mesh after setting it

### LINKED ISSUE #3783: [Feature Request]: Broadcast newly set fixed position immediately
### Platform

NRF52, ESP32

### Description

While the new set fixed position admin message is working well the fixed position is being broadcast on the regular position broadcast interval so it does not go out immediately.  Since people are using this feature often on hard to access nodes it would be good if we can immediately broadcast the position packet the first time so that the verification that the position set properly is available in the apps.

thebentern:
Closes #3783


garthvh (approved):
Works perfectly, I think this will offer the feedback people want.

ndoo:
Is this code path possibly making the gps.enable from my earlier PR #3758 redundant or is it touching some other state enabling GPS doesn't?

Issue #3811, opened 2024-05-06
[Feature Request]: Support Ublox M9N  


hdngr:
### Platform

NRF52, ESP32

### Description

# Overview
Canaryone is designed with the Ublox M9N.

# Problem
M6,7,8,9 and especially M10 all run different versions of firmware, and may communicate with dramatically different protocols.

Currently, the GPS module handles all chips that are `!M10`.

# Solution
Refactor code to selectively apply commands to chips given there version and protocol needs.

**As of May 6th, 2024 - DRAFT PRs to follow**

hdngr:
cc @lolsborn @GPSFan

GPSFan:
The u-blox standard precision product line can be roughly divided into 2 categories, Pre F9/M9 and post F9/M9. The F9/M9 generation (Chip type 9140) introduced the VALSET/VALGET/VALDEL key value pair configuration framework. It also retained some of the older configuration protocol. As the newer versions of the protocol were introduced, the older configuration method has been deprecated and in newer generations much has been removed. The Neo-M9N firmware is currently 4.04, and hasn't been updated since lat Sept, 2020.
The M9 is a hungry beast, but also is a very good receiver, combining an M9 and an NRF52 is somewhat puzzling as even in power save mode the M9 consumes more than an M10 in full acquisition mode.
The current GPS.cpp handles u-blox generations 6,7,8 and 10. Although the M10 support is not as complete as I would like it to be, and the support has been sort of "patched together" over time. A more structured approach might be a goal for Meshtastic V3. There is also the support for non-u-blox receivers, as they can be much more cost effective GNSS solutions, and there are a lot of variations in those receivers.

thebentern:
Closing this for now

PR #3816, opened 2024-05-07
Gps research

hdngr:
Begin researching M9 integration re:  #3811.

A couple of things:
1.  Please provide feedback on if this overall structure looks alright.  I don't think I'm breaking anything, but without any unit testing it is kind of a free for all
2. @GPSFan it would be helpful if you could give some feedback on what the critical configs are to make the M9 functional...  We'll translate them into `32.01`

Feel free to write them as comments in the code here or in #3811.

cc @lolsborn

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -396,7 +396,7 @@ bool GPS::setup()
             gnssModel = GNSS_MODEL_UNKNOWN;
         }
 #else
-        gnssModel = GNSS_MODEL_UC6580;
+        gnssModel = GNSS_MODEL_UC6580;∏
 #endif
 
         if (gnssModel == GNSS_MODEL_MTK) {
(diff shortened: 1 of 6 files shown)


GPSFan:
I have an M9 receiver and will be able to test it on my ESP32 based system, but until I get my I/O breakout for my RAK system (on order, shipped) I won't be able to do any testing on an NRF52.
I will review the attached code.

hdngr:
> The hwVersion for the M9 is 90000 not 190000.

That would be logical, wouldn't it.  But its not, lol.

image

hdngr:
> I have an M9 receiver and will be able to test it on my ESP32 based system, but until I get my I/O breakout for my RAK system (on order, shipped) I won't be able to do any testing on an NRF52. I will review the attached code.

Totally.  Just feeding ideas would be helpful.  Have not dug into if they are defaults yet, but seems like setting `RMC` and `GCA` and then turning off all of the other chatty stuff might be the key thing to fix.

GPSFan:
Well, I'll be... shows you what a few years of not using the old M9 does..

src/gps/GPS.cpp
```diff
@@ -646,7 +652,33 @@ bool GPS::setup()
...
+                msglen = makeUBXPacket(0xb5, 0x62, sizeof(_message_1HZ), _message_1HZ);
+                _serial_gps->write(UBXscratch, msglen);
+
+                if (getACK(0x06, 0x08, 500) != GNSS_RESPONSE_OK) {
+                    LOG_WARN("Unable to set GPS update rate.\n");
+                } else {
+                    LOG_INFO("Successfully set GPS update rate.!\n");
+                }
+                delay(1000);
+            }
```
hdngr, line 678: What other messages should go here?  Not super familiar with all of these protocols.  A dump of the hex code for M9 would be helpful BUT, a pointer directly to config.  Would be helpful.

For instance, what does `_message_VALSET_DISABLE_NMEA_RAM` correspond to?  Would it be [3.10.13 UBX-CFG-NMEA (0x06 0x17)](https://content.u-blox.com/sites/default/files/u-blox-M9-SPG-4.04_InterfaceDescription_UBX-21022436.pdf)

hdngr: @GPSFan ^^...  cc: @lolsborn 

@GPSFan if you made comments in here with what configs our commands you think we absolutely need, we can grind them out.  We'd appreciate it! 🙏

fifieldt:
Hi @hdngr @lolsborn , are you still interested in M9 support? I've just merged some changes to the gps code which will make it easier to add M9 specific things. Though, the new code conflicts pretty badly with this pull request, so let's move back to the issue #3811 for discussion?

PR #3840, opened 2024-05-09
T-Echo screen and button performance

todd-herbert:
Two small tweaks which I believe improve the T-Echo UX. Changes affect only the T-Echo.

### Disable ghost-pixel tracking
"Ghost pixels" referring to the after-image left by setting black pixels to white with a fast refresh.
These artifacts are fairly faint for this model of display, and there doesn't seem to be much benefit in tracking / actively avoiding them with the T-Echo. Disabling this results in fewer full-refreshes; a smoother experience.

### Increase the "click ms" value
This value determines how long a button waits to register a double-press, before accepting the current input as a single click. This PR increases the value from 250ms to 400ms. I don't believe this is perceptible when single-pressing, as the display refresh operation is fairly slow.

The benefit is that the double, triple, and quad presses are easier to execute.

To enable this change, the PR allows for variant-specific "click ms" values. If not specified, the current values are used as a default.

### DIFF
--- a/src/ButtonThread.cpp
+++ b/src/ButtonThread.cpp
@@ -52,8 +52,8 @@ ButtonThread::ButtonThread() : OSThread("Button")
 
 #if defined(BUTTON_PIN) || defined(ARCH_PORTDUINO)
     userButton.attachClick(userButtonPressed);
-    userButton.setClickMs(250);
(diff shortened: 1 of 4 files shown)


runestone74:
Really nice, much easier with quad click for backlight now :)

PR #3845, opened 2024-05-10
Revert Sensirion SHT4X sensors due to size constraints

thebentern:
Reverts meshtastic/firmware#3792

Hopefully we can find an alternative library


pr000t:
I went through the manufacturer library but I can suggest the Adafruit library if that is preferable

caveman99:
Apart from that issue here, which turned out to be a red herring the Adafruit sensor lib is preferred for our firmware since it re-uses many parts of the sensor stack. if you can alter the support to use the adafruit libs that pr would vcertainly be welcomed :-)

PR #3860, opened 2024-05-11
exclude serial module for T-Echo, saves 3000 bytes

caveman99:


variants/t-echo/variant.h
```diff
@@ -206,6 +206,9 @@ External serial flash WP25R1635FZUIL0
 // To debug via the segger JLINK console rather than the CDC-ACM serial device
 // #define USE_SEGGER
 
+// T-Echo does not have a free serial port for this module
```
thebentern, line 209: We do have serial.override_console_serial_port that lets you use the NMEA and Cal Topo modes to stream sentences over USB. Maybe we can gate those off with a finer grained macro? 
Exclude serial is still a good one to have for really small platforms like STM32 possibly?

caveman99: I didn't consider that. Will revert.

PR #3889, opened 2024-05-14
Don't reboot for certain config prefs and make accelerometer thread re-entrant

### LINKED ISSUE #3780: [Feature Request]: Only Reboot on save if necessary for Device, Display and Ambient Lighting Config
### Platform

NRF52, ESP32

### Description

Similar to the changes made for LoRa Config to support saving the ignoreIncoming array without a reboot, it would be good if we can update some of the settings to not require a device reboot, especially if we can provide immediate feedback on the device screen / LED's.

### Device
* Double Tap as Button
* Heartbeat LED
* Timezone

## Display
* Always point north
* Wake screen on tap or motion
* Flip screen
* Screen on for
* Carousel Interval
* GPS Format 
* Display Units

## Ambient Lighting
* State
* Color
* Current

thebentern:
Closes #3780 minus ambient lighting, which seems tricky at the moment

Needs some testing

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -228,6 +228,8 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
             if (gps != nullptr)
                 gps->enable();
 #endif
+            // Send our new fixed position to the mesh for good measure
+            positionModule->sendOurPosition();
         }
         break;
     }
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -26,6 +26,9 @@
 #if !MESHTASTIC_EXCLUDE_GPS
 #include "GPS.h"
 #endif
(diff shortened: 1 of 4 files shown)


src/modules/AdminModule.cpp
```diff
@@ -221,8 +224,6 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
             nodeDB->setLocalPosition(r->set_fixed_position);
             config.position.fixed_position = true;
             saveChanges(SEGMENT_DEVICESTATE | SEGMENT_CONFIG, false);
-            // Send our new fixed position to the mesh for good measure
```
caveman99, line 224: This seems accidental removal?

thebentern: That's what I get for copy and pasting changes from another branch downstream of master changes

caveman99 (approved):

PR #3894, opened 2024-05-15
Lock Portduino to MAGIC_USB_BATTERY_LEVEL for now

jp-bennett:


thebentern:
its-magic-shia-labeouf-gif-300x168

thebentern:
Sad... github did not preserve gif

jp-bennett:
This does get the Pi to show plugged in rather than at 0 battery.

PR #3895, opened 2024-05-15
fix native compilation for linux PCs

mverch67:
- fix compiler guard PORTDUINO_LINUX_HARDWARE
- fix logic for GPIO_lines search max loop

It looks like the native target is running only on machines that have SPI and GPIO (e.g. raspberry), otherwise it will crash with a message that spidev0.0 cannot be opened. It's not simulating SpiChip anymore.

Note: this PR relies on https://github.com/meshtastic/framework-portduino/pull/27 and needs pull in the corresponding platform-native.

### DIFF
--- a/arch/portduino/portduino.ini
+++ b/arch/portduino/portduino.ini
@@ -1,6 +1,6 @@
 ; The Portduino based sim environment on top of any host OS, all hardware will be simulated
 [portduino_base]
-platform = https://github.com/meshtastic/platform-native.git#9881bf3721d610cccacf5ae8e3a07839cce75d63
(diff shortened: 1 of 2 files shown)


jp-bennett:
> For tests I've added a target _native-linux-pc_, which runs well on my fedora linux PC. Should we add that target permanently?

I don't love it. At least not for working around an issue.

mverch67:
> > For tests I've added a target _native-linux-pc_, which runs well on my fedora linux PC. Should we add that target permanently?
> 
> I don't love it. At least not for working around an issue.

I think the issue was introduced in #3049 by defining PORTDUINO_LINUX_HARDWARE. Normally this could be detected during runtime, or configured into the config file.

caveman99 (approved):

PR #3905, opened 2024-05-16
automatically propose to setup virtualenv if platformio is having issues in native build

Jorropo:
I'm not sure it's quite the right thing to do, it tooks me a dozen minutes to find which version of platformio would work when trying to build on ubuntu jammy.

### DIFF
--- a/bin/build-native.py
+++ b/bin/build-native.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+
+import subprocess
+import sys
+
+from readprops import readProps
+
+
+def run(*command, check=True):
+    subprocess.run(command, check=check)
+
+
+def bash(command: str, /, *, check=True):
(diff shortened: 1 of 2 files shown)


Jorropo:
> ```
> bin/build-native.py:3:0
>  3:0  low  Consider possible security implications associated with the subprocess module.  bandit/B404
> 10:0  low  subprocess call - check for execution of untrusted input.                       bandit/B603
> ```

Yes, I know, that the point, is there a way to to acknowledge theses so CI is happy ?

caveman99:
Same problem is on your other not-so-trivial PR. Newer trunk is really picky about these things. I'll have a look how to either tell it to not treat these as warnings or ignore 'low' category alltogether.

Jorropo:
I got a better idea

PR #3926, opened 2024-05-17
Amend T-Deck variant.h to improve readability 

joneyes:
Changes to improve readability of T-Deck TFT:
Set T-Deck screen colour to white rather than green. Increase default brightness to 200

### DIFF
--- a/variants/t-deck/variant.h
+++ b/variants/t-deck/variant.h
@@ -18,7 +18,9 @@
 #define TFT_OFFSET_ROTATION 0
 #define SCREEN_ROTATE
 #define SCREEN_TRANSITION_FRAMERATE 5
-#define BRIGHTNESS_DEFAULT 130 // Medium Low Brightness
+#define BRIGHTNESS_DEFAULT 200 // Medium Brightness
+// Set a white on black display
+#define TFT_MESH COLOR565(0xFF, 0xFF, 0xFF)
 
 #define HAS_TOUCHSCREEN 1
 #define SCREEN_TOUCH_INT 16


joneyes:
I had previously achieved the same result in a different way. When I posted images on a Facebook group, there was significant interest. 
It's not just me, a lot of users find the default  green Tdeck characters unreadable in sunlight.
This was the simplest fix I could do. If you can implement it a better way, please go ahead, but I suspect it will be a lot more complex?

mverch67:
Increasing the default brightness will also increase the power consumption of the TFT display and thus reduce the runtime on batteries. As already suggested you can use the keyboard control to adapt the brightness, but I would not touch the default as not all users know how to change the brightness.

joneyes:
Ok, I'll take the adjustable brightness comment.

Thoughts on the colour change? Everyone that has seen it prefers white text as it is much more readable in bright sunlight.

I am happy to raise a retrospective enhancement request, to use it to gauge user opinions?

ChrDel:
> > Ok, I'll take the adjustable brightness comment.
> > Thoughts on the colour change? Everyone that has seen it prefers white text as it is much more readable in bright sunlight.
> > I am happy to raise a retrospective enhancement request, to use it to gauge user opinions?
> 
> I don’t see the need to change it. There’s been little to no issue so far. If there’s a personal preference, then it would be something folks can manually compile for themselves.

I dont agree with that statement. I’m actually running custom builds now only to have a white text… For people using devices in sunny places there is no way to see anything on this screen even at full brightness… white does make it more readable… If I understood how to code this I would even go with an actual white background and black text which is the most readable outdoor in the sun…

joneyes:
I've already made a change to support it, and submitted the pull request.
I don't think it's going anywhere as there's no accompanying feature  request.
It will need submitting again to back out the increased default brightness and I'll do that tomorrow when I am back home. Thank-you for supporting my opinion demonstrating that it's not a single person whim.

ChrDel:
> > > > Ok, I'll take the adjustable brightness comment.
> > > > Thoughts on the colour change? Everyone that has seen it prefers white text as it is much more readable in bright sunlight.
> > > > I am happy to raise a retrospective enhancement request, to use it to gauge user opinions?
> > > 
> > > 
> > > I don’t see the need to change it. There’s been little to no issue so far. If there’s a personal preference, then it would be something folks can manually compile for themselves.
> > 
> > 
> > I dont agree with that statement. I’m actually running custom builds now only to have a white text… For people using devices in sunny places there is no way to see anything on this screen even at full brightness… white does make it more readable… If I understood how to code this I would even go with an actual white background and black text which is the most readable outdoor in the sun…
> 
> And I don’t agree with yours. I use mine outdoors. Is it sometimes difficult to see? Sure, as is any screen. But I don’t think it’s as difficult as is being made out to be.

I think it's sad to dismiss people's opinion based on your own and not to offer options... 
This is an open source project that feels very close to the opinion of a few people and not open to what others things or may need... I'm really glad it works for you. Unfortunately for others it doesn't but this is being dismissed and ignored... 
If you really don't have a problem either your eye sight is amazing or the sun where you live is not that bright. 
Some people have disabilities and other live in scorching deserts burned by the sun others in rainy places. Why is this supposed to be only working for people that are young, healthy and live in low sun to no sun settings? 
If you do not like white just because green is cooler and is the meshtastic color, why not just offer an option for users to chose what works best for them and be all inclusivee instead of discriminate amongst people disabilities, or place they live...

PR #3934, opened 2024-05-19
Added "Hops away" on display

capslockapocalypse:
Added in logic that displays "hops away" instead of signal strength on the display when the node is more than 0 hops away.
Any feedback to tidy up is appreciated.


caveman99 (approved):

thebentern (approved):

capslockapocalypse:
Apologies If I'm not doing this right, first time using github and contributing. I'm assuming I need to get the Trunk Check Runner to pass? and that it's hanging on the extensions file.

thebentern:
> Apologies If I'm not doing this right, first time using github and contributing. I'm assuming I need to get the Trunk Check Runner to pass? and that it's hanging on the extensions file.

Everything looks good on that front. You just need to revert your commit to the .vscode/extensions.json file. :-)

capslockapocalypse:
Thank you for your help and patience, if this doesn't work I'll just retry the whole pull request.

thebentern (approved):

HarukiToreda:
> Added in logic that displays "hops away" instead of signal strength on the display when the node is more than 0 hops away. Any feedback to tidy up is appreciated.

do you have some screenshots to show it looks?

PR #3935, opened 2024-05-19
Screen changes from time deltas to timestamps after 15 minutes

todd-herbert:
Changes the time deltas to timestamps in some situations, ~~for E-Ink displays.~~ **extended to all displays**

Conditions:
  * Timezone must be set
  * Timestamp must be at least 15 minutes in the past. Until this point, time delta is displayed

Impact:
  * E-Ink UI is more static, fewer refreshes: avoids refreshing every minute to update time-deltas
  * If configured correctly (if timezone set), much higher screen-timeout values are suitable (longer before screensaver appears)

___

Allows users to disable the E-Ink screensaver, by setting screen timeout to zero.
If disabled, a cautionary message is printed to the on-screen log buffer at startup.

### DIFF
--- a/src/PowerFSM.cpp
+++ b/src/PowerFSM.cpp
@@ -348,12 +348,18 @@ void PowerFSM_setup()
 
     powerFSM.add_transition(&stateDARK, &stateDARK, EVENT_CONTACT_FROM_PHONE, NULL, "Contact from phone");
 
-    powerFSM.add_timed_transition(&stateON, &stateDARK,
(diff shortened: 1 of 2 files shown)


caveman99:
Looking good to me. One point: is it useful to implement alternating behaviour for eInk only? i like the 15 minute cutoff for the time display, why not just enable it for all displays?

todd-herbert:
Can certainly enable it everywhere if you think it's useful. It was just me trying to tread lightly!

todd-herbert:
Just making minor adjustments

PR #3937, opened 2024-05-20
Fix GPS toggle on Heltec v3

slabua:
The definition of PIN_GPS_EN was missing from the heltec_v3 variant.
Originally, I thought it was removed for some reason, and perhaps was it indeed? But while testing this change, it seems like I have encountered no malfunctioning with the GPS.
If I am missing some important detail, I'd be happy to hear some news about it.

### DIFF
--- a/variants/heltec_v3/variant.h
+++ b/variants/heltec_v3/variant.h
@@ -18,6 +18,8 @@
 #define ADC_ATTENUATION ADC_ATTEN_DB_2_5 // lower dB for high resistance voltage divider
 #define ADC_MULTIPLIER 4.9
 
+#define PIN_GPS_EN 46 // GPS power enable pin
+
 #define USE_SX1262
 
 #define LORA_DIO0 -1 // a No connect on the SX1262 module


b8b8:
Heltec V3 doesnt normally have a GPS and an external one needs to be added by the user. The pins you actually use for TX RX and EN are set by the user with the app. Why explicitly define them now and remove configurability?

slabua:
That's what I thought too, but without that line, the specified GPIO on the app never turns off when triple clicking, despite the message showing up on the screen.

todd-herbert:
> That's what I thought too, but without that line, the specified GPIO on the app never turns off when triple clicking, despite the message showing up on the screen.

I remember we chatted about this on the discord server recently. A couple of people have been looking into it since then, it might be worth jumping back over and having a read of [#alphanauts-testing](https://discord.com/channels/867578229534359593/1039674919346249728) and then the [thread that's splintered off](https://discord.com/channels/867578229534359593/1241937144382230588).

It seems like for the most part, the code is actually working as intended. `PIN_GPS_EN` apparently only changes if your [GPS Update Interval setting](https://meshtastic.org/docs/configuration/radio/position/#gps-update-interval) is longer than 15 minutes (+ however long it takes to get a GPS lock).

It seems like the is probably one bug, where a triple press won't move the GPS hardware from a "standby" state to a true "powered-off" state.

slabua:
@todd-herbert thanks! yes sorry I couldn't find anymore the thread on discord the other day.
I will check on discord and also try different timeouts.
So far I have tested the voltage on the pin itself, and across the gps module power pins upon triple clicking, to rule out the transistor situation.

PR #3949, opened 2024-05-21
Send own node-info earlier and move others to the end of want-config flow

### LINKED ISSUE #3740: [Feature Request]: Make NodeInfo Optional on WantConfig
### Platform

NRF52, ESP32

### Description

Right now want config also returns the whole nodedb, which is usually 100 when nodes are on MQTT, would be great for the apps and things like the Arduino library if we could have the want config just bring back the config and do the node info separately after we have got config data back.

thebentern:
@garthvh request.
Needs client testing.

Closes #3740 

Addendum:
Clients can send a special want_config_id of **69420** to omit the step of sending other nodeinfos, which is particularly useful for serial proto connected MCUs.


thebentern:
This has been tested on both Android and iOS clients by multiple testers

PR #3955, opened 2024-05-22
Update the Heltec board battery level read accuracy.

Heltec-Aaron-Lee:
Update the Heltec board battery voltage read parameter.


HarukiToreda:
> Update the Heltec board battery voltage read parameter.

Hey Aaron, you need to sign the CLA for this to be pushed.

Cisien:
I can confirm these values work as an override in the current firmware by setting the override to 5.1205 (4.9*1.045)

mkgin:
"*1.045" 5 bytes of bureaucracy...

garthvh:
Hi @Heltec-Aaron-Lee we really need access to the fork of meshtastic you are giving out to customers, needs to be available per the GPL3, it appears that the protobufs are out of sync and this is becoming a support issue.  We also need the CLA signed before this can get merged.

Heltec-Aaron-Lee:
Hi @garthvh sorry for the delay. It is already signed, I will pay attention to it in the future.

PR #3971, opened 2024-05-24
Fix MQTT queue use-after-free

mskvortsov:
There is an issue with the MQTT queue implementation. Enqueued ServiceEnvelope items contain fields pointing to objects that may be deallocated in another thread before the fields are accessed. This leads to the publishing of garbage.

One approach would be just encapsulating the fields (MeshPacket and both strings) into a new QueueItem structure by value. However, since ServiceEnvelope seems to be used only in the context of MQTT.cpp, changing field annotation at the protobuf level also seems reasonable.

Requires https://github.com/meshtastic/protobufs/pull/503.


thebentern:
Couldn't we check if the pointers are valid instead of converting these field to value via the protobuf annotations?

mskvortsov:
Reference counting for the MeshPacket objects in the base allocation pool could be an option. That’s a bit more intrusive change though.

mskvortsov:
I've experimented with adding a hidden refCount field to meshtastic_MeshPacket (hidden in the sense of existing only at C++ code level) and actually freeing up the packet from packetPool only if the counter becomes 1. Unfortunately, this approach done right requires some additional support from Nanopb: at the moment there is no good way to extend message A to A* with an auxiliary field added so that other message B encapsulating A will also use this extended A*.

So I would stick with the solution that was originally proposed.

PR #3974, opened 2024-05-26
Fix time updates from client device and potentially incorrect UI frame receiving 'toggle watch face' button tap

andrew-moroz:
This fixes a couple things:

* Time updates from a BLE client device. `RTC::perhapsSetRTC` ignores time qualities that are less than or equal to the current one, and the quality passed by `Position::trySetRtc` will be either `RTCQualityNet` or `RTCQualityNTP`. Consequently, updates from e.g. a connected phone were previously being ignored after the first one. This would be fine, except for the fact that the T-Watch appears to "pause" its RTC when shut down, such that the time it reads upon powering on is the same as when it was shut down. In that case, since the watch's time quality is still `RTCQualityNTP` when it powers back on, I found it could sometimes stay significantly out of sync with the phone for long periods of time.
* The 'toggle watch face' button tap handler expects the button to exist in UI frame 0, but previously this wasn't always the case. Should work as intended now, with the watch face always in frame 0 except in case of a critical fault.

### DIFF
--- a/src/gps/RTC.cpp
+++ b/src/gps/RTC.cpp
@@ -96,13 +96,17 @@ void readFromRTC()
  *
  * If we haven't yet set our RTC this boot, set it from a GPS derived time
  */
-bool perhapsSetRTC(RTCQuality q, const struct timeval *tv)
+bool perhapsSetRTC(RTCQuality q, const struct timeval *tv, bool forceUpdate)
 {
     static uint32_t lastSetMsec = 0;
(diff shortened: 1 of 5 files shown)


andrew-moroz:
I should add that, while this works well, I'm not certain it is much more than a bandaid... it's odd that the RTC would remain powered by its dedicated coin cell when the watch is off, and yet not mark time.

thebentern (approved):

PR #3975, opened 2024-05-26
Bump portduino to pick up improvements to reboot()

jp-bennett:
Two fixes here. One is to properly catch errors when calling execv() in the reboot() function, to prevent a crash. The other is to skip the --erase flag when calling a reboot, to prevent continual erase on reboot.


Talie5in (commented):
Compiled against arm64 (rPi3)

Log extract on reboot below.

```
INFO  | 05:28:07 22 Rebooting
DEBUG | 05:28:07 22 final reboot!
Portduino is starting, HWID=1, VFS root at /root/.portduino/default
Setting up Meshtastic on Portduino...
Using /etc/meshtasticd/config.yaml as config file
INFO  | ??:??:?? 0 

//\ E S H T /\ S T / C
```

thebentern (approved):

PR #3977, opened 2024-05-27
added AHTx0 sensor

fzellini:
Hi, I have added the AHT10 sensor to the project, so, with inexpensive modules containing bmp280&AHT20 you have pressure, temp & humidity.

aht10


platformio.ini
```diff
@@ -130,6 +130,7 @@ lib_deps =
   adafruit/Adafruit PM25 AQI Sensor@^1.0.6
   adafruit/Adafruit MPU6050@^2.2.4
   adafruit/Adafruit LIS3DH@^1.2.4
+  https://github.com/adafruit/Adafruit_AHTX0.git#2.0.5
```
caveman99, line 133: if possible use the Platformio registry instead of Github. that would be
`adafruit/Adafruit AHTX0@^2.0.5`

src/configuration.h
```diff
@@ -128,6 +128,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define LPS22HB_ADDR_ALT 0x5D
 #define SHT31_4x_ADDR 0x44
 #define PMSA0031_ADDR 0x12
+#define AHT10_ADDR 0x38
```
caveman99, line 131: NCP5623 uses this address already. Please add scanner code to detemine which device is actually connected

src/detect/ScanI2CTwoWire.cpp
```diff
@@ -256,7 +256,10 @@ void ScanI2CTwoWire::scanPort(I2CPort port)
                     type = BMP_280;
                 }
                 break;
-
+            case AHT10_ADDR:
+                LOG_INFO("AHT10 sensor found at address 0x%x\n", (uint8_t)addr.address);
+                type = AHT10;
+                break;
```
caveman99, line 262: See above comment, however this only applies if HAS_NCP5623 is defined.

src/mesh/generated/meshtastic/telemetry.pb.h
```diff
@@ -57,7 +57,9 @@ typedef enum _meshtastic_TelemetrySensorType {
     /* Lite On LTR-390UV-01 UV Light Sensor */
     meshtastic_TelemetrySensorType_LTR390UV = 21,
     /* AMS TSL25911FN RGB Light Sensor */
-    meshtastic_TelemetrySensorType_TSL25911FN = 22
+    meshtastic_TelemetrySensorType_TSL25911FN = 22,
+    /* AHT10/AHT20 Temperature and Humidity Sensor */
+    meshtastic_TelemetrySensorType_AHT10 = 23
```
caveman99, line 62: please revert all generated files as this would break merge commits. You need to change the protobuf repository instead.

src/modules/Telemetry/Sensor/AHT10.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+bool AHT10Sensor::getMetrics(meshtastic_Telemetry *measurement)
+{
+    LOG_DEBUG("AHT10Sensor::getMetrics\n");
+
+    sensors_event_t humidity, temp;
+    aht10.getEvent(&humidity, &temp);
+
+    measurement->variant.environment_metrics.temperature = temp.temperature;
+    measurement->variant.environment_metrics.relative_humidity = humidity.relative_humidity;
```
caveman99, line 32: Is this wise? you quote the sensor is used in addition to BME280. This will overwrite the temperature value. Maybe you can set this field only if the BME280 was NOT detected.

caveman99 (changes requested):

caveman99:
You can add the sensor definiton here: https://github.com/meshtastic/protobufs/blob/5f78a06c0fe5d0bd99c2fe206165212bdce89da0/meshtastic/telemetry.proto#L347

src/modules/Telemetry/Sensor/AHT10.cpp
fzellini: Both sensors provide the temperature, so taking it from one or the other is the same thing.

src/configuration.h
fzellini: NCP5623 and AHT10 does not have a simple method to check their identity, so if NCP5623 is defined (present in specific cards), no AHT10 detection will be performed.

src/detect/ScanI2CTwoWire.cpp
fzellini: see above

src/modules/Telemetry/Sensor/AHT10.cpp
caveman99: The problem is the measurements vary and which chi is used is not deteministic. I'd prefer the Bosch chip over the AHT

caveman99 (changes requested):

fzellini:
If BMP280+AHT20 detected, grab only humidity from AHT20...

DEBUG | ??:??:?? 12 [EnvironmentTelemetryModule] BMP280Sensor::getMetrics
INFO  | ??:??:?? 12 [EnvironmentTelemetryModule] AHTX0+BMP280 module detected: using temp from BMP280 and humy from AHTX0AHT10Sensor::getMetrics

fzellini:
> You can add the sensor definiton here: https://github.com/meshtastic/protobufs/blob/5f78a06c0fe5d0bd99c2fe206165212bdce89da0/meshtastic/telemetry.proto#L347

I have added sensor definition  in my cloned repository

fzellini:
> Please revert changes to the gitignore and gitmodules files.

Done

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "protobufs"]
 	path = protobufs
-	url = https://github.com/meshtastic/protobufs.git
+	url = https://github.com/fzellini/meshtastic-protobufs
```
alexander-smyslov, line 3: Please revert back

fzellini:
Done

Il giorno ven 31 mag 2024 alle ore 16:31 Alexander Smyslov <
***@***.***> ha scritto:

thebentern:
.gitignore changes are still present. Please revert it

fzellini:
Please check now.

Il giorno ven 31 mag 2024 alle ore 17:49 Ben Meadors <
***@***.***> ha scritto:

Enter-tainer:
hi, thank you for adding aht20 support!

> If BMP280+AHT20 detected, grab only humidity from AHT20

Could please explain the decision here? I found that aht20's temperature accuracy  is \pm 0.3, while bmp280 is \pm 1. So technically aht20 is more accurate?

fzellini:
Not my decision. If you read the all expanded thread ,   caveman99 says.. "I'd prefer the Bosch chip over the AHT".  So i have implemented in that way. But if ATH20 is more accurate, i agree with you

PR #3979, opened 2024-05-27
User experience improvement - app battery icon

neilhao:
For Station G2 or any board that does not have the BATTERY_PIN function defined, the current firmware can accurately determine that the board does not have a battery installed.

However, the current firmware will send batteryChargePercent = 0 to the APP for boards without battery, and the APP will display a battery empty icon. Users hope that boards without BATTERY_PIN will display a plug icon in the APP.

Ps. This is a PR about user experience and does not involve any bug fixes.

### DIFF
--- a/src/PowerStatus.h
+++ b/src/PowerStatus.h
@@ -59,9 +59,18 @@ class PowerStatus : public Status
     int getBatteryVoltageMv() const { return batteryVoltageMv; }
 
     /**
-     * Note: 0% battery means 'unknown/this board doesn't have a battery installed'
+     * Note: for boards with battery pin, 0% battery means 'unknown/this board doesn't have a battery installed'
      */
+#ifdef BATTERY_PIN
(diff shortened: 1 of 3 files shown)


thebentern (approved):

src/PowerStatus.h
```diff
@@ -59,9 +59,18 @@ class PowerStatus : public Status
     int getBatteryVoltageMv() const { return batteryVoltageMv; }
 
     /**
-     * Note: 0% battery means 'unknown/this board doesn't have a battery installed'
+     * Note: for boards with battery pin, 0% battery means 'unknown/this board doesn't have a battery installed'
      */
+#ifdef BATTERY_PIN
```
thebentern, line 64: I think we also need to check for HAS_PMU, otherwise T-Beams which do not have a battery pin but use the AXP chip would not report correctly

neilhao: I almost broke something :) Thanks for the reminder

thebentern (approved):

thebentern (approved):

PR #3980, opened 2024-05-27
Temporary: Move 6.7 framework to S3/C3 cores and move base esp32 (old stuff) back to 6.3.2 framework

thebentern:
Closes #3961

PR #3985, opened 2024-05-29
Fix original esp32 boot init panic

mskvortsov:


mskvortsov:
Still no luck. Now some builds fail because of iram text segment overflow - the very reason why libc was switched to nano in the first place.

thebentern:
> Still no luck. Now some builds fail because of iram text segment overflow - the very reason why libc was switched to nano in the first place.

I started taking a stab at splitting things out and moving the original ESP32 back to 6.3.2. https://github.com/meshtastic/firmware/pull/3980
Still needs some love

mskvortsov:
Arduino-esp32 framework uses a build of esp-idf sdk that turns on a set of compile-time optimization flags targeting speed instead of size, which leads to IRAM segment blowup*. As these flags can't be reverted after building, and sdk re-building is probably undesirable, IRAM size reduction becomes tricky.

SDK includes IRAM-placed auxiliary SPI flash chip drivers - a compromise but reasonable candidate for removal. Removing them frees up several hundred bytes just enough to make the firmware fit.

In the long term, something more systematic should be done about IRAM usage. Currently, I don't see any alternative to making a custom SDK build.

(*) Proper meshtastic code uses only a tiny portion of the segment, no chance for reduction there.

thebentern (approved):

PR #3999, opened 2024-05-31
Run tzset() and localtime() in getTZOffset() to ensure proper timezone offset

### LINKED ISSUE #3812: [Bug]: POSIX Timezone offset incorectly on T-Echo - but correct on a Haltec v3
### Category

Other

### Hardware

T-Echo

### Firmware Version

2.3.9.0afe2d4

### Description

Using the correct CST6CDT,M3.2.0,M11.1.0 time zone code on a T-Echo results in the time being off by 1hr on the device.
If I change it to a wrong code of CST5CDT,M3.2.0,M11.1.0 then it shows correctly on the device.
Using the correct time zone code CST6CDT,M3.2.0,M11.1.0 on a Haltec v3 shows the correct time.
All devices tested on 2.3.9.0afe2d4, I also tested via both GPS and phone provided time - both were affected.

### Relevant log output

_No response_

jp-bennett:
Fixes #3812 
My guess is that we run tzset before we have a valid date, so the DST element can't happen. This moves that to every time we attempt to calculate the offset, as well as moving to localtime() to make sure we get the proper timezone offset.

### DIFF
--- a/src/gps/RTC.cpp
+++ b/src/gps/RTC.cpp
@@ -222,14 +222,11 @@ bool perhapsSetRTC(RTCQuality q, struct tm &t)
  */
 int32_t getTZOffset()
 {
-    tzset();
-    time_t now;
+    time_t now = getTime(false);
     struct tm *gmt;
-    struct tm *local;
-    now = time(NULL);
     gmt = gmtime(&now);
-    local = localtime(&now);
-    return (int32_t)difftime(mktime(local), mktime(gmt));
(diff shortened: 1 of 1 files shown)


jp-bennett:
I misunderstood the problem the first time around (and had a local node broadcasting a bogus time). The solution is actually really simple: Just use the real time for our "now" variable, to calculate the offset.

thebentern (approved):

PR #4001, opened 2024-05-31
Don't send potentially bogus timestamps with fixed location

jp-bennett:
I'm seeing a situation where a "rogue" node thinks it has a valid timestamp, but doesn't. It's using fixed position, so the wrong timestamp is overwriting a time provided by another node that actually has a GPS.


wnagele:
@thebentern this will result in no time source availability in a mesh where no GPS nodes are present. I would suggest to still allow NTP otherwise this is a pretty drastic change.

thebentern:
Yes, I think NTP quality is probably a good compromise of allowing time source injections while avoiding poor quality times. Care to make a PR?

> @thebentern this will result in no time source availability in a mesh where no GPS nodes are present. I would suggest to still allow NTP otherwise this is a pretty drastic change.

wnagele:
Here you go: https://github.com/meshtastic/firmware/pull/4068

PR #4041, opened 2024-06-05
Triple-press not disabling GPS

todd-herbert:
Fixes an issue where triple-pressing the user button may not change the output at `PIN_GPS_EN`

Replaces a boolean, which didn't differentiate between "off" and "standby", with an enum.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -768,6 +768,15 @@ GPS::~GPS()
 void GPS::setGPSPower(bool on, bool standbyOnly, uint32_t sleepTime)
 {
     LOG_INFO("Setting GPS power=%d\n", on);
+
+    // Record the current state
+    if (on)
+        powerState = AWAKE;
+    else if (!on && standbyOnly)
+        powerState = STANDBY;
+    else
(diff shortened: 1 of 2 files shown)


src/gps/GPS.cpp
```diff
@@ -767,7 +774,19 @@ GPS::~GPS()
 
 void GPS::setGPSPower(bool on, bool standbyOnly, uint32_t sleepTime)
 {
-    LOG_INFO("Setting GPS power=%d\n", on);
+    // Record the current state
```
caveman99, line 777: please don't use lego comments :-) The syslog facility does not process them well. Use sprintf-style replacements instead.

caveman99 (changes requested):

src/gps/GPS.cpp
todd-herbert: Ah right, I wasn't aware, sorry. I've probably sprinkled a few more of those around the code-base that'll need tidying up, now that you mention it.

thebentern (approved):

caveman99 (approved):

PR #4043, opened 2024-06-05
Screen set frames update

### LINKED ISSUE #4038: [Bug]: Screen reverting back to last message after setFrame reload
### Category

Other

### Hardware

T-Echo

### Firmware Version

2.3.11

### Description

I wanted to make a small contribution in relation to Screen Frame reload issue ( may be not such a big one) , but every time setFrame run the page reverts back to Message or Module frame , but not the same frame that was viewed by user before, which could be annoying having to manually switch back to say Status frame.
Here is the change needed, please have a look and comment if needed. I am happy to submit PR if all OK. 
There are some caveats: After Resetting NodeDB or enabling / disabling a Module , the frame will not necessarily revert back to originally viewed frame. But after the device rebooted and there are more than 5 nodes in the list , it is quite stable.
Tested on T-Echo.
https://github.com/meshtastic/firmware/compare/master...meshi-bit:meshtastic-echo-echo:screen-frames

### Relevant log output

_No response_

slash-bit:
This PR fixes #4038 
every time setFrame run the page reverts back to Message or Module frame , but not the same frame that was viewed by user before, which could be annoying having to manually switch back to say Status frame.
So I have added lines where it would keep the same Frame number that user was one prior to reloading the frames. If that frame still available.
There may be some more enhancements needed in the future to accommodate when total number of frames changes ( like enabling/disabling modules or resetting Node DB.
For now, testing on T-Echo seems reasonably stable and stays on the same frame while device runs.

### DIFF
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -2165,26 +2165,7 @@ void Screen::setFrames(bool holdPosition)
 
         else 
             ui->switchToFrame(oldFrame); // Otherwise, go back to the same frame
-
-        }
-    else //in case a new incoming message is received, we want to show the message frame or if waypoint is received, show the waypoint frame
-        {
-        if (devicestate.has_rx_text_message) {
-            if (moduleFrames.size() > 0) 
-                ui->switchToFrame(moduleFrames.size() + 1);
-            
-            else 
-                ui->switchToFrame(0);
-            
-            }
-        if (devicestate.has_rx_waypoint) {
-            if (moduleFrames.size() > 0) 
-                ui->switchToFrame(moduleFrames.size() + 2);
-            
-            else 
-                ui->switchToFrame(1);
-            }
-        }
+}
     oldNumFrames = numframes; // Store how many frames we have, in case we want to "restore position" next time
 }
 
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -2151,7 +2151,7 @@ void Screen::setFrames()
     setFastFramerate(); // Draw ASAP
 
     if (currentFrameNum > numframes - 1) { // If we were on a frame that no longer exists
-        ui->switchToFrame(numframes - 2); // Attempt to return to last frame
+        ui->switchToFrame(0); // Return to Frame 0 , if we can't 
         } 
     else {
         ui->switchToFrame(currentFrameNum); // Attempt to return to same frame after rebuilding the frames
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -2056,7 +2056,8 @@ void Screen::setScreensaverFrames(FrameCallback einkScreensaver)
 // restore our regular frame list
 void Screen::setFrames()
 {
-    LOG_DEBUG("showing standard frames\n");
+    uint8_t currentFrameNum = ui->getUiState()->currentFrame;
+    LOG_DEBUG("Showing standard frame number %d\n", currentFrameNum);
     showingNormalScreen = true;
 
 #ifdef USE_EINK
@@ -2147,6 +2148,13 @@ void Screen::setFrames()
                     // just changed)
 
     setFastFramerate(); // Draw ASAP
+
+    if (currentFrameNum > numframes - 1) { // If we were on a frame that no longer exists
+        ui->switchToFrame(numframes - 2); // Attempt to return to last frame
+        } 
+    else {
+        ui->switchToFrame(currentFrameNum); // Attempt to return to same frame after rebuilding the frames
+        }    
 }
 
 void Screen::handleStartBluetoothPinScreen(uint32_t pin)
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -2054,11 +2054,12 @@ void Screen::setScreensaverFrames(FrameCallback einkScreensaver)
 #endif
 
 // restore our regular frame list
-void Screen::setFrames(bool holdPosition = false)
+void Screen::setFrames(bool holdPosition)
 {
-    uint8_t currentFrameNum = ui->getUiState()->currentFrame;
-    LOG_DEBUG("Showing standard frame number %d\n", currentFrameNum);
-    
+    // Which frame we are currently showing. We might want to return here after the new frames are set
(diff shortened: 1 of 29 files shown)


src/graphics/Screen.cpp
```diff
@@ -2143,6 +2145,13 @@ void Screen::setFrames()
                     // just changed)
 
     setFastFramerate(); // Draw ASAP
+
+    if (currentFrameNum > numframes - 1) { // If we were on a frame that no longer exists
+        ui->switchToFrame(numframes - 2); // Attempt to return to last frame
+        }
```
caveman99, line 2151: What if that one doesn't exist as well? i'd rather return to the first frame, there's always one frame.

caveman99 (changes requested):
Just something else to consider. We have pop-up screens especially in the Audio and Canned Messages module (PTT; Message Status, Delivery report and so on...), but also for functionality that is enabled with keypresses. These screens still need to take fcocus, but of course should return to the last screen shown and not the first one

todd-herbert:
> Just something else to consider. We have pop-up screens especially in the Audio and Canned Messages module (PTT; Message Status, Delivery report and so on...), but also for functionality that is enabled with keypresses. These screens still need to take fcocus, but of course should return to the last screen shown and not the first one

I'm not familiar enough with the modules to know exactly what needs to be accommodated, but I can certainly imagine the sort of attention to detail it'll take to catch all those little cases.

I hadn't considered this before, but now that you mention all this, I wonder if rather than trying to catch *every* case, it might be better to just target one:

If we changed the signature of `setFrames()` to`setFrames(bool holdPosition=false)`, we could apply this new behavior only to "node status updates", in [`Screen::handleStatusUpdate()`](https://github.com/meshtastic/firmware/blob/d0ca616c19028ae4609744d15f5fd340430265e5/src/graphics/Screen.cpp#L2643C5-L2648C15), avoiding status updates for text messages, input events, and UI events (from modules)?

src/graphics/Screen.cpp
todd-herbert: Ah I think I might have caused some of this confusion. I made a (probably unclear) suggestion that it would be good to store `numFrames` at the end of the method, so that next time `setFrames()` is called, we can detect whether we were previously on the log buffer / settings frames, and make sure we return to them, even if the number of frames has now changed.

slash-bit:
Hi guys, sorry attention drifted. I got your points and thanks @caveman99  and @todd-herbert  for attention to detail.
I think I got it . This code is not tested yet, will test it tonight. Will try to enable various modules and send some messages to ensure the focus is changing to the new message. Just wanted your verification of code in a meantime. Thanks
https://github.com/meshtastic/firmware/compare/master...slash-bit:meshtastic-echo-echo:screen-setFrames-update

todd-herbert:
> Hi guys, sorry attention drifted

Hey no problem, really excited to see some action back on this one!

slash-bit:
Great stuff, I think we are getting there! 
I have left one comment regarding assignment of oldNumFrames, please have a look.
Will try code on the T-Echo this evening.

slash-bit:
Hey all, I have done some tests on the latest branch.
The code works, but there might be more tweaks needed to polish it.
 -The frame holds after status update.
- increasing number of frames, ( because new nodes found ( when total number of nodes is below 5) still holds.
- Enabling additional module ( Telemetry for example), still holds. Obviously reboot knocks it to frame 0 , but after reboot and frame chosen. It stays.
What can be improved:
For some reason when deleting a node from DB manually, the frame reverts back to 0. Don't quite understand why , as it should be using the same Screen::handleStatusUpdate as with new node added. 
Also, when new incoming Message arrives (Screen::handleTextMessage) the screen jumps to frame 0. 
The frame 0 could be an additional Module ( like Telemetry) , therefore frame 0 isn't always new message.
I quickly switched to official 2.3.12 firmware and it has the same behavior, so that seems an existing issue.
Secondly, when New Way Point arrives it is suppose to switch to waypoint frame, but it jumps to frame 0.
We can probably fix those issues as well. If that's OK, I can have a go and try to work on setFrames() further , the send you new PR.

slash-bit:
Yeah, thanks. I suspected something wasn't right with devicestate.has_rx_text_message during limited testing that I have done . Will revert. Then other possible solution maybe to use handleTextMessage as a starting point to trigger switch to a message frame.
https://github.com/meshtastic/firmware/blob/39c9f92c6e7ee1639051a967d761ed5d642f4fcd/src/graphics/Screen.cpp#L2658
I am also not a user of the Waypoint or Telemetry in fact, but just reading commented likes in the code, I understand that's the intention to switch to those frames.
Yes, it ll be nice to clarify about canned message , so we may get that sorted as well. I crack on with looking for a solution, and better stay away from the heated debate in the discord :-)

todd-herbert:
(Whoops, accidentally pushed here by mistake, when I meant to sent it as a PR)

todd-herbert:
One more thing I spotted: I don't think the firmware code handles "deleting a waypoint" yet, at least not on the device.
There's no record in the firmware code of what changes are being made to the waypoint packet to indicate that it has been deleted. I imagine we can look in the app code though and see which fields it's using to trigger this.

slash-bit:
> he way I was handling it was over-thinking the whole thing. I don't have this code done tonight sorry, but if you want to have a go at doing the first part too, I was thinking maybe with two new Observables, then passing a "TargetFrame" enum to setFrames, instead of the bool.

:-) Aha , I was thinking the same , pass enum instead of boolean. But then I got distracted by .has_rx_text_message.
Yeah, let me try this. Yes, it will need couple of new variables defined, like txtFrame and wpFrame
they will derive from the amount of modules activated ( I believe modules frames always go first) , so every new module activated will shift  txtFrame up by one.

slash-bit:
> One more thing I spotted: I don't think the firmware code handles "deleting a waypoint" yet, at least not on the device. There's no record in the firmware code of what changes are being made to the waypoint packet to indicate that it has been deleted. I imagine we can look in the app code though and see which fields it's using to trigger this.

This can be interesting one to find you.

todd-herbert:
Actually, maybe it'd even work to record the value of numframes just before drawTextMessageFrame etc are called?

PR #4051, opened 2024-06-07
Hopefully fix T3S3 / EoRA-S3 usb boot issue

### LINKED ISSUE #4012: [Bug]: ESP32-S3 Boards with Native USB Not Booting without USB Host Serial
### Category

Hardware Compatibility, Serial

### Hardware

Other

### Firmware Version

2.3.10.d19607b

### Description

Connect CDEBYTE EoRa-S3 to 5V charger (not host computer) using USB-C to USB-A or USB-C to USB-C cable.

Device does not boot and OLED never displays any output. User LED does not light up or blink.

As this board uses the ESP32-S3's native USB pins (schematic is in the [EoRa-S3-900TB downloads page](https://www.cdebyte.com/products/EoRa-S3-900TB/4#Downloads) -> hardware data), I suspect the issue and fix would be related to a whole bunch of USB connectivity issues the T-Beam Supreme also had, due to that board also wiring the USB-C port to the native USB pins of its ESP32-S3: #3944 #3912 #3856 #3736 #3597.

### Relevant log output

_No response_

thebentern:
Closes #4012 
Please test

### DIFF
--- a/boards/CDEBYTE_EoRa-S3.json
+++ b/boards/CDEBYTE_EoRa-S3.json
@@ -5,12 +5,12 @@
     },
     "core": "esp32",
     "extra_flags": [
-      "-D CDEBYTE_EORA_S3",
-      "-D ARDUINO_USB_CDC_ON_BOOT=1",
-      "-D ARDUINO_USB_MODE=0",
-      "-D ARDUINO_RUNNING_CORE=1",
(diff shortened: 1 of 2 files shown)


thebentern:
This is not needed. Cannot reproduce problem on .11 alpha

PR #4052, opened 2024-06-07
Fix: two button presses sometimes required to wake screen

todd-herbert:
On some devices, after a long period of inactivity, the first press of the user button does not reliably wake the screen. A second button press is often required. This PR hopefully corrects (or at least reduces) this issue. 

Tested working on T-Beam v1.2


thebentern (approved):

madeofstown:
I've also experienced this bug on Heltec Tracker v1.1 and Heltec V3. Hopefully it fixes those too.

todd-herbert:
I'm less and less confident with this one, but hopefully!

Issue #2885, opened 2023-10-12
[Bug]: Device becomes unresponsive or powers off if Bluetooth is disabled


cordelster:
### Category

BLE

### Hardware

Rak4631

### Firmware Version

2.1.23.04bbdc6

### Description

Tested across multiple RAK devices to expedite and confirm.
Desired effect is to have a Repeater or Router with USB Serial and Bluetooth disabled.
All devices set to US region, each given two PSK channels with one being "admin" slot 2:
In each case, setting Bluetooth disabled seems to create a condition where the device becomes unresponsive to admin commands first, later followed by becoming completely unresponsive and later appears to be powered off (no LED activity). The time to become unresponsive can take up to 72 hours.
Tested with/without power (battery only).

Equipment:
RAK4631
RAK19007
RAK18001 - PWM Buzzer module (on some devices which don't appear to have any bearing on issue/full disclosure in case it becomes relevant)

Tested each of the following modes with Bluetooth disabled, eventually becomes unresponsive regardless of USB serial setting:
Client/router - because issues in router were previously known
Router
Repeater

All devices where Bluetooth is enabled have not exhibited issues. Snip from serial debug shows only Power status when and after the device faults. No other clues are obvious.

### Relevant log output

```Shell
DEBUG | ??:??:?? 8781 [Router] Incoming message was filtered 0x5185b8ba
DEBUG | ??:??:?? 8788 [Power] Battery: usbPower=0, isCharging=0, batMv=4186, batPct=98
DEBUG | ??:??:?? 8808 [Power] Battery: usbPower=0, isCharging=0, batMv=4172, batPct=96
DEBUG | ??:??:?? 8828 [Power] Battery: usbPower=0, isCharging=0, batMv=4188, batPct=98
DEBUG | ??:??:?? 8848 [Power] Battery: usbPower=0, isCharging=0, batMv=4190, batPct=98
DEBUG | ??:??:?? 8868 [Power] Battery: usbPower=0, isCharging=0, batMv=4162, batPct=95
DEBUG | ??:??:?? 8886 [RadioIf] (bw=250, sf=9, cr=4/8) packet symLen=2 ms, payloadSize=39, time 205 ms
DEBUG | ??:??:?? 8886 [RadioIf] Lora RX (id=0x4b25c3d2 fr=0xfd to=0xff, WantAck=0, HopLim=5 Ch=0xb9 encrypted rxSNR=10.5 rxRSSI=-73)
[...]

PR #4055, opened 2024-06-09
Workaround to disable bluetooth on NRF52

todd-herbert:
Not a fix, but a workaround for #2885. 
Implements a suggestion discussed on the Meshtastic Discord server:

If `config.bluetooth.enabled` is false, Bluefruit is setup as normal, then advertising is stopped, and the tx power is reduced to minimum. This means that the SoftDevice, and the brown-out-detection, are enabled as usual.

With this workaround, and Bluetooth disabled:
* the device does not show in scans
* it is not possible to connect to the device using Bluetooth

This workaround seems to avoid the issue of #2885, and RadioLib -706 errors (discussed in #NRF52 on Discord). I have not experienced any lock-ups or crashes during testing, even with uptime > 3 days. It would be good to get more test time on this before rolling out.

I don't have equipment to measure the power consumption over time, but crude current measurement with a digital multimeter suggests that the workaround is similar to running with Bluetooth completely uninitialized. The workaround is *at most* a few perfect more efficient than running normally, with Bluetooth at idle.

### DIFF
--- a/src/platform/nrf52/NRF52Bluetooth.cpp
+++ b/src/platform/nrf52/NRF52Bluetooth.cpp
@@ -215,6 +215,19 @@ void NRF52Bluetooth::shutdown()
     Bluefruit.Advertising.stop();
 }
 
+void NRF52Bluetooth::startDisabled()
+{
+    LOG_DEBUG("Initializing NRF52 Bluetooth, then disabling. (workaround)\n");
+
+    // Setup Bluetooth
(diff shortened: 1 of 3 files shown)


caveman99 (approved):

thebentern:
Looks good. 
I want to hold this one till the release after this upcoming alpha though, just in case we encounter lockups on stop advertising.

mike6789:
I will test this soon with power monitoring equipment, hope it fixes my issue

PR #4071, opened 2024-06-10
support RAK10701 touchscreen

DanielCao0:
Support touchscreen for the Board RAK10701 , which uses the RAK14014 to implement the touchscreen function .  The touch chip model is FT6336U.


src/input/TouchScreenImpl1.cpp
```diff
@@ -55,11 +55,11 @@ void TouchScreenImpl1::onEvent(const TouchEvent &event)
 
     switch (event.touchEvent) {
     case TOUCH_ACTION_LEFT: {
-        e.inputEvent = static_cast<char>(meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar_RIGHT);
+        e.inputEvent = static_cast<char>(meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar_LEFT);
```
caveman99, line 58: This may upset other implementations. can i have @mverch67 opinion on this?

caveman99 (changes requested):

src/input/TouchScreenImpl1.cpp
DanielCao0: It may affect other platforms, but this seems to be a bug when tested. TOUCH_ACTION_LEFT should swipe from right to left.

mverch67: There is a reason for the original code. If you swipe from right to left the TOUCH_ACTION_LEFT is generated. If you swipe a screen from right to left then the screen moves to the left, disappears and the screen on the right reveals. So the trigger InputEventChar_RIGHT as in the original code is correct.

DanielCao0: Thank you. I see what you mean.

caveman99:
@DanielCao0 no need to close the pull request, you can simply keep pushing commits to your branch and they will appear here...

DanielCao0:
I'm still modifying the code, I'll pull again, I'm adding the configuration of the virtual keyboard

caveman99:
@DanielCao0 i'll mark the Pull Request as a 'draft' and we can keep it open. When you are satisfied, mark as 'review' :-)

PR #4086, opened 2024-06-12
Add Heltec Capsule Sensor V3 to the Meshtastic source code

Heltec-Aaron-Lee:
This commit includes the following functions for Capsule Sensor V3:
- Add Heltec Capsule Sensor V3 hardware definition,
- Button status read.
- Battery level read,
- Charging status detection.

The following functions are tested before submission:
- The whole project compile, build, and firmware update,
- Bluetooth connect to an Android/IOS phone,
- GPS read;
- LoRa transmission and reception,
- Mesh network communication.

Going to do in the next PR:
- Various sensor drives and running on Capsule Sensor V3;
- Sensor data read, monitor, and transmit.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -33,6 +33,7 @@ default_envs = tbeam
 ;default_envs = wio-e5
 ;default_envs = radiomaster_900_bandit_nano
 ;default_envs = radiomaster_900_bandit_micro
+;default_envs = heltec_capsule_sensor_v3
 
 extra_configs =
   arch/*/*.ini
--- a/src/ButtonThread.cpp
+++ b/src/ButtonThread.cpp
@@ -41,7 +41,11 @@ ButtonThread::ButtonThread() : OSThread("Button")
     }
 #elif defined(BUTTON_PIN)
     int pin = config.device.button_gpio ? config.device.button_gpio : BUTTON_PIN; // Resolved button pin
+#if defined(HELTEC_CAPSULE_SENSOR_V3)
+    this->userButton = OneButton(pin, false, false);
+#else
     this->userButton = OneButton(pin, true, true);
+#endif
     LOG_DEBUG("Using GPIO%02d for button\n", pin);
 #endif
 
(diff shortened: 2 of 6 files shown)


caveman99 (approved):

nolith:
Hello @Heltec-Aaron-Lee 

I was playing with an heltec sensor capsule v3 with GPS.

I was hoping to run it as a TRACKER mode where the device sends the position and then goes into a sleep mode until the next event.

From the battery runtime I have the impression that the device isn't sleeping. Is this a limitation of the firmware or the hardware?

For my project it will be ok even to use a deep sleep state as long as the device can get back online, get a fix, and send a position on a given interval.

thebentern:
> Hello @Heltec-Aaron-Lee 
> 
> 
> 
> I was playing with an heltec sensor capsule v3 with GPS.
> 
> 
> 
> I was hoping to run it as a TRACKER mode where the device sends the position and then goes into a sleep mode until the next event.
> 
> 
> 
> From the battery runtime I have the impression that the device isn't sleeping. Is this a limitation of the firmware or the hardware?
> 
> 
> 
> For my project it will be ok even to use a deep sleep state as long as the device can get back online, get a fix, and send a position on a given interval.

Device role of Tracker and power.is_power_saving true will give you the behavior you want.

nolith:
Thanks @thebentern 

In the iOS UI that setting is not available for the heltec sensor capsule, but I was able to set it using the python CLI and the admin channel.

I'll report back on battery usage.

If I understood correctly, now my only relevant setting is the positon broadcast interval, that one will set my sleep schedule, is that correct?

fifieldt:
@andres-a-git , should work fine!

andres-a-git:
> @andres-a-git , should work fine!

I was already trying to flash it over wifi serial, but did not finalize the flashing. I assumed that the serial over wifi could be also used, but discovered that the Arduino IDE may not help. Luckily stumbled to the FW page:
https://github.com/meshtastic/meshtastic.github.io/tree/master/firmware-2.5.9.936260f

From here I found the special capsule FW and downloaded the update version.
image



The update .bin wth long name was not accepted by the Heltec WiFi flasher page, so I renamed it to firmware.bin and uploaded it to the capsule over Wifi. Now the capsule works as before, but with new FW :)

Would it be possible to add the option for Capsule to download the FW bin already from the flasher.meshtastic.org page like it is done with UF2 files on RAK devices?

Issue #4087, opened 2024-06-12
[Feature Request]: Include hop_start and hop_limit in JSON message for MQTT uplink


ab0oo:
### Platform

NRF52, ESP32, RP2040, Linux Native, other

### Description

I would like to add hop_start and hop_limit numbers to the JSON formatted message in MQTT uplink, in addition to having hop_count.  This will allow network managers to quickly identify stations using absurdly large hop limits and help guide them to a properly tuned network.

PR #4088, opened 2024-06-12
feature-mqtt: add hop_start and hop_limit to MQTT uplink

ab0oo:
This PR adds "hop_start" and "hop_limit" to the MQTT JSON uplink message. 
This is useful in network management, helping to identify stations with larger-than-recommended hop-limits and lets network managers guide them to better choices.

References:  https://github.com/meshtastic/firmware/issues/4087

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -903,7 +903,6 @@ std::string MQTT::meshPacketToJson(meshtastic_MeshPacket *mp)
     if (mp->hop_start != 0 && mp->hop_limit <= mp->hop_start)
         jsonObj["hops_away"] = new JSONValue((unsigned int)(mp->hop_start - mp->hop_limit));
     jsonObj["hop_start"] = new JSONValue((unsigned int)(mp->hop_start));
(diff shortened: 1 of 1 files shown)


src/mqtt/MQTT.cpp
```diff
@@ -902,7 +902,9 @@ std::string MQTT::meshPacketToJson(meshtastic_MeshPacket *mp)
         jsonObj["snr"] = new JSONValue((float)mp->rx_snr);
     if (mp->hop_start != 0 && mp->hop_limit <= mp->hop_start)
         jsonObj["hops_away"] = new JSONValue((unsigned int)(mp->hop_start - mp->hop_limit));
-
+    jsonObj["hop_start"] = new JSONValue((unsigned int)(mp->hop_start));
```
GUVWAF, line 905: I would only add `hop_start` when it's not 0 because that usually means a node on the path had too old firmware. 

Furthermore, we don't need `hop_limit` because it can be derived from `hop_start` and `hops_away`.

GUVWAF (changes requested):

GUVWAF (approved):

PR #4109, opened 2024-06-14
Update Screen.cpp - Sleeping screen with owner.short_name

Dorn8010:
Added owner.short_name to "Sleeping" screen message. Didn't check variable availablility - not C++ experienced. :/
Pls recheck.

### DIFF
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -305,6 +305,8 @@ static bool haveGlyphs(const char *str)
 /// Used on eink displays while in deep sleep
 static void drawDeepSleepScreen(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t x, int16_t y)
 {
+    const char *idText = owner.short_name;
+
     // Next frame should use full-refresh, and block while running, else device will sleep before async callback
     EINK_ADD_FRAMEFLAG(display, COSMETIC);
     EINK_ADD_FRAMEFLAG(display, BLOCKING);
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -310,7 +310,7 @@ static void drawDeepSleepScreen(OLEDDisplay *display, OLEDDisplayUiState *state,
     EINK_ADD_FRAMEFLAG(display, BLOCKING);
 
     LOG_DEBUG("Drawing deep sleep screen\n");
-    drawIconScreen("Sleeping...", display, state, x, y);
+    drawIconScreen( (owner.short_name.empty() ? "Sleeping." : owner.short_name + " sleeping."), display, state, x, y);
 }
 
 /// Used on eink displays when screen updates are paused
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -312,7 +312,7 @@ static void drawDeepSleepScreen(OLEDDisplay *display, OLEDDisplayUiState *state,
     EINK_ADD_FRAMEFLAG(display, BLOCKING);
 
     LOG_DEBUG("Drawing deep sleep screen\n");
-    drawIconScreen( (owner.short_name.empty() ? "Sleeping." : owner.short_name + " sleeping."), display, state, x, y);
+    drawIconScreen( (idText.empty() ? "Sleeping." : idText + " sleeping."), display, state, x, y);
 }
 
 /// Used on eink displays when screen updates are paused
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -314,12 +314,12 @@ static void drawDeepSleepScreen(OLEDDisplay *display, OLEDDisplayUiState *state,
     LOG_DEBUG("Drawing deep sleep screen\n");
     
     // Check if idText is not assigned OR empty OR has non printable characters
-    if (!idText || strlen(idText) == 0 || !(haveGlyphs(idText)) ) {
+    if (!idText || strlen(idText) == 0 || !haveGlyphs(idText)) {
         drawIconScreen("Sleeping", display, state, x, y);
     } else {
-        // Create a new string that is the concatenation of idText and " sleeping."
+        // Create a new string that is the concatenation of idText and " sleeping"
         char displayStr[30]; // Declare a character array for the display string 5+buffer for changes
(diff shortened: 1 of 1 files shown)


todd-herbert (approved):
Looks good on my devices 👍

src/graphics/Screen.cpp
```diff
@@ -281,12 +281,27 @@ static void drawFunctionOverlay(OLEDDisplay *display, OLEDDisplayUiState *state)
...
-    drawIconScreen("Sleeping...", display, state, x, y);
+    
+    // Check if idText is empty
+    if (strlen(idText) == 0) {
+        drawIconScreen("Sleeping.", display, state, x, y);
+    } else {
+        // Create a new string that is the concatenation of idText and " sleeping."
+        size_t idTextLen = strlen(idText);
+        char displayStr[idTextLen + 12]; // Allocate enough space for the new string
+        strcpy(displayStr, idText); // Copy idText to the new string
```
thebentern, line 299: Perhaps more consolidated to write in this way with a single expression?
sprintf(displayStr, "%s sleeping.",  owner.short_name);

thebentern: Also, no need to declare idText this way. You can just reference owner.short_name

Dorn8010: Absolutely guys. I'm not a C++ programmer and just tried that the code works. I also don't like the strlen check on a maybe unassigned var. So if you guys know it better - pls change.

Dorn8010: Thought about something like 
if (!idText || strlen(idText) == 0) {
for the check of idText ?

todd-herbert: Actually, is it possible to get into a situation where short name is has zero length? It might not be an issue? Now I think about it though, I'm not sure what would happen with emoji short names. I should check if this is a problem with the existing E-Ink screensaver code too.

todd-herbert:
Just an update: the new `haveGlyphs` function will help you catch any unprintable characters.
Something like:
```cpp
if (haveGlyphs(owner.short_name)) {
    // use the short name
}
else {
    // don't use the short name
}
```
___

A bit of house keeping: can we also [move the haveGlyphs definition](https://github.com/meshtastic/firmware/pull/4115#issuecomment-2170947765) inside `#ifdef USE_EINK` eventually? Just to get rid of compiler warnings that it's not used.

src/graphics/Screen.cpp
Dorn8010, line 1: idText used for common style with the following drawScreensaverOverlay.
snprintf for mem safety and 30 chars gives enough space for later changes of text

Dorn8010:
> Just an update: the new `haveGlyphs` function will help you catch any unprintable characters. Something like:
> 
> ```c++
> if (haveGlyphs(owner.short_name)) {
>     // use the short name
> }
> else {
>     // don't use the short name
> }
> ```
> 
> A bit of house keeping: can we also [move the haveGlyphs definition](https://github.com/meshtastic/firmware/pull/4115#issuecomment-2170947765) inside `#ifdef USE_EINK` eventually? Just to get rid of compiler warnings that it's not used.

Waited with this. If OK for you guys, I would add the short_name and HEX ID at the boot screen e.g. EU_868:ad24:!A1B2C3 - so every device can immediately be identified at startup and the unique ID is visible - and then we need this haveGlyphs also when using the other devices.
For logical naming - I would like to change the name of the functinon into haveNoGlyphs - since a true answer would mean exactly this.

src/graphics/Screen.cpp
```diff
@@ -305,12 +305,25 @@ static bool haveGlyphs(const char *str)
 /// Used on eink displays while in deep sleep
 static void drawDeepSleepScreen(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t x, int16_t y)
 {
+    const char *idText = owner.short_name;
+
```
todd-herbert, line 309: Don't worry too much about creating `idText` for common style. The only reason I used it in `drawScreensaverOverlay` was because it helped me read the code, so only use it here if it helps you.

src/graphics/Screen.cpp
```diff
@@ -305,12 +305,25 @@ static bool haveGlyphs(const char *str)
...
     // Next frame should use full-refresh, and block while running, else device will sleep before async callback
     EINK_ADD_FRAMEFLAG(display, COSMETIC);
     EINK_ADD_FRAMEFLAG(display, BLOCKING);
 
     LOG_DEBUG("Drawing deep sleep screen\n");
-    drawIconScreen("Sleeping...", display, state, x, y);
+    
+    // Check if idText is not assigned OR empty OR has non printable characters
+    if (!idText || strlen(idText) == 0 || !haveGlyphs(idText)) {
```
todd-herbert, line 317: I think `!idText` and `strlen(idText)==0` are technically checking the same thing (gotta love C++). I think maybe it's not so important to check this though: `owner.short_name` is assigned when the node is initially set up, and if something goes wrong, we probably won't catch with `strlen(idText) == 0` anyway.

todd-herbert: > For logical naming - I would like to change the name of the functinon into haveNoGlyphs - since a true answer would mean exactly this.

I wouldn't change it to `haveNoGlyhps`. The reason it was called `haveGlyphs` is because it checks if our font has all the "glyphs" (characters) needed to print the string. It returns false if we're missing a glyph which we need.

The fact that I had to explain this though suggests that the name probably isn't clear enough! Maybe `fontHasChars(idText)` or something similar?

todd-herbert (commented):
> If OK for you guys, I would add the short_name and HEX ID at the boot screen e.g. EU_868:ad24:!A1B2C3

No strong opinion on this one!

> and then we need this haveGlyphs also when using the other devices.

It's possible, although I'm not sure about the idea. We're hiding the missing characters here for cosmetic purposes, but in some other situations, it might be important that we continue to identify them with `¿`.

src/graphics/Screen.cpp
Dorn8010: OK. I think I got the use from reading the code - just thought the name would be more consistent as your fontHasChars. I leave it for now.

todd-herbert:
I still think this is a great idea. Keen to see it included!

PR #4115, opened 2024-06-15
E-Ink screensaver validates short name

todd-herbert:
Adds a function to Screen.cpp which checks if the OLED library is able to render all characters in a string.
Using this function, the E-Ink screensaver may decide not to draw the device's short name.

This new `haveGlyphs` function may be relevant for https://github.com/meshtastic/firmware/pull/4109#discussion_r1640919772

With short name `wp1`
normal E-ink screensaver

With short name `Pkī`, or `📜`
noGlyphs

### DIFF
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -277,6 +277,30 @@ static void drawFunctionOverlay(OLEDDisplay *display, OLEDDisplayUiState *state)
     }
 }
 
+/// Check if the display can render a string (detect special chars; emoji)
+static bool haveGlyphs(const char *str)
+{
+#if defined(OLED_UA) || defined(OLED_RU)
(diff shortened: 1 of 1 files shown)


thebentern (approved):

caveman99:
@todd-herbert this is causing a warning on non eink display compiles, casue the function is not used anywhere else. Wanna pull it into the e-ink conditional or are you planning to use it in other places as well?

todd-herbert:
> @todd-herbert this is causing a warning on non eink display compiles, casue the function is not used anywhere else. Wanna pull it into the e-ink conditional or are you planning to use it in other places as well?

No plans elsewhere, sorry I didn't get to this in time. I'll make sure it gets moved inside that conditional.

PR #4116, opened 2024-06-16
Show compass on waypoint frame; clear when waypoint deleted

todd-herbert:
Re-draws screen frames when a new waypoint is received.
Removes the waypoint frame from the screen when waypoint is deleted.

Relevant for https://github.com/meshtastic/firmware/pull/4043#issuecomment-2168427308


caveman99 (approved):

thebentern (approved):

thebentern:
Looks good.
Out of scope, but I want to circle back to the waypoint screen and actually provide the heading circle like we do for node positions. It'll be really useful with the new digital compass support.

slash-bit (approved):
Looks good , guys

todd-herbert:
@thebentern 
> I want to circle back to the waypoint screen and actually provide the heading circle like we do for node positions. It'll be really useful with the new digital compass support.

I'm not 100% sure how all the new digital compass stuff works, but the waypoint frame is now showing the compass in basically the same way as the NodeInfo frames, so fingers crossed.

jp-bennett:
> @thebentern 
> > I want to circle back to the waypoint screen and actually provide the heading circle like we do for node positions. It'll be really useful with the new digital compass support.
> 
> I'm not 100% sure how all the new digital compass stuff works, but the waypoint frame is now showing the compass in basically the same way as the NodeInfo frames, so fingers crossed.

I'll try to give it a test run tomorrow.

jp-bennett:
Seems to work as expected, with the display pointing the way to the waypoint. One note, adding a waypoint from Android does not add that waypoint on the local radio itself.

src/graphics/Screen.cpp
```diff
@@ -2133,9 +2240,12 @@ void Screen::setFrames()
     if (devicestate.has_rx_text_message && shouldDrawMessage(&devicestate.rx_text_message)) {
         normalFrames[numframes++] = drawTextMessageFrame;
     }
-    // If we have a waypoint - show it next, unless it's a phone message and we aren't using any special modules
+
+    // If we have a waypoint (not expired, not deleted)
     if (devicestate.has_rx_waypoint && shouldDrawMessage(&devicestate.rx_waypoint)) {
```
jp-bennett, line 2245: IMO, drop the shouldDrawMessage() check here. Navigating to a self-dropped waypoint seems like a good thing.

src/graphics/Screen.cpp
```diff
@@ -1596,6 +1595,112 @@ static void drawNodeInfo(OLEDDisplay *display, OLEDDisplayUiState *state, int16_
     drawColumns(display, x, y, fields);
 }
 
+/// Draw the last waypoint we received
+static void drawWaypointFrame(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t x, int16_t y)
```
jp-bennett, line 1599: What would it take to move this into WaypointModule.drawFrame()?

src/graphics/Screen.cpp
todd-herbert: Makes sense to me!

src/graphics/Screen.cpp
todd-herbert: It'd be nice to have it contained like that. Honestly, I just modified the existing `drawWaypointFrame` method right where it was without giving it too much thought. 

> What would it take

A bunch of graphics stuff seems to lean on static functions inside of Screen.cpp, but outside the screen class. I don't think any of the frames drawn directly by modules can make use of these functions right now. Exposing those tidily to the mesh module classes would probably be the main thing to deal with.

jp-bennett: This is going to take a bit more untangling. I don't think the PR should wait on it.

todd-herbert:
Short-term solution. Future plan is to instead make waypoint module itself responsible for drawing the frame.

jp-bennett (approved):

PR #4117, opened 2024-06-16
Add RAKwireless WisMesh Hub (RAK2560/RAK9154)

beegee-tokyo:
RAKwireless is planning to offer the [RAK2560 SensorHub](https://store.rakwireless.com/products/sensor-hub) and [RAK9154 Solar Battery Lite](https://store.rakwireless.com/products/rak-battery-lite-solar-power-solution-rak9154) preflashed with Meshtastic Firmware as a Meshtastic Router or Repeater under the name WisMesh Hub.

This pull request adds support for the RAK2560 device and the communication protocol for the RAK9154 Solar Battery Lite to Meshtastic.

image

The RAK2560 device is based on the RAK4630 (used on RAK4631) stamp module and is a low power device with LoRa and BLE connectivity. 
- LoRa 868-930 MHz support
- Nordic nRF52840 MCU with Semtech SX1262 LoRa transceiver
- 12 VDC over power supply or solar panel
- Embedded antenna (external antenna optional)
   - High efficiency (over 75%) LoRa Band 860~930 MHz
   - Support 700\~960 MHz and 1700\~2170 MHz.
- IP66-rated waterproof enclosure

The RAK9154 is a solar panel/battery combination with 
- Battery capacity: 5200 mAh
- Operating temperature: -20～60° C
- Dimensions: 180 x 130 x 60 mm
- Outputs: one 12 VDC output for Sensor Hub
- Integrated BMS and heater
- Supported mounting pole diameter: 55～80 mm (Other mounting options possible)

WisMesh Hub in Meshtastic mobile app (temperature/humidity is optional and requires RAk1901 module inside the RAK2560):
image

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,8 +2,9 @@
     // See http://go.microsoft.com/fwlink/?LinkId=827846
     // for the documentation about the extensions.json format
     "recommendations": [
-        "ms-vscode.cpptools",
-        "platformio.platformio-ide",
-        "trunk.io"
+        "platformio.platformio-ide"
     ],
+    "unwantedRecommendations": [
+        "ms-vscode.cpptools-extension-pack"
+    ]
 }
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,9 +2,8 @@
     // See http://go.microsoft.com/fwlink/?LinkId=827846
     // for the documentation about the extensions.json format
     "recommendations": [
-        "platformio.platformio-ide"
+		"ms-vscode.cpptools",
+        "platformio.platformio-ide",
+        "trunk.io"
     ],
-    "unwantedRecommendations": [
-        "ms-vscode.cpptools-extension-pack"
-    ]
 }
--- a/src/modules/Telemetry/Sensor/RAK9154Sensor.cpp
+++ b/src/modules/Telemetry/Sensor/RAK9154Sensor.cpp
@@ -1,3 +1,4 @@
+#ifdef HAS_RAKPROT 
 #include "RAK9154Sensor.h"
 #include "../mesh/generated/meshtastic/telemetry.pb.h"
 #include "TelemetrySensor.h"
@@ -187,3 +188,4 @@ bool RAK9154Sensor::isCharging()
 {
 	return (dc_cur > 0) ? true : false;
 }
+#endif // HAS_RAKPROT
--- a/src/modules/Telemetry/Sensor/RAK9154Sensor.cpp
+++ b/src/modules/Telemetry/Sensor/RAK9154Sensor.cpp
@@ -1,0 +1,189 @@
+#include "RAK9154Sensor.h"
+#include "../mesh/generated/meshtastic/telemetry.pb.h"
+#include "TelemetrySensor.h"
+#include "configuration.h"
+
+#include <RAK-OneWireSerial.h>
+#include "concurrency/Periodic.h"
+
+using namespace concurrency;
+
+#define BOOT_DATA_REQ
+
+RAK9154Sensor::RAK9154Sensor() : TelemetrySensor(meshtastic_TelemetrySensorType_SENSOR_UNSET, "RAK1954") {}
+
+static Periodic *onewirePeriodic;
+
+static SoftwareHalfSerial mySerial(HALF_UART_PIN); // Wire pin  P0.15
+
+static uint8_t buff[0x100];
+static uint16_t bufflen = 0;
+
+static int16_t dc_cur = 0;
+static uint16_t dc_vol = 0;
+static uint8_t dc_prec = 0;
+static uint8_t provision = 0;
+
+static void onewire_evt(const uint8_t pid, const uint8_t sid, const SNHUBAPI_EVT_E eid, uint8_t *msg, uint16_t len)
+{
+	switch (eid)
+	{
+	case SNHUBAPI_EVT_RECV_REQ:
+	case SNHUBAPI_EVT_RECV_RSP:
+		break;
+
+	case SNHUBAPI_EVT_QSEND:
+		mySerial.write(msg, len);
+		break;
+
+	case SNHUBAPI_EVT_ADD_SID:
+		// LOG_INFO("+ADD:SID:[%02x]\r\n", msg[0]);
+		break;
+
+	case SNHUBAPI_EVT_ADD_PID:
+		// LOG_INFO("+ADD:PID:[%02x]\r\n", msg[0]);
+#ifdef BOOT_DATA_REQ
+		provision = msg[0];
+#endif
+		break;
+
+	case SNHUBAPI_EVT_GET_INTV:
+		break;
+
+	case SNHUBAPI_EVT_GET_ENABLE:
+		break;
+
+	case SNHUBAPI_EVT_SDATA_REQ:
+
+		// LOG_INFO("+EVT:PID[%02x],IPSO[%02x]\r\n",pid,msg[0]);
+		// for( uint16_t i=1; i<len; i++)
+		// {
+		//     LOG_INFO("%02x,", msg[i]);
+		// }
+		// LOG_INFO("\r\n");
+		switch (msg[0])
+		{
+		case RAK_IPSO_CAPACITY:
+			dc_prec = msg[1];
+			if (dc_prec > 100)
+			{
+				dc_prec = 100;
+			}
+			break;
+		case RAK_IPSO_DC_CURRENT:
+			dc_cur = (msg[2] << 8) + msg[1];
+			break;
+		case RAK_IPSO_DC_VOLTAGE:
+			dc_vol = (msg[2] << 8) + msg[1];
+			dc_vol *= 10;
+			break;
+		default:
+			break;
+		}
+
+		break;
+	case SNHUBAPI_EVT_REPORT:
+
+		// LOG_INFO("+EVT:PID[%02x],IPSO[%02x]\r\n",pid,msg[0]);
+		// for( uint16_t i=1; i<len; i++)
+		// {
+		//     LOG_INFO("%02x,", msg[i]);
+		// }
+		// LOG_INFO("\r\n");
+
+		switch (msg[0])
+		{
+		case RAK_IPSO_CAPACITY:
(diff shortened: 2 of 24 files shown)


beegee-tokyo:
Test installation (with internal antenna)
image

.vscode/extensions.json
```diff
@@ -2,8 +2,9 @@
...
     "recommendations": [
-        "ms-vscode.cpptools",
-        "platformio.platformio-ide",
-        "trunk.io"
+        "platformio.platformio-ide"
     ],
+    "unwantedRecommendations": [
+        "ms-vscode.cpptools-extension-pack"
+    ]
 }
```
caveman99, line 10: Please don't check in the vscode files. This will cause merge conflicts all over the place.

src/modules/Telemetry/Sensor/RAK9154Sensor.cpp
```diff
@@ -0,0 +1,191 @@
+#ifdef HAS_RAKPROT
```
caveman99, line 1: I don't think the RAK9154 support is places correctly. The Telemetry folder is for devices that actually transmit values over Lora. This looks more like gathering values for internal processing. If this is only ever used once for this particular device, you can as well put it in the variant folder and skip deriving it from TelemetryModule. If you add the OSThread class to inheritance you can use the runOnce model without having to define it as a sensor (which it isn't in the meshtastic sense)

src/modules/Telemetry/Sensor/RAK9154Sensor.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+void RAK9154Sensor::setup()
+{
+	// Set up oversampling and filter initialization
+}
+
+bool RAK9154Sensor::getMetrics(meshtastic_Telemetry *measurement)
+{
+	return true;
```
caveman99, line 174: Alternatively if the values really should be able to go out over the mesh, you need to implementthese functions :-)

variants/rak4631/variant.h
```diff
@@ -228,9 +228,18 @@ SO GPIO 39/TXEN MAY NOT BE DEFINED FOR SUCCESSFUL OPERATION OF THE SX1262 - TG
...
 #define PIN_GPS_PPS (17) // Pulse per second input from the GPS
 
+// On RAK2560 the GPS is be on a different UART
+#ifdef HAS_RAKPROT
+// #define GPS_RX_PIN PIN_SERIAL2_RX
+// #define GPS_TX_PIN PIN_SERIAL2_TX
+// #define PIN_GPS_EN PIN_3V3_EN
+// Disable GPS
+#define MESHTASTIC_EXCLUDE_GPS 1
+#else
```
caveman99, line 238: This does not only disable GPS but removes all GPS code from the binary. I assume goal is to have a GPS active on Serial2?

caveman99 (changes requested):
Thanks for the contribution, this is a very useful device in the line-up cause we have many requests for ready-made devices. Please see my review comments, hoping i judged the intent of the changes correctly.

caveman99:
also worth noting: We do have an ENUM for devices to identify who is actually talking to us. Right now this code is masquerading as a RAK4631. I'd really like to assign this its own RAK2560 label or even better call it WISMESH-HUB. If you agree i'd open up a pull request to the devicenum protobuf.

beegee-tokyo:
Hi @caveman99 

(1) "Please don't check in the vscode files" 
Not sure what happened,  I guess my PIO installation messed it up. I will reverse that.

(2) " if the values really should be able to go out "
Personally I am not sure if a device will/should report it's battery level over the mesh. I guess that having them in the telemetry in power metrics should be sufficient. I will move the RAK9154 functions into the variant folder and check if compilation works. If users want these values over the mesh, we can add this later.

(3) "I assume goal is to have a GPS active on Serial2?"
The RAK2560 doesn't have GPS by default, I did some tests through Serial 2, but the GPS module was not detected. That's why I removed GPS for now. I hope with the help of my R&D colleagues at RAK I get it to work, but for now it doesn't.

(4) "I'd really like to assign this its own RAK2560 label"
Yeah, I wanted that, but I was not sure how to do it. For the name itslef,  "WisMesh Hub" as a name is still under discussion. Better we use the RAK2560 as name, similar as for the RAK10700, which is a pre-assembled device as well.

I will work on the changes and push again.

caveman99:
1) regarding the vscode stuff, it does that all over the place. Just have to be careful to not check the changes in, we supply the files as a default config.
4) the procedure is to configure a new values in meshtastic/protobufs - same for new (remote) sensors. I'll do that for the RAK2560 now.

beegee-tokyo:
I am sure I enabled "maintainer edits", but it might be related to this strange response I got to the CLA:
image
First it said signed, then it says "not signed yet". I clicked the "recheck" several times already.
image

beegee-tokyo:
image

variants/rak4631/variant.h
Talie5in: @caveman99 Just thinking out loud, should this go into its own variant/rak2560/ - otherwise some code gets clobbered (eg, the JLink CDC pin assignment, unsure if that's used but noticed it while going through the changes)

beegee-tokyo: @Talie5in 
I can create extra variant files for RAK2560, but I wanted to avoid double maintenance, RAK2560 is just a RAK4631 on a different PCB with a solar power battery. With both in one variant.h I have to work only with one file if there are changes required.

Open for suggestions @caveman99

caveman99: really depends on how much it diverges from a stock 4631+Baseboard. Right now there's only a handful of defines. In the long term i think i'd prefer a separate variant.h file, since this is more or less a fixed configuration while we have to account for all sorts of modules on the wisblock boards. We broke out an EINK version too, since the defines were diverging too much and it got harder and harder to keep them apart.

thebentern:
The RAK2560 HW_MODEL protobufs should be available downstream now.

beegee-tokyo:
> The RAK2560 HW_MODEL protobufs should be available downstream now. @caveman99 this would just need to be pretty much a carbon copy of the RAK4631 but with the variant.h changes and the additional macro?

@thebentern 
Anything I need add in the variants files?

thebentern:
> @thebentern Anything I need add in the variants files?

I believe all you should need is to pull over what you have for the RAK4631 variant.h, since they're identical except for the serial port changes and additional macros.

variants/rak2560/platformio.ini
```diff
@@ -0,0 +1,24 @@
...
+  https://github.com/RAKWireless/RAK13800-W5100S.git#1.0.2
+  rakwireless/RAKwireless NCP5623 RGB LED library@^1.0.2
+  beegee-tokyo/RAKwireless RAK12034@^1.0.0
+  https://github.com/beegee-tokyo/RAK-OneWireSerial.git
+debug_tool = jlink
+; If not set we will default to uploading over serial (first it forces bootloader entry by talking 1200bps to cdcacm)
+;upload_protocol = jlink
+extra_scripts = 
+	${env.extra_scripts}
+	../firmware/variants/rak2560/create_uf2.py
```
caveman99, line 24: you can drop the `../firmware` part, this doesn't work if the local folder is named differently. `variants/rak2560/create_uf2.py` should pick the script up just fine. The platfirmio.ini always assumes workspace root.

variants/rak2560/platformio.ini
```diff
@@ -0,0 +1,24 @@
+; The very slick RAK wireless RAK 4631 / 4630 board - Unified firmware for 5005/19003, with or without OLED RAK 1921
+[env:rak2560]
+extends = nrf52840_base
+board = wiscore_rak4631
+board_check = true
+build_flags = ${nrf52840_base.build_flags} -Ivariants/rak2560 -D RAK_4631
+  -L "${platformio.libdeps_dir}/${this.__env__}/bsec2/src/cortex-m4/fpv4-sp-d16-hard"
+  -DGPS_POWER_TOGGLE ; comment this line to disable triple press function on the user button to turn off gps entirely.
+  -DHAS_RAKPROT=1 ; Define if RAk OneWireSerial is used (disables GPS)
+build_src_filter = ${nrf52_base.build_src_filter} +<../variants/rak2560> +<mesh/eth/> +<mesh/api/> +<mqtt/>
```
caveman99, line 10: src/platform/nrf52/architecture.h needs to be amended to pick up the new hardware model, before the plain 4630 is found.

GUVWAF:
Also maybe slightly unrelated to this PR and needless to say, but if you plan to pre-flash them as Router/Repeater please make sure to not copy completely the same configuration to each device, because all devices will get the same unique NodeNum and that leads to issues on the mesh.

caveman99 (approved):

beegee-tokyo:
@caveman99 @thebentern 

Thanks a lot for the support to merge this in.

@GUVWAF 
We will flash a plain Meshtastic firmware with no setup (not even as router or repeater).

beegee-tokyo:
@caveman99 @thebentern 

Last question

Created a firmware from original sources for RAK2560.
The Meshtastic App now shows "Hardware model: UNRECOGNIZED"
Does it need an app update to show the model correct?

geeksville:
hi @beegee-tokyo I just happened to notice this bug and your comment.  Yeah.  Clients will eventually get the latest protobuf code and then properly convert the enum (which they don't recognize without those new strings) to the device name.

(long time no see - I hope all is great for you -kevin).

beegee-tokyo:
@geeksville 

All good here, busy with RAKwireless WisBlock. And as RAKwireless is now jumping full speed on the Meshtastic train, I am on your trails now.

No problem for now with the module name, it will take some weeks (I guess) before you can officially buy it. 
I am just ahead with the preparation and testing.

geeksville:
@beegee-tokyo  cool beans.  I've been away from meshtastic for a while but I've decided to devote some time to **really** decreasing power consumption.  I'm initially using a RAK4630 that I had in a drawer as my reference platform for this work.  It seems great!

beegee-tokyo:
@geeksville 
Let me know if you need anything. I am sitting literally on the source 😁 
Btw, we have distributor in US, Rokland in Florida (I think) if you need any parts.

PR #4123, opened 2024-06-16
Add rak4631_dap variant for debugging with NanoDAP debug probe device.

geeksville:
Feedback/review requested.  Docs for this are in:  https://github.com/meshtastic/meshtastic/pull/1295 

Thoughts?

### DIFF
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -254,7 +254,7 @@ jobs:
           chmod +x ./output/device-update.sh
 
       - name: Zip firmware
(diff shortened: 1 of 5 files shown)


thebentern:
@geeksville looks good. Can you also add `board_level = extra`? That will prevent the CI from picking it up in release asset bundles.

geeksville:
> @geeksville looks good. Can you also add `board_level = extra`? That will prevent the CI from picking it up in release asset bundles.

ooh that's great!  will do.

caveman99 (approved):

PR #4129, opened 2024-06-17
Add support for Heltec HRU-3601

ndoo:
Board is very similar to the Heltec HT-C62 based boards (Heltec HT62 variant) but due to wiring of SK6812 Neopixel LED to GPIO2 it becomes incompatible due to the regular HT-C62 dev board using a simple LED on the same GPIO. Depends on [protobufs#521](https://github.com/meshtastic/protobufs/pull/521).

Works:

* SK6812 Neopixel on GPIO2
* [GXCAS GXHTV3](https://www.lcsc.com/product-detail/Temperature-Sensors_GXCAS-GXHTV3C_C5441730.html) (SHTC3 compatible)

Not working:

* SX1262 initialization after deep sleep (seems like an issue out of scope of the PR)

Won't fix:

* Battery reading - Board has no voltage divider on VBAT (board has a 1.25mm pitch "JST" style connector and a TP4054 charge IC)
* Main thread LED - Board has no LED on simple GPIO

Board schematic: [HRU3601.pdf](https://github.com/user-attachments/files/15874850/HRU3601.pdf)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -242,7 +242,7 @@ void setup()
     initDeepSleep();
 
     // power on peripherals
-#if defined(TTGO_T_ECHO) && defined(PIN_POWER_EN)
+#if defined(PIN_POWER_EN)
     pinMode(PIN_POWER_EN, OUTPUT);
     digitalWrite(PIN_POWER_EN, HIGH);
     // digitalWrite(PIN_POWER_EN1, INPUT);
--- a/src/platform/esp32/architecture.h
+++ b/src/platform/esp32/architecture.h
@@ -101,6 +101,8 @@
 #define HW_VENDOR meshtastic_HardwareModel_STATION_G1
 #elif defined(DR_DEV)
 #define HW_VENDOR meshtastic_HardwareModel_DR_DEV
+#elif defined(HELTEC_HRU_3601)
+#define HW_VENDOR meshtastic_HardwareModel_HELTEC_HRU_3601
 #elif defined(HELTEC_V3)
 #define HW_VENDOR meshtastic_HardwareModel_HELTEC_V3
 #elif defined(HELTEC_WSL_V3)
--- a/variants/heltec_hru_3601/platformio.ini
+++ b/variants/heltec_hru_3601/platformio.ini
@@ -1,0 +1,9 @@
+[env:heltec-hru-3601]
+extends = esp32c3_base
+board = adafruit_qtpy_esp32c3
+build_flags =
+  ${esp32_base.build_flags}
+  -D HELTEC_HRU_3601
+  -I variants/heltec_hru_3601
+lib_deps = ${esp32c3_base.lib_deps}
+  adafruit/Adafruit NeoPixel @ ^1.12.0
--- a/variants/heltec_hru_3601/variant.h
+++ b/variants/heltec_hru_3601/variant.h
@@ -1,0 +1,40 @@
+#define BUTTON_PIN 9
+
+#define HAS_SCREEN 0
+#define HAS_GPS 0
+#undef GPS_RX_PIN
+#undef GPS_TX_PIN
+
+#define USE_SX1262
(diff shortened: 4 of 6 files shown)


src/main.cpp
```diff
@@ -407,6 +407,9 @@ void setup()
     Wire.begin();
     i2cScanner->scanPort(ScanI2C::I2CPort::WIRE);
 #elif defined(I2C_SDA) && !defined(ARCH_RP2040)
+#if defined(I2C_PWR_ENABLE)
```
caveman99, line 410: Looking through the source i have a better suggestion. Use PIN_POWER_EN instead. Right now the use is limited to the T-ECHO device but it's the only variant that actually uses that define. so you can remove the limitation to T-ECHO in Line 244 and use that instead.

variants/heltec_hru_3601/platformio.ini
```diff
@@ -0,0 +1,8 @@
+[env:heltec-hru-3601]
+extends = esp32c3_base
+board = adafruit_qtpy_esp32c3
+board_level = extra
+build_flags = 
+  ${esp32_base.build_flags}
+  -D PRIVATE_HW
```
caveman99, line 7: If you have a on-sale device i'd recommend registering a public enumerator in the protobuf registry for it

variants/heltec_hru_3601/variant.h
```diff
@@ -0,0 +1,38 @@
...
+#define SX126X_DIO2_AS_RF_SWITCH
+#define SX126X_DIO3_TCXO_VOLTAGE 1.8
+
+// Board has I2C connected to UART0 pins, and no other hardware serial port
+#define UART_TX -1
+#define UART_RX -1
+
+// Board has I2C connected to U0RXD and U0TXD; I2C module power is controlled on Vext_Ctrl
+#define I2C_PWR_ENABLE 1
+#define I2C_SDA 21
```
caveman99, line 31: See above, i suggest using PIN_POWER_EN

caveman99 (changes requested):

src/main.cpp
ndoo: Thank you, it actually works properly using PIN_POWER_EN, my implementation did not set the GPIO direction correctly anyway. After ungating PIN_POWER_EN and using that code path, the SHTC3 clone gets detected fine in the environment telemetry module.

variants/heltec_hru_3601/platformio.ini
ndoo: I'm not actually working for Heltec but just a random user of this device, but I have opened a PR in protobufs for it: https://github.com/meshtastic/protobufs/pull/521

ndoo:
I noticed that the SX1262 does not initialize when the board wakes up from deep sleep, with `SX126x init result -2` followed by a critical error.

I've double checked the pins in variant.h against the schematic and it does not appear to be caused by this PR but a regression/issue elsewhere.

caveman99 (approved):

caveman99:
Looks good to me, I'll review the protobuf change now.

caveman99:
Protobufs are in place now, just need to change from PRIVATE_HW to a unique tag you chose and add the translation to architecture.h of the ESP platform source.

src/platform/esp32/architecture.h
```diff
@@ -101,6 +101,8 @@
 #define HW_VENDOR meshtastic_HardwareModel_STATION_G1
 #elif defined(DR_DEV)
 #define HW_VENDOR meshtastic_HardwareModel_DR_DEV
+#elif defined(HELTEC_HRU_3601)
```
ndoo, line 104: The current commit in this PR already uses HELTEC_HRU_3601 tag so it should be good to merge

ndoo:
Rebase on HEAD in an attempt to make CI happy

caveman99 (approved):

PR #4134, opened 2024-06-18
Show owner.short_name on boot (and E-Ink sleep screen)

Dorn8010:
Had to install Visual Studio Code and trunk. Manual editing on my linux Python IDE didn't work so well :)
Thx for your patience guys.

Now all the changes done and compiled on my device - old pull request closed.
All checked, compiled and it works (locally Heltec V3 & Wireless Paper 1.1).

Didn't add the HEX ID on the boot screen - not to clutter the space.
But since I think the HEX ID should be displayed for consistency of the whole Meshtastic ecosystem somewhere on a screen, I will try to come up with a solution we agree on later.
For now boot looks like REGION:short_name

Hope this is all OK with you,
Greetings from Austria


todd-herbert:
> Didn't add the HEX ID on the boot screen - not to clutter the space.
But since I think the HEX ID should be displayed for consistency of the whole Meshtastic ecosystem somewhere on a screen, I will try to come up with a solution we agree on later.
For now boot looks like REGION:short_name

Sorry for the slow reply! If I'm honest, I'm not sure how everyone would feel about generally moving towards the Hex ID instead of the short name. I don't think it's a bad idea but might need some wider discussion. Let's try with these changes here though, at least as a start.

I know there are some quite small OLED screens being used out there. I'll try it out today on my T-Beam, which has a slightly smaller screen that Wireless Paper. Also to check what happens with region UNSET, or when booting after a factory reset. I'll get back to you!

Dorn8010:
Pls take your time guys. I'm new in this so apologize the newbie errors I make all the time :) 
The hex ID is the backbone of the net, so it should be shown somewhere.
For the normal user the name is better - but if you want to call e.g. a silent node, the HEX ID is crucial.
I would see the HEX ID like the main connection to a device and logically a crucial information (like the unique id in a database) which should be shown somewhere. The boot screen is a bad idea anyways, since its shown shortly - so a normal screen, like the one where the amount of connected nodes is shown, would be better. 
I thought behind the LongFast mode info in small font.
Since I come from software design and strategy, I'm more into concept than programming as you can see :)))

PS: If you don't like the idText pointer solution I can change all of them in the Screen.cpp.

Dorn8010:
PPS: I have two new classes a StringDB for RAM optimized string management and storage to flash and a msgDB to safe and handle the msgs and also for a get CLI call of the last X messages e.g. filtered by a node.
Since there is a big RAM and flash gap between the devices MT runs on, I don't know where to put the code that you guys can see it but I don't disturb the workflow. Maybe you give me a hint.

todd-herbert:
After testing this out today, the only thought I had on this specific PR is that maybe `owner.short_name` could be drawn as part of `drawIconScreen`, (and `drawOEMIconScreen`), and shown on the right. This is only a matter of taste though, so only a suggestion!

eink boot eink sleep
oled boot

> Since I come from software design and strategy, I'm more into concept than programming as you can see :)))
> PS: If you don't like the idText pointer solution I can change all of them in the Screen.cpp.

Hey no real objection to the `idText` pointer. Just a heads up though: the firmware code is definitely layer-upon-layer of revision by different contributors, so it's a bit of a patchwork. I think you could go mad trying to standardize it 😉 

___

>I would see the HEX ID like the main connection to a device and logically a crucial information (like the unique id in a database) which should be shown somewhere. 

>The boot screen is a bad idea anyways, since its shown shortly - so a normal screen,
I thought behind the LongFast mode info in small font.

> PPS: I have two new classes a StringDB for RAM optimized string management and storage to flash and a msgDB to safe and handle the msgs and also for a get CLI call of the last X messages e.g. filtered by a node.
Since there is a big RAM and flash gap between the devices MT runs on, I don't know where to put the code that you guys can see it but I don't disturb the workflow. Maybe you give me a hint.

Some interesting thoughts there! It's all bit beyond me; I mostly just spot bugs and tweak stuff. Have you come across the [Meshtastic Discord server](https://discord.com/invite/ktMAKGBnBs)? A lot of the devs are active there; it seems like the hub for discussion about big picture stuff like this.

thebentern:
@todd-herbert I like your proposed alternative for the consistency of placement of the short name

todd-herbert:
> the firmware code is definitely a patchwork from different contributors

(Not to be derogatory or anything! Just a sign of the large number of people who have given their time)

todd-herbert:
@Dorn8010 @thebentern Any objections or notes on https://github.com/Dorn8010/firmware-chgs/pull/1?

PR #4140, opened 2024-06-19
Fix protobuf structs handling

mskvortsov:
This fixes minor warnings as reported by a newer compiler (as in #3960).

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -141,11 +141,6 @@ NodeDB::NodeDB()
     if (channelFileCRC != crc32Buffer(&channelFile, sizeof(channelFile)))
         saveWhat |= SEGMENT_CHANNELS;
 
-    if (!devicestate.node_remote_hardware_pins) {
-        meshtastic_NodeRemoteHardwarePin empty[12] = {meshtastic_RemoteHardwarePin_init_default};
-        memcpy(devicestate.node_remote_hardware_pins, empty, sizeof(empty));
-    }
-
     if (config.position.gps_enabled) {
         config.position.gps_mode = meshtastic_Config_PositionConfig_GpsMode_ENABLED;
         config.position.gps_enabled = 0;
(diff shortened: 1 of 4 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -141,11 +141,6 @@ NodeDB::NodeDB()
     if (channelFileCRC != crc32Buffer(&channelFile, sizeof(channelFile)))
         saveWhat |= SEGMENT_CHANNELS;
 
-    if (!devicestate.node_remote_hardware_pins) {
```
caveman99, line 144: Is there a replacement for this?

src/mqtt/MQTT.cpp
```diff
@@ -482,7 +482,8 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp, const meshtastic_MeshPacket &
 
     auto &ch = channels.getByIndex(chIndex);
 
-    if (&mp_decoded.decoded && strcmp(moduleConfig.mqtt.address, default_mqtt_address) == 0 &&
+    assert(mp_decoded.which_payload_variant == meshtastic_MeshPacket_decoded_tag);
```
caveman99, line 485: Please don't introduce new asserts. We're trying to get rid of them.

caveman99 (changes requested):

src/mesh/NodeDB.cpp
mskvortsov: There is node_remote_hardware_pins_count. But then the whole if block would be a no-op, wouldn't it?

src/mqtt/MQTT.cpp
mskvortsov: Done.

src/mesh/NodeDB.cpp
mskvortsov: (I meant it would be a no-op not literally, but logically: there seems to be no reason for additionally erasing node_remote_hardware_pins, considering its usage in the rest of the code.)

caveman99: That code is from @thebentern i am sure there was a reason for this.

caveman99 (approved):

src/mesh/NodeDB.cpp
thebentern: I believe this was to initialize value for devices that had a device state from before the addition of that property.

caveman99: @thebentern you reckon we still need this?

PR #4147, opened 2024-06-20
heltec-ht62-esp32c3-sx1262: Assign GPIO18 and GPIO19 to I2C

ndoo:
I2C is not defined on schematics, but these are the only free GPIO exposed on the castellated pads.

There are pin mux conflicts: GPIO18 = USB_DN and GPIO19 = USB_DP.

Since the HT-DEV-ESP provides a CP2102, re-use these ports for I2C instead since it's more useful (attaching peripherals vs. connecting to a host PC). Boards which use the native ESP32 USB will require a separate variant (refer to Heltec HRU-3601 variant for example).

Perhaps a "heltec_esp32c3_usb" variant would make things easier?

Took the opportunity to clean up the variant's header files too.

Relevant forum post/discussion in [Heltec HT-CT62 support - Meshtastic’s future?#14](https://meshtastic.discourse.group/t/heltec-ht-ct62-support-meshtastics-future/11391/14).

### DIFF
--- a/variants/heltec_esp32c3/variant.h
+++ b/variants/heltec_esp32c3/variant.h
@@ -1,16 +1,52 @@
-#define BUTTON_PIN 9
+/*
+ * Heltec HT-CT62, Heltec HT-DEV-ESP V2 and Heltec HT-DEV-ESP V3
+ *
+ * Heltec HT-CT62 Schematic: https://resource.heltec.cn/download/HT-CT62/HT-CT62_Schematic_Diagram.pdf
+ * Heltec HT-CT62 Reference Design: https://resource.heltec.cn/download/HT-CT62/HT-CT62_Reference_Design.pdf
+ * Note: Reference design is just an application note, but no actual product exists
+ *
+ * Heltec HT-DEV-ESP V3 Schematic: https://resource.heltec.cn/download/HT-DEV-ESP/HT-DEV-ESP_V2_Sch.pdf
+ * Heltec HT-DEV-ESP V3 Schematic: https://resource.heltec.cn/download/HT-DEV-ESP/HT-DEV-ESP_V3_Sch.pdf
+ * Note: Heltec sells HT-DEV-ESP boards
+ *
+ * Note: GPIO0 and GPIO1 are connected to 32kHz crystal on HT-DEV-ESP boards and in reference design
+ *
+ */
 
-// LED pin on HT-DEV-ESP_V2 and HT-DEV-ESP_V3
-// https://resource.heltec.cn/download/HT-CT62/HT-CT62_Reference_Design.pdf
-// https://resource.heltec.cn/download/HT-DEV-ESP/HT-DEV-ESP_V3_Sch.pdf
-#define LED_PIN 2 // LED
-#define LED_INVERTED 0
-
-#define HAS_SCREEN 0
+// Module does not have enough free pins for a UART GPS
 #define HAS_GPS 0
 #undef GPS_RX_PIN
 #undef GPS_TX_PIN
 
+// USER_SW (HT-DEV-ESP V2/V3)
+#define BUTTON_PIN 9
+
+/*
+ * LED (HT-DEV-ESP_V2/V3)
+ * HT-CT62 reference schematic connects the LED to GPIO18 but HT-DEV-ESP schematic connects the LED to GPIO2
+ */
+#define LED_PIN 2
+#define LED_INVERTED 0
+
+/*
+ * I2C
+ * Not defined on schematics, but these are the only free GPIO exposed on the castellated pads.
+ * There are pin mux conflicts: GPIO18 = USB_DN and GPIO19 = USB_DP. Since the HT-DEV-ESP provides a CP2102, re-use these ports
+ * for I2C instead since it's more useful (attaching peripherals vs. connecting to a host PC). Boards which use the native ESP32
+ * USB will require a separate variant (refer to Heltec HRU-3601 variant for example). Perhaps a "heltec_esp32c3_usb" variant
+ * would make things easier?
+ */
+#define I2C_SCL 18
+#define I2C_SDA 19
+
+// Since we have I2C, probe for screens
+#define HAS_SCREEN 1
+
+// U0TXD, U0RXD to CP2102 (HT-DEV-ESP_V2/V3)
+#define UART_TX 21
+#define UART_RX 20
+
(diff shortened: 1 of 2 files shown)


variants/heltec_esp32c3/variant.h
```diff
@@ -1,16 +1,43 @@
...
-#define LED_INVERTED 0
+/*
+ * Heltec HT-CT62, Heltec HT-DEV-ESP V2 and Heltec HT-DEV-ESP V3
+ * Heltec HT-CT62 Schematic: https://resource.heltec.cn/download/HT-CT62/HT-CT62_Reference_Design.pdf
+ * Heltec HT-DEV-ESP V3 Schematic: https://resource.heltec.cn/download/HT-DEV-ESP/HT-DEV-ESP_V2_Sch.pdf
+ * Heltec HT-DEV-ESP V3 Schematic: https://resource.heltec.cn/download/HT-DEV-ESP/HT-DEV-ESP_V3_Sch.pdf
+ */
 
 #define HAS_SCREEN 0
+
```
caveman99, line 9: If you define I2C pins, why not activate the oled screen code?

variants/heltec_esp32c3/variant.h
```diff
@@ -1,16 +1,43 @@
...
+// Board does not have enough free pins for a UART GPS, use an I2C GPS
 #define HAS_GPS 0
```
caveman99, line 11: I2C GPS is not supported on Meshtastic.

variants/heltec_esp32c3/variant.h
```diff
@@ -1,16 +1,43 @@
...
+ * I2C
+ * Not defined on schematics, but these are the only free GPIO exposed on the castellated pads.
+ * There are pin mux conflicts: GPIO18 = USB_DN and GPIO19 = USB_DP. Since the HT-DEV-ESP provides a CP2102, re-use these ports
+ * for I2C instead since it's more useful (attaching peripherals vs. connecting to a host PC). Boards which use the native ESP32
+ * USB will require a separate variant (refer to Heltec HRU-3601 variant for example). Perhaps a "heltec_esp32c3_usb" variant
+ * would make things easier?
+ */
+#define I2C_SCL 18
+#define I2C_SDA 19
+
```
caveman99, line 35: What about the original Pins 0 and 1?

caveman99 (changes requested):

variants/heltec_esp32c3/variant.h
LyndonBlack: Not sure that anyone using this particular board would want to hook up a oled, way easier to spend the extra couple of dollars and get the Heltec V3 with the screen included.

variants/heltec_esp32c3/variant.h
LyndonBlack: Thanks for that. I had kinda come to work that out yesterday but just wish i had dug a little deeper before spending a lot of time mucking around with it. A man with a little knowledge is a man with no knowledge at all.

variants/heltec_esp32c3/variant.h
ndoo: > What about the original Pins 0 and 1?

These are used for a 32kHz crystal oscillator on the HT-DEV-ESP32 and reference design, https://resource.heltec.cn/download/HT-DEV-ESP/HT-DEV-ESP_V3_Sch.pdf and https://resource.heltec.cn/download/HT-CT62/HT-CT62_Reference_Design.pdf respectively

variants/heltec_esp32c3/variant.h
ndoo: > I2C GPS is not supported on Meshtastic.

Fair enough, let me remove the remark about GPS then. The limited I/O is really getting tough to make decisions with...

GPIO18 = I2C_SCL (could have been USB_DN or UART1)
GPIO19 = I2C_SDA (could have been USB_DP or UART1)
GPIO20 = U0RXD
GPIO21 = U0TXD

I feel like this has the most utility for most "classic" MCU hackers who are used to getting to their boards over TTL serial and want to have I2C peripherals.

Maybe U0RXD and U0TXD can be sacrificed for GPS too, making the port do double duty as flashing UART and GPS UART?

Not really sure what's preferred/most useful at this point. So I guess this PR is becoming a bit of an RFC.

variants/heltec_esp32c3/variant.h
ndoo: > If you define I2C pins, why not activate the oled screen code?

Good point, will enable it then. I'm assuming the BLE pairing PIN generation checks to see if the OLED was actually initialized then.

ndoo:
Pushed changes to address the review feedback.

Still very much an RFC, given the very limited GPIO available, we have to figure out what mapping makes the most sense for the community.

Why Heltec exposed GPIO that were already connected to the SX1262 beats me, probably just due to PCB routing limitations or not using blind vias, with however many layers they are using.

variants/heltec_esp32c3/variant.h
ndoo: > Thanks for that. I had kinda come to work that out yesterday but just wish i had dug a little deeper before spending a lot of time mucking around with it. A man with a little knowledge is a man with no knowledge at all.

There is a way to get both I2C and GPS UART working, which is to sacrifice UART0 entirely and reassign the pins to GPS.

That means that you have to disconnect the GPS while putting the ESP32 in download mode, flash your firmware, then reconnect the GPS. There isn't even 1 more pin of GPIO free to control the enable pin of the GPS.

Alternatively, we can use GPIO0 or GPIO1 to drive the GPS enable, but then HT-DEV-ESP board users have to desolder the 32kHz crystal to free up GPIO0 and GPIO1.

variants/heltec_esp32c3/variant.h
ndoo: As commented in the other thread, GPIO0 and GPIO1 can be freed up for HT-DEV-ESP users by desoldering the 32kHz crystal and passives. I don't think this crystal provides any benefit to Meshtastic anyway; the HT-CT62 module has a 40MHz crystal onboard for the ESP32-C3 already.

The 32kHz crystal is just an extra clock source and desoldering it won't affect HT-DEV-ESP board functionality. And anyone building a Meshtastic device around the HT-CT62 is not likely to bother with adding a 32kHz crystal either.

So this is another option we can consider too.

caveman99 (approved):

PR #4149, opened 2024-06-20
Make serial port on wio-sdk-wm1110 board work

geeksville:
By disabling the (inaccessible) adafruit USB

### DIFF
--- a/boards/wio-sdk-wm1110.json
+++ b/boards/wio-sdk-wm1110.json
@@ -7,13 +7,6 @@
     "cpu": "cortex-m4",
     "extra_flags": "-DARDUINO_WIO_WM1110 -DNRF52840_XXAA",
     "f_cpu": "64000000L",
-    "hwids": [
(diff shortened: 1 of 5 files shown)


geeksville:
oh - anyone know if it possible to suppress trunk checks on a particular python script?  the script that failed is used by platformio directly and is valid when imported in that environment.  alternatively I could delete  it for now...

caveman99:
Depends on the failing linter. Most either support online comments or a point file in the trunk subdir.

geeksville:
ok - i'll read up on trunk. no prob.

thebentern:
To be honest, it wouldn't hurt my feelings if we turned trunk off altogether on the python scripts. I'm not a fan off all of the linting of configuration and scripting languages. Seems excessive.

thebentern (approved):

caveman99 (approved):

PR #4153, opened 2024-06-21
Heltec Tracker v1 and v1.1 - Power Config

### LINKED ISSUE #3760: [Bug]: Heltec Wireless Tracker Battery issues - possible fix
### Category

Other

### Hardware

Heltec Wireless Tracker

### Firmware Version

2.3.7.30fbcab

### Description

THE PROBLEM
The heltec wireless tracker has poor battery performance and battery monitoring doesn't seem to work at all. It always shows that it is powered by USB and will over drain the battery and potentially cause a fire. 

POSSIBLE FIX
There is a recent (within the last 15 days) post on the Meshtastic discourse forums with a potential fix that I'm hoping can be implemented. https://meshtastic.discourse.group/t/heltec-tracker-v1-and-v1-1-power-config/12016/4

### Relevant log output

_No response_

mkgin:
TL;DR.

Output high to measure voltage in the same why as other heltec device forces the ESP32 to try to pump 3.3v into the base of Q3.
Solution, use internal pullup and pulldown resistor

Fixes #3760

WORK IN PROGRESS

Progress so far:
* Hasn't caught fire yet.
* Battery measurements seems ok

TODO:

* Measure current, check voltages at relevant pins.
  * Not in a position to measure differences for a few days.
* take a better look at code
* CLA issues? credit to OP?
* check circuit for tracker v1.0 and add it if there is the same issue


mkgin:
As mentioned [here...](https://github.com/meshtastic/firmware/issues/3760#issuecomment-2247769450) haven't been able to do anything with this. Have a chance to look at this again now. 

Todo:
* add the original HELTEC_TRACKER to the `#ifdef`
* test ( I only have 1.1)
* measure

PR #4156, opened 2024-06-21
Standardize lat/lon position logs

### LINKED ISSUE #4150: [Feature Request]: Identify log data that may need privacy redaction
### Platform

### Description

It would be nice if logs that are output with coordinates in them could be identified and redacted by connected clients via a regex.

thebentern:
Closes #4150
New standard is `lat=xxx lon=xxx`


garthvh (approved):
Looks great

PR #4170, opened 2024-06-24
Display alerts

jp-bennett:
This replaces multiple oddball functions in Screen.cpp with startAlert() and endAlert() functions, that take either a callback function, or simple string.


jp-bennett:
This actually works now. Not sure I'm excited about just moving the static functions into Screen.h, but it gets it done.

src/modules/AdminModule.cpp
```diff
@@ -811,7 +816,12 @@ void AdminModule::handleGetChannel(const meshtastic_MeshPacket &req, uint32_t ch
 void AdminModule::reboot(int32_t seconds)
 {
     LOG_INFO("Rebooting in %d seconds\n", seconds);
-    screen->startRebootScreen();
+    screen->startAlert([](OLEDDisplay *display, OLEDDisplayUiState *state, int16_t x, int16_t y) -> void {
+        uint16_t x_offset = display->width() / 2;
+        display->setTextAlignment(TEXT_ALIGN_CENTER);
+        display->setFont(FONT_MEDIUM);
+        display->drawString(x_offset + x, 26 + y, "Rebooting...");
+    });
```
todd-herbert, line 829: It feels like now we're repeating this same piece of code whenever we want to reboot / shutdown, rather than calling it from a central location.

src/graphics/Screen.cpp
```diff
@@ -2225,59 +2190,13 @@ void Screen::setFrames()
...
-
-void Screen::handleRebootScreen()
-{
-    LOG_DEBUG("showing reboot screen\n");
-    showingNormalScreen = false;
-#ifdef USE_EINK
-    EINK_ADD_FRAMEFLAG(dispdev, DEMAND_FAST); // Use fast-refresh for next frame, no skip please
-    EINK_ADD_FRAMEFLAG(dispdev, BLOCKING);    // Edge case: if this frame is promoted to COSMETIC, wait for update
-    handleSetOn(true);                        // Power-on to show rebooting screen (PowerFSM should handle?)
-#endif
```
todd-herbert, line 2331: These little details are quite important for Heltec Wireless Paper and T-Echo. They make sure that these special frames do actually draw.

todd-herbert (commented):
I'm actually wondering if it might be good to handle these special cases separately, rather than directly as alerts.

src/graphics/Screen.cpp
jp-bennett: Agreed. Need to make sure they all make it back in.

src/modules/AdminModule.cpp
jp-bennett: Made a helper function that just takes the string, to get rid of this boilerplate.

thebentern (approved):

jp-bennett:
> I'm actually wondering if it might be good to handle these special cases separately, rather than directly as alerts.

If we need to we can, but I'm hopeful the ability to define a custom callback will make it universally useful.

todd-herbert (commented):
Hey I'm not sure a one-size-fits-all approach like this works with the E-Ink flags. I'm not sure, might get away with it; I just remember it taking a bit of close attention to flag all the frames right to make sure they each do their own specific job.

jp-bennett:
> Hey I'm not sure a one-size-fits-all approach like this works with the E-Ink flags. I'm not sure, might get away with it; I just remember it taking a bit of close attention to flag all the frames right to make sure they each do their own specific job.
> 
> I can give it a test run later today and see if it's okay to handle them all like this.

Hmm. I think I have one of the e-ink devices around here. I'll see if I can get some testing in, too. A repeatable case where it's broken will be helpful to decide on an approach.

todd-herbert:
Thanks. The whole thing just makes me a bit nervous about trying to catch everything.

todd-herbert (approved):
I think it's probably all okay!

PR #4194, opened 2024-06-28
Clear vector after complete config state

thebentern:


mverch67:
I'd still suggest to check for empty size(9 otherwise it'll end up in an endless loop, e.g.:

```
    case STATE_SEND_FILEMANIFEST: {
        LOG_INFO("getFromRadio=STATE_SEND_FILEMANIFEST, size=%d\n", filesManifest.size());
        if (filesManifest.size() > 0) {
            fromRadioScratch.which_payload_variant = meshtastic_FromRadio_fileInfo_tag;
            if (config_state < filesManifest.size()) {
                fromRadioScratch.fileInfo = filesManifest.at(config_state);
                config_state++;
                // last element
                if (config_state == filesManifest.size()) {
                    filesManifest.clear();
                    state = STATE_SEND_COMPLETE_ID;
                    config_state = 0;
                }
            }
        }
        else {
            state = STATE_SEND_COMPLETE_ID;
        }
        break;
    }
```

PR #4199, opened 2024-06-28
Make the logs Colorful!

jp-bennett:
Add some color to the log output. Still looking for how to render this properly in vscode.

### DIFF
--- a/src/RedirectablePrint.cpp
+++ b/src/RedirectablePrint.cpp
@@ -71,20 +71,49 @@ size_t RedirectablePrint::vprintf(const char *format, va_list arg)
     return len;
 }
 
-size_t RedirectablePrint::log(const char *logLevel, const char *format, ...)
+size_t RedirectablePrint::vprintf(const char *logLevel, const char *format, va_list arg)
+{
+    va_list copy;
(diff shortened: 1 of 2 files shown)


thebentern (approved):

geeksville:
IMO this slightly breaks automated parsing of  logs.  IMO we shouldn't assume on the device side that the viewer will be reading it on a vt100ish screen.  They (like me) might be feeding those logs into another tool or saving them in a text file.

IMO we might want to revisit it in the future.

geeksville:
btw - if you use the meshtastic python client it will already properly render inside of vscode.

geeksville:
actually - thinking a bit more on this, it isn't too bad (though not ideal): because whenever the client talks to the device via the API log messages will switch to being sent encapsulated in LogRecords (and we won't do coloring inside those messages).

So I think fine (but putting control chars into text files for people who are just logging to a file is not great).

jp-bennett:
@geeksville , @thebentern half-joked that this would inevitably lead to yet another config option, to turn it off. =) If it becomes a problem, we can look into making it optional, or turned off on some platforms

PR #4201, opened 2024-06-29
Deprecate Router Client role (and make it Client)

### LINKED ISSUE #4185: [Feature Request]: Admit Router Client was a mistake
### Platform

### Description

Router client was a bad choice, make router client behave exactly as client does and disable BLE for router so that the role is only used for infrastructure nodes as intended.

thebentern:
Closes #4185

### DIFF
--- a/src/modules/esp32/StoreForwardModule.cpp
+++ b/src/modules/esp32/StoreForwardModule.cpp
@@ -319,8 +319,8 @@ ProcessMessage StoreForwardModule::handleReceived(const meshtastic_MeshPacket &m
 #ifdef ARCH_ESP32
     if (moduleConfig.store_forward.enabled) {
 
-        // The router node should not be sending messages as a client. Unless he is a ROUTER_CLIENT
-        if ((getFrom(&mp) != nodeDB->getNodeNum()) || (config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER_CLIENT)) {
+        // The router node should not be sending messages as a client
+        if ((getFrom(&mp) != nodeDB->getNodeNum())) {
 
             if ((mp.decoded.portnum == meshtastic_PortNum_TEXT_MESSAGE_APP) && is_server) {
                 auto &p = mp.decoded;
--- a/src/mesh/FloodingRouter.cpp
+++ b/src/mesh/FloodingRouter.cpp
@@ -22,7 +22,6 @@ bool FloodingRouter::shouldFilterReceived(const meshtastic_MeshPacket *p)
     if (wasSeenRecently(p)) { // Note: this will also add a recent packet record
         printPacket("Ignoring incoming msg we've already seen", p);
         if (config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER &&
-            config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER_CLIENT &&
             config.device.role != meshtastic_Config_DeviceConfig_Role_REPEATER) {
             // cancel rebroadcast of this message *if* there was already one, unless we're a router/repeater!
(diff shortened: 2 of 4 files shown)


garthvh (approved):
Looks good

src/modules/esp32/StoreForwardModule.cpp
GUVWAF, line 1: Shall we then also get rid of the requirement that a StoreForward server needs to be a `ROUTER`/`ROUTER_CLIENT` (line 536)? I think that might also be a reason why it was used a lot.

GUVWAF: Oh, I now see this was discussed in #4185. 
Store & Forward doesn't make sense for a mobile node, but I have a powered T-Beam inside I use as Store & Forward router and a RAK on the roof with much better coverage. I would rather have the T-Beam `CLIENT_MUTE` and still be able to get missed messages.

thebentern: How should we determine to put s&f in client mode? We used Router + Router client to determine this previously.
Instead of deprecating Router Client, should we perhaps rename to STATION to indicate a normal client which is at a fixed location and enable the serving mode of s&f for that role and router?

GUVWAF: I'd prefer to have it not be dependent on the role, but a dedicated S&F setting.

garthvh: > How should we determine to put s&f in client mode? We used Router + Router client to determine this previously. Instead of deprecating Router Client, should we perhaps rename to STATION to indicate a normal client which is at a fixed location and enable the serving mode of s&f for that role and router?

Let's not create another traffic issue while fixing this mistake, making store and forward easier to enable will likely negate gains from fixing router client misuse.

madeofstown: I don't believe this is going to fix the issue. People who want Store and Forward are just going to have to run in Router mode now and manage with the admin channel.

garthvh: > I don't believe this is going to fix the issue. People who want Store and Forward are just going to have to run in Router mode now and manage with the admin channel.

I see very little store and forward traffic compared to nodes set to router client.

GUVWAF:
Before releasing this, we do need to decide on something because currently all StoreForward `ROUTER_CLIENT`s will not work anymore because they will be made `CLIENT`.

garthvh:
> Before releasing this, we do need to decide on something because currently all StoreForward `ROUTER_CLIENT`s will not work anymore because they will be made `CLIENT`.

Not sure this is a big enough group right now to get too concerned, it would also be good to just disable SF on the default key.

GUVWAF:
Yeah, although I think most of StoreForward routers currently are `ROUTER_CLIENT`. 

Shall we just add a setting `is_server` to the StoreForward config and set it automatically to true if you had the module enabled as `ROUTER_CLIENT`?

Talie5in:
Reading through the Issue comments and this PR..

I still think this is a good opportunity to change how Client/Client Mute are portrayed and/or named and/or described.

Client/Client Mute would be better off renamed to Client Stationary (alternatively Stationary Node?) and Client Mobile (alternatively Mobile Node?), which works well with Client Hidden (Hidden Node?) that exists now.

I feel this better defines the three primary roles (outside of TAK) that really need a better description to fit their (prefered) roles from a network perspective and would make selecting the right role more intuitive.

Otherwise, be good to finally rid of ROUTER_CLIENT misuse!

madeofstown:
@Talie5in, I agree, Client_Mute gives the impression that it won't transmit.

PR #4202, opened 2024-06-29
Move waypoint

jp-bennett:
Seems to be ready. More cleanups, moving code out of Screen and into the Waypoint module.

### DIFF
--- a/src/graphics/Screen.h
+++ b/src/graphics/Screen.h
@@ -82,11 +82,6 @@ class Screen
 #define SEGMENT_WIDTH 16
 #define SEGMENT_HEIGHT 4
 
-static uint16_t displayWidth, displayHeight;
-
-#define SCREEN_WIDTH displayWidth
-#define SCREEN_HEIGHT displayHeight
-
 /// Convert an integer GPS coords to a floating point
 #define DegD(i) (i * 1e-7)
 
@@ -437,6 +432,11 @@ class Screen : public concurrency::OSThread
     int32_t runOnce() final;
 
     bool isAUTOOled = false;
+
+    // Screen dimensions (for convenience)
+    // Defined during Screen::setup
+    uint16_t displayWidth = 0;
+    uint16_t displayHeight = 0;
 
   private:
     FrameCallback alertFrames[1];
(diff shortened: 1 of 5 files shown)


src/graphics/Screen.h
```diff
@@ -82,6 +82,14 @@ class Screen
 #define SEGMENT_WIDTH 16
 #define SEGMENT_HEIGHT 4
 
+static uint16_t displayWidth, displayHeight;
+
+#define SCREEN_WIDTH displayWidth
+#define SCREEN_HEIGHT displayHeight
```
todd-herbert, line 88: Just from some initial investigation, the waypoint compass rose issue comes from how these are accessed. I'll take a look at it in the next few hours unless you're already working on it!

todd-herbert:
Hope it's not intrusive for me to have chipped in on this one. Please do revert or alter if you'd prefer.

src/modules/WaypointModule.cpp
```diff
@@ -2,6 +2,11 @@
 #include "NodeDB.h"
 #include "PowerFSM.h"
 #include "configuration.h"
+#ifdef HAS_SCREEN
+#include "gps/RTC.h"
+#include "graphics/Screen.h"
+#include "main.h"
+#endif
```
todd-herbert, line 9: It looks like this is causing CI to fail for some targets too, because of the `drawFrame` parameters. I took a peek at the EnvironmentTelemetry module is doing, but it looks like it's just including the OLED library, screen or no screen, which doesn't seem quite right either.

todd-herbert:
I'd be keen to streamline what's happening in the waypoint module a bit more eventually, but that feels like a job for another day. Also think it'd be nice at some point to extract all the compass drawing to its own method, but have to be careful about that so it still plays nice with the compass calibration frame.

jp-bennett:
@todd-herbert I'll hop in and fix the build fails in a bit. I know what it needs, just ran out of steam last night. Thanks for the fixes, looks great.

todd-herbert:
Sorry, timezones and all that! Looks good!

PR #4204, opened 2024-06-29
Power.cpp wasn't using the INA sensor if it was available

akohlsmith:
The firmware can use the INA series of sensors for measuring battery voltage and current if it's detected, but the device metrics weren't using them. 

This PR attempts to do that, based on the existing code in Power.cpp.


src/Power.cpp
```diff
@@ -413,35 +427,80 @@ class AnalogBatteryLevel : public HasBatteryLevel
...
+        } else if (nodeTelemetrySensorsMap[meshtastic_TelemetrySensorType_INA3221].first == ina_addr) {
+            ret = ina3221Sensor.getBusVoltageMv();
+
+        /* don't know how to read voltage, return 0mV */
+        } else {
+            ret = 0;
         }
-        return 0;
+
+        return ret;
```
todd-herbert, line 471: I always get the sense that maybe it's good to avoid reworking stuff like this just for the sake of it, to keep the diff down, unless the PR is explicitly tidying or something, but I'm not sure if there's an official stance on it.

src/Power.cpp
```diff
@@ -532,6 +591,10 @@ bool Power::analogInit()
 
     batteryLevel = &analogLevel;
     return true;
+#elif HAS_TELEMETRY && !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR && !defined(ARCH_PORTDUINO)
+    batteryLevel = &analogLevel;
+    return true; // TODO ABK: should be hasINA()
+
```
todd-herbert, line 597: Definitely gotta sort this TODO first

src/modules/Telemetry/Sensor/VoltageSensor.h
```diff
@@ -8,6 +8,7 @@ class VoltageSensor
 {
   public:
     virtual uint16_t getBusVoltageMv() = 0;
+    virtual int16_t getCurrentMa() = 0;
```
todd-herbert, line 11: Just guessing, but rather than actually trying to figure out what's up with RAK9154, see if you can get away with:
```cpp
virtual int16_t getCurrentMa() { return 0; }
```

todd-herbert (commented):
I don't actually have the hardware to test this out with the INA, but I'll give it a smoke test tomorrow on some devices which are using the ADC.

Just watch out for the linting too. Gotta get trunk set up!

todd-herbert:
Just another thought, it'd probably be handy too to edit the initial PR comment to describe exactly what was going wrong, and how this PR fixes it (or even better, open an issue and link it). Will make it easier to follow for anyone drifting in here with no context.

src/Power.cpp
akohlsmith: I was trying to get rid of the line wrapping; we've got some insanely long variable names that blow out the lines and then the wrapping makes everything harder to read. I'm trying to not change much but the voltage function was almost identical and immediately below the one I added.

todd-herbert: That's a pretty valid reason 👍

src/modules/Telemetry/Sensor/VoltageSensor.h
todd-herbert: Never mind, looks like you got it with that last commit!

src/Power.cpp
akohlsmith: > Definitely gotta sort this TODO first

got this addressed now. Was way easier than I thought at first glance.

src/Power.cpp
```diff
@@ -379,7 +385,15 @@ class AnalogBatteryLevel : public HasBatteryLevel
 #endif
 #ifdef EXT_CHRG_DETECT
         return digitalRead(EXT_CHRG_DETECT) == ext_chrg_detect_value;
+
+#elif HAS_TELEMETRY && !defined(ARCH_PORTDUINO) && !defined(HAS_PMU) && !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR
+        if (hasINA()) {
+            return (getINACurrent() > 0) ? OptTrue : OptFalse;
+        } else {
+            return OptFalse;
+        }
```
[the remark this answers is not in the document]
akohlsmith: I think the logic is sound; the larger code block looks like this;

```c
#elif HAS_TELEMETRY && !defined(ARCH_PORTDUINO) && !defined(HAS_PMU) && !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR
        if (hasINA()) {
            return (getINACurrent() > 0) ? OptTrue : OptFalse;
        } else {
            return OptFalse;
        }
#else

        return isBatteryConnect() && isVbusIn();
#endif
```

What I can do instead is remove both the `else` and the `#else` so that even if we SHOULD have an INA and don't, it still falls through and returns `isBatteryConnect() && isVbusIn()` -- I think I like that logic better, thanks for pointing that out!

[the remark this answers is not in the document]
todd-herbert: In the original version, I think most devices which do not have an INA sensor would have always returned OptFalse there, but looks like your changes have got it all sorted out now!

src/Power.cpp
```diff
@@ -379,9 +385,15 @@ class AnalogBatteryLevel : public HasBatteryLevel
 #endif
 #ifdef EXT_CHRG_DETECT
         return digitalRead(EXT_CHRG_DETECT) == ext_chrg_detect_value;
-#else
-        return isBatteryConnect() && isVbusIn();
+
+#elif HAS_TELEMETRY && !defined(ARCH_PORTDUINO) && !defined(HAS_PMU) && !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR
+        if (hasINA()) {
+            return (getINACurrent() > 0) ? OptTrue : OptFalse;
```
todd-herbert, line 391: One more thought actually. I've never used these current sensors, but is there a chance that some users have wired these with the current flowing the other way?

todd-herbert:
Just with a look through, no big issues are jumping out at me. I gave it a quick test and it doesn't seem to disrupt ADC readings or anything.

It does assume that users have the INA sensor wired in a specific configuration, yeah? 

I don't actually have any of these sensors to test it out so it'd be great for someone else to give it a run and a close look over!

warrenguy:
> The firmware can use the INA series of sensors for measuring battery voltage and current if it's detected, but the device metrics weren't using them. What was therefore happening was that while the power telemetry looked good, the device telemetry was just always reporting 101% (plugged in).

Anecdotally, I'm pretty sure the device metrics appeared to already use the correct (same as power metrics) value for me when using INA219 and (and since [a few days ago](https://github.com/meshtastic/firmware/commit/23ac6b65141914496ba108580cf0377f538db2e4) INA3221) sensors. I'll check that again more closely, and have a look at your PR here.

src/Power.cpp
warrenguy: > One more thought actually. I've never used these current sensors, but is there a chance that some users have wired these with the current flowing the other way?

I have these sensors wired to measure power usage by the device itself (rather than net current to/from the battery), so this value will always be greater than zero and so this new behaviour would be surprising...

However I also use the three channel INA3221 to measure the device voltage/current on one channel, and solar input on another channel. I've considered using the third channel to measure net flow to the battery, which I think is the assumption being made here for the expected wiring...

Perhaps a configuration item for "INA address and channel" for each of "device" and "battery" power makes sense? Or some configuration for the criteria used to determine charging/not charging... :shrug:

warrenguy: But the direct answer to your question is probably yes, I'm sure these get wired "wrong", and if you are only measuring current in one direction, you could just ignore the sign.

akohlsmith: this is incorrect; they don't care about direction -- I have wired mine such that negative current means the battery is discharging and positive current means it is charging. I spoke with @todd-herbert in Discord and will make a note of this in the source.

akohlsmith:
> Just with a look through, no big issues are jumping out at me. I gave it a quick test and it doesn't seem to disrupt ADC readings or anything.
> 
> It does assume that users have the INA sensor wired in a specific configuration, yeah?

Yes - it is assuming that the INA is wired in series with the battery positive lead, such that the more positive side of the measurement is facing the device, and the more negative side of the measurement is facing the battery positive lead. This leads to the concept of "negative current" representing the battery discharging and "positive current" representing the battery accepting charge.

src/Power.cpp
warrenguy: I don't completely follow, but this behaviour is definitely peculiar to a specific configuration which I don't believe to be a given...

todd-herbert: > I have these sensors wired to measure power usage by the device itself (rather than net current to/from the battery), so this value will always be greater than zero and so this new behaviour would be surprising...

This is the scenario I could imagine happening.

> But the direct answer to your question is probably yes, I'm sure these get wired "wrong", and if you are only measuring current in one direction, you could just ignore the sign.

Also the device would report as "powered" when discharging, and "on battery" when charging, I think?

todd-herbert:
> Just another thought, it'd probably be handy too to edit the initial PR comment to describe exactly what was going wrong, and how this PR fixes it (or even better, open an issue and link it). Will make it easier to follow for anyone drifting in here (me tomorrow, oops) with no context.

I still think even more detail here might helpful! An exact description of what you see happening right now, and what the PR would help clear up things like this:

>Anecdotally, I'm pretty sure the device metrics appeared to already use the correct (same as power metrics) value for me when using INA219 sensors, but I could be mistaken... I'll check that again more closely, and have a look at your PR here.

Now that there are two (potentially overlapping) PRs, I'd love to get a new issue opened (with "walk-me-though-it" level details of the problem). Hopefully we can catch the eye of there of some other devs who have worked with these INA sensors and get a bit more feedback in a general sense; and a consensus on how to go with this bug.

thebentern:
This PR has merge conflicts now based on a previously merged one.

Mictronics:
I wonder why the code needs to be more complex for that.
In actual code, the INA sensor is considered in Power.cpp for battery voltage reading under two conditions:
* BATTER_PIN must be defined
* the INA sensor address must be set in the apps power configuration

The working example is the rp2040-lora variant. This board has no onboard voltage divider for battery sensing, hence an external is needed through one of the ADC pins.
However, I have defined the BATTERY_PIN with a dummy value 0xFF to enable all the battery voltage reading code by preprocessor. And I have set the INA address in the power configuration.
The results is that the battery voltage&percentage is present in device metrics and voltage&current are in environment metrics.
In the existing code with that solution the plugged/unplugged indication is not working, but that fine I guess since the usage of an INA sensor may implies that external battery management is used.

This PR also creates a philosophy problem about the charge/discharge current. e.g. on my INA nodes the positive current indication means a battery discharge, while negative is charging the battery from solar.

IMHO, it looks like _I have a solution, let's find the problem._

akohlsmith:
> I wonder why the code needs to be more complex for that. In actual code, the INA sensor is considered in Power.cpp for battery voltage reading under two conditions:
> 
> * BATTER_PIN must be defined
> * the INA sensor address must be set in the apps power configuration

Why would `BATTERY_PIN` need to be defined if there is no voltage divider connected to an analog input pin?

> The working example is the rp2040-lora variant. This board has no onboard voltage divider for battery sensing, hence an external is needed through one of the ADC pins. However, I have defined the BATTERY_PIN with a dummy value 0xFF to enable all the battery voltage reading code by preprocessor. And I have set the INA address in the power configuration. The results is that the battery voltage&percentage is present in device metrics and voltage&current are in environment metrics. In the existing code with that solution the plugged/unplugged indication is not working, but that fine I guess since the usage of an INA sensor may implies that external battery management is used.

Is this dummy/magic value of 0xff mean something special to the code, or does having an invalid pin (as opposed to no definition or a value of -1 meaning "no pin") do something deeper in the code? It seems it would make a lot more sense to either not define `BATTERY_PIN` at all (similar to how you don't define the pin for an LED or user button if you don't have one) or to define its value as -1 which means no pin (not sure if it's a 32-bit or 8-bit value, 0xff could very well be interpreted as -1 if it's only an 8 bit value). I'm fine with either method but prefer the former as it is logical. The code is a hodgepodge of these odd little exceptions and it'd be nice to try to nudge it more toward a consistent way to do things.

> This PR also creates a philosophy problem about the charge/discharge current. e.g. on my INA nodes the positive current indication means a battery discharge, while negative is charging the battery from solar.

Sure, but short of adding another configuration option (and trying to get it into the phone apps, python utility and web client) it seems like there isn't a good way to define this, and I'd really, really prefer not to have another compile-time option for this. I chose "current flow from battery point of view" as it seemed logical to me, but do understand it can be logical in the other direction for other people with other applications. I'd prefer a config option but it seems that might be useful in a second PR to address it rather than making this one larger.

> IMHO, it looks like _I have a solution, let's find the problem._

Agreed; once I understand your comment more thoroughly I will try to do the same and see if I can clean this up significantly.

Mictronics:
> > I wonder why the code needs to be more complex for that. In actual code, the INA sensor is considered in Power.cpp for battery voltage reading under two conditions:
> > 
> > * BATTER_PIN must be defined
> > * the INA sensor address must be set in the apps power configuration
> 
> Why would `BATTERY_PIN` need to be defined if there is no voltage divider connected to an analog input pin?
> 
BATTERY_PIN will enable the entire analog battery voltage reading code via preprocessor. Without the power status will always report back 101% and never reads anything.

> > The working example is the rp2040-lora variant. This board has no onboard voltage divider for battery sensing, hence an external is needed through one of the ADC pins. However, I have defined the BATTERY_PIN with a dummy value 0xFF to enable all the battery voltage reading code by preprocessor. And I have set the INA address in the power configuration. The results is that the battery voltage&percentage is present in device metrics and voltage&current are in environment metrics. In the existing code with that solution the plugged/unplugged indication is not working, but that fine I guess since the usage of an INA sensor may implies that external battery management is used.
> 
> Is this dummy/magic value of 0xff mean something special to the code, or does having an invalid pin (as opposed to no definition or a value of -1 meaning "no pin") do something deeper in the code? It seems it would make a lot more sense to either not define `BATTERY_PIN` at all (similar to how you don't define the pin for an LED or user button if you don't have one) or to define its value as -1 which means no pin (not sure if it's a 32-bit or 8-bit value, 0xff could very well be interpreted as -1 if it's only an 8 bit value). I'm fine with either method but prefer the former as it is logical. The code is a hodgepodge of these odd little exceptions and it'd be nice to try to nudge it more toward a consistent way to do things.
> 

0xFF is like RADIOLIB_NC which is 32bit, but for ordinary GPIO an 8bit value is required. The value itself doesn't matter much, the preprocessor define is needed.
See Power.cpp, Power::analogInit(), the entire reading of the battery level is guarded by #ifdef BATTERY_PIN. 
```
#ifdef BATTERY_PIN
.
.
.
    batteryLevel = &analogLevel;
    return true;
#else
    return false;
#endif
```
If not defined the variable  batteryLevel is not initialized, hence the sensor doesn't exist.
`static HasBatteryLevel *batteryLevel; // Default to NULL for no battery level sensor`
See rp2040-lora variant for example.

> > This PR also creates a philosophy problem about the charge/discharge current. e.g. on my INA nodes the positive current indication means a battery discharge, while negative is charging the battery from solar.
> 
> Sure, but short of adding another configuration option (and trying to get it into the phone apps, python utility and web client) it seems like there isn't a good way to define this, and I'd really, really prefer not to have another compile-time option for this. I chose "current flow from battery point of view" as it seemed logical to me, but do understand it can be logical in the other direction for other people with other applications. I'd prefer a config option but it seems that might be useful in a second PR to address it rather than making this one larger.
> 

Being an electronics engineer, I learned more than 30 years ago that a current drawn by a load has a positive sign. And I consider the node hardware as being the load on the battery.
From my childhood I remember that analog dash ampere meter in cars. A needle swing to the right, positive scale, was discharging the battery, while a left swing, negative scale, was charging. But world has changed since.

> > IMHO, it looks like _I have a solution, let's find the problem._
> 
> Agreed; once I understand your comment more thoroughly I will try to do the same and see if I can clean this up significantly.

PR #4215, opened 2024-06-30
INA3221 sensor: use for bus voltage & environment metrics

warrenguy:
fixes bugs:

- ina3221 not used for device or environment telemetry
  - fix: add to sensors available for [environment](https://github.com/meshtastic/firmware/pull/4215/files#diff-1ddd840d2edd6b3dc5912e2743c19f142810af4a4fad75b846acd5293323f02eR326) and [device](https://github.com/meshtastic/firmware/pull/4215/files#diff-27c64c64eb25fee7c89f6a5c1251b4254e28ecf8d4959ccd2a866e12516435b3R444-R448) telemetry
- ina3221 getPower method overwrites environment telemetry
  - fix: write to the [correct fields](https://github.com/meshtastic/firmware/pull/4215/files#diff-6450b2eef36c3aeb8cf57c4950b77580681bec46c61ffe90f37a8321b22b2571R54-R60) on the measurement struct depending on the measurement variant

small improvements to the ina3221 implementation:

- add vars to define channels to use for battery voltage (for getBusVoltage) and environment metrics (default to CH1 for both)
- DRY up the sensor measurement collection code a bit


thebentern:
Has this been tested against power telemetry and environmental telemetry running simultaneously?

warrenguy:
> Has this been tested against power telemetry and environmental telemetry running simultaneously?

yes, I have been running this branch with both enabled and it's working as expected

warrenguy:
Note, the fix herein for _"ina3221 getPower method overwrites environment telemetry"_ is particular to the INA3221, due to it being a 3-channel sensor.

The other supported INA sensors have a single channel. These write to `measurement->variant.environment_metrics.{voltage|current}` for both environment and power metrics calls, which happens to work only because the `voltage` and `current` members of the `environment_metrics` variant happen to align exactly with `ch3_voltage` and `ch3_current` members of the `power_metrics` variant.

This seems like good fortune rather than intention, but it does explain why the e.g. INA219 uses channel 3 for its only channel, which I was curious about.

I intend to do some more work on the power metrics and INA sensors, so I'll tidy up the above and make it more explicit separately.

thebentern (approved):

PR #4233, opened 2024-07-05
Fix exclude macros

mverch67:
#3469 and  #3337 introduced macros to exclude modules or to minimize the build.

In the meantime some of these macros are broken and lead to compiler/linker errors when in use.

This PR fixes these compiler/linker errors.


mverch67:
As this may happen again and again when merging PRs I'd like to suggest to include a dummy target into the build matrix with a define e.g. `MESHTASTIC_MINIMIZE_BUILD=1`
so that there is a basic automated process to verify the correctness of the build exclusion macros.

Talie5in:
I thought I caught most of these when doing #4049

Nice catches @mverch67

thebentern (approved):

PR #4235, opened 2024-07-05
fix python warning in uf2conf

geeksville:
the old regex worked but was technically incorrect.  fixes: Generating NRF52 uf2 file
/home/kevinh/development/meshtastic/firmware/bin/uf2conv.py:195: SyntaxWarning: invalid escape sequence '\s'
  words = re.split('\s+', line)
Converting to uf2, output size: 1458688, start address: 0x26000

### DIFF
--- a/bin/uf2conv.py
+++ b/bin/uf2conv.py
@@ -1,39 +1,38 @@
 #!/usr/bin/env python3
-import sys
-import struct
-import subprocess
-import re
+import argparse
 import os
 import os.path
-import argparse
+import re
+import struct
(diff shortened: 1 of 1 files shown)


geeksville:
The only real change is

> words = re.split("\\s+", line)
instead of \s.

Alas, I'm tired of fighting with trunk.  It keeps wanting to reformat this python file per our coding conventions.  So I guess we should just let it do that (because we'll have to do that eventually anyways).

thebentern (approved):

PR #4239, opened 2024-07-06
stm32wl: add AES256 assert for it's cryptoengine

Jorropo:
This impl always use AES256 to encrypt packets, assert this is the case. Tbh I think this could be moved to CrossPlatformCryptoEngine.cpp.


GUVWAF (approved):

caveman99:
This is pointless, as it would only cause merge conflicts when pulling in the ongoing STM32WL work. The crypto engine has been replaced to use rweather a long time ago. Please refer to the stm32wl branch instead of master for this.

PR #4244, opened 2024-07-07
Fix BLE logging on nrf52

geeksville:
Two changes - the first one is not required (though it will be useful in the future, because I'm about to add an option to the python client to use BLE indicate instead of notify - for lossless log message transmission).  But the second one is an oops I missed in the review of the LogRecord changes.  See the commit for a description.


geeksville:
also at some point I'll probably send in a PR to unify the two subclasses that implement BluetoothApi.  Currently they are related but rather than having everyone reach BLE land by touching something like bluetooth->.  We have lots of #ifdef that chooses nrf52Bluetooth or esp32Bluetooth.

Eventually it would be good to remove those ifdefs and use inheritance/polymorphism to do the right thing?

thebentern (approved):

PR #4245, opened 2024-07-07
Transition to STATE_SEND_COMPLETE_ID instead

thebentern:
Make STATE_SEND_FILEMANIFEST behave like STATE_SEND_OTHER_NODEINFOS does


thebentern:
Issue with fromnum actually. Closing

PR #4148, opened 2024-06-20
Moar LR1110 Targets

caveman99:

PR #4252, opened 2024-07-08
Cleanup NRF s140 Softdevice variants

fifieldt:
Note: This idea is originally from @caveman99 and should be credited as such. Submitting as a separate PR so the work in meshtastic/firmware#4148 can be a bit cleaner and Seeed boards can build while that work is ongoing.

The nrf52 boards that depend on the v7 softdevice all use the same code and linker files. Rather than duplicate the code, keep it all together with the platform.


markbirss:
So this make more sense, my bunch of PR's is duplicating files

#4248

#4253 

#4254 

@fifieldt we should look at doing the same for the xiao\_ble target

fifieldt:
Thanks for checking @markbirss . xiao_ble is already changed to this style in this patch, I believe?

markbirss:
> Thanks for checking @markbirss . xiao\_ble is already changed to this style in this patch, I believe?

Ok, we also need to remove some files from that variant folder ?

fifieldt:
@markbirss - this patch has a 

`git mv variants/xiao_ble/softdevice src/platform/nrf52/`

 ... so I believe (hope!) those files will magically disappear after it merges :)

thebentern (approved):

PR #4258, opened 2024-07-09
Add wio-sdk-wm1110 to build.

fifieldt:
The wio-sdk-wm1110 is distinct from the wio-tracker-wm1110, with different platformio build options and pin config.

This change adds the wio-sdk-wm1110 to the CI matrix so firmware is built as part of release.


thebentern:
I also need to make sure the hex from our previous build-nrf52.sh makes it into the release assets so that folks can flash from SWD.

PR #4261, opened 2024-07-10
Fix error when compiling with "#define MESHTASTIC_EXCLUDE_SCREEN 1"

tuxphone:
Fix compile error when excluding screen (#define MESHTASTIC_EXCLUDE_SCREEN 1) on target nrf52

NRF52Bluetooth::onPairingPasskey() tries to create and write to OLEDDisplay *display, which is not defined when excluding the screen code.

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -94,7 +94,11 @@ int MeshService::handleFromRadio(const meshtastic_MeshPacket *mp)
     } else if (mp->which_payload_variant == meshtastic_MeshPacket_decoded_tag && !nodeDB->getMeshNode(mp->from)->has_user &&
                nodeInfoModule) {
         LOG_INFO("Heard a node on channel %d we don't know, sending NodeInfo and asking for a response.\n", mp->channel);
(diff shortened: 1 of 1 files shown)


thebentern:
Looks good. Can you trunk fmt to fix the formatting issues?

src/platform/nrf52/NRF52Bluetooth.cpp
```diff
@@ -295,6 +295,7 @@ bool NRF52Bluetooth::onPairingPasskey(uint16_t conn_handle, uint8_t const passke
 {
     LOG_INFO("BLE pairing process started with passkey %.3s %.3s\n", passkey, passkey + 3);
     powerFSM.trigger(EVENT_BLUETOOTH_PAIR);
+    #ifndef MESHTASTIC_EXCLUDE_SCREEN
```
mverch67, line 298: shouldn't this read `#if !MESHTASTIC_EXCLUDE_SCREEN`

thebentern: Good call.

PR #4273, opened 2024-07-11
Don't send node info interrogation when ch. util is >25%

thebentern:
Another attempt to curb excessive non-text traffic


GUVWAF (changes requested):
I think we should move this to the request instead of the reply, otherwise a node will keep asking and not get a response. Maybe cap the request at 25% and keep this at 40% so you don't get in this endless loop.
The request is [here](https://github.com/meshtastic/firmware/blob/eabec5ae3451c0edc33979453c8af635d538c538/src/mesh/MeshService.cpp#L94). (Probably better to move this to `NodeDB::updateFrom()`?)

thebentern:
> I think we should move this to the request instead of the reply, otherwise a node will keep asking and not get a response. Maybe cap the request at 25% and keep this at 40% so you don't get in this endless loop.

Good idea. The send wantresponse is probably the more valuable one to snub on a congested mesh anyway

GUVWAF (approved):

PR #4274, opened 2024-07-11
Move up telemetry defaults to every 30 minutes

thebentern:
Docs will need to be updated, but I think this is a positive change as networks grow


GUVWAF (approved):
One other caveat: both iOS and Android consider a node "online" if it was heard in the last 15 min. When increasing this and if you don't have position or other periodic broadcasts, a node will bounce between "online" and "offline" all the time.

thebentern:
> One other caveat: both iOS and Android consider a node "online" if it was heard in the last 15 min. When increasing this and if you don't have position or other periodic broadcasts, a node will bounce between "online" and "offline" all the time.

Yes, at this point I believe the only default to 15 minutes of the core traffic would be positions.

PR #4280, opened 2024-07-13
Fix for serial monitoring and I2C for Vision Master e213

HarukiToreda:
The board did not allow serial monitoring while on boot mode, i was able to fix this by adding a board variant. I also corrected the i2c pins. I was able to test it with a cardkb


HarukiToreda:
Still known issues, 

- When it cold boots it doesn't scan the i2c fast enough and doesn't recognize the keyboard. Needs to be soft  rebooted
- When on flash mode (Middle button+ boot button) it has a different serial port number than when powered on on my windows computer. Not sure if there's a fix for that.

boards/heltec_vision_master_e213.json
```diff
@@ -0,0 +1,39 @@
...
+      "-DHELTEC_VISION_MASTER_E213",
+      "-DARDUINO_USB_CDC_ON_BOOT=1",
+      "-DARDUINO_USB_MODE=0",
+      "-DARDUINO_RUNNING_CORE=1",
+      "-DARDUINO_EVENT_RUNNING_CORE=1"
+    ],
+    "f_cpu": "240000000L",
+    "f_flash": "80000000L",
+    "flash_mode": "qio",
+    "hwids": [["0x303A", "0x1001"]],
```
todd-herbert, line 19: >When on flash mode (Middle button+ boot button) it has a different serial port number than when powered on on my windows computer. Not sure if there's a fix for that.

I'm experimenting with use the two different hardware ids for this device. If nothing else, it seems to make the PlatformIO serial monitor find the second port correctly.
```
      "hwids": [
        [
          "0x303A",
          "0x1001"
        ],
        [
          "0x303A",
          "0x0002"
        ]
````

I'd like to play with this a little bit more on my end and see if we can get an even better solution.

File: boards/heltec_vision_master_e213.json [hunk 2]
```diff
@@ -0,0 +1,39 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "esp32s3_out.ld",
+      "partitions": "default_8MB.csv"
+    },
+    "core": "esp32",
+    "extra_flags": [
+      "-DBOARD_HAS_PSRAM",
+      "-DHELTEC_VISION_MASTER_E213",
```
todd-herbert, line 10: I think maybe(?) we should just remove `"-DHELTEC_VISION_MASTER_E213"` from here, just because it looks like the other boards.json files leave this definition to variant.h

variants/heltec_vision_master_e213/variant.h
```diff
@@ -1,8 +1,13 @@
...
 // #define LED_PIN 18
 
+#ifdef HELTEC_VISION_MASTER_E213
+#define HELTEC_VISION_MASTER_E213 true
+
 // Enable bus for external periherals
-#define I2C_SDA SDA
-#define I2C_SCL SCL
+#define I2C_SCL 38
+#define I2C_SDA 39
```
todd-herbert, line 10: Were these working when defined as `SDA` and `SCL`, once you fixed pins_arduino.h?

variants/heltec_vision_master_e213/variant.h
todd-herbert, line 1: I'm keen to rearrange this whole file so it's the same as the Wireless Paper variant.h files, although that's honestly just housekeeping. I'm more than happy to take care of that, if you're okay with it?

todd-herbert (commented):
> When it cold boots it doesn't scan the i2c fast enough and doesn't recognize the keyboard. Needs to be soft rebooted

Worst case scenario, we might be able to add a *small* delay at boot, just for this board (+other vision master boards?). I'm still hoping we can figure out something that fixes the Serial and this I2C issue all-at-once. I'm looking at it today!

Do see if you can get [trunk](https://docs.trunk.io/check/advanced-setup/cli) set up, so you don't have to battle this one every time!
image

variants/heltec_vision_master_e213/variant.h
HarukiToreda: Ill test it

variants/heltec_vision_master_e213/variant.h
HarukiToreda: that'd be very helpfull, thank you

variants/heltec_vision_master_e213/variant.h
HarukiToreda: yeah it works after I changed this back to sda

boards/heltec_vision_master_e213.json
HarukiToreda: removed

HarukiToreda:
I2c scans properly now on boot.

PR #4281, opened 2024-07-13
Fix error when compiling with "#define MESHTASTIC_EXCLUDE_SCREEN 1"

tuxphone:
Fix compile error when excluding screen (#define MESHTASTIC_EXCLUDE_SCREEN 1) on target nrf52

NRF52Bluetooth::onPairingPasskey() tries to create and write to OLEDDisplay *display, which is not defined when excluding the screen code.


tuxphone:
My trunk fmt sees no issues

PR #4282, opened 2024-07-13
Let StoreForward server send history to phoneAPI

GUVWAF:
This allows a node configured as StoreForward server to use its PSRAM for an extended text message history (max. ~11,000) for sending to a connected client app. When connecting to the server, you automatically get the packets you missed since the last connection. Text messages will not be placed in the `toPhoneQueue` anymore to avoid duplicates and have more room for other packets.

It ended up being quite a redesign of the StoreForwardModule to remove the `packetHistoryTXQueue` which could become quite large and was actually just a filtered duplicate. Instead, we just loop through the `packetHistory` itself.

I also automatically set the new `is_server` configuration to true for nodes that were `ROUTER_CLIENT` and had the StoreForwardModule enabled, because that was the previous behavior. If that's not desired I can remove it.


thebentern (approved):

thebentern:
Super cool!

garthvh:
Can we make it not function on the default key?

GUVWAF:
Okay, I can add that restriction for the request over LoRa. Would need to be added to the docs as well then.

GUVWAF:
Latest commit adds the restriction.

thebentern (approved):

thebentern:
Looks good to me

PR #4289, opened 2024-07-15
Align NRF52 and ESP32 BLE behavior

todd-herbert:
Disables BLE on NRF52 targets with router role


thebentern:
I don't think this is in line with the bluetooth behavior on esp32. ESP32 only controls the shutting down of Bluetooth via the PowerFSM states. Take a look at the isRouter states transitions to `&stateNB` and particularly the guard in `nbEnter`. 

I think we'll need to test this thoroughly or we could introduce bluetooth behaviors in all roles though. I would actually place a guard in that nbEnter to ensure it only affects ROUTERs and REPEATERs to head off those issues.

thebentern:
> In the context of router_client depreciation, do you think that this PR is important to try rework somehow, or are you happy to leave it for now? If it is important, do you have a preferred way of implementing it?

I would say we're not in an immediate rush to get it out there right now, but I definitely would prioritize consistency between the two platforms if we're going to shutdown BLE on Routers.

todd-herbert:
All good, I think I definitely just got the wrong idea about the router client stuff somehow.

PR #4294, opened 2024-07-15
Migrate to new defaults

thebentern:


src/mesh/NodeDB.cpp
```diff
@@ -648,6 +648,26 @@ void NodeDB::loadFromDisk()
     if (state == LoadFileResult::SUCCESS) {
         LOG_INFO("Loaded OEMStore\n");
     }
+
+    // 2.4.X - configuration migration to update new default intervals
+    if (moduleConfig.version < DEVICESTATE_CUR_VER) {
```
todd-herbert, line 653: I wonder if it might be good to hard-code this as `if(moduleConfig.version < 23)`, to prevent it running again in the future, when `DEVICESTATE_CUR_VER` increments again? Optional though!

todd-herbert (approved):
I haven't built it for a test run, but I can in a few hours if it helps.

From a read through it looks good!
* module config *won't* be erased, because `DEVICESTATE_MIN_VER` remains unchanged
* module config *will* be updated on first boot, because `DEVICE_STATE_CUR_VER` has increased
* module config *won't* be updated on every boot, because `moduleConfig.version` is increased to `DEVICE_STATE_CUR_VER` after the initial update

src/mesh/NodeDB.cpp
jp-bennett: Agree. Real numbers are the way to go.

thebentern: The one time mama said not to use a constant! Good call, guys. 😆

todd-herbert (changes requested):
I just gave this a run on T-Beam and caught something unexpected.
It seems like [`NodeDB::initModuleConfigIntervals`](https://github.com/meshtastic/firmware/blob/0f5fdfbab780e3dcdb3a957efd142d9440f52cf3/src/mesh/NodeDB.cpp#L426-L432) is using [default_broadcast_interval_secs](https://github.com/meshtastic/firmware/blob/master/src/mesh/Default.h#L9), which is defined as 900 seconds.

thebentern:
> I just gave this a run on T-Beam and caught something unexpected. It seems like [`NodeDB::initModuleConfigIntervals`](https://github.com/meshtastic/firmware/blob/0f5fdfbab780e3dcdb3a957efd142d9440f52cf3/src/mesh/NodeDB.cpp#L426-L432) is using [default_broadcast_interval_secs](https://github.com/meshtastic/firmware/blob/master/src/mesh/Default.h#L9), which is defined as 900 seconds.

Wowie! How did I miss that land mine?! 
I think I'll zero those out so that they coalesce into my getConfiguredOrDefault() so we can move the defaults around again on those who don't set it to a different value.

todd-herbert (approved):

PR #4296, opened 2024-07-16
Serial Mode for Ecowitt WS85 weather station.

tavdog:
This PR adds a new SerialModule mode for the WS85 ultrasonic anemometer / weather station.  Attaching the station via a serial connection allows for Meshtastic to read the weather data and transmit it at 5 minute intervals as Telemetry data to the mesh.  

To enable this mode you must use the cli to issue `meshtastic --set serial.mode 6` as this mode is not present in any of the other clients at the time of this PR.


src/modules/SerialModule.cpp
```diff
@@ -198,8 +199,157 @@ int32_t SerialModule::runOnce()
...
+                }
+                if (gotwind) {
+
+                    LOG_INFO("WS85 : %i %.1fg%.1f %.1fv %.1fv\n", atoi(windDir), strtof(windVel, nullptr),
+                             strtof(windGust, nullptr), batVoltageF, capVoltageF);
+                }
+                if (gotwind && millis() - lastAveraged > averageIntervalMillis) {
+                    // calulate average and send to the mesh
+                    float velAvg = 1.0 * velSum / velCount;
+                    // float dirAvg = dirSum / dirCount; // TODO : this needs some trig to be accurate
```
thebentern, line 315: Is this comment still accurate?

src/modules/SerialModule.cpp
```diff
@@ -198,8 +199,157 @@ int32_t SerialModule::runOnce()
                     }
                 }
             }
+
 #if !defined(TTGO_T_ECHO) && !defined(CANARYONE)
-            else {
+            else if ((moduleConfig.serial.mode == meshtastic_ModuleConfig_SerialConfig_Serial_Mode_WS85)) {
+                static unsigned int lastAveraged = 0;
```
thebentern, line 205: Given the extensive parsing logic and math involved, I wonder if we should make a void function to extract this block to, so that it doesn't bloat runOnce()

tavdog: Sure I can do that.  All that logic is pretty confusing.

src/modules/SerialModule.cpp
tavdog: The TODO comment is in fact old. I'll remove it.

thebentern (approved):

mikechristiansen:
@tavdog Love your work here. Just setting up a WS85 meshtastic station. 
I see the serial data is reporting rain. Any idea how much work it would be to report this over the mesh? I guess this would require changes in the enviromental protobuffs? 
Cheers, Mike.

tavdog:
> @tavdog Any idea how much work it would be to report this over the mesh? I guess this would require changes in the enviromental protobuffs? Cheers, Mike.

Apparently the station reports RainIntSum and Rain values and I'm not sure what exactly that Sum interval is.  But yes it would require adding those two items to the protobufs if we wanted to include the rain values.

PR #4313, opened 2024-07-21
Just a bit of security hygiene. 

jp-bennett:


src/platform/esp32/main-esp32.cpp
```diff
@@ -93,6 +93,7 @@ void esp32Setup()
 {
     uint32_t seed = esp_random();
     LOG_DEBUG("Setting random seed %u\n", seed);
+    randomSeed(seed);
```
GUVWAF, line 96: I remember that setting the seed for ESP32 was removed on purpose: #2374

caveman99: Spot on. Maybe leave it in, commented out with an annotation to NOT set that again :-)

caveman99: I would even suggest to remove the seed call too. the variable is not used except for the (now removed) call.

jp-bennett: Absolutely, because without that explanation, it looks like *not* making the call to randomSeed() is what breaks the random generator.

caveman99 (approved):

thebentern (approved):

PR #4317, opened 2024-07-22
Fix de/compression buffer overflows in TAK packets

### LINKED ISSUE #3841: [Bug]: Remotely reachable stack buffer overflow due to the use of unishox2_decompress_simple
### Category

Other

### Hardware

Not Applicable

### Firmware Version

2.3.7.30fbcab

### Description

There is a stack buffer overflow when decompressing ATAK messages here:

https://github.com/meshtastic/firmware/blob/75dc8cccecd52da78f1d69eeb4eb017fbc68f26c/src/modules/AtakPluginModule.cpp#L114
https://github.com/meshtastic/firmware/blob/75dc8cccecd52da78f1d69eeb4eb017fbc68f26c/src/modules/AtakPluginModule.cpp#L119
https://github.com/meshtastic/firmware/blob/75dc8cccecd52da78f1d69eeb4eb017fbc68f26c/src/modules/AtakPluginModule.cpp#L126
https://github.com/meshtastic/firmware/blob/75dc8cccecd52da78f1d69eeb4eb017fbc68f26c/src/modules/AtakPluginModule.cpp#L133

The function `unishox2_decompress_simple` does not take the size of the output buffer into account, always assuming it has enough space for the result. Therefore, if an ATAK message contains a compressed field that decompresses into more bytes than is reserved in a `meshtastic_TAKPacket`, then this will write out of the bounds of that `uncompressed` field, overwriting other values on the stack (potentially even a pushed return address).

This is actually the same issue as what triggered #3573, except there it was _compression_ that triggered the overflow. The user who reported it sent a base64-encoded message near the maximum size. Due to the high entropy of the message the unishox2 compression actually increased its length from 220 to 245 bytes, also causing an out of bounds write on the stack. Before #3606 was merged it looks like decompression there was affected too with received messages.

The correct function to call would be `unishox2_decompress`/`unishox2_compress` and making sure that `UNISHOX_API_WITH_OUTPUT_LEN` is defined as 1. (#3606 disabled the use of unishox2 for "app" messages, but I'm not sure if not doing the same in the ATAK plugin was an oversight.)

This is a vulnerability that could be exploited by sending a malicious message to a Meshtastic node (the ATAK plugin does not need to actively used just sending a message of this type on a channel the node is on is enough). In the firmware I looked at (the LilyGo T3-S3), this would not be directly exploitable to gain RCE due to the use of stack cookies, but I don't know if that's the case for all supported hardware. Even without knowing the stack cookie value, this can be used to repeatedly crash a node.

### Relevant log output

_No response_

thebentern:
Closes #3841


caveman99 (approved):

src/mesh/compression/unishox2.h
```diff
@@ -373,7 +331,7 @@ extern int unishox2_compress_lines(const char *in, int len, UNISHOX_API_OUT_AND_
                                    const char *usx_freq_seq[], const char *usx_templates[], struct us_lnk_lst *prev_lines);
 /**
  * More Comprehensive API for de-compressing array of strings \n
- * This function is not be used in conjunction with unishox2_compress_lines()
+ * This function is not be used in conjuction with unishox2_compress_lines()
```
gretel, line 334: think the speling was right before?

PR #4320, opened 2024-07-23
[WIP] Add Configuration support for Air350

fifieldt:
**Note: I haven't managed to get this working in testing.** Sharing here for visibility, in the hope that someone who knows more can use it.

The Airoha Air350 GNSS Chip is a part of several Seeed boards. At present, it is auto-detected as a L76K chip (GNSS_MODEL_MTK), which works. However, this means the chip is not fully configured and power management is probably not optimised.

This patch adds code to use the version information command to automatically detect the Airoha chips.


variants/wio-tracker-wm1110/variant.h
```diff
@@ -100,6 +100,25 @@ extern "C" {
 
 #define LR1110_GNSS_ANT_PIN (32 + 5) // P1.05 37
 
+#define HAS_GPS 1
+#define GNSS_Airoha
```
thebentern, line 104: Let's do GNSS_AIROHA to keep our macros all upper

fifieldt:
Discovery: the Air530 I received is actually a Air530z, which has different instructions: https://doc.openluat.com/article/2296/0

fifieldt:
Going to start over.

PR #4303, opened 2024-07-19
Support Seeed Tracker-T1000-E

Dylanliacc:
#### Summary
This PR introduces support for the T1000-E board, integrating the T1000xSensor and incorporating various customizations and adaptations specific to the board's features. Below are the detailed changes and additions:

#### Board Support: T1000-E
- **Added support for the T1000-E board:** The new configuration files and settings tailored to the T1000-E board have been included, ensuring compatibility and proper functioning with the Meshtastic system.

#### Sensor Integration: T1000xSensor
- **Integrated T1000xSensor:** The T1000xSensor has been added to the project, providing enhanced sensor capabilities and ensuring it works seamlessly with the T1000-E board.

#### ButtonThread Customization
- **Macro Definitions for T1000-E:** Specific macro definitions have been introduced within the ButtonThread to leverage the unique features of the T1000-E board. These definitions optimize button handling and ensure smooth operation tailored to the T1000-E's hardware.

#### Customization for LR1110 in LR11x0Interface.cpp
- **DIO Pin Configuration:** The LR11x0Interface.cpp file has been customized based on the DIO pin configuration of the T1000-E's LORALR1110. These modifications ensure proper communication and functionality of the LR1110 module on the T1000-E board.

#### GPS Adaptation: AG3335 Sensor
- **Adaptation for AG3335 Sensor:** The GPS component has been adapted to support the AG3335 sensor, ensuring accurate and reliable GPS data acquisition. This integration improves the overall GPS functionality of the T1000-E board within the Meshtastic system.

#### Future Plans and Improvements
- **Class-based GPS Implementation:** In future updates, I plan to implement a dedicated class for the GPS functionality instead of using macro definitions. This will enhance code modularity, readability, and maintainability.

#### Draft PR and Request for Assistance
This is my first contribution to the Meshtastic project, and I am submitting this as a draft PR. There may be areas that need improvement, and I kindly ask for your patience and understanding as I navigate through this process. Any assistance or feedback from the community would be greatly appreciated to help refine and enhance this implementation.

Thank you for considering this PR and for any help or guidance you can provide.

Best regards,

PR #4328, opened 2024-07-25
Fix T1000-E GPS - some changes went missing from #4303?

markbirss:
Some changes from @Dylanliacc T1000-E branch went missing in the merge it seems, and GPS is not detected on master

https://github.com/meshtastic/firmware/pull/4303

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -400,14 +400,15 @@ bool GPS::setup()
     int msglen = 0;
 
     if (!didSerialInit) {
+#ifdef GNSS_AIROHA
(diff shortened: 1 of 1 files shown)


thebentern (approved):

fifieldt:
@markbirss - this was because of #4319 .

The extra baud rate code probably isn't needed anymore, see a few lines down - should automatically get your 115200 since it's set in variant.h.

The GPS probably broke because it's not a UC6580, which we now have auto-detection code for.

markbirss:
> @markbirss - this was because of #4319 .
> 
> The extra baud rate code probably isn't needed anymore, see a few lines down - should automatically get your 115200 since it's set in variant.h.

Ok, I see, I may have not looked at it long enough 😳

Issue #3984, opened 2024-05-28
[Feature Request]: Use PDOP from GPS devices to determine if we have a good enough position to save airtime 


garthvh:
### Platform

NRF52, ESP32, RP2040, Linux Native

### Description

We are currently not using the PDOP value to throw out bad position values causing smart position to use a lot of airtime while the unit is stationary.

garthvh:
![image](https://github.com/meshtastic/firmware/assets/1795163/8bef04e7-3c38-4e30-b26e-fc8ace3e826a)
Example image from @pdxlocations of a stationary vehicle generating 10K position packets with smart position on

fifieldt:
@GPSFan , do you happen to have numbers for what makes a 'bad' DOP we could throw out?

GPSFan:
DOP is a measure of how poor the geometry of the sats used in the fix is. Bad DOP means bad location accuracy, but you can have good DOP and a lousy fix. What constitutes bad DOP varies by whom you ask and in what context. For a surveyor, anything less than 1.5 is BAD. For a hiker 3-4 might be fine, as any position is better than no position. A good view of the sky should get you good DOP.  In a canyon, the DOP might drop to 10+. Do you want to throw out the only position you have even if it is poor? Columbus would have killed for 100km accuracy.

In any case, Meshtastic only uses the GGA and RMC sentences so HDOP is all we get. Unfortunately the GSA sentence that contains the other DOPS can overload the serial input, without providing any thing more useful. In u-blox receivers its' firmware follows the NMEA spec and will only out GSA data for 12 sats. Most other receivers don't follow the spec and will output GSA data for as many sats as they are tracking. My Wireless Tracker was tracking 32 sats this morning and while it might be interesting for some users which sats were tracked, most don't care. 32 sats worth of GSA data is a lot, to deal with for only PDOP, HDOP and VDOP.

In the normal use case, Meshtastic does not allow the receiver to acquire more sats than necessary for the initial fix. This can lead to bad DOP even with a good sky view. Over time, the receiver may be able to acquire more sats. 
The GPS (u-blox) is then put to sleep for a while to save power. Receiver power can be switched off as well, in both cases the receiver will have to re-acquire a fix after it wake up. This is usually quite fast, but that leaves little  time to acquire new sats.

fifieldt:
Thank you so much for the most excellent explanation,@GPSFan!


I've been tracking some of the position data from my local mesh recently. Some examples below looking at 1 week of data.


[...]

PR #4342, opened 2024-07-28
Allow RTC to auto-fix in case of noisy neighbour with delayed clock.

eljojo:
### Problem

There's a node misconfigured in my area, it's broadcasting every 10 seconds a position packet with timestamp set to two weeks in the past.

Notice time of `1720219611` two weeks ago.
```
INFO  | ??:??:?? 6 [Router] Received position from=0xb01eb761, id=0x6adafb99, portnum=3, payloadlen=17
DEBUG | ??:??:?? 6 [Router] POSITION node=b01eb761 l=17 lat=454006399 lon=-756663217 msl=66 hae=0 geo=0 pdop=0 hdop=0 vdop=0 siv=0 fxq=0 fxt=0 pts=0 time=1720219611
DEBUG | ??:??:?? 6 [Router] Upgrading time to quality Net
DEBUG | 18:46:51 6 [Router] Read RTC time as 1720219611
INFO  | 18:46:51 6 [Router] updatePosition REMOTE node=0xb01eb761 time=1720219611 lat=454006399 lon=-756663217
DEBUG | 18:46:51 7 [Router] Node status update: 38 online, 40 total
```

### Proposed fix
If a new position packet comes from the net, which is at least 3 days ahead of our current time, we'll jump-forward to that new time.

```
DEBUG | 18:47:48 64 [Router] POSITION node=7f1e7692 l=30 lat=454098944 lon=-757268480 msl=69 hae=0 geo=0 pdop=112 hdop=0 vdop=0 siv=16 fxq=0 fxt=0 pts=1722156207 time=1722165205
INFO  | 18:47:48 64 [Router] Reapplying external time to correct clock drift 1722165205 secs. A neighbour likely has a broken clock.
DEBUG | 18:47:48 64 [Router] tv_sec=1722165205, now=1720219611, diff=0
DEBUG | 07:13:25 64 [Router] Read RTC time as 1722165205
INFO  | 07:13:25 64 [Router] updatePosition REMOTE node=0x7f1e7692 time=1722165205 lat=454098944 lon=-757268480
DEBUG | 07:13:25 64 [Router] Node status update: 0 online, 40 total
```

### DIFF
--- a/src/gps/RTC.cpp
+++ b/src/gps/RTC.cpp
@@ -113,6 +113,11 @@ bool perhapsSetRTC(RTCQuality q, const struct timeval *tv, bool forceUpdate)
         // Every 12 hrs we will slam in a new GPS or Phone GPS / NTP time, to correct for local RTC clock drift
         shouldSet = true;
         LOG_DEBUG("Reapplying external time to correct clock drift %ld secs\n", tv->tv_sec);
+    } else if (q == currentQuality && tv->tv_sec > (zeroOffsetSecs + 3 * 24 * 60 * 60)) {
+        // Sometimes nodes have RTC issues and broadcast time in the past. Jump clock forward if our time is "older" than 3 days.
+        shouldSet = true;
+        LOG_INFO("External time is more than 3 days in the future. Setting RTC to %ld. A neighbour likely has a broken clock.\n", tv->tv_sec);
+        LOG_DEBUG("new_time=%ld, old_time=%ld, diff=%ld\n", tv->tv_sec, zeroOffsetSecs, (tv->tv_sec - zeroOffsetSecs));
     } else {
         shouldSet = false;
         LOG_DEBUG("Current RTC quality: %s. Ignoring time of RTC quality of %s\n", RtcName(currentQuality), RtcName(q));


src/gps/RTC.cpp
```diff
@@ -113,6 +113,12 @@ bool perhapsSetRTC(RTCQuality q, const struct timeval *tv, bool forceUpdate)
         // Every 12 hrs we will slam in a new GPS or Phone GPS / NTP time, to correct for local RTC clock drift
         shouldSet = true;
         LOG_DEBUG("Reapplying external time to correct clock drift %ld secs\n", tv->tv_sec);
+    } else if (q >= RTCQualityFromNet && tv->tv_sec > (zeroOffsetSecs + 3 * 24 * 60 * 60)) {
```
eljojo, line 116: Maybe this would make the upgrade safer? so it only tries to fix if we have a clock of low quality

```suggestion
    } else if (currentQuality == RTCQualityFromNet && tv->tv_sec > (zeroOffsetSecs + 3 * 24 * 60 * 60)) {
```

eljojo:
I noticed this warning while building. Is this something I'd fix by type casting?

```
src/gps/RTC.cpp: In function 'bool perhapsSetRTC(RTCQuality, const timeval*, bool)':
src/gps/RTC.cpp:116:50: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     } else if (q == currentQuality && tv->tv_sec > (zeroOffsetSecs + 3 * 24 * 60 * 60)) {
                                       ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

RichieFrame:
I am thinking out loud here, but there is probably a better solution, which is to ignore timestamps from a position update in the case where the sats in view is equal to 0, as generally that means it is a fixed position node and not getting the time from a satellite.

Additionally, does the firmware ignore time sync from the mesh when it has its own GPS fix?

thebentern:
> I am thinking out loud here, but there is probably a better solution, which is to ignore timestamps from a position update in the case where the sats in view is equal to 0, as generally that means it is a fixed position node and not getting the time from a satellite.
> 
> Additionally, does the firmware ignore time sync from the mesh when it has its own GPS fix?

I think this is the right idea, but we should look at the DOP flags instead of num sats. 
I don't think difference in times is a good metric for whether we should ignore or not. It's a pretty tenuous criteria.

thebentern:
Tracking effort in #3984

PR #4343, opened 2024-07-29
Event mode: Enforce reliable hop limit and disallow default public MQTT 

thebentern:


GUVWAF (approved):

src/mesh/FloodingRouter.cpp
```diff
@@ -46,6 +46,11 @@ void FloodingRouter::sniffReceived(const meshtastic_MeshPacket *p, const meshtas
                 meshtastic_MeshPacket *tosend = packetPool.allocCopy(*p); // keep a copy because we will be sending it
 
                 tosend->hop_limit--; // bump down the hop count
+#if EVENT_MODE
+                if (tosend->hop_limit > 2) {
+                    tosend->hop_limit = 2;
```
GUVWAF, line 51: This will mess up the "hops away" count if you care about that. Can be "fixed" by changing "hop start" as well.

jp-bennett: Good catch. I think the latest code takes care of it.

PR #4344, opened 2024-07-29
Cleanup for Air530z GPS

fifieldt:
It turns out that the Air530z is a GNSS_MODEL_MTK. Our existing code detects and configures it properly. This patch removes a couple of AIROHA ifdefs since they are not required for a working GPS.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -400,12 +400,6 @@ bool GPS::setup()
     int msglen = 0;
 
     if (!didSerialInit) {
-#ifdef GNSS_AIROHA
-        if (tx_gpio && gnssModel == GNSS_MODEL_UNKNOWN) {
-            probe(GPS_BAUDRATE);
-            LOG_INFO("GPS setting to %d.\n", GPS_BAUDRATE);
-        }
-#else
 #if !defined(GPS_UC6580)
 
         if (tx_gpio && gnssModel == GNSS_MODEL_UNKNOWN) {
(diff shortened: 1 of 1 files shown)


markbirss:
@fifieldt This change also affects Tracker T1000-e GPS

fifieldt:
yes @markbirss , I believe setting #define GPS_BAUDRATE 115200 in the T1000e variant.h is incorrect: the Air530z defaults to 9600. Try removing that line?

markbirss:
> yes @markbirss , I believe setting #define GPS\_BAUDRATE 115200 in the T1000e variant.h is incorrect: the Air530z defaults to 9600. Try removing that line?

Does not seem to help.

fifieldt:
What's breaking? What is it detecting as? Can you confirm if the Air530 in the T1000E is the same as the Air530 available for sale from the Seeed store (I.e. an Air530z)?

markbirss:
@failed "Failed to find UBlox & MTK GNSS Module using baudrate"

I cannot confirm only have T1000-E

@Dylanliacc can you confirm ?

fifieldt:
Thanks for looking at it. If it's not too much trouble, what are the baud rates it tries?

markbirss:
It does try couple 9600,4800, 57600 and 115200

DEBUG | ??:??:?? 2 \[GPS\] Probing for GPS at 9600                                   
WARN  | ??:??:?? 6 \[GPS\] Failed to find UBlox & MTK GNSS Module using baudrate 9  
600

fifieldt:
Working on something I think will fix t1000e

https://github.com/meshtastic/firmware/compare/master...fifieldt:meshtastic-firmware:l89-2

fifieldt:
Working on the fix in #4394

PR #4345, opened 2024-07-29
Add setuptools to CI/CD basic setup

fifieldt:
Build was failing due to missing setuptools.


fifieldt:
This needs the pip version instead.

PR #4355, opened 2024-07-30
PA FAN Disable

gjelsoe:
Ability to disable PA FAN if RF95_FAN_EN is set, doesn't require reboot.


thebentern (approved):

gjelsoe:
Can figure out, why Trunk is failing.

PR #4365, opened 2024-08-01
"Scan and Select" input for Canned Messages

todd-herbert:
### What
A new input source for the canned message module, using a single push-button connected to GPIO.

Short press: scroll through the pre-set messages
Long press: send the currently highlighted message to primary channel

### Why
Some boards have an additional built-in button, which is currently unused. The proposed "scan and select" input is an attempt to make productive use of the feature.

PR enables canned messages as default for Heltec VME290 and VME213 boards.
By default, the module is *partially configured*: pins and input method are set, however no messages are defined.
Pressing the configured button during this state alerts on-screen: "no canned messages". This will prompt the user to either configure messages, or disable the module entirely.

[Doc for proposed changes](https://github.com/meshtastic/meshtastic/compare/master...todd-herbert:meshtastic-doc:scan-and-select)

https://github.com/user-attachments/assets/09206c27-6f30-4029-930b-99d2538afe1b

https://github.com/user-attachments/assets/df6067f3-955b-43d7-a29d-7ed36b4d9480

### DIFF
--- a/src/graphics/EInkDynamicDisplay.cpp
+++ b/src/graphics/EInkDynamicDisplay.cpp
@@ -375,7 +375,7 @@ void EInkDynamicDisplay::hashImage()
 
     // Sum all bytes of the image buffer together
     for (uint16_t b = 0; b < (displayWidth / 8) * displayHeight; b++) {
-        imageHash += buffer[b];
+        imageHash ^= buffer[b] << b;
     }
 }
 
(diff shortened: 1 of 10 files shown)


todd-herbert:
Failing CI, needs some more guarding.

thebentern:
Ooooh so this is a single additional button on GPIO in addition to a user button?

tropho23:
This is very cool indeed

todd-herbert:
I'd be keen to get this in for 2.4.2 if possible, to try catch as many new VME213 and VME290 boards with this PRs default module config. Otherwise they'll probably miss out on the feature until they erase the flash next.

thebentern:
Let's do it!

PR #4373, opened 2024-08-02
Remove OSTimer

fifieldt:
This code is not included anywhere, and none of the functions are called by code elsewhere in meshtastic.


fifieldt:
I think the build failures are unrelated to this patch.

PR #4374, opened 2024-08-02
Fix build to not use incorrect OneButton version

geeksville:
OneButton pushed out a new update today that has a different API rather than just use whichever new version they push, stay on 2.5.x until someone sees a need to update.  Fixes build for wm1100 tracker.

### DIFF
--- a/arch/stm32/stm32.ini
+++ b/arch/stm32/stm32.ini
@@ -32,5 +32,5 @@ lib_deps =
   https://github.com/caveman99/Crypto.git#f61ae26a53f7a2d0ba5511625b8bf8eff3a35d5e
 
 lib_ignore =
-  mathertel/OneButton
+  mathertel/OneButton@^~2.5.0
   Wire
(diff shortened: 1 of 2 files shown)


geeksville:
btw @thebentern - I think the reason recent CI builds are now failing is because of this also.

thebentern:
@geeksville yep it definitely is 😅 
I actually submitted a fix PR and it got merged and released upstream so I think we can peg this one to 2.6.1 
https://github.com/mathertel/OneButton/releases/tag/2.6.1

geeksville:
ooh I see !
2.6.1 will have your fix bravo!

thebentern (approved):

geeksville:
hmm @thebentern  it seems like the platformio version fetching isn't working with their release name for 2.6.1.   It only finds 2.6.0 if the platformio.ini file is updated.  I think they might have strict rules and https://github.com/mathertel/OneButton/releases botched the name of the release (not calling it literally 2.6.1 but instead 'fixing blah blah).  This apparently breaks the github auto version resolution for 'releases' (notice how interestingly https://github.com/mathertel/OneButton says the latest release is 2.6.0 still)

thebentern:
Weird. Perhaps we can just use the github tag?

PR #4376, opened 2024-08-02
fix #4367 make USB power detection work correctly on seeed trackers

### LINKED ISSUE #4367: [Bug]: Seeed Tracker T1000-E tracker doesn't detect when USB power is applied
### Category

Other

### Hardware

Other

### Firmware Version

master (private build)

### Description

Some of our behaviors change when the device realizes it is connected to USB power.  I just connected my T1000-E to my desktop and happened to notice this (low priorty/minor) problem:

```
DEBUG | 15:45:39 155470 [Power] Battery: usbPower=0, isCharging=0, batMv=4114, batPct=94
```

usbPower should be 1 in this message (because I was connected to USB power)
I suspect we need to add some variant code so that this detection happens (but I haven't yet looked at the schematic).  If someone else wants to investigate/fix this then great, otherwise I'll give it a shot sometime in the next few days

(not a big problem but cc @Dylanliacc fyi )

### Relevant log output

```Shell
DEBUG | 15:45:39 155470 [Power] Battery: usbPower=0, isCharging=0, batMv=4114, batPct=94
```

geeksville:
for wio tracker 1110 and 1000-E and possibly other nrf52 boards. The problem was that nrf52 power stuff wasn't generating regular powerstatus notifications (because that code was guarded by a batteryLevel check which was null for those boards).  So I've cleaned up the battery status stuff a bit and we now have fewer special cases.
Tested on a 1000-E, tracker 1110 and a rak4631 board.

(fix #4367)

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -461,7 +461,7 @@ class AnalogBatteryLevel : public HasBatteryLevel
 #endif
 };
 
-AnalogBatteryLevel analogLevel;
+static AnalogBatteryLevel analogLevel;
 
 Power::Power() : OSThread("Power")
 {
@@ -560,6 +560,10 @@ bool Power::setup()
 {
     bool found = axpChipInit() || analogInit();
 
+#ifdef NRF_APM
+    found = true;
+#endif
+
     enabled = found;
     low_voltage_counter = 0;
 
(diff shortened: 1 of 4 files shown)


thebentern:
Legend! 
I be willing to bet that's a regression from some of the recent work on using INA sensors for device battery level voltage.

thebentern (approved):

fifieldt:
nice! Care to add to the wio-sdk-wm1110 too?

geeksville:
> nice! Care to add to the wio-sdk-wm1110 too?

hmm - I'm not sure if it will work there because on that board the USB isn't really going to the nrf52840, instead it is going to a little USB serial chip (oops).  I'll check the schematic though.

geeksville:
@fifieldt just checked schematic.  Alas, this approach won't work for that board.  

Btw is a bummer they forgot to include (at-least) a voltage-divider circuit to let the CPU sense the battery voltage (as far as I can tell - haven't checked super carefully).

I hope they included battery level sensing in the 1000-E board because if they didn't: ouch.

geeksville:
oh they did for the 1000-E.  good!

```
#define BATTERY_PIN 2
#define ADC_MULTIPLIER (2.0F)

#define ADC_RESOLUTION 14
#define BATTERY_SENSE_RESOLUTION_BITS 12
```

fifieldt:
ah! I forgot about that. Thank you very much for investigating.

fifieldt (approved):

Issue #4378, opened 2024-08-03
[Bug]: nrf52 boards are not entering deep sleep when battery is critically low (make system_off mode work on nrf52)


geeksville:
### Category

Hardware Compatibility

### Hardware

All nrf52 boards

### Firmware Version

master (private build)

### Description

I just happened upon this code in Power.cpp:

```
#ifdef ARCH_NRF52
                // We can't trigger deep sleep on NRF52, it's freezing the board
                LOG_DEBUG("Low voltage detected, but not triggering deep sleep\n");
#else
                LOG_INFO("Low voltage detected, triggering deep sleep\n");
                powerFSM.trigger(EVENT_LOW_BATTERY);
#endif
```

Freezing boards is bad, but IMO it is **super** important to enter deep sleep anytime we are able to detect that the battery is critically low.  Because to not do this will kill lipos.  

I'm happy to look into this as I continue my adventure of improving nrf52 power consumtion.  But opening an issue in case any of ya'll have comments on this theory?

todd-herbert:
I built one node with a small solar enclosure from RAK. [(Unify Enclosure Solar IP65 100x75x38)](https://store.rakwireless.com/products/unify-enclosure-ip65-100x75x38-solar) 

* RAK4631
* RAK19007
* 1000mAh LiPo cell (to fit claimed 55x34x6mm space below mounting plate)

I was concerned that this battery could run flat. I made a [quick-and-dirty modification](https://github.com/meshtastic/firmware/compare/master...todd-herbert:meshtastic-firmware:nasty-solar-hack) to the firmware to add some level of software protection.

In essence: 
* Early in boot, check voltage
  - If below 3.7V, enter *pseudo deep-sleep*: `delay()` 30 minutes, then reboot
  - If above 3.7V, continue boot
* If `powerStatus` reports below 3.45V, reboot the node. This triggers the low voltage check.

This current draw in this *pseudo deep-sleep* state is very good; a few µA from memory? *However*, below a certain voltage, I believe the buck converter on the RAK19007 baseboard switches to a linear mode. Between a certain voltage range (3.5V down to 3.3V?) the current is higher; almost 1mA at worst.

Even in this 1mA region, the current is low enough to protect the battery from many days of poor weather. Despite the small battery, I haven't actually had this happen yet since deployment, although it worked very well in testing.

[...]

PR #4382, opened 2024-08-03
Remove LED_INVERTED, see below for why ;-)

geeksville:
While working on #4378 I noticed a funny problem: the blinking system LED was on during deep-sleep.  Initially I thought it was some weird sleep hw config thing but it turns out it was easier but more pervasive.

We had two different preprocessor symbols which both meant approximately the same thing LED_INVERTED and LED_STATE_ON (though their polarity was opposite). Some variant files were setting one, others were setting the other, and others were setting both. heh.

In the case of the board I was testing (seeed tracker wio 1100) it was only setting one and the default behavior for the other (for all boards) was incorrect.  So I did a grep and it seems like LED_STATE_ON was used more often, so I kept that one and removed LED_INVERTED everywhere.

### DIFF
--- a/src/DebugConfiguration.h
+++ b/src/DebugConfiguration.h
@@ -3,8 +3,8 @@
 #include "configuration.h"
 
 // DEBUG LED
-#ifndef LED_INVERTED
-#define LED_INVERTED 0 // define as 1 if LED is active low (on)
+#ifndef LED_STATE_ON
+#define LED_STATE_ON 1
 #endif
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


caveman99:
All changed variant targets are esp32 boards. Could it be this was tool chain dependant at one time?

geeksville:
Good theory.  IDK but I kinda doubt it because both are just using standard cpp predefines.  It might just be back when I started it I had both (because stupid  or just random codereuse history ;-) ).

thebentern (approved):

PR #4383, opened 2024-08-03
Make lora radio reset reliable on wio-tracker-1100 and lower lr11x0 power consumption in sleep

geeksville:
Two related changes:

* Fix wio-tracker-1110 lora radio reset GPIO assignment 
This fixes flaky lora radio init on this board.  Prior to this change while debugging I would often see the lr11x0 fail to init.

* No need to keep lr11x0 radio config during sleep anymore, also stop TCXO

I think the problem (at least on the board I'm using for power testing
a wio tracker 1110) was that actually the RESET GPIO was not correct for
the radio.  This led to the radio not being properly reinited after exiting
sleep mode.

Now that the GPIO is fixed I can enter deep sleep (fully shutting down
radio) and then later when the CPU resets, it can successfully init the
radio and send packets.

After this seeming success, I also turned off the TCXO during sleep and that
worked as well.

### DIFF
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -285,17 +285,15 @@ template <typename T> bool LR11x0Interface<T>::isActivelyReceiving()
 
 template <typename T> bool LR11x0Interface<T>::sleep()
 {
-    // Not keeping config is busted - next time nrf52 board boots lora sending fails  tcxo related? - see datasheet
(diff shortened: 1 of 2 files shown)


geeksville:
@caveman99 does this seem reasonable to you?  From looking at commits it seems like you did most of the awesome lr11x0 stuff.

thebentern (approved):

fifieldt:
Flashed this on a wio-tracker-1100, and nothing caught fire :) Thanks, @geeksville .

PR #4385, opened 2024-08-04
Configure pin modes of selected pins before attempting to write to them in Remote Hardware module

Kealper:
This is a small change to properly configure the correct pin mode before attempting to write to the specified pins in the mask in the Remote Hardware module. This change fixes Remote Hardware writes on RP2040-based devices.

Previously, the pin modes were set after the writes had taken place, and it seems like the Arduino core for ESP32 implicitly set the target pin to `OUTPUT` automagically before writing if it wasn't set prior. This does not appear to be the case with the `arduino-pico` core so writes to a misconfigured pin would just silently fail.


fifieldt:
Nice find!

GUVWAF (approved):
Nice find indeed!

PR #4392, opened 2024-08-05
Fix python call in UF2 generation.

fifieldt:
The call to generate UF2 files in the platformio custom script was a bare call to python. In some environments, this command won't exist in this way. Instead, use the standard env approach to find the right python.

Without this change, a user may see:
"""
Generating UF2 file
sh: 1: python: not found
"""

Additionally, add the shebang line on line 1 so this script can be executed standalone if needed.

### DIFF
--- a/bin/platformio-custom.py
+++ b/bin/platformio-custom.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 # trunk-ignore-all(ruff/F821)
 # trunk-ignore-all(flake8/F821): For SConstruct imports
 import sys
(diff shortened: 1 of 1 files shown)


thebentern (approved):

mrekin:
Will this work on windows?
I got some error now, but don't know if this change is issue
`
\x91\xa8\xe1⥬\xa5 \xad\xa5 㤠\xa5\xe2\xe1\xef \xad\xa0\xa9\xe2\xa8 㪠\xa7\xa0\xad\xad\xeb\xa9 \xaf\xe3\xe2\xec.
*** [.pio\build\nrf52_promicro_diy_tcxo\firmware.hex] Error 1
`

PR #4399, opened 2024-08-05
Make UF2 build command windows-friendly

fifieldt:
As reported by @mrekin, the previous changes to the platformio custom build script may not work on windows. Change to use python3 instead of a call to /usr/bin/env python3.


thebentern (approved):

mrekin:
Maybe better use something like {sys.executable} instead?

fifieldt:
Damn! Thanks for checking @mrekin .

{sys.executable} works on linux, will send an updated.

PR #4402, opened 2024-08-06
Use sys.executable to refer to python.

fifieldt:
Thanks to @mrekin for testing the build on Windows. The previous fix for this UF2 call did not work.
sys.executable should fix it.


thebentern:
3rd time's charm ;-)

PR #4421, opened 2024-08-08
Genericise genericisable variant defines

fifieldt:
This got a bit out of hand! It all started seeing comments from board makers, noting we had multiple overlapping variables for the same thing ...
 "FIXME: for some reason both are used in /src"
 " FIXME - we really should define LORA_CS instead"
 "[added for] Compatibility with variant file configuration structure"

Defines proliferated based on specific chip names, even though there was no different functionalities between chips. Re-use was minimal. For example, SX126X_CS
 and LORA_CS were commonly aliased everywhere, and used more or less
interchangably.

Working theories:
1. There are a minimum set of pins that tend to be used for LORA radios for which we need variables. These are LORA_* flags, and when a developer sets them in  variant.h , they should expect relatively common and sane behaviour.
The current set is: LORA_SCK, LORA_MISO, LORA_MOSI, LORA_CS, LORA_RESET, LORA_BUSY, LORA_DIO0
2. If there's a need to do different things with these pins for a particular radio in our code we check USE_[RADIONAME] eg #if defined(USE_SX1268). See main.cpp for an example of how we do it right now for creating interfaces.
3. For unique pins of a lora radio, make a new [RADIONAME]_* variable as we've always done!

I think that this direction will make it easier to add new variants, and particularly new variants with different lora radios. At the very least, it removes a few lines from most variant.h files ;)

So very tired.


fifieldt:
CI / build-esp32 (tlora-v1) / build-esp32 (pull_request_target)  - failed due to a timeout of github API, appears to be unrelated to code.

jp-bennett:
> CI / build-esp32 (tlora-v1) / build-esp32 (pull_request_target) - failed due to a timeout of github API, appears to be unrelated to code.

Bravo on the work. I've poked CI to see if it'll complete that failure this time.

fifieldt:
Hey! Thanks for the review and taking the time to provide education too, much appreciated. This was still a worthwhile activity for me for learning the code base, so I don't mind :)

What do you think of flipping this idea - doubling down on the chipname variables and getting rid of the lora_* ones?

caveman99:
> What do you think of flipping this idea - doubling down on the chipname variables and getting rid of the lora_* ones?

Not an easy one either. We just streamlined a few of these to replace the RF95 named variables :-)

PR #4423, opened 2024-08-09
Adds MESHTASTIC_EXCLUDE_TZ option

jp-bennett:
While timezone support is amazing for targets with screens, it's useless for repeaters, and is a few more bytes to shave off.


jp-bennett:
On the WIO-e5, usage goes from 
``Flash: [==========]  98.4% (used 258060 bytes from 262144 bytes)``
to 
``Flash: [==========]  98.3% (used 257640 bytes from 262144 bytes)``

According to the map file, there are still timezone related functions in the libraries, so there is more savings to be had here.

PR #4425, opened 2024-08-09
adds MINIMUM_SAFE_FREE_HEAP macro

jp-bennett:
Prepping for landing the PKI PR. This gets away from determining how much free memory is safe based on NodeDB size, and instead sets it to a conservative 1.5k. In testing, this hasn't been a problem for any devices.


fifieldt (approved):

jp-bennett:
Pulling into the 2.5 branch with the other PKI changes

PR #4427, opened 2024-08-09
Split factory reset into config and device variants

thebentern:


thebentern:
@garthvh @andrekir @Hunter275 I think most folks will want the existing factory reset function in the apps to send the `factory_reset_config` now which resets everything except for the BLE bonds.

@ianmcorvidae I think the `factory_reset_device` makes sense in CLI. I'm wondering if we add an optional argument to the existing command to specify that we're factory reseting the entire device?

PR #4428, opened 2024-08-09
Do not forward private messages via MQTT, allow telemetry

musznik:
This commit introduces a feature that enhances security by preventing the sending of private messages between nodes when the MQTT uplink is enabled. The motivation behind this change is to ensure that sensitive information is not inadvertently transmitted over potentially insecure channels. Public messages broadcast to address 0xffffffff, along with telemetry data, remain unrestricted to maintain the core functionality of the network. 

This feature is configurable via the command 

`meshtastic --set mqtt.secure_messages [true|false]`

providing users with flexibility and control over their network's security.

### DIFF
--- a/src/mesh/Default.h
+++ b/src/mesh/Default.h
@@ -19,6 +19,7 @@
 #define default_mqtt_username "meshdev"
 #define default_mqtt_password "large4cats"
 #define default_mqtt_root "msh"
+#define default_secure_messages true
 
 #define IF_ROUTER(routerVal, normalVal)                                                                                          \
     ((config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER) ? (routerVal) : (normalVal))
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -408,6 +408,7 @@ void NodeDB::installDefaultModuleConfig()
     strncpy(moduleConfig.mqtt.password, default_mqtt_password, sizeof(moduleConfig.mqtt.password));
     strncpy(moduleConfig.mqtt.root, default_mqtt_root, sizeof(moduleConfig.mqtt.root));
     moduleConfig.mqtt.encryption_enabled = true;
+    moduleConfig.mqtt.secure_messages = true;
 
     moduleConfig.has_neighbor_info = true;
     moduleConfig.neighbor_info.enabled = false;
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -257,7 +257,16 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
 #if !MESHTASTIC_EXCLUDE_MQTT
         // Only publish to MQTT if we're the original transmitter of the packet
         if (moduleConfig.mqtt.enabled && p->from == nodeDB->getNodeNum() && mqtt) {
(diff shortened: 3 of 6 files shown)


garthvh:
There is already granular control of uplink and downlink by channel, why add another less granular setting?

GUVWAF:
So I assume you want others not to read your direct messages, which use the normal channel key. Let's wait for the better solution: #4379, which implements public key encryption per destination.

For now, indeed don't enable uplink on an insecure channel.

musznik:
> There is already granular control of uplink and downlink by channel, why add another less granular setting?

> So I assume you want others not to read your direct messages, which use the normal channel key. Let's wait for the better solution: https://github.com/meshtastic/firmware/pull/4379, which implements public key encryption per destination.

However, I'd like to point out that even with public key encryption in place, when messages are uplinked to MQTT, they are decrypted, which means they can be transmitted to the server in an unencrypted form. This creates a potential risk, as users have no control over what happens to the data after it's decrypted, which could compromise privacy.

That's why I believe it's valuable to introduce an additional option to control whether private messages can be sent to MQTT. My proposal isn't intended to replace future improvements but rather to complement them, giving users the ability to better secure their data during the transition period and beyond if they deem it necessary.

jp-bennett:
'However, I'd like to point out that even with public key encryption in place, when messages are uplinked to MQTT, they are decrypted'
How do you figure? They definitely won't be by random listening nodes along the way. Is the problem that the destination node is uploading received DMs? Even then, they won't do so without JSON turned on.

musznik:
> 'However, I'd like to point out that even with public key encryption in place, when messages are uplinked to MQTT, they are decrypted' How do you figure? They definitely won't be by random listening nodes along the way. Is the problem that the destination node is uploading received DMs? Even then, they won't do so without JSON turned on.
> 
> This does raise an interesting question. The PKI DMs aren't really on an existing channel. How should they interact with MQTT?

When I mentioned that messages are decrypted before being uplinked to MQTT, I was specifically referring to the process occurring at the node that is configured to send data to the MQTT broker. In this context, after the node receives a message (including DMs), it processes and potentially decrypts the packet to interpret or forward it. If the node is set up to uplink messages to MQTT and JSON output is enabled, these decrypted messages could then be sent in plaintext to the MQTT broker.

You’re correct that random nodes along the way won't decrypt the messages; the decryption would only happen at the intended destination node or a node explicitly configured to process the message for MQTT.

This brings us to the point about how PKI DMs, which aren't really tied to an existing channel, should interact with MQTT. Since PKI DMs are designed to be secure, their integration with MQTT needs careful handling to ensure that sensitive information isn't unintentionally exposed when transmitted to MQTT brokers. A possible approach could involve setting a stricter policy for handling DMs within the MQTT integration, such as not forwarding them at all or doing so only under specific, secure conditions.

This is where my proposal for an additional option comes in—by allowing users to control whether private messages (including DMs) can be sent to MQTT, we can ensure that sensitive data remains secure, even in scenarios where JSON is enabled or other conditions could lead to unintentional exposure.

jp-bennett:
@musznik There's some good stuff to think about here. My initial thought is that PKI DMs will be treated as a channel 8, with an uplink/downlink setting of its own. I think that accomplishes what you're looking for here.

PR #4431, opened 2024-08-10
2.5 Changes

thebentern:
- PKI infrastructure
- Optionals on telemetry
- New admin message for setting time only
- Client Notifications on phone API
- New Short Turbo Preset
- Deprecated VLong Slow
- Deprecated some device config fields and migrated to new Security config
- Moved NodeInfoLite to use smaller UserLite proto
- Added new LocalStats telemetry variant to send over phone API every 15 minutes


jp-bennett:
@thebentern I think we can pull the trigger on this. It may be worth making a 2.4 branch in case we want to mint another release there before 2.5 is ready for beta.

PR #4442, opened 2024-08-11
Improves ignore messages in Router.cpp

charlieh0tel:
- Improves ignore messages in Router.cpp
- Cleans up flow using standard "goto out" pattern to handle cleanup.


thebentern:
I'm a big fan of guard clauses, but less of a fan of goto statements. Perhaps we can take a similar approach in the if blocks but instead of the goto we can do early return ala:
```
packetPool.release(p);
return;
```

charlieh0tel:
The Linux kernel uses to goto pattern because it avoids copypasta, which has been demonstrated to cause bugs.

But I will change it to copypasta + nlx per your request.

thebentern:
> copypasta, which has been demonstrated to cause bugs.

In my experience if copypasta is concern, it's typically a code-smell that extracting a method to encapsulate the duplicated behavior is called for. That could also be a QoL change as a part of this cleanup. FWIW, I understand gotos can be used effectively, but I've also seen them used to create some truly awful spaghetti code. We've gotten this far without using them, so I'd rather not open pandora's box. 😅

charlieh0tel:
I did it the copypasta + nlx way you requested.    I think it is ready to merge.   PTAL.

charlieh0tel:
Caught to head of master.    Should be good to go.  PTAL @thebentern

thebentern (approved):

Issue #346, opened 2020-08-29
Make default xmit power for Japan 13 dBm


geeksville:
Based on comment by Yuki:

```
According to Japanese rules.

Max 1mW 915MHz - 930MHz No radio license registration
Max 20mW 920.6MHz - 928MHz No radio license registration
Max 250mW 920.6MHz - 923.4MHz Radio license registration required

It would be great if you could change the default powerlevel and
frequency band from the Android app.
```

and https://www.everythingrf.com/rf-calculators/dbm-to-watts

It seems like we should set the default xmit power when in Japan to 13 dBm.  Probably by adding a REGION_DEFAULT_POWER definition?

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/about-the-japanese-version/1171/6

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/about-the-japanese-version/1171/11

PR #4446, opened 2024-08-12
JP frequency - 20mW limit, change freqs to avoid duty cycle

fifieldt:
Thanks to user Goyath on Discord, we discovered that in Japan the 250mW radio level requires licensing, and 20mW is the practical limit. We also discovered that a duty cycle of 10% is needed on most frequencies:

* CH 24-38 920.5-923.5 20mW no airtime restrictions
* CH 39-61 923.5-928.1 20mW 10% airtime

This patch limits the power and restricts the frequency to those with no airtime restrictions.

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -53,8 +53,10 @@ const RegionInfo regions[] = {
 
     /*
         https://lora-alliance.org/wp-content/uploads/2020/11/lorawan_regional_parameters_v1.0.3reva_0.pdf
+        https://www.arib.or.jp/english/html/overview/doc/5-STD-T108v1_5-E1.pdf
+        https://qiita.com/ammo0613/items/d952154f1195b64dc29f
      */
-    RDEF(JP, 920.8f, 927.8f, 100, 0, 16, true, false, false),
+    RDEF(JP, 920.5f, 923.5f, 100, 0, 13, true, false, false),
 
     /*
         https://www.iot.org.au/wp/wp-content/uploads/2016/12/IoTSpectrumFactSheet.pdf


Talie5in:
May need to account for remapping channel slots.

JPN Slot 20 been default for Public LongFast and been above 923Mhz
Screenshot_20240812_203923_Chrome

Screenshot_20240812_201731_Chrome
Screenshot_20240812_193229_Adobe Acrobat
Screenshot_20240812_193137_Adobe Acrobat
Screenshot_20240812_193116_Adobe Acrobat

fifieldt:
Conclusion from discord discussion is:


* CH 24-38 920.5-923.5 20mW no airtime restrictions
* CH 39-61 923.5-928.1 20mW 10% airtime

src/mesh/RadioInterface.cpp
```diff
@@ -53,8 +53,10 @@ const RegionInfo regions[] = {
 
     /*
         https://lora-alliance.org/wp-content/uploads/2020/11/lorawan_regional_parameters_v1.0.3reva_0.pdf
+        https://www.arib.or.jp/english/html/overview/doc/5-STD-T108v1_0-E1.pdf
+        https://qiita.com/ammo0613/items/d952154f1195b64dc29f
      */
-    RDEF(JP, 920.8f, 927.8f, 100, 0, 16, true, false, false),
+    RDEF(JP, 920.6f, 927.8f, 10, 0, 13, true, false, false),
```
fifieldt, line 59: An alternative to this is:
    RDEF(JP, 920.6f, 923.5f, 100, 0, 13, true, false, false),
    
    to use just the frequencies with no airtime restrictions.

GUVWAF: I think that would be the best option, then you still have 2.9MHz, which is way more than e.g. EU868.

In any case this will be a breaking change though, but if there are not a lot of users in Japan yet, that might not be a big problem.

fifieldt: Thanks @GUVWAF , I've updated the patch.

fifieldt:
It looks like we discovered this some time ago:

https://github.com/meshtastic/firmware/issues/346
https://meshtastic.discourse.group/t/about-the-japanese-version/1171/4

but it got left off the task list somehow

numpad0:
Just passing by & just FYI, 920MHz low-power in JP is indeed license free but modems need to be type certified and 〶[R]xxx-xxxx marked for full legal conformance, similar to Wi-Fi cards

truedat101:
> 〶

Thanks for this post ... we are trying to figure out the list of compliant devices so we can maintain them for Japan.  Esp since there are like 7 public nodes in Japan and only in Tokyo , we would like to improve this situation.  Still trying to figure out why this is the case....

PR #4455, opened 2024-08-13
Shorter timeout for some NodeInfo messages

jp-bennett:
A NodeInfo message sent directly to the node, with wantAck set to true, triggers a higher priority NodeInfo packet with a quicker cool off period. We try to trigger this when a response indicates a PKI decrypt failure.


jp-bennett:
Dang it, fixing 2.5 clobbered this PR.

PR #4464, opened 2024-08-14
Add PKI channel for MQTT

jp-bennett:
If we are connecting to MQTT, this adds the root/2/e/PKI topic for PKI messages.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -152,7 +152,8 @@ void MQTT::onReceive(char *topic, byte *payload, size_t length)
                     LOG_INFO("Ignoring downlink message we originally sent.\n");
             } else {
                 // Find channel by channel_id and check downlink_enabled
-                if (strcmp(e.channel_id, channels.getGlobalId(ch.index)) == 0 && e.packet && ch.settings.downlink_enabled) {
+                if ((strcmp(e.channel_id, "PKI") && e.packet) ||
+                    (strcmp(e.channel_id, channels.getGlobalId(ch.index)) == 0 && e.packet && ch.settings.downlink_enabled)) {
                     LOG_INFO("Received MQTT topic %s, len=%u\n", topic, length);
                     meshtastic_MeshPacket *p = packetPool.allocCopy(*e.packet);
                     p->via_mqtt = true; // Mark that the packet was received via MQTT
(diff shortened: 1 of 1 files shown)


src/mqtt/MQTT.cpp
```diff
@@ -161,8 +162,11 @@ void MQTT::onReceive(char *topic, byte *payload, size_t length)
                         p->channel = ch.index;
                     }
 
+                    // PKI messages get accepted even if we can't decrypt
+                    if (router && p->which_payload_variant == meshtastic_MeshPacket_encrypted_tag && p->channel == 0)
```
GUVWAF, line 166: Better to check for `(strcmp(e.channel_id, "PKI")`  to avoid not decrypting when you happen to have a channel hash of 0.

Also, I think here we want to check whether the `to` and `from` are in our NodeDB to avoid routing PKI messages for nodes that do not talk via your channels.

src/mqtt/MQTT.cpp
```diff
@@ -377,6 +381,11 @@ void MQTT::sendSubscriptions()
 #endif // ARCH_NRF52
         }
     }
+#if !MESHTASTIC_EXCLUDE_PKI
+    std::string topic = cryptTopic + "PKI/#";
+    LOG_INFO("Subscribing to %s\n", topic.c_str());
+    pubSub.subscribe(topic.c_str(), 1);
```
GUVWAF, line 387: Let's only subscribe to the "PKI" channel when you have at least one channel downlink enabled.

src/mqtt/MQTT.cpp
```diff
@@ -513,8 +531,12 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp, const meshtastic_MeshPacket &
             // FIXME - this size calculation is super sloppy, but it will go away once we dynamically alloc meshpackets
             static uint8_t bytes[meshtastic_MeshPacket_size + 64];
             size_t numBytes = pb_encode_to_bytes(bytes, sizeof(bytes), &meshtastic_ServiceEnvelope_msg, env);
-
-            std::string topic = cryptTopic + channelId + "/" + owner.id;
+            std::string topic;
+            if (mp.pki_encrypted) {
```
GUVWAF, line 535: What's the channel index of PKI messages? I.e. what's line 500 (`auto &ch = channels.getByIndex(chIndex);`) going to return?

If line 500 doesn't give you a valid channel, you'll never get here for PKI messages.

jp-bennett: PKI messages leave channel index set to 0.

GUVWAF: Then this currently only works when you have uplink enabled on channel 0.

jp-bennett: I figured there would be edge cases I hadn't spotted yet. I'll get to work on these.

src/mqtt/MQTT.cpp
jp-bennett: Latest change to 2.5 branch fixes all of this, except checking for to/from. I'm still pondering that one.

PR #4467, opened 2024-08-15
feature: default to fuzzy GPS location on the Default Channel

### LINKED ISSUE #4412: [Feature Request]: Please default to fuzzy GPS location on the Default Channel.
### Platform

ESP32

### Description

Setting up and installing new firmware, I need to rush to disable precise location before my exact location gets broadcast out to the mesh.  This is insecure and not a good default. Please consider switching to a fuzzy location of 1 mile as the default.

mrfyda:
First contribution to the repo, attempting to fix #4412.

The original issue suggested using a default fuzzy location with 700m precision. This pull requests instead is using ~1500m precision which is the same default as for MQTT Map Report module.


thebentern (approved):

thebentern:
I think this is a reasonable change. @jp-bennett @garthvh any objections?

garthvh:
> I think this is a reasonable change. @jp-bennett @garthvh any objections?

Might be good to target 2.5, the circles may not even be in the play store yet.  These imprecise positions just seem wrong without the visualization in place.

jp-bennett (approved):

ianmcorvidae:
I have one thought, which is that Android currently has options for 32 (high precision), 16 (medium precision) and 11 (low precision) only, and shows an input box for other options (but only if it's already outside of those three + disabled). I think we should either use one of those here, or (better, IMO) add more options to the default available ones on Android to include 14. Either adding medium-low at 14 (and medium-high, maybe, but 16 is already kinda that), or adding a way to select everything in something like 10-19 or 11-16 or such (whatever range iOS allows, probably).

garthvh:
> I have one thought, which is that Android currently has options for 32 (high precision), 16 (medium precision) and 11 (low precision) only, and shows an input box for other options (but only if it's already outside of those three + disabled). I think we should either use one of those here, or (better, IMO) add more options to the default available ones on Android to include 14. Either adding medium-low at 14 (and medium-high, maybe, but 16 is already kinda that), or adding a way to select everything in something like 10-19 or 11-16 or such (whatever range iOS allows, probably).

iOS does 10...19 for config and map display and defaults to 13.

garthvh:
And I chose 13 as the default as it would not be precise geolocation data under the California law.

garthvh:
IMG_3060

thebentern:
@mrfyda Let's default to 13 since that has a legal basis, and can we get that option added to the android app?@jamesarich

jamesarich:
> @mrfyda Let's default to 13 since that has a legal basis, and can we get that option added to the android app?@jamesarich

Can we make this an issue on the android repo so it doesn't get lost in the shuffle, pls. Not sure when I'll have some more cycles to burn.

thebentern:
> Can we make this an issue on the android repo so it doesn't get lost in the shuffle, pls. Not sure when I'll have some more cycles to burn.

https://github.com/meshtastic/Meshtastic-Android/issues/1197

akrabu:
I haven't tested the alpha firmware yet, but I have noticed on the iOS app that the fuzzy circles can be very... visually noisy and distracting.  I can't imagine how messy the map would appear with _everyone_ having fuzzy circles by default.  

I'd like to request that the circles only show when a node is tapped on - or at least a way to turn the wider circles off with the options.  Thanks!

garthvh:
> I haven't tested the alpha firmware yet, but I have noticed on the iOS app that the fuzzy circles can be very... visually noisy and distracting. I can't imagine how messy the map would appear with _everyone_ having fuzzy circles by default.
> 
> I'd like to request that the circles only show when a node is tapped on - or at least a way to turn the wider circles off with the options. Thanks!

The whole point is to see at a glance that it is imprecise without having to click on it, the circles are transparent and can overlap fine.

PR #4468, opened 2024-08-15
Router and sensor are impolite

### LINKED ISSUE #4466: [Feature Request]: Please Enable More Frequent Device Metrics/Telemetry for Router Role Devices ONLY in Crowded Mesh
### Platform

NRF52

### Description

The improvements to Metrics/Telemetry broadcasting for busy meshes are great at reducing congestion. 

However, it's also created a bit of an issue for busy mountain top Routers. Routers are by definition meant to be located high and clear of obstructions. This means they are very likely to see more than 40 nodes and also see the highest channel utilization in any given mesh. 

This has created somewhat of an issue after upgrading my main router to 2.4.x. I am now seeing metrics updated maybe once every few days due to the exceptional viewshed it possesses which overlooks the entirety of the south Salish Sea from Vancouver down to Puget Sound.  The channel util at the site was perpetually hovering around 50% based on my last on site visit. This has grown substantially since the unit was installed in March when it was reliably around 20%. Since then I have had a steady decline in frequency of updates corresponding with increase in channel utilization but the metrics were received more than often enough to be useful data. Upon upgrading to 2.4.0 a few weeks ago the updates have almost ceased entirely.

Routers mode devices are also remote and the metrics/telemetry are instrumental in preemptively spotting issues such as under charging and moisture ingress. Having multiple days pass between received telemetry information significantly hampers this ability. 

My ask is that the frequency scaling for Metrics/Telemetry on Router mode ONLY nodes be altered in a future release. Perhaps an interval of every X hours regardless of number of online nodes and channel utilization? Perhaps 3 hours? In a low traffic mesh the current scaling could still be observed. Or simply allow the user specified interval to be observed if it is set to a value greater than 60 minutes?

Once again, this request is for Router mode nodes ONLY.  

I've attached screenshots from my device and another user to illustrate the infrequency of useful telemetry reception. 
CRT1_Infrequent_updates_2
CRT1_Infrequent_Updates

Thank you!

thebentern:
Closes #4466

### DIFF
--- a/src/modules/Telemetry/DeviceTelemetry.cpp
+++ b/src/modules/Telemetry/DeviceTelemetry.cpp
@@ -16,12 +16,14 @@
 int32_t DeviceTelemetryModule::runOnce()
 {
     refreshUptime();
+    bool isImpoliteRole = config.device.role == meshtastic_Config_DeviceConfig_Role_SENSOR ||
+                          config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER;
     if (((lastSentToMesh == 0) ||
          ((uptimeLastMs - lastSentToMesh) >=
           Default::getConfiguredOrDefaultMsScaled(moduleConfig.telemetry.device_update_interval,
                                                   default_telemetry_broadcast_interval_secs, numOnlineNodes))) &&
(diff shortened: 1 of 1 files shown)


GUVWAF (approved):
Very impolite indeed.

PR #4469, opened 2024-08-15
Re-compute correct timeslot on applyModemConfig

### LINKED ISSUE #3892: [Bug] [V3 fix] slotTimeMsec is constant but should be recalculated on startup based on spreading factor and bandwidth
### Category

Other

### Hardware

Not Applicable

### Firmware Version

2.x

### Description

This is a fix for V3 after some discussion in discord.

In `RadioInterface.h` we set some initial values for `slotTimeMsec`, `preambleTimeMsec`, and `maxPacketTimeMsec`, but the second and third of those are recalculated on startup in order to match the actual modem settings in use. However, `slotTimeMsec` is not, so it uses the fixed values for spreading factor and bandwidth that appear in `RadioInterface.h` (sf 9, bw 125) regardless of radio settings. We should update this since the actual LongFast values would create a longer slot time (~77ms rather than ~42), so this could in theory lead to some failures in channel activity detection (since we're not leaving enough time for the process to happen).

The formula may also need or want adjustment, or formulas that use this value. Possibly the maximum contention window size will be too large with the longer slot time, or other issues. This should be tested/simulated, probably.

### Relevant log output

_No response_

thebentern:
Per our conversation in contributors room in Discord
Closes #3892

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -520,6 +520,7 @@ void RadioInterface::applyModemConfig()
     saveChannelNum(channel_num);
     saveFreq(freq + loraConfig.frequency_offset);
 
+    slotTimeMsec = computeSlotTimeMsec(bw, sf);
     preambleTimeMsec = getPacketTime((uint32_t)0);
     maxPacketTimeMsec = getPacketTime(meshtastic_Constants_DATA_PAYLOAD_LEN + sizeof(PacketHeader));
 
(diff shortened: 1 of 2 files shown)


GUVWAF:
Looks good, but we might want to decrease `CWmax` then also. Currently the absolute maximum delay on LongFast before rebroadcasting is 2^8 * `slotTimeMsec` = 10.7 seconds, which will become 19.7 seconds. Maybe we can decrease `CWmax` to 7 such that we get 9.86 seconds?

thebentern:
> Looks good, but we might want to decrease `CWmax` then also. Currently the absolute maximum delay on LongFast before rebroadcasting is 2^8 * `slotTimeMsec` = 10.7 seconds, which will become 19.7 seconds. Maybe we can decrease `CWmax` to 7 such that we get 9.86 seconds?

See this is why I added you :-)

GUVWAF (approved):

PR #4482, opened 2024-08-17
Add BMP388 as a new pressure and temp sensor

dhskinner:
Hi this PR adds BMP388 as a new pressure and temperature sensor for the environment module

Its linked to [meshtastic/protobuf PR#540](https://github.com/meshtastic/protobufs/pull/540) that added the necessary enums to [telemetry.proto](https://github.com/meshtastic/protobufs/blob/master/meshtastic/telemetry.proto)

This follows pretty much same pattern as other sensors, except that the bmp3xx is intended as a singleton - the reason is that I'm intending to integrate interrupt based reads in future to take advantage of this sensors FIFO (this will be a separate/future PR). 

Appreciate your review and any comments, or merge if its ok

Thanks

Dave


fifieldt (approved):

src/modules/Telemetry/EnvironmentTelemetry.cpp
```diff
@@ -54,6 +55,7 @@ AHT10Sensor aht10Sensor;
 MLX90632Sensor mlx90632Sensor;
 DFRobotLarkSensor dfRobotLarkSensor;
 NAU7802Sensor nau7802Sensor;
+extern BMP3XXSensor bmp3xxSensor;
 #ifdef T1000X_SENSOR_EN
```
caveman99, line 59: why the extermal reference?

src/modules/Telemetry/Sensor/BMP3XXSensor.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "../mesh/generated/meshtastic/telemetry.pb.h"
+#include "BMP3XXSensor.h"
+#include "TelemetrySensor.h"
+#include <Adafruit_BMP3XX.h>
+#include <typeinfo>
+
+BMP3XXSensor::BMP3XXSensor() : TelemetrySensor(meshtastic_TelemetrySensorType_BMP3XX, "BMP3XX"){}
+
+BMP3XXSensor bmp3xxSensor;
+
```
caveman99, line 14: Please put that in the telemetry init and reference it from there.

caveman99 (changes requested):

src/modules/Telemetry/EnvironmentTelemetry.cpp
dhskinner: This is fixed now - the reason was BMP388 sensor really should be accessed through only one instance to avoid conflicting calls. The latest commit to this PR adds a singleton wrapper around the the underlying BMP3XXSensor class to sort this properly

src/modules/Telemetry/Sensor/BMP3XXSensor.cpp
dhskinner: Ok that's done

caveman99 (approved):

PR #4483, opened 2024-08-17
Move NeighborInfo interval default to 6 hours (double NodeInfo)

thebentern:
As a part of the default MQTT broker traffic management, we've noticed a disproportionate percentage of packets lately are NeighborInfo. Given that this data is really only useful for mesh topology visualizations, I'm proposing moving it to a 6-hour interval, which is double that of its most similar companion: NodeInfo.


garthvh:
Do 6, twice a day is fine

thebentern:
I agree. Sounds good. So we'll double both the node info broadcast interval default and minimum allowed.

jp-bennett:
@thebentern Take a second look at this. I think you accidentally dropped the ms to second scaling, so Neighborinfo happens about 1000 times more often than intended.

Talie5in:
> @thebentern Take a second look at this. I think you accidentally dropped the ms to second scaling, so Neighborinfo happens about 1000 times more often than intended.

Can confirm, opened a Issue
https://github.com/meshtastic/firmware/issues/4502

PR #4487, opened 2024-08-17
Don't reject network time updates unintentionally

jp-bennett:
Minor flaw in the new time setting logic, was refusing to set time from the network for the first 24 hours after boot. Namely because the default value of 0 was unintentionally indicating that time was set within the first second of boot. Solution is to treat lastSetFromPhoneNtpOrGps of 0 as a special case.


garthvh:
Nobody likes unix epoch

garthvh (approved):

PR #4489, opened 2024-08-17
Don't reject network time updates unintentionally

jp-bennett:


garthvh (approved):

jp-bennett:
This is fairly annoying, and a simple fix. Do we want to re-cut 2.4.3 for it?

PR #4492, opened 2024-08-18
Make it possible to define TCXO and XTAL radio modules within one variant

Nestpebble:
This changes the way that the SX1262 is initialised, so that if it fails to initialise with the DIO3 voltage reference set, it will retry with the voltage set to zero. By this, there is no need to fully define whether the attached radio has TCXO or not during compile.

This is mainly for the pro-micro variant, but will work with other DIY boards or variants that come in both flavours. Potential candidates are the RP2040 modules, as there are some low-cost versions that utilise the same cheap modules.

Huge credit to Todd Herbert (@todd-herbert) for doing so much of the heavy lifting and heavy thinking associated with this. It would not have been possible to do this without you!


thebentern:
This is great! I think there are also TXCO variants in the TLoRA line that have required specific pio targets, so we can perhaps try this on them. Anytime we can consolidate, I'm on board.

Nestpebble:
> This is great! I think there are also TXCO variants in the TLoRA line that have required specific pio targets, so we can perhaps try this on them. Anytime we can consolidate, I'm on board.

I think I've made this universal enough, so just adding `#define TCXO_OPTIONAL to the variant should engage it. It's only for SX1262, but should work with SX126x boards, because thats where this particular switch lives. I did have a think about other variants, but I don't know enough about the PiPico etc.

Anyway, I'm calling this good enough for now.

Nestpebble:
@thebentern I keep getting notifications on this. Do I need to do something to make it merge properly?

PR #4493, opened 2024-08-18
Add 4 bytes of random nonce to PKI

jp-bennett:
Our 32 bits of Nonce really aren't enough. This adds an additional 4 bytes to the raw bytes of PKI packets, then extracts them during decrypt. This brings the total overhead of a PKI packet to 12 bytes.


garthvh (approved):
Probably good advice

PR #4498, opened 2024-08-18
Only accept PKI messages for MQTT downlink for known nodes

GUVWAF:
We're potentially receiving a lot of DMs on an MQTT root topic if we have downlink enabled on a (private) channel, but not all are of interest. Let's only accept PKI messages when we know the transmitter and receiver, as then it's (more) likely they would normally have communicated via a channel we have downlink enabled.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -164,10 +164,14 @@ void MQTT::onReceive(char *topic, byte *payload, size_t length)
 
                     // PKI messages get accepted even if we can't decrypt
                     if (router && p->which_payload_variant == meshtastic_MeshPacket_encrypted_tag &&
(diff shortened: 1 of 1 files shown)


thebentern (approved):

thebentern:
Seems like a good idea. Most of these would get zero-hopped on the default broker anyway, so we'd only be dealing with private channels or a private broker instance.

GUVWAF:
> Most of these would get zero-hopped on the default broker anyway

True, but still it lowers the processing effort of the node :)

jp-bennett:
My only hesitation here is the case where one or both nodes have rolled out of the nodeDB, on a busy mesh.

jp-bennett (approved):

PR #4511, opened 2024-08-20
Add heartbeat LED for HT-VME290 and HT-VME213

todd-herbert:
Not populated on original board, however revisions are now shipping which do have the LED

### DIFF
--- a/variants/heltec_vision_master_e213/pins_arduino.h
+++ b/variants/heltec_vision_master_e213/pins_arduino.h
@@ -3,7 +3,7 @@
 
 #include <stdint.h>
 
-static const uint8_t LED_BUILTIN = -1; // Board has no built-in LED, despite what schematic shows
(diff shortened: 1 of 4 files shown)


variants/heltec_vision_master_e213/pins_arduino.h
```diff
@@ -3,7 +3,7 @@
 
 #include <stdint.h>
 
-static const uint8_t LED_BUILTIN = -1; // Board has no built-in LED, despite what schematic shows
+static const uint8_t LED_BUILTIN = 45; // Board has no built-in LED, despite what schematic shows
```
fifieldt, line 6: Is that comment still valid after this change?

todd-herbert: Good spotting!

fifieldt (approved):

PR #4514, opened 2024-08-20
Fix array out of bounds read.

Mictronics:
Causes lately a warning:
`src/modules/AdminModule.cpp:355:18: warning: iteration 4 invokes undefined behavior [-Waggressive-loop-optimizations]`

`sizeof(r->get_module_config_response.payload_variant.remote_hardware.available_pins);`will return a value of 80.
`meshtastic_RemoteHardwarePin available_pins[4];` = (4+15+1) * 4

Someone more familiar with this part of code please check if `available_pins_count`is what we want here.


Mictronics:
Got this error one time only on first reboot after flashing the fix.
```
ERROR | ??:??:?? 18 [Router] Can't decode protobuf reason='invalid wire_type', pb_msgdesc 0x10080c5c
ERROR | ??:??:?? 18 [Router] Error decoding protobuf module!
ERROR | ??:??:?? 18 [Router] Can't decode protobuf reason='invalid wire_type', pb_msgdesc 0x10080c5c
ERROR | ??:??:?? 18 [Router] Error decoding protobuf module!
```
Following reboots are all clean.

thebentern (approved):

thebentern:
Good find.

PR #4523, opened 2024-08-21
Initial support for RadioMaster Bandit.

gjelsoe:
Everything but the Accelerometer is working.
The Accelerometer is detected as LIS3DH but is a STK8B53 and it's not supported (WIP).

AmbientLighting :
Will be turned on be default.
Change to AmbientLightingThread.h with @todd-herbert so it will turn off lighting when shutdown() is issued.
Different colors for the two buttons can be defined in variant.h.


gjelsoe:
Seems like the scripts are failing with an **ModuleNotFoundError: No module named 'poetry'** error..

thebentern (approved):

PR #4527, opened 2024-08-21
for #4154 Only enable Vext regulator when needed for either Screen or GPS

geeksville:
(just opening now for CI notifications)

notes to self:

## Changes in this PR

* change GPS code to use virtual gpio for pwr enable (so that we can manage more cleverly)
* change TFT code to use a virtual gpio for backlight enable
* cleanup a bunch of redundant variant defs which led to an explosion of unneeded ifdefs
* use virt GPIOs inside screen and GPS
* confirm that power savings works.  Done: with this change when the GPS and screen are both powered down the power consumption is about 10mA lower on the USB 5V in (I haven't yet tested when running on battery, but this matches expected behavior based on the regulator datasheet)
* confirmed that if the screen is needed the regulator is powered up as needed.

## TODO still
 
* test that GPS still works!


thebentern (approved):

GPSFan:
Please make sure that there is some way to keep the GPS powered all the time. In the past setting GPS Update Interval to <=10 sec would do this. The UC6580 GNSS chip in the Wireless Tracker is very good, there are use cases where power consumption is secondary to position accuracy.

geeksville:
@GPSFan no worries on that aspect.  This doesn't change the GPS power up/down stuff at all.  If the GPS was previously enabled it will still be enabled after this change.  The only change here is: when GPS is disabled (and screen is also disabled) turn off the regulator that powers them (to prevent the 10mA drain from that regulator).

PR #4556, opened 2024-08-26
set current time to system time in portduino build

pccr10001:
To have better looking in log for portduino.


src/platform/portduino/PortduinoGlue.cpp
```diff
@@ -1,6 +1,7 @@
 #include "CryptoEngine.h"
 #include "PortduinoGPIO.h"
 #include "SPIChip.h"
+#include "RTC.h"
```
jp-bennett, line 4: Trunk is complaining that these are out of alphabetical order. Can you swap RTC and SPIChip?

pccr10001: update soon

PR #4562, opened 2024-08-26
Don't compare nodeDB macaddr to owner.macaddr, because in rare cases …

### LINKED ISSUE #4559: [Bug]: Apparently finding our own node num in the nodedb can cause us to pick a new nodenum!
### Category

Other

### Hardware

Not Applicable

### Firmware Version

v2.5.0.ab7de7f and 2.4.x

### Description

Per investigation by @dm5tt in https://github.com/meshtastic/firmware/issues/4555#issuecomment-2309754360:

Repeatedly generating a new node ID due to (our) nodenum being found in the internal db is super bad.

Investigate and fix as needed.

### Relevant log output

_No response_

jp-bennett:
…that may be unset.

Should fix #4559

I'm not aware of any edge cases this would break. As I understand it, we always use the real mac address in our owner object, even if using a generated nodeID.


thebentern (approved):

GUVWAF (approved):

geeksville (approved):
LGTM!

PR #4569, opened 2024-08-27
Lilygo T3S3 E-Paper support

mverch67:
Just a new hardware definition is missing in protobuf / architecture.h
(and adding to build matrix and web-flasher)

image

### DIFF
--- a/variants/tlora_t3s3_epaper/platformio.ini
+++ b/variants/tlora_t3s3_epaper/platformio.ini
@@ -14,10 +14,10 @@ build_flags =
   -DEINK_LIMIT_FASTREFRESH=10          ; How many consecutive fast-refreshes are permitted
   -DEINK_LIMIT_RATE_BACKGROUND_SEC=30  ; Minimum interval between BACKGROUND updates
   -DEINK_LIMIT_RATE_RESPONSIVE_SEC=1   ; Minimum interval between RESPONSIVE updates
-;   -D EINK_LIMIT_GHOSTING_PX=2000      ; (Optional) How much image ghosting is tolerated
-  -DEINK_BACKGROUND_USES_FAST          ; (Optional) Use FAST refresh for both BACKGROUND and RESPONSIVE, until a limit is reached.
-  -DEINK_HASQUIRK_GHOSTING             ; Display model is identified as "prone to ghosting"
-  -DEINK_HASQUIRK_WEAKFASTREFRESH      ; Pixels set with fast-refresh are easy to clear, disrupted by sunlight
+  -DEINK_HASQUIRK_VICIOUSFASTREFRESH   ; Identify that pixels drawn by fast-refresh are harder to clear
+  ;-DEINK_LIMIT_GHOSTING_PX=2000        ; (Optional) How much image ghosting is tolerated
+  ;-DEINK_BACKGROUND_USES_FAST          ; (Optional) Use FAST refresh for both BACKGROUND and RESPONSIVE, until a limit is reached.
+
 lib_deps =
   ${esp32s3_base.lib_deps}
   https://github.com/meshtastic/GxEPD2#b202ebfec6a4821e098cf7a625ba0f6f2400292d
--- a/variants/tlora_t3s3_epaper/platformio.ini
+++ b/variants/tlora_t3s3_epaper/platformio.ini
@@ -1,0 +1,23 @@
+[env:tlora-t3s3-epaper]
+extends = esp32s3_base
+board = tlora-t3s3-v1
+board_check = true
+upload_protocol = esptool
+
+build_flags = 
+  ${esp32_base.build_flags} -D TLORA_T3S3_EPAPER -I variants/tlora_t3s3_epaper
+  -DGPS_POWER_TOGGLE
+  -DEINK_DISPLAY_MODEL=GxEPD2_213_BN
+  -DEINK_WIDTH=250
+  -DEINK_HEIGHT=122
+  -DUSE_EINK_DYNAMICDISPLAY            ; Enable Dynamic EInk
+  -DEINK_LIMIT_FASTREFRESH=10          ; How many consecutive fast-refreshes are permitted
+  -DEINK_LIMIT_RATE_BACKGROUND_SEC=30  ; Minimum interval between BACKGROUND updates
+  -DEINK_LIMIT_RATE_RESPONSIVE_SEC=1   ; Minimum interval between RESPONSIVE updates
+;   -D EINK_LIMIT_GHOSTING_PX=2000      ; (Optional) How much image ghosting is tolerated
+  -DEINK_BACKGROUND_USES_FAST          ; (Optional) Use FAST refresh for both BACKGROUND and RESPONSIVE, until a limit is reached.
+  -DEINK_HASQUIRK_GHOSTING             ; Display model is identified as "prone to ghosting"
+  -DEINK_HASQUIRK_WEAKFASTREFRESH      ; Pixels set with fast-refresh are easy to clear, disrupted by sunlight
+lib_deps =
+  ${esp32s3_base.lib_deps}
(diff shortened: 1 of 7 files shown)


thebentern (approved):

variants/tlora_t3s3_epaper/platformio.ini
```diff
@@ -0,0 +1,23 @@
...
+  -DEINK_WIDTH=250
+  -DEINK_HEIGHT=122
+  -DUSE_EINK_DYNAMICDISPLAY            ; Enable Dynamic EInk
+  -DEINK_LIMIT_FASTREFRESH=10          ; How many consecutive fast-refreshes are permitted
+  -DEINK_LIMIT_RATE_BACKGROUND_SEC=30  ; Minimum interval between BACKGROUND updates
+  -DEINK_LIMIT_RATE_RESPONSIVE_SEC=1   ; Minimum interval between RESPONSIVE updates
+;   -D EINK_LIMIT_GHOSTING_PX=2000      ; (Optional) How much image ghosting is tolerated
+  -DEINK_BACKGROUND_USES_FAST          ; (Optional) Use FAST refresh for both BACKGROUND and RESPONSIVE, until a limit is reached.
+  -DEINK_HASQUIRK_GHOSTING             ; Display model is identified as "prone to ghosting"
+  -DEINK_HASQUIRK_WEAKFASTREFRESH      ; Pixels set with fast-refresh are easy to clear, disrupted by sunlight
```
todd-herbert, line 20: I'd maybe go easy on some of these values. I don't have this device, but I believe it's the same display as in Heltec Wireless Paper V1.1. From memory, the fast-refresh operation right now is quite vicious. It'd definitely consider using the same profile as with Wireless Paper V1.1:
```
  -D USE_EINK_DYNAMICDISPLAY            ; Enable Dynamic EInk
  -D EINK_LIMIT_FASTREFRESH=5           ; How many consecutive fast-refreshes are permitted
  -D EINK_LIMIT_RATE_BACKGROUND_SEC=30  ; Minimum interval between BACKGROUND updates
  -D EINK_LIMIT_RATE_RESPONSIVE_SEC=1   ; Minimum interval between RESPONSIVE updates
  ;-D EINK_BACKGROUND_USES_FAST         ; Disabled, so display will perform a full refresh once user input ceases
  -D EINK_HASQUIRK_VICIOUSFASTREFRESH   ; Identify that pixels drawn by fast-refresh are harder to clear

  ; I don't remember how bad the image ghosting is, but you might need this
  ;-D EINK_LIMIT_GHOSTING_PX=2000
  ```
  I don't think I made any attempt to optimize the fast-refresh operation for this display. It's using the values the official GxEPD2 library uses. It might be possible room to get a bit more performance out of the panel with some careful adjustments.

Comment (mverch67) [reply variants/tlora_t3s3_epaper/platformio.ini:20]: I didn't see any ghosting and the fastrefresh=10 looks fine to me. I copied from the Heltec E213 (it's PC1 not BN), but let me try your suggestion if I can spot any difference...

Comment (todd-herbert) [reply variants/tlora_t3s3_epaper/platformio.ini:20]: It's definitely been a few months since I've powered on that older Wireless Paper model, so I don't remember exactly how rough things were. I just have memories of it tolerating fast refresh fairly poorly is all. It's also not impossible that the panel used with this T3S3 model is a slightly different revision too, with different characteristics, so will have to defer to your first-hand experience with it!

Comment (todd-herbert) [reply variants/tlora_t3s3_epaper/platformio.ini:20]: Removing `EINK_BACKGROUND_USES_FAST` will also use full-refresh for any low priority display updates, which is another way to potentially preserve the display health, if it seems like it might be an issue.

Comment (mverch67) [reply variants/tlora_t3s3_epaper/platformio.ini:20]: Uuuh, it got much worse and is crashing now and booting in a loop
`DEBUG | ??:??:?? 72 [RadioIf] SX126x standby failed with error -707`

Comment (todd-herbert) [reply variants/tlora_t3s3_epaper/platformio.ini:20]: As a result of changing the profile? I definitely don't know why it would have had any impact like that. Those macros just influence some high-level decision making in when to use full vs fast refresh. Nothing at a driver level.

**Comment by canadamadman** [comment]
i have device. would like to help with testing. i don't know how to code, well, but I'm willing to sacrifice my device in the name of development.

File: variants/tlora_t3s3_epaper/platformio.ini (hunk 1 above)
Comment (mverch67) [reply variants/tlora_t3s3_epaper/platformio.ini]: I've updated the defines. Looks ok now.

File: variants/tlora_t3s3_epaper/platformio.ini [hunk 2]
```diff
@@ -14,10 +14,10 @@ build_flags =
   -DEINK_LIMIT_FASTREFRESH=10          ; How many consecutive fast-refreshes are permitted
   -DEINK_LIMIT_RATE_BACKGROUND_SEC=30  ; Minimum interval between BACKGROUND updates
   -DEINK_LIMIT_RATE_RESPONSIVE_SEC=1   ; Minimum interval between RESPONSIVE updates
-;   -D EINK_LIMIT_GHOSTING_PX=2000      ; (Optional) How much image ghosting is tolerated
-  -DEINK_BACKGROUND_USES_FAST          ; (Optional) Use FAST refresh for both BACKGROUND and RESPONSIVE, until a limit is reached.
-  -DEINK_HASQUIRK_GHOSTING             ; Display model is identified as "prone to ghosting"
-  -DEINK_HASQUIRK_WEAKFASTREFRESH      ; Pixels set with fast-refresh are easy to clear, disrupted by sunlight
+  -DEINK_HASQUIRK_VICIOUSFASTREFRESH   ; Identify that pixels drawn by fast-refresh are harder to clear
+  ;-DEINK_LIMIT_GHOSTING_PX=2000        ; (Optional) How much image ghosting is tolerated
+  ;-DEINK_BACKGROUND_USES_FAST          ; (Optional) Use FAST refresh for both BACKGROUND and RESPONSIVE, until a limit is reached.
```
todd-herbert, line 19: I'm not sure it'd handle constantly doing 10 fast refreshes back-to-back with the current driver, but now that `EINK_BACKGROUND_USES_FAST` is no-longer defined, that sort of back-to-back fast refresh will be much rarer; so the display can probably tolerate it.

todd-herbert (approved):
Just keep an eye in particular on the display's border, for darkening and white-noise as an early warning sign.

variants/tlora_t3s3_epaper/platformio.ini
mverch67: You mean that not optimal values can physically damage the display? Or is it "aging" faster? Did you ever manage to "break" an e-ink using wrong settings?

todd-herbert: > You mean that not optimal values can physically damage the display?

If left unchecked yes, although in my experience, the image quality significantly deteriorates before damage is irreversible.

> Or is it "aging" faster?

Quite possibly, although unless something is set *really* wrong, I doubt it's too much of an issue.

> Did you ever manage to "break" an e-ink using wrong settings?

I've never had one pushed to breaking point before, although I've run them down into some pretty rough states. If things are going wrong, it tends to become visible quite quickly.

> I've observed the display now for a couple of hours. I don't see something suspicious.

A couple of hours would probably be long enough for any problems to start to show, so it's a good sign.

> Every now and then it is doing a full refresh even without pressing a button (probably due to EINK_HASQUIRK_VICIOUSFASTREFRESH).

That's the behavior when `EINK_BACKGROUND_USES_FAST` is not defined. If an urgent display update happens, like in response to button press, fast-refresh is used, however the display will redraw the same screen image ~30 seconds later using a full refresh. Any non-urgent updates will also use full-refresh.

This configuration in particular is what's protecting this display right now. The frequent full-refreshing during downtime is helping prevent charge imbalances accumulating from this displays fast-refresh operation, which is currently quite aggressive. 

There's a flowchart over on #3316 if you ever have to try make sense of my code spaghetti there.

PR #4508, opened 2024-08-20
Remove heltec-specific gps code from main.cpp

fifieldt:
After the recent GPS power work we have an clear set of definitions for turning GPS on and off. Rather than manipulating specific heltec tracker-related pins in main setup, the relevant power management code in the GPS module will turn things on/off later as needed.

PR #4588, opened 2024-08-30
Nuke proxy variable

S5NC:
Functionality consistent, was only a proxy (source: https://github.com/search?q=repo%3Ameshtastic%2Ffirmware+VGNSS_CTRL&type=code) after https://github.com/meshtastic/firmware/pull/4508

Surely there is an extension to detect this type of thing?


fifieldt (approved):
Thanks!

caveman99:
This is remains from auto-detecting if we had a V1.0 or V1.1 board. Since we have separate targets now these are moot.

PR #4592, opened 2024-08-30
Set high priority for text messages

### LINKED ISSUE #4576: [Feature Request]: Always Prioritize Text Messages
### Platform

### Description

Currently when a network becomes congested we don't do a good enough job protecting text messages and the related routing from position and telemetry packets.

Prioritize text messages above all other packet types, especially on the known key.

GUVWAF:
Use the new `HIGH` priority for text messages in the Tx queue. Fixes #4576.

When working on this, I saw that `fixPriority()` was called in `MeshPacketQueue::enqueue()`, but that is always *after* encryption, while it was checking `p->decoded` without checking the payload variant. Not sure why this never failed, but I think it wouldn't set the ACK priority correctly before. I therefore moved it to before encryption (confirmed there's no other MeshPacketQueue than the `txQueue`).

### DIFF
--- a/src/mesh/MeshPacketQueue.cpp
+++ b/src/mesh/MeshPacketQueue.cpp
@@ -40,19 +40,22 @@ void fixPriority(meshtastic_MeshPacket *p)
     // We might receive acks from other nodes (and since generated remotely, they won't have priority assigned.  Check for that
     // and fix it
     if (p->priority == meshtastic_MeshPacket_Priority_UNSET) {
-        // if acks give high priority
         // if a reliable message give a bit higher default priority
-        p->priority = (p->decoded.portnum == meshtastic_PortNum_ROUTING_APP)
-                          ? meshtastic_MeshPacket_Priority_ACK
-                          : (p->want_ack ? meshtastic_MeshPacket_Priority_RELIABLE : meshtastic_MeshPacket_Priority_DEFAULT);
+        p->priority = (p->want_ack ? meshtastic_MeshPacket_Priority_RELIABLE : meshtastic_MeshPacket_Priority_DEFAULT);
+        if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
+            // if acks/naks give very high priority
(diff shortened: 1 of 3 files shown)


garthvh:
Closes https://github.com/meshtastic/firmware/issues/4576

thebentern (approved):

gitbisector:
Happy to hear you are taking another look at this otherwise boring piece of code too. 

Been profiling TX queue occupancy and how long messages sit in there. Upgraded to master this morning including this change. Can see priority 10, 64, 70, 100 and 120 in use now. Still, some messages get stuck for more than 15 minutes while others of same priority bypass it. Started with measuring through msg counts rather than time in my graphs as GUVWAF noted time might stay at 0. 
image

I do have working code that uses stable same-priority order without adding any storage but can't use the heap implementation with that. As station under my control does have rx_time I'll use that to do before/after measurements.

gitbisector:
BTW, expect  your commit that gives proper priority to ack and raises text message above telemetry is going to improve user experience a bunch.

GUVWAF:
Thanks for the analysis, that's interesting. I might have underestimated how big the influence of this is in really busy meshes. Would be nice to have a cost-effective solution, and I'm wondering if we should make more priority levels for certain messages, e.g. responses to requests.

gitbisector:
Have some log proof now this is backwards. Even says: return 'true' if pi1 is ordered before p2.

`/// @return "true" if "p1" is ordered before "p2"
bool CompareMeshPacketFunc(const meshtastic_MeshPacket *p1, const meshtastic_MeshPacket *p2)
{
    assert(p1 && p2);
    auto p1p = getPriority(p1), p2p = getPriority(p2);

    // If priorities differ, use that
    // for equal priorities, order by id (older packets have higher priority - this will briefly be wrong when IDs roll over but
    // no big deal)
    return (p1p != p2p)
               ? (p1p < p2p)                                                 // prefer bigger priorities
               : ((p1->id & ID_COUNTER_MASK) >= (p2->id & ID_COUNTER_MASK)); // Mask to counter portion, prefer smaller packet ids
}`

PR #4608, opened 2024-09-01
Same priority packets processed in enqueue order

gitbisector:
@GUVWAF, 

This change fixes the priority comparator to greater-than and treats packets within a single priority in fifo order.

Finally seeing some acks again in our very busy mesh and hours long logging shows no more huge latency tail for random packets that happen to have a high sequence number.

### DIFF
--- a/src/mesh/MeshPacketQueue.cpp
+++ b/src/mesh/MeshPacketQueue.cpp
@@ -16,13 +16,9 @@ bool CompareMeshPacketFunc(const meshtastic_MeshPacket *p1, const meshtastic_Mes
 {
     assert(p1 && p2);
     auto p1p = getPriority(p1), p2p = getPriority(p2);
-
     // If priorities differ, use that
-    // for equal priorities, order by id (older packets have higher priority - this will briefly be wrong when IDs roll over but
-    // no big deal)
-    return (p1p != p2p)
-               ? (p1p < p2p)                                                 // prefer bigger priorities
-               : ((p1->id & ID_COUNTER_MASK) >= (p2->id & ID_COUNTER_MASK)); // Mask to counter portion, prefer smaller packet ids
+    // for equal priorities, order stays as enqueued
+    return (p1p > p2p); // prefer bigger priorities
 }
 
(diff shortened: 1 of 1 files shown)


src/mesh/MeshPacketQueue.cpp
```diff
@@ -16,13 +16,9 @@ bool CompareMeshPacketFunc(const meshtastic_MeshPacket *p1, const meshtastic_Mes
...
     auto p1p = getPriority(p1), p2p = getPriority(p2);
-
     // If priorities differ, use that
-    // for equal priorities, order by id (older packets have higher priority - this will briefly be wrong when IDs roll over but
-    // no big deal)
-    return (p1p != p2p)
-               ? (p1p < p2p)                                                 // prefer bigger priorities
-               : ((p1->id & ID_COUNTER_MASK) >= (p2->id & ID_COUNTER_MASK)); // Mask to counter portion, prefer smaller packet ids
+    // for equal priorities, order stays as enqueued
+    return (p1p > p2p); // prefer bigger priorities
```
GUVWAF, line 21: Indeed it has been wrong the whole time!

I still think that if the priorities are equal, we should prioritize other node’s packets above those of ourselves for the reasons I laid out here: https://github.com/meshtastic/firmware/pull/4561#issuecomment-2313059473. Text messages have high priority now, so it should not really influence normal messaging on a node that is busy forwarding other people’s packets. Also if another node rebroadcasts before you, other people's messages will be removed anyway. 

I think this can be done by: 
`return (p1p != p2p) ? (p1p > p2p) : (getFrom(p1) != nodeDB->getNodeNum() && getFrom(p2) == nodeDB->getNodeNum())`.

GUVWAF (commented):
Really like this solution, it’s even simpler than the previous implementation. I left one comment though.

src/mesh/MeshPacketQueue.cpp
gitbisector: I'll make the change and test it.

gitbisector:
> Really like this solution, it’s even simpler than the previous implementation. I left one comment though.
> 
> Also: can you please sign the CLA?

Working on the CLA, expect legal rep to reply tomorrow as today is a holiday here.

PR #4610, opened 2024-09-02
Fix display of certain Unicode symbols

Coloradohusky:
This PR fixes the display of all characters within the 0xC2.. Unicode character range, from ¡ to ¿ (including the degree symbol °, which previously showed up as ¿). Previously, this fix was only applied to OLED_PL, OLED_UA, and OLED_RU, but now it applies to other OLEDs as well. Tested on my RAK1921 and my Heltec Tracker.

### DIFF
--- a/src/graphics/Screen.h
+++ b/src/graphics/Screen.h
@@ -309,7 +309,7 @@ class Screen : public concurrency::OSThread
     static char customFontTableLookup(const uint8_t ch)
     {
         // UTF-8 to font table index converter
-        // Code form http://playground.arduino.cc/Main/Utf8ascii
+        // Code from http://playground.arduino.cc/Main/Utf8ascii
         static uint8_t LASTCHAR;
         static bool SKIPREST; // Only display a single unconvertable-character symbol per sequence of unconvertable characters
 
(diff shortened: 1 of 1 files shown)


src/graphics/Screen.h
```diff
@@ -322,13 +322,19 @@ class Screen : public concurrency::OSThread
...
+        switch(last) {
         case 0xC2: {
             SKIPREST = false;
             return (uint8_t)ch;
         }
+        }
+
+        // We want to strip out prefix chars for two-byte char formats
+        if (ch == 0xC2)
+            return (uint8_t)0;
```
todd-herbert, line 334: I think this is causing issues with `OLED_PL`, `OLED_UA` and `OLED_RU`. If using those variants, `switch(last)` is being closed early here, leaving their `case` statements floating loose in the method.

todd-herbert (changes requested):

src/graphics/Screen.h
Coloradohusky: You're right, added the switches back where they belong

todd-herbert (approved):

PR #4613, opened 2024-09-02
Add MAX17048 lipo fuel gauge

dhskinner:
Hi this PR adds MAX17048 lipo fuel gauge as a new battery sensor for the telemetry & power module

Its linked to https://github.com/meshtastic/protobufs/pull/540 that added the necessary enums to [telemetry.proto](https://github.com/meshtastic/protobufs/blob/master/meshtastic/telemetry.proto)

This follows pretty much same pattern as other Telemetry sensors, as well as adding requisite code to integrate this sensor into the device power functionality in similar fashion to the axp192 power management IC

Appreciate your review and any comments

Thanks

Dave

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -146,6 +146,7 @@ lib_deps =
   adafruit/Adafruit VEML7700 Library@^2.1.6
   adafruit/Adafruit SHT4x Library@^1.0.4
   adafruit/Adafruit TSL2591 Library@^1.4.5
+  adafruit/Adafruit MAX1704X@^1.0.3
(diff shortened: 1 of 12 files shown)


caveman99:
Code looks fine in principle. A few targets are not building. You need help fixing those?

dhskinner:
> Code looks fine in principle. A few targets are not building. You need help fixing those?

Thanks for the offer much appreciated / it took a little longer than expected but finally got everything going ok... just needed to get the right combination of compiler directives. All works now so when the tests have run if you could take a look that would be great

Thanks again

dhskinner:
Ran into [this issue here](https://github.com/orgs/community/discussions/5634) - will resubmit another PR for this

PR #4621, opened 2024-09-03
fix RF switch for Tracker E

caveman99:
this creates the same switch matrix we originally had in godmode. Still not working on my version.

### DIFF
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -232,8 +232,7 @@ template <typename T> void LR11x0Interface<T>::startReceive()
 
     // We use a 16 bit preamble so this should save some power by letting radio sit in standby mostly.
     // Furthermore, we need the PREAMBLE_DETECTED and HEADER_VALID IRQ flag to detect whether we are actively receiving
-    int err = lora.startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF,
-                                RADIOLIB_IRQ_RX_DEFAULT_FLAGS | RADIOLIB_IRQ_PREAMBLE_DETECTED | RADIOLIB_IRQ_HEADER_VALID,
+    int err = lora.startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF, RADIOLIB_IRQ_RX_DONE,
                                 0); // only RX_DONE IRQ is needed, we'll check for RADIOLIB_LR11X0_IRQ_PREAMBLE_DETECTED and
                                     // RADIOLIB_LR11X0_IRQ_SYNC_WORD_HEADER_VALID in isActivelyReceiving
     assert(err == RADIOLIB_ERR_NONE);
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -50,11 +50,23 @@ template <typename T> bool LR11x0Interface<T>::init()
 
     limitPower();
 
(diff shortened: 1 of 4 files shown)


caveman99:
This should be merged since at the current state SX1262 and SX1280 won't work either.

src/mesh/SX126xInterface.cpp
```diff
@@ -267,9 +264,8 @@ template <typename T> void SX126xInterface<T>::startReceive()
 
     // We use a 16 bit preamble so this should save some power by letting radio sit in standby mostly.
     // Furthermore, we need the PREAMBLE_DETECTED and HEADER_VALID IRQ flag to detect whether we are actively receiving
-    int err = lora.startReceiveDutyCycleAuto(preambleLength, 8,
-                                             RADIOLIB_SX126X_IRQ_RX_DEFAULT | RADIOLIB_SX126X_IRQ_PREAMBLE_DETECTED |
-                                                 RADIOLIB_SX126X_IRQ_HEADER_VALID);
+    int err = lora.startReceiveDutyCycleAuto(
+        preambleLength, 8, RADIOLIB_IRQ_RX_DEFAULT_FLAGS | RADIOLIB_IRQ_PREAMBLE_DETECTED | RADIOLIB_IRQ_HEADER_VALID);
```
GUVWAF, line 268: `RADIOLIB_IRQ_HEADER_VALID` is now already in `RADIOLIB_IRQ_RX_DEFAULT_FLAGS`, so can be safely removed here and for the SX128x.

src/mesh/LR11x0Interface.cpp
```diff
@@ -218,9 +232,9 @@ template <typename T> void LR11x0Interface<T>::startReceive()
 
     // We use a 16 bit preamble so this should save some power by letting radio sit in standby mostly.
     // Furthermore, we need the PREAMBLE_DETECTED and HEADER_VALID IRQ flag to detect whether we are actively receiving
-    int err = lora.startReceive(
-        RADIOLIB_LR11X0_RX_TIMEOUT_INF, RADIOLIB_LR11X0_IRQ_RX_DONE,
-        0); // only RX_DONE IRQ is needed, we'll check for PREAMBLE_DETECTED and HEADER_VALID in isActivelyReceiving
+    int err = lora.startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF, RADIOLIB_IRQ_RX_DONE,
```
GUVWAF, line 235: Latest two commits in RadioLib make this consistent with SX126x again. If you point it to that commit, you can use: 
`startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF, RADIOLIB_IRQ_RX_DEFAULT_FLAGS)`.

thebentern (approved):

GUVWAF (approved):
Looks good. This also includes the fix for the header error in RadioLib, so we should also get less CRC failures now :)

caveman99:
> Looks good. This also includes the fix for the header error in RadioLib, so we should also get less CRC failures now :)

My hope was that this fix (header error) would make my T1000E work again, but alas its stilld eaf with this 
firmware.

Talie5in:
> > Looks good. This also includes the fix for the header error in RadioLib, so we should also get less CRC failures now :)
> 
> My hope was that this fix (header error) would make my T1000E work again, but alas its still deaf with this firmware.

Hey @caveman99 - i was just doing iterative testing through 2.5.0 TP versions, and this is where the t1000e breaks for me after this change.... (dont hear anything other then a really close by Rak4631).

If i go back to 2.5.0.33eb073 i get lots of nodes showing up...
Can see me going through it here - https://discord.com/channels/867578229534359593/1039674919346249728/1283614633311080582

PR #4623, opened 2024-09-04
Gather canned message magic numbers into header defines.

robertfisk:
The canned message module accepts a bunch of magic numbers from input modules to perform system actions and navigation. This PR bundles them together as header defines for easier reference in `CannedMessageModule.h`.

Note that we are not touching any magic numbers used internally by the keyboards / input modules, or any standard ascii control characters.


fifieldt:
This is a good idea, many thanks for submitting it.

I'm not so familiar with this module, but two questions:
1. src/modules/CannedMessageModule.h is wrapped in #ifdef HAS_SCREEN , so just including the header doesn't necessarily result in these variables being available. If there's a chance any of the places these variables need to be used could be triggered on a device without a screen, this should be addressed.
2. Canned Message Module seems like a convenient place to put this, but are there other options worth considering?

caveman99 (approved):

caveman99:
These could go int inputbroker as well, since we may need them wothout canned messages in the future. I'd like to move any input device into inputbroker. Canned Message is the only client at the moment, but doesn't need to stay the only client.

fifieldt (approved):

robertfisk:
I've moved the defines to InputBroker.h now as suggested. This way doesn't require an additional #include in the various keyboards.

HarukiToreda:
> I've moved the defines to InputBroker.h now as suggested. This way doesn't require an additional #include in the various keyboards.

would this perhaps fix a known issue when the environmental screen is enabled the left and righ buttons on the keyboard become unresponsive and can't switch to the next or previous frames? or is this an unrelated issue?

caveman99 (approved):

robertfisk:
> would this perhaps fix a known issue when the environmental screen is enabled the left and righ buttons on the keyboard become unresponsive and can't switch to the next or previous frames? or is this an unrelated issue?

I'm not familiar with the environment screen or that bug sorry. I suspect it is unrelated.

PR #4627, opened 2024-09-05
Fix NMEA message settings for AG3335

fifieldt:
In collaboration with the great @GPSFan, we present fixed NMEA message configuration for the AG3335.

"only RMC & GGA are used by Meshtastic", and added some delays to see if it assists.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -507,18 +507,28 @@ bool GPS::setup()
             delay(250);
         } else if (gnssModel == GNSS_MODEL_AG3335) {
 
-            _serial_gps->write("$PAIR066,1,0,1,0,0,1*3B"); // Enable GPS+GALILEO+NAVIC
+            _serial_gps->write("$PAIR066,1,0,1,0,0,1*3B");
+            delay(250); // Enable GPS+GALILEO+NAVIC
 
-            // Configure NMEA (sentences will output once per fix)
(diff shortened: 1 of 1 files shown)


GPSFan:
This issue may need some hardware debugging, if anyone has an opened Seeed T1000E who could loan it for a while it might be useful.

src/gps/GPS.cpp
```diff
@@ -507,18 +507,28 @@ bool GPS::setup()
...
+            delay(250); // GSV OFF
+            _serial_gps->write("$PAIR062,4,1*3B");
+            delay(250); // RMC ON
+            _serial_gps->write("$PAIR062,5,0*3B");
+            delay(250); // VTG OFF
+            _serial_gps->write("$PAIR062,6,0*38");
+            delay(250); // ZDA ON
+
+            _serial_gps->write("$PAIR080,1*2F");
+            delay(250); // Fitness mode!
```
caveman99, line 530: I couldn't make these commands work on the T1000E either, but noticed the CR/LF is missing. (\r\n) Even with those they show no effect. Are we sure TX is even connected on this chip?

fifieldt: See comments in gnss on discourse today. I can't get any commands working either :(

fifieldt: The CRLF does need yo be fixed though :)

caveman99:
Closing because of merge conflicts, and manually testing this it doesn't make a difference anyway.

PR #4628, opened 2024-09-05
Added support for DIY waveshare Core1262-868M + esp32doit-devkit-v1

zelifcam:
**Added support for DIY waveshare Core1262-868M + esp32doit-devkit-v1**

Common DIY US Amazon parts with optional OLED, GPS ,BUTTON, NOTIFY. 
Jokingly referred to as the Amazon Special (aspec).


src/mesh/generated/meshtastic/mesh.pb.h
```diff
@@ -199,6 +199,7 @@ typedef enum _meshtastic_HardwareModel {
     /* M5 esp32 based MCU modules with enclosure, TFT and LORA Shields. All Variants (Basic, Core, Fire, Core2, Paper) https://m5stack.com/ */
     meshtastic_HardwareModel_M5STACK_COREBASIC = 77,
     meshtastic_HardwareModel_M5STACK_CORE2 = 78,
+    meshtastic_HardwareModel_ASPEC = 79,
```
fifieldt, line 202: These are generated from https://github.com/meshtastic/protobufs - probably want to propose this change there instead :)

zelifcam: My mistake.  It's actually not that important to define it as a new hardware device.  I think it falls in line with the DIY that already exists.  I'm removing the changes to the other files related to it being a new hardware device.

PR #4638, opened 2024-09-06
Add missing linefeeds to gps code

fifieldt:
As reported by @caveman99, the required CRLFs were missing from the AG3335 setup code.


GPSFan:
@caveman99 Did this resolve the issue with the L89 not responding to commands?

Issue #4384, opened 2024-08-04
[Bug]: "Wake On Tap Or Motion" setting not being honored


blarson:
### Category

Other

### Hardware

Rak4631

### Firmware Version

2.4.0

### Description

The device does not appear to honor the Display "Wake On Tap Or Motion" configuration setting when running in Client role (other roles not tested).

I've tried setting and clearing this flag via the Apple iPhone app and via the web client, and neither appears to work. The device continues to wake the screen on motion, regardless of whether the option is set or cleared.

It also does not appear to reboot after adjusting "Wake On Tap Or Motion" alone. It does, however, reboot when applying other Display settings.

garthvh:
Have you installed an accelerometer?

blarson:
Yes, the ST LIS3DH accelerometer on the RAK19026 base board.

It always wakes on motion, even with the setting disabled.

Ruledo:
I just realized this same bug with v2.5 and RAK1904 installed.
The setting in the client is not honored by the device.

Ruledo:
I tested the web client as well, same result.
The toggle appears to be set off but it still wakes on motion and when rebooted the setting toggle is enabled again.

PR #4642, opened 2024-09-06
tryfix #4384

caveman99:
don't assume we want that functionality if the Accelerometer was found. This is only for T-Watch


src/main.cpp
```diff
@@ -635,8 +635,6 @@ void setup()
 #if !MESHTASTIC_EXCLUDE_I2C
 #if !defined(ARCH_PORTDUINO) && !defined(ARCH_STM32WL) && !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR
     if (acc_info.type != ScanI2C::DeviceType::NONE) {
-        config.display.wake_on_tap_or_motion = true;
```
thebentern, line 638: We may consider adding this to the RAK4631 default config, since the WisMesh pocket comes with it. It's kind of a neat feature to enable OOTB. 😄

thebentern (approved):

src/main.cpp
caveman99: i actually moved these to the default config for t-watch, feel free to add there.

PR #4643, opened 2024-09-06
Adds the ok_to_mqtt bit

jp-bennett:
This adds the ok_to_mqtt bit on outgoing messages, as set by the device config. It also checks for the bit on incoming messages before uploading to an MQTT server with a known key.

Depends on https://github.com/meshtastic/protobufs/pull/573


caveman99 (approved):

caveman99:
Principle is ok, CI carnage needs to be addressed :-)

src/mesh/Router.cpp
```diff
@@ -424,6 +424,9 @@ meshtastic_Routing_Error perhapsEncode(meshtastic_MeshPacket *p)
 
     // If the packet is not yet encrypted, do so now
     if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
+        p->decoded.has_ok_to_mqtt = true;
+        p->decoded.ok_to_mqtt = config.lora.config_ok_to_mqtt;
```
GUVWAF, line 428: Here, based on your setting, you decide whether other people's packets will be OK to MQTT.

GUVWAF (changes requested):
Besides my comment, isn't there a better way to set this at only one place? E.g. in `MeshService::sendToMesh()`? 
I think this would then also account for MQTT proxy.

**Edit**: I now see MQTT proxy is already accounted for with this, but anything that goes over the PhoneAPI is not, so e.g. if someone is doing a serial to MQTT bridge it can't honor the setting.

caveman99 (changes requested):
As discussed, can we introduce a bitflag field instead?

src/mesh/Router.cpp
jp-bennett: As many of these as I added, it's not surprising I got the logic wrong in one of them.

jp-bennett:
> As discussed, can we introduce a bitflag field instead?

I love the idea, but I forgot how ridiculous Protobufs is. There is no uint8 type. The official recommendation is to use uint32 and hope it only takes only one or two bytes. *sigh*

src/mesh/Router.cpp
jp-bennett: > Here, based on your setting, you decide whether other people's packets will be OK to MQTT.

I was under the impression that a packet that got to this point, still decoded, would have to come from a local source, either phone or firmware. How does this code path actually get taken?

GUVWAF: We try to decrypt every received packet (in `Router::handleReceived()`), because we might need to do something with the payload. Then the only place it gets encrypted again is here, just before transmitting. So every packet we could decrypt would still be decrypted here.

jp-bennett: Got it. So just need to add the check that it is from the local node.

GUVWAF: Yes. And if you have it here, it doesn't need to be set also in the modules.

thebentern (approved):

src/mesh/Router.cpp
```diff
@@ -378,6 +378,8 @@ bool perhapsDecode(meshtastic_MeshPacket *p)
         // parsing was successful
         p->which_payload_variant = meshtastic_MeshPacket_decoded_tag; // change type to decoded
         p->channel = chIndex;                                         // change to store the index instead of the hash
+        if (p->decoded.has_bitfield)
+            p->decoded.want_response |= p->decoded.bitfield & 1 << 1;
```
GUVWAF, line 382: Had to look up whether `<<` is actually before `&`, can we add brackets? And maybe a macro `BITFIELD_WANT_RESPONSE_SHIFT` and `BITFIELD_OK_TO_MQTT_SHIFT` (=0)?

jp-bennett: > Had to look up whether `<<` is actually before `&`, can we add brackets? And maybe a macro `BITFIELD_WANT_RESPONSE_SHIFT` and `BITFIELD_OK_TO_MQTT_SHIFT` (=0)?

Done.

GUVWAF: Don't want to be nitpicky here, but I think we should use them in `MQTT.cpp` as well, then it's immediately clear what it means.

jp-bennett: > Don't want to be nitpicky here, but I think we should use them in `MQTT.cpp` as well, then it's immediately clear what it means.

Lol, Came to that conclusion, too.

GUVWAF (approved):

PR #4647, opened 2024-09-07
Simplify GNSS Probe code

fifieldt:
This patch takes inspiration from our I2CDetect code where we have many sensors that can be detected rather simply. It creates a new macro,
PROBE_SIMPLE(Chip name, Command to run, response, Driver, timeout)

and converts existing simple cases to use this macro.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -1184,6 +1184,15 @@ int GPS::prepareDeepSleep(void *unused)
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


fifieldt:
Not expecting any impact from this change since it's just moving code around, but for good measure I have tested on:
tbeam (ublox)
heltec-wireless-tracker (UC6580)
tracker-t1000-e (AG3335)
wio-tracker-wm1110 (L76K)

and it behaves exactly the same as before :)

caveman99 (approved):

PR #4660, opened 2024-09-09
Fix for MESHTASTIC_MINIMIZE_BUILD

dhskinner:
This is a minor fix so minimal build works when any/all of these flags are set:

- MESHTASTIC_MINIMIZE_BUILD
- MESHTASTIC_EXCLUDE_I2C
- MESHTASTIC_EXCLUDE_GPS
- MESHTASTIC_EXCLUDE_SCREEN

No change to functionality


fifieldt (approved):
makes sense, thanks!

caveman99 (approved):

PR #4666, opened 2024-09-09
Minor fix

fifieldt:
It turns out setting a map value with the index notation causes an lookup that can be avoided with emplace. Apply this to one line in the StoreForward module.


caveman99 (approved):

caveman99:
the Portduino change needs to be refactored methinks.

thebentern (approved):

PR #4679, opened 2024-09-11
Add option to preserve private key for factory reset (config)

thebentern:


jp-bennett (approved):

jp-bennett:
@thebentern I've tested this on my Rak and tbeam. The factoryReset() was the main culprit, once everything else was actually copying the right number of bits to the right places.

And then the disableBluetooth is because I was seeing the Rak reboot before the file save was finished. Turns out the mobile device disconnects and then reconnects during the file operation, leading to a crash and reboot. @garthvh has indicated that this is also a problem when wiping the nodeDB, so added the function there, as well.

Regardless, private key saving is *finally* working as expected on both my test devices.

Issue #4675, opened 2024-09-10
[Bug]: position timestamps are ignored even when the node's time is unset


nhecker:
### Category

Other

### Hardware

Other

### Firmware Version

2.5.0.9e55e6b

### Description

Hi there,

Using 2.5.0.9e55e6b, a freshly booted node with no BT connection, no GPS, and `meshtastic_Position_LocSource_LOC_UNSET` will receive a position packet with a nonzero timestamp and will ignore the timestamp (`Ignoring time from mesh because it has a unknown or manual source`). This is not desirable because even if the timestamp isn't known to be dependable, it would be better than the nothing that we have.

Using 2.4.2.5b45303, a freshly booted node with no BT connection, no GPS, and `meshtastic_Position_LocSource_LOC_UNSET` will receive a position packet with a nonzero timestamp and will *not* ignore the timestamp. Instead the RTC is set with the time, based on that timestamp. This behavior makes sense to me, as we're using the best time source we have available to us even if it's not yet known to be trustworthy.

I'm looking at what changed and https://github.com/meshtastic/firmware/pull/4593/files caught my eye. I think I see what that PR was trying to accomplish, but it seems to be overly broad in that it misses an opportunity to set the time at least once, at or near boot time. I'm considering expanding the `if` on 132 so that if we're just booted up and don't have *any* time set, it'll use the value in the timestamp. Feel free to comment or close if this doesn't make sense, but I'd be happy to contribute a PR if it'd help.

Thanks!

nhecker:
https://github.com/meshtastic/firmware/pull/4680

PR #4680, opened 2024-09-11
fix: accept mesh time if nothing better

nhecker:
This (hopefully) fixes bug number [4675](https://github.com/meshtastic/firmware/issues/4675). I'm still working on validating that, but as a first-time contributor to this project that's gonna take a hot minute. Commits and comments welcome.

Happy to respond to feedback, let me know if anything looks amiss.


nhecker:
I agree that it does allow the mesh to act as a time source, and I reviewed @thebentern's PR before submitting the associated bug. I'm not trying to step on that work at all; I liked it. I simply wanted to allow mesh time from untrusted sources under one condition: when we don't have _any_ clue about what the time may be.

At least, that's the intent of this PR. I'm still getting familiar with the codebase so if there's a better way to do that I'm all ears.

PR #4690, opened 2024-09-12
Define SX126X_ANT_SW for the RAK11200 on the RAK19007 base

RealityAnomaly:
Fixes #3338, allows the RAK13300 and GNSS module to function correctly when used on the RAK19007 base with the RAK11200 core module by activating the 3.3V power rail correctly.

PR #4691, opened 2024-09-12
Rak11200 pin name correction

S5NC:
Functionality consistent, some related comments on https://github.com/meshtastic/firmware/pull/4690.

If `SX126X_POWER_EN` can be initialised in `SX126xInterface.cpp`, so can `SX126X_ANT_SW`.

Please trunk fmt


fifieldt (approved):
From reading the comments here and in the other bug, this seems to make sense.

It also cleans up main.cpp a little, which I like :)

PR #4692, opened 2024-09-12
First stab at ESP32-C6 support for TLora-C6

caveman99:
Many things don't work

- [x] Bosch Sensortec and other environment sensors
- [ ] NimBLE
- [ ] Webserver
- [ ] paxcounter


lyusupov:
> Many things don't work
> 
> * Bosch Sensortec and other environment sensors
> * Webserver

All that stuff has no issues with SoftRF [**Eco Edition**](https://github.com/lyusupov/SoftRF/wiki/Eco-Edition) 

> * NimBLE

Author "NimBLE for Arduino" says NO for ESP32-C6 https://github.com/h2zero/NimBLE-Arduino/issues/685

[**Arduino BLE**](https://github.com/arduino-libraries/ArduinoBLE) is known to work nicely with **ESP32-C6**



**Bluedroid BLE** works as well but takes approx. 450-500K more of flash memory space than **Arduino BLE** does.

caveman99:
Thanks for the pointers, i did wanna push out what i got till now so others can have a look at it. waiting for my actual device in the mail. So we'll need to port the newer ESP targets to Arduino BLE i guess.

lyusupov:
> So we'll need to port the newer ESP targets to Arduino BLE i guess.

Ok.
Be aware that this minor tweak for Arduino BLE for ESP32-C6 is required at this moment: https://github.com/arduino-libraries/ArduinoBLE/pull/364

PR #4694, opened 2024-09-13
Only return a time if we have a fix

fifieldt:
NMEA GGA Messages have a GPS Quality Indicator which indicates whether a GPS fix was achieved. Use this to determine whether we should attempt to set a time or not.

Also, fix up some of the custom airoha code and comments.


fifieldt:
the check failures don't look like my fault, I swear!

jp-bennett:
We broke web again! >< Yeah, not @fifieldt 's fault.

caveman99 (approved):

fifieldt:
Is this patch still necessary after 
#4705 
And 
#4704

?

This could delay setting a good time and the issues we've seen in the past are probably well solved by the other patches...

jp-bennett:
> Is this patch still necessary after 
> #4705 
> And 
> #4704
> 
> ?
> 
> This could delay setting a good time and the issues we've seen in the past are probably well solved by the other patches...

I don't think either of those would address a timestamp in the future, where this one seems to.

thebentern:
@fifieldt on voice chat yesterday we also discussed simply relaxing the time aggressive guarding window of 12 hours when a GPS is detected. That would solve this issue in a somewhat different way. We would have a temporarily inaccurate time until a good fix is acquired and then the GPS would correct the time.

fifieldt:
Closing this now the alternate fix is in :) Or at least if I understand correctly!

PR #4707, opened 2024-09-14
Allow only one traceroute towards and one back per minute

thebentern:
Untested so far

### DIFF
--- a/src/modules/TraceRouteModule.cpp
+++ b/src/modules/TraceRouteModule.cpp
@@ -12,6 +12,17 @@ bool TraceRouteModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, m
 void TraceRouteModule::alterReceivedProtobuf(meshtastic_MeshPacket &p, meshtastic_RouteDiscovery *r)
 {
     auto &incoming = p.decoded;
+    uint32_t now = millis();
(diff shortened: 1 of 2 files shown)


caveman99:
is this only for passing through requests or also for phoneAPI originated ones? i'd return an error for display on the client if it was a local API request.

src/modules/TraceRouteModule.cpp
```diff
@@ -12,6 +12,17 @@ bool TraceRouteModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, m
 void TraceRouteModule::alterReceivedProtobuf(meshtastic_MeshPacket &p, meshtastic_RouteDiscovery *r)
 {
     auto &incoming = p.decoded;
+    uint32_t now = millis();
+    if (incoming.request_id && lastTraceRouteBackMs && (now - lastTraceRouteBackMs) < (60 * 1000)) {
```
GUVWAF, line 16: `lastTraceRouteTowardMs` and `lastTraceRouteBackMs` don't seem to be set, but if you want to limit the reply generation, this should go only in `allocReply()`, because otherwise intermediate hops will not add themselves if another traceroute went through them already in the last minute.

PR #4713, opened 2024-09-15
feat: enable remote admin to set/clear fixed positions

andrekir:
This update enables changing and removing `fixed_position` for remote nodes via `set_fixed_position` & `remove_fixed_position` AdminMessages.

When fixed positions were introduced, they were set using standard Position packets, which lacked any way to establish trust in the data. To avoid spoofing any node's position (not just our own) it made sense to block remote fixed positions. 

However, with the introduction of secure admin messages, this restriction is no longer necessary.

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -257,34 +257,26 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
         break;
     }
     case meshtastic_AdminMessage_set_fixed_position_tag: {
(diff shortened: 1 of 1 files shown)


jp-bennett:
This makes sense to me. We no longer have insecure admin, and with the filtering of precise location on MQTT, there's also less chance of abuse there.

thebentern (approved):

noon92:
This will allow manual remote entry of gps coords via admin?

thebentern:
Yes

PR #4715, opened 2024-09-15
Don't use PKC on a non-primary channel unless specifically requested

jp-bennett:
Since we still have some uses where we want to send to a specific node over a specific channel, this attempts to catch those cases and honor the request. Specifically the admin and serial channel.

### DIFF
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -185,9 +185,12 @@ ErrorCode Router::sendLocal(meshtastic_MeshPacket *p, RxSource src)
             handleReceived(p, src);
         }
 
-        if (!p->channel) { // don't override if a channel was requested
-            p->channel = nodeDB->getMeshNodeChannel(p->to);
-            LOG_DEBUG("localSend to channel %d\n", p->channel);
+        if (!p->channel && !p->pki_encrypted) { // don't override if a channel was requested
+            meshtastic_NodeInfoLite *node = nodeDB->getMeshNode(p->to);
+            if (node && node->user.public_key.size == 0) {
+                p->channel = node->channel;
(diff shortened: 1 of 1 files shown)


thebentern (approved):

src/mesh/Router.cpp
```diff
@@ -478,10 +481,20 @@ meshtastic_Routing_Error perhapsEncode(meshtastic_MeshPacket *p)
...
         meshtastic_NodeInfoLite *node = nodeDB->getMeshNode(p->to);
-        if (!owner.is_licensed && config.security.private_key.size == 32 && p->to != NODENUM_BROADCAST && node != nullptr &&
-            node->user.public_key.size > 0 && p->decoded.portnum != meshtastic_PortNum_TRACEROUTE_APP &&
-            p->decoded.portnum != meshtastic_PortNum_NODEINFO_APP && p->decoded.portnum != meshtastic_PortNum_ROUTING_APP &&
-            p->decoded.portnum != meshtastic_PortNum_POSITION_APP) {
+        // We may want to retool things so we can send a PKC packet when the client specifies a key and nodenum, even if the node
+        // is not in the local nodedb
+        if (
+            // Don't use PKC with Ham mode
+            !owner.is_licensed &&
```
thebentern, line 488: Wondering if we can encapsulate some of the shared logic for detecting PKC intent for a packet into a method that returns a bool, since the criteria gets fairly lengthy and could grow.

GUVWAF (approved):

src/mesh/Router.cpp
jp-bennett: There's probably a few places through the code where massive if statements could use this treatment.

PR #4716, opened 2024-09-15
Feat: Add configurable neighbor removal interval multiplier

Rogelio96:
- Refactored the NeighborInfoModule CleanUpNeighbors function to use a configurable multiplier for determining the neighbor removal interval.
- Introduced a new variable in place of the hardcoded '2' multiplier, which can now be configured in userPrefs.
- This change allows users to customize how long nodes stay on their lists before being removed, offering more flexibility.
- Default multiplier remains set to 2.


thebentern:
I'm not sure this is a control we want to expose to the Neighbor Info module. We're trying to keep it as simple and default as possible, especially because it can potentially be used for routing in the future.

Rogelio96:
Thank you both for your feedback. I appreciate the direction you're aiming for and understand the need to keep things simple. I'll keep these points in mind for future contributions.

Rogelio96:
@thebentern @GUVWAF Thank you again for your feedback. My goal is to let clients customize how quickly nodes drop from their list. I seek your advice on how I could move forward with this goal in a way that aligns with the project as a whole. I appreciate your time. -Rogelio

Issue #284, opened 2020-07-15
loss of GPS lock not updated in the GUI


geeksville:
but the problem isn't the GUI.  It is that we only publish GPS updates when we have lock.  Once we lose lock we stop publishing.  Really - we should do one last publish when the state changes from locked to not locked (so the rest of the app establishes the new proper !locked state).

This fix should go into 1.0 (master).

From this forum report:

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/bug-tbeam-never-returns-to-no-sats-state/741/2

Professr:
Mind if I fix this one?

geeksville:
Feel free to fix any one you want! ;-)

Professr:
Hmm.  NEMA already looks ok, but UBlox always reports its last known fix.  Fortunately it seems to update the number of satellite locks it has, so I integrated that into the logic that determines whether the module has a lock or not.  It takes 5-10 minutes to really pick up on the loss of lock, but it does go back to "No sats" at that point.

PR #4721, opened 2024-09-15
Separate GPS and NTP RTCQuality logic and allow GPS time to always set us

thebentern:
Alternative to #4694 

This would allow for the T-Echo in question to be temporarily _wrong_ but eventually correct the time with a time from GPS with a fix.


fifieldt:
I haven't had time to look into this, but in an ancient issue: https://github.com/meshtastic/firmware/issues/284#issuecomment-658559972 it's noted that ublox puts its last known fix in NMEA messages, which could include an old time.  Will try and take a look early next week, but by all means push ahead :)

jp-bennett (approved):

PR #4727, opened 2024-09-16
Make Ublox code more readable

fifieldt:
Ublox GNSS chips comes in a myriad of versions and settings. Presently our configuration code treats all of Ublox as a monolith and in one long function does a lot of branching based on versions.

This patch:
1. Separates version detection from setup code
2. Separates setup into methods for each of: Ublox 6, Ublox 7/8/9, and Ublox10.
3. Adds a macro for sending a Ublox message.

The objective of these changes was to make the code much shorter and more readable.


fifieldt:
Uploading for early sanity check -- I've still in the middle of testing.

GPSFan:
The u-blox code and the GPS init code in general can use some sprucing up. A while ago I did a slash/hack on the u-blox identifying code, it worked for v6/v7/M8/M10, and actually reduced the overall code size. Time flies and sheep happens, never got around to doing a PR, and it bit rotted with the many changes and new parts that are now supported.

I am not a programmer and welcome insights from folks that actually know what they are doing. I'm a hardware guy I've been doing u-blox stuff since the Neo-7 was a brand new part, and fiddling with GPS since the mid 90s.

I would like to explore the possibility of changing the probing so that there could be a:
#define GNSS_Model xxxx in variant.h
Where xxxx could be:
not defined and default to PROBE in GPS.cpp
or:
PROBE
UBLOX
ATGM336
TAU1202
L76K
L76B
L89
etc.......

That would keep the existing probing ability but allow individual platforms to pre-define a GNSS Model to reduce the code footprint as well as probe time.


What do you think?

fifieldt:
Thanks, yeah, one of my hopes for this patch was that it would also make it easier for GPS experts who are not necessarily coders to wrangle the public messages.

On the probe code, certainly room to do more :) For example there are probably some variants that have a ublox soldered on them that could save a lot of time if baudate 384000 was set in variant.h .Saving that for a different patch though!

fifieldt:
I've tested this on Ublox 6 and it works the same as before the patch, which was the aim!

thebentern:
Looks good. Let's merge this one after we cut a 2.5.1 release

thebentern (approved):

PR #4735, opened 2024-09-16
Remove lora_isp4520

fifieldt:
Per Ben: "We haven't supported any NRF5832 based board in quite some time. It's relatively resource constrained compared to the NRF52840"


madejackson:
Does it makes sense to remove this while the wio-e5 is being in there and actively tested which has 50% less / only 256kb of flash?

I'm planning on doing some tests for the isp4520 in the near future (as sensor role).

fifieldt:
The code was quite a way off working, whereas I think the wio kinda does

PR #4742, opened 2024-09-17
More useful PKC logging

jp-bennett:
Eliminates a log line or two that hasn't been useful, hides most of the derived shared key, and explicitly lists the public keys that were used for the DH step.


fifieldt (approved):

jp-bennett:
@thatSFguy
This has better logging for DM issues. If you want to grab a build artifact from the CI run, you can test this now. It will almost certainly be in 2.5.1, too.

Issue #3690, opened 2024-04-21
PMSA Sensor no detected[Bug]: 


MALAONE1:
### Category

Hardware Compatibility

### Hardware

RAK 4631, TTGOs

### Firmware Version

2.3.6

### Description

I have two PMSA 0031 sensors that are supposed to be Meshtastic compatible. I have one from RAK Wisbock that I can plug in directly and I have one that I can solder to the I2C interface. The sensor is not recognized or auto detected by the RAK 4631, Lora T3S3 and TTGO Lora32 1.6. The air quality module has been enabled in the telemetry module. This was not recognized in the previous firmware either

shodan8192:
Hello, I have RAK4631 with RAK12039 (PMSA003I). This sensor become detectable on I2C bus after ~3 seconds from powering up, when scan is already finished.

Inserting `delay(4000)` after `digitalWrite(AQ_SET_PIN, HIGH)` in this code should help : https://github.com/meshtastic/firmware/blob/75dc8cccecd52da78f1d69eeb4eb017fbc68f26c/src/main.cpp#L371-L375

PR #4743, opened 2024-09-17
PMSA0031 sensors require ~3secs before coming up on I2C

fifieldt:
As reported by @MALAONE1 and debugged by @shodan8192 , PMSA0031s on a RAK4631 take 3 seconds before they can become detectable on I2c.

Add a delay(4000) before I2C scan if the air quality sensor pin is defined.

Fixes https://github.com/meshtastic/firmware/issues/3690

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -393,6 +393,7 @@ void setup()
     // RAK-12039 set pin for Air quality sensor
     pinMode(AQ_SET_PIN, OUTPUT);
     digitalWrite(AQ_SET_PIN, HIGH);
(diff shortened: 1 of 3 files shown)


caveman99:
This can be solved more elegantly by doing a rescan when the actual module initializes. We do the same with cardkb for the same reason. Gonna have a look at it later on.

fifieldt:
Elegant sounds better!

fifieldt (commented):
elegant :)

thebentern (approved):

Issue #4605, opened 2024-09-01
[Feature Request]: Expose an I2C interface for Heltec Mesh Node T114


todd-herbert:
### Platform

NRF52

### Description

(Opened on behalf of LupusWorax on the Meshtastic Discord server. I don't myself have this device)

The one I2C interface currently used by Heltec Mesh Node T114 is connect directly to the PCF8563TS RTC (?)
The pins (`P0.26` and `P0.27`) are not exposed via the headers or the jumper connectors.
If I understand it correctly, the NRF52840 datasheet suggests it should be possible to define a second I2C bus.

Could we create a second I2C bus, mapped to the device's header pins? Are there any technical reasons why this wouldn't be possible or would be unwise?

lyusupov:
> The one I2C interface currently used by Heltec Mesh Node T114 is connect directly to the PCF8563TS RTC (?)

<img width="450" alt="image" src="https://github.com/user-attachments/assets/18955cac-9baa-4c4d-9356-f00d8291122d">


<img width="749" alt="image" src="https://github.com/user-attachments/assets/5019a0c4-5387-434a-b5e7-ee688f8bd138">

todd-herbert:
Ah I had a funny feeling I'd heard that the RTC was actually populated on the boards that were being shipped. If it's not, it might be an option to just remap the pins for the first I2C bus? Although there could be an argument for leaving that first I2C bus where Heltec has defined it too, in case that RTC does get included in future revisions?

lyusupov:
Heltec have borrowed mapping for most of nRF52840 GPIO pins from LilyGO T-Echo to ease porting of the firmware.
RTC and FLASH are using the same GPIOs as T-Echo as well.
Heltec just decided to save some money on the components that they think are not in use.

thebentern:
> I don't myself have this device

Can you request a couple in the budget room? Probably any of the other Heltec Vision Master hardware missing in your inventory as well.

lyusupov:
> Ah I had a funny feeling I'd heard that the RTC was actually populated on the boards that were being shipped.

The board delivered to me yesterday has neither RTC nor SPI FLASH.

<img  alt="image" src="https://github.com/user-attachments/assets/495bb690-c01b-464a-a27f-2d33edc334d3">

HarukiToreda:
> ed to me yesterday has neither RTC nor SPI FLASH.

According to Heltec, these are installed upon request when you place the order.

todd-herbert:
[...]

PR #4745, opened 2024-09-17
I2C for Heltec Mesh Node T114

todd-herbert:
Resolves https://github.com/meshtastic/firmware/issues/4605

Adds an I2C bus, assigned to P0.13 (SDA) and P0.16 (SCL), which are available on the header pins.

This is implemented using a second I2C bus.
Bus 0 retains its original assignment from Heltec: routed to the footprint for an RTC which is not populated on the current hardware version.

Another option would be to simply reassign the pins for the currently unused I2C bus 0. 
Is there reason to consider this instead?

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -210,6 +210,16 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define MINIMUM_SAFE_FREE_HEAP 1500
 #endif
 
+#ifndef WIRE_INTERFACES_COUNT
+// Officially an NRF52 macro
+// Repurposed cross-platform to identify devices using Wire1
+#if defined(I2C_SDA1) || defined(PIN_WIRE_SDA)
+#define WIRE_INTERFACES_COUNT 2
+#elif HAS_WIRE
+#define WIRE_INTERFACES_COUNT 1
+#endif
+#endif
+
 /* Step #3: mop up with disabled values for HAS_ options not handled by the above two */
 
 #ifndef HAS_WIFI
--- a/src/detect/ScanI2CTwoWire.cpp
+++ b/src/detect/ScanI2CTwoWire.cpp
@@ -162,13 +162,13 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
     Melopero_RV3028 rtc;
 #endif
 
-#ifdef I2C_SDA1
+#if defined(I2C_SDA1) || (defined(NRF52840_XXAA) && (WIRE_INTERFACES_COUNT == 2))
     if (port == I2CPort::WIRE1) {
         i2cBus = &Wire1;
     } else {
 #endif
         i2cBus = &Wire;
-#ifdef I2C_SDA1
+#if defined(I2C_SDA1) || (defined(NRF52840_XXAA) && (WIRE_INTERFACES_COUNT == 2))
     }
 #endif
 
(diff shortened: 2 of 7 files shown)


src/detect/ScanI2CTwoWire.cpp
```diff
@@ -162,13 +162,13 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
     Melopero_RV3028 rtc;
 #endif
 
-#ifdef I2C_SDA1
+#if defined(I2C_SDA1) || (defined(NRF52840_XXAA) && (WIRE_INTERFACES_COUNT == 2))
     if (port == I2CPort::WIRE1) {
```
caveman99, line 166: feels like the (defined(NRF52840_XXAA) && (WIRE_INTERFACES_COUNT == 2)) can be combined into a single macro, or even just define I2C_SDA1 - as it is now there's a lot of duplicate code.

caveman99 (changes requested):

src/detect/ScanI2CTwoWire.cpp
todd-herbert: I did try out the idea of defining `I2C_SDA1`, but it messed with the preprocessor logic for the `Wire.begin`.
I do agree though, it's messy. I'll go back and see what options I can come up with to tidy it.

todd-herbert: I'm thinking either:
 ```cpp
#if defined(PIN_SDA1) || defined(PIN_WIRE1_SDA)
```
or
```cpp
#if WIRE_INTERFACES_COUNT == 2
```
and setting `WIRE_INTERFACES_COUNT` once in configuration.h

What are your thoughts on these as options?

lyusupov:
image


Think about to use this order:
 
Signal|Pin
---|---
SCL|13
SDA|16

image

todd-herbert:
> **Think about to use this mapping:**
> Signal 	Pin
> SCL 	13
> SDA 	16

That's definitely a reasonable suggestion. 

My first hope was to match the qwIIC mapping Heltec used on their other Vision Master boards, but this wasn't possible because VCC and GND are reversed, so I didn't pay too much attention.
image


If the pin-out in your photos is another common layout for I2C devices, then definitely it's worth swapping the SCL and SDA assignment, so at least it can be convenient in some cases.

caveman99 (approved):

src/configuration.h
```diff
@@ -210,6 +210,16 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define MINIMUM_SAFE_FREE_HEAP 1500
 #endif
 
+#ifndef WIRE_INTERFACES_COUNT
+// Officially an NRF52 macro
+// Repurposed cross-platform to identify devices using Wire1
+#if defined(I2C_SDA1) || defined(PIN_WIRE_SDA)
```
fifieldt, line 216: @todd-herbert  shouldn't this be PIN_WIRE_SDA1 here?

fifieldt: PIN_WIRE1_SDA* rather :)

fifieldt: https://github.com/meshtastic/firmware/pull/4840

todd-herbert:
@fifieldt Ah thank you for spotting that!

pagex:
Wondering how were you flashing T114 as there is no available at web flasher?

PR #4747, opened 2024-09-17
Implement optional hops_away on NodeInfo/Lite

thebentern:
@garthvh question from the app consumer perspective: Do you want the current node to have a null hops_away or hops_away of zero? I elected to use the former.


caveman99 (approved):
LGTM

ianmcorvidae (approved):
To the best of my knowledge (& ability to look on my phone) this looks good!

ianmcorvidae:
Just saw the client-dev perspective question (albeit not directed at me). I think null is better as well. The local node is basically -1 hops "away" but null is better I think, haha

PR #4748, opened 2024-09-17
Implement CoT detail support in TAKPacket

thebentern:
@niccellular once this goes live, think we should do a major version bump on the ATAK Plugin, since it's a breaking change for the non-standard CoTs?


src/modules/AtakPluginModule.cpp
```diff
@@ -52,6 +52,10 @@ meshtastic_TAKPacket AtakPluginModule::cloneTAKPacketData(meshtastic_TAKPacket *
     } else if (t->which_payload_variant == meshtastic_TAKPacket_chat_tag) {
         clone.which_payload_variant = meshtastic_TAKPacket_chat_tag;
         clone.payload_variant.chat = {0};
+    } else if (t->which_payload_variant == meshtastic_TAKPacket_detail_tag) {
+        clone.which_payload_variant = meshtastic_TAKPacket_detail_tag;
+        clone.payload_variant.detail.size = t->payload_variant.detail.size;
+        memcpy(clone.payload_variant.detail.bytes, clone.payload_variant.detail.bytes, t->payload_variant.detail.size);
```
fifieldt, line 58: memcpy is source, dest, size, yeah? Is the source and dest supposed to be the same here?

thebentern: Doh! Good eyes.

src/modules/AtakPluginModule.cpp
```diff
@@ -55,7 +55,7 @@ meshtastic_TAKPacket AtakPluginModule::cloneTAKPacketData(meshtastic_TAKPacket *
     } else if (t->which_payload_variant == meshtastic_TAKPacket_detail_tag) {
         clone.which_payload_variant = meshtastic_TAKPacket_detail_tag;
         clone.payload_variant.detail.size = t->payload_variant.detail.size;
-        memcpy(clone.payload_variant.detail.bytes, clone.payload_variant.detail.bytes, t->payload_variant.detail.size);
+        memcpy(clone.payload_variant.detail.bytes, t->payload_variant.detail.size, t->payload_variant.detail.size);
```
fifieldt, line 58: maybe t->payload_variant.detail.bytes?

thebentern: Coffee isn't kicking in well yet this morning. I'm 0 for 2 😆

PR #4769, opened 2024-09-18
t-echo: fix phantom pps led on shutdown

augustozanellato:
T-Echo have GPS PPS connected to both P1.04 and to an active high led (see [schematics](https://github.com/Xinyuan-LilyGO/T-Echo/blob/main/T-Echo_Schematic.pdf)), on shutdown the led is faintly on (see attached image) causing unnecessary battery drain.

This PR addresses the issue by defining the PPS pin in target config and by pulling it low on shutdown.

image

### DIFF
--- a/src/platform/nrf52/main-nrf52.cpp
+++ b/src/platform/nrf52/main-nrf52.cpp
@@ -253,9 +253,16 @@ void cpuDeepSleep(uint32_t msecToWake)
 #endif
 #endif
 
-#ifdef HELTEC_MESH_NODE_T114
+#ifdef PIN_GPS_PPS
     nrf_gpio_cfg_default(PIN_GPS_PPS);
     detachInterrupt(PIN_GPS_PPS);
+#ifdef TTGO_T_ECHO
+    pinMode(PIN_GPS_PPS, OUTPUT);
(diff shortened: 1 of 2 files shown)


lyusupov:
> This PR addresses the issue by defining the PPS pin in target config and by pulling it low on shutdown.
```
#ifdef TTGO_T_ECHO
    pinMode(PIN_GPS_PPS, OUTPUT);
    digitalWrite(PIN_GPS_PPS, LOW);
#endif
```

> [!CAUTION]
> This **will likely increase deep sleep current** - the L76K PPS drain will follow connected MCU's GPIO pin path _(the one been pulled LOW)_.

augustozanellato:
> This will likely increase deep sleep current - the L76K PPS drain will follow connected MCU's GPIO pin path (the one been pulled LOW).

There's a 1k resistor between MCU and PPS, not sure how much that affects calculations. Sadly I don't have a PPK I can use to actually measure the impact of this PR. I noticed that setting pinMode to INPUT_PULLDOWN has the same effect, but again I don't know about its effect on power draw.

On the other hand I think part of this PR should be still merged excluding the `#ifdef TTGO_T_ECHO` part

PR #4774, opened 2024-09-19
Actions: Semgrep Images have moved from returntocorp to semgrep

fifieldt:
https://hub.docker.com/r/returntocorp/semgrep notes: "We've moved!
 Official Docker images for Semgrep now available at semgrep/semgrep."

Patch updates our CI workflow for these images.


caveman99 (changes requested):
CI says nope :-)

PR #4776, opened 2024-09-19
Revert "DetectionSensor: broadcast all state changes"

thebentern:
Reverts meshtastic/firmware#4767

@augustozanellato reverting this for now. I think we should add a new configuration to the protobufs to manage whether or not to notify on the state change _back_ to the false detection state. Many use-cases do not want this extra chatter, for example the initial one we tested on: doppler and PIR motion sensor events, as well as recently I've seen use of button as a "man down" detector.

### DIFF
--- a/src/modules/DetectionSensorModule.cpp
+++ b/src/modules/DetectionSensorModule.cpp
@@ -49,16 +49,10 @@ int32_t DetectionSensorModule::runOnce()
 
     // LOG_DEBUG("Detection Sensor Module: Current pin state: %i\n", digitalRead(moduleConfig.detection_sensor.monitor_pin));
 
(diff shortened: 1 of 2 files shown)


augustozanellato:
Yeah, that was one of my questions when implementing that PR.
My usecase is tracking power outages so at least for some use cases the extra mesh traffic makes sense.
Do you have a suggestion for the config flag name? I can't come up with anything.

thebentern:
> Do you have a suggestion for the config flag name? I can't come up with anything.
Perhaps a bool `broadcast_all_state_transitions`?

augustozanellato:
That’s make sense, thanks. PR coming in a couple hours.

Issue #4655, opened 2024-09-08
[Bug]: Flip Screen not working on Heltec T114


TheBamse:
### Category

Other

### Hardware

Other

### Firmware Version

2.4.3.91d6612 and 2.5.0.9e55e6b

### Description

When setting the Flip Screen option via either python CLI (version 2.4.0) or Android app (version 2.4.4) to true it has no effect on device.

Tried factory erasing and a reflash with both firmware versions on both devices with identical results (an unflipped screen :))

### Relevant log output

```Shell
Example CLI --info excerpt:
 "display": {
    "screenOnSecs": 60,
    "flipScreen": true,
    "headingBold": true,
    "gpsFormat": "DEC",
    "autoScreenCarouselSecs": 0,
    "compassNorthTop": false,
    "units": "METRIC",
    "oled": "OLED_AUTO",
    "displaymode": "DEFAULT",
    "wakeOnTapOrMotion": false,
    "compassOrientation": "DEGREES_0"
  }
```

todd-herbert:
I think that setting's original intention might have been to fix display output coming out mirrored on some OLED displays, but you're keen to have it rotate the display image 180 degrees right? That'd probably something that's pretty achievable in the near future, although would have to check that no other TFT displays use it to fix mirroring.

TheBamse:
The docs is a bit ambigiuos on this;
"Flip Screen
If enabled, the screen will be rotated 180 degrees, for cases that mount the screen upside down"

It could be interpreted as mirroring but since I was after rotating it to flip it upside-down I just assumed the documentation was referring to that ;)

todd-herbert:
> The docs could perhaps be a bit ambiguous on this(?);
> If enabled, the screen will be rotated 180 degrees, for cases that mount the screen upside down"

Oh, you could totally be right on that! In any case I'll definitely look into in the near future.

PR #4786, opened 2024-09-20
Honor flip & color for Heltec T114 and T190

todd-herbert:
Resolves https://github.com/meshtastic/firmware/issues/4655
Tested working on T114, but needs testing on T190

To-do: update `lib_deps` to point at https://github.com/meshtastic/st7789/pull/1 once merged
___

* Honors `display.config.flip_screen` when set by user
* Honors `TFT_MESH` color if defined in variant.h

### DIFF
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -1670,11 +1670,9 @@ void Screen::setup()
     static_cast<SH1106Wire *>(dispdev)->setSubtype(7);
 #endif
 
-#ifdef USE_ST7789
-// Heltec T114 and T190: honor a custom text color, if defined in variant.h
-#ifdef TFT_MESH
+#if defined(USE_ST7789) && defined(TFT_MESH)
+    // Heltec T114 and T190: honor a custom text color, if defined in variant.h
     static_cast<ST7789Spi *>(dispdev)->setRGB(TFT_MESH);
-#endif
 #endif
 
     // Initialising the UI will init the display too.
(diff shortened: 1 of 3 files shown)


src/graphics/Screen.cpp
```diff
@@ -1670,6 +1670,13 @@ void Screen::setup()
     static_cast<SH1106Wire *>(dispdev)->setSubtype(7);
 #endif
 
+#ifdef USE_ST7789
+// Heltec T114 and T190: honor a custom text color, if defined in variant.h
+#ifdef TFT_MESH
+    static_cast<ST7789Spi *>(dispdev)->setRGB(TFT_MESH);
+#endif
+#endif
+
```
caveman99, line 1679: #if defined(USE_ST7789) && defined(TFT_MESH)
[...]
#endif

caveman99 (changes requested):
See inline comments. 

EDIT: did you upstream the lib changes?

thebentern:
Lib changes were just merged this morning

todd-herbert:
I think @HarukiToreda is going to give this a test-run on his T190s later today to check for any unexpected weirdness too 🙏

HarukiToreda:
i tested the flip on the T190, works great

AaronJackson:
I recall trying a similar change to enable flipping on the Wireless Tracker, which caused it to crash at boot. I think due to the negative display offset for those boards.

https://github.com/meshtastic/firmware/blob/18aac0ba25e0a638fd7cbab4b73c95105892b88e/variants/heltec_wireless_tracker_V1_0/variant.h#L25-L26

In the end I just hard coded and did a build for that specific board to force the flip (I designed my case upside down...)

```diff
-#define TFT_OFFSET_X 26
-#define TFT_OFFSET_Y -1
+#define TFT_OFFSET_X -26
+#define TFT_OFFSET_Y 1
```

and changed the `tft->setRotation(x)` from 3 to 2.

I'm not in a position to test this at the moment unfortunately

PR #4817, opened 2024-09-22
Mesh Modules specify when handling keyboard input

### LINKED ISSUE #4681: [Bug]: Waypoint breaks Cardkb and Tdeck left and right buttons/trackball
### Category

Other

### Hardware

Heltec V3

### Firmware Version

2.4.3

### Description

when someone shares a waypoint on the mesh and it shows on the screen, the left and right buttons on the cardkb and the tackball no longer work to go to the next or previous frames. Someone recently shared a waypoint and all my nodes were affected, it got fixed when the waypoint screen was removed. 

### Relevant log output

_No response_

todd-herbert:
Resolves #4681

### Background
* Screen class navigates between frames with left / right keyboard input.
* Navigation is suppressed when a module frame is shown, to allow canned message text input.

### Problem
* When waypoint module frame is shown, left / right keyboard navigation is suppressed indefinitely: UI is "stuck"

### Solution
* Virtual method added to MeshModule class; allow modules to explicitly state when handling keyboard input.
* When not handling keyboard input, allow normal left / right UI keyboard navigation, even on module frames.
* _(Bonus)_ CardKB key-combo to dismiss Waypoint frame, for convenience.


src/input/kbI2cBase.cpp
```diff
@@ -295,6 +295,7 @@ int32_t KbI2cBase::runOnce()
             case 0x9b: // fn+s      INPUT_BROKER_MSG_SHUTDOWN
             case 0xac: // fn+m      INPUT_BROKER_MSG_MUTE_TOGGLE
             case 0x9e: // fn+g      INPUT_BROKER_MSG_GPS_TOGGLE
+            case 0x8e: // fn+w      INPUT_BROKER_MSG_WAYPOINTFRAME_DISMISS
```
thebentern, line 298: Should we make something more generic to also dismiss text message frames as well?
Fn+D for delete / dismiss perhaps

todd-herbert: That could be a good call. Would you be happy for me to revert that combo for now, and address it separately in the near future?

thebentern: What about just switching the combo to Fn+D for now and continuing the rest of the PR as is? Then we can follow up with extending the functionality to the text message frames later. :-)

todd-herbert: I think it might be a bit late in the evening now to trust my brain with a quick push for this change, but I'll get back on it tomorrow, and might as well figure out the logistics of the general-dismissal thing while I'm at it.

HarukiToreda: > mbo to Fn+D for now and continuing the rest of the PR as is? Then we can follow up with extending the functionality to the text message frames later. :-)

will it be deleting both or just the frame it's on? if you do one click of it, and it deletes my waypoint and my message could leave a possibility of deleting an important message

HarukiToreda: by choosing fn+w it focuses on just the way to delete, but possibility of stand alones to also set a way point in the future. Just a thought

thebentern: >will it be deleting both or just the frame it's on? if you do one click of it, and it deletes my waypoint and my message could leave a possibility of deleting an important message

I should be able to remove either a waypoint or a textmessage frame if I am currently on it in the frame index.

todd-herbert:
> * CardKB key-combo to dismiss Text Message and Waypoint frames, for convenience.

@thebentern @HarukiToreda Are you guys happy with this mapped to Fn + Delete? No objections to changing it if you'd prefer something different.

HarukiToreda:
> y with this mapped to Fn + Delete? No objections to changing it if you'd prefer something different.

yeah looks like the intuitive shortcut combination that will be easy to remember.

thebentern (approved):

PR #4819, opened 2024-09-22
Account for size of Envelope when allocating buffer.

caveman99:
```
INFO  | 09:29:20 568 [mqtt] Subscribing to msh/2/e/LongFast/+
INFO  | 09:29:20 568 [mqtt] Subscribing to msh/2/json/LongFast/+
INFO  | 09:29:20 568 [mqtt] Subscribing to msh/2/e/PKI/+
DEBUG | 09:29:20 568 [mqtt] Publishing enqueued MQTT message
ERROR | 09:29:20 568 [mqtt] Panic: can't encode protobuf reason='bytes size exceeded'

assert failed: size_t pb_encode_to_bytes(uint8_t*, size_t, const pb_msgdesc_t*, const void*) mesh-pb-constants.cpp:18 (0)
```

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -32,6 +32,9 @@ static MemoryDynamic<meshtastic_ServiceEnvelope> staticMqttPool;
 
 Allocator<meshtastic_ServiceEnvelope> &mqttPool = staticMqttPool;
 
+// FIXME - this size calculation is super sloppy, but it will go away once we dynamically alloc meshpackets
+static uint8_t bytes[meshtastic_MqttClientProxyMessage_size + 30]; // 12 for channel name and 16 for nodeid
+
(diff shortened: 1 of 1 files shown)


caveman99:
Discussion: 

- Anybody got a clever idea how to access these maximum sizes programmatically?
- Can we define this buffer only once in the class protected section? would save quite some ram...

thebentern:
> Anybody got a clever idea how to access these maximum sizes programmatically?

Completely different approach, but I think we should talk to the app devs about potentially having them build the topics. They should have all of the information available to do so.
> Can we define this buffer only once in the class protected section? would save quite some ram..

Sound good to me.

caveman99:
> Completely different approach, but I think we should talk to the app devs about potentially having them build the topics. They should have all of the information available to do so.

The same struct is used later on to either push to a proxy client or publish to MQTT directly. The initial error was happening on a real broker connection from an ESP32. So we'd have to split paths earlier.

PR #4805, opened 2024-09-21
Update classes on protobufs update

scruplelesswizard:
This PR changes the update process for Protobufs. Every time a change is made to the `meshtastic` submodule in a PR, the `regen-protos.sh` will be run and the changes committed to the PR's branch.

This enables class regeneration on Dependabot's PRs and also on any other users PR where `meshtastic/` is updated. Additionally it ensures that on any PRs where the protobufs have changed that the classes match.

PR #4818, opened 2024-09-22
[Board]: Support for M5Stack CoreS3

lboue:
Add support for M5Stack CoreS3 variant:

- [M5Stack CoreS3](https://shop.m5stack.com/products/m5stack-cores3-esp32s3-lotdevelopment-kit)
- [LoRa868_V1.1 Module](https://shop.m5stack.com/products/lora-module-868mhz-v1-1) with RA01H (SX1276)

- [x] Radio LoRa part
- [ ] LCD display

image

PR #4823, opened 2024-09-22
Support manual Protobuf update process

scruplelesswizard:
This re-enables the update process originally used for protobufs, lost in #4805


scruplelesswizard:
Testing in #4818

scruplelesswizard:
`workflow_dispatch` trigger tested in 
https://github.com/scruplelesswizard/firmware/actions/runs/10983167024

PR #4830, opened 2024-09-23
Removed extra Chip select in SPI object

petrkr:
This is useless here. It may colidate with other SPI devices on BUS later.

RadioLib take care about ChipSelect itself. So use CS on SPI object is pointless and may cause problems later.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -679,8 +679,8 @@ void setup()
     SPI.begin();
 #else
     // ESP32
-    SPI.begin(LORA_SCK, LORA_MISO, LORA_MOSI, LORA_CS);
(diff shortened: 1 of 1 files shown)


caveman99:
SPI is not exclusively used by radiolib, sometimes we share the same bus with TF Card or Display driver. We take care of this by using a LockingHal underneath the Radiolib driver. ESP32 needs to know about the CS for these other devices. There's never been a clash observed.

petrkr:
> SPI is not exclusively used by radiolib, sometimes we share the same bus with TF Card or Display driver. We take care of this by using a LockingHal underneath the Radiolib driver. ESP32 needs to know about the CS for these other devices. There's never been a clash observed.

Yes, other device has own CS, that is what is about this PR. It is also prerequisite for SPI Ethernet support.

Tf card will use different CS than Display aswell, so it is pointless initilize SPI object with one CS for lora pin, it may cause collisions

PR #4697, opened 2024-09-13
Refactor AccelerometerThread.h 

dhskinner:
This PR is to refactor AccelerometerThread.h into its own folder and tease apart the code into individual classes per sensor to make it easier to add more sensors in future - for more info refer to #4662 

It picks up all 6 existing accelerometers, plus adds a basic wake-on-move implementation for a new sensor ICM-20948 (the intent is add digital motion processing for ICM-20948 in a new PR to come)

Issue #4662, opened 2024-09-09
Additional accelerometers (code hygiene)


dhskinner:
### Discussed in https://github.com/meshtastic/firmware/discussions/4614

<div type='discussions-op-text'>

<sup>Originally posted by **dhskinner** September  3, 2024</sup>
Hi - currently there's a couple of accelerometers and motion sensors supported by Meshtastic that are all in one file at [.../src/AccelerometerThread.h](https://github.com/meshtastic/firmware/blob/master/src/AccelerometerThread.h) which makes it a little awkward to add new ones

Can I suggest it could be the right time to move AccelerometerThread.h into its own folder and tease apart the code into individual classes per sensor, and to improve code hygiene - like:

```
/src
    |->motion
        |-> AccelerometerThread.h
        |->MPU6050.cpp
        |->MPU6050.h
        |->...etc...
```

The reason for raising this is I'll shortly add a PR to add the ICM-20948 (n.b. the newer big brother from the MPU-6050) but it will need a couple of new code and header files to fully enable the onboard 9-Axis digital motion processor which is quite a cool... it would be a bit odd to add a bunch of new files into the [.../src/](https://github.com/meshtastic/firmware/tree/master/src) directory

Am more than happy to do the work and submit a small PR first for some light refactoring, but.... before I start I wanted to check in - will this annoy our admins by adding new folders and moving a bit of code around or is it do-able?

[edit - as of today (Sep 3rd) I did a scan of all open PR's and there doesn't seem to be any I could spot where anyone else is making changes to AccelerometerThread.h in parallel]
</div>

@thebentern appreciate if you'd be able to take a look and let me know thoughts?

thebentern:
Looks like a good organizational strategy compared to what we have now. Excited to get more IMU and accelerometer support going

dhskinner:
> Have you had a look at how the temperature and other sensors are broken out at modules/telemetry/sensor? Is that the kind of approach you're thinking of following?

Yep am thinking pretty much the same approach as for  TelemetrySensor and  VoltageSensor 

> Looks like a good organizational strategy compared to what we have now. Excited to get more IMU and accelerometer support going

OK I'll do the refactoring and submit a PR in next few days. Currently I only have one of the existing accelerometers to test against (an MPU6050) so will see how it goes and might need to ask a favour for some help to test and confirm everything works as expected when its done. Much appreciated!

gjelsoe:
[...]

PR #4831, opened 2024-09-23
 Refactor AccelerometerThread.h

dhskinner:
This PR is to refactor AccelerometerThread.h into its own folder and tease apart the code into individual classes per sensor to make it easier to add more sensors in future - for more info refer to https://github.com/meshtastic/firmware/issues/4662

It picks up all 6 existing accelerometers, plus adds a basic wake-on-move implementation for a new sensor ICM-20948 (the intent is add digital motion processing for ICM-20948 in a new PR to come)

@thebentern, @fifieldt and @caveman99 - this is a do-over of [#4697](https://github.com/meshtastic/firmware/pull/4697) that you reviewed last week. Sorry had to cancel that one because of an issue [raising the PR from another organisation account](https://github.com/orgs/community/discussions/5634). Comments raised have been fixed up in this new PR

This new one should be good to review/test/merge or let me know if anything needs fixing? Thanks


src/motion/STK8XXXSensor.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
+#ifndef _STK8XXX_SENSOR_H_
+#define _STK8XXX_SENSOR_H_
+
```
fifieldt, line 4: Should there be an #endif here, rather than at the end of the file?

dhskinner: Hi - yep good question - what I'm concerned about is making sure that each sensor only gets compiled and linked once

The #ifndef is an [include guard](https://includeguardian.io/article/pragma-once-vs-ifndef) to prevent the CPP header file being imported and compiled twice i.e. that's the reason the #if <-> #endif wraps around most of the whole .h file 

In theory #pragma once also does this, but to be honest I can't actually tell if the build chain for platform.io respects the #pragma once directive on all platforms, so am playing it safe and throwing them in.

fifieldt: Thanks for the explanation :)

fifieldt (approved):

fifieldt (approved):

thebentern (approved):

PR #4834, opened 2024-09-23
Teardown bluetooth phoneAPI better and fix client notification issue

### LINKED ISSUE #4388: [Bug]: Bluetooth connectivity drops repeatidly when MQTT JSON is activated
### Category

Other

### Hardware

T-Beam, T-Lora v2 1.6

### Firmware Version

2.4.0.46d7b82 Beta

### Description

I have 2 T-Beams and a T-Lora board set to use MQTT and either Wi-Fi or Cellphone as proxy. So far so good, when out of RF range, MQTT is relaying messages.

However, I tried to set the "JSON Enabled" to true so I can integrate in an easy way some data to use with my home automation.

Once turned on and saved, the selected device reboot and reconnect to my mobile device for few seconds and the BLE connectivity drop for few seconds and comes back on. This stats loops until I am quick enough to disable the JSON output OR that I use another device through the admin channel to deactivate it.

The hardware device will not reboot, it continues to work as supposed except for the BLE connectivity.

### Relevant log output

```Shell
Not yet
```

thebentern:
Closes #4388 
Adding a ton of explicit tearing down seems to improve stability a lot between BLE disconnects. 
Also fixed an inconsistency in how we were handling ClientNotification vs other FromRadio variants with their own queues
Tested on Heltec V3 and RAK-4631.

PS: Also hoping this has a side-effect of fixing some of the nodedb / config wipes on BLE disconnect and reconnect cycles sporadically.

### DIFF
--- a/src/mesh/MeshService.h
+++ b/src/mesh/MeshService.h
@@ -92,6 +92,9 @@ class MeshService
     /// Return the next MqttClientProxyMessage packet destined to the phone.
     meshtastic_MqttClientProxyMessage *getMqttClientProxyMessageForPhone() { return toPhoneMqttProxyQueue.dequeuePtr(0); }
 
+    /// Return the next ClientNotification packet destined to the phone.
+    meshtastic_ClientNotification *getClientNotificationForPhone() { return toPhoneClientNotificationQueue.dequeuePtr(0); }
+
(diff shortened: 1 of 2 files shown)


jp-bennett (approved):
Not seeing anything problematic.

PR #4843, opened 2024-09-24
Build message in printBytes, to not spam BLE log

jp-bennett:
For @garthvh


fifieldt (approved):
FWIW I checked our current code and found our largest label so far is 52

fifieldt:
built, tested on t114, works fine.

fifieldt (approved):

PR #4820, opened 2024-09-22
Consolidate variant build steps

scruplelesswizard:
## Description

Creates an action with common build steps across the following variants:
- ESP32
- ESP32-C3
- ESP32-S3
- NRF52
- Raspberry Pi 2040
- STM32

### Iterations
- #4806

PR #4845, opened 2024-09-24
Revert "Consolidate variant build steps"

fifieldt:
Reverts meshtastic/firmware#4820


fifieldt:
Rolling forward instead.

PR #4847, opened 2024-09-24
sed doesn't like newlines

scruplelesswizard:


fifieldt (approved):

fifieldt (changes requested):
Need to fix the rm -rf . line at the end. Causes nrf52 to fail

fifieldt (approved):

PR #4848, opened 2024-09-24
chomp trailing newline

scruplelesswizard:
Adds a chomping indicator to strip trailing newlines

https://yaml.org/spec/1.2.2/#8112-block-chomping-indicator


fifieldt (approved):
Makes sense.

PR #4859, opened 2024-09-24
fix #4844

### LINKED ISSUE #4844: [Enhancement]: Disable onboard buzzer on wakeup
### Platform

ESP32

### Description

On devices with an onboard buzzer in their firmware, like the T1000-e, when the device wakes to transmit in SENSOR or TRACKER roles (possibly with other power settings?) it plays the bootup tune.
I brought this up on discord and @caveman99 mentioned that I should open a ticket because it was "counter productive" for these roles and I agree, it's likely an unintended edge-case. 
It would be nice to have either a setting to disable the bootup tune, or possibly, detect when waking with these roles enabled and know not play it.
I dunno about others, but I would be happy if my backpack didn't sing every 30 minutes :)

caveman99:


src/main.cpp
```diff
@@ -624,7 +624,13 @@ void setup()
...
 #endif
 
-    playStartMelody();
+    // only play start melody when role is not tracker or sensor
+    if (config.power.is_power_saving == true && (config.device.role == meshtastic_Config_DeviceConfig_Role_TRACKER ||
+                                                 config.device.role == meshtastic_Config_DeviceConfig_Role_TAK_TRACKER ||
+                                                 config.device.role == meshtastic_Config_DeviceConfig_Role_SENSOR))
+        LOG_DEBUG("Tracker/Sensor: Skipping start melody\n");
+    else
+        playStartMelody();
```
thebentern, line 633: What about shutdown melody?

caveman99: i don't think that's triggered on sensor sleep mode. At least  i didn't find a reference for it.

PR #4867, opened 2024-09-25
Fix (some ?) memory alignment issues on the crypto part - resulting in crashes or strange behavior

### LINKED ISSUE #4855: [Bug]: RP2040 - Decrypt or Encrypt (during remote admin for example) crashes the device
### Category

Hardware Compatibility

### Hardware

Other

### Firmware Version

2.5 +

### Description

Hi,

Since 2.5+, my rp2040-lora was unusable. When trying to remote admin, the devices just crash.
I finally took time to do some investigations and found out it crashes just after the "Attempting PKI decryption" message.

I found out that there was some pointer dereference on "extraNonce" on the "encryptCurve25519" and "decryptCurve25519" functions causing this crash. The pointer itself is not bad (valid address) but was not word-aligned on the memory. While it may work on most hardware (with non-guaranted results), it makes rp2040 crash.

Good news is I have a fix, which you may use or not, depending on how you want to handle this situation.

Firstly, I replaced the dereferences "*extraNonce" with a memcpy, like that on 2 or 3 places : 

` *extraNonce = extraNonceTmp;    ==> memcpy(extraNonce, &extraNonceTmp, 4); // do not use dereference on non aligned pointers`

I tested, and it worked further until another crash. I again found out a lot of pointer dereference into the byte buffer in the aes code, which is not very good news (you can't replace every dereference, and don't want to touch the aes code !), but can be fixed in this case with a simple alignment directive on the buffer itself during declaration (the buffer was not aligned at all !) : 

```
static uint8_t bytes[MAX_RHPACKETLEN] __attribute__ ((__aligned__(4)));
static uint8_t ScratchEncrypted[MAX_RHPACKETLEN] __attribute__ ((__aligned__(4)));
```

With these 2 tricks, my rp2040 is finally working fine on 2.5+ on decrypt AND crypt (tested on 2.5.1). 
The directive is not sufficient, because the code use as nonce the last 4 bytes of the buffer, and this buffer does not always have a size modulo 4. But you can easily change by using another way to get a nonce. 

Also, it may explain arbitrary crashes on other devices.

I'm not sure I explained it right, but I can answer questions :)

Thanks !

### Relevant log output

_No response_

TheMalkavien:
Hi,

As discussed into the #4855 issue, here is my humble PR.

It fixes #4855 by avoiding the use of non aligned pointers. On some hardware, we do not care at all, but for some the result is not guaranteed, and it makes some hardware crash (rp2040 crashes, Heltecv3 PKI_FAILED, ...).

There is 2 main changes :
- Firstly, avoiding to dereference the extraNonce pointer, wich is pointing to a bytebuffer, at an address we cannot be sure is properly aligned (it depends on the size of the buffer, and the alignment of the buffer itself).
- Secondly, the two crypto bytebuffers are now aligned at declaration, it makes the use of pointers inside, a little more safe (and it fixes multiples crashes during the crypto algorithm which use pointers to the buffers).

As discussed, these two changes are a quick way to fix crashes or bugs linked to the non-alignment of some objects. It will probably not fix all the alignment problems in the future (or even in the present), but it should be considered to avoid such difficult to debug behavior.
Maybe the compiler can provide some useful flags for that, and of course be very careful when using pointers (as usual ? :p ).

Tested and working on RP2040-lora and Heltecv3

Thanks !


thebentern (approved):

jp-bennett:
It looks sane to me, though memory alignment isn't something I'm used to worrying about (obviously). I should have a rp2040 platform by the end of today to test with.

caveman99 (approved):

GUVWAF:
I'm wondering whether we should align the `radiobuf` (used when receiving) as well: 
https://github.com/meshtastic/firmware/blob/9bebad2dbe3650c133d6efac5e028d52ce7fa12a/src/mesh/RadioInterface.h#L94 

As we're doing pointer arithmetic for the payload afterwards:
https://github.com/meshtastic/firmware/blob/9bebad2dbe3650c133d6efac5e028d52ce7fa12a/src/mesh/RadioLibInterface.cpp#L392

TheMalkavien:
Hi,

I'd say, every structure should be aligned for performance at least. I'm rusty, but I found it strange that the compiler does not align global variables by default.

Anyway, in your example, even if the start of the buffer is properly aligned, there is a risk that if sizeof(PacketHeader) is not modulo "word", your pointer will be not aligned.
I'ts not a problem if you use it byte by byte, but using it with a dereferrence will mostly cause problems.

Everything I say must be verified, I'm not myself an expert in memory alignment or gcc specificities.

GUVWAF:
@TheMalkavien Thanks, makes sense. We force `sizeof(PacketHeader)` to be 16, but that might thus be an issue as well. Not sure how far we should go with this, but at least this PR looks good to me.

GUVWAF (approved):

jp-bennett (approved):

jp-bennett:
#4878 adds a test case that crashes the rp2040. I've not yet been able to trigger the Heltec v3 flaw. This patch does fix that crash and get the test suite passing again on rp2040.

Issue #4775, opened 2024-09-19
[Bug]: Unclear Issues with LR1110; "my T-1000-E is deaf"


fifieldt:
### Category

Hardware Compatibility

### Hardware

Seeed Card Tracker T1000-E

### Firmware Version

2.5.0

### Description

As discussed in Discord over the past 3 weeks, there are multiple instances of people reporting issues with LR1110 radios. This issue is being created to serve as a central place for recording this information and coordinate testing.

caveman99:
> @Talie5in has identified #4621 as the breakpoint.

Are the other devices on the same firmware level or older/newer? We have radiolib changes scattered in there somewhere.

fifieldt:
Currently trying: 33eb073535c2bece2353b9fd76d0dda5d120ac05 with a later version of RadioLib...

fifieldt:
OK, https://github.com/meshtastic/firmware/commit/33eb073535c2bece2353b9fd76d0dda5d120ac05 with RadioLib 7.0.0 results in the bad outcome.

By my logic, using the good meshtastic commit with just radiolib changes means the issue might be in RadioLib!

fifieldt:
RadioLib 6.6.0 (working version) is May 29 https://github.com/jgromes/RadioLib/commit/954477b0c0589e6373692ce569ca117191667c30 .

Stepping forward through RadioLib with known-good Meshtastic version.

fifieldt:
June 1    7a0b5bae9906905f32cd840ec79b2665f509f717 - OK

Talie5in:
> > @Talie5in has identified #4621 as the breakpoint.
> 
> Are the other devices on the same firmware level or older/newer? We have radiolib changes scattered in there somewhere.

The one example is my WisMeshPockt (Rak4631) on 2.5.0 Alpha (through the entire time I was testing different 2.5.0 TechPreview build - it was out in my car (50m away) and was the only device it would hear past https://github.com/meshtastic/firmware/commit/33eb073535c2bece2353b9fd76d0dda5d120ac05

GUVWAF:
That RadioLib commit introduces differences in the IRQ flags handling in `startReceive()`, so if using Meshtastic commit https://github.com/meshtastic/firmware/commit/33eb073535c2bece2353b9fd76d0dda5d120ac05 it is expected to break, but PR #4621 also should handle those API changes.

fifieldt:
[...]

PR #4885, opened 2024-09-27
Potential fix for bad Rx performance on T1000-E

GUVWAF:
This seems to fix the part of the issue from #4775 where it only hears SX126x/LR11x0 radios that are close.
While the RadioLib example mentions `setRfSwitchTable()` should be called before `begin()`, we called `setDioAsRfSwitch()` (which gets called inside `setRfSwitchTable()`) at this place before https://github.com/meshtastic/firmware/commit/2f0c19ebeac0054cb8b8f417b9f2952d18532fce.

### DIFF
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -3,6 +3,15 @@
 #include "configuration.h"
 #include "error.h"
 #include "mesh/NodeDB.h"
+#ifdef LR11X0_DIO_AS_RF_SWITCH
+#include "rfswitch.h"
+#else
+static const uint32_t rfswitch_dio_pins[] = {RADIOLIB_NC, RADIOLIB_NC, RADIOLIB_NC, RADIOLIB_NC, RADIOLIB_NC};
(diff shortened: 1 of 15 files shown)


caveman99:
Let's try it. setDioAsRfSwitch is godmode only btw... :-) I just went by the documentation.

caveman99 (approved):

GUVWAF:
@caveman99 Can you try if this doesn't break on another LR11x0 dev board you have?

caveman99:
> @caveman99 Can you try if this doesn't break on another LR11x0 dev board you have?
I will tomorrow morning. Also i wanna put the Switch matrix into variant.h or next to it for each board and just leave the default in the firmware code. Not have a giant ifdef construct there.

thebentern (approved):

caveman99:
> @caveman99 Can you try if this doesn't break on another LR11x0 dev board you have?

seems to work fine with all boards, can you review my changes?

thebentern (approved):

GUVWAF:
Looks good and clean!

PR #4888, opened 2024-09-27
Ignore seen phoneapi packets

thebentern:
Turns out some of the Phone API spam is actually replaying the same packets. This completely silences the Rate limit notification spam on my devices.

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -8,7 +8,7 @@ plugins:
       uri: https://github.com/trunk-io/plugins
 lint:
   enabled:
-    - trufflehog@3.82.4
+    - trufflehog@3.82.5
     - yamllint@1.35.1
     - bandit@1.7.10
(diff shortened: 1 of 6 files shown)


GUVWAF:
Do we know why it is replaying the same packets?

thebentern:
> Do we know why it is replaying the same packets?

No, I have been trying to figure out if it's somehow related to the push / pull system of notify and read value from characteristic or something going haywire in the bluetooth stack in iOS. I have not reproduced it on the Android side, but I have not done much testing there.

> I'm not sure if all clients actually set a packet ID or leave it 0, which would not work with this.

I know iOS is generating packet IDs, but I'll double check if Android as well.

jp-bennett:
Or just make an exception for packets with num of 0

thebentern:
> Or just make an exception for packets with num of 0

Easy addition. I'll whip this up

PR #4892, opened 2024-09-28
Only log "Setting DIO2 as RF switch" when doing so and fix battery level

dahanc:
…do it. Also, if there's an error setting DIO2, log the error code.

Super minor, but I mentioned this in Discord a few days ago, and it seemed like @jp-bennett thought it was a good idea?

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -1068,10 +1068,9 @@ bool Power::axpChipInit()
 #if !MESHTASTIC_EXCLUDE_I2C && !defined(ARCH_PORTDUINO) && !defined(ARCH_STM32WL)
 
 /**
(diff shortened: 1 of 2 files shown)


GUVWAF:
This indeed fixes battery level values of `-0.00` for boards without MAX17048.

@dhskinner can you double-check if the MAX17048 still works for you when using this?

dahanc:
Oops, I'm not good at GitHub; I didn't actually intend for the battery level commit to be included in this PR. I guess I should've made the changes on separate branches?

GUVWAF:
@dahanc No worries, since the first one is pretty small, I think it's okay like this.

caveman99 (approved):

PR #4898, opened 2024-09-29
When importing config, keep Bluetooth on and defer rebooting until co…

### LINKED ISSUE #4812: [Bug]: Export and import configuration on Heltec T114 does not work
### Category

Other

### Hardware

Heltec Mesh Node T114

### Firmware Version

2.5

### Description

After setting up the T114 I exported the config before doing some testing. When I tried to re import the complete config again the only thing that was reverted after a clean flash was the node name, every other module, device, external notification, canned messages, position etc. was untouched and empty. 

### Relevant log output

_No response_

dahanc:
…nfig is committed

With this change, importing the config from Android over Bluetooth hangs with the progress bar showing 92%, but I think all of the config does get imported. Maybe the device reboots before sending an acknowledgement for the commit edit settings packet?

fixes #481

[Download firmware-2.5.4.677e091.zip. This artifact will be available for 90 days from creation](https://nightly.link/meshtastic/firmware/actions/artifacts/1992271328.zip)

nfig from Android over Bluetooth hangs with the progress bar showing 92%, but I think all of the config does get imported. Maybe the device reboots before sending an acknowledgement for the commit edit settings packet?

fixes #4812

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -592,7 +592,8 @@ void AdminModule::handleSetConfig(const meshtastic_Config &c)
 
 void AdminModule::handleSetModuleConfig(const meshtastic_ModuleConfig &c)
 {
-    disableBluetooth();
+    if (!hasOpenEditTransaction)
+        disableBluetooth();
     switch (c.which_payload_variant) {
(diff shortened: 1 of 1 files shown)


thebentern (approved):

thebentern:
Great find!

PR #4907, opened 2024-09-30
Consolidate and shrink down the re-used strings in logs

thebentern:
Tiny bit of saving some precious flash on RAK4631.


caveman99 (approved):

fifieldt:
nice one

PR #4912, opened 2024-09-30
Fix #4911 : Partially rework some code to remove warnings about potential non-aligned memory accesses

### LINKED ISSUE #4911: [Bug]: More potential non aligned memory accesses (focused on RP2040 but not only)
### Category

Hardware Compatibility

### Hardware

Other

### Firmware Version

2.5+

### Description

Hi,

After the issue #4855 was resolved by my little patch, I was thinking this was just the visible part. After discussing it on the discord I tried the -Wcast-align gcc flag which throws a warning when the code uses some **potential** non aligned memory pointer.
The **potential** is important here because it checks at the compilation time, not run time, so it does not know if it will unaligned or not, it just knows it could be depending on the rest of the code.

The result is interesting, some warning have to be fixed, some not (imho). I will not paste here all the warnings, because there is 90 of them when compiling for the rp2040. Il will just put them by files : 

1. .pio\libdeps\rp2040-lora\Crypto\ChaCha.cpp : 3 warning
2. .pio/libdeps/rp2040-lora/INA3221/src/INA3221.cpp : 77 warnings
3. .pio\libdeps\rp2040-lora\RadioLib\src\modules\LR11x0\LR11x0.cpp : 2 warnings
4. /src/gps/GPS.cpp : 1 warning
5. /src/gps/NMEAWPL.cpp : 2 warnings
6. /src/mesh/aes-ccm.cpp : 2 warnings
7. /src/mesh/CryptoEngine.cpp : 1 warning
8. /src/mesh/RadioInterface.cpp : 1 warning
9. /src/mesh/RadioLibInterface.cpp : 1 warning

At first, I thought it was a hard work to come. But I suggest we do not care about warnings in the libraries. We can trust the library to "know better than me how to handle pointers". That means we have to be careful of what we gave it in input.
Then I suggest not to touch aes-ccm, as it contrains important crypto functions, I saw the copyright so this is maybe a library too. Trying to change a crypto function may lead to very big problems, I prefer to avoid this one.

I will focus on the others, which will be profitable for others platform as well : 
4. /src/gps/GPS.cpp : 1 warning
5. /src/gps/NMEAWPL.cpp : 2 warnings
7. /src/mesh/CryptoEngine.cpp : 1 warning
8. /src/mesh/RadioInterface.cpp : 1 warning
9. /src/mesh/RadioLibInterface.cpp : 1 warning

With very little effort, I think we can fix 6 warnings, but more important with some simple change we can better sanitize a lot of buffer accesses, especially the radiobuffer.

I'll make a PR very soon to show you what I am thinking of so you will decide what to do on this topic.

Regards,

### Relevant log output

_No response_

TheMalkavien:
Hi,

As discussed in the issue #4911, here is my approach to contain the memory alignment issue

There is some very easy fixes based on memcpy insted of pointer arithmetic and dereferrence. This is a quick win.

But I had to think of another way to work with the radiobuffer bytebuffer, because of how it was previously handled, the only way was using pointer arithmetic to access payload after the header.

So I just replaced the byte buffer by a struct containing header then payload : 

```
/**
 * This structure represent the structured buffer : a PacketHeader then the payload. The whole is
 * MAX_LORA_PAYLOAD_LEN + 1 length
 * It makes the use of its data easier, and avoids manipulating pointers (and potential non aligned accesses)
 */
typedef struct {
    /** The header, as defined just before */
    PacketHeader header;

    /** The payload, of maximum length minus the header, aligned just to be sure */
    uint8_t payload[MAX_LORA_PAYLOAD_LEN + 1 - sizeof(PacketHeader)] __attribute__ ((__aligned__));

} RadioBuffer;
```

As you can see, this is very simple. And I replaced all the pointer code by struct accesses : 

```
PacketHeader *h = (PacketHeader *)radiobuf;
h->from
```
becomes :
`radioBuffer.header.from`

and
`radiobuf + sizeof(PacketHeader)`
becomes :
`radioBuffer.payload`

Its more elegant too :)

The result is, less memory-aligned warnings, and a more safe code to use, and depending of your coding style, more readable I guess.
I quickly tested it on RP2040, HTv3 and HT Tracker 1.1 (to test the gps part) without seeing any issue.

This should Fix #4911 (partially at least) and gives a good start (or just ideas) to a more robust code.

Regards,


caveman99 (approved):
LGTM. We need to make sure the packetheader stays the same size and alignment no matter what processor architecture we're running on.

PR #4916, opened 2024-09-30
Regenerate public key on boot, to avoid accidental mismatch.

jp-bennett:
A failure mode that we might accidentally enable is an old or bogus public key paired with a new private key. It's trivial to regenerate the public key on boot to avoid this

[Download firmware-2.5.4.5cf6ffc.zip. This artifact will be available for 90 days from creation](https://nightly.link/meshtastic/firmware/actions/artifacts/1997929574.zip)

ld or bogus public key paired with a new private key. It's trivial to regenerate the public key on boot to avoid this.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -133,7 +133,7 @@ NodeDB::NodeDB()
         config.security.is_managed = config.device.is_managed;
     }
 
-#if !(MESHTASTIC_EXCLUDE_PKI_KEYGEN)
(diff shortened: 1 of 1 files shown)


thebentern (approved):

caveman99 (approved):

caveman99:
why is STM32 failing all of a sudden on this?

jp-bennett:
> why is STM32 failing all of a sudden on this?

I probably goofed up the ifdef

Issue #4738, opened 2024-09-16
Addtional Sensor for Farm Animals Health


nextechDI:
Greetings, congratulations to all the developers and collaborators, this project is fantastic, I wanted to contribute about use cases and sensors, I think that the meshtastic firmware for monitoring farm animals has a lot of potential by allowing several sensors to be integrated with the same micro, but I would recommend adding the MLX90614 temperature sensor to the telemetry firmware, which is the most suitable for sensing temperature in animals. I would also recommend adding the MAX30102 heart rate sensor and oximeter, which offers an excellent quality-price ratio for this function. These two sensors together Those that already include the meshtastic firmware would allow it to perform very well as a farm animal tracking device.

dhskinner:
Hi @nextechDI I have exactly zero of those sensors but don't mind at all contributing an integration if you're ok to do some testing. From a quick skim both have reasonable library support:
- [Adafruit-MLX90614-Library](https://registry.platformio.org/libraries/adafruit/Adafruit%20MLX90614%20Library)
- [sparkfun/SparkFun MAX3010x Pulse and Proximity Sensor Library](https://registry.platformio.org/libraries/sparkfun/SparkFun%20MAX3010x%20Pulse%20and%20Proximity%20Sensor%20Library)

Also consider that you can get other info from the accelerometer such as what-do-cows-do-when-no-one-is-looking?  (if only we had the data... those cows are up to something I'm sure)

@thebentern might need to add some new fields to [https://github.com/meshtastic/protobufs](https://github.com/meshtastic/protobufs) for oxygen level and pulse rate? Appreciate your thoughts

thebentern:
@dhskinner that was super spooky timing 👀 
https://github.com/meshtastic/protobufs/pull/590

fifieldt:
Popular issue :) I've also just acquired these sensors so can contribute to code and testing.

thebentern:
I'll push a branch after the protobuf classes regenerate and we'll collab.

fifieldt:
Hi @nextechDI , we've just about got a prototype for telemetry using those sensors you mentioned. Do you have devices available to help test in the real world? We can send updated firmware...

nextechDI:
> Hi @nextechDI I have exactly zero of those sensors but don't mind at all contributing an integration if you're ok to do some testing. From a quick skim both have reasonable library support:
> 
[...]

PR #4927, opened 2024-10-02
Add health telemetry module

thebentern:
Adds a new telemetry module covering health aspects - body temperature, oxygen saturation and heart rate. Currently the MLX90614 IR temperature sensor and MAX30102 temperature/oxygen saturation/heart rate sensors are supported.

The module will add a frame to devices with a screen, and broadcast to the mesh based on the set interval.

Fixes https://github.com/meshtastic/firmware/issues/4738


fifieldt:
For reference - looked up emissivity and cows and humans are both 0.98, so that's a good value if we need to set that for temperature sensors. Pigs also >0.95.  If this needs to go on furry animals like a mouse, then we need to revisit :)

fifieldt:
I just wrote up the MLX90614 temp sensor quickly as an EnvironmentalTelemetry sensor to get it working. It works now. So we can steal the code for HealthTelemetry testing :)

fifieldt:
OK, and there's the MAX30102 - or, at least the temperature bit - working.

fifieldt:
now for the fun bit?

fifieldt:
Big difference when compared with EnvironmentalTelemetry seems to be that we need to get multiple readings in order to get SPO2 or heart rate. This might end up being more like GPS where we turn it on to get a 'fix', and then sleep it.

fifieldt:
After the last set of changes, spO2 and temperature both look accurate.

Heart rate is too high - tried increasing bus speed to 400Khz but no change, also tried taking 8 seconds (original: 4 secs) of samples and no change.

Also, only works on nrf52 due to some hard-coded references to SCL/SDA.

fifieldt:
Fixed the SCL/SDA stuff by switching libraries for MAX90614.

Screen works!

fifieldt:
"Common libs for environmental measurements in telemetry module are not included in native / portduino" - which is why those builds are failing. Need to fix.

fifieldt:
Also need to wrap the screen stuff in a HAS SCREEN it seems.

fifieldt:
"It works!"

HarukiToreda:
any sneak peek on how the screen will display this data?

thebentern:
DEBUG | ??:??:?? 104 [HealthTelemetryModule] heart_rate=187(1), sp02=100(1)(Sending): temperature=33.500000, heart_bpm=187, spO2=100
DEBUG | ??:??:?? 104 [HealthTelemetryModule] Partially randomized packet id 424126784
INFO  | ??:??:?? 104 [HealthTelemetryModule] Sending packet to mesh

PR #4933, opened 2024-10-02
preliminary Othernet Dreamcatcher Support

caveman99:
Need to adapt to 2301 final version. Will probably be able to autodetect that.


fifieldt:
Saw your posts on https://forums.othernet.is/t/othernet-short-term-and-long-term-status-update/7946 :) Seems sad news for the Othernet project

markbirss:
Awesome effort @caveman99

src/detect/ScanI2CTwoWire.cpp
```diff
@@ -406,6 +406,7 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
                 SCAN_SIMPLE_CASE(FT6336U_ADDR, FT6336U, "FT6336U touchscreen found\n");
                 SCAN_SIMPLE_CASE(MAX1704X_ADDR, MAX17048, "MAX17048 lipo fuel gauge found\n");
                 SCAN_SIMPLE_CASE(MLX90614_ADDR, MLX90614, "MLX90614 IR temp sensor found\n");
+                SCAN_SIMPLE_CASE(TPS65233_ADDR, TPS65233, "TPS65233 BIAS-T found\n");
```
fifieldt, line 409: since TPS65233_ADDR is 0x60 , don't you need to de-conflict with the BME-280 ?

src/main.cpp
```diff
@@ -298,6 +298,11 @@ void setup()
     digitalWrite(VEXT_ENABLE, VEXT_ON_VALUE); // turn on the display power
 #endif
 
+#if defined(BIAS_T_ENABLE)
```
fifieldt, line 301: Is a BIAS_T likely to be included in future variants? Otherwise, is this better in variant-specific code to keep main clean?

variants/dreamcatcher/platformio.ini
```diff
@@ -0,0 +1,27 @@
+[env:dreamcatcher] ; 2301, latest revision
+extends = esp32s3_base
+board = esp32s3box
```
fifieldt, line 3: Just checking: is there a new board to add to /boards here?

variants/dreamcatcher/variant.h
```diff
@@ -0,0 +1,107 @@
+#undef I2C_SDA
+#undef I2C_SCL
+#define I2C_SDA 16 // I2C pins for this board
+#define I2C_SCL 17
+
+#define I2C_SDA1 45
```
fifieldt, line 6: Do you need
#define WIRE_INTERFACES_COUNT 2
?

variants/dreamcatcher/variant.h
```diff
@@ -0,0 +1,107 @@
...
+#define USE_LR1120
+
+#define SPI_MISO 37
+#define SPI_MOSI 39
+#define SPI_SCK 38
+#define SDCARD_CS 40
+
+#define PIN_BUZZER 48
+
+#define GPS_RX_PIN 10
```
fifieldt, line 33: Just checking - need HAS_GPS 1?

fifieldt (commented):
Left some minor comments for you

src/main.cpp
caveman99: There are 2 kinds of BIAS-T on the board. Once supplies a switchable DC offset (14 or 18V) to power an active antenna and switch its polarity. we generally want that off. The other possibility is a 5V DC offset on the antenna port to power an active antenna (also possible for GPS). So i'd leave that in without additional guards as it could be useful for other board designs. I used existing definitions for those switches where possible and added just those that were'nt available with existing methods.

These beasts have 6 GPIO's controlling the switching (!) :-)

variants/dreamcatcher/platformio.ini
caveman99: nope, there is a stock ESP32 solder core on the board, and this is it.

variants/dreamcatcher/variant.h
caveman99: The board does not have a soldered GPS. There's a serial header to connect one. Other uses would be the serial module. This was more a reminder to me what the (changing) pinout for those pins was.

variants/dreamcatcher/variant.h
caveman99: no, if you don't define it it will auto-adjust. configuration.h line 221...
```

#ifndef WIRE_INTERFACES_COUNT
// Officially an NRF52 macro
// Repurposed cross-platform to identify devices using Wire1
#if defined(I2C_SDA1) || defined(PIN_WIRE1_SDA)
#define WIRE_INTERFACES_COUNT 2
#elif HAS_WIRE
#define WIRE_INTERFACES_COUNT 1
#endif
#endif
```

caveman99:
Just adding a picture of the latest installment, so that people know what we are talking about.

d032d967bca847dd50fe0bb44e51071789297a97

markbirss:
@caveman99 

i went to have a look also, just the product not in stock at the moment and the price is bit high

markbirss:
I still have the older version looks quite different as it was just sat add-on for Raspberry Pi
image

variants/dreamcatcher/variant.h
caveman99: i'll change them to comments, so people know what to look for when defining pins for either use case. the header is dead center on the board.

caveman99:
> @caveman99
> 
> i went to have a look also, just the product not in stock at the moment and the price is bit high

I don't think it will come back into stock, these firmware images are for existing devices to be repurposed after the original service is discontinued at the end of 2024.

caveman99:
> I still have the older version looks quite different as it was just sat add-on for Raspberry Pi !

yes, this was a regular sat RX without lora demod. At that stage you needed a satellite dish. They switched to Lora to allow reception with a bare antenna without reflector. The kit you have is a good data receiver, but there has not been coverage for these for years.

thebentern (approved):

src/detect/ScanI2CTwoWire.cpp
caveman99: Also BME280 has Addr 0x76 or 0x77 - it's just the subtype register that's 0x60. Guarded anyway

fifieldt: ah! sorry for making you go to the extra effort :(

fifieldt (approved):

PR #4938, opened 2024-10-03
Add a Userprefs Timezone String, to be replaced in the web flasher

jp-bennett:


thebentern (approved):

thebentern (approved):

GUVWAF:
Maybe we can do the same with the `BUILD_TIME` macro? Currently it's still rebuilding the whole firmware every time, because this changes in the platformio.ini file.

GUVWAF:
Maybe we can do the same with the `BUILD_TIME` macro? Currently it's still rebuilding the whole firmware every time, because this changes in the platformio.ini file.

thebentern:
> Maybe we can do the same with the `BUILD_EPOCH` macro? Currently it's still rebuilding the whole firmware every time, because this changes in the platformio.ini file.

GENIUS! I've been almost regretting the mechanism for that one because I'm commenting it out locally to avoid rebuilds. 🤣

thebentern:
Actually I think I have a quick way of fixing this the BUILD_EPOCH. PR incoming!

PR #4943, opened 2024-10-04
Leave the build epoch commented and uncomment when CI runs

thebentern:
Fixes the annoying side-effect of the Platform IO having changes for every build and rebuilding the ENTIRE firmware every time.


GUVWAF (approved):
Even simpler!

fifieldt (approved):
<3

jp-bennett (approved):
Sed for the win

PR #4956, opened 2024-10-05
Retire PPR Boards

fifieldt:
These boards appear to be no longer supported by Othernet (which itself has questions of viability?).
Retire the PPR and PPR1 boards/variants.


caveman99 (approved):
Afaik these never went beyond prototype stage

PR #4958, opened 2024-10-05
Remove rak4631_epaper_onrxtx variant

fifieldt:
Appears to be a testing variant of rak4631_epaper. Due to little information available, let's remove it for now.


caveman99:
This one was instated here ... https://github.com/meshtastic/firmware/pull/2391 - what about we make a separate repository in the org to collect DIY templates/examples for variant files?

markbirss:
@caveman99 @fifieldt 

Yes, i must just test that target again, with the current board support, also have some other wip variants for example too add e-ink and tft to wio-tracker

markbirss:
> it would have been more beneficial to have added the support of RAK's RAK14000 E-Paper Module, which I bought assuming this variant was for, but it is not. I agree with it's removal too

"RAK's RAK14000 E-Paper" is in the rak6431_epaper variant ?

HarukiToreda:
> > it would have been more beneficial to have added the support of RAK's RAK14000 E-Paper Module, which I bought assuming this variant was for, but it is not. I agree with it's removal too
> 
> "RAK's RAK14000 E-Paper" is in the rak6431_epaper variant ?

where can I find this variant? I only see this rak4631_eink one and it does not work. screen is attached and it does not respond with the eink display module attached.
image

markbirss:
@HarukiToreda 

you have your flex cable connected incorrectly ? (also it best to secure the e-ink board with screws)
 
image

HarukiToreda:
> @HarukiToreda
> 
> you have your flex cable connected correctly ? to face upward it connects to the backside of the e-ink board
> 
> (also it best to secure the e-ink board with screws)
> 
> image

attached the flex cable to the to the connector in the back and it works, thank you. This has been sitting on my drawer for months

PR #4976, opened 2024-10-06
First version of a DeepSleep state for the RP2040

TheMalkavien:
Hi,

As I use mainly RP2040 with solar power, I needed a way to reduce power consumption, or being able to put the node to sleep during very low battery charge.

The RP2040 can be put to deepsleep, I did not find a lot of example (well precisely I did find multiple way to do so, but not always very clear to me), so I worked on a first version based on the multiple examples I found.

My changes are based on parts of the pico-extra lib (https://github.com/raspberrypi/pico-extras/).

What it does : 
 - RP2040 is now able to "shutdown" (forever sleep)
 - RP2040 is now able to DeepSleep for a given time - works well when power state is LowBatt !

What is does not : 
 - RP2040 cannot fully restore its state after being awoken, so it just reboots to become operational again.

I measured the power consumption during deep sleep to around 3 mah. Its still high, but way better than the 30-35 when it is awake.

There is A LOT of place for improvement, I'm not sure I can do better myself cause I'm not an expert of this device, nor a good developer, but it is really a good start (and works for me).

I updated some RP2040 libraries to be able to use pico-extra.

Issue : Rarely, the radio chip does not want to go to sleep, even after the 30 seconds delay. Maybe this delay is to short, maybe the channel is too occupied, I dont know.

P.S. : Sorry my Trunk addon for VS does not work anymore, it crashes so the code may not be very clean.

Regards,


TheMalkavien:
Hum I see that the rak11310 misses a #define ledOff(pin) pinMode(pin, INPUT) in variants.h.

Maybe the best is for now to limit my changes for the RP2040-lora only ? I did not test on another rp2040 devices. 
What do you think ?

thebentern (approved):

caveman99 (approved):

Issue #4687, opened 2024-09-11
[Feature Request]: Is there an option on/off bluetooth with q10 bb? 


sim1988son:
### Platform

ESP32

### Description

I'm searching through the entire program/code and I can't find the on/off bluetooth option from the device position?

todd-herbert:
I think #4977 does a lot of the ground-work for this, but I think it probably only adds the hotkey for the M5Stack CardKB. 
It looks like Q10 BB / T-Deck keyboard would each need a few lines of code added to `input/kbI2cBase.cpp` (and someone to test).

hayschan:
Is the open source hardware Blackberry keyboard adapter boar [solderparty/pmod_bbq10_keyboard](https://github.com/solderparty/pmod_bbq10_keyboard) supported by Meshtastic now?

![image](https://github.com/user-attachments/assets/3ebfe83e-e3d9-4012-b21c-54a38b97320e)

Issue #4315, opened 2024-07-22
[Feature Request]: Turn off Bluetooth on Standalone devices by pressing Fn+b


HarukiToreda:
### Platform

ESP32

### Description

Standalone devices are by design meant to be ran without a phone, and as such it would be ideal to give them the ability to use the keyboard and with the key conbination Fn+b it should dissable the bluetooth and restart the device as if done by a phone.  If there's a need for the bluetooth again it should be able to toggle it back on with the same key conbination. 

Being able to completely shut off the bluetooth will also allow ESP32 devices to cut the power consumption by half since it currently spends 59mA when on.

PR #4977, opened 2024-10-06
Toggle Bluetooth with Fn+b shortcut

HarukiToreda:
Problem:
As many are aware, ESP32 devices are known for their high power consumption. For instance, the Heltec ESP32 V3 draws around 110mA when powered on with the screen active and connected to a phone via Bluetooth. The Bluetooth radio alone is responsible for approximately 50mA of that consumption. For keyboard-based standalone devices, which rarely need Bluetooth other than for changing settings, users were forced to keep Bluetooth on regardless of necessity. There was no way to toggle Bluetooth on or off without physically connecting the device to a computer via serial or using the admin channel, which required another node for access.

Solution:
I implemented a new feature that allows users to turn off Bluetooth on keyboard devices by pressing Fn+b and turn it back on when needed. This enhancement significantly improves power efficiency for these devices.

Result:
With Bluetooth off, the device now consumes only 55mA. When combined with Power Save mode, the consumption can drop as low as 11mA, a substantial reduction from the previous 110mA. Users can still easily reconnect to a phone using the shortcut when necessary, offering greater flexibility and extended battery life.

Issue discussed on #4315
and can possible be repurposed for #4687 by someone else.

### DIFF
--- a/src/input/InputBroker.h
+++ b/src/input/InputBroker.h
@@ -18,6 +18,7 @@
 #define INPUT_BROKER_MSG_RIGHT 0xb7
 #define INPUT_BROKER_MSG_FN_SYMBOL_ON 0xf1
 #define INPUT_BROKER_MSG_FN_SYMBOL_OFF 0xf2
+#define INPUT_BROKER_MSG_BLUETOOTH_TOGGLE 0xAA
 
 typedef struct _InputEvent {
     const char *source;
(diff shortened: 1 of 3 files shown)


thebentern:
@HarukiToreda thought about just shutting down `bluetooth->enabled(false)` as an alternative? Completely disabling and rebooting seems slower and heavy handed

HarukiToreda:
I thought about it, but unless I fully reboot the node, I could not get the bluetooth to fully turn off to achieve the power savings as they do now.

> ut just shutti

HarukiToreda:
I'm going to give it a test to see for sure.

HarukiToreda:
I discussed on options with @todd-herbert was able to prevent a reboot using the disableBluetooth(); command, current tested at 47-55mA, it doesn't require a reboot to turn off, but it does need reboot to turn back on.

caveman99 (approved):

PR #4982, opened 2024-10-07
Remove has_rx * on installDefaultDeviceState

### LINKED ISSUE #4007: [Bug]: Factory reset not actually resetting node.
### Category

Other

### Hardware

Heltec Wireless Tracker

### Firmware Version

2.3.10.d19670b

### Description

Performing a factory reset from the Android app on Heltec Wireless Tracker v1.1 appears to leave old messages and nodes (now all unknown name) on the device. 

Steps to reproduce:
1. Be connected to node via Bluetooth to your Android device. 
2. Go to settings and select the Factory Reset option. Confirm. Whatch node reboot.
3. Re-pair and reconnect to same Android device.
4. See message on node screen. It's from ??? 2 hours ago.(I did receive the message 2 hours ago before driving to a different city on different side of mountain range and factory resetting. Pretty sure it's not being retransmitted.)
5. Scroll through screen and see a bunch of Unknown Name nodes. 
(This is without setting a region from the Android app)

Update:
The Node now also displays on screen that it is connected to other nodes. I still haven't set up a region since the factory reset.
Update:
### Relevant log output

_No response_

thebentern:
Closes #4007

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -626,6 +626,8 @@ void NodeDB::installDefaultDeviceState()
     // devicestate.node_db_lite_count = 0;
     devicestate.version = DEVICESTATE_CUR_VER;
     devicestate.receive_queue_count = 0; // Not yet implemented FIXME
+    devicestate.has_rx_waypoint = false;
+    devicestate.has_rx_text_message = false;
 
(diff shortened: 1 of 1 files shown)


fifieldt (approved):
Makes sense!

PR #4997, opened 2024-10-08
On T114 do no wake into loader instead of application.

### LINKED ISSUE #4651: [Bug]: Heltec T114 does not resume to application after forced shutdown
### Category

Other

### Hardware

Other

### Firmware Version

2.5.0.d6dac17

### Description

A long press and "Shutting down" on the Heltec T110 put it into an approx 18 uA drawing sleep.
A press to wake it resulted in waking to the firmware loading disk emulation rather than the "Resuming...".

### Relevant log output

_No response_

jhps:
In shutdown, on button press, wake back to application rather than into the loader.
Suggested by lyusupov and implemented by todd-herbert.
fixes #4651

### DIFF
--- a/src/platform/nrf52/main-nrf52.cpp
+++ b/src/platform/nrf52/main-nrf52.cpp
@@ -270,6 +270,11 @@ void cpuDeepSleep(uint32_t msecToWake)
         delay(msecToWake);
         NVIC_SystemReset();
     } else {
+        // Resume on user button press
+        // https://github.com/lyusupov/SoftRF/blob/81c519ca75693b696752235d559e881f2e0511ee/software/firmware/source/SoftRF/src/platform/nRF52.cpp#L1738
+        constexpr uint32_t DFU_MAGIC_SKIP = 0x6d;
+        sd_power_gpregret_set(0, DFU_MAGIC_SKIP); // Equivalent NRF_POWER->GPREGRET = DFU_MAGIC_SKIP
+
         // FIXME, use system off mode with ram retention for key state?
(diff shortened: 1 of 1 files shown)


fifieldt (approved):

thebentern (approved):

caveman99:
Just to follow up on this - we clean up stale issues once in a while. No ill is meant by that. Sometimes it brings them back to life, like this one :-)

Monolith18:
I think this is a bug not a feature when it comes to T-Echo not autostarting. Since day 1 (don't remember which FW was that with my first T-Echo) autostart was something normal, out of the box. To put T-Echo unit into DFU mode user can quickly double press reset (RST) button. That's it. I don't know how RAK units work and what is the button configuration, but for T-Echo there was absolutely no problem nor conflict between autostart and using buttons for other functions. Switch on is one press RST button. Double press RST is DFU mode. Long press other button is shutdown.
Applications? Simple one. Unit can operate autonomously. I am not saying even about solar panels or power banks. Imagine T-Echo unit in remote cabin somehwere in the country, but with AC supply from the main grid. T-Echo can be powered with USB cable permanently and when there will be long power outage (remote locations are prone for that) eventually battery will run out of juice. Once power outage is finished and restored unit could start without anyones attention connecting back to the meshtastic grid.
I am not coder, so don't know how to fix that or revert back apart from going back to 2.5.4 FW which is last good one I know, but that means loosing new stuff and fixes from new FWs. If it is possible, I would like to have this feature back for T-Echo units.
Thanks!

PR #5008, opened 2024-10-09
Use `M5STACK_COREBASIC` instead for `M5STACK` to avoid confusion

lboue:
We can use `M5STACK_COREBASIC` instead for `M5STACK` which is a brand and not a model to avoid confusion.
` M5STACK_COREBASIC = 77;`

In fact, it's clear that some users have tried to use the M5Stack variant with a M5Stack Core2 or CoreS3 module.

### DIFF
--- a/src/platform/esp32/architecture.h
+++ b/src/platform/esp32/architecture.h
@@ -102,8 +102,8 @@
 #define HW_VENDOR meshtastic_HardwareModel_PRIVATE_HW
 #elif defined(NANO_G1)
 #define HW_VENDOR meshtastic_HardwareModel_NANO_G1
-#elif defined(M5STACK_CORE)
-#define HW_VENDOR meshtastic_HardwareModel_M5STACK_CORE
+#elif defined(M5STACK_COREBASIC)
+#define HW_VENDOR meshtastic_HardwareModel_M5STACK_COREBASIC
 #elif defined(STATION_G1)
 #define HW_VENDOR meshtastic_HardwareModel_STATION_G1
 #elif defined(DR_DEV)
--- a/src/platform/esp32/architecture.h
+++ b/src/platform/esp32/architecture.h
@@ -102,8 +102,8 @@
 #define HW_VENDOR meshtastic_HardwareModel_PRIVATE_HW
 #elif defined(NANO_G1)
 #define HW_VENDOR meshtastic_HardwareModel_NANO_G1
-#elif defined(M5STACK)
-#define HW_VENDOR meshtastic_HardwareModel_M5STACK
+#elif defined(M5STACK_CORE)
+#define HW_VENDOR meshtastic_HardwareModel_M5STACK_CORE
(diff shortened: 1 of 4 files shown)


variants/m5stack_core/platformio.ini
```diff
@@ -7,7 +7,7 @@ build_src_filter =
 build_flags = 
   ${esp32_base.build_flags} -I variants/m5stack_core
   -DILI9341_DRIVER
-  -DM5STACK
+  -DM5STACK_CORE
   -DUSER_SETUP_LOADED
```
caveman99, line 11: you need to update src/platform/esp32/architecture.h and a few other places too.

caveman99 (changes requested):

variants/m5stack_core/platformio.ini
lboue: It should be OK now.

src/platform/esp32/architecture.h
```diff
@@ -102,8 +102,8 @@
 #define HW_VENDOR meshtastic_HardwareModel_PRIVATE_HW
 #elif defined(NANO_G1)
 #define HW_VENDOR meshtastic_HardwareModel_NANO_G1
-#elif defined(M5STACK)
-#define HW_VENDOR meshtastic_HardwareModel_M5STACK
+#elif defined(M5STACK_CORE)
+#define HW_VENDOR meshtastic_HardwareModel_M5STACK_CORE
 #elif defined(STATION_G1)
```
caveman99, line 107: This, on the other hand, needs a change to the protobufs :-) meshtastic_HardwareModel_M5STACK_CORE is not defined there, meshtastic_HardwareModel_M5STACK is. Since it's not used anywhere else, you can leave the old value

caveman99 (changes requested):

src/platform/esp32/architecture.h
lboue: Yes, this is the protobufs PR: https://github.com/meshtastic/protobufs/pull/598

lboue: I think this is why `M5STACK_COREBASIC` has been defined is protobufs. So we can use `M5STACK_COREBASIC` instead for `M5STACK` which is a brand and not a model to avoid confusion.

caveman99: Let's leave the hardware enum as it is, it's not showing up anywhere in writing. more like a macro for code...

caveman99:
sorry, corebasic is not core. The original support was made for the 2018 M5Stack Core. And this is what constitutes "M5STACK = 42". All other newer devices can be catered for and we can change the macros at will, but please don't discontinue use of slot 42 as this would be a breaking change AND waste. 

Please change this PR in a way that it leaves the existing support functional and adds the new devices.

lboue:
Thanks for the explanation

Issue #5000, opened 2024-10-08
Direct Messages no longer sent over MQTT in JSON


JohnathonMohr:
### Category

Other

### Hardware

Heltec V3

### Firmware Version

2.5.4.8d288d5

### Description

I have my node configured to forward traffic to my own local MQTT server to use with Home Assistant, such as what is described at https://meshtastic.org/docs/software/integrations/mqtt/home-assistant/. Everything was working as expected with the 2.4 Beta firmware, and I had automations set up to notify me of channel and direct messages.

After updating to 2.5 Beta (excited for PKI), I was disappointed to find that DMs were no longer getting forwarded to the JSON topic in MQTT. I have found that traffic such as telemetry, nodeinfo, and even channel messages are still forwarded to MQTT using the settings I already had configured, but DMs are just missing.

I’ve confirmed it isn’t simply a matter of going to a different topic, as I’ve set up a listener for “#” (grabbing everything) and sending DMs, and they do not get sent.

Maybe they show up in protobufs (I don’t have anything setup to be able to deserialize that topic)? I just know it’s missing in JSON.

Talie5in:
@JohnathonMohr  Between two 2.5.x devices with PKC (Green Lock) on both sender/receiver in a Android/iOS App?
If so, that would make sense, as you would no longer have the decryption key (that was originally at the Channel Level for DM's) to be able to decode the packet.

2.4.x <-> 2.5.x where PKC is not been used for DM's should however still be through MQTT.

JohnathonMohr:
Yes, this is between 2 nodes with 2.5.x, using their unique keys (green lock in the app). I originally observed this in the app, but also confirmed same behavior when messaging via the in-device hosted web sites (not sure if you meant specifically using the apps to message, or just that the apps display the green locks).
Is the decryption key not stored in the Meshtastic device? If it's stored in the device (my assumption) what's preventing decryption before sending to MQTT (just like decryption with the channel key was done previously)?

Talie5in:
Do you have encryption enabled in your MQTT config?

https://meshtastic.org/docs/configuration/module/mqtt/#encryption-enabled

But, typically most MQTT based tooling decodes the protobuf and ChannelKey (eg. Default Public LongFast is a known key, so anyone could of before 2.5.x decrypted your DMs or channel messages over MQTT)

JohnathonMohr:
[...]

PR #5012, opened 2024-10-09
Possibly forward PKC DMs over MQTT

jp-bennett:


thebentern (approved):

ianmcorvidae (approved):
This looks like the fix for the spot I was looking at (unless we need to do any additional guards when MQTT encryption is turned off?); haven't dug through completely to be sure it's the only spot. We may need to mark the packet as pki_encrypted (even though we're not sure it is) with how the MQTT code is currently written, since I think it chooses the publish topic and such based on that. Or push similar checks to this into there as well

jp-bennett:
@ianmcorvidae I *think* this logic is already present in MQTT itself.

JohnathonMohr:
@jp-bennett just judging from the title - do you expect this to fix #5000? Or is this unrelated?

jp-bennett:
> @jp-bennett just judging from the title - do you expect this to fix #5000? Or is this unrelated?

I believe this is related, yes. Please let me know if this fixes your issue.

GUVWAF:
@jp-bennett I don't think this will do anything, it would only lead to: 
https://github.com/meshtastic/firmware/blob/ad8747d914e670bf7be43dfe3b58f2fda598c7c1/src/mqtt/MQTT.cpp#L540-L542
Because if it was already decoded and it was PKC, `pki_encrypted` would be set.

PR #5014, opened 2024-10-10
Fix U-Blox detection code.

fifieldt:
Recently there have been reports of intermittent difficulties detecting U-Blox chips. This patch proposes a new approach that should be more reliable.

Previously we were fighting with NMEA messages to try and send binary commands. We unusually also tried changing the Baud rate of U-Blox chips, something we don't do with any other GPS.

It turns out U-Blox has another method to disable NMEA messages. PUBX,40 is a text-based command, supported on all the U-Blox versions we care about that can set the rate of NMEA messages to zero.

This is what we attempt to do with all other GPS and it works quite well.

So this patch alters the probe code to:
1. Remove UBX binary code to stop NMEA messages
2. Remove code that tries to reset UBX chips to 9600 baud
3. Add UBX proprietary text commands messages to stop the NMEA flood
4. Improve log strings sent to the user.

Tested on Ublox 6, Ublox 9, and Ublox 10 on multiple devices.

Also tested on several devices with
non-Ublox GPS to ensure it does not interfere with their detection (heltec-wireless-tracker, wio-tracker-wm11110)

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -1106,6 +1106,11 @@ GnssModel_t GPS::probe(int serialSpeed)
     // Close all NMEA sentences, valid for L76K, ATGM336H (and likely other AT6558 devices)
     _serial_gps->write("$PCAS03,0,0,0,0,0,0,0,0,0,0,,,0,0*02\r\n");
     delay(20);
+    // Close NMEA sequences on Ublox
+    _serial_gps->write("$PUBX,40,GLL,0,0,0,0,0,0*5C\r\n");
+    _serial_gps->write("$PUBX,40,GSV,0,0,0,0,0,0*59\r\n");
+    _serial_gps->write("$PUBX,40,VTG,0,0,0,0,0,0*5E\r\n");
+    delay(20);
 
     // Unicore UFirebirdII Series: UC6580, UM620, UM621, UM670A, UM680A, or UM681A
     PROBE_SIMPLE("UC6580", "$PDTINFO", "UC6580", GNSS_MODEL_UC6580, 500);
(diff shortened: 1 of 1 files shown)


caveman99:
the speed change code, oh yes. That's inside since we foolishly lowered the TX speed to 4800 in some firmware versions to tackle buffer overflows and people had trouble coming back when we reverted that - that was about a year ago ...

src/gps/GPS.cpp
```diff
@@ -1138,35 +1143,10 @@ GnssModel_t GPS::probe(int serialSpeed)
...
-        _serial_gps->write(_message_prt, sizeof(_message_prt));
-        delay(500);
-        serialSpeed = 9600;
-#if defined(ARCH_NRF52) || defined(ARCH_PORTDUINO) || defined(ARCH_STM32WL)
-        _serial_gps->end();
-        _serial_gps->begin(serialSpeed);
-#elif defined(ARCH_RP2040)
-        _serial_gps->end();
-        _serial_gps->setFIFOSize(256);
-        _serial_gps->begin(serialSpeed);
```
caveman99, line 1165: I assume Fifo Size is already 256 here, just reopening?

caveman99 (approved):

src/gps/GPS.cpp
fifieldt: I believe so - checking.

fifieldt: Yes , set at the start of  GPS::probe

fifieldt: (line 1093 or so)

thebentern (approved):

fifieldt:
Also tested on UBlox 7 and it worked there too :)

PR #5023, opened 2024-10-10
Add localhost exception for dontMqttMeBro

jp-bennett:
The OkToMqtt bit is great for blocking unintended traffic uploads to the public MQTT broker. It's also a real pain for users trying to do legitimate local data capture. This patch seeks a middle road, making an allowance for an MQTT server on localhost. Yes this could be abused, but it's less problematic than a popular fork that just strips the bit altogether.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -544,7 +544,7 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp, const meshtastic_MeshPacket &
         }
 
         // check for the lowest bit of the data bitfield set false, and the use of one of the default keys.
-        if (!isFromUs(&mp_decoded) && mp_decoded.decoded.has_bitfield &&
+        if (!isFromUs(&mp_decoded) && strcmp(moduleConfig.mqtt.address, "127.0.0.1") != 0 && mp_decoded.decoded.has_bitfield &&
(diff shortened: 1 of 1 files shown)


fifieldt:
@jp-bennett , what do you think about also adding unroutable IANA private ranges :
* 10.0.0.0/8 IP addresses: 10.0.0.0 – 10.255.255.255
* 172.16.0.0/12 IP addresses: 172.16.0.0 – 172.31.255.255
* 192.168.0.0/16 IP addresses: 192.168.0.0 – 192.168.255.255

jp-bennett:
@fifieldt I like it. I just suspect that logic would take up a decent chunk of flash. It's usually done with tegex, after all.

fifieldt:
@jp-bennett , something like ... ?

<string>moduleConfig.mqtt.address.substr(0,2) == "10."
<string>moduleConfig.mqtt.address.substr(0,4) == "172.1"
<string>moduleConfig.mqtt.address.substr(0,7) == "192.168."

jp-bennett:
> @jp-bennett , something like ... ?
> 
> ```
> <string>moduleConfig.mqtt.address.substr(0,2) == "10."
> <string>moduleConfig.mqtt.address.substr(0,4) == "172.1"
> <string>moduleConfig.mqtt.address.substr(0,7) == "192.168."
> ```

Yeah, that would pretty much do it. I'd probably use strncmp(), but I think the logic works.

PR #5025, opened 2024-10-10
Uplink DMs not to us if MQTT encryption enabled

GUVWAF:
We'll mark the encrypted version of the packet as PKI encrypted when we could not decode it, it's a DM not to us and the hash matches. Then, we don't drop the packet immediately when `mp_decoded` is not actually decoded, but only when `encryption_enabled` is not true. Also we should check whether either the encrypted or decrypted version is marked as PKI encrypted to upload it to the PKI topic.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -573,6 +573,7 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp_encrypted, const meshtastic_Me
             env->packet = (meshtastic_MeshPacket *)&mp_decoded;
             LOG_DEBUG("portnum %i message\n", env->packet->decoded.portnum);
         } else {
+            LOG_DEBUG("nothing, pkt not decrypted\n");
             return; // Don't upload a still-encrypted PKI packet if not encryption_enabled
         }
 
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -524,9 +524,9 @@ void MQTT::publishQueuedMessages()
     }
 }
 
-void MQTT::onSend(const meshtastic_MeshPacket &mp, const meshtastic_MeshPacket &mp_decoded, ChannelIndex chIndex)
+void MQTT::onSend(const meshtastic_MeshPacket &mp_encrypted, const meshtastic_MeshPacket &mp_decoded, ChannelIndex chIndex)
 {
-    if (mp.via_mqtt)
+    if (mp_encrypted.via_mqtt)
         return; // Don't send messages that came from MQTT back into MQTT
     bool uplinkEnabled = false;
     for (int i = 0; i <= 7; i++) {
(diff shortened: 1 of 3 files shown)


src/mqtt/MQTT.cpp
```diff
@@ -559,21 +555,25 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp, const meshtastic_MeshPacket &
...
-            env->packet = (meshtastic_MeshPacket *)&mp;
+            env->packet = (meshtastic_MeshPacket *)&mp_encrypted;
             LOG_DEBUG("encrypted message\n");
-        } else if (mp_decoded.which_payload_variant ==
-                   meshtastic_MeshPacket_decoded_tag) { // Don't upload a still-encrypted PKI packet
+        } else if (mp_decoded.which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
             env->packet = (meshtastic_MeshPacket *)&mp_decoded;
             LOG_DEBUG("portnum %i message\n", env->packet->decoded.portnum);
+        } else {
+            return; // Don't upload a still-encrypted PKI packet if not encryption_enabled
```
ianmcorvidae, line 576: semi-unrelated to the main topic of the PR but I think some sort of `LOG_DEBUG` should go in the third case so we don't print a log line without a newline (bug existed before the change too, I think, but spotting it now).

src/mqtt/MQTT.cpp
```diff
@@ -559,21 +555,25 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp, const meshtastic_MeshPacket &
             return;
         }
     }
-    if (mp.pki_encrypted || ch.settings.uplink_enabled) {
-        const char *channelId = mp.pki_encrypted ? "PKI" : channels.getGlobalId(chIndex);
+    // Either encrypted packet (we couldn't decrypt) is marked as pki_encrypted, or we could decode the PKI encrypted packet
+    bool isPKIEncrypted = mp_encrypted.pki_encrypted || mp_decoded.pki_encrypted;
```
ianmcorvidae, line 559: The comment on line 540 seems inconsistent with the comment down here and I'm not sure which is right -- if we get a PKI-encrypted packet that we _can_ decode (i.e., to us), should it hit the branch above? I guess it looks like it'd fall through both of the branches within the above, so I suppose it probably needn't bother.

ianmcorvidae (approved):
Two thoughts that are potentially just nitpicking, but I think this looks good regardless.

src/mqtt/MQTT.cpp
GUVWAF: I'll add a debug line indeed.

src/mqtt/MQTT.cpp
GUVWAF: Ah, I see what you mean. It's only not decoded when it's PKI encrypted and not to us, will update the wording there.

ianmcorvidae: Ah, I think I see my confusion here, problem on my end. I thought that line was saying "if it was ever PKI encrypted, it will not have been decoded", which isn't true if it was succesfully decoded (as this line shows). Looking again, I guess what's meant is more "if this is a PKI message we can't decrypt, the name 'mp_decoded' is slightly misleading because it will not actually be decoded, therefore we need to guard against that".

So yeah, issue with understanding on my end. Carry on :)

GUVWAF: Haha, yes, that was indeed the intention behind that comment.

thebentern (approved):

PR #5033, opened 2024-10-11
Fix SH1107 - Set Geometry 128x128

markbirss:
USE_SH1107 needs fix to set correct geometry

screen_geometry = GEOMETRY_128_128;


caveman99 (approved):
uh

PR #5040, opened 2024-10-12
Implement rebroadcast mode NONE

### LINKED ISSUE #5030: [Feature Request]: add "none" to rebroadcast mode for all* roles as a configurable option
### Platform

Cross-Platform

### Description

Right now the only role that truly doesnt participate in meshing is client_mute. 

I believe there should be a "no rebroadcast" option added to the drop-down list so that at least sensor and tracker role can use this.

A scenario where you have a room full of powered sensors or a single powered (always awake) weather station would benefit from the higher telemetry priority in the sensor role while still not wanting to contribute to meshing other packets and cause congestion. Currently if there is high channel utilization a client_mute device would stop sending or scale telemetry. This would mean a weather station would never be able to pass on data to the rest of the busy mesh. 

The same could can be said for always powered (car?) trackers that prioritize position data but we still dont want to participate in meshing. 

Client_Hidden_**mute** would also be configurable if this option was added.

thebentern:
Closes #5030

### DIFF
--- a/src/mesh/FloodingRouter.cpp
+++ b/src/mesh/FloodingRouter.cpp
@@ -35,6 +35,12 @@ bool FloodingRouter::shouldFilterReceived(const meshtastic_MeshPacket *p)
     return Router::shouldFilterReceived(p);
 }
 
+bool FloodingRouter::isRebroadcaster()
+{
+    return config.device.role != meshtastic_Config_DeviceConfig_Role_CLIENT_MUTE &&
+           config.device.rebroadcast_mode != meshtastic_Config_DeviceConfig_RebroadcastMode_NONE;
+}
+
 void FloodingRouter::sniffReceived(const meshtastic_MeshPacket *p, const meshtastic_Routing *c)
 {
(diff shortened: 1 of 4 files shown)


src/mesh/FloodingRouter.cpp
```diff
@@ -45,7 +51,7 @@ void FloodingRouter::sniffReceived(const meshtastic_MeshPacket *p, const meshtas
     }
     if (!isToUs(p) && (p->hop_limit > 0) && !isFromUs(p)) {
         if (p->id != 0) {
-            if (config.device.role != meshtastic_Config_DeviceConfig_Role_CLIENT_MUTE) {
+            if (isRebroadcaster()) {
```
GUVWAF, line 54: Maybe we should change the debug log on line 65 as well.

GUVWAF (approved):

src/mesh/FloodingRouter.cpp
thebentern: Good call. Pushed

GUVWAF (approved):

PR #5046, opened 2024-10-13
DataRegion struct

lboue:
DataRegion struct from #5043 PR
Ralated to PR #5043.

This structure is used for on-screen display.


fifieldt:
Hi! Thank you as always for your work. I don't quite have enough context to understand the aim here. Would you be able to write something up, or join us in the discord to chat for a bit?

PR #5050, opened 2024-10-13
fix compile error for BMA423

andyshinn:
Minor typo preventing BMA423 compiling.


thebentern:
This spelling fix actually breaks compilation for the version we have pinned

PR #5052, opened 2024-10-13
Ws85 updates : set want_ack, high_priority, add temperature.

tavdog:
For more reliable data transmission, set want_ack and priority_max in the packet.  Wind data only gets sent every 5 minutes and any node with ws85 enabled will most likely be a remote weather node with no other purpose than to send out the wind telemetry. 

Also add parse temperature from the WS85 and add it to the telemetry packet.

### DIFF
--- a/src/modules/SerialModule.cpp
+++ b/src/modules/SerialModule.cpp
@@ -252,8 +252,12 @@ void SerialModule::sendTelemetry(meshtastic_Telemetry m)
         pb_encode_to_bytes(p->decoded.payload.bytes, sizeof(p->decoded.payload.bytes), &meshtastic_Telemetry_msg, &m);
     p->to = NODENUM_BROADCAST;
     p->decoded.want_response = false;
-    p->want_ack = true;
-    p->priority = meshtastic_MeshPacket_Priority_HIGH;
+    if (config.device.role == meshtastic_Config_DeviceConfig_Role_SENSOR) {
+        p->want_ack = true;
+        p->priority = meshtastic_MeshPacket_Priority_HIGH;
+    } else {
+        p->priority = meshtastic_MeshPacket_Priority_RELIABLE;
+    }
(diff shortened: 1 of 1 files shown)


src/modules/SerialModule.cpp
```diff
@@ -252,7 +252,8 @@ void SerialModule::sendTelemetry(meshtastic_Telemetry m)
         pb_encode_to_bytes(p->decoded.payload.bytes, sizeof(p->decoded.payload.bytes), &meshtastic_Telemetry_msg, &m);
     p->to = NODENUM_BROADCAST;
     p->decoded.want_response = false;
-    p->priority = meshtastic_MeshPacket_Priority_RELIABLE;
+    p->want_ack = true;
+    p->priority = meshtastic_MeshPacket_Priority_MAX;
```
GUVWAF, line 256: Please don't set it higher than `ACK`, because those need to get priority always to avoid excessive retransmissions and flooding. You could use `meshtastic_MeshPacket_Priority_HIGH`.

caveman99: also what about framing this change with device.role.sensor? because that would be the appropriate role for this kind of packets.

tavdog: I changed to high from max.  The sensor role could work but the ws85 sensor node can't go to sleep or it'll miss the serial wind data that comes in every 2 seconds.  Sensor role could muck that up i think.

thebentern: Sensor role should only sleep if you have power.is_power_saving enabled

tavdog: > also what about framing this change with device.role.sensor? because that would be the appropriate role for this kind of packets.
> 
> E.G. use 'reliable' and no want_ack from all other roles...

are you proposing that I check for sensor_role here and only set want_ack and high priority if sensor_role is set ?

thebentern (approved):

PR #5063, opened 2024-10-14
UserPrefs - Added Configurable Hops

medentem:
- Added configurable hops to UserPrefs


src/mesh/NodeDB.cpp
```diff
@@ -308,7 +308,11 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
 #else
     config.lora.modem_preset = meshtastic_Config_LoRaConfig_ModemPreset_LONG_FAST;
 #endif
+#ifdef USERPREFS_CONFIG_LORA_HOPS
+    config.lora.hop_limit = USERPREFS_CONFIG_LORA_HOPS;
+#else
```
caveman99, line 313: Given the history we've had with this setting, i am strongly not a fan of making it bakeable into firmware.

medentem: It's use case dependent, and is currently manipulated via the`USERPREFS_EVENT_MODE` user prefs for valid reasons. What is the specific concern?

medentem:
Closing to rebase and clean up PR.

src/mesh/NodeDB.cpp
caveman99: People giving out compiled firmware with this set to "7" and marketing it as "meshtastic range boosted special hacker version"

thebentern: I'm not in favor of this either. We're working on making an interface to allow users to custom define userPrefs for custom builds. Allowing users to bake in 7 hop default firmware is like handing out the plans on how to build a mesh killing gun.

PR #5064, opened 2024-10-14
remove want_response flag from packets going to the phoneAPI

caveman99:
This fixes position sometimes not updating in the phone apps.

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -310,6 +310,12 @@ void MeshService::sendToPhone(meshtastic_MeshPacket *p)
         }
     }
 
+    // in case we have a position packet, the phone does not like the want_response flag. Remove it here.
+    // NB: sendToPhone is always operating on a copy, so it is ok to modify it here.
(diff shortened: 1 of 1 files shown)


andrekir:
if the `want_response` is intended, I think we should patch this on the app side instead. otherwise clients can't tell `position requests` from regular position broadcasts.

the reason to filter out `position requests` is all clients except iOS send an empty position payload, and if we use that data it clears the node's last known good position.

caveman99:
The want_response is intended, it's a deliberate query for other nodes to send their positions. It's triggered by a multi-button-press, a hotkey from the canned messages keyboard or the api call 'request position'. Also the first position broadcast after booting a node will have the flag set, and this particular isntance is what is failing to update the app display.

andrekir:
thanks for the confirmation. I prepared a fix on the app side: https://github.com/meshtastic/Meshtastic-Android/commit/af56fff2162d4f13a1fc70719750fed3111729b4

 this issue will be resolved in the next app release.

PR #5065, opened 2024-10-14
UserPrefs - Added # of hops to userprefs

medentem:
Quite simply, added # of hops to userprefs.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -308,7 +308,11 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
 #else
     config.lora.modem_preset = meshtastic_Config_LoRaConfig_ModemPreset_LONG_FAST;
 #endif
+#ifdef USERPREFS_CONFIG_LORA_HOPS
+    config.lora.hop_limit = USERPREFS_CONFIG_LORA_HOPS;
+#else
     config.lora.hop_limit = HOP_RELIABLE;
+#endif
 #ifdef USERPREFS_CONFIG_LORA_IGNORE_MQTT
     config.lora.ignore_mqtt = USERPREFS_CONFIG_LORA_IGNORE_MQTT;
 #else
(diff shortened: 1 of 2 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -306,7 +306,11 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
 #else
     config.lora.modem_preset = meshtastic_Config_LoRaConfig_ModemPreset_LONG_FAST;
 #endif
+#ifdef USERPREFS_CONFIG_LORA_HOPS
+    config.lora.hop_limit = USERPREFS_CONFIG_LORA_HOPS;
+#else
```
caveman99, line 311: Given the history we've had with this setting, i am strongly not a fan of making it bakeable into firmware. This will immediately lead to People giving out compiled firmware with this set to "7" and marketing it as "meshtastic range boosted special hacker version"

medentem: Thanks for commenting again. The other PR was a mess. To provide a rebuttal, I think this is use case dependent. There are valid use cases for changing this value for a group of people. Example: events, rural settings with higher hop needs due to terrain and network topology.

Also, there is nothing to stop someone from maintaining a fork and doing this anyway to get a distribution out there.

caveman99: I just noticed every time someone touches this seting it ended in disaster, because many don't know the implications. I am not the judge for this PR; just voicing my concerns.

thebentern: I'm not in favor either. We're working on making an interface to allow users to custom define userPrefs for custom builds. Allowing users to bake in 7 hop default firmware is like handing out the plans on how to build a mesh killing gun.

medentem: @rcarteraz I didn't mean that as a threat at all. Apologies if it came off that way. My point is only that if someone is skilled enough to clone the repo, build and flash the firmware, they are very likely also knowledgeable to fork it. My argument is that keeping hops out of userprefs is not a significant deterrent. Furthermore, this is simply to set a default value in a custom firmware package. The precedent has already been set that this may be appropriate for some circumstances - example: it was done for the meshtastic firmware for defcon.

PR #5072, opened 2024-10-15
Add MQTT exception for private IP address server

JohnathonMohr:
Related to issue #5000 and implementing what was suggested in #5023.

Allowing MQTT publishing regardless of the flag setting if the server being published to is local (for local automation scenarios).

I have not completed testing yet, but wanted to get early feedback.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -556,7 +556,8 @@ void NodeDB::installDefaultChannels()
 
 void NodeDB::resetNodes()
 {
-    clearLocalPosition();
+    if (!config.position.fixed_position)
+        clearLocalPosition();
     numMeshNodes = 1;
     std::fill(devicestate.node_db_lite.begin() + 1, devicestate.node_db_lite.end(), meshtastic_NodeInfoLite());
     devicestate.has_rx_text_message = false;
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -35,6 +35,8 @@ Allocator<meshtastic_ServiceEnvelope> &mqttPool = staticMqttPool;
 // FIXME - this size calculation is super sloppy, but it will go away once we dynamically alloc meshpackets
 static uint8_t bytes[meshtastic_MqttClientProxyMessage_size + 30]; // 12 for channel name and 16 for nodeid
 
+static bool isMqttServerAddressPrivate = false;
+
 void MQTT::mqttCallback(char *topic, byte *payload, unsigned int length)
 {
     mqtt->onReceive(topic, payload, length);
@@ -236,6 +238,11 @@ MQTT::MQTT() : concurrency::OSThread("mqtt"), mqttQueue(MAX_MQTT_QUEUE)
                                                                      default_map_position_precision);
             map_publish_interval_msecs = Default::getConfiguredOrDefaultMs(
                 moduleConfig.mqtt.map_report_settings.publish_interval_secs, default_map_publish_interval_secs);
+        }
+
+        isMqttServerAddressPrivate = isPrivateIpAddress(moduleConfig.mqtt.address);
+        if (isMqttServerAddressPrivate) {
+            LOG_INFO("MQTT server is a private IP address.");
         }
 
 #if HAS_NETWORKING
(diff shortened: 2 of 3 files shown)


fifieldt (approved):
Thanks for working on this and congratulations on your first patch to Meshtastic :)

jp-bennett (changes requested):
Does this catch something sneaky like 10.mydomain.com? I started working on this a couple days ago, and went down that rabbit hole. I think we should check that we only have dots and decimals in the address.

Talie5in:
@jp-bennett and @JohnathonMohr 

Might be able to do that with some regex to make sure it's an IP firsT?

`#include <regex>

bool MQTT::isValidIpAddress(const char ip[])
{
    // Use a simple regular expression to check for valid IPv4 addresses.
    const std::regex ipPattern("^([0-9]{1,3}\\.){3}[0-9]{1,3}$");
    return std::regex_match(ip, ipPattern);
}

bool MQTT::isPrivateIpAddress(const char ip[]) 
{
    // First, check if it's a valid IPv4 address.
    if (!isValidIpAddress(ip)) {
        return false;
    }

    // Check for common private IP address ranges.
    if (strcmp(ip, "127.0.0.1") == 0 || 
        strncmp(ip, "10.", 3) == 0 || 
        strncmp(ip, "192.168", 7) == 0) {
        return true;
    }

    // Early exit if it's not a 172 address.
    if (strncmp(ip, "172.", 4) != 0) {
        return false;
    }

    // Ensure the format is correct (checking that the IP has at least two octets and no buffer overflows).
    if (strlen(ip) < 7 || ip[6] != '.') {
        return false;
    }

    // Parse the second octet.
    char octet2[3] = { ip[4], ip[5], '\0' };
    int octet2Num = atoi(octet2);

    // Return true if the second octet is within the private range for 172.x.x.x.
    return octet2Num >= 16 && octet2Num <= 31;
}`

jp-bennett:
> Might be able to do that with some regex to make sure it's an IP first?

I can only imagine how much flash space adding regex to the firmware is going to cost. We only need to check once per boot. Just make a char array of "0123456789." and a nested for loop.

Talie5in:
@jp-bennett Yup, didnt think of that, how about cstring as this is already loaded in many places in the firmware?

```
#include <cstring> // for strlen

bool MQTT::isValidIpAddress(const char ip[])
{
    int numDots = 0;
    int length = strlen(ip);  // Get the length of the string

    // IPv4 addresses should be between 7 and 15 characters (e.g., "0.0.0.0" to "255.255.255.255").
    if (length < 7 || length > 15) {
        return false;
    }

    const char *ptr = ip;

    // Check if the string is empty or null-terminated properly.
    if (*ptr == '\0') {
        return false;
    }

    while (*ptr) {
        int octet = 0;
        int digitCount = 0;

        // Parse each octet (up to three digits).
        while (*ptr >= '0' && *ptr <= '9') {
            octet = octet * 10 + (*ptr - '0');
            ptr++;
            digitCount++;

            // Ensure no octet exceeds 255 or is longer than three digits.
            if (octet > 255 || digitCount > 3) {
                return false;
            }
        }

        // After each octet, we expect either a dot (.) or the end of the string.
        if (*ptr == '.') {
            numDots++;
            ptr++;  // Move past the dot.
            if (numDots > 3) {
                return false;  // Too many dots.
            }
        } else if (*ptr != '\0') {
            return false;  // Invalid character found.
        }
    }

    // We should have exactly 3 dots and 4 octets (number of dots = 3).
    return numDots == 3;
}
```

JohnathonMohr:
Good catch on decimal prefixes of domains. I'll include the check in the next iteration (avoiding regex).

fifieldt:
Thanks for the additional review folks.

JohnathonMohr:
@Talie5in, it's great to see your idea of confirming a valid IPv4, but for the simple purpose of determining whether to consider the server as a private address or not, I think it's more than required.

As long as we can determine it's *not* a valid domain (i.e. it has only digits and dots) I think that's sufficient. If we determine it's a private address after that, *even if it's not a valid address*, I don't think it matters since a connection will never be able to be made anyway.

Talie5in:
@JohnathonMohr All goodly!

PR #5081, opened 2024-10-16
Ensure the MQTT address is an IPv4 before determining it's private

JohnathonMohr:
Continuing from #5072, ensuring no sneaking subdomains make an address appear like a private IPv4.

This is also now tested. :)

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -706,9 +706,15 @@ bool MQTT::isValidJsonEnvelope(JSONObject &json)
 
 bool MQTT::isPrivateIpAddress(const char address[])
 {
+    // Min. length like 10.0.0.0, max like 192.168.255.255
+    size_t length = strlen(address);
+    if (length < 8 || length > 15) {
+        return false;
+    }
+
     // Ensure the address contains only digits and dots.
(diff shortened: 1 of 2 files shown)


JohnathonMohr:
@jp-bennett @fifieldt Next iteration is here. :)

fifieldt (approved):

fifieldt (commented):
Hey, is it worth checking the length?

jp-bennett (approved):

JohnathonMohr:
> Hey, is it worth checking the length?

We could. I opted not to initially for the same reason of just checking for digits and dots: even if it's not valid, a connection won't be made.

It's easy enough though, so I can if you want.

JohnathonMohr:
Ok, length check added. 😅

fifieldt (approved):

thebentern (approved):

CamFlyerCH:
I have to add a non-standard port number (like 192.168.1.84:2883) and the detection is not detecting it as a private IP anymore.

JohnathonMohr:
> I have to add a non-standard port number (like 192.168.1.84:2883) and the detection is not detecting it as a private IP anymore.

@CamFlyerCH yep, I realized that earlier. The fix is in #5084.

PR #5084, opened 2024-10-16
Account for port number in MQTT server

JohnathonMohr:
The port number might also be specified in the MQTT server address. If it is, we can’t let that fail validation of a private address.

Also adding some more format validation.

~~**I still need to test this, but want to make this gap known early.**~~
✅ Tested.


JohnathonMohr:
@jp-bennett @fifieldt FYI

thebentern (approved):

fifieldt (approved):
Argh, thanks for this!

JohnathonMohr:
Any ideas why the build suddenly failed? Is it transient?

fifieldt:
Possibly. Re-running.

PR #5086, opened 2024-10-17
Coerce minimum telemetry interval of 30 minutes on defaults and make new default interval one hour

thebentern:
Starting to see Telemetry traffic quickly become the predominant generator of traffic on the public mesh. This is _another_ attempt to back some of that off to make more headroom for text messaging, node discovery, and position updates, as well as prevent inadvertent setting of these values to spammy intervals. I don't need to see everybody's battery level updated every 5 minutes. ;-)

The new standard would be:
1. No more frequent than 30-minute broadcasts of any telemetry on the default channel
2. New default of one-hour (up from 30 minutes) for regular interval broadcasts of telemetry.

### DIFF
--- a/src/mesh/Default.cpp
+++ b/src/mesh/Default.cpp
@@ -43,6 +43,15 @@ uint32_t Default::getConfiguredOrDefaultMsScaled(uint32_t configured, uint32_t d
     return getConfiguredOrDefaultMs(configured, defaultValue) * congestionScalingCoefficient(numOnlineNodes);
 }
 
+uint32_t Default::getConfiguredOrMinimumValue(uint32_t configured, uint32_t minValue)
+{
+    // If zero, intervals should be coalesced later by getConfiguredOrDefault... methods
+    if (configured == 0)
+        return configured;
+
+    return configured < minValue ? minValue : configured;
+}
+
 uint8_t Default::getConfiguredOrDefaultHopLimit(uint8_t configured)
 {
 #if USERPREFS_EVENT_MODE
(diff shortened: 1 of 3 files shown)


andrekir:
something to consider in the future is coalescing the Telemetry oneOfs into a single broadcast. the 16 byte header makes splitting up in separate small packets very inneficient.

GUVWAF (approved):
LGTM, but maybe after this we should consider changing the "online" status of nodes in the Android/iOS app from 15 min. to [2 hours](https://github.com/meshtastic/firmware/blob/fbb6778415dac8d6e7e33d8c79510f77ecdb42fa/src/mesh/NodeDB.cpp#L953), which is what the firmware uses. DeviceTelemetry might be the only packet that keeps a node "online", so it means they will switch between online and offline all the time now.

andrekir:
good point! both Android and iOS are using 2 hours for online nodes, so we're all set there.

GUVWAF:
@andrekir Ah, looks like iOS is indeed using two hours already, but Android still uses 15min., right? https://github.com/meshtastic/Meshtastic-Android/blob/700d8d3460eff858c216c4eef5e9d254fcfd5811/app/src/main/java/com/geeksville/mesh/NodeInfo.kt#L189

andrekir:
@GUVWAF `NodeInfo` is now exclusive to the API and was replaced with [NodeEntity](https://github.com/meshtastic/Meshtastic-Android/blob/master/app/src/main/java/com/geeksville/mesh/database/entity/NodeEntity.kt#L189).

Technologyman00:
Is there anyway to get around the limit? This killed my weather station as it can no longer update my private MQTT server fast enough

jp-bennett:
> Is there anyway to get around the limit? This killed my weather station as it can no longer update my private MQTT server fast enough
> 

Internally we've talked about the potential for use cases outside the normal defaults, and so far we've settled on encouraging people to compile the source themselves.

Alternatively, we could consider allowing faster updates when not on the default channel.

PR #5087, opened 2024-10-17
Initial NODENUM_BROADCAST_NO_LORA implementation with NeighborInfo module

### LINKED ISSUE #5082: [Enhancement]: Limit neighbor info module to no-lora transport (MQTT)
### Platform

Cross-Platform

### Description

The current state of Neighbor info module:

1. Serves no real world use outside of topology visualizations in third party maps
2. Not supported in any of the official apps or clients
3. Can easily create a significant amount of traffic with fairly large packets on larger meshes

thebentern:
Closes #5082


GUVWAF:
Maybe we could add an `isDM()` method that checks for `to != NODENUM_BROADCAST && to != NODENUM_BROADCAST_NO_LORA`?

GUVWAF:
I think when this is applied, we can also relax the default and minimum broadcast intervals again? E.g. 30 min. and 15 min., respectively, would be OK if it's only going to MQTT and the PhoneAPI.

garthvh:
I don't think we should relax anything, this thing is still traffic monster, these packets will pop back up over mqtt.  Why save this?

Issue #5089, opened 2024-10-17
[Feature Request]: Add Feedback for Screenless devices for enabling/disabling the GPS


Technologyman00:
### Platform

NRF52

### Description

The SeeedStudio T1000-E doesn't have a screen but has a buzzer and light, so it would be really nice to get some feedback when the button is triple clicked to enable/disable the GPS. I do this frequently and without feedback it is hard to know if it worked or not.

PR #5090, opened 2024-10-18
Add buzzer feedback on GPS toggle

Technologyman00:
Addresses #5089. Screen-less devices don't have any feedback for a triple click of the button so it maybe difficult to ensure the GPS was disabled

### DIFF
--- a/src/ButtonThread.cpp
+++ b/src/ButtonThread.cpp
@@ -171,6 +171,7 @@ int32_t ButtonThread::runOnce()
                     gps->toggleGpsMode();
                     if (screen)
                         screen->forceDisplay(true); // Force a new UI frame, then force an EInk update
+                    playBeep();
                 }
                 break;
 #endif
--- a/src/ButtonThread.cpp
+++ b/src/ButtonThread.cpp
@@ -169,8 +169,9 @@ int32_t ButtonThread::runOnce()
             case 3:
                 if (!config.device.disable_triple_click && (gps != nullptr)) {
                     gps->toggleGpsMode();
-                    if (screen)
+                    if (screen) {
                         screen->forceDisplay(true); // Force a new UI frame, then force an EInk update
+                    }
                     playBeep();
                 }
                 break;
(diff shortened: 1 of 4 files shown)


fifieldt:
This is a great idea, thank you!

Technologyman00:
@fifieldt Thank you! After playing around with it, it probably should have some audio cue for on vs off, but I am not a musician.

fifieldt:
Maybe since the disable sequence is a triple push, the associated buzzer should be a triple sound?

Technologyman00:
Triple click enables and disables. I will play around and see

b8b8:
triple low sound after disabling, triple high sound after enabling?

Technologyman00:
The most recent commit sounds like this: https://youtu.be/VHLxfraEB0k?si=R2N3lvpfyCn2dF4E

Technologyman00:
Don't merge yet the noise is backwards according to what the app reports

fifieldt:
Thanks so much for providing the video - helps a lot as I haven't time to flash and test. I think the sound is good enough to ship - we can always change later if needed :)

@Technologyman00 - are you familiar with the trunk linting/formatting tool? We use it to keep out files formatted the same. It looks like it needs to be run on the files from this patch. Could you do that? If not , let us know and someone will help. Other than that, good to go.

fifieldt:
We should also make sure this works OK on devices that don't have a buzzer.

Technologyman00:
Ok I just tested it on my T114 V1 and it works. I don't have a different one to test.

fifieldt:
OK, looks like buzzer.h is included by main and not guarded by PIN_BUZZER so will have no impact on devices without buzzers

fifieldt (changes requested):
Good to go apart from trunk format.

Technologyman00:
Someone else can do it or tell me why its been stuck doing this for the past while 
image

fifieldt (approved):
Let's not make trunk take the fun out of your first patch.

Nestpebble:
@Technologyman00 re: https://github.com/meshtastic/firmware/pull/5090#issuecomment-2421210770

Trunk doesn't work on windows for some reason. There's workarounds, either making it work in Windows by placing the files manually or running it in WSL, but I found the quickest way for a casual update was gitpod. YMMV.

Love this idea and implementation, btw. I have lots of screenless GPS devices, and this is really cool.

PR #5093, opened 2024-10-18
Add `-p` flag

madeofstown:
Add the `-p` flag to the `mkdir` command so it doesn't fail when the folder already exists.


fifieldt:
Hi, and welcome to Meshtastic :)

Are there situations where /etc doesn't exist? WSL?

madeofstown:
> Hi, and welcome to Meshtastic :)
> 
> Are there situations where /etc doesn't exist? WSL?

If there are nobody has submitted an issue about it yet.

PR #5098, opened 2024-10-19
add the non routable portions of the 44Net/HamNet/AMPRNet IP blocks to the private address comparison

rbrtio:
This PR adds the 44Net aka HamNet IP space to the private IP address check which enables the exchange of data via MQTT.

This IP address space is commonly used by Ham Radio operators for data communications.

https://www.ardc.net/44net/

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -750,6 +750,11 @@ bool MQTT::isPrivateIpAddress(const char address[])
         return true;
     }
 
+    // Check to see if it is a HamNet address which we also consider private.
+    if (strncmp(address, "44.", 3) == 0) {
+        return true;
+    }
+
     // See if it's definitely not a 172 address.
     if (strncmp(address, "172", 3) != 0) {
         return false;


fifieldt:
Hi @rbrtio , thanks for your first patch to Meshtastic!

Could you provide a bit more information on how you're using 44Range for MQTT servers?

As I understand it, 44Range is Internet routable, so it might be a bit different than the other ranges that have been added to the code here. The original intention of the exceptions to the MQTT opt-out were for "local" use.

rbrtio:
Thank you for your review @fifieldt 

Although some of the 44Net address space is Internet routable, the vast majority of it is utilized by Ham Radio operators to connect resource located on private networks.

Ham Radio operators can request an IP address allocation located within this network block to connect resources used for projects within the hobby.  The operator must identify ahead of time if they plan to expose their allocation via BGP to the Internet, but this is not commonplace and is primarily concentrated in 3 /16 address blocks.

I hope this helps explain the use of the address space.

I have identified a change I need to make as it is no longer the entire 44.0.0.0/8 network that is part of 44Net, but only the 44.0.0.0/9 and 44.128.0.0/10 blocks which would encompass the addresses from 44.0.0.1 - 44.191.255.255 so I will need to follow the pattern created by the parsing of the 172 address space.

Thanks again for your feedback and please let me know how you would like to proceed.  In the meanwhile I will prepare the changes required to allow only the /9 and /10 blocks noted above.

jp-bennett:
Seems like this should then be gated for unencrypted HAM packets only, if we proceed. Adding a general exception for routable IPs seems like a bad idea IMO.

rbrtio:
@jp-bennett please note that I identified the routable addresses are concentrated in 3 /16 network blocks, which can be excluded, that leaves 189 /16 network blocks that are not routable.

This change doesn't impact me, I was just trying to be proactive.  I can close the PR if there is no current need for the change.

jp-bennett:
@rbrtio that handles my primary concern, then.

fmoessbauer:
Please note, that this address space does not solely belong to AMPR net. As of the latest change in 2019, only the following blocks are AMPR:

- `44.0.0.0/9` for USA subnets
- `44.128.0.0/10` for others

rbrtio:
@fmoessbauer thank you, I highlighted that in one of my above comments, I will update the title of the PR to eliminate any confusion.

caveman99:
i am -1 on this. there's a RFC on private address space. Clearly defined as 10.* 172.something.* and 192.168.* - you can include the 169.254 ad-hoc space if you want, that's blocked 'forever' too.

Nothing prevents 44.* from being re-allocated for non-HAM use if the need arises. there's no law blocking that. 1.1.1.1 has a history i don't wanna repeat here.

rbrtio:
It sounds like this in general this change is frowned upon so I will withdrawl this PR and submit a different PR to add the IPIPA address space identified in RFC 3927 that @caveman99 identified.

PR #5101, opened 2024-10-20
Move 115200 baud GNSS probe earlier

thebentern:
Since most GPS modules are defaulted to 9600 or 115200, moving the 115200 probe earlier should statistically speed up detection time and avoid wasted cycles on irrelevant baud rates


fifieldt:
Move 38400 before 4800, for ublox

fifieldt (approved):
ship it

jp-bennett (approved):

PR #5102, opened 2024-10-20
Optimise GPS Baud Rate cycle

fifieldt:
Previously, our baud rate cycled through one list twice.

There were some rarer baudrates in there, so this code separates out those into a dedicated list that is only run through if detection fails for common bauds. We also only run through each baud rate once.


fifieldt:
This is currently untested, please don't merge until we test :)

fifieldt:
Beginning testing.

fifieldt:
* T1000-E (AG3335) OK (and speed much improved)

fifieldt:
OK, tested with 7 different GPSes across several devices and this works nicely. Ready for review.

src/gps/GPS.cpp
```diff
@@ -937,10 +950,10 @@ void GPS::down()
...
 
         if (softsleepSupported) {
-            // How long does gps_update_interval need to be, for GPS_HARDSLEEP to become more efficient than GPS_SOFTSLEEP?
-            // Heuristic equation. A compromise manually fitted to power observations from U-blox NEO-6M and M10050
-            // https://www.desmos.com/calculator/6gvjghoumr
-            // This is not particularly accurate, but probably an impromevement over a single, fixed threshold
+            // How long does gps_update_interval need to be, for GPS_HARDSLEEP to become more efficient than
+            // GPS_SOFTSLEEP? Heuristic equation. A compromise manually fitted to power observations from U-blox NEO-6M
+            // and M10050 https://www.desmos.com/calculator/6gvjghoumr This is not particularly accurate, but probably an
+            // impromevement over a single, fixed threshold
```
Technologyman00, line 956: impromevement should be improvement

src/gps/GPS.cpp
```diff
@@ -413,30 +413,42 @@ int GPS::getACK(uint8_t *buffer, uint16_t size, uint8_t requestedClass, uint8_t
     return 0;
 }
 
+/**
+ * @brief  Setup the GPS based on the model detected.
+ *  We detect the GPS by cyling through a set of baud rates, first common then rare.
```
Technologyman00, line 418: cyling should be cycling

src/gps/GPS.h
```diff
@@ -76,13 +76,20 @@ class GPS : private concurrency::OSThread
...
 #endif
   private:
-    const int serialSpeeds[6] = {9600, 115200, 38400, 4800, 57600, GPS_BAUDRATE};
+#if GPS_BAUDRATE_FIXED
+    // if GPS_BAUDRATE is specified in variant, only try that.
+    const int serialSpeeds[1] = {GPS_BAUDRATE};
+#else
+    const int serialSpeeds[3] = {9600, 115200, 38400};
+#endif
+    const int rareSerialSpeeds[3] = {4800, 57600, GPS_BAUDRATE};
```
thebentern, line 85: I think rare serial speeds should be inside of the else

fifieldt: would fail to build due to the reference in gps.cpp

fifieldt: I can modify the array though.

thebentern:
@fifieldt do we want to apply the GPS_BAUDRATE_FIXED macro to variants as a part of this PR as well?

fifieldt:
> @fifieldt do we want to apply the GPS_BAUDRATE_FIXED macro to variants as a part of this PR as well?

I don't think there's a need set GPS_BAUDRATE_FIXED directly in variant.h? Just adds work for hardware makers when setting GPS_BAUDRATE does the job. Or are you saying there are other variants to set baud rates for?

thebentern:
> I don't think there's a need set GPS_BAUDRATE_FIXED directly in variant.h? Just adds work for hardware makers when setting GPS_BAUDRATE does the job. Or are you saying there are other variants to set baud rates for?

I'm being a bonehead and the #ifdef GPS_BAUDRATE block in configuration.h sailed over my head.

thebentern (approved):

PR #5115, opened 2024-10-21
add RFC 3927 IP address space to private IP checks

rbrtio:
Per the comments of #5098 add the RFC 3927 IP address block (169.254.0.0/16), also referred to as IPIPA, to the private address checks for MQTT functionality.

[RFC3927 Reference](https://datatracker.ietf.org/doc/html/rfc3927)


fifieldt (approved):
Thanks for this, appreciate your dedication to contributing even in the face of negative feedback in your first patch!

PR #5118, opened 2024-10-21
Update meshtasticd.service

yNosGR:
Adding restart on service failure with 3 seconds between restart to stop fast restart loops.
Adding StartLimitBurst and StartLimitInterval to limit it to 5 restarts in 200 seconds.

### DIFF
--- a/bin/meshtasticd.service
+++ b/bin/meshtasticd.service
@@ -1,12 +1,16 @@
 [Unit]
 Description=Meshtastic Native Daemon
 After=network-online.target
+StartLimitInterval=200
+StartLimitBurst=5
 
 [Service]
 User=root
 Group=root
 Type=simple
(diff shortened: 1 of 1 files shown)


bin/meshtasticd.service
```diff
@@ -1,12 +1,16 @@
...
 After=network-online.target
+StartLimitInterval=200
+StartLimitBurst=5
 
 [Service]
 User=root
 Group=root
 Type=simple
 ExecStart=/usr/sbin/meshtasticd
+Restart=always
```
thebentern, line 12: Perhaps on-failure makes more sense than always?

jp-bennett: That would need an audit of whether we properly use return codes for all of our exit conditions.

jp-bennett (approved):

PR #5154, opened 2024-10-26
dummy for config transfer

caveman99:


mverch67:
fyi: related to device_ui.proto

PR #5155, opened 2024-10-26
Icarus - Custom PCB

Babyyoda777:
Hi,
Hopefully I have sufficiently fixed the issues from before :) Changed it to private hardware and removed my protobufs.

### DIFF
--- a/variants/icarus/platformio.ini
+++ b/variants/icarus/platformio.ini
@@ -1,11 +1,12 @@
 [env:icarus]
 extends = esp32s3_base
 board = icarus
+board_level = extra
 board_check = true
 board_build.mcu = esp32s3
 upload_protocol = esptool
 upload_speed = 921600
-platform_packages = framework-arduinoespressif32@https://github.com/PowerFeather/powerfeather-meshtastic-arduino-lib/releases/download/2.0.16a/esp32-2.0.16.zip
+platform_packages = framework-arduinoespressif32@https://github.com/PowerFeather/powerfeather-meshtastic-arduino-lib/releases/download/2.0.16b/esp32-2.0.16.zip
 lib_deps =
   ${esp32s3_base.lib_deps}
 build_unflags =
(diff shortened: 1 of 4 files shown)


variants/icarus/variant.h
```diff
@@ -0,0 +1,30 @@
+// Icarus has a 1.3 inch OLED Screen
+#define USCREEN_SSD106
```
caveman99, line 2: That looks like a typo. you sure?

variants/icarus/platformio.ini
```diff
@@ -0,0 +1,19 @@
+[env:icarus]
+extends = esp32s3_base
+board = icarus
+board_level = extra
+board_check = true
+board_build.mcu = esp32s3
+upload_protocol = esptool
+upload_speed = 921600
+platform_packages = framework-arduinoespressif32@https://github.com/PowerFeather/powerfeather-meshtastic-arduino-lib/releases/download/2.0.16b/esp32-2.0.16.zip
```
caveman99, line 9: Is there something special about that framework? just curious.

caveman99 (approved):

variants/icarus/platformio.ini
Babyyoda777: hi, yes it adds improved power management for esp32s

variants/icarus/variant.h
Babyyoda777: it was in the seeedstudio esp32s3 variant.h and im using a similar screen, the firmware works so no issues there haha

caveman99: That is not really authoritative ... :-) 

I looked it up, that's definitely a typo. Should be SCREEN_SSD1306 (or any other screen model, these are usually auto detected) and you define them only if you have problem with autodetect. May be the reason it works for you despite the typo.

Babyyoda777: ah ok fixed :)

PR #5156, opened 2024-10-27
De-duplicate Ambient LED management code

fifieldt:
We currently have 4 different places where we have the logic for modifying LED state of the various types of Ambient LEDs, ExternalNotificationModule::SetExternalOff
ExternalNotificationModule::SetExternalOn
AmbientLightingThread::setLighting
AmbientLightingThread::setLightingOff

This patch de-duplicates the methods in ExternalNotification to a single method, using a boolean to toggle whether we're turning things on or off.

### DIFF
--- a/src/modules/ExternalNotificationModule.cpp
+++ b/src/modules/ExternalNotificationModule.cpp
@@ -93,7 +93,7 @@ int32_t ExternalNotificationModule::runOnce()
             nagCycleCutoff = UINT32_MAX;
             LOG_INFO("Turning off external notification: ");
             for (int i = 0; i < 3; i++) {
-                setExternalOff(i);
+                setExternalState(i, false);
                 externalTurnedOn[i] = 0;
                 LOG_INFO("%d ", i);
             }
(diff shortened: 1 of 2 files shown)


fifieldt:
Github diff doesn't make it easy to see: here's past off and on:
image

image

fifieldt:
Where I think this could go next -- make AmbientLighting responsible for LED stuff and call it from ExternalNotification.

thebentern (approved):

caveman99 (approved):

caveman99:
> Where I think this could go next -- make AmbientLighting responsible for LED stuff and call it from ExternalNotification.

We are trying to resolve those cross-depencencies of the modules. In principal it is good but can we use the notification system we have in the firmware for IPC. E.G. Ambient lighting offering a subscription service for LED changes and the External Notifications just using it (if available)

fifieldt:
Thanks, I appreciate the guidance. Shall we merge this one as a start - since it'll be a while before I can get to the next steps?

thebentern (approved):

PR #5160, opened 2024-10-27
Update GPSUpdateScheduling.cpp

spiffysec:
Default value is too short, resulting in unstable GPS locks on T1000-E (possibly others). Fix has been tested an confirmed working with no adverse effects, by multiple users. Also discussed at length on Discord

### DIFF
--- a/src/gps/GPSUpdateScheduling.cpp
+++ b/src/gps/GPSUpdateScheduling.cpp
@@ -70,8 +70,7 @@ bool GPSUpdateScheduling::isUpdateDue()
 // Have we been searching for a GPS position for too long?
 bool GPSUpdateScheduling::searchedTooLong()
 {
-    uint32_t maxSearchMs =
-        Default::getConfiguredOrDefaultMs(config.position.position_broadcast_secs, default_broadcast_interval_secs);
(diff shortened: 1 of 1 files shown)


Nestpebble (commented):
I was going to suggest making it T1000E specific, like this:
```cpp

// Have we been searching for a GPS position for too long?
bool GPSUpdateScheduling::searchedTooLong()
{
#if !defined(_VARIANT_TRACKER_T1000_E_)
    uint32_t maxSearchMs =
        Default::getConfiguredOrDefaultMs(config.position.position_broadcast_secs, default_broadcast_interval_secs);

#elif defined(_VARIANT_TRACKER_T1000_E_)
// T1000000000E tracker seems to need much longer to search for a decent lock
// The default is 5 minutes (300s). 15 minutes seems to work ok.
    uint32_t maxSearchMs = 15*60*1000;

#endif
```

Maybe this is too cumbersome, tho?

thebentern:
Interesting. efault_broadcast_interval_secs

thebentern:
Updated this with the minimum time coerce logic for now. We'll be revisiting this for a more long term fix that takes into account the extended cold lock times vs subsequent ones, but for now, this fixes a lot of folks having GPS issues

PR #5164, opened 2024-10-27
Add region BR

tmedicci:
Create a new region-specific band for Brazil. Initially, this region will be a copy of the existing `ANZ` band. This is done to avoid any confusion for Brazilian users and the Meshtastic community.

Also, this region allows the users to keep using the project's MQTT server with the appropriate default root topic `msh/BR` (instead of using the `msh/ANZ`, which was always confusing.

Finally, according to Brazil's telecommunications agency (ANATEL), the ISM band can be used for LoRa. This includes the 915-928 MHz band (the same used by `ANZ`), but the 902-905 MHz can be used too. Then, keeping BR as a separate region makes it easier to deal with its particularities and use cases for future implementations.

# Testing

This PR was tested with a patched version of the Python client (to be sent upstream soon). The default MQTT root topic is `msh/BR` (as supposed) and the device can still communicate with the old `ANZ` region.

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -65,6 +65,11 @@ const RegionInfo regions[] = {
     RDEF(ANZ, 915.0f, 928.0f, 100, 0, 30, true, false, false),
 
     /*
+        https://www.gov.br/anatel/pt-br/regulado/radiofrequencia/radiacao-restrita
+     */
+    RDEF(BR, 915.0f, 928.0f, 100, 0, 30, true, false, false),
+
+    /*
         https://digital.gov.ru/uploaded/files/prilozhenie-12-k-reshenyu-gkrch-18-46-03-1.pdf
 
         Note:


tmedicci:
Hi @thebentern, it seems there is an automated CI job to open a PR regarding the updated

garthvh:
> Hi @tmedicci , welcome to Meshtastic and thanks for adding Brazil.
> 
> No problem having a separate region, but could you give us some screenshots from the regulations that show the frequencies and power restrictions?
> 
> Is 902-905 MHz usable now? Are there any power differences for that range to 915-928?

Andre is in Brazil and is familiar with the regulations, I think I the recommendation is to use ANZ

andrekir:
Countries are grouped into regions based on shared frequency profiles. Brazil and most of South America uses the ANZ/AU915 frequency plan. This new region would be a duplicate of ANZ.

tmedicci:
> Countries are grouped into regions based on shared frequency profiles. Brazil and most of South America uses the ANZ/AU915 frequency plan. This new region would be a duplicate of ANZ.

Yes, it's a duplicate for now. Pay attention, however, to the Brazilian community and the default root topic for MQTT (which being `msh/ANZ` makes no sense for the users). Apart from that, most of the Brazilian users use AU915, however ANATEL also allows using the 902-905 MHz band (which doesn't make part of this change yet).

tmedicci:
> Hi @tmedicci , welcome to Meshtastic and thanks for adding Brazil.
> 
> No problem having a separate region, but could you give us some screenshots from the regulations that show the frequencies and power restrictions?
> 
> Is 902-905 MHz usable now? Are there any power differences for that range to 915-928?

Yes, 902-905 MHz is usable (not yet implemented on Meshtastic). The following table (from https://informacoes.anatel.gov.br/legislacao/resolucoes/2023/1834-resolucao-759) shows it:
image

The power restrictions are the same for the 915-928MHz band.

andrekir:
> Yes, it's a duplicate for now.

Once a new region is added, it can't be changed later.

> the default root topic for MQTT (which being msh/ANZ makes no sense for the users)

This is true for 90% of all countries. The default MQTT topic is just a suggestion, users can set any root topic they want.

tmedicci:
> Once a new region is added, it can't be changed later

Yes, I know. It would be something like `BR_902`. I meant that it was worth creating a separate region to avoid confusion.

> This is true for 90% of all countries. The default MQTT topic is just a suggestion, users can set any root topic they want.

Yes, this is true, but most users use the default value. This is always a question for the Brazilian community ("Why are there only nodes from Australia"/"How to chat with other people in Brazil"). The meshtastic community (and coverage) in Brazil has started growing only recently: having a specific region for Brazil (despite being a duplicate) that also sets an appropriate topic would help quite a lot to increase adoption.

caveman99:
The band plan is a technical description of allowed frequencies. From your screenshot i see it starts at 905 (not 902) and i'd be interested for a native speaker to evaluate the pitfalls in the fine print. It would not be the first time someone claiming 'it's fine to use that' and when you read on you are not allowed to use it for some quirky footnote reson.

The other argument (root topic) is something that is not the objective of this band plan, but a convenience measure that can also be implemented in the apps (ios does that already for the US states iirc) If we went by your reasoning why it is simportant ot have a COPY of ANZ just so that BR is the root topic, we would have 195 radio regions tomorrow. I don't wanna scroll through this list setting my region.

tmedicci:
> The band plan is a technical description of allowed frequencies. From your screenshot i see it starts at 905 (not 902) and i'd be interested for a native speaker to evaluate the pitfalls in the fine print. It would not be the first time someone claiming 'it's fine to use that' and when you read on you are not allowed to use it for some quirky footnote reason.

I'm sorry, the correct screenshot:
image

I didn't claim "it's fine to use", I just said that in Brazil two bands are possible. Most users use AU915, but 902-905 is usable. Check the "DESTINAÇÃO" column: you can check that the content of the 902-915 lines is the same of the 915-927,75 one.

> 
> The other argument (root topic) is something that is not the objective of this band plan, but a convenience measure that can also be implemented in the apps (ios does that already for the US states iirc) If we went by your reasoning why it is important ot have a COPY of ANZ just so that BR is the root topic, we would have 195 radio regions tomorrow. I don't wanna scroll through this list setting my region.

I see... well, I'll try to figure out how to do that. There aren't so many devices in Brazil and the users are always asking why there are so many people in Australia appearing on their maps. This PR was intended to 1) "fix" the default topic and 2) differentiate the ANZ-like region from a yet-to-be-implemented 902-905 band.

I'll close this PR, then.

PR #5165, opened 2024-10-27
Native config.d

jp-bennett:
New approach to make configs a bit easier. Adds the config.d folder, pre-populated with small config files, that simply need to be renamed to be activated.


src/platform/portduino/PortduinoGlue.cpp
```diff
@@ -100,33 +101,20 @@ void portduinoSetup()
...
     if (configPath != nullptr) {
         std::cout << "Using " << configPath << " as config file" << std::endl;
-        try {
-            yamlConfig = YAML::LoadFile(configPath);
-        } catch (YAML::Exception &e) {
-            std::cout << "Could not open " << configPath << " because of error: " << e.what() << std::endl;
-            exit(EXIT_FAILURE);
-        }
+        loadConfig(configPath);
     } else if (access("config.yaml", R_OK) == 0) {
```
charlieh0tel, line 112: TOCTOU bug here and below.    (No obvious security issue.)    A better practice is to just try to open the file and deal with failures from open(2).

jp-bennett: Good point.

PR #5168, opened 2024-10-28
Cherrry pick bin/config-dist.yml from TFT-GUI-Work

fifieldt:


jp-bennett:
I think this one also depends on a LovyanGFX lib bump.

caveman99 (approved):

PR #5173, opened 2024-10-28
cherry-pick:  USB mode=1 messed up the debug log

fifieldt:


mverch67:
If we'd keep the non-fancy version of the T-Deck (for users to decide what they prefer) then it would be beneficial to set 
-DARDUINO_USB_MODE=1 for the less fancy version, because it shows all esp-idf logs as well (e.g. stack-dump when crashing) like for the tbeam.

Fancy TFT should leave it as -DARDUINO_USB_MODE=0 as the logs from the two tasks are intertwined and unreadable.
Overwriting the definition from the boards .json must be done via `build_unflags` definition.

Similar for the ARDUINO_EVENT_RUNNING_CORE. Would leave the current value for non-fancy.

If it's agreed to keep the non-fancy version I'd push a new t-deck `platformio.ini` that contains targets for both versions which would render this cherry-picked commit superfluous.

caveman99:
> If it's agreed to keep the non-fancy version I'd push a new t-deck `platformio.ini` that contains targets for both versions which would render this cherry-picked commit superfluous.

I'd vote for keeping both legacy and fancy versions of firmware that support the new UI.

fifieldt:
This needs a solution other than cherry pick, and I don't have a tdeck to test with. Closing assuming it'll be fixed in the tft branch :)

PR #5178, opened 2024-10-28
Default rebroadcast mode for Router and Repeater to ignore problematic portnums

thebentern:


src/mesh/Router.cpp
```diff
@@ -591,19 +591,22 @@ void Router::handleReceived(meshtastic_MeshPacket *p, RxSource src)
...
+#endif
+        if (shouldIgnoreNonstandardPorts && p->which_payload_variant == meshtastic_MeshPacket_decoded_tag &&
             (p->decoded.portnum == meshtastic_PortNum_ATAK_FORWARDER || p->decoded.portnum == meshtastic_PortNum_ATAK_PLUGIN ||
              p->decoded.portnum == meshtastic_PortNum_PAXCOUNTER_APP || p->decoded.portnum == meshtastic_PortNum_IP_TUNNEL_APP ||
              p->decoded.portnum == meshtastic_PortNum_AUDIO_APP || p->decoded.portnum == meshtastic_PortNum_PRIVATE_APP ||
              p->decoded.portnum == meshtastic_PortNum_DETECTION_SENSOR_APP ||
              p->decoded.portnum == meshtastic_PortNum_RANGE_TEST_APP ||
              p->decoded.portnum == meshtastic_PortNum_REMOTE_HARDWARE_APP)) {
-            LOG_DEBUG("Ignoring packet on blacklisted portnum during event");
+            LOG_DEBUG("Ignoring packet on blacklisted portnum for CORE_PORTNUMS_ONLY");
```
caveman99, line 606: don't we have this nifty is_one_of macro?

caveman99 (approved):

caveman99:
Just for clarification, these are defaults, not runtime enforced? cause if i wanna have a router that will also forward audio, i would set the role and then switch the core packet restraint off again...

thebentern:
> Just for clarification, these are defaults, not runtime enforced? cause if i wanna have a router that will also forward audio, i would set the role and then switch the core packet restraint off again...

Right. You set the Role to Router, then you go change the Rebroadcast mode to All after that.

src/mesh/Router.cpp
thebentern: Hah! I didn't even think about, but's a perfect application for it. Changes will be pushed shortly

PR #5190, opened 2024-10-29
fix: don't broadcast public keys if the user is licensed

andrekir:
prevents NodeInfo broadcasts with public keys when the user is licensed.

### DIFF
--- a/src/modules/NodeInfoModule.cpp
+++ b/src/modules/NodeInfoModule.cpp
@@ -81,6 +81,12 @@ meshtastic_MeshPacket *NodeInfoModule::allocReply()
         ignoreRequest = false; // Don't ignore requests anymore
         meshtastic_User &u = owner;
 
+        // Strip the public key if the user is licensed
+        if (u.is_licensed && u.public_key.size > 0) {
+            u.public_key.bytes[0] = 0;
+            u.public_key.size = 0;
+        }
+
         LOG_INFO("sending owner %s/%s/%s", u.id, u.long_name, u.short_name);
         lastSentToMesh = millis();
         return allocDataProtobuf(u);


thebentern (approved):

jp-bennett:
Do we need this, with the is_licensed change I pushed earlier?

JohnathonMohr:
Please excuse my ignorance, but I'm curious to learn why this is necessary. I understand ham has to operate unencrypted, so the licensed node can't send anything encrypted, but why prevent another non-licensed node from sending something encrypted. Must a licensed node also not _receive_ anything encrypted?

thebentern:
> Do we need this, with the is_licensed change I pushed earlier?

Not sure your change will have an effect if they already have keys generated.

andrekir:
> Do we need this, with the is_licensed change I pushed earlier?

yes. public keys are still being sent in the NodeInfo broadcasts when `is_licensed` is true.

jp-bennett:
> Please excuse my ignorance, but I'm curious to learn why this is necessary. I understand ham has to operate unencrypted, so the licensed node can't send anything encrypted, but why prevent another non-licensed node from sending something encrypted. Must a licensed node also not _receive_ anything encrypted?

Last time I checked, I came to the conclusion that it was a problem, yes. Of course I can't currently find that source.

jp-bennett:
> > Do we need this, with the is_licensed change I pushed earlier?
> 
> yes. public keys are still being sent in the NodeInfo broadcasts when `is_licensed` is true.

Got it. Yeah, looks good then.

fromthebeanbag:
I tried to find the legislative answer to amateur radio and encryption and only went down a rabbit hole. In Australia atleast from https://www.acma.gov.au/amateur-radio-operating-procedures I found the following. 

But receiving signals as an amateur would mean the other party could only be another amateur and thus only unencrypted anyway. This not allowing reception of unencrypted would just be additional checks and balances.

Except

Encryption/scrambling
Transmissions from an amateur station must not be encrypted or scrambled, except for signals used to control a satellite, signals used to control a remote amateur station or by stations participating in emergency services operations or exercises.

Re-transmission
If you re-transmit another station’s transmission, you must have the other station’s permission and indicate it is a retransmission.

PR #5195, opened 2024-10-31
Don't skip GPS serial speeds, and always land on GPS_BAUDRATE

jp-bennett:
Two tiny tweaks to GPS search. First, it really looks like the speedSelect skip can trigger each time through the loop of serialSpeeds, which means that we could skip some speeds every time. And second, rather than just always defaulting to 9600 after not finding anything, we should default to the defined speed in GPS_BAUDRATE. (Credit to Technologyman00 on Discord for that idea)


Technologyman00:
You need to also edit https://github.com/meshtastic/firmware/blob/GPS-search-tweak/src%2Fgps%2FGPS.cpp#L434

fifieldt (approved):

Issue #5152, opened 2024-10-26
[Bug]: fromRadioQueue is full


hb9hha:
### Category

BLE

### Hardware

T-Echo

### Firmware Version

2.5.7.f77c87d

### Description

Hello meshtastic team !

I have a tricky bug with my t-echo.

Background:
I am using a T-Echo with my android phone using BLE. A personal MQTT server is configured with the "Proxy to client enabled" turned ON. I am using the default LongFast EU_868 modem settings. Node list is at the max 100 nodes and ChUtil is around 20%. Traffic is somewhat intense.

Symptoms:
The T-Echo stop working after a few hours. Android app says connected to a radio but last transmit time is not "now" but minutes/hours ago. Messages are not received any more. The blinking heartbeat blue LED is not blinking. The T-echo seems dead. A reset with the RST button solve the issue for a while.

Relevant logs:
First, all go smoothly for a few hours. I included standard logs from received packets and MQTT traffic.
Then, I think the e-paper screen is doing a FULL refresh.
Soon after that, we are seeing BUG logs:
DEBUG | 12:37:05 9726 [Screen] BUG! fromRadioQueue is full! Discarding! (id=0x55b78200 fr=0xe8 to=0xff, WantAck=0, HopLim=4 Ch=0x1 Portnum=67 rxtime=1729946226 rxSNR=-8.25 rxRSSI=-109 via M

And then the logs loops for the next 15 minutes.

I am not sure where to look. Is that a MQTT related problem? Is that a problem with the e-paper display? Do I have too much traffic to process? I do not have such behaviours with the T3S3 nor with the T-Beam connected with Wi-Fi and MQTT. T-echo users from my area are seeing similar behaviour with their t-echo. It seems persistent with this model.

Let me know if I can dig deeper and get some more debug logs.

Thanks for the great work !

73 de HB9HHA

### Relevant log output

```Shell
DEBUG | 12:36:50 9711 New ToRadio packet
INFO  | 12:36:50 9711 Got MqttClientProxy message
INFO  | 12:36:50 9711 Received MQTT topic msh/EU_868/2/e/LongFast/!7c5a4724, len=129
DEBUG | 12:36:50 9711 Using channel 1 (hash 0x8)
DEBUG | 12:36:50 9711 Expanding short PSK #1
DEBUG | 12:36:50 9711 Using AES128 key!
[...]

PR #5212, opened 2024-11-01
Drop oldest packet from radio when queue is full

GUVWAF:
Potential fix for #5152.

It looks like the radio was enqueuing packets so quickly that the Router could not handle them fast enough. Then eventually the `fromRadioQueue` gets full, in which case the Router thread was never activated again because `setReceivedMessage()` wasn’t called. I now rewrote the logic such that it discards the oldest when it’s full and always notifies the Router to dequeue.
I tested it by faking that the queue is full, and it seems to work.

From the issue it looks like again the full e-ink refresh takes too much time (maybe in combination with MQTT proxy). This actually sounded quite familiar to what I’ve experienced once also with the T-Echo when I was in the big mesh in Vancouver, but I didn’t have serial logs then. @thebentern @caveman99 what about increasing `MAX_RX_FROMRADIO` from 4 to let’s say 8, to avoid dropping packets when the Router can’t handle them quickly enough? Or maybe only for devices with e-ink screens?


thebentern:
Looks good to me. The MAX_RX_FROMRADIO increase sounds good. We'll have to take another look at memory footprint on the different platforms after that increase.

thebentern (approved):

GUVWAF:
After a second look at the logs it really looks like it was caused due to a couple MQTT proxy packets received back-to-back during a full e-ink refresh. I actually don't think it's worth increasing it since this would normally not happen via LoRa.

PR #5228, opened 2024-11-02
Spell check all Code

Technologyman00:
I went through and spell checked all the code. I didn't want to change any significant code, so I only spell checked comments and some strings for logging. I did change the wording of Celsius in a conversion function. I also tried my best to not change regional spellings as somethings are spelled differently in the UK vs the US.


fifieldt:
Thanks for this. This is good to go except the edits to files in src/mesh/generated . These files are automatically generated and will be overwritten. Changes here might cause merge conflicts. Could you remove those files from the commit?

For reference, they come from the protobuf repo.

PR #5229, opened 2024-11-03
Fix displays showing "GPS Not Present" until first lock

fifieldt:
https://github.com/meshtastic/firmware/pull/5160 introduced a change which made first publication of GPS information take up to 15mins. For that initial period, displays would show "No GPS Present", even if one was detected.

This change fixes that bug, triggering publication immediately after a GPS module is detected.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -447,10 +447,12 @@ bool GPS::setup()
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


fifieldt:
Tested on wio-tracker-wm1110, t114, tbeam.

thebentern (approved):

PR #5242, opened 2024-11-03
Exclude preferred routing roles from nodeinfo interrogation behavior

### LINKED ISSUE #5238: [Feature Request]: Add option to disable automatic nodeinfo requests from node
### Platform

NRF52, ESP32, RP2040

### Description

We have a small problem on our [slovenian] mesh where strategically positioned nodes (routers) have their nodedb filled up pretty fast. This means that nodes are getting removed when the limit (of 100) is reached.

This becomes a real problem when this node hears new packets (position/neighbourinfo/telemetry) from the nodes that it removed and since the node is not present in nodedb, it will sent its nodeinfo to the mesh (also requesting nodeinfo from the unknown node, resulting in two packets being sent over the mesh).

This cycle constantly repeats when you have more than 100 active nodes.

I think the responsible part of the code is in [`src/mesh/MeshService.cpp#89`](https://github.com/meshtastic/firmware/blob/d00acc5d7bd8c936e5d8088463355b08398a2012/src/mesh/MeshService.cpp#L89). A new setting would have to be added and an if statement checking that setting when attempting to automatically broadcast node information.

This change would only require a protobuf change in a single Config protobuf definition, in my opinion `DeviceConfig` would be a best fit.

thebentern:
Closes #5238


fifieldt (approved):

GUVWAF (approved):

src/mesh/MeshService.cpp
```diff
@@ -85,8 +85,9 @@ int MeshService::handleFromRadio(const meshtastic_MeshPacket *mp)
                                                                               //  ignore our request for its NodeInfo
     } else if (mp->which_payload_variant == meshtastic_MeshPacket_decoded_tag && !nodeDB->getMeshNode(mp->from)->has_user &&
                nodeInfoModule) {
-        LOG_INFO("Heard new node on channel %d, sending NodeInfo and asking for a response.", mp->channel);
-        if (airTime->isTxAllowedChannelUtil(true)) {
+        if (airTime->isTxAllowedChannelUtil(true) && !IS_ONE_OF(config.device.role, meshtastic_Config_DeviceConfig_Role_ROUTER,
+                                                                meshtastic_Config_DeviceConfig_Role_REPEATER)) {
+            LOG_INFO("Heard new node on channel %d, sending NodeInfo and asking for a response.", mp->channel);
             nodeInfoModule->sendOurNodeInfo(mp->from, true, mp->channel);
         } else {
```
fifieldt, line 92: This will create a lot of logs on routers saying "Skip sending NodeInfo due to > 25 percent channel util."

Change to "Skip sending NodeInfo due to > 25 percent channel util or we are are ROUTER"? ;)

thebentern: Good catch!

fifieldt (approved):

Issue #5193, opened 2024-10-31
Optimise GPS probe code - one message per family


fifieldt:
In the GPS probe code (
https://github.com/meshtastic/firmware/blob/aae346aef74b72d5914a470d7f19f18bce2f0ec3/src/gps/GPS.cpp#L1137 ), we write commands on the serial line and determine which GPS we have based on the result.

GPS units in the same family sometimes use the same command, but return different results (eg AG3335 and AG3332 both use $PAIR021*39). Currently we run the command once per GPS. Instead we should run each command only once per family, record the result, and select the GNSS MODEL based on the result.

This will mean probing takes a shorter time.

absolut3ego:
can you assign this to me?

fifieldt:
Welcome to Meshtastic, we're happy to have you!


One way I thought about implementing this was to change the PROBE_SIMPLE macro.

At the end, save the command we sent as a variable (lastcmd = TOWRITE). Then, at the start, check if TOWRITE == lastcmd and perform the check there. Would require changing GPS::getACK so that it can handle returning the buffer or accepting a pointer to a buffer.

Let us know if you need any pointers!

absolut3ego:
Hi @fifieldt,

I have a few questions. First, can you tell me how the PROBE_SIMPLE macro currently works? To make sure that my modification can work correctly with your project. Second, how do you want me to solve this issue? As I initially understood, I need to add something to store the results of each GPS family, so that I can call a family many times without having to re-execute the PROBE_SIMPLE macro. But according to your suggestion, I understand it in a different way, which is to save the last command. If the next command matches it, I will get the result that was just called to return. Can you explain this issue more clearly? I need to understand clearly to be able to have the correct solution.

PR #5244, opened 2024-11-04
Fix issue #5193: Optimise GPS probe code - one message per family

absolut3ego:
I have upgraded the PROBE_SIMPLE macro. So that it can not need to do everything again when calling the same message. I simply added some variables to save the last command and based on it to execute the next commands, if it matches the latest command I do not need to do the steps in the PROBE_SIMPLE macro again but instead I will get the result and return it.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -40,6 +40,9 @@ GPS *gps = nullptr;
 
 GPSUpdateScheduling scheduling;
 
+static const char *lastcmd = nullptr;
+static GPS_RESPONSE cachedResponseStatus = GNSS_RESPONSE_NONE;
+
 /// Multiple GPS instances might use the same serial port (in sequence), but we can
 /// only init that port once.
 static bool didSerialInit;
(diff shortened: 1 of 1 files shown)


fifieldt:
Thanks for this! Taking a look :)

fifieldt:
Two things you can do while waiting for review:
* Sign the Contributor License Agreement https://cla-assistant.io/meshtastic/firmware?pullRequest=5244
* Learn about the Trunk lint tool - it's how we keep the format of our code consistent. You'll note one of the checks is failing.

fifieldt (changes requested):
Thanks for this. A great beginning.

Further changes are needed. At the moment this checks whether we got a response to the command, which is a good start. 

However, we actually care about the contents. Since the response to the command could be OK, but might correspond to a model number we don't support.

For example, we send the command  $PCAS06,1*1A to look for ATGM336H or ATGM332D. When we see a response of $GPTXT,01,01,02,HW=ATGM336H we know the chip we have is an ATGM336H. If we see a response of $GPTXT,01,01,02,HW=ATGM332D we know it's an ATGM332D. However, we could also see $GNTXT,01,01,01,PCAS inv format*23. where it's an invalid command for the chip connected , or it could be $GPTXT,01,01,02,HW=ATGM330B , which we don't support.

So the code should be extended to make the buffer from the first time we run the command available, so we can search it next time.

fifieldt:
Hi @absolut3ego , did you need any help or pointers with this? I have a lot of GPS to test with, so if you want to make the code so can  test it for you :)

fifieldt:
Hi @absolut3ego , do you still have time to work on this?

fifieldt:
Superseded by https://github.com/meshtastic/firmware/pull/6114

PR #5264, opened 2024-11-06
Smarter traffic scaling

thebentern:
Closes #964 

This PR addresses a few issues with the current mesh size based intervalled traffic management:

1. Smaller networks (backpacking, small towns, etc) of <= 40 are encouraged to transmit more frequently.
2. Sensors and trackers don't get telemetry / position throttled because these are crucial data points for their respective roles.
3. The coefficient for throttling back the interval has more judicious values for the faster presets now, so it's less of a one size fits all poorly approach.


GUVWAF (approved):
Looks good!

PR #5267, opened 2024-11-06
Only PKC encrypt when packet originates from us

GUVWAF:
This should fix the issue when a node using firmware <2.5 sends a direct message to a 2.5 node, and it hops through another 2.5 node. Currently, as reported on Discord, the intermediate 2.5 node would PKC encrypt it with its own public key when the DM was received on its primary channel, because it could decrypt it and then before rebroadcasting, this passes: 
https://github.com/meshtastic/firmware/blob/982190936dfe47b20a7147d4a6627a3ade7ac3c5/src/mesh/Router.cpp#L495-L496

Only when a packet originates from us, we should PKC encrypt.


GUVWAF:
No, right? In your case it's sent from a 2.4 device, then goes through a 2.5 device and received by a 2.5 node.

psyh0x:
Exactly

caveman99 (approved):

PR #5269, opened 2024-11-07
Fix device flashing scripts so they work with esptool when it's installed via pipx

jeremiah-k:
When installing esptool via `pipx install esptool` it doesn't work to use `python -m esptool`.  

In linux `esptool.py` is used and in Windows `esptool`.

This PR adds checks to first determine which command should be used, before executing the commands to flash the devices.

### DIFF
--- a/bin/device-install.bat
+++ b/bin/device-install.bat
@@ -2,13 +2,32 @@
 
 set PYTHON=python
 
+:: Determine the correct esptool command to use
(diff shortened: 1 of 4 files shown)


fifieldt (approved):
Thank you very much for your detailed fix. Welcome to Meshtastic!

jeremiah-k:
> Thank you very much for your detailed fix. Welcome to Meshtastic!

Thank you, sir!

PR #5219, opened 2024-11-01
musl compatibility

vidplace7:
Existing meshtasticd code currently fails to compile with musl.

This PR seeks to resolve compilation errors for musl so that it can be used with [buildroot-meshtastic](https://github.com/vidplace7/buildroot-meshtastic).

Including `#include <sys/types.h>` resolves issues caused by musl not including `uint` and `ulong` type definitions by default.

Cousin PR: https://github.com/meshtastic/framework-portduino/pull/36

PR #5270, opened 2024-11-07
uClibc compatibility

vidplace7:
Similar to https://github.com/meshtastic/firmware/pull/5219

This PR resolves compile errors when compiling with `uClibc`, so that it may be used with [buildroot-meshtastic](https://github.com/buildroot-meshtastic/buildroot-meshtastic).

Notably, `uClibc` is the natively supported toolchain for the LuckFox Pico.


src/mesh/StreamAPI.h
```diff
@@ -3,6 +3,7 @@
 #include "PhoneAPI.h"
 #include "Stream.h"
 #include "concurrency/OSThread.h"
+#include <cstdarg>
```
vidplace7, line 6: `va_list` type definition is not included in `uClibc` by default.

fifieldt (approved):
Thanks!

arch/portduino/portduino.ini
```diff
@@ -32,6 +32,7 @@ build_flags =
   -Isrc/platform/portduino
   -DRADIOLIB_EEPROM_UNSUPPORTED
   -DPORTDUINO_LINUX_HARDWARE
+  -lstdc++fs
```
vidplace7, line 35: Allows `std::filesystem` linking to function on `uClibc`

PR #5280, opened 2024-11-08
rak10701 (rak wismeshtap)  optimization

DanielCao0:
* Improve the processing speed of virtual keyboards
* Fixed turned off gps caused the screen not to display
* Changed the default screen sleep time to 30s
* Rename platform rak10701 -> rak wismeshtap


platformio.ini
```diff
@@ -29,7 +29,7 @@ default_envs = tbeam
 ;default_envs = rak4631
 ;default_envs = rak4631_eth_gw
 ;default_envs = rak2560
-;default_envs = rak10701
+default_envs = rak_wismeshtap
```
fifieldt, line 32: Probably want to leave this commented

DanielCao0: Fixed

DanielCao0:
This is an urgent merge request, hope to get help, thank you .

fifieldt (approved):
I reviewed the code and confirmed it only affected the particular device.

There may be a less invasive way to perform the exclusions, but since the committer labeled it urgent I would be OK merging this and performing cleanup later.

DanielCao0:
> I reviewed the code and confirmed it only affected the particular device.
> 
> There may be a less invasive way to perform the exclusions, but since the committer labeled it urgent I would be OK merging this and performing cleanup later.

I found a issue, uf2 of master 10701 can not be used, my branch is also the same issue, is it because the firmware of nrf52 is too large? The 4631 branch does not have this issue@fifieldt

DanielCao0:
a73fbdaecce5d79a74d7e31c754bcbe
Whether uf2 cannot be used has much to do with flash ?

DanielCao0:
> a73fbdaecce5d79a74d7e31c754bcbe Whether uf2 cannot be used has much to do with flash ?

Fixed .     uf2 doesn't work because the firmware is too big
-DMESHTASTIC_EXCLUDE_WIFI=1
-DMESHTASTIC_EXCLUDE_EXTERNALNOTIFICATION=1
-DMESHTASTIC_EXCLUDE_WAYPOINT=1

caveman99 (approved):

PR #5281, opened 2024-11-08
Add new port number for airborne application(s)

pjalocha:
I am very sorry, I forgot to actually change the name of the new port !
I fixed it in the next push to the same branch, but I am not sure how to proceed now.
Sorry for the mess.

I propose a new port number for airborne applications like OGN-Tracer or FANET or ADS-L whch have well establilshed compact position packet formats. It is more efficient to put position in this format, which takes like 16 bytes for OGN than use the GPS generic format where many elements are missing like climb rate ro turn rate.

The new port number would allow efficient (low rate) position reports by objects like amateur balloons.

The formatting would be simple: I define ID's for particulat packet type and they would be embedded as a pack of bytes, the decoding of the pack would follow the particular system.

If this gets accepted, I will proceed to define those ID's

Thank you.

BTW. I am beta-receiving Meshtantic on ground receivers of the Open Glider Network (OGN) and I wonder if it would be worth to upload this traffic to a server of yours ?

### DIFF
--- a/src/mesh/generated/meshtastic/portnums.pb.h
+++ b/src/mesh/generated/meshtastic/portnums.pb.h
@@ -127,7 +127,7 @@ typedef enum _meshtastic_PortNum {
     /* PowerStress based monitoring support (for automated power consumption testing) */
     meshtastic_PortNum_POWERSTRESS_APP = 74,
     /* Airborne tracking: allow existing systemw with compact position formats put those on Meshtantic */
-    meshtastic_PortNum_POWERSTRESS_APP = 75,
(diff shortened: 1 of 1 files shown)


fifieldt (changes requested):
As @gjelsoe said, this change needs to be ade in the protobufs repo.

pjalocha:
Thank you, I will make the changes, and push them.

pjalocha:
I have done the hopefully correct now change in the meshtantic/protobufs and made a pull request there.
You might then remove this pull request and treat the other one.
Thank you.

caveman99:
this is a null change. closing in favor of protobuf PR's

PR #5286, opened 2024-11-08
Add setting to transmit NeighborInfo over LoRa

GUVWAF:
As discussed in #5082 and https://github.com/meshtastic/protobufs/pull/618. 

When setting `neighbor_info.transmit_over_lora` to true (default to false), next to sending it to MQTT/PhoneAPI, it will also broadcast it over LoRa to be able to obtain it from remote nodes. This happens only when not using the default channel as primary, and the channel utilization allows it. 

Furthermore, the absolute minimum you can set the NeighborInfo broadcast interval to is bumped to 14400 (4 hours).

### DIFF
--- a/src/mesh/Default.h
+++ b/src/mesh/Default.h
@@ -19,7 +19,7 @@
 #define default_node_info_broadcast_secs 3 * 60 * 60
 #define default_neighbor_info_broadcast_secs 6 * 60 * 60
 #define min_node_info_broadcast_secs 60 * 60 // No regular broadcasts of more than once an hour
-#define min_neighbor_info_broadcast_secs 2 * 60 * 60
+#define min_neighbor_info_broadcast_secs 4 * 60 * 60
 
 #define default_mqtt_address "mqtt.meshtastic.org"
 #define default_mqtt_username "meshdev"
--- a/src/modules/NeighborInfoModule.cpp
+++ b/src/modules/NeighborInfoModule.cpp
@@ -121,7 +121,12 @@ Will be used for broadcast.
 */
 int32_t NeighborInfoModule::runOnce()
 {
-    sendNeighborInfo(NODENUM_BROADCAST_NO_LORA, false);
+    if (moduleConfig.neighbor_info.transmit_over_lora && !channels.isDefaultChannel(channels.getPrimaryIndex()) &&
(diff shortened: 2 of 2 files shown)


thebentern (approved):

fifieldt (approved):

macvenez:
>This happens only when not using the default channel as primary, and the channel utilization allows it.

Hi, does this mean that NeighborInfo data is only transmitted to the default channel if it's not primary? I don't know if there's still this limitation but I remember position data was sent only on primary channel, would this mean we'll have neighbor info on default channel but not position?

What's the purpose of this? Wouldn't this still generate the same traffic?
Thanks!

GUVWAF:
Hi @macvenez, indeed, periodic broadcasts only go out on the primary channel, so if that is the default channel, it won’t allow it. It was added upon request here https://github.com/meshtastic/protobufs/pull/618.

When you have a private primary channel, it will automatically use a different frequency slot. When using `EU_868` there is only one, so indeed then it would lead to the same channel utilization on the mesh.

ragnarblackmane:
@GUVWAF 

So:
-if a default channel (e.g. MediumFast), is primary: no NeighborInfo over LoRa (as per pull #5286);
-if a default channel (e.g. MediumFast), is secondary: no NeighborInfo over LoRA (normal behaviour since periodic broadcasts happen only on Primary channel);
-wherever can be used more than one frequency (e.g. USA): NeighborInfo traffic goes on another frequency (since a private primary channel is used);
-in EU (single slot): we'll have a bit less channel utilization because by effectively disabling NeighborInfo over LoRa on primary+default channel.

Am i right?

GUVWAF:
- Yes, if also the default key is used.
- Depends on what is your primary, but yes, no NeighborInfo goes out via a secondary channel.
- Yes (unless you specifically change the frequency slot again).
- Correct.

todd2982:
I need to verify this more, but it seems this feature broadcasts info about clients set with the "CLIENT_HIDDEN" preset to other nodes.

fifieldt:
Ah! that's no good!

fifieldt:
Nodeinfo defaults the broadcast interval for NodeInfo for HIDDEN nodes to UINT_MAX

 667         moduleConfig.neighbor_info.update_interval = UINT32_MAX;

fifieldt:
How about this: https://github.com/meshtastic/firmware/pull/5464

PR #5300, opened 2024-11-10
Exclude some niche modules by default and populate exclude_modules

thebentern:


fifieldt (approved):

fifieldt:
Needed to fix the failing jobs: https://github.com/meshtastic/firmware/pull/5301

fifieldt (approved):
Thanks, like the logic

andrekir (approved):

garthvh:
> What's the purpose of adding `#define NO_EXT_GPIO 1`
> 
> For the T-Echo?
> 
> Because it seems to cause Critical Error #3 as seen in #6294

Because there is no GPIO available.

Logicbloke:
> Because there is no GPIO available.

However, the flag seems to be causing the exclusion of modules that are essential to the operation of the LoRa radio. I didn't manage to pinpoint which, but commenting it out fixes the problem in #6294

Thanks!

PR #5301, opened 2024-11-10
Web now(?) comes in a /build subdirector in the tar

fifieldt:
So let's move everything to where it's expected and cleanup before trying to unzip.


fifieldt:
(merged to try and fix the build)

PR #5303, opened 2024-11-10
Package file move - include dotfiles

fifieldt:
Adds shopt -s dotglob nullglob to ensure we get 'hidden' files in our move command.


fifieldt:
(merging to try and fix the build)

PR #5310, opened 2024-11-11
Don't send to public channel

gjelsoe:
`p->to` wasn't set and had the same value as broadcast, it's now set to our own NodeNum.

### DIFF
--- a/src/modules/DetectionSensorModule.cpp
+++ b/src/modules/DetectionSensorModule.cpp
@@ -58,7 +58,7 @@ int32_t DetectionSensorModule::runOnce()
     // moduleConfig.detection_sensor.minimum_broadcast_secs = 30;
     // moduleConfig.detection_sensor.state_broadcast_secs = 120;
     // moduleConfig.detection_sensor.detection_trigger_type =
-    //          meshtastic_ModuleConfig_DetectionSensorConfig_TriggerType_LOGIC_HIGH;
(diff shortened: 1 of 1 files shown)


caveman99:
Why would you want to send this to the originating nodenumber?

GUVWAF:
But this means it will not be sent out over LoRa at all. 

You can make your primary channel private and add the public channel as a secondary channel if you don't want to send it on the public channel.

gjelsoe:
That kind of the issue, they don't do that and ends up spamming the channel.

GUVWAF:
But not sending it out over LoRa at all is not a proper solution.

gjelsoe:
That fair..

Closing it.

thebentern:
FWIW, I liked the previous idea of adding the configuration of DMing to a specific nodenum destination. Particularly for the use case of detecting a state, I think often the desired recipient is one node on the mesh rather than the whole mesh

gjelsoe:
That is also the one I would prefer and Garth wanted at solution that didn't involve changes that might not get into the apps.

garthvh:
A cli only solution will get essentially zero use.

garthvh:
I agree this should not be on the default key anymore, the choosing of a channel or an individual node is not something We are currently set up to manage.

gjelsoe:
Is this acceptable ?

thebentern (approved):

Issue #5259, opened 2024-11-05
[Bug]: Low Battery voltage detection threshold is way  too low for the most LDOs being used on ESP32 boards


dm5tt:
### Category

Other

### Hardware

Other

### Firmware Version

2.5.x

### Description

Did a bit of code reviewing.. still hunting our ESP32 LittleFS corruption bug.

**Problem**

If I understand this piece of code correctly:

```
Power.cpp, 714:
    // If we have a battery at all and it is less than 0%, force deep sleep if we have more than 10 low readings in
    // a row. NOTE: min LiIon/LiPo voltage is 2.0 to 2.5V, current OCV min is set to 3100 that is large enough.
    //
    if (batteryLevel && powerStatus2.getHasBattery() && !powerStatus2.getHasUSB()) {
        if (batteryLevel->getBattVoltage() < OCV[NUM_OCV_POINTS - 1]) {
            low_voltage_counter++;
            LOG_DEBUG("Low voltage counter: %d/10", low_voltage_counter);
            if (low_voltage_counter > 10) {
#ifdef ARCH_NRF52
                // We can't trigger deep sleep on NRF52, it's freezing the board
                LOG_DEBUG("Low voltage detected, but not triggering deep sleep");
#else
                LOG_INFO("Low voltage detected, triggering deep sleep");
                powerFSM.trigger(EVENT_LOW_BATTERY);
#endif
            }
        } else {
            low_voltage_counter = 0;
        }
    }
}
```

We go to Deep Sleep with our ESP32 device as soon the battery reaches 3100mV

```
#else // LiIon
#define OCV_ARRAY 4190, 4050, 3990, 3890, 3800, 3720, 3630, 3530, 3420, 3300, 3100
#endif
```

Most (>90%) of our devices are using a 3V3 LDO that will crap out as soon the battery goes below 3.4V (the good ones) or even higher if bad LDOs (looking at you AMS1117) are being used. This effect is almost binary.

This renders the Low Battery protection basically useless.

**Suggested Solution**

1. Increase generic minimum detection to at least 3420mV or even 3530mV. I'd vote vote 3530mv as the device should survive a while on deep sleep.

2. The threshold should be configurable from the board definition files

thebentern:
I'm in favor of at least bumping this on at least a subset of variants in a generic way if it makes a more reliable experience with device flash persistence.

HarukiToreda:
[...]

PR #5312, opened 2024-11-11
Don't attempt to save NodeDB on low-batt shutdown to prevent FS corruption

thebentern:
Addresses part of #5259 by skipping the NodeDB save on low battery shutdown. This should result in minimal loss of data, because the NodeDB is saved in other domain events within the firmware regularly. Plus in any regard, it's better than potentially losing the whole enchilada.

@todd-herbert we can collaborate on this branch to try to figure out the low voltage warning startup halt, or we work that separately and push this one forward.
I was researching the best ways to persist the status that we shut down due to low battery, so that we can check attempt to run some ADC samples on the next boot. It looks like RTC slow memory could be an option: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/memory-types.html#rtc-slow-memory


thebentern:
Also, I'm going to get a harness setup for my PSU so that I can reproduce this condition on demand *hopefully*

caveman99 (approved):

dm5tt:
Reproducing it shouldn't be too complicated. 

You can start with pulling and plugging the power cable into the PSU out and in fstly by trying to abort the early boot-up sequence.

todd-herbert:
@dm5tt Do you feel that the corruption is happening on boot, rather than during low power shutdown?

todd-herbert (approved):

PR #5315, opened 2024-11-11
Fix sending duplicate packets to PhoneAPI/MQTT

GUVWAF:
Accidentally introduced with #5232.

When receiving repeated packets due to dropped ACKs/replies, we should re-do any routing or sending ACK/replies, but we shouldn’t resend it to the PhoneAPI or MQTT. 

To check this, I moved the PacketHistory to the Router class (which is needed for #2856 anyway).


thebentern (approved):

fifieldt (approved):

GUVWAF:
@thebentern Just found that in some cases this actually causes it to not send to the PhoneAPI when it actually should. Will try to find a better fix...

VigibotDev:
Thanks for solving. Going test :)

VigibotDev:
Testing 2.5.13 for duplicate message it's not ok for now. Easy to reproduce for me feel free ask more testing.

Screenshot_20241114_180949_Meshtastic
Screenshot_20241114_180838_Meshtastic

VigibotDev:
OK I must update the other side also because it can receive twice also

VigibotDev:
Solved for me :)
Screenshot_20241114_183511_Meshtastic
Screenshot_20241114_183834_Meshtastic

PR #5322, opened 2024-11-12
Reduce the flash usage of wismeshtap platform

DanielCao0:
Reduce the flash usage of wismeshtap platform ,  If the flash usage is too high, the .uf2 cannot be used


variants/rak_wismeshtap/platformio.ini
```diff
@@ -10,8 +10,16 @@ build_flags = ${nrf52840_base.build_flags} -Ivariants/rak_wismeshtap -D RAK_4631
...
   -DMESHTASTIC_EXCLUDE_WIFI=1
-  -DMESHTASTIC_EXCLUDE_EXTERNALNOTIFICATION=1
+  -DMESHTASTIC_EXCLUDE_PAXCOUNTER=1
   -DMESHTASTIC_EXCLUDE_WAYPOINT=1
+  -DMESHTASTIC_EXCLUDE_DETECTIONSENSOR=1
+  -DMESHTASTIC_EXCLUDE_RANGETEST=1
+  -DMESHTASTIC_EXCLUDE_REMOTEHARDWARE=1
+  -DMESHTASTIC_EXCLUDE_AUDIO=1
+  -DMESHTASTIC_EXCLUDE_STOREFORWARD=1
+  -DMESHTASTIC_EXCLUDE_STOREFORWARD=1
```
fifieldt, line 20: you have STOREFORWARD twice

DanielCao0: Fixed.  Thank you.

fifieldt (approved):
Range test is pretty popular in my local communities. Does it work if you keep it in?

caveman99:
How likely is it this target will include the wisblock ethernet port? i think that's a good chunk of code we can guard. I can add a define for it.

DanielCao0:
> Range test is pretty popular in my local communities. Does it work if you keep it in?

Thanks for your advice. I've resumed the distance test.  As long as the usage does not exceed 99%, the uf2 can run normally
RAM:   [==        ]  21.7% (used 53928 bytes from 248832 bytes)
Flash: [==========]  98.1% (used 799848 bytes from 815104 bytes)

DanielCao0:
> How likely is it this target will include the wisblock ethernet port? i think that's a good chunk of code we can guard. I can add a define for it.

This device does not support ethernet port.

fifieldt (approved):

variants/rak_wismeshtap/platformio.ini
```diff
@@ -10,8 +10,14 @@ build_flags = ${nrf52840_base.build_flags} -Ivariants/rak_wismeshtap -D RAK_4631
   -DEINK_WIDTH=250
   -DEINK_HEIGHT=122
   -DMESHTASTIC_EXCLUDE_WIFI=1
-  -DMESHTASTIC_EXCLUDE_EXTERNALNOTIFICATION=1
+  -DMESHTASTIC_EXCLUDE_PAXCOUNTER=1
   -DMESHTASTIC_EXCLUDE_WAYPOINT=1
+  -DMESHTASTIC_EXCLUDE_DETECTIONSENSOR=1
+  -DMESHTASTIC_EXCLUDE_REMOTEHARDWARE=1
+  -DMESHTASTIC_EXCLUDE_AUDIO=1
```
thebentern, line 17: These can be removed because they are handled in the parent nrf52.ini:
-DMESHTASTIC_EXCLUDE_PAXCOUNTER=1
-DMESHTASTIC_EXCLUDE_AUDIO=1

variants/rak_wismeshtap/platformio.ini
```diff
@@ -10,8 +10,14 @@ build_flags = ${nrf52840_base.build_flags} -Ivariants/rak_wismeshtap -D RAK_4631
   -DEINK_WIDTH=250
   -DEINK_HEIGHT=122
   -DMESHTASTIC_EXCLUDE_WIFI=1
-  -DMESHTASTIC_EXCLUDE_EXTERNALNOTIFICATION=1
+  -DMESHTASTIC_EXCLUDE_PAXCOUNTER=1
   -DMESHTASTIC_EXCLUDE_WAYPOINT=1
+  -DMESHTASTIC_EXCLUDE_DETECTIONSENSOR=1
+  -DMESHTASTIC_EXCLUDE_REMOTEHARDWARE=1
```
thebentern, line 16: This one can be removed because it is handled in the root platform.ini:
-DMESHTASTIC_EXCLUDE_REMOTEHARDWARE=1

variants/rak_wismeshtap/platformio.ini
DanielCao0: Thank you. Fixed

PR #5326, opened 2024-11-12
Update variant.h for T-Echo RGB Matrix LED

Lma-Mike-Zulu:
Changed/Corrected mapping of (left) RGB Matrix LED.

Confirmed by reviewing the vendors sceamtic ( My version 2021/6/23 has wrong labeling for colors ) and checking the datasheet of the Matrix LED FM-B2020RGBA-HG mentioned in the scematic.

Tryed it out with my sample of T-Echo delivered 11/2024 seeing Pin 0.14 ( 0 +14) blinked blue with the "old" code and trying green led with External Messaging module succsessfully on Pin P1.01 (32 + 1)

### DIFF
--- a/variants/t-echo/variant.h
+++ b/variants/t-echo/variant.h
@@ -43,17 +43,18 @@ extern "C" {
 #define NUM_ANALOG_OUTPUTS (0)
 
 // LEDs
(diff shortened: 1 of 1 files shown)


thebentern:
Your assertions are correct. Unfortunately, changing to new (correct) LED scheme in broke the previous iterations of T-Echos. :-/

Issue #5333, opened 2024-11-13
[Feature Request]: Appropriate 900MHz band limits for Indonesia?


ndoo:
### Platform

Cross-Platform

### Description

Dear devs,

There's 2 possibly relevant current sections of Indonesian law (https://jdih.kominfo.go.id/produk_hukum/view/id/862/t/peraturan+menteri+komunikasi+dan+informatika+nomor+2+tahun+2023) regarding 900 MHz band in Indonesia for IoT:

1)

![Screenshot 2024-11-13 100042](https://github.com/user-attachments/assets/6736681b-82a4-4a59-930f-32ff90d60bd3)

2)

![Screenshot 2024-11-13 101417](https://github.com/user-attachments/assets/d44bbd9e-04e5-461f-998c-671dcd96b0ee)

"Sesuai dengan Standar Teknis yang berlaku" is translated by Google to "In accordance with applicable Technical Standards".

For the life of me, I can't find any reference to this in any current regulation in the search page of [JDIH Kemkominfo - Search](https://jdih.kominfo.go.id/produk_hukum/pencarian).

What's an appropriate direction to take to determine the appropriate Tx power for ID_923? It looks like the Meshtastic community in Indonesia is currently using SG_923.

Would adding ID_923 at 26 dBm or 20 dBm (which?), 250 kHz b/w, 1% duty cycle be a good idea?

garthvh:
A 1% duty cycle won't work for meshtastic, you might contact the lorawan alliance and see if they have thoughts.

ndoo:
> A 1% duty cycle won't work for meshtastic, you might contact the lorawan alliance and see if they have thoughts.

Good point. Actually, the route to go would be the SRD allocation at 920 - 923 MHz, so I'll continue looking for anything that might provide the allowed EIRP for it.

ndoo:
Looks like this is the most recent document that specifies EIRP for 920 - 923 MHz band - at <= 500 mW (27 dBm):

[Regulation of the Minister of Communication and Information Technology Number 35 of 2015 concerning Technical Requirements for Short-Range Telecommunication Tools and Equipment (Regulation of the Minister of Communication and Information Technology Number 35 of 2015 concerning Technical Requirements for Short-Range Telecommunication Tools and Equipment)](https://jdih.kominfo.go.id/produk_hukum/view/id/510/t/peraturan+menteri+komunikasi+dan+informatika+nomor+35+tahun+2015+tanggal+31+desember+2015)

[...]

PR #5334, opened 2024-11-13
Add Indonesia region

ndoo:
# Add Indonesia region

Protobufs: https://github.com/meshtastic/protobufs/pull/622
Discussion/regulatory document links in related issue (and code comments): https://github.com/meshtastic/firmware/issues/5333

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -148,6 +148,18 @@ const RegionInfo regions[] = {
     RDEF(PH_915, 915.0f, 918.0f, 100, 0, 24, true, false, false),
 
     /*
+        Indonesia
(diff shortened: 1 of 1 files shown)


caveman99 (commented):
This is a copy of the KR band plan. Please use that. the 27dBm is theoretical anyway since the most powerful radios we use can do 22dBm max (without PA)

caveman99:
This is a copy

ndoo:
> This is a copy of the KR band plan. Please use that. the 27dBm is theoretical anyway since the most powerful radios we use can do 22dBm max (without PA)

Got it. Thank you. 

Though, with regulatory gain fully implemented, it may be desirable at some point to implement this anyway.

PR #5338, opened 2024-11-13
Decrease max nodes for NRF52 to 80 as workaround to prevent FS blowouts

thebentern:


fifieldt:
You sure we can't fix this with even shorter logs?

caveman99:
> You sure we can't fix this with even shorter logs?

Log's don't go on the node filesystem.

caveman99 (approved):

fifieldt (approved):

Issue #5295, opened 2024-11-10
[Feature Request]: Define user button pin in userPrefs file


CypressXt:
### Platform

Cross-Platform

### Description

Hi there :wave:, 
I would be nice to be able to define the default user button pin (`PIN_BUTTON`) number in the `userPrefs` allowing this setting to persist even after a reset of the board.

Issue #4580, opened 2024-08-28
[Feature Request]: Add position to userPrefs.h


rcarteraz:
### Platform

NRF52, ESP32, RP2040, Linux Native

### Description

We need to add the ability to set the device's position (latitude, longitude, altitude) in the userPrefs.h file. This feature would serve two important purposes:

Provide a hardcoded fallback if something happens to the device that has a manually set position (such as remotely deployed nodes).
Which also solves the issue of the complaints about not being able to set the position remotely.

gergy008:
I'm still a bit of a git newbie - I just come across this while I was adding custom MQTT server to the userprefs.h file. If no-one else solves this issue I'll attempt my first pull request after some testing and see if I can resolve this

PR #5341, opened 2024-11-13
Adds fixed GPS, BUTTON_PIN and BLE code to userPrefs.h

gjelsoe:
Following will be added to `userPrefs.h`
- Adds fixed GPS (#5295)
- Adds fixed BUTTON_PIN (#4580)
- Adds fixed Bluetooth pairing code

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -270,7 +270,7 @@ NodeDB::NodeDB()
         config.position.fixed_position = true;
 #endif
     }
-#endif    
+#endif
     saveToDisk(saveWhat);
 }
 
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -824,7 +824,7 @@ void NodeDB::loadFromDisk()
     // disk we will still factoryReset to restore things.
 
     // static DeviceState scratch; We no longer read into a tempbuf because this structure is 15KB of valuable RAM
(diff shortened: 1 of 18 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -246,6 +246,31 @@ NodeDB::NodeDB()
         config.position.gps_mode = meshtastic_Config_PositionConfig_GpsMode_ENABLED;
         config.position.gps_enabled = 0;
     }
+#ifdef USERPREFS_FIXED_GPS
+    if (myNodeInfo.reboot_count == 1) { // Check if First boot ever or after Factory Reset.
+        meshtastic_Position FixedGPS = meshtastic_Position_init_default;
```
thebentern, line 251: Let's do `fixedGPS` to stick with camelcase for locals

gjelsoe: So drop `if (myNodeInfo.reboot_count == 1) {` ?

fifieldt: I think Ben means change the name of the FixedGPS variable to fixedGPS

gjelsoe: Oh.....
No problem in changing to `fixedGPS` but since everything else in `userPrefs.h` has a `USERPREFS_` prefix, shouldn't be `USERPREFS_fixedGPS` ?

fifieldt: down two lines: meshtastic_Position FixedGPS

fifieldt (approved):

PR #5355, opened 2024-11-14
RP2040: Update core; add mDNS support

GUVWAF:
Updated the core, hopefully to fix potential FreeRTOS issues. 
It also includes a new `SimpleMDNS` library that does work with FreeRTOS. Added a "meshtastic" service as there is no HTTP/HTTPS service, which the Android app now also looks for: https://github.com/meshtastic/Meshtastic-Android/pull/1401. Tested this and it indeed finds the IP address now.

Lastly, added the `-g` option to link the source files for debugging.

### DIFF
--- a/src/mesh/wifi/WiFiAPClient.cpp
+++ b/src/mesh/wifi/WiFiAPClient.cpp
@@ -20,7 +20,7 @@
 #include <ESPmDNS.h>
 #include <esp_wifi.h>
 static void WiFiEvent(WiFiEvent_t event);
-#else // ARCH_RP2040
+#elif defined(ARCH_RP2040)
 #include <SimpleMDNS.h>
 #endif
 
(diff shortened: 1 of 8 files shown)


fifieldt (approved):
Checked that it only affects rpi, changes seem fine.

src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -59,19 +61,21 @@ static void onNetworkConnected()
...
         } else {
-            LOG_INFO("mDNS responder started");
             LOG_INFO("mDNS Host: Meshtastic.local");
+#ifdef ARCH_ESP32
             MDNS.addService("http", "tcp", 80);
             MDNS.addService("https", "tcp", 443);
-        }
-#else // ESP32 handles this in WiFiEvent
-        LOG_INFO("Obtained IP address: %s", WiFi.localIP().toString().c_str());
+#else
```
fifieldt, line 72: is it worth guarding this in case other architectures have wifi in the future?

fifieldt (approved):

broglep:
Great to see MDNS support for the actual meshtastic tcp port, wanted to suggest that as well. 
 Is there any particular reason why not always announcing `MDNS.addService("meshtastic", "tcp", 4403);`, instead just for `ARCH_RP2040` ?

GUVWAF:
Good question, it seemed a bit redundant to me, but it would indeed be more correct.

PR #5364, opened 2024-11-16
Add sudo to apt-get commands for Raspbian Build

fifieldt:
Without sudo, inadequate permissions to runs the commands meant the build was failing.


fifieldt:
merging to fix build

Issue #4969, opened 2024-10-06
[Feature Request]: Add all 3 PKI admin key slots to userPrefs file


thespillmonkey:
### Platform

Cross-Platform

### Description

Add a way to define admin keys 0, 1, and 2 to userprefs on compile

PR #5368, opened 2024-11-16
Allows all 3 PKI keys to be added to userPrefs.h (#4969) and a tool.

gjelsoe:
Allows all 3 PKI keys to be added to userPrefs.h and solves #4969 
A little tool for converting base64 PKI Keys to decoded byte that userPrefs.h can understand.


fifieldt (approved):
Thank you for this :)

src/mesh/NodeDB.cpp
```diff
@@ -407,9 +407,30 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
...
+
+    // Check if USERPREFS_ADMIN_KEY_0 is non-empty
+    if (sizeof(USERPREFS_ADMIN_KEY_0) > 0) {
+        memcpy(config.security.admin_key[numAdminKeys].bytes, USERPREFS_ADMIN_KEY_0, 32);
+        config.security.admin_key[numAdminKeys].size = 32;
+        numAdminKeys++;
+    }
+
+    // Check if USERPREFS_ADMIN_KEY_1 is non-empty
+    if (sizeof(USERPREFS_ADMIN_KEY_1) > 0) {
```
thebentern, line 421: Should these be wrapped in #ifdef USERPREFS_ADMIN_KEY_N?
Wondering if trying to run a sizeof() on a macro which is not present might cause a panic.

gjelsoe: USERPREFS_ADMIN_KEY_N aren't macros.

thebentern (approved):

Issue #5348, opened 2024-11-14
[Bug]: External Notification : output_ms is ignored RAK 4631 with RAK 19003 using GPIO Buzzer.  Alert sounds continuously.


vitaminrad:
### Category

Hardware Compatibility

### Hardware

Rak4631

### Firmware Version

2.5.9.936260f

### Description

I'm using a RAK 4631 with a RAK 19003 carrier board.  I want a super loud buzzer alert for incoming messages.  I'm using the unused Tx pin on the RAK 19003, which maps to GPIO 19.  In the External Notification module, I have configured the pin using buzzer_gpio, with a nag_timeout of 60 seconds and an output_ms of 1000 ms.

The buzzer sounds but does not stop after 1000 ms.  It simply continues to ring until I depress a user button connected to the Rx pin (GPIO pin 20).  The buzzer sounds until the nag_timeout is reached.  In the past, using a RAK 19007 board, it behaved as expected.

Oscilloscope shows a clean signal.  I don't see any serial data or anything weird.

Configuring the device using the buzzer_gpio and enabling use_pwm it seems to work but there is hesitation between ringtones.

PR #5375, opened 2024-11-17
Bug fixed in ExternalNotificationModule

gjelsoe:
While `nagging` setExternalState wasn't written to Buzzer & Vibra so output was never toggled.

Possible fix for #5348


thebentern (approved):

vitaminrad:
Thank you @gjelsoe

PR #5376, opened 2024-11-17
Cleanup static files from bad Web UI bundle on 2.5.13 release

thebentern:


src/mesh/NodeDB.cpp
```diff
@@ -823,6 +823,7 @@ void NodeDB::loadFromDisk()
         0; // Mark the current device state as completely unusable, so that if we fail reading the entire file from
     // disk we will still factoryReset to restore things.
 
+    rmDir("/static/static"); // Remove bad static web files bundle from initial 2.5.13 release
```
caveman99, line 826: maybe check if it exists first?

caveman99 (approved):

PR #5383, opened 2024-11-17
Minimize time between channel scan and actual transmit

GUVWAF:
We should really try to minimize the time between the Channel Activity Detection procedure (`scanChannel()`) and the actual transmit (`startTransmit()`) to minimize the chance someone else is starting a packet during this time. I therefore moved a call to `millis()` and a debug print to after `startTransmit()`, and added a note about trying to keep this time as short as possible.


thebentern (approved):
Seems like a good micro-optimization

PR #5386, opened 2024-11-17
/api/v1/fromradio: add OPTIONS handler for CORS.

### LINKED ISSUE #5385: [Bug]: Web client initial load is very slow (90 seconds) in Chrome (compared to Firefox - 4 seconds)
### Category

WiFi

### Hardware

Heltec V3

### Firmware Version

2.5.11.8e2a3e5

### Description

Chrome (~90 seconds):

https://github.com/user-attachments/assets/6945fdc0-bc10-4005-93c8-74faf714507e

Firefox (~4 seconds):

https://github.com/user-attachments/assets/60737f5f-e9b8-48ba-b4e9-88adb8af74ee

I think this is because Chrome re-establishes the connection (TCP+TLS) for each request: 
Screenshot 2024-11-17 at 3 17 12 PM

Screenshot 2024-11-17 at 3 18 42 PM

Connection keep-alive is supposed to work, and make a huge difference, as seen in Firefox, where each request takes only ~60 ms.

I thought I figured out the reason Chrome connections don't keep alive but I found some conflicting clues so now I'm not so sure I understand why.

### Relevant log output

_No response_

cpatulea:
This avoids hitting the 404 Not Found handler, which breaks connection keep-alive, so this change fixes a big performance regression for Web Client in Chrome: https://github.com/meshtastic/firmware/issues/5385

Tested on Heltec V3.

Fixes #5385


thebentern (approved):

thebentern:
Ah, so they get pre-flighted now? Good deal. Thanks for the fix.
Once you sign the CLA, I can get this merged in.

cpatulea:
Sorry, I need a bit more time for the CLA.

cpatulea:
CLA is signed.

PR #5390, opened 2024-11-18
Streamline board definitions for first tech preview.

caveman99:
TBD: Indicator


arch/nrf52/nrf52.ini
```diff
@@ -1,10 +1,10 @@
 [nrf52_base]
 ; Instead of the standard nordicnrf52 platform, we use our fork which has our added variant files
-platform = platformio/nordicnrf52@^10.5.0
+platform = platformio/nordicnrf52@^10.6.0
 extends = arduino_base
 platform_packages =
   ; our custom Git version until they merge our PR
-  framework-arduinoadafruitnrf52 @ https://github.com/geeksville/Adafruit_nRF52_Arduino.git
+  framework-arduinoadafruitnrf52 @ https://github.com/geeksville/Adafruit_nRF52_Arduino.git#4f591d0f71f75e5128fab9dc42ac72f1696cf89f
```
thebentern, line 7: We may want to move this to our fork

mverch67 (approved):

arch/nrf52/nrf52.ini
caveman99: 9.3 compiler works fine, but it increases binary size slightly. we need to check for optimizations, especially for RAK4631, that gets to 99%

PR #5391, opened 2024-11-18
add smiley emoji

jcyrio:
Added smiley emoji. Accepts 5 different common Unicode smiley emojis as input, so that the phone user doesn't have to remember which one is compatible.


thebentern (approved):

thebentern:
Looks good. Can you run a trumk fmt?

fifieldt (approved):

PR #5397, opened 2024-11-19
Windows Support - Trunk and Platformio

ThatKalle:
Ths PR aims to extend the support for Windows users by leveraging the [devcontainer](https://code.visualstudio.com/docs/devcontainers/containers) feature.

It's targeted towards the tft-gui-work branch as I've used a T-Deck Plus to test and validate, that said, the changes are not really tft-gui specific persay, so maybe it would be better suited somewhere else.

The two issues with developing using Windows is that Trunk.io and Platformio is not fully supported.

## Trunk.io

Trunk.io works nicely within the devcontainer simply by installing the vscode extension during build.
```json
  "customizations": {
    "vscode": {
      "extensions": [
        "Trunk.io"
      ]
    }
  },
```
It finds and uses the configuration within `/.trunk/`.

Installing the same extension on Windows natively results in all kinds of errors.

## PlatformIO

PlatformIO does have very much the same sort of issue as Trunk.io, it does not play nicely on Windows native, but does work from within the devcontainer.
```json
  "customizations": {
    "vscode": {
      "extensions": [
        "platformio.platformio-ide"
      ]
    }
  },
```

A feature that needs a bit of a helping hand however is the ability to Upload the built firmware to the device using the serial connection (in the case of T-Deck Plus).

Due to how the forwarding of devices work from Windows to WSL2 to Devcontainer not only do we need to make some changes to the container itself, but also to Windows and WSL.

### Windows 

Using [usbipd-win](https://github.com/dorssel/usbipd-win) we're able to send a connected device down WSL2.
```bat
> usbipd list
Connected:
BUSID  VID:PID    DEVICE
4-7    303a:1001  USB Serial Device (COM18), USB JTAG/serial debug unit
> usbipd bind --busid=4-7
> usbipd attach --wsl --busid=4-7 --auto-attach
```

### WSL2

Within WSL2 we'll see the device
```bash
$ lsusb
Bus 001 Device 005: ID 303a:1001 Espressif USB JTAG/serial debug unit
$ sudo dmesg | grep tty
[ 2944.930440] cdc_acm 1-1:1.0: ttyACM0: USB ACM device
```

But, we also need to setup [udev](https://en.wikipedia.org/wiki/Udev) rules for PlatformIO supported boards/devices. [ref](https://docs.platformio.org/en/stable/core/installation/udev-rules.html)

```bash
$ sudo apt-get install udev -y
$ curl -fsSL https://raw.githubusercontent.com/platformio/platformio-core/develop/platformio/assets/system/99-platformio-udev.rules | sudo tee /etc/udev/rules.d/99-platformio-udev.rules
$ sudo service udev restart
```

The `ttyACM0` device should now be ready to be sent over to the devcontainer to be used by Platformio from within VSCode.

### devcontainer

The `"runArgs": ["--device=/dev/ttyACM0"],` needs to be adjusted to match your system. On my machine it ended up being `/dev/ttyACM0`, It's probably not the case for every system, hence commented out.

The same udev rules for PlatformIO is also added to the devcontainer via Dockerfile.

## GPG

The devcontainer needs the `gpg` and `gnupg2` packages to allow for GPG signed commits to be created.


koga73:
Related to Windows support, I built the tft-gui-work branch on Windows with PlatformIO but was getting this error:

> lto-wrapper.exe: fatal error: CreateProcess: No such file or directory

As a workaround I removed the `-flto` compile flag from my local and this resolved the problem.
Will this PR address `lto-wrapper.exe` missing?

ThatKalle:
> Related to Windows support, I built the tft-gui-work branch on Windows with PlatformIO but was getting this error:
> 
> > lto-wrapper.exe: fatal error: CreateProcess: No such file or directory
> 
> As a workaround I removed the `-flto` compile flag from my local and this resolved the problem. Will this PR address `lto-wrapper.exe` missing?

The error message looks very familiar.

This PR will not resolve the fact that `lto-wrapper.exe` is missing, can't be found, or can't find what it's looking for.

However, it does allow Windows users to do the building and uploading using a devcontainer _with some setup_, rather than locally. The devcontainer does not experience the `lto-wrapper.exe` ralated error during my testing, most likely due to it being a more supported Linux box, rather than a Windows one.

.devcontainer/Dockerfile
```diff
@@ -20,6 +26,12 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
...
     zip \
+    usbutils \
+    hwdata \
+    gpg \
+    gnupg2 \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
-RUN pipx install platformio==6.1.15
\ No newline at end of file
+RUN pipx install platformio==6.1.15
```
mverch67, line 35: maybe add `install --upgrade meshtastic` ? :)

.devcontainer/Dockerfile
```diff
@@ -1,6 +1,12 @@
+# trunk-ignore-all(trivy/DS002)
+# trunk-ignore-all(trivy/DS026)
+# trunk-ignore-all(checkov/CKV_DOCKER_2)
+# trunk-ignore-all(checkov/CKV_DOCKER_3)
 FROM mcr.microsoft.com/devcontainers/cpp:1-debian-12
```
caveman99, line 5: Don't ignore, rather fix the annotations

caveman99 (changes requested):

fifieldt (approved):

.devcontainer/Dockerfile
fifieldt: I think there's a chance that might conflict, since this is designed for dev-envs?

fifieldt:
Thanks a lot for your dedication and excellent write-up, @ThatKalle .

PR #5403, opened 2024-11-20
Make heart emoji usable

jcyrio:
Made heart emoji usable by matching it with the most commonly used heart emojis on smartphones. Beforehand it was only matching the "black heart suit" emoji, which I don't think I've ever seen anyone use before (https://graphemica.com/%E2%99%A5). 

Also cleaned/shortened code a little by using a "msg" constant instead of repeating "reinterpret_cast<const char *>(mp.decoded.payload.bytes)" dozens of times.


fifieldt (approved):
Thanks for the cleanup

Issue #5382, opened 2024-11-17
[Feature Request]: Add --web argument to device-install(.sh/.bat) scripts to bundle web app littlefs bin


thebentern:
### Platform

ESP32

### Description

This argument will change the logic to esptool flash littlefswebui-{version}.bin instead of littlefs-{version}.bin

GUVWAF:
Closed by #5405.

PR #5405, opened 2024-11-20
--web added to device-install(.sh/.bat)

gjelsoe:
Solves #5382


fifieldt (approved):

bin/device-install.bat
```diff
@@ -49,8 +52,14 @@ IF EXIST %FILENAME% IF x%FILENAME:update=%==x%FILENAME% (
     ) else (
         %ESPTOOL_CMD% --baud 115200 write_flash 0x260000 bleota-s3.bin
     )
-    for %%f in (littlefs-*.bin) do (
-        %ESPTOOL_CMD% --baud 115200 write_flash 0x300000 %%f
+    IF %WEB_APP%==1 (
+        for %%f in (littlefsweb-*.bin) do (
```
GUVWAF, line 56: Shouldn't this be `littlefswebui-`? Same for the .sh script.

gjelsoe: You are right, though I did changed it.

gjelsoe: It should be correct now, is a new pr needed?

gjelsoe: Np

PR #5410, opened 2024-11-21
add canned message and keyboard in indicator board

Dylanliacc:
Add canned message and keyboard in indicator board  (The basic functionality comes from the support of T-Watch-S3, with some macro definitions enabled）


mverch67:
Rather than creating a `#define` mess in the code which will become more difficult to maintain in future I'd like to suggest using a single `#define` macro for the feature that this macro enables, e.g. `USE_VIRTUAL_KEYBOARD` which would then replace all device specific macros (such as` #if defined(USE_VIRTUAL_KEYBOARD)` replaces `#if defined(T_WATCH_S3 || defined(SENSECAP_INDICATOR)` everywhere in the code).
This way if in future a new device is added it just needs to put this feature macro into the variant file and good is. No need to change code at ten different places and no risk to break things by changing existing code and simpler to review and more likely and easier to get approved.

Dylanliacc:
> Rather than creating a `#define` mess in the code which will become more difficult to maintain in future I'd like to suggest using a single `#define` macro for the feature that this macro enables, e.g. `USE_VIRTUAL_KEYBOARD` which would then replace all device specific macros (such as` #if defined(USE_VIRTUAL_KEYBOARD)` replaces `#if defined(T_WATCH_S3 || defined(SENSECAP_INDICATOR)` everywhere in the code). This way if in future a new device is added it just needs to put this feature macro into the variant file and good is. No need to change code at ten different places and no risk to break things by changing existing code and simpler to review and more likely and easier to get approved.

I will make some modifications tomorrow ,I just worried that the modifications you mentioned earlier might have some strange effects on the T_WATCH_S3 and RAK10043, as they involve both canned messages and UI rendering

Dylanliacc:
Added virtual keyboard macro and enabled Indicator I have made adjustments to the macro definition according to mverch's suggestions, but due to the inability to test TWatch_S3 and RAK14014, i have temporarily not made any changes to their code

thebentern:
> Added virtual keyboard macro and enabled Indicator I have made adjustments to the macro definition according to mverch's suggestions, but due to the inability to test TWatch_S3 and RAK14014, i have temporarily not made any changes to their code

I cleaned up the macros a bit by adding a DISPLAY_CLOCK_FRAME and applying USE_VIRTUAL_KEYBOARD to those variants. I will test for regressions against the T-Watch S3. I have a WisMesh Tap on the way and I can test that one as well.

PR #5413, opened 2024-11-21
Check for OkToMqtt flag presence before uplinking to MQTT

thebentern:
Untested at the moment


GUVWAF (changes requested):
I think we should do this in `MQTT:onSend()` as it's called on two places and there is logic for the OK to MQTT bit already there.

thebentern:
Good catch. That one somehow eluded me. Moved things over to the existing block of logic in mqtt->onsend.

PR #5414, opened 2024-11-21
Telemetry can respond to want-response for LocalStats variant

### LINKED ISSUE #4563: [Feature Request]: Add the absolute amount of received Meshtastic packets to telemetry
### Platform

NRF52, ESP32, RP2040, Linux Native, other

### Description

Problem: It's at the moment it's impossible to determine the utilization of a Router remotely. Without these data we cannot determine the health of the HF chain nor if the setup/position of the entire device can be improved.

Right now following data can be gathered using the Python script:

```
~$ /home/had/.local/bin/meshtastic --host 192.x.x.x --request-telemetry --dest \!279xxxx
Connected to radio
Sending telemetry request to !279xxxx on channelIndex:0 (this could take a while)
Telemetry received:
Battery level: 81.00%
Voltage: 4.00 V
Total channel utilization: 42.40%
Transmit air utilization: 13.42%
Uptime: 109625 s
```

It would be great to have an additional absolute counter of received, valid Meshtastic packets.

Example:

`Received Packets: 20000`

On the other side "Battery level" and "Voltage" is IMHO redundant.

thebentern:
Closes #4563

### DIFF
--- a/src/modules/Telemetry/DeviceTelemetry.cpp
+++ b/src/modules/Telemetry/DeviceTelemetry.cpp
@@ -77,9 +77,10 @@ meshtastic_MeshPacket *DeviceTelemetryModule::allocReply()
         // Check for a request for device metrics
         if (decoded->which_variant == meshtastic_Telemetry_device_metrics_tag) {
             LOG_INFO("Device telemetry reply to request");
-
-            meshtastic_Telemetry telemetry = getDeviceTelemetry();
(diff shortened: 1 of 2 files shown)


GUVWAF:
"le" is calling and wants to be in as well, by the way ;)

thebentern:
Le sigh 😅

PR #5415, opened 2024-11-21
Update build-native.sh

madeofstown:
Device-install.sh and device-update.sh are not used on Linux native platform, skip copying to release directory after build and copy native-install.sh and native-run.sh instead.

### DIFF
--- a/bin/build-native.sh
+++ b/bin/build-native.sh
@@ -28,4 +28,3 @@ platformio pkg update --environment native || platformioFailed
 pio run --environment native || platformioFailed
 cp .pio/build/native/program "$OUTDIR/meshtasticd_linux_$(uname -m)"
 cp bin/native-install.* $OUTDIR
-cp bin/native-run.* $OUTDIR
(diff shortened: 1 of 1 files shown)


fifieldt (approved):

bin/build-native.sh
```diff
@@ -27,5 +27,5 @@ rm -r $OUTDIR/* || true
 platformio pkg update --environment native || platformioFailed
 pio run --environment native || platformioFailed
 cp .pio/build/native/program "$OUTDIR/meshtasticd_linux_$(uname -m)"
-cp bin/device-install.* $OUTDIR
-cp bin/device-update.* $OUTDIR
+cp bin/native-install.* $OUTDIR
+cp bin/native-run.* $OUTDIR
```
thebentern, line 31: Native install is good, but native-run has pio build commands and doesn't really make sense for the release assetts

madeofstown: 👌, only added that in for cases where people intend to test the build before installing.

fifieldt (approved):

PR #5417, opened 2024-11-21
Fix admin key loading from userPrefs.h

Mictronics:
* Remove static declaration for admin keys from userPrefs.h.
* Fix bug in loading key slot 2 introduced with 89469fc
* Load admin keys to slots in defined order. Don't depend on counter.
* Load hard coded admin keys in case the config file has empty slots.


src/mesh/NodeDB.cpp
```diff
@@ -888,6 +903,54 @@ void NodeDB::loadFromDisk()
...
+    // Initialize admin_key_count to zero
+    byte numAdminKeys = 0;
+    uint16_t sum = 0;
+#ifdef USERPREFS_USE_ADMIN_KEY_0
+    for (uint8_t b = 0; b < 32; b++) {
+        sum += config.security.admin_key[0].bytes[b];
+    }
+    if (sum == 0) {
+        numAdminKeys += 1;
+        LOG_INFO("Admin 0 key zero. Loading hard coded key from user preferences.\n");
```
thebentern, line 916: Can you remove the newlines on the logs?

gjelsoe:
> * Load hard coded admin keys in case the config file has empty slots.

Doesn't that mean, keys can't be removed anymore ?

gjelsoe:
And possible add
`#if defined(USERPREFS_ADMIN_KEY_0) || defined(USERPREFS_ADMIN_KEY_1) || defined(USERPREFS_ADMIN_KEY_2)`
    `saveToDisk(SEGMENT_CONFIG);`
`#endif`
so there will only be one `saveToDisk(SEGMENT_CONFIG)` and not potentially 3 in a row.

Mictronics:
> > * Load hard coded admin keys in case the config file has empty slots.
> 
> Doesn't that mean, keys can't be removed anymore ?

Correct, keys can only be changed but not deleted/zeroized.
However, I consider a hard coded key the highest priority. I mean there is a reason that they are hard coded. If keys are subject of [frequent] change they should not be stored in code.

Mictronics:
> And possible add `#if defined(USERPREFS_ADMIN_KEY_0) || defined(USERPREFS_ADMIN_KEY_1) || defined(USERPREFS_ADMIN_KEY_2)` `saveToDisk(SEGMENT_CONFIG);` `#endif` so there will only be one `saveToDisk(SEGMENT_CONFIG)` and not potentially 3 in a row.

That would mean a saveToDisk is executed on every boot as soon as at least one key is stored in code. Whether config file needs an update or not.

gjelsoe:
> > And possible add `#if defined(USERPREFS_ADMIN_KEY_0) || defined(USERPREFS_ADMIN_KEY_1) || defined(USERPREFS_ADMIN_KEY_2)` `saveToDisk(SEGMENT_CONFIG);` `#endif` so there will only be one `saveToDisk(SEGMENT_CONFIG)` and not potentially 3 in a row.
> 
> That would mean a saveToDisk is executed on every boot as soon as at least one key is stored in code. Whether config file needs an update or not.

Correct, forgot to mention that lines 920, 935 and 950 should be removed.
Then there will only be one `saveToDisk` for all 3 keys.

gjelsoe:
> > Doesn't that mean, keys can't be removed anymore ?
> 
> I mean there is a reason that they are hard coded. If keys are subject of [frequent] change they should not be stored in code.

It's also to ease the setup of a node. But this could be fixed by having a `#define KEEP_ADMIN_KEY_0` or something like it, if it should be persistent. IMHO

Mictronics:
> > > And possible add `#if defined(USERPREFS_ADMIN_KEY_0) || defined(USERPREFS_ADMIN_KEY_1) || defined(USERPREFS_ADMIN_KEY_2)` `saveToDisk(SEGMENT_CONFIG);` `#endif` so there will only be one `saveToDisk(SEGMENT_CONFIG)` and not potentially 3 in a row.
> > 
> > 
> > That would mean a saveToDisk is executed on every boot as soon as at least one key is stored in code. Whether config file needs an update or not.
> 
> Correct, forgot to mention that lines 920, 935 and 950 should be removed. Then there will only be one `saveToDisk` for all 3 keys.

That's clear. However, at the moment a saveToDisk is only executed for a key that has been transferred from code memory into config file, because that key slot was empty (`sum == 0`). [protobuf initializes/defaults to zero]
With your change proposal, after checking the three key slots in config, a saveToDisk is executed regardless of what is/was inside config. saveToDisk depends just on what is hard coded (the defined keys) but not what is actually stored in config. (the content of key slots)

gjelsoe:
> > > > And possible add `#if defined(USERPREFS_ADMIN_KEY_0) || defined(USERPREFS_ADMIN_KEY_1) || defined(USERPREFS_ADMIN_KEY_2)` `saveToDisk(SEGMENT_CONFIG);` `#endif` so there will only be one `saveToDisk(SEGMENT_CONFIG)` and not potentially 3 in a row.
> > > 
> > > 
> > > That would mean a saveToDisk is executed on every boot as soon as at least one key is stored in code. Whether config file needs an update or not.
> > 
> > 
> > Correct, forgot to mention that lines 920, 935 and 950 should be removed. Then there will only be one `saveToDisk` for all 3 keys.
> 
> That's clear. However, at the moment a saveToDisk is only executed for a key that has been transferred from code memory into config file, because that key slot was empty (`sum == 0`). [protobuf initializes/defaults to zero] With your change proposal, after checking the three key slots in config, a saveToDisk is executed regardless of what is/was inside config. saveToDisk depends just on what is hard coded (the defined keys) but not what is actually stored in config. (the content of key slots)

Your are right.
If it's not an issue having 1-3 saved then there are no need to change it.

PR #5420, opened 2024-11-22
Seems like the last DIY board that's not "extra"

jp-bennett:


tomuk5:
Hi, so now we can no longer flash any DIY ESP32 boards from the web flasher or download them as part of the precompiled Zip bundles?

tomuk5:
Please revert this change as this breaks the ability to flash DIY Boards from the web flasher tool and the last remaining DIY firmware image is now missing from the main pre-compiled zip bundles.

@jp-bennett please can you explain or provide some reasoning as to what it was you were trying to fix or improve by removing the last remaining DIY firmware?

jp-bennett:
Hi @tomuk5 . The intent is to make this consistent with our other targets. The firmware should get built for releases, just not as part of the per-commit CI run.

tomuk5:
At present there is no DIY firmware available for download as part of the latest release or available to be flashed via the web flashing tool.
After this commit to remove the building of the firmware as part of the per commit CI run, is there then other work that has yet to be completed to build the firmware for releases - as at present it’s now missing entirely.

Issue #626, opened 2021-01-04
On boards with BATTERY_PIN defined, check for missing voltage divider net


geeksville:
at power up:
enable internal pullup and see if the raw ADC value is saturated at Vcc
then enable internal pulldown and see if raw ADC value is stuck at ground

If both of these are true the voltage divider resistors must not be installed.  Disable battery sensing.

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/ttgo-lora32-v2-0-battery-level/1989/18

thebentern:
Closing

PR #5425, opened 2024-11-22
Support for the ClimateGuard RadSens Geiger-Muller tube

jake-b:
This PR is to support the [ClimateGuard RadSense dosimeter](https://github.com/climateguard/RadSens/blob/master/extras/datasheets/RadSens_datasheet_ENG.pdf).
A fun and educational sensor for Meshtastic; not for safety critical applications.

This is an I2C device which integrates nicely into the firmware along side other sensors.

This pairs with a pull [request to add a field and enumeration to the protobufs](Support for the ClimateGuard RadSens Geiger-Muller tube #626)

Changes:

- `detect/ScanI2CTwoWire.cpp - adds the ability to find the device on the I2C bus during scan
- `modules/Telemetry/Sensor/CGRadSensSensor.*` - adds support for reading the sensor
- `modules/Telemetry/EnvironmentTelemetry.cpp`- updates to handle the new sensor, also updates the display code to not show temperature/humidity unless it is present in the measurement.

Thanks for your consideration of this, my first contribution to Meshtastic


src/mesh/generated/meshtastic/telemetry.pb.h
```diff
@@ -77,7 +77,9 @@ typedef enum _meshtastic_TelemetrySensorType {
     /* MLX90614 non-contact IR temperature sensor */
     meshtastic_TelemetrySensorType_MLX90614 = 31,
     /* SCD40/SCD41 CO2, humidity, temperature sensor */
-    meshtastic_TelemetrySensorType_SCD4X = 32
+    meshtastic_TelemetrySensorType_SCD4X = 32,
+    /* ClimateGuard RadSens, radiation, Geiger-Muller Tube */
+    meshtastic_TelemetrySensorType_RADSENS = 33
```
caveman99, line 82: Please don't commit any files named 'generated' - these will usually cause merge conflicts and will be overwritten anyway. The build errors can be ignored and will be fixed by the protobuf PR:

protobufs
caveman99, line 1: Please don't check in any protobuf submdule changes. This will bomb the merge process for sure.

src/modules/Telemetry/EnvironmentTelemetry.cpp
```diff
@@ -243,6 +250,10 @@ void EnvironmentTelemetryModule::drawFrame(OLEDDisplay *display, OLEDDisplayUiSt
     if (lastMeasurement.variant.environment_metrics.weight != 0)
         display->drawString(x, y += _fontHeight(FONT_SMALL),
                             "Weight: " + String(lastMeasurement.variant.environment_metrics.weight, 0) + "kg");
+
+    if (lastMeasurement.variant.environment_metrics.radiation != 0)
+        display->drawString(x, y += _fontHeight(FONT_SMALL),
+                            "Radiation: " + String(lastMeasurement.variant.environment_metrics.radiation, 2) + "µR/h");
```
caveman99, line 256: Screen estate is valuable. maybe just Rad:

caveman99 (changes requested):
Thanks for your contribution, for some minor changes see inline comments.

GUVWAF:
Since this is a rather specific sensor and we're very close to the flash limit on nRF52 devices, I think it would be good to introduce a macro like `MESHTASTIC_EXCLUDE_NICHE_SENSORS` in order to exclude this and maybe some other sensors for that platform.

caveman99 (approved):

caveman99:
> Since this is a rather specific sensor and we're very close to the flash limit on nRF52 devices, I think it would be good to introduce a macro like `MESHTASTIC_EXCLUDE_NICHE_SENSORS` in order to exclude this and maybe some other sensors for that platform.

Do you care to have it inside of this, or can i merge it and create another PR? i wanted to introduce exclude_ethernet too...

GUVWAF:
> Do you care to have it inside of this, or can i merge it and create another PR? i wanted to introduce exclude_ethernet too...

I'm fine with merging this as is. PR looks good otherwise.

GUVWAF (approved):

sourceholder:
@jake-b, would you be able to share a screenshot of how the Geiger data is displayed in the app?  I only seen battery level, temperature and humidity logged before. I am wondering how the app handles other units.

jake-b:
@sourceholder The iOS (and I assume Android) apps do not surface this information currently.  That would be an enhancement request to the iOS app and is beyond the scope of this pull request. 

The data is surfaced on the Environment Metrics page of the node’s display. It will be shown on the display of other nodes that also run a firmware that understands the protobuf field that contains the radiation value.

Also if you are using MQTT, JSON and Home Assistant the value will appear in the JSON topic.

PR #5426, opened 2024-11-23
Cleans up visibility in GPS.h

charlieh0tel:
Cleans up visibility in GPS.h to pave the way for additional clean ups and additions (gpsd support).


charlieh0tel:
This is ready to go.

PR #5439, opened 2024-11-25
Fix memory leaks by adding missing `free()` calls before early returns in `MQTT::onReceive`

### LINKED ISSUE #5387: [Bug]: Wi-Fi Connectivity Loss on T-Beam v1.2 (Potential Memory Leak)
### Category

Other

### Hardware

T-Beam

### Firmware Version

2.5.13.1a06f88

### Observed Behavior:
My T-Beam v1.2 loses Wi-Fi connectivity after 30~90 minutes and requires a reset to reconnect.

### Additional Information:
I suspect a memory leak in the firmware as the issue seems related to memory exhaustion.

Attached are memory usage stats collected from the `/json/report` API during Wi-Fi operation.

I'm willing to help debug this issue and would appreciate instructions on how to proceed.

memory_usage_plot
[memory_usage_log.csv](https://github.com/user-attachments/files/17792974/memory_usage_log.csv)

### Relevant log output

_No response_

CTassisF:
This fix addresses memory leaks in the `MQTT::onReceive` function by ensuring that dynamically allocated resources (`e.channel_id`, `e.gateway_id` and `e.packet`) are properly freed before each early return. Previously, these resources were only freed at the end of the function, leaving them unhandled in certain exit paths. Adding the missing `free()` calls prevents memory leaks and ensures proper resource cleanup in all scenarios.

Fixes #5387

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -167,17 +167,26 @@ void MQTT::onReceive(char *topic, byte *payload, size_t length)
                     if (isFromUs(p)) {
                         LOG_INFO("Ignore downlink message we originally sent");
                         packetPool.release(p);
+                        free(e.channel_id);
+                        free(e.gateway_id);
+                        free(e.packet);
                         return;
                     }
                     if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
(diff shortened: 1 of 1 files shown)


fifieldt (approved):
Appreciated!

PR #5443, opened 2024-11-25
Don't powersave on Wifi

### LINKED ISSUE #5440: [Bug]: Llilygo T-Deck wifi dropouts due to default power settings
### Category

WiFi

### Hardware

T-Deck

### Firmware Version

2.5.7

### Description

When I updated the firmware (and doing a full erase at the same time) on my Llilygo T-Deck to 2.5.*, the option under the "Power" config called "Select if powered from a low-current source (i.e. solar), to minimize power consumption as much as possible" is turned ON by default. With most other hardware I use (RAK4631, Heltec V3 etc), this is OFF by default. This option  being on causes the Wifi to only remain connected for around 10 secs at a time, then it disconnects for 10 secs (unreachable via ping), then reconnects, continually. I think this option should be OFF by default, optionally turned on, as it took me days to work out what was going on.

### Relevant log output

_No response_

thebentern:
Powersaving sleep transition is already cancelled when bluetooth is connected. This change rectifies the behavior with Wifi. Sleep transitions will be completely removed if Wifi is enabled.

Closes #5440

### DIFF
--- a/src/PowerFSM.cpp
+++ b/src/PowerFSM.cpp
@@ -383,7 +383,7 @@ void PowerFSM_setup()
     // See: https://github.com/meshtastic/firmware/issues/1071
     // Don't add power saving transitions if we are a power saving tracker or sensor or have Wifi enabled. Sleep will be initiated through the
     // modules
-    if ((isRouter || config.power.is_power_saving && !isWifiAvailable()) && !isTrackerOrSensor) {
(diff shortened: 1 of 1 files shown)


fifieldt:
This makes sense!

fifieldt (approved):

PR #5464, opened 2024-11-27
CLIENT_HIDDDEN should never broadcast NeighborInfo

fifieldt:
We set the broadcast interval to UINT_MAX in the NodeDB setup. Howwever, in theory there would be a case where a HIDDEN node could reveal itself through this module.

This change means it will never send, and will always wait UINT_MAX before trying again if called.

### DIFF
--- a/src/modules/NeighborInfoModule.cpp
+++ b/src/modules/NeighborInfoModule.cpp
@@ -121,6 +121,9 @@ Will be used for broadcast.
 */
 int32_t NeighborInfoModule::runOnce()
 {
+    if (config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT_HIDDEN) {
+        return UINT_MAX;
+    }
     if (moduleConfig.neighbor_info.transmit_over_lora && !channels.isDefaultChannel(channels.getPrimaryIndex()) &&
         airTime->isTxAllowedChannelUtil(true) && airTime->isTxAllowedAirUtil()) {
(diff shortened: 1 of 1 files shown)


GUVWAF:
Not sure this is needed, it’s an optional module already, there are other modules that would let a `CLIENT_HIDDEN` transmit.

Also, I don’t think it addresses the point made in #5286. It’s about *another* node broadcasting the node number of a `CLIENT_HIDDEN` when it has NeighborInfo enabled with LoRa transmit. 
Anyway I don’t really see a problem with this, because when the `CLIENT_HIDDEN` transmits anything, it transmits its node number in the clear already, and it gets rebroadcasted by other nodes as well. From the docs: _“Device that only broadcasts as needed for stealth or power savings.”_, which doesn’t mean it cannot be detected in any case.

fifieldt:
Thanks for the review :) I was just looking at what DeviceTelemetryModule does and trying to make a parallel. I bow to those with more experience about whether this is something that needs to be addressed or not - this patch took very little effort :P

GUVWAF:
Yeah, DeviceTelemetryModule is enabled by default, NeighborInfoModule is not. I think if someone wants this on a `CLIENT_HIDDEN` then we should allow it, although it’s a weird choice.

fifieldt:
OK! Will wait another day in case someone else has an opinion and will abandon.

fifieldt:
Abandoning, not needed.

PR #5471, opened 2024-11-28
Convert userprefs to a json file instead of header file which has to be included everywhere

thebentern:
For your consideration is a change to the userPrefs scheme which has a more strategic direction for ad-hoc agent builds. One of the main goals of userPrefs was to enable ad-hoc builds with custom default values. However, boxing this into a header file was always limiting from a web app consumer perspective. 

In the past I attempted to bridge this gap by using clang to generate a manifest of the userprefs into a json file which we would then later attempt to tokenize and replace values from user input. That approach was problematic and still had the issue of macros not being defined unless the userPrefs header was included in every single file that required use of them. 

This new json based approach ensures availability globally by injecting them as build flags in our custom PIO script, still allows proper macro expansion, and will make consuming in an automated build application much simpler.

image


panaceya:
Nice description, but it's not clear how to use it. Please provide more examples.

userPrefs.jsonc
```diff
@@ -0,0 +1,37 @@
...
+  // "USERPREFS_EVENT_MODE": "1",
+  // "USERPREFS_FIXED_BLUETOOTH": "121212",
+  // "USERPREFS_FIXED_GPS": "",
+  // "USERPREFS_FIXED_GPS_ALT": "0",
+  // "USERPREFS_FIXED_GPS_LAT": "48.85873920",
+  // "USERPREFS_FIXED_GPS_LON": "2.294508368",
+  // "USERPREFS_LORACONFIG_CHANNEL_NUM": "31",
+  // "USERPREFS_LORACONFIG_MODEM_PRESET": "meshtastic_Config_LoRaConfig_ModemPreset_SHORT_FAST",
+  // "USERPREFS_SPLASH_TITLE": "DEFCONtastic",
+  "USERPREFS_TZ_STRING": "tzplaceholder                                         "
```
fifieldt, line 33: did you want to merge with tzplaceholder in here?

fifieldt (commented):
Looks good - just had one comment about the uncommented thing - tzplaceholder.

userPrefs.jsonc
thebentern: Yep, this is one macro we have defined already in the userPrefs.h to replace at flash time.

fifieldt (approved):

thebentern:
> Nice description, but it's not clear how to use it. Please provide more examples.

Works almost entirely the same way as the header file except it is `"MACRO": "value"` instead of `#define MACRO value` and you no longer had to include the header file. These now get injected as build flags by platformio globally

PR #5477, opened 2024-11-29
Added femtofox configs

noon92:
Adding several yaml config files for the femtofox/luckfox project to `available.d/`.
Perhaps these should be in a subdirectory? Right now there's not much in `available.d/`, but this may change.


fifieldt:
Hi! Yes, actually, a subdirectory would be great. Lead the way for the future masses!

noon92:
> Hi! Yes, actually, a subdirectory would be great. Lead the way for the future masses!

Made requested changes

PR #5481, opened 2024-11-29
Add RAK12035VB Soil Moisture Sensor support

Justin-Mann:
Introduce the RAK12035VB sensor as an environmental telemetry sensor, including necessary calibration checks and default values. Update relevant files to integrate the sensor into the existing telemetry system.

I did as good a job as I could up to this point, but some correction may be necessary.. 

I kinda had an idea of how I wanted things to go, and I think the basic detection and registration methods being used for the sensors are probably enough so I did not implement any guards specific to the RAK soil monitoring.  

This hardware is not just one module, but a couple.. RAK12023 and RAK12035 is the component stack, the RAK12023 does not seem to matter much and allows for multiple RAK12035 devices to be used.  *I did not code support for multiple RAK12035, but this could be handled in code if desired...  *I wrote very simple support to default the configuration values for the soil moisture levels.. this could be imporved, but is a complex interaction so I would suggest doing what I did until the mobile apps are enhanced to support the new environmental data points.. (perhaps something could be built into the mobile admin gui to perform a water monitor config routine with feedback). 

The main issue I have atm.. 
***********************************
in ScanI2CTwoWire.cpp ln.  347...
***********************************
I defined a case for devices with address  0x20.. I don't know where I got the values I did (probably left over from the pervious night's coding session from a copy and paste), but they do not work.. that check does not work.. I tried to run down registry uuid locations for both devices, but came up short.. and.. I'm apparently bad at this, so if anyone has more refined detection methods I could use or knowledge around this any help would be appreciated.
****


fifieldt:
Thanks for the updates and welcome to Meshtastic :)

As you've probably seen, the protobufs don't like to be checked in -- they're automatically generated, so its best to remove them from this commit. If you need help working with the git submodules let us know, they're a bit tricky :)

caveman99:
@Justin-Mann i'd really like to include this, but due to heavy lifting in the sensor code and a few unfortunate checkins this PR is now conflicted beyond repair. Do you object if i sent in a new cleaned up PR and credited you as the author?

Justin-Mann:
feel free.  I just havn't really dealt with the tests on the other software
platforms failing and have not had a ton of time to look into it more.. was
actually just thinking about restructuring a little bit and trying again in
the near future.. but if you feel like writing it correctly.. on the newest
branch.. feel free man.  I'm probably online all day today if I can help in
any way.

Thanks,
Justin
#mcluvin7930

fifieldt:
I added a PR to the protobufs repository for this sensor:

https://github.com/meshtastic/protobufs/pull/657

fifieldt:
@Justin-Mann , @caveman99 - I started the 'clean' patch at:

https://github.com/meshtastic/firmware/pull/6259

. @Justin-Mann , if you have time, perhaps you could try this detection code?

Justin-Mann:
I do and will get back with results later today. Thank you for the help and
resolution.

I have taken a couple shots at resolving this againt the newest master, but
i just couldnt get it to work.. only my old branch would function.  The
behavior with the new branches was that the device just got stuck in a
reboot loop. I appreciate the help and will test and do what i can to make
it more effective (logic to support multiple sensors). Appreciate you guys.

fifieldt:
Closing this as we are focusing on the new patch at #6529

PR #5484, opened 2024-12-01
Add generic thread module

tavdog:
Add a developer module aimed at showing how to make a simple thread module.  It can be used to have your node do an action on any interval you want.  If enabled with the MESHTASTIC_INCLUDE_GENERIC_THREAD flag it will just print "Genericthread running" to serial every 10 seconds.  Maybe somebody wants to have their node reboot every 24 hours. . . 

I'm not attached to this being added to main so if nobody thinks it's useful feel free to reject this PR. . hehe


fifieldt:
Thanks for this. I like it - nice and clear how to make a module.

thebentern:
Looks good. Since this is the only MESHTASTIC_INCLUDE macro, should we make it fall in line with the other EXCLUDE ones and add the build flag to remove it in the platformio.ini?

GUVWAF:
Was there a reason this got closed? I think it was a nice example for a module that acts periodically.

tavdog:
I'm not sure what happened. Maybe I thought it got merged so I closed it.

GUVWAF:
I see, no problem. Let's get it merged, I think the only comment left was to use `MESHTASTIC_EXCLUDE_GENERIC_THREAD` instead, and add this by default to the `platformio.ini`.

thebentern (approved):

GUVWAF (approved):

fifieldt (approved):

src/modules/Modules.cpp
```diff
@@ -131,6 +135,9 @@ void setupModules()
 
 #if !MESHTASTIC_EXCLUDE_DROPZONE
         dropzoneModule = new DropzoneModule();
+#endif
+#if MESHTASTIC_INCLUDE_GENERIC_THREAD
```
fifieldt, line 139: should this be !MESHTASTIC_EXCLUDE_GENERIC_THREAD_MODULE ?

PR #5487, opened 2024-12-01
SimRadio: clean-up and emulate collisions

GUVWAF:
This cleans up some of the guarding for the simulator. Actually all Portduino variants have `HAS_RADIO` still set to 0, but only when the SimRadio is actually initialized, we should handle the packets from the simulator. `HAS_RADIO` can now thus be set to 1 for Portduino as well, and when it doesn’t have a radio configured, it will fall back to SimRadio.

Also, the simulator can’t handle PKC packets, so this avoids using it when SimRadio is used.

Furthermore, SimRadio now emulates collisions when it receives a packet while it is currently sending or receiving another one. Those bad packets are now counted as well, and sent with the LocalStats.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -83,7 +83,7 @@ NRF52Bluetooth *nrf52Bluetooth = nullptr;
 #include "STM32WLE5JCInterface.h"
 #endif
 
-#if !HAS_RADIO && defined(ARCH_PORTDUINO)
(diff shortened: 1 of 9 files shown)


GUVWAF:
Converted to draft for now, would probably be nice to be able to disable the emulation of collisions.

PR #5495, opened 2024-12-03
Add Heltec Mesh Node T114 No Screen board definition

Heltec-Aaron-Lee:
Add Heltec Mesh Node T114 No Screen board definition.

The default BlueTooth connection pin code is `123456`

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -38,6 +38,7 @@ default_envs = tbeam
 ;default_envs = heltec_vision_master_e213
 ;default_envs = heltec_vision_master_e290
 ;default_envs = heltec_mesh_node_t114
+;default_envs = heltec_mesh_node_t114_noscreen
 extra_configs =
 	arch/*/*.ini
 	variants/*/platformio.ini
--- a/variants/heltec_mesh_node_t114_noscreen/platformio.ini
+++ b/variants/heltec_mesh_node_t114_noscreen/platformio.ini
@@ -1,0 +1,17 @@
+; First prototype nrf52840/sx1262 device
+[env:heltec-mesh-node-t114-noscreen]
+extends = nrf52840_base
+board = heltec_mesh_node_t114
+debug_tool = jlink
+
+# add -DCFG_SYSVIEW if you want to use the Segger systemview tool for OS profiling.
+build_flags = ${nrf52840_base.build_flags} -Ivariants/heltec_mesh_node_t114_noscreen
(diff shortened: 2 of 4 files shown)


lyusupov:
Aaron, what's up with Heltec T114 **V1** replacement program ?

I've made a request via AliExpress on October 13th and the Hectec Seles personnel responded that the request was accepted.
Today is December 3rd, few days before completion of 2 month period, but the replacement (1 piece of T114 V2) is still NOT shipped out.

mverch67:
IMO this PR is a bit like shooting sparrows with cannons. There is no need for another variant as it is likely that the variant could be requested at runtime. The ST7789 display driver supports SDA+DC readings of the driver ID. 

I did run a slightly modified sketch of [this](https://github.com/prenticedavid/ST7789_kbv/blob/master/examples/ST7789_readreg_diag/ST7789_readreg_diag.ino) with the heltec T114 and got the following result:

`read reg(4):  ID = 0x8181B3`

Heltec-Aaron-Lee:
@mverch67 

Due to the inability of the screenless version of T114 to display the Bluetooth connection pin code on the display, this update has been submitted in response to numerous user requests.

Heltec-Aaron-Lee:
Hi @lyusupov 

Very sorry about that. I just confirmed with my colleague that all reissues will be completed within this week.

thebentern:
> Due to the inability of the screenless version of T114 to display the Bluetooth connection pin code on the display, this update has been submitted in response to numerous user requests.

I believe Manuel is suggesting we might be able to probe those registers and detect presence of the display or not to set the Bluetooth pin mode default config.

Heltec-Aaron-Lee:
Haha @thebentern 

Indeed verification via register is the best way, but adding a new hardware model can quickly solve many of the current user problems, and our upcoming `nRF+SX1262` module can also use this model, making it convenient for more people to DIY their own Meshtastic hardware

mverch67:
@Heltec-Aaron-Lee 
Haha, no objections to leave this PR open for users who want a quick-and-dirty solution and use it for their DIY compilation of this firmware for the no_screen variant or the upcoming nrf+SX162 module.

I compared the code with the existing variant and found only one single change:
image

Copy and paste of the entire variant because of this single change obviously doesn't make much sense and will also confuse the user with so many variants. So I think we'd need another PR that queries the ST7789 driver at runtime and sets a corresponding flag to set the initial bluetooth password.

thebentern:
We could also just create a new environment target in PIO with no new HW_MODEL and define the flags in the ini instead of the variant, if the intent is to have a variant without a screen which is more DIY. Presumably a user could add an I2C based oled display as well

PR #5506, opened 2024-12-04
Add MACAddress to config.yaml

jp-bennett:
Allows setting a virtual MAC Address in config.yaml

### DIFF
--- a/src/platform/portduino/PortduinoGlue.cpp
+++ b/src/platform/portduino/PortduinoGlue.cpp
@@ -69,6 +69,22 @@ void portduinoCustomInit()
     portduinoAddArguments(child, childArguments);
 }
 
+void getMacAddr(uint8_t *dmac)
+{
+    if (settingsStrings[mac_address].length() > 11) {
+        dmac[0] = std::stoi(settingsStrings[mac_address].substr(0, 2), nullptr, 16);
+        dmac[1] = std::stoi(settingsStrings[mac_address].substr(2, 2), nullptr, 16);
+        dmac[2] = std::stoi(settingsStrings[mac_address].substr(4, 2), nullptr, 16);
+        dmac[3] = std::stoi(settingsStrings[mac_address].substr(6, 2), nullptr, 16);
+        dmac[4] = std::stoi(settingsStrings[mac_address].substr(8, 2), nullptr, 16);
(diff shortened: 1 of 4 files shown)


thebentern (approved):

jp-bennett:
> Humans are generally not good at generating random numbers, I'm afraid this would higher the chance of `deadbeef` node number collisions even more.
> 
> We could probably also try getting an Ethernet MAC if there's no Bluetooth, right?

Honestly I was just copying from my local Ethernet device. We could encourage that, or maybe even automate it.

GUVWAF:
Yeah, I think that would be nice. 
Or, even when there's no Bluetooth/Ethernet, generate a random one and write it to the config file (in case you do a "factory reset").

jp-bennett:
> Yeah, I think that would be nice. Or, even when there's no Bluetooth/Ethernet, generate a random one and write it to the config file (in case you do a "factory reset").

OK, I'll see about adding another option, to specify the network device to pull the MAC from. Should we consider the blank mac address an error condition? I'm thinking maybe yes, to prevent the proliferation of bogus devices.

GUVWAF:
Sounds good, and yes, I think that should generate an (easily interpretable) error.

src/platform/portduino/PortduinoGlue.cpp
```diff
@@ -137,6 +163,13 @@ void portduinoSetup()
         }
     }
 
+    uint8_t dmac[6];
+    getMacAddr(dmac);
+    if (dmac[0] == 128 && dmac[1] == 0 && dmac[2] == 0 && dmac[3] == 0 && dmac[4] == 0 && dmac[5] == 1) {
+        std::cout << "*** Blank MAC Address not allowed! " << std::endl;
```
GUVWAF, line 169: What about: `"MAC address should be set in 'config.yaml'"`? Then it's easier to figure out how to fix it.

jp-bennett: Yes. Still a WIP. I'll finalize the error message once the new options are available.

jp-bennett:
Actually, is there a good reason we have the getMacAddr() function in portduino at all? That's not a standard Arduino function. I'm very much inclined to just pull the whole thing into the portduinoGlue file. Makes it all way easier

GUVWAF:
I think it would indeed by fine to remove it, it's indeed not an Arduino function.

Issue #5508, opened 2024-12-04
[Bug]: Node rebooting from receiving invalid packet


liquidraver:
### Category

Other

### Hardware

T-Echo, Station G2

### Firmware Version

2.5.15

### Description

Opening a new bugreport, because I' not sure it's related to #5458 or #5498.
My T-Echo and my Station G2 is randomly rebooting, they are never alive for 24hrs.
The reboot happens usually after some hours uptime when I make a traceroute.

### Relevant log output

```Shell
.[36m.[0m..[34m.DEBUG .[0m| 20:08:21 15484 [RadioIf] .[34m.Started Tx (id=0x3ef60d7e fr=0xac to=0x90, WantAck=0, HopLim=0 Ch=0x8 encrypted rxtime=1733339301 priority=120).[0m
.[36m.[0m..[34m.DEBUG .[0m| 20:08:21 15484 [RadioIf] .[34m.Packet TX: 452ms.[0m
.[36m.[0m..[34m.DEBUG .[0m| 20:08:22 15484 [RadioIf] .[34m.Completed sending (id=0x3ef60d7e fr=0xac to=0x90, WantAck=0, HopLim=0 Ch=0x8 encrypted rxtime=1733339301 priority=120).[0m
.[36m.[0m..[34m.DEBUG .[0m| 20:08:28 15490 [Power] .[34m.Battery: usbPower=1, isCharging=1, batMv=4554, batPct=100.[0m
.[36m.[0m..[34m.DEBUG .[0m| 20:08:29 15491 [RadioIf] .[34m.Lora RX (id=0xc101d05c fr=0x90 to=0xac, WantAck=1, HopLim=2 Ch=0x8 encrypted rxSNR=6.5 rxRSSI=-38 hopStart=2).[0m
.[36m.[0m..[34m.DEBUG .[0m| 20:08:29 15491 [RadioIf] .[34m.Packet RX: 460ms.[0m
.[36m.[0m..[34m.DEBUG .[0m| 20:08:29 15491 [Router] .[34m.Found existing packet record for fr=0xSTATIONG2TRACEDNODE,to=0xTECHOWHOSTARTEDTHETRACE,id=0xc101d05c.[0m
[...]

PR #5509, opened 2024-12-04
Don't use channel index for encrypted packet

GUVWAF:
Temporary fix for #5508.

During filtering the packet is not yet decrypted, and thus the channel is still the hash, not an index. As a temporary solution, always use the primary channel for sending the ACK. This will be properly fixed with the NextHopRouter.


thebentern (approved):

GUVWAF:
Let's also remove the assert in `getKey()` that actually caused the reboot.

fifieldt (approved):

PR #5511, opened 2024-12-05
add ignore_direct radio testing option

Jorropo:
This allows to exercise the meshing at a hardware level while the devices are meters appart and I find that useful while testing.
https://github.com/meshtastic/protobufs/pull/631


Jorropo:
I hadn't the chance to try the feature yet because after 5 minutes I didn't managed to regen the protos in the python cli yet but the code is very simple and I can fix it if I see later it doesn't work as it should.

Jorropo:
https://github.com/meshtastic/protobufs/pull/631#issuecomment-2520625615

PR #5515, opened 2024-12-06
Added support for the LR1121 radio to the NRF52 Pro-Micro

Nestpebble:
Added support for the LR1121 radio, tested as functional with an E80-900M2213S from CDEbyte.


Nestpebble:
@markbirss I tried it and it works.
image

markbirss:
> Added support for the LR1121 radio, tested as functional with an E80-900M2213S from CDEbyte.

I have also myself tried other values for "#define LR11X0_DIO3_TCXO_VOLTAGE 1.6"

1.8 and 3.0

lyusupov:
[**Ebyte reference source code**](https://github.com/KunYi/EBYTE_E80_DEMO/tree/cc19a4a0f4c60681477bd31522d41206598e9e21) uses 1.8 V of TCXO power supply voltage.

image

lyusupov:
## E80-xxxM2213S_UserManual_EN_v1.0

DIO7 is a participant of RF switch matrix.

image

markbirss:
> ## E80-xxxM2213S_UserManual_EN_v1.0
> DIO7 is a participant of RF switch matrix.
> 
> image

But the E80 don't have a additional connection for it ?

Nestpebble:
> ## E80-xxxM2213S_UserManual_EN_v1.0
> DIO7 is a participant of RF switch matrix.
> 
@caveman99 Can you comment on this? I'm cribbing your content here...

caveman99:
Not familiar with the actual E8 design here, but if you don't wanna use GNSS scanning the DIO7 can be left undefined. For good measure i would define it as it is switched to the right position by the lib then for sure. In the rfswitch.h simply add a 3d switch column for DIO7 that is always false except in the GNSS case.

caveman99 (approved):

Nestpebble:
@lyusupov thanks for pointing me in that direction. This E80 module (at least) is based on the lr1121, and gnss and WiFi scan are only available on the lr1110 and lr1120. Nevertheless, I will add the additional switch as suggested by @caveman99, so that if such a feature is added in future, or a different module is used, it won't be activated by mistake.

thebentern (approved):

PR #5522, opened 2024-12-07
Improve Lost and Found

### LINKED ISSUE #5514: Improve lost and found mode
### Platform

NRF52, ESP32, RP2040

### Description

I think there are several ways lost and found mode could be improved:

- Reduce power in lost and found mode
- Enable precise location to avoid sending imprecise locations
- Reduce battery usage (Could stop forwarding packets)
- Allow position requests
- Make the node emit RF in a way to make it easier to direction find by making it a unique signal. (Doesn’t have to be Lora)

fifieldt:
As suggested by @RCGV1, this patch sets LOST_AND_FOUND role to
* Broadcast full precision by default
* Save battery by not re-broadcasting

Fixes https://github.com/meshtastic/firmware/issues/5514

### DIFF
--- a/src/mesh/Channels.cpp
+++ b/src/mesh/Channels.cpp
@@ -134,7 +134,11 @@ void Channels::initDefaultChannel(ChannelIndex chIndex)
     channelSettings.psk.bytes[0] = defaultpskIndex;
     channelSettings.psk.size = 1;
     strncpy(channelSettings.name, "", sizeof(channelSettings.name));
-    channelSettings.module_settings.position_precision = 13; // default to sending location on the primary channel
+    if (config.device.role == meshtastic_Config_DeviceConfig_Role_LOST_AND_FOUND) {
+        channelSettings.module_settings.position_precision = 32; // lost and found defaults to full precision
+    } else {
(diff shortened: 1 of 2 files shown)


garthvh:
We are still waiting on more clarity on broadcasting full precision on the unencrypted channels or known keys, and we need to have a mqtt data policy in place before adding any full precision features.

b8b8:
> As suggested by @RCGV1, this patch sets LOST_AND_FOUND role to
> 
>     * Broadcast full precision by default
> 
>     * Save battery by not re-broadcasting
> 
> 
> Fixes #5514

does it also make sense to increase the default position transmission interval to say 5min for this role, regardless of the previous setting and smart position settings?

fifieldt:
> does it also make sense to increase the default position transmission interval to say 5min for this role, regardless of the previous setting and smart position settings since the device would probably not be moving?

It's already 5mins with smart position off, I believe...

b8b8:
ok cool, ive never actually used lost and found mode yet so looks like its already covered.

fifieldt:
Any updates on the data policy things?

PR #5532, opened 2024-12-09
Update .gitmodules to specify device-ui branch

j6k4m8:
@mverch67 @mariotti this addresses the comment I made [here](https://github.com/meshtastic/firmware/pull/3259/files#r1875106777) about setting a default git submod branch. 

Feel free to use or discard as you like!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,7 @@
 [submodule "lib/device-ui"]
 	path = lib/device-ui
 	url = https://github.com/meshtastic/device-ui.git
+	branch = master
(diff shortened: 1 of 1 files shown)


mverch67:
Doesn't make sense at this point as it must be removed before this branch is merged into master. Also, hiding such setting somewhere in an invisible subfolder is somewhat confusing if someone wants the normal git submodule behavior.

I'd rather suggest to install the github extensions in visual code:
image

Then the desired checkout as well as pull/push can be easily performed with the click of a button:
image

PR #5533, opened 2024-12-09
Add screen detection function

Heltec-Aaron-Lee:
Add screen detection function for Heltec Mesh Node T114 board.

- If T114 has a screen, the BlueTooth connection pin code will be displayed on the screen.
- If T114 is without a screen, the BlueTooth connection pin code will be set to `123456` by default.


caveman99:
@Heltec-Aaron-Lee thanks for the submission, is it ok if i take this code and refactor it a bit?

thebentern (approved):

fifieldt:
Let's get this in first and refactor afterward.

fifieldt (approved):

PR #5534, opened 2024-12-09
If a packet is heard multiple times, rebroadcast using the highest hop limit

erayd:
Sometimes a packet will be in the TX queue waiting to be transmitted, when it is overheard being rebroadcast by another node with a higher remaining hop limit. When this occurs, modify the pending packet in the TX queue to avoid unnecessarily wasting hops.

This is intended to help solve the problem of a router node with excellent coverage (A) hearing a packet bounce around between several poorly-placed nodes while waiting for the opportunity to rebroadcast. On a busy mesh, this can mean that the first copy of the packet that (A) hears may have had its hop limit prematurely reduced, and a better version is subsequently received.

This is more relevant to the late-rebroadcast scenario implemented in https://github.com/meshtastic/firmware/pull/5528, but is still applicable even without that mode, particularly when having to contend with poorly-placed rooftop routers.

For example:
1. (E) transmits a packet with hop limit 3
2. (B) & (D) both overhear it. (D) starts rebroadcasting it first with hop limit 2, (B) overhears (D) and waits.
3. (B) and (C) are both ready to transmit. (C) happens to go first, rebroadcasting with hop limit 1.
4. (A) hears the packet from (C), and is about to rebroadcast with hop limit 0, when it hears (B) starting to send.
5. (B) rebroadcasts the packet with hop limit 2.
6. (A) hears the packet again from (B), this time with hop limit 2.
7. (A) adjusts the hop limit on the packet in its TX queue from 0 to 1, and then transmits the packet.

example


fifieldt:
This makes sense to me

GUVWAF:
Implementation looks good, but I'm not entirely sure about the idea behind it, as it effectively means you're artificially increasing the hop limit for certain branches of the mesh. Not sure this is necessarily a bad thing, as it wouldn't happen if the optimal route was chosen in the first place, but it does go against the idea of ​​limiting the amount of rebroadcasts by setting a proper hop limit.

Also, a small issue with this is that a traceroute result will show a route with more hops than the `hops_away` accompanying the packet, which is also used to add “unknown” nodes to the route.

erayd:
> I'm not entirely sure about the idea behind it, as it effectively means you're artificially increasing the hop limit for certain branches of the mesh. Not sure this is necessarily a bad thing, as it wouldn't happen if the optimal route was chosen in the first place, but it does go against the idea of ​​limiting the amount of rebroadcasts by setting a proper hop limit.

The way I see it, is that it is *correcting* an unintended (and suboptimal) lowering of hop count for those sections. Two different ways of looking at the same situation perhaps? This wouldn't change the number of rebroadcasts in that area, but it would allow the packet to ultimately travel further, rather than getting lost between high sites due to routing shenanigans at the origin area.

> Also, a small issue with this is that a traceroute result will show a route with more hops than the hops_away accompanying the packet, which is also used to add “unknown” nodes to the route.

Thanks for flagging this. I do not yet understand how traceroute is implemented (have yet to investigate that part of the source), so will dig into that to ensure it is treated properly.

GUVWAF:
> This wouldn't change the number of rebroadcasts in that area

For this specific packet, it does (otherwise this won't have any effect). 

> Two different ways of looking at the same situation perhaps? 

I guess it depends on whether users are already accounting for this or not. Even with this solution, consuming hops too early might still happen if the packet already left the Tx queue, or if the first packet arrived with a hop limit of 0. So, if users set their hop limit based on the maximum they have seen, this only leads to more unnecessary rebroadcasts. Only if they set it exactly to the one needed for the optimal route, then this would indeed higher the success rate without any additional unnecessary rebroadcasts.

GUVWAF:
> I do not yet understand how traceroute is implemented 

It will be called upon as a module, and then modifies the payload, before adding it to the Tx queue again. When you see a duplicate packet, it will be filtered out and thus the modules are not called for it, so the original traceroute result will be in the Tx queue.
I've struggled with this before, since you need to be careful with handling duplicate packets multiple time, for example because it shouldn't be sent to the phone and MQTT twice.

erayd:
> For this specific packet, it does (otherwise this won't have any effect).

Can you clarify? I'm not seeing where the extra rebroadcast is in the local area. The only additional rebroadcast should be out of the problem area, by whichever relaying device would otherwise have seen a hop limit of 0 (in this case, A).

The fundamental question is, do you see this behaviour as desirable or undesirable in the general case? If the latter, would you be more comfortable if I implemented this as part of #5528, to apply *only* to `ROUTER_LATE`?

I think it's definitely warranted in the `ROUTER_LATE` case, because those nodes are *intended* to be serving isolated pockets of nodes that do a poor job amongst themselves first. But with other roles, the usefulness is much more marginal.

GUVWAF:
It might just be we’re not agreeing on what the “(local/problem) area” is. In my opinion it doesn’t really matter where the additional rebroadcasts happen, if they are unnecessary then it should be avoided. It is unnecessary when an end-node rebroadcasts even though there is no new receiver, but it is especially relevant for direct messages, where most nodes do not need to receive the packet at all. For example here, if node 0 sends a DM to node 1 with a hop limit of 2, it also goes to node 2 and 3, and ends up with a hop limit of 0 at node 4. However, when it arrives later via 5, with this PR node 4 would still rebroadcast.

image


> do you see this behaviour as desirable or undesirable in the general case?

If users are conservative in setting the hop limit, then it would be desirable, but in my experience this is mostly not the case. 

> I think it's definitely warranted in the ROUTER_LATE case, because those nodes are intended to be serving isolated pockets of nodes that do a poor job amongst themselves first. 

The `ROUTER_LATE` should serve isolated pocket of nodes that did not hear the packet yet. The first packet that arrives at the `ROUTER_LATE` should not be coming from those isolated pockets of nodes that consume too many hops.

> But with other roles, the usefulness is much more marginal. 

I agree this will happen more frequently if there are `ROUTER_LATE` roles in the mesh. However, it will then also affect `ROUTER`s and `REPEATER`s, because even when the route via a `ROUTER_LATE` uses less hops, its packet likely arrives later than a suboptimal route. 

Maybe we can come to a compromise and only introduce this after `ROUTER_LATE` and the NextHopRouter (#2856) are introduced? With the NextHopRouter unnecessary relaying for DMs will already be limited, while using this PR, for broadcasts the reachability might in some cases be improved.

erayd:
> It might just be we’re not agreeing on what the “(local/problem) area” is.

I think that may well be the case. I'm looking at this from the principle of *"The packet should be spread as widely as possible across the mesh with the fewest number of hops"*. In that context, what I consider to be unnecessary rebroadcasts are those generated by clients who missed seeing a prior nearby repetition of the packet (e.g. due to terrain obstructions). Whereas you seem to be coming at it from the perspective of *"Users are probably trying to reach only people very close to them, and packets should be discouraged from travelling longer distances."* (unless I have badly misunderstood your example scenario in the above post).

> In my opinion it doesn’t really matter where the additional rebroadcasts happen, if they are unnecessary then it should be avoided.

Agreed. However, currently we do not have a way to divine the user's intent - they might be trying to reach node 1, but they may also be wanting to reach node 6, or perhaps even something further away than that which would require relaying via (6). THis is particularly the case with broadcast packets (e.g. if (6) and a closer node share a common channel, the user may desire them *both* to receive the packet).

With DMs, it's more straightforward - the destination node ID is in the packet header. Albeit that we don't (yet) have a mechanism to determine where on the network that node might be located, and which hops may be necessary to reach it. Looking forward to #2856 for that!

> > do you see this behaviour as desirable or undesirable in the general case?
> 
> If users are conservative in setting the hop limit, then it would be desirable, but in my experience this is mostly not the case.

Is it not that users bump the hop limit up *because* they are struggling to reach their target otherwise? I know that is certainly the case at the edges of our local mesh here... distant users increase their hopcount, because otherwise their packets don't reliably make it through the mess of mountains & hills to the person 400km away that they're trying to communicate with.

> > I think it's definitely warranted in the ROUTER_LATE case, because those nodes are intended to be serving isolated pockets of nodes that do a poor job amongst themselves first.
> 
> The `ROUTER_LATE` should serve isolated pocket of nodes that did not hear the packet yet. The first packet that arrives at the `ROUTER_LATE` should not be coming from those isolated pockets of nodes that consume too many hops.

This makes no sense. `ROUTER_LATE` does not exist as a one-way thing to only *deliver* packets to those pockets of users. For it to properly serve those users, also exists to *receive* packets from them and relay those out to the rest of the mesh. In many cases, due to the topography, that node will be the only path available for such traffic to take.

> > But with other roles, the usefulness is much more marginal.
> 
> I agree this will happen more frequently if there are `ROUTER_LATE` roles in the mesh. However, it will then also affect `ROUTER`s and `REPEATER`s, because even when the route via a `ROUTER_LATE` uses less hops, its packet likely arrives later than a suboptimal route.

By the time the `ROUTER_LATE` version of a packet arrives, anything with a `ROUTER` or `REPEATER` role would have already rebroadcast that packet unless `ROUTER_LATE` is the first time it was received. This optimisation is only relevant to packets that are *already* sitting in the TX queue; it doesn't retain them there to wait for better options. It's purely opportunistic. Hence why I say the benefits are more marginal for other roles, because with other roles things get flushed out of the TX queue much more rapidly. The sooner they move, the less this helps.

> Maybe we can come to a compromise and only introduce this after `ROUTER_LATE` and the NextHopRouter (#2856) are introduced? With the NextHopRouter unnecessary relaying for DMs will already be limited, while using this PR, for broadcasts the reachability might in some cases be improved.

I'd be OK with that 🙂. It's a feature I'm keen to have available on our local mesh sooner than that - because it's clear from experience that we definitely do need it - but we can always just use custom binaries on our sites where that matters, so waiting until after #2856 and #5528 are merged won't cause us any problems.

erayd:
> Anyway, let's merge this after ROUTER_LATE and the NextHopRouter (and hopefully after fixing the traceroute issue).

Sounds good. And yes, I agree that the traceroute issue you've identified should be resolved before this is merged 🙂

src/mesh/RadioLibInterface.cpp
```diff
@@ -336,6 +336,18 @@ void RadioLibInterface::startTransmitTimerSNR(float snr)
     }
 }
 
+/**
+ * Clamp the hop limit to the greater of the hop count provided, or the hop count in the queue
+ */
+void RadioLibInterface::clampHopsToMax(NodeNum from, PacketId id, uint32_t hop_limit)
```
GUVWAF, line 342: Can you add this function to the SimRadio also?

erayd: I'm not familiar with the SimRadio part of the codebase yet (or at least, I don't recognise the name), but assuming there are no show-stoppers, yes 🙂.

GUVWAF: It's used with https://github.com/meshtastic/Meshtasticator?tab=readme-ov-file#interactive-simulator to simulate multiple nodes on a Linux system (or Docker container).

It should just work if you simply copy this function :)

GUVWAF: Or, maybe better to implement it in `RadioInterface` directly.

erayd: Does `RadioInterface` have access to the TX queue? I didn't think it did... that's why I put this in `RadioLibInterface` in the first place.

GUVWAF: Yeah, sorry, it doesn't. Indeed it has to be in `RadioLibInterface` and then unfortunately needs to be copied to `SimRadio`.

erayd:
@caveman99 Please don't force-push my branches. No issue with you keeping it up to date if you want to, but unnecessarily breaking sync is irritating. Merge commits work just fine.

garthvh:
> @caveman99 Please don't force-push my branches. No issue with you keeping it up to date if you want to, but unnecessarily breaking sync is irritating. Merge commits work just fine.

This is our process, having one branch managed differently than the others is not really going to work.

erayd:
> This is our process, having one branch managed differently than the others is not really going to work.

#5528 indicates otherwise, as you folks have been using merge commits there without issue. And on several other PR branches, including some of them by @caveman99. So using merge commits is clearly accepted practice here.

Why is doing things in a way that doesn't break sync OK for those branches, but not this one?

caveman99:
Preferred is rebase, sometimes you can't rebase, then we use merge commits.

erayd:
@caveman99 

> Preferred is rebase, sometimes you can't rebase, then we use merge commits.

Roger. Looking at the PR list though, I'm still seeing a fair bit of merge activity in cases where a rebase would have worked fine (including my initial example of #5528) - so it doesn't seem that this preference is being adhered to in practice across the board. If that is something you want to insist on though, then I will just put up with the annoyance as the price of engaging with this project.

Rebase-first-by-maintainers for regularly keeping every PR updated on a repo this busy does seem an odd way of doing things IMO. Given the frequency of updates, you'd be breaking things on a near-constant basis. If this preference is something you want to insist on, might I suggest that you mention it in the default notes when somebody creates a PR? Currently it just requests that editing by maintainers be enabled, but makes no mention that one's work might get repeatedly rebased without warning. That is a useful thing to know upfront, especially for PRs that may involve a notable number of distinct commits between pushes.

erayd:
@GUVWAF What do you think of the new approach as a way of resolving the traceroute issue?

It seemed a bit dirty IMO to put exceptions in there for traceroute, so now it simply reprocesses the packet if it arrives again with a better hop limit while still in the TX queue (instead of modifying the hop limit of the existing queued version).

src/mesh/FloodingRouter.cpp
```diff
@@ -29,6 +29,13 @@ bool FloodingRouter::shouldFilterReceived(const meshtastic_MeshPacket *p)
             // cancel rebroadcast of this message *if* there was already one, unless we're a router/repeater!
             if (Router::cancelSending(p->from, p->id))
                 txRelayCanceled++;
+        } else if (iface && p->hop_limit > 0) {
+            // If we overhear a duplicate copy of the packet with more hops left than the one we are waiting to
+            // rebroadcast, then remove the packet currently sitting in the TX queue and use this one instead.
+            if (iface->removePendingTXPacket(getFrom(p), p->id, p->hop_limit - 1)) {
```
erayd, line 35: Note to self: need to also catch packets that arrived initially with `hop_limit` of `0`, and then subsequently with something higher.

As currently implemented, such packets will never hit the TX queue (because they weren't eligible for rebroadcast the first time around), and therefore will never be noticed as eligible for reprocessing.

GUVWAF:
@erayd Unfortunately this won't work as it will lead to duplicate packets in the apps and on MQTT as I mentioned above. Even if it's still in the Tx queue, it has already been sent to the phone here: 
https://github.com/meshtastic/firmware/blob/b2808063d0488bcd2a2c3b17aa134ecf865513eb/src/modules/RoutingModule.cpp#L32
and to MQTT here: 
https://github.com/meshtastic/firmware/blob/b2808063d0488bcd2a2c3b17aa134ecf865513eb/src/mesh/Router.cpp#L280

erayd:
@GUVWAF Roger, and thanks. Will keep thinking...

foch:
Hello, I opened a ticket recently that matches this implementation.
https://github.com/meshtastic/firmware/issues/7572
Is there any development on-going on your side? I can try to help, but I've not quite understood the traceroute issue.

erayd:
> Is there any development on-going on your side?

I had forgotten I had this one sitting there! Will try to pick it back up again soon.

> I can try to help, but I've not quite understood the traceroute issue.

The issue is that a traceroute response arriving later with more hops left on it contains different data than the packet which is currently sitting in the queue. So it's not enough to simply modify the hop limit of the already queued packet (as we do with other packets), but rather the traceroute packet needs to be re-processed or modified accordingly.

src/mesh/FloodingRouter.cpp
h3lix1: This was fun. Fixed, but required adding max_hops to seen recently.

h3lix1:
## Summary
- Rework hop-limit upgrade handling so routers swap in better packets without losing duplicate suppression.
- Use packet history to save the best hop limit seen (This will increase memory 3 bytes/packet)
- Improve TX queue tooling/logging to make it obvious when and why entries are replaced

## Details
### Router hop-limit flow
- Refactored `FloodingRouter::shouldFilterReceived()` and `NextHopRouter::shouldFilterReceived()` to call `wasSeenRecently()` once, capture the new `wasUpgraded` flag, and run upgrade replacement before the duplicate early-return. This lets the routers reuse the upgraded frame while still filtering true dupes.
- Gate the upgrade path on `IS_ROUTER_ROLE()` and require `hop_limit > 0`, then call `removePendingTXPacket()` with the new packet’s hop limit so only worse queued copies are removed. Log messages now report the threshold being applied.

### Packet history tracking
- Extended `PacketHistory::wasSeenRecently()` with an optional `wasUpgraded` out-parameter and persisted the maximum hop limit observed for each packet. Seeing a higher hop count sets `wasUpgraded`, suppresses the duplicate fast path, and keeps the higher value even if later retransmissions arrive with fewer hops.
- Updated `PacketRecord` metadata and comments to reflect the new “highest hop limit”

### TX queue utilities and logging
- Added `MeshPacketQueue::getPacketFromQueue()` so queue lookups can reuse the same search logic, and simplified `find()` to delegate to it.
- Marked `RadioLibInterface::removePendingTXPacket()` as an override and switched the diagnostic to log packet IDs in hex, matching other debugging output.

## Testing
- Manual log inspection while passing upgraded hop-limit traffic using a production node set to ROUTER_LATE.

### Example logs

```
DEBUG | 15:55:41 190 [Router] Packet History - Hop limit upgrade: packet 0x6a2f0d5d from hop_limit=1 to hop_limit=2
DEBUG | 15:57:45 315 [Router] Packet History - Hop limit upgrade: packet 0x69051dd8 from hop_limit=0 to hop_limit=1
DEBUG | 15:57:48 318 [Router] Packet History - Hop limit upgrade: packet 0x8f299f10 from hop_limit=1 to hop_limit=2
DEBUG | 15:58:36 366 [Router] Packet History - Hop limit upgrade: packet 0x0424a99c from hop_limit=0 to hop_limit=2
```
```
DEBUG | 15:55:41 190 [Router] Dropping pending-TX packet 0x6a2f0d5d with hop limit 0
DEBUG | 15:57:48 318 [Router] Dropping pending-TX packet 0x8f299f10 with hop limit 0
DEBUG | 16:06:25 834 [Router] Dropping pending-TX packet 0x60a989c2 with hop limit 3
```

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Station G2

h3lix1:
@GUVWAF Can you assign this to me so I can submit for review? Thank you.

thebentern:
> @GUVWAF Can you assign this to me so I can submit for review? Thank you.

Sadly github doesn't allow assignments to people not in the organization

erayd:
> @GUVWAF Can you assign this to me so I can submit for review? Thank you.

I have hit the review button. Can't assign it, but I can certainly click that on your behalf 🙂

erayd:
Ignore the force-commit sorry. I have force-reversed it, and all should be well again.

Was forcing a different branch for a different feature, and accidentally pushed my old work over the top of this one at the same time.

erayd:
...twice! 😭

Have fixed it again, and updated my local copy of it to stop that happening a third time. Apologies for the notification spam.

src/mesh/FloodingRouter.cpp
```diff
@@ -21,10 +21,33 @@ ErrorCode FloodingRouter::send(meshtastic_MeshPacket *p)
 
 bool FloodingRouter::shouldFilterReceived(const meshtastic_MeshPacket *p)
 {
-    if (wasSeenRecently(p)) { // Note: this will also add a recent packet record
+    bool wasUpgraded = false;
+    bool seenRecently =
+        wasSeenRecently(p, true, nullptr, nullptr, &wasUpgraded); // Updates history; returns false when an upgrade is detected
+
+    // Handle hop_limit upgrade scenario for routers
+    if (wasUpgraded && IS_ROUTER_ROLE() && iface && p->hop_limit > 0) {
```
foch: I don't understand why you would limit this logic to Routers. It can be valuable for all nodes.
Let's take the example described in the original schema of the PR.
Imagine Node A,B and E are Clients, and Note C and D are poorly placed Routers

The wait time is much lower on Routers than Clients, so message goes quickly from E to D, then to C and to A.
It takes more time to go from E to B (long wait time at B level) and then is forwarded to A which gets it a second time.
Then, if A is a Client and not a Router, if we apply the logic to Routers only it will not update the message in the queue, and will not be transmitted.

See also the issue I opened https://github.com/meshtastic/firmware/issues/7572, before finding that a related PR (this one) was already created.

src/mesh/FloodingRouter.cpp
```diff
@@ -21,10 +21,33 @@ ErrorCode FloodingRouter::send(meshtastic_MeshPacket *p)
...
+    }
+
+    if (seenRecently) {
         printPacket("Ignore dupe incoming msg", p);
         rxDupe++;
 
+        // Handle ROUTER_LATE specific logic. Do not send to the regular queue.
+
+        if (config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER_LATE && iface) {
+            iface->clampToLateRebroadcastWindow(getFrom(p), p->id);
```
GUVWAF: This is already handled in `perhapsCancelDupe()` later on.

src/mesh/FloodingRouter.cpp
```diff
@@ -21,10 +21,33 @@ ErrorCode FloodingRouter::send(meshtastic_MeshPacket *p)
...
+    // Handle hop_limit upgrade scenario for routers
+    if (wasUpgraded && IS_ROUTER_ROLE() && iface && p->hop_limit > 0) {
+        // wasSeenRecently() reports false in upgrade cases so we handle replacement before the duplicate short-circuit
+        // If we overhear a duplicate copy of the packet with more hops left than the one we are waiting to
+        // rebroadcast, then remove the packet currently sitting in the TX queue and use this one instead.
+        uint8_t dropThreshold = p->hop_limit; // remove queued packets that have fewer hops remaining
+        if (iface->removePendingTXPacket(getFrom(p), p->id, dropThreshold)) {
+            LOG_DEBUG("Processing upgraded packet 0x%08x for rebroadcast with hop limit %d (dropping queued < %d)", p->id,
+                      p->hop_limit, dropThreshold);
+            return false; // Reprocess for routing only, skip app delivery
```
GUVWAF: How do we now ensure it is not again sent to the app?

src/mesh/FloodingRouter.cpp
```diff
@@ -21,10 +21,33 @@ ErrorCode FloodingRouter::send(meshtastic_MeshPacket *p)
...
+    bool seenRecently =
+        wasSeenRecently(p, true, nullptr, nullptr, &wasUpgraded); // Updates history; returns false when an upgrade is detected
+
+    // Handle hop_limit upgrade scenario for routers
+    if (wasUpgraded && IS_ROUTER_ROLE() && iface && p->hop_limit > 0) {
+        // wasSeenRecently() reports false in upgrade cases so we handle replacement before the duplicate short-circuit
+        // If we overhear a duplicate copy of the packet with more hops left than the one we are waiting to
+        // rebroadcast, then remove the packet currently sitting in the TX queue and use this one instead.
+        uint8_t dropThreshold = p->hop_limit; // remove queued packets that have fewer hops remaining
+        if (iface->removePendingTXPacket(getFrom(p), p->id, dropThreshold)) {
```
GUVWAF, line 40: A downside of removing and then re-adding is that the packet thus also gets re-ordered in the queue.

src/mesh/PacketHistory.h
```diff
@@ -16,8 +16,9 @@ class PacketHistory
         PacketId id;
         uint32_t rxTimeMsec;              // Unix time in msecs - the time we received it,  0 means empty
         uint8_t next_hop;                 // The next hop asked for this packet
+        uint8_t hop_limit;                // The hop limit when we received this packet
         uint8_t relayed_by[NUM_RELAYERS]; // Array of nodes that relayed this packet
-    };                                    // 4B + 4B + 4B + 1B + 3B = 16B
+    };                                    // 4B + 4B + 4B + 1B + 1B + 3B = 17B (will be padded to 20B)
```
GUVWAF, line 21: Interesting, I was working on an optimization of the NextHopRouter and also needed this entry (although then the hop limit with which we sent it out). Since it will be padded to 20B, we can increase `NUM_RELAYERS` to 6 for free.

src/mesh/FloodingRouter.cpp
h3lix1: This is fair - I made the assumptions that only the routers would need to repeat messages, but client role is absolutely valid in this use case. I'll make this valid for clients as well, but avoiding other roles like tracker.

erayd: > This is fair - I made the assumptions that only the routers would need to repeat messages, but client role is absolutely valid in this use case. I'll make this valid for clients as well, but avoiding other roles like tracker.

Every role other than CLIENT_MUTE, including TRACKER, may rebroadcast packets unless configured not to.

See the following: https://github.com/meshtastic/firmware/blob/cc3ff1504ac5694aa43bc3c59b9ed80e1252094c/src/mesh/FloodingRouter.cpp#L80-L84

h3lix1: Learning read the code instead leaning on my flawed knowledge about the different roles and just use this function. This works better.

src/mesh/FloodingRouter.cpp
h3lix1: Yup. You're right. Removing.

src/mesh/FloodingRouter.cpp
h3lix1: Blast. you're right again. We don't.

Re-working logic to still filter, but add it to the queue before filtering.

src/mesh/FloodingRouter.cpp
h3lix1: This is true. This seemed best since it would allow to replace traceroute packets (with new contents) and makes the flow consistent. Looking briefly at RadioInterface() it bases its decision on SNR and the airtime slot, etc. Changing the packet already in the queue (other than deleting) looked like it might trigger a race condition with the ISR and Router thread. This seemed safer.

GUVWAF:
@h3lix1 Now we're back to the original reason why this PR wasn't merged yet. Either we return `false` in `wasSeenRecently()` such that all logic gets redone, but that has the issue that it also leads to duplicates at the app. With the current approach, it is simply rebroadcasted, but then it is not handled by modules like TraceRoute.

I feel like we might want to use `TransportMechanism` to denote this is a duplicate packet and avoid sending it twice to the client, but I'm happy to hear other suggestions.

h3lix1:
@GUVWAF I went a different route, but one that I like better. This new route will re-run all the modules, except not send to the phone. After looking at it more, traceroute is only one of the things that might want to be updated with the updated information. Nodeinfo should also have the latest/best information as well.

All the modules (thankfully) handle this very well. Most key off of (from, id) which stays the same, and will just do an in-place replacement or StoreForward that has its own history queue.

Hopefully this is an acceptable alternate path.

GUVWAF:
Sorry for keeping you busy all the time ;) 
This looks quite good, but it’s not only `sendToPhone()` that's problematic. Calling the modules again also triggers things like the TextMessageModule, which also triggers e.g. the [InkHUD](https://github.com/meshtastic/firmware/blob/db2f79b6c467fc77ac7c024fb6838bb8d27c0324/src/graphics/niche/InkHUD/Applets/User/DM/DMApplet.cpp#L9) that might display duplicate texts, for example. 

I’m wondering if we really should only re-do traceroute, as I indeed can also not think of any other module that needs the double treatment. I looked at which modules implement `alterReceivedProtobuf()` and for the others they either use that for when receiving packets from the phone (Position, ATAK), or it will remain unchanged (NeighborInfo).

h3lix1:
@GUVWAF You're doing fantastic at finding all the cases I overlooked. I owe you a drink sometime. Maybe two.

Refactored to only handle the traceroute and nodedb update cases, since I really want the nodeDB to show the best number of hops. Agreed this is likely the better route, although elegance is a little lacking in nexthoprouter and floodingrouter's shouldFilterReceived ...

src/modules/TraceRouteModule.cpp
```diff
@@ -153,6 +153,20 @@ void TraceRouteModule::alterReceivedProtobuf(meshtastic_MeshPacket &p, meshtasti
...
+{
+    if (mp.which_payload_variant != meshtastic_MeshPacket_decoded_tag || mp.decoded.portnum != meshtastic_PortNum_TRACEROUTE_APP)
+        return;
+
+    meshtastic_RouteDiscovery decoded = meshtastic_RouteDiscovery_init_zero;
+    if (!pb_decode_from_bytes(mp.decoded.payload.bytes, mp.decoded.payload.size, &meshtastic_RouteDiscovery_msg, &decoded))
+        return;
+
+    handleReceivedProtobuf(mp, &decoded);
+    meshtastic_MeshPacket copy = mp;
```
GUVWAF: The original packet will not be modified using this. It also can't, because it's a `const`.

src/mesh/NextHopRouter.cpp
```diff
@@ -32,7 +36,33 @@ bool NextHopRouter::shouldFilterReceived(const meshtastic_MeshPacket *p)
...
+    if (wasUpgraded && isRebroadcaster() && iface && p->hop_limit > 0) {
+        // Upgrade detection bypasses the duplicate short-circuit so we replace the queued packet before exiting
+        uint8_t dropThreshold = p->hop_limit; // remove queued packets that have fewer hops remaining
+        if (iface->removePendingTXPacket(getFrom(p), p->id, dropThreshold)) {
+            LOG_DEBUG("Processing upgraded packet 0x%08x for relay with hop limit %d (dropping queued < %d)", p->id, p->hop_limit,
+                      dropThreshold);
+
+            if (nodeDB)
+                nodeDB->updateFrom(*p);
+            if (traceRouteModule && p->which_payload_variant == meshtastic_MeshPacket_decoded_tag &&
```
GUVWAF, line 57: This would need to be guarded with `#if !MESHTASTIC_EXCLUDE_TRACEROUTE`.

GUVWAF:
I did remember that there were quite some things here causing this to be not so easy to fix, and I now slowly start remembering which ;)

In principle this looks good enough to me, but there's an issue with the `const` MeshPacket.

src/modules/TraceRouteModule.cpp
h3lix1: Oops, that was why I made it mutable before. I thought I was being smart by changing it.

src/mesh/NextHopRouter.cpp
h3lix1: 🤦‍♂️ Why do we even have that lever? :) No worries, removing the headers, too.

image

h3lix1:
Will re-merge. Not sure how the rebase pulled in so much v4 stuff into this MR.

h3lix1:
Verified in router and router_late and router mode on a popular router.

DEBUG | 18:26:44 2639 [Router] Packet History - Hop limit upgrade: packet 0xa53a790d from hop_limit=2 to hop_limit=3
DEBUG | 18:26:44 2639 [Router] Processing upgraded packet 0xa53a790d for rebroadcast with hop limit 3 (dropping queued < 3)

It does still work in router mode, but the packet has almost always left the queue first before upgrading. In router_late, the queue is perpetually full, so its effectiveness might not be as high in very busy networks.

h3lix1:
@GUVWAF Let me know the steps left to get this pushed to the develop branch.

src/modules/TraceRouteModule.cpp
GUVWAF: @h3lix1 This seems to be still there. `const` should be removed and then it should modify the original struct.

h3lix1: Ah, I see what you mean now.

This is not the original mp, but the one that just arrived that will be replacing the one that was just removed from the queue. It creates a copy before handing it off to alterReceivedProtobuf(). 
```
    handleReceivedProtobuf(mp, &decoded);
    meshtastic_MeshPacket copy = mp;
    alterReceivedProtobuf(copy, &decoded);
```
This is where the new mp (now "copy") becomes mutable, and is modified in alterReceivedProtobuf to be be re-queued and sent. At least that is how I'm reading the code.

I didn't want to modify the original frame because it just seemed like the right thing to do to keep the original frame intact and make a copy, but I can also modify so mp is isn't a const. (I'll need to cast the variable with const_cast in shouldFilterReceived)

If you want it this way, let me know.

GUVWAF:
Please see my remaining comment about the `const` MeshPacket above.

h3lix1:
@GUVWAF I responded...

edit: And I was wrong. (facepalm) Anyway.. It should now work.

src/modules/TraceRouteModule.cpp
h3lix1: I was incorrect in the above. alterReceivedProtobuf does not add it back to the queue to be sent.. Now fixing to use the same mp.

h3lix1:
@erayd - do you have write access to merge? :)

erayd:
> @erayd - do you have write access to merge? :)

Nope. Will need one of the other devs to do that.

PR #5538, opened 2024-12-09
Initialize dmac array to nulls

jp-bennett:
I've had some crashing on runs where there is no mac address source. It seems to be in the printBytes() function, which really confuses me, but making sure to initialize to nulls here is an obvious step.


jp-bennett:
Oh, duh, console hasn't been initialized. Shout-out to allanmac on discord for pointing this out, as I'd forgotten it was a thing.

PR #5540, opened 2024-12-10
Add portduino-buildroot variant

vidplace7:
Adds new variant `portduino-buildroot`, reliant upon [platform-native#6](https://github.com/meshtastic/platform-native/pull/6).

This will allow us to leverage cross-builder systems like Buildroot and OpenWRT (or others that use a similar convention).

### DIFF
--- a/arch/portduino/portduino.ini
+++ b/arch/portduino/portduino.ini
@@ -1,6 +1,6 @@
-; The Portduino based sim environment on top of any host OS, all hardware will be simulated
+; The Portduino based 'native' environment. Currently supported on Linux targets with real LoRa hardware (or simulated).
 [portduino_base]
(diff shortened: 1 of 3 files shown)


vidplace7:
When [platform-native#6](https://github.com/meshtastic/platform-native/pull/6) is merged I'll include the platform-native update in this PR and un-mark as Draft.

thebentern:
@vidplace7 platform change is merged

arch/portduino/portduino.ini
```diff
@@ -1,6 +1,6 @@
 ; The Portduino based sim environment on top of any host OS, all hardware will be simulated
```
caveman99, line 1: While we're at it, this comment is outdated. Not purely a sim environment any more. Would you do us the honor to update it as well? :-)

thebentern (approved):

PR #5548, opened 2024-12-10
Portduino: Move meshtasticd/web out of /usr/share/doc/

vidplace7:
Second attempt after #5323 was reverted, with changes to address the previous issue.

---

`/usr/share/doc` is the home of **documentation** on most Linux distros.
On minimal / embedded distros (such as buildroot) this directory is often removed by default to save disk space; akin to stripping manpages.

This PR moves `meshtasticd/web` out of `/usr/share/doc` and into `/usr/share` on Portduino platforms.

Includes a symlink from the new location to the previous location, to ease the transition.
https://manpages.debian.org/testing/debhelper/dh_link.1.en.html

PR #5550, opened 2024-12-10
Portduino: fix transitional symlinks for /usr/share/doc/

vidplace7:
Should fix issue with symlinks not being created properly in #5548


vidplace7:
Finally working now :+1:

PR #5553, opened 2024-12-12
Initiate magnetometer based compass calibration from button presses

danwelch3:
- only active for BMX160 accelerometers on RAK_4631
- replace automatic calibration on power on with button triggered calibration
- set 5 presses to trigger 30s calibration
- set 6 presses to trigger 60s calibration (useful if unit is not handheld, ie vehicle mounted)
- show calibration time remaining on calibration alert screen


fifieldt:
Thanks for this! It looks like you may have checked in changes to the protobufs. These are automatically generated , so please update the patch not to include these :)

danwelch3:
Thanks for pointing that out.  I think I successfully removed the changes to protobufs from the patch.

thebentern (approved):

PR #5557, opened 2024-12-13
Add new endpoint to retrieve node info

andrepcg:
I've been using the HTTP endpoint a lot and it would be cool if Nodes could be exposed there as well.

For that I introduced a new `/json/nodes` endpoint which returns a array of nodes tied to a user (`has_user`)


thebentern (approved):

fifieldt (approved):

fifieldt:
Thanks for the cool new feature.

PR #5561, opened 2024-12-13
First attempt to adding hardware support for NRF52 SPI SD Card

Woutvstk:
My first time contributing to an open source project so not very confident in what i'm doing.

Changes to
FSCommon:
initializing SD library for NRF52. Progress: No compile error, but SD card does not get initialized properly yet added ifdef ARCH_ESP32 conditions around esp32 SD library functions

memget: added ifdef conditional statements

StoreForwardModule.cpp: added ifdef conditional statements

Rak4631 platfromIO.ini and variant.h:
added arduino-libraries/SD@^1.3.0 library to libdeps defined HAS_SDCARD and SPI pins

Arduino SD library. Made changes to library because using namespace SDLIB in header file caused ambiguity problems Not sure this is the right way of adding a library, also, how do i implement changes to the library permanently to the project?

Am I going somewhat in the right direction with these changes? Tell me your thoughts, thanks


garthvh:
There is already an open draft pull from one of our primary firmware developers for this 

https://github.com/meshtastic/firmware/pull/5083

caveman99 (changes requested):
There is a lot of duplicate code in there. what about instead defining a couple of wrapper functions to emulate the original ESP32 sdcard driver? Also the test file on init is not neccessary. Ihe way it is not it makes the code really problematic to maintain.

caveman99 (approved):

PR #5571, opened 2024-12-14
TFT cherry-pick: PacketAPI

fifieldt:
* Adds the PacketAPI files from the Meshtastic UI branch (code goes in, but not used yet!)
* Moves PhoneAPI's handleStartConfig function from private to protected


mverch67:
Note that class PacketAPI has an association to PacketServer which serves the API. Those classes should be merged together.

fifieldt:
Abandoning this, since I think we're close to merging the branch itself :)

PR #5575, opened 2024-12-15
(Experimental) Portduino: WolfSSL Support

vidplace7:
Adds experimental support for [WolfSSL](https://www.wolfssl.com/), the native SSL implementation on OpenWRT.

Leveraging `libwolfssl` should enable hardware offloading on OpenWRT devices, which is typically enabled when available.

> With a 20-100 kB build size, wolfSSL is up to 20 times smaller than OpenSSL. wolfSSL is a better choice for resource constrained environments – where every byte matters.

See: https://www.wolfssl.com/documentation/manuals/wolfssl/chapter13.html


vidplace7:
Closing until I get a chance to revisit.

PR #5590, opened 2024-12-16
Allow specifying NodeInfo broadcast channel via userPrefs

medentem:
- Added an optional userpref value `USERPREFS_NODEINFO_BROADCAST_CHANNEL_HASH `. When that is set to the hash for a given channel (as computed in Channels.cpp `generateHash`), nodeinfo will be sent out automatically on that channel rather than the primary channel.

- By default, this setting is not available and merging this will not affect current behavior because it is behind ifdefs and a preference value. This gives some time to test the concept further. I have tested this with my own nodes and it behaves as expected so far.

- If a user sets the hash to the wrong value, it will simply fallback to current behavior.

- If you wish to compute the hash for a channel, you can use this fiddle: https://jsfiddle.net/20tcodpb/1/ You need the channel name and the key bytes. 

- The default value in the userprefs.jsonc file is the hash for LongFast (0x08). With this value, a user could set their primary channel to something more private, send their locations there, but also auto discover on LongFast by sending nodeinfo to that channel.


src/modules/PositionModule.cpp
```diff
@@ -315,6 +316,13 @@ void PositionModule::sendOurPosition()
     // If we changed channels, ask everyone else for their latest info
     LOG_INFO("Send pos@%x:6 to mesh (wantReplies=%d)", localPosition.timestamp, requestReplies);
     sendOurPosition(NODENUM_BROADCAST, requestReplies);
+
+#ifdef USERPREFS_CONFIG_DISCOVERY_CHANNEL_HASH
+    // If the user wants discovery on a different channel, send only nodeinfo to that channel
+    // sendOurNodeInfo() will automatically handle routing a nodeinfo packet to the correct channel
+    assert(nodeInfoModule);
+    nodeInfoModule->sendOurNodeInfo();
```
GUVWAF, line 324: Please remove this part. We shouldn't just send out a NodeInfo **every** time immediately after you're sending out a position.

GUVWAF (changes requested):

src/modules/PositionModule.cpp
medentem: That's fair. What are your thoughts on this -> I think many people use the smart broadcast mechanism to do wardriving node discovery, or they use it because its a reliable way to pickup nearby nodes as you move around (ie. send a packet when my position changes sufficiently, receive responses). What would be a way to maintain this in principle, but address your concern of sending it every time. 

Would 1 out of every 4 position sends work?

GUVWAF: How is wardriving with the right primary channel Greek to them when they already set up a private primary channel for precise location sharing, and LongFast as secondary to communicate with the public? 
You’re trying to achieve parallel communication channels using the same physical medium. If you want to educate users, you should educate them that you can’t just duplicate traffic on a very low bandwidth mesh.

medentem: Done

userPrefs.jsonc
```diff
@@ -34,4 +34,5 @@
   // "USERPREFS_USE_ADMIN_KEY_0": "{ 0xcd, 0xc0, 0xb4, 0x3c, 0x53, 0x24, 0xdf, 0x13, 0xca, 0x5a, 0xa6, 0x0c, 0x0d, 0xec, 0x85, 0x5a, 0x4c, 0xf6, 0x1a, 0x96, 0x04, 0x1a, 0x3e, 0xfc, 0xbb, 0x8e, 0x33, 0x71, 0xe5, 0xfc, 0xff, 0x3c }",
   // "USERPREFS_USE_ADMIN_KEY_1": "{}",
   // "USERPREFS_USE_ADMIN_KEY_2": "{}"
+  // "USERPREFS_CONFIG_DISCOVERY_CHANNEL_HASH": "8"
```
GUVWAF, line 37: Let's change this to something like `USERPREFS_NODEINFO_BROADCAST_CHANNEL_HASH`, since it's not a normal config option.

GUVWAF (approved):

thebentern:
I have gone back and forth on this, but I think ultimately, if we make userprefs more publicly consumable, this non-standard behavior which has some problematic implications in the apps would become too easily accessed by users. I'm in favor of keeping this fully out of the mainline codebase and folks can compile into a fork if they really want it. 

In the long term, I think we should work on addressing the deficiencies that make this sort of behavior change desirable for some users.

garthvh:
> I have gone back and forth on this, but I think ultimately, if we make userprefs more publicly consumable, this non-standard behavior which has some problematic implications in the apps would become too easily accessible by users. I'm in favor of keeping this fully out of the mainline codebase and folks can compile into a fork if they really want it.
> 
> In the long term, I think we should work on addressing the deficiencies that make this sort of behavior change desirable for some users.

Agreed, this is a huge change that makes explaining to users how things work much more complicated.

medentem:
> ...which has some problematic implications in the apps

Genuinely curious - what are these problematic implications?

medentem:
Understood but this should not increase traffic - it simply moves the standard nodeinfo broadcast to a different channel. Right?

garthvh:
> Understood but this should not increase traffic - it simply moves the standard nodeinfo broadcast to a different channel. Right?

Makes everything more complex for no benefit, known as a can of worms.

GUVWAF:
I totally agree it makes explaining it to users more difficult. However, I believe it wouldn't have an influence on the apps, since it's based on the channel hash and not an index, and it falls back to the primary if there's an issue.

Anyway, I don't mind whether this is only in userPrefs or only on a fork; either option is fine with me.

medentem:
That was my thought. It shouldn't affect the client apps, and it doesn't add any traffic. It simply shifts nodeinfo broadcast to a diff channel. And the default channel hash in the userPrefs file is LongFast. 

So practically speaking, this just moves nodeinfo to LongFast for those that are using Primary for something more private because of the auto position sends that occur there. @thebentern thoughts? Maintaining a fork is an option, but it's a hassle and this seemed pretty benign.

garthvh:
Making thing more complicated to explain to users when things are already too complicated is not a win, practically speaking the bulk of the user base should be the primary consideration.

medentem:
Dude. I get that... but who would you be explaining this to if its a UserPref only utilized if compiling from source? Even if it were explained, its not complicated. The concept is already explained here: https://meshtastic.org/docs/configuration/tips/#creating-a-private-primary-with-default-secondary

I think you leave it out of the docs entirely, but if you were to add it, the explanation would be added to that section of the docs. 

"Your nodes information is also automatically sent over this primary channel which generally helps discovery other nearby nodes. Because you're making the primary channel private, that default behavior may limit node discovery. You can optionally automatically send your nodes information over a more public secondary channel by using [insert variable or instruction here]"

Again, don't think it needs to be in the docs at this point, but the concept is really not complex.

PR #5604, opened 2024-12-18
extend mqtt address length again. some cloud providers have long hostnames.

xhci:
Related to #2117 

Some cloud providers use long hostnames for MQ instances, and it can be cost prohibitive to use a custom hostname. The default hostname in my case is 67 characters long.


garthvh:
This exceeds the max packet size.

PR #5608, opened 2024-12-19
Remove unnecessary memcpy for PKI crypto

esev:
The PKI/Curve25519 encryption and decryption methods use separate buffers for input and output. They do not encrypt/decrypt in place. It is not necessary to have a separate `ScratchEncrypted` space for these methods to use.

### DIFF
--- a/src/mesh/CryptoEngine.cpp
+++ b/src/mesh/CryptoEngine.cpp
@@ -58,10 +58,16 @@ void CryptoEngine::clearKeys()
  * Encrypt a packet's payload using a key generated with Curve25519 and SHA256
  * for a specific node.
  *
- * @param bytes is updated in place
+ * @param toNode The MeshPacket `to` field.
+ * @param fromNode The MeshPacket `from` field.
+ * @param remotePublic The remote node's Curve25519 public key.
+ * @param packetId The MeshPacket `packet_id` field.
+ * @param numBytes Number of bytes of plaintext in the bytes buffer.
+ * @param bytes Buffer containing plaintext input.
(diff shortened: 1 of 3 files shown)


thebentern (approved):

esev:
> Does this break the case where we attempt to decrypt with PKC, it fails, and we then attempt to decrypt with the classic encryption scheme?

For `crypto->decryptCurve25519` the output goes into the `bytes` buffer, so that won't impact Channel decryption. The corresponding `pb_decode_from_bytes` does overwrite the original data, but if it fails the existing code returns without attempting Channel decryption.

That's how I convinced myself that this was safe. Please let me know if I missed something!

esev:
@jp-bennett I wonder if there might be a pre-existing bug here?

The decrypt attempts for the Channels happens in a `for` loop. `crypto->decrypt()` will modify that `bytes` buffer in place. Should I move that memcpy from line 360 to within the for loop on line 368?

jp-bennett:
> @jp-bennett I wonder if there might be a pre-existing bug here?
> 
Yeah, looks like it. It's a pretty narrow case, where multiple channels have the same hash. Still, good catch.
> The decrypt attempts for the Channels happens in a `for` loop. `crypto->decrypt()` will modify that `bytes` buffer in place. Should I move that memcpy from line 360 to within the for loop on line 366 (within `if (channels.decryptForHash(chIndex, p->channel)) {`)? That way each pass through the loop sees the original encrypted bytes. And the memcpy only happens if the node has a channel hash that matches as well.

Seems like a reasonable fix, yes.

thebentern (approved):

Issue #5543, opened 2024-12-10
[Feature Request]: CryptoEngine.cpp questions & suggestions


esev:
### Platform

Cross-Platform

### Description

Original: https://discord.com/channels/867578229534359593/1022240589091192853/1313927254400569355

@jp-bennett FYI

In case this is useful: When reimplementing the PKI encryption in my integration there were a couple parts in the firmware that were unclear at first.
- The nonce uint32/64 values are all stored in little-endian format. I'm not super familiar with the hardware side, so I had to look that up. It'd be helpful to have a comment about that.
- I don't understand why the `packetId` is a uint64 in the nonce generation code. It seems like it should be a uint32 as the most significant 32-bits are overwritten by `extraNonce`.
- I wasn't sure why there is an `if (extraNonce)`. Why is the memcpy conditional? Are there cases where `extraNonce` isn't supplied and where the packetId value is larger than 32-bits?
- It would be nice if the constants for M (8) & nonce size (13) were given names in CryptoEngine.cpp. The comment about L=2 was helpful in aes-ccm.cpp, as were the names of the function parameters. I just think it would be clearer to have those as named variables or comments in CryptoEngine.cpp.
- I got confused about what `auth` was being used for. I thought it was the "additional *auth*enticated data" found in AEAD since space was allocated right after it for the extra nonce. It wasn't immediately clear that this was the output of aes-ccm, when encrypting, and not an input. `authTag` might avoid ambiguity. Maybe just a little packet header ASCII art diagram would clear it up too.

Once I figured out those bits, it interoperated perfectly with github.com/pschlump/AesCCM

PR #5613, opened 2024-12-19
Clarify the nonce values/sizes

esev:
This PR is not intended to have any functional changes, it is only meant to document/clarify how the nonce value is computed for #5543.

I've looked as far back as [v2.0.0.18ab874]( https://github.com/meshtastic/firmware/blob/v2.0.0.18ab874/src/mesh/Router.cpp) and as best I can tell the argument used for `packetId` has always been populated with the `uint32_t` MeshPacket `id` field. When encoded into the nonce in little-endian format as a `uint64_t` only the first 32-bits contain the `id` value. The remaining most significant 32-bits that follow have always been set to 0. The `extraNonce` then overwrites those most significant 32-bits if it is provided.

I found this partial overwriting of the 64-bit packetId field with the extraNonce and the mismatch between the MeshPacket.id field type to be confusing when reading the code. The comment in the header file about how the values are concatenated also misled me to believe the extraNonce came after the sending node number.

As long as extraNonce is described as always being 0 when using the AES-CTR (Channel/classic) algorithm, all prior implementations where it was assumed that the packetId was 64-bit will be unchanged and be fully backward compatible; the most significant bits will remain 0 as they were in v2.0.0.18ab874. For the newer AES-CCM  (PKC/PKI) algorithm, I don't think the full 64-bit packetId has ever been populated in the nonce without being half overwritten by extraNonce, so I think there shouldn't be any issues with backward compatibility. This allows for a unified description of how the nonce is computed that works for both algorithms.

I'll start this as a draft. @jp-bennett & @caveman99 I welcome any feedback on this. I'm only intending to make things clearer, and if this PR doesn't do that then it isn't needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved packet encryption and decryption consistency by standardizing packet identifier handling.
  * Updated nonce generation to use a defined layout for packet IDs, optional nonce values, source nodes, and encryption counters.
  * Maintained consistent behavior across Curve25519 and AES-CTR packet protection.
* **Tests**
  * Updated cryptographic tests to verify the revised packet identifier format.

### DIFF
--- a/src/mesh/CryptoEngine.cpp
+++ b/src/mesh/CryptoEngine.cpp
@@ -61,7 +61,7 @@ void CryptoEngine::clearKeys()
  * @param bytes is updated in place
  */
 bool CryptoEngine::encryptCurve25519(uint32_t toNode, uint32_t fromNode, meshtastic_UserLite_public_key_t remotePublic,
-                                     uint64_t packetNum, size_t numBytes, uint8_t *bytes, uint8_t *bytesOut)
+                                     uint32_t packetId, size_t numBytes, uint8_t *bytes, uint8_t *bytesOut)
 {
     uint8_t *auth;
     long extraNonceTmp = random();
@@ -77,7 +77,7 @@ bool CryptoEngine::encryptCurve25519(uint32_t toNode, uint32_t fromNode, meshtas
         return false;
     }
     crypto->hash(shared_key, 32);
-    initNonce(fromNode, packetNum, extraNonceTmp);
+    initNonce(fromNode, packetId, extraNonceTmp);
 
     // Calculate the shared secret with the destination node and encrypt
     printBytes("Attempt encrypt with nonce: ", nonce, 13);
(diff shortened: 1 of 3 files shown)


esev:
One thing I'm still not clear on: Is extraNonce intended to be interpreted as an integer value by clients, or is it intended to be treated as raw bytes. If it is intended to be read as an integer (counter), can I clarify in the comments that this is also written to the nonce in little-endian format?

jp-bennett:
Sorry to take so long to get back to this, but it looks correct to me. We'll obviously want to triple-check this in real-world tests, as repeating a nonce value is cryptographically a huge problem.  I think of the extraNonce value as raw bytes, but it's probably also equally correct to think of it as an integer written in little-endian.

4np:
I have actually been looking at the nonce from the client perspective, and it's quite unclear to me what is actually means or is supposed to do. From the protobufs I understand that whatever nonce is _sent_ in the `ToPhone` protobuf payload, should be returned by the `FromPhone` payload. However, I receive many BT payloads with nonce `0` and one with the nonce I sent in the `ToPhone`. 

Are the payloads with `nonce = 0` the 'bonus' payloads that I didn't specifically request? Or is the first payload _until_ the payload with the `nonce` I requested all combined the full response payload?

If I look at the iOS codebase, they use an auto increment UInt32, but from the protobuf documentation I understand it's just a random value to validate a response belong to the request (so it could just be `UInt32.random(in: 0..<UInt32.max))`.

Is there an expectation of order in the nonce values in the firmware?

caveman99:
Rebased onto develop, base retargeted from master, two commits squashed into one with authorship preserved.

One fix to the original patch: the header comment put the block counter at bytes 4 to 7, which is where `extraNonce` lives. The AES-CTR counter is the last 4 bytes, since `encryptAESCtr` calls `setCounterSize(4)`. The header now points at the diagram in CryptoEngine.cpp instead of repeating it.

Nonce bytes are unchanged. Every caller passes `MeshPacket.id`, a `uint32_t`, so bytes 4 to 7 were always zero before `extraNonce` overwrote them. Diffing both versions of `initNonce` over 20M inputs gives zero mismatches, and both reproduce the captured nonce in test_crypto.

caveman99:
@esev can you please confirm this is still your intended PR?

PR #5651, opened 2024-12-23
Add NXP_SE050 detection

jp-bennett:
Adds support for detecting the NXP SE050 crypto chip.


variants/portduino/platformio.ini
```diff
@@ -6,6 +6,7 @@ build_flags = ${portduino_base.build_flags} -O0 -I variants/portduino
   -I /usr/include
   !pkg-config --libs libulfius --silence-errors || :
   !pkg-config --libs openssl --silence-errors || :
+  -li2c
```
vidplace7, line 9: Doesn't this belong in `arch/portduino/portduino.ini`?

jp-bennett: Yes, it probably does. Good spot!

fifieldt (approved):

PR #5663, opened 2024-12-24
More accurately determine if MQTT uses the default server

esev:
This avoids needing to use strcmp each time a packet is sent as well.

Tested with the following values for `moduleConfig.mqtt.address`:
- An [empty string](https://github.com/meshtastic/firmware/blob/175ff218f1e8830b501dab26f8f2aa9e40d55051/src/mqtt/MQTT.cpp#L416) (`isDefaultMqttServer=true`)
- mqtt.meshtastic.org (`isDefaultMqttServer=true`)
- mqtt.meshtastic.org:1883 (`isDefaultMqttServer=true`)
- 192.168.1.10:1883 (`isDefaultMqttServer=false`)

### DIFF
--- a/src/mesh/Channels.cpp
+++ b/src/mesh/Channels.cpp
@@ -318,7 +318,7 @@ bool Channels::anyMqttEnabled()
 {
 #if USERPREFS_EVENT_MODE
     // Don't publish messages on the public MQTT broker if we are in event mode
-    if (strcmp(moduleConfig.mqtt.address, default_mqtt_address) == 0) {
+    if (mqtt && mqtt.isUsingDefaultServer()) {
         return false;
     }
 #endif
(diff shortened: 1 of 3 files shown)


fifieldt:
Is it worth also adding this to Channels.cpp ( Channels::anyMqttEnabled() ) to keep it consistent across the code base?

esev:
Nice catch! It makes sense to me.

I checked all the callers of Channels::anyMqttEnabled(), and they each checked to see that `mqtt` was valid too (or the caller was MQTT). So I don't expect the behavior to change by adding the same `mqtt &&` check inside Channels::anyMqttEnabled().

fifieldt (approved):

thebentern (approved):

PR #5667, opened 2024-12-25
Generate a coverage report for End to end tests

esev:
For coverage to be saved, the application under test needs to exit cleanly, calling all exit handling logic. This happens when the application calls `exit()`. I've modified the simulator and the crypto unit test so they exit and save the coverage details.

Demo: see the `code-coverage-report` artifact on https://github.com/esev/meshtastic-firmware/actions/runs/12497423501

The `code-coverage-report/coverage_src.info` file should be compatible with most coverage tools like CodeCov/Coveralls.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,12 +8,15 @@ on:
 jobs:
   test-simulator:
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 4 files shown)


thebentern (approved):

thebentern:
I've used Coveralls in some other projects, so I'll add it to my list to get that added for the firmware

PR #5668, opened 2024-12-26
Unset received SNR/RSSI values upon receiving packet via MQTT

GUVWAF:
When nodes receive a packet via LoRa, they publish it to MQTT without subtracting a hop, but with their received SNR/RSSI added to the packet. This means that when another node then receives this packet via MQTT, they will see *their* SNR/RSSI with `hops_away = 0`. Let's unset these values as they are confusing.


fifieldt:
Ahhhh, so that's what that is! Nice!

fifieldt (approved):

PR #5670, opened 2024-12-26
fix for nrf52 lfs assert boot loop

tavdog:
I actually don't have a way to test this fix because the corrupted filesystem is hard to reproduce but I do know that a call to FSCom.format() did fix my issue when I called it explicitly before FSCom.begin().

### DIFF
--- a/src/FSCommon.cpp
+++ b/src/FSCommon.cpp
@@ -56,10 +56,12 @@ extern "C" void lfs_assert(const char *reason)
     LOG_ERROR("LFS assert: %s", reason);
     lfs_assert_failed = true;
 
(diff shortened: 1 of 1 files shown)


thebentern (approved):

fifieldt (approved):

grievertime:
> I actually don't have a way to test this fix because the corrupted filesystem is hard to reproduce but I do know that a call to FSCom.format() did fix my issue when I called it explicitly before FSCom.begin().

Can reproduce daily with cold weather, can I suggest putting it in the fridge? (I know it sounds stupid but it's the second day in a row that I get the same issue with old firmware)

PR #5677, opened 2024-12-27
Upgrade to LovyanGFX 1.2

fifieldt:
We need a new version for incoming TFT branch. This is the latest version, and @mverch67 says it works. What's not to like?


mverch67:
1.2 is the latest release. Seems to work, too.

fifieldt:
Cool! Let's go to that!

mverch67 (approved):

PR #5693, opened 2024-12-29
Add SE05X library

jp-bennett:


berlincount:
wow, that's a serious Secure Element - what hardware has this?

PR #5661, opened 2024-12-24
Add czech oled localization

kyberpunk:
Hi, I would like to add support for Czech and Slovak specific characters on displays. It covers special characters with diacritic which are not shown in default localization (like ěščřžáéíóúů ...). There are hundreds of meshers in these countries, so I believe it will make them happy. It happens to me very often that I cannot properly read messages from the others on the display.

Wish you nice holidays and Happy New Year!

PR #5695, opened 2024-12-29
Fixed localization on bigger screens

kyberpunk:
Hi. There is a bug, that if the firmware is compiled with bigger screen, then default text localization is used even if custom localization build flag is set. It is caused by this [condition](https://github.com/meshtastic/firmware/blob/master/src/graphics/ScreenFonts.h#L19) in the beginning.

I've fixed macros to always use the localized glyph array (if available) and just change font size assignments for different displays. Will appreciate any idea how to make it even more readable.
Relates to #5661.

### DIFF
--- a/src/graphics/ScreenFonts.h
+++ b/src/graphics/ScreenFonts.h
@@ -16,63 +16,61 @@
 #include "graphics/fonts/OLEDDisplayFontsCS.h"
 #endif
 
+#ifdef OLED_PL
+#define FONT_SMALL_LOCAL ArialMT_Plain_10_PL
+#else
+#ifdef OLED_RU
+#define FONT_SMALL_LOCAL ArialMT_Plain_10_RU
+#else
+#ifdef OLED_UA
+#define FONT_SMALL_LOCAL ArialMT_Plain_10_UA // Height: 13
+#else
+#ifdef OLED_CS
+#define FONT_SMALL_LOCAL ArialMT_Plain_10_CS
+#else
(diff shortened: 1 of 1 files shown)


src/graphics/ScreenFonts.h
```diff
@@ -16,49 +16,53 @@
...
-#define FONT_LARGE ArialMT_Plain_24 // Height: 28
+#define FONT_LARGE_LOCAL ArialMT_Plain_24 // Height: 28
 #endif
 #endif
+
+#if (defined(USE_EINK) || defined(ILI9341_DRIVER) || defined(ILI9342_DRIVER) || defined(ST7701_CS) || defined(ST7735_CS) ||      \
+     defined(ST7789_CS) || defined(USE_ST7789) || defined(HX8357_CS)) &&                                                         \
+    !defined(DISPLAY_FORCE_SMALL_FONTS)
+// The screen is bigger so use bigger fonts
+#define FONT_SMALL FONT_MEDIUM_LOCAL // Height: 19
```
fifieldt, line 59: What happens here if none of the local fonts are defined?

kyberpunk: Font all font sizes there is always fallback to default implementation like here: https://github.com/kyberpunk/firmware/blob/fix-localization-on-big-screens/src/graphics/ScreenFonts.h#L42.

caveman99 (approved):

kyberpunk:
Rebased and resolved conflicts from #5821. Can you merge it please, or is there something I should improve?

PR #5696, opened 2024-12-29
support indicator sensors through Rp2040 serial

caveman99:
this *should* work with the original RP2040 firmware factory loaded. It works better with the updated version from https://github.com/caveman99/indicator_rp2040 (preliminary location)


fifieldt:
Very cool!

thebentern (approved):

fifieldt:
@caveman99 , it seems this made the build unhappy ....

caveman99:
Fix is already underway, I ran out of time yesterday.

PR #5697, opened 2024-12-30
Draft - Bloom Filter Probabilistic Routing

medentem:
This is a draft for discussion and review:

**Proposal**

Expand `relay_node` to 4 bytes and include a 13-byte (104-bit) Bloom filter in the packet header. Each hop adds up to N uint32_t node IDs. The filter uses 2 hash functions.

**Why a Bloom Filter?**

- Space Efficiency: Instead of listing 60 node IDs (240 bytes), a 13-byte filter is far smaller.
- Reduce Redundant Rebroadcasts: We use the filter to decide whether a new hop adds sufficient new coverage to justify forwarding.
- Acceptable False Positives: The ~37% false positive rate (worst case - assumes each node has 20 recently seen nodes to add) is fine for our probabilistic approach; it still drastically reduces traffic.

**False Positives - A Known Limitation**

Bloom filters can give false positives (claiming “node X is in the set” when it actually isn’t), but they never give false negatives (if it says “not in the set,” then it’s definitely not in the set).

For our routing use case, these occasional false positives are acceptable because we only need a rough idea of which nodes are covered to make probabilistic rebroadcast decisions.

**Impact**

- Fewer Duplicates: Less congestion and lower power consumption. Simulations suggest ~30% fewer transmissions, depending on topology and node density. Results below.


GUVWAF:
First thing: this can only be done for version 3.0, because extending the header would be a breaking change.

Sounds interesting, I’m curious to the results of simulations. Note that the interactive simulator is hard to use for scale though, as you’ll be simulating all nodes on one machine. 

One thing I would change is the number of neighbors per hop. 20 is really a lot, while it’s much more common you exceed 3 hops (which is actually 2 hops because you'll need to add the first neighbors already when the hop limit is still 3).
Also, while the bloom filter doesn’t give false positives, if you can’t hear a node directly it doesn’t necessarily mean it cannot hear you.

I’m also not sure if this overhead (duplicating the header size) is worth it if you only base the probability of rebroadcasting on it. With this you could also implement rebroadcasting until you don’t have unique receivers to serve after receiving other nodes’ rebroadcasts.

medentem:
> First thing: this can only be done for version 3.0, because extending the header would be a breaking change.

Makes sense. 

> 
> Sounds interesting, I’m curious to the results of simulations. Note that the interactive simulator is hard to use for scale though, as you’ll be simulating all nodes on one machine.

For sure. Will need some real world after simulating the effects. By the way, I'm having trouble with the simulator. Is there anything off the top of your head that might need to change given the expanded header? I changed the config value to 32 bytes and looked through the code but nothing jumped out at me. But for some reason the packet is losing the header value when flowing through the simulator. 

> 
> One thing I would change is the number of neighbors per hop. 20 is really a lot, while it’s much more common you exceed 3 hops (which is actually 2 hops because you'll need to add the first neighbors already when the hop limit is still 3). Also, while the bloom filter doesn’t give false positives, if you can’t hear a node directly it doesn’t necessarily mean it cannot hear you.

This will be the interesting part. In an ideal state, all a nodes **current** neighbors should be in the filter. But the best we can do is approximate current neighbors because we don't have frequent heartbeats to say "I'm still here". Right now I'm filtering down to nodes heard directly within the last hour because that is the minimum interval a nodeinfo packet goes out. Open to thoughts on this. I don't think we can be sure that 20 is too much. For example, the Chicago mesh is very dense. I think during their scheduled NETs, 20 is low. 

And also, we can't pack the bloom filter with a fixed size with too many nodes or the False Positive Rate will increase. A  node will more often think "I am also covering X  node that is in this filter" when it actually is not. A non-zero false positive rate  should be fine, but we wouldn't want it to climb too high.

> 
> I’m also not sure if this overhead (duplicating the header size) is worth it if you only base the probability of rebroadcasting on it. With this you could also implement rebroadcasting until you don’t have unique receivers to serve after receiving other nodes’ rebroadcasts.

Probabilistic forwarding can be aggressive but still leave room for some level of potential duplication. I view it as being almost inversely proportional to the false positive rate. For as often as we may accidentally think we cover a node, we allow some potential for rebroadcast. But your idea is super interesting. Would have to think about that.

GUVWAF:
> Is there anything off the top of your head that might need to change given the expanded header? I changed the config value to 32 bytes and looked through the code but nothing jumped out at me. But for some reason the packet is losing the header value when flowing through the simulator. 

I also responded to your question here: https://github.com/meshtastic/firmware/pull/5629#issuecomment-2565164966

The header is not sent in "raw" form to the simulator (or any client app); the fields are parsed and then added to the MeshPacket protobuf message. So you either have to add your field to that protobuf message and forward it in the simulator, or as workaround send it with the payload and parse it from there.

src/modules/NodeInfoModule.cpp
garthvh, line 1: This change should not be a part of this feature, was already closed.

garthvh (changes requested):

caveman99:
I don't think we can afford donating 8% of our usable payload for better guessing where to send DM's.

medentem:
> I don't think we can afford donating 8% of our usable payload for better guessing where to send DM's.

This is not for DMs. This would be for all broadcasts.

src/modules/NodeInfoModule.cpp
medentem: I will remove. This is not ready to merge right now.

GUVWAF:
> And its much more than a guess.  

The current implementation only uses it to calculate the probability of rebroadcasting, it's not doing anything deterministically. A similar thing is done now using SNR as metric (but then at least it will rebroadcast when nobody else did), and we'd need to simulate whether the improvement is worth it, considering the extra airtime 16 bytes on each packet takes.

medentem:
> > And its much more than a guess.
> 
> The current implementation only uses it to calculate the probability of rebroadcasting, it's not doing anything deterministically. A similar thing is done now using SNR as metric (but then at least it will rebroadcast when nobody else did), and we'd need to simulate whether the improvement is worth it, considering the extra airtime 16 bytes on each packet takes.

Yes that is true. But unless we include the full node list, with full ids, anything implemented will by definition be probabilistic. The act of compressing a list of node ids into a bloom filter adds that effect. I completely agree we need to simulate and test to see if its worth it. I do think there is a lot of potential here.. needs to be proven. To your earlier point, this may be able to replace a hop limit. Or, at least change its purpose slightly.

medentem:
I have modified the simulator and supporting code to test this out including logging to see exactly how coverage is being handled. You can find it forked here: https://github.com/medentem/Meshtasticator

My early tests indicate it is working as expected. I changed the parameters slightly to bias more strongly towards rebroadcast when a node has at least 1 unique neighbor to extend the range of the packet. This can be tuned by changing the value for `COVERAGE_SCALE_FACTOR` in `MeshTypes.h`. The lowest probability of rebroadcast remains at 0.2 to balance the worst case FPR of the bloom filter.

The rebroadcast probability is calculated using this formula:

`float forwardProb = BASE_FORWARD_PROB + (coverageRatio * COVERAGE_SCALE_FACTOR);`

Some example values and the resulting probability:
BFP = BASE_FORWARD_PROB
CSF = COVERAGE_SCALE_FACTOR
A `forwardProb` of 0 means it will definitively not rebroadcast. A `forwardProb` of 1 means it will definitively rebroadcast.

| BFP | # Neighbors | # Not Covered | coverageRatio | CSF | forwardProb | 
| ---- | ------------ | --------------- | -------------- | ---- | ------------ |
| 0.2 | 10 | 1 | 0.1 | 2 | 0.4 |
| 0.2 | 15 | 4 | 0.26 | 2 | 0.66 | 
| 0.2 | 3 | 3 | 1 | 2 | 1.00 |
| 0.2 | 20 | 3 | 0.15 | 2 | .55 |

This probabilistic based approach is not really the point of the change. It's the bloom filter that can be used to adapt to node coverage. How that is leveraged could be changed. For example, if a node is sure that it has an uncovered node, we could simply pass it along every time and handle the packet as is currently done. Conversely, if a node thinks it has no additional nodes that it can reach, we could prevent rebroadcast. No probabilities involved aside from the FPR inherent to a bloom filter. 

Speaking of the bloom filter. There is a formula to calculate the false positive rate you can expect given `n` number of items added to the filter, `m` filter size in bits and `k` the number of hash functions used. I selected 16 bytes because it could accommodate 60 unique nodes with a 37% false positive rate. At worst case, this would mean that at each hop a node would be adding 20 unique nodes to the bloom filter.

A more likely scenario for the mesh in my area is about 35 unique nodes within a rebroadcast area (hops = 3). In that case the FPR drops to only 18%. Meaning approx ~18% of the time a node could think its neighbor was already covered in the filter. 

Here is a little snippet to run the math to approximate bloom filter FPR: https://jsfiddle.net/x8obhm34/

** In Summary **
There are 3 critical levers to be pulled with this implementation:

1. [The bloom filter settings](https://github.com/medentem/firmware/blob/b1ba75a041ae3286a45bf37c85ba672612263db0/src/mesh/MeshTypes.h#L47) - size (bits) and # of hashes applied to each neighbor being added to the filter
2. The set of neighbors we allow a node to add - limited by a [max number](https://github.com/medentem/firmware/blob/b1ba75a041ae3286a45bf37c85ba672612263db0/src/mesh/MeshTypes.h#L44), [max age](https://github.com/medentem/firmware/blob/b1ba75a041ae3286a45bf37c85ba672612263db0/src/mesh/MeshTypes.h#L65), etc.
3. [How the comparison of coverage is actually applied to the decision of whether or not to rebroadcast](https://github.com/medentem/firmware/blob/b1ba75a041ae3286a45bf37c85ba672612263db0/src/mesh/FloodingRouter.cpp#L79) - currently probabilistic, but could be different.

More test results will follow.

caveman99:
Do the 16 bates really need to be appended to every packet header or can surrounding nodes cache this info? more like a nodeinfo light?

medentem:
> Do the 16 b<tes really need to be appended to every packet header or can surrounding nodes cache this info? more like a nodeinfo light?

That is an intriguing question. I suppose it is a tradeoff. We'd be adding 16 bytes of storage per node in the nodedb. But this could be like a nodeinfo packet where each node says "here are the nodes I cover". That could be at the same interval as nodeinfo, or updated when position moves. 

The downside is that it could be stale quickly in a more dynamic environment, or where nodes are not using position so it is harder to detect when a change in coverage is likely. 

I am very open to this concept though. It's a good alternate approach to explore.

thebentern:
Topologies in many meshes are pretty dynamic with all of the portable nodes moving around as well as nodes coming off and online at unpredictable times. IMO, this approach would have to demonstrate undeniable benefits to justify the 16 byte overhead.

medentem:
> Topologies in many meshes are pretty dynamic with all of the portable nodes moving around as well as nodes coming off and online at unpredictable times. 

More reason it would be more effective on every packet.

Here are some results from the simulator using the following logic when determining whether to rebroadcast:
- If no new coverage added, probability reduced to 0.2 (unlikely to rebroadcast)
- If 1 new node covered, probability 0.8 (likely to rebroadcast)
- If > 1 new node covered, probability 1 (will rebroadcast)
The same node layout was utilized for baseline and coverage filter comparison. I tried to keep the timing of the packets consistent, but it isn't perfect. Node 0 was always the original broadcasting node in this test.

Here is the node layout:
Screenshot 2025-01-02 at 10 34 03 AM

**CURRENT ROUTING APPROACH BASELINE**

Screenshot 2025-01-02 at 10 29 35 AM

There is a linear increase in air utilization of all nodes, with nodes 3 and 4 using the most due to their location relative to the other nodes. 

When a packet is routed throughout the entire network of nodes, as seen below, nodes 5 and 6 always rebroadcast despite offering no additional coverage.

Screenshot 2025-01-02 at 10 36 38 AM

**ROUTING WITH COVERAGE FILTER**

Using the same node placement, here is the channel utilization **with** the coverage filter enabled.

Screenshot 2025-01-02 at 10 20 36 AM

The nodes that offer the most coverage utilize the most airtime, where nodes that offer less unique coverage start to taper off. 

Screenshot 2025-01-02 at 10 18 35 AM

When a node is routed to the full mesh, you can see that node 5 and 6 do not rebroadcast. This is because they do not offer additional coverage and their probability of rebroadcast is reduced to 0.2. 

**Next Steps**

In order to understand the longer term impact I need to rerun these tests over a longer period of time with more packets flowing, and ideally with many more nodes. After adding 10 nodes I start to encounter issues plotting, so I will try to figure out a way around the current limitations.

medentem:
**Scenario - relatively equidistant cluster of nodes**

Node Placement
Screenshot 2025-01-02 at 12 30 32 PM

All broadcasts sent from Node 0.

Existing routing approach:
Screenshot 2025-01-02 at 12 37 13 PM

Coverage filter applied:
Screenshot 2025-01-02 at 12 30 26 PM

Note the reduction in airtime after nodedbs are established and the coverage filter goes to work.

**First broadcast from node 0**

Existing routing approach:
Screenshot 2025-01-02 at 12 37 23 PM

Coverage filter applied:
Screenshot 2025-01-02 at 12 30 45 PM

There are no rebroadcasts, partly in this first case because nodedbs aren't established so coverage is unknown and assumed to be 0 for each node. 

**Second broadcast from node 0**

Existing routing approach:
Screenshot 2025-01-02 at 12 37 33 PM

Coverage filter applied:
Screenshot 2025-01-02 at 12 31 00 PM

3 rebroadcasts compared to 2 with the coverage filter. NodeDBs not fully saturated, but close.

**Third broadcast from node 0**

Existing routing approach:
Screenshot 2025-01-02 at 12 37 42 PM

Coverage filter applied:
Screenshot 2025-01-02 at 12 31 08 PM

**Another broadcast from node 0**

Existing routing approach:
Screenshot 2025-01-02 at 12 38 00 PM

Coverage filter applied:
Screenshot 2025-01-02 at 12 31 24 PM

**Summary**

There is a marked and predictable decrease in airtime utilization from nodes with nothing to add to the overall coverage of the mesh. But because it is a non-zero probability (0.2), we still occasionally see a rebroadcast from a node that has nothingt to add.

GUVWAF:
Thanks for the initial results. Indeed it makes sense that the end-points do not rebroadcast, but the same can be achieved when the nodes set an appropriate hop limit (2 in the first scenario, 0 in the last). It would be more interesting to see how it performs when e.g. one out of two nodes should rebroadcast to reach a certain receiver.

> I tried to keep the timing of the packets consistent, but it isn't perfect. 

You can use a [script](https://github.com/meshtastic/Meshtasticator/blob/master/INTERACTIVE_SIM.md#usage-with-script) to have the nodes transmit at reproducible times. 

However, to really test this at scale in all kinds of (randomly generated) scenarios, it might be worth looking into emulating your proposal in the discrete-event simulator, which would need to be implemented somewhere [here](https://github.com/meshtastic/Meshtasticator/blob/master/loraMesh.py#L208-L209). Then we can see the trade-off between reachability (since it might be you're not rebroadcasting when you should) versus channel utilization more easily.

medentem:
> Thanks for the initial results. Indeed it makes sense that the end-points do not rebroadcast, but the same can be achieved when the nodes set an appropriate hop limit (2 in the first scenario, 0 in the last). It would be more interesting to see how it performs when e.g. one out of two nodes should rebroadcast to reach a certain receiver.

Is that realistic? I don't think most people will know enough to be able to tune their hop limits. Are you referring to something else?

> 
> > I tried to keep the timing of the packets consistent, but it isn't perfect.
> 
> You can use a [script](https://github.com/meshtastic/Meshtasticator/blob/master/INTERACTIVE_SIM.md#usage-with-script) to have the nodes transmit at reproducible times.
> 
> However, to really test this at scale in all kinds of (randomly generated) scenarios, it might be worth looking into emulating your proposal in the discrete-event simulator, which would need to be implemented somewhere [here](https://github.com/meshtastic/Meshtasticator/blob/master/loraMesh.py#L208-L209). Then we can see the trade-off between reachability (since it might be you're not rebroadcasting when you should) versus channel utilization more easily.

I'll take a look.

GUVWAF:
> Is that realistic? I don't think most people will know enough to be able to tune their hop limits. Are you referring to something else?

I's probably not perfect, but you can (and should) fine-tune it. While the default is 3, it's good to only increment by one when you notice you cannot reach some nodes. And it would be even better to use less than 3 when that's not necessary. 
The same holds for assigning roles (router/repeater/client-mute); humans aren't perfect at this, but the bloom filter and asymmetric links with probabilistic rebroadcasting isn't perfect either. On the other hand, with manual settings you don't have any overhead.

medentem:
I have updated the simulator to run these tests at a larger scale per your recommendation @GUVWAF. You can view the changes here: https://github.com/meshtastic/Meshtasticator/pull/21

The existing managed flooding approach is maintained and a new bloom filter approach has been added. Simply change the config to toggle between them. With the bloom filter approach there are a few **really important** changes:

1. When testing the bloom filter routing, set hop limit to something very high (like 15) to simulate effectively removing the limit. I needed to keep a value in there because it is used to track and calculate the reachability metric, but by increasing it to something very high, you can see how the routing behaves with "no limit"
2. SNR based delays are bypassed completely when testing the bloom filter. They should be less useful because of the coverage filter. And you can see a pretty interesting outcome in the results.
3. Roles are not considered at all. This simulates "what if devices didn't need a role". Because the bloom filter is adaptive to coverage, nodes with excellent coverage will naturally bias towards rebroadcast.

Lastly, this is using a piecewise probability function as follows:

1. If a node offers 0 additional nodes worth of coverage, the probability of rebroadcast is 20%.
2. If a node offers 1 additional node, the probability of rebroadcast is 80%.
3. If a node offers > 1 additional node, the probability of rebroadcast is 100%.

This is highly configurable ^ and worth tweaking as we think through how we could leverage knowledge of coverage. I actually a probability function that scales based on the coverage ratio would be better, but I have not tested that yet.

**Simulated Managed Flood - 3 hops**
Screenshot 2025-01-03 at 2 25 19 PM
Screenshot 2025-01-03 at 2 25 14 PM

**Simulated Bloom Routing - 15 hops**

Screenshot 2025-01-03 at 2 24 42 PM
Screenshot 2025-01-03 at 2 24 36 PM

medentem:
This test is using a scaling probability based on the ratio of new coverage a node adds to the network. For 100 nodes:

**Simulated Managed Flood - 3 hops**
Screenshot 2025-01-03 at 2 59 18 PM
Screenshot 2025-01-03 at 2 59 14 PM

Simulated Bloom Routing - 15 hops
Screenshot 2025-01-03 at 3 00 32 PM
Screenshot 2025-01-03 at 3 00 29 PM

medentem:
In very dense topologies (250 nodes), it appears a combined approach of SNR delay + bloom filter is the most effective and allowing the bloom filter to completely reject rebroadcast if a node adds 0 coverage (reduce baseline probability to 0.0).

**Simulated Managed Flood - 3 hops**
Screenshot 2025-01-03 at 3 42 05 PM

**Simulated Bloom Routing - 15 hops**
Screenshot 2025-01-03 at 3 53 02 PM

medentem:
After further testing it appears as though a combined approach of SNR delay and bloom filter is the most effective. However, the hop_limit and device role could be still obsolete with this approach.

I added two additional metrics to the log output of the simulator. Number of packets dropped by delay/hop limit (aka. the current managed flooding approach) and Number of packets dropped by coverage (aka. the new approach).

In this test, 35 nodes were used.

Screenshot 2025-01-03 at 8 57 21 PM

Screenshot 2025-01-03 at 8 58 30 PM

Despite the Bloom routing approach with 15 hops, the results are essentially the same because the hop limit is no longer necessary to prevent runaway packet rebroadcasting.

medentem:
With 100 nodes.

Screenshot 2025-01-03 at 9 11 25 PM
Screenshot 2025-01-03 at 9 10 08 PM

GUVWAF:
Thanks for the results, really appreciate the effort you put into this and it’s good to see the simulator in use.

For your first result it seemed bypassing the SNR-based delays caused it to have the packets spread out much further (38 seconds average delay versus 13.6), likely because then it’s based on channel utilization which would be quite high. In general I don’t think it’s really a positive result, because a 13% increase in Tx airtime for all nodes is not insignificant, and you can see the “usefulness” is rather low; only 32.57% of the packets actually contain a new message.

When using a scaling probability based on the ratio of new coverage a node adds, it looks more promising, and adding back the SNR delay seems indeed the most effective. However, I don’t find the results really spectacular. The percentage of nodes reached is a bit higher, but the Tx air utilization is too. 
Also, I commented on your PR for the simulator, since it looks like you increased the header size to 32 bytes for both methods.

fifieldt:
Just wanted to say: this is enjoyable to read. Thanks for the effort.

medentem:
> Thanks for the results, really appreciate the effort you put into this and it’s good to see the simulator in use.
> 

Thank you. And I appreciate your analysis because its helping me understand the simulator and various considerations. 

> For your first result it seemed bypassing the SNR-based delays caused it to have the packets spread out much further (38 seconds average delay versus 13.6), likely because then it’s based on channel utilization which would be quite high. In general I don’t think it’s really a positive result, because a 13% increase in Tx airtime for all nodes is not insignificant, and you can see the “usefulness” is rather low; only 32.57% of the packets actually contain a new message.
> 

Agreed. I have more tests to run. I think I overlooked ensuring the current  node handling the packet adds itself to the coverage filter. 😒 There will certainly be cases when a node receives a packet but wasn't yet in the filter. 

> When using a scaling probability based on the ratio of new coverage a node adds, it looks more promising, and adding back the SNR delay seems indeed the most effective. However, I don’t find the results really spectacular. The percentage of nodes reached is a bit higher, but the Tx air utilization is too. Also, I commented on your PR for the simulator, since it looks like you increased the header size to 32 bytes for both methods.

Good catch. I've updated now. Will re run tests. I also want to try a non linear probability function. Thank you again for the feedback and guidance.

medentem:
@GUVWAF - I fixed the bug you found with the header length as well as the bug where the transmitting node was not including itself in the coverage filter. I also added a log print of the routing approach in use for the simulation so there is no confusion.

Lastly, I visualized the obvious probability functions we could use, but ultimately I landed on the current approach which is a linear function that scales up the slope which is based on the ratio of new nodes against the full neighbor count.

Screenshot 2025-01-04 at 7 05 13 PM

I have not tested the other functions, but I suspect that we want a steep ramp to 100% probability if the new coverage a node offers approaches 20% of its own neighbors. 

Here are the results for 10 nodes with the aforementioned fixes using both routers.

Screenshot 2025-01-04 at 7 02 53 PM
Screenshot 2025-01-04 at 7 02 33 PM

In this case the bloom filter outperforms the current approach in every metric, including a 20% reduction in avg tx air utilization, a 4% increase in deliverability and a 24% increase in % of packets w/ new data. 

Next few tests will ramp up the node count and see what happens.

medentem:
With 25 nodes:

Screenshot 2025-01-04 at 7 16 11 PM
Screenshot 2025-01-04 at 7 15 34 PM

Similar results at 25 nodes, where the Bloom filter outperforms. I am suspecting the performance will degrade as soon as our mesh size results in an oversaturated bloom filter. False positives will increase which will over suppress traffic. We shall see....

medentem:
At 50 nodes bloom still seems to outperform current approach.

Screenshot 2025-01-04 at 7 19 24 PM
Screenshot 2025-01-04 at 7 18 43 PM

medentem:
With 150 nodes still looks promising. Last test at 250 is next.

Screenshot 2025-01-04 at 7 26 42 PM
Screenshot 2025-01-04 at 7 24 08 PM

medentem:
At 250 nodes still an improvement.

Screenshot 2025-01-04 at 7 48 56 PM
Screenshot 2025-01-04 at 7 39 45 PM

GUVWAF:
Great, looks like it really has potential. 

For version 3.0 (which will likely still take quite a while), we can redesign the header and make a separate one for broadcasts and DMs. Then for broadcasts we can include the bloom filter and for DMs we can use next-hop routing. This would also make the payload limit more or less the same, since broadcasts don’t have 12 bytes PKI overhead.

I still have a couple of concerns: 
- What about incomplete or outdated NodeDB data? The NodeDB size is limited and when you don’t have the channel key nodes transmit with, they are not persisted. Also, `hops_away` is only updated when you receive a packet originating from that node, and it might move in the meantime, or RF conditions change. 
- What about asymmetric links? Currently the simulator doesn’t take this into account, but we’ve seen it happens quite often that you can hear someone, but they can’t hear you.

medentem:
> * What about incomplete or outdated NodeDB data? The NodeDB size is limited ... 

- nodeinfo broadcasts should prefer LongFast if it exists on the device, otherwise fallback to primary. It's more important we understand the nodes around us and users do not understand how this works when they move channels around.
- We don't need to know all the nodes nearby perfectly, just that there are new edges to serve. I think this is accomplished by filtering the nodedb down to nodes we've heard within a reasonable lookback period, and sorting it by strongest SNR to weakest. What do you think about this? I am curious for some feedback here.

> ....and when you don’t have the channel key nodes transmit with, they are not persisted. 

In this scenario, you're saying that Node A has Node B in their coverage list because of a shared channel. But Node A is broadcasting on a channel Node B does not share in this case. Node B may rebroadcast a packet it personally doesn't care about. Is that a bad thing? I think it does not matter so much, because there may be others later in the chain that do care. The mesh isn't just for packets that a given node "understands".

> Also, `hops_away` is only updated when you receive a packet originating from that node, and it might move in the meantime, or RF conditions change.

I am not using `hops_away`... can you explain the idea here? If the general concern is the changing conditions, I believe that is accounted for in two ways as best as it can be: 

1. The coverage filter is regenerated and sent on every packet. It's fidelity is therefore higher than say a separate mechanism to share coverage less frequently.
2. The coverage filter is derived from the most recently seen nodes with the highest link quality. Stale nodes are not considered as "covered" because they have not been seen recently enough. The lookback period is something we could debate. 

> * What about asymmetric links? Currently the simulator doesn’t take this into account, but we’ve seen it happens quite often that you can hear someone, but they can’t hear you.

I think this is equally a problem for the current router and the added bloom filter. In other words, this problem does not get worse or better with this. I'll have to keep thinking about this. Perhaps I can add this dynamic to the simulator. I could see this being correlated to higher gain + better positioned nodes vs. low gain + poorly positioned nodes. The latter node could hear from the former, but not be able to get a signal back.

GUVWAF:
> The mesh isn't just for packets that a given node "understands".

Yes, actually that was my point. If someone is using a private channel, these nodes are not persisted in the NodeDB, so you don't account for them in your coverage even if you need to rebroadcast their packets.

> I am not using hops_away... can you explain the idea here?

You do, right? This is in `getDistinctRecentDirectNeighborIds()`:

```cpp
// Check if this node is a direct neighbor (hops_away == 0)
if (!node.has_hops_away || node.hops_away != 0) {
```

> The coverage filter is regenerated and sent on every packet. 

Yes, but your coverage is calculated based on `hops_away` of nodes in your DB, which is only updated when you receive a packet originating from that node.

> The lookback period is something we could debate.

Indeed, but it's a bit tricky. A node that's not sending anything frequently is a good citizen of the mesh, but it might still be interested in receiving packets.

> I think this is equally a problem for the current router and the added bloom filter.

I don't think it is. The current router uses SNR as a proxy for "new coverage", whereas the bloom filter uses the number of nodes it can *hear* directly.

medentem:
> > The mesh isn't just for packets that a given node "understands".
> 
> Yes, actually that was my point. If someone is using a private channel, these nodes are not persisted in the NodeDB, so you don't account for them in your coverage even if you need to rebroadcast their packets.
> 

Got it. Ok, I understand what you mean. I think this is why using LongFast for nodeinfo should be the default. I had a separate PR that located the LongFast channel by its hash, and used it for nodeinfo broadcasts regardless of which channel position it was in. It should then fallback to the primary channel if the LongFast channel was removed entirely, in which case it is likely that the user is very intentionally opting out of the broader mesh. But there is another way to mitigate as well....

A secondary, small, locally stored **counting** bloom filter could keep track of nodes that are not stored in nodeinfo for exactly this purpose. It would have its own lifecycle and take very little space while providing an approximation of new coverage. Ideally, we'd just store a separate lightweight nodedb for these "unknown channel" nodes. But I am not as familiar with the space constraints this would put pressure on. 

The lightweight nodedb would be easier to implement and offer the same functionality of the current implementation. The counting bloom filter would probably save a lot of space but would be more complex and would provide an approximation of new coverage.

> > I am not using hops_away... can you explain the idea here?
> 
> You do, right? This is in `getDistinctRecentDirectNeighborIds()`:
> 
> ```c++
> // Check if this node is a direct neighbor (hops_away == 0)
> if (!node.has_hops_away || node.hops_away != 0) {
> ```
> 

My fault. Was thinking of the python implementation. You are right that at any given moment of time, the actual coverage will be possibly different than what we think it is, but I think other mechanisms would be mitigating like position based packets, aging out nodes after an hour, etc. Yes... a node may move out of range. What that means is that at worst, we end up rebroadcasting because we think its there when it isn't. 

> > The coverage filter is regenerated and sent on every packet.
> 
> Yes, but your coverage is calculated based on `hops_away` of nodes in your DB, which is only updated when you receive a packet originating from that node.
> 

This is by design. We only want immediate neighbors. You are saying what if you have a cluster of nodes, N of which have not sent their own packets, then we will not see it as a neighbor? Is this likely to happen that a node takes a long time to generate its own traffic?

> > The lookback period is something we could debate.
> 
> Indeed, but it's a bit tricky. A node that's not sending anything frequently is a good citizen of the mesh, but it might still be interested in receiving packets.
> 

Yes, but there is a window during which it is essentially guaranteed that a packet will go out from that node because of the other automatic broadcast mechanisms in place. Right? 

> > I think this is equally a problem for the current router and the added bloom filter.
> 
> I don't think it is. The current router uses SNR as a proxy for "new coverage", whereas the bloom filter uses the number of nodes it can _hear_ directly.

How is that different in this case.

**Current**
Node A broadcasts a packet and is unaware of Node B. Node B hears it with a very low SNR and rebroadcasts quickly.

**Bloom**
Node A broadcasts a packet and is unaware of Node B. Node A broadcasts a packet. Node B hears it with a very low SNR and checks if it can add coverage. Because it is furthest away, it is likely it has more to offer, but it checks to make sure. If it adds coverage, rebroadcasts quickly.

In the latest simulations, the SNR mechanism is still intact. I don't see how the bloom filter makes this scenario worse.

I will see what happens in the simulator.

GUVWAF:
> I  think this is why using LongFast for nodeinfo should be the default.

I’m also talking about nodes that do not use LongFast at all. Their packets are currently routed fine over an existing mesh, but, especially when `BASELINE_REBROADCAST_PROBABILITY` is 0, it may be those packets never reach the destination anymore. The same holds for an infrastructure node that does not have any channel keys (to avoid them being compromised). 

Currently nodes with `has_user = false` will only be in memory as long as the node doesn’t reboot and there is enough space in the NodeDB.

> What that means is that at worst, we end up rebroadcasting because we think its there when it isn't

Yes, but that also means the node that should rebroadcast won’t if it hears that node, e.g. when a node moved from being in range of node A to node B. Moreover, if B didn’t have any other coverage, it won’t rebroadcast in the first place.

> Yes, but there is a window during which it is essentially guaranteed that a packet will go out from that node because of the other automatic broadcast mechanisms in place. Right? 

Yes, the current default minimum is a one hour interval for DeviceTelemetry (assuming no `REPEATER` or `CLIENT_HIDDEN` roles). That’s a long time in which it may move or environmental changes occur.

> If it adds coverage, rebroadcasts quickly.

Yes, but if it thinks it doesn’t add coverage (because it can’t hear someone), it won’t rebroadcast, even though nodes might be able to hear them. And the other way around, you might rebroadcast because you think you add coverage (and that causes other nodes to shut up), while actually you don’t.
SNR is also not perfect, but the question is if the bloom filter based on `hops_away` towards you is still a better proxy for new coverage compared to SNR if there are asymmetric links, especially with `BASELINE_REBROADCAST_PROBABILITY = 0`.

medentem:
I just spent an hour or so understanding the code :-D, and I see that as long as the device is not rebooted, unknown (`has_user = false`) nodes will still be in the node_db until the devices boots and runs the cleanup function.

It is clear that these dynamics need to be modeled in the simulator. I can work on adding them to see the effects.

> > I  think this is why using LongFast for nodeinfo should be the default.
> 
> I’m also talking about nodes that do not use LongFast at all. Their packets are currently routed fine over an existing mesh, but, especially when `BASELINE_REBROADCAST_PROBABILITY` is 0, it may be those packets never reach the destination anymore. The same holds for an infrastructure node that does not have any channel keys (to avoid them being compromised).
> 

Packets will continue to flow as long as the node thinks it can reach additional nodes. So if a node is fully opted out of LongFast (is this even common?), it will have other channels and presumably it has other nodes it intends to communicate with that also have opted out of LongFast. In which case it will have nodes it can serve and will rebroadcast. 

> Currently nodes with `has_user = false` will only be in memory as long as the node doesn’t reboot and there is enough space in the NodeDB.
> 

Regardless, I think I can make a space efficient unknown node mechanism that can mitigate this scenario.

> > What that means is that at worst, we end up rebroadcasting because we think its there when it isn't
> 
> Yes, but that also means the node that should rebroadcast won’t if it hears that node, e.g. when a node moved from being in range of node A to node B. Moreover, if B didn’t have any other coverage, it won’t rebroadcast in the first place.
> 

In essence, this is an issue with stale knowledge of coverage. That is true. If the mesh changes a lot in a given period of time, the broadcast strategy will be flawed. To what extent that actually happens or matters will be something to play out. 

If there are a very small number of nodes establishing a chain, it probably matters more. Particularly because the minimum rebroadcast probability is zero.

> > Yes, but there is a window during which it is essentially guaranteed that a packet will go out from that node because of the other automatic broadcast mechanisms in place. Right?
> 
> Yes, the current default minimum is a one hour interval for DeviceTelemetry (assuming no `REPEATER` or `CLIENT_HIDDEN` roles). That’s a long time in which it may move or environmental changes occur.
> 

True, but again, how likely is this. Some proportion of these nodes will also use position broadcasts or emit other packets that update its presence relative to other nodes.

> > If it adds coverage, rebroadcasts quickly.
> 
> Yes, but if it thinks it doesn’t add coverage (because it can’t hear someone), it won’t rebroadcast, even though nodes might be able to hear them. And the other way around, you might rebroadcast because you think you add coverage (and that causes other nodes to shut up), while actually you don’t. SNR is also not perfect, but the question is if the bloom filter based on `hops_away` towards you is still a better proxy for new coverage compared to SNR if there are asymmetric links, especially with `BASELINE_REBROADCAST_PROBABILITY = 0`.

That makes sense. Yes. More testing required and need this in the simulator. More to come.

medentem:
One additional benefit is that this will likely extend range where networks need many hops, but without risking a lot of rebroadcast noise unnecessarily. The elimination of the need for a hop_limit is a big benefit in my mind. Provided your other concerns are addressed.

medentem:
Added two mechanisms to mitigate your concerns. They will require testing obviously:

**Dealing w/ Stale or Unknown Coverage**
- We keep track of the last time we heard any node to understand more about the status of this node relative to the network
- If the `secondsSinceLastNodeHeard` is beyond the seconds configured as indicating our coverage is stale, just fallback to the regular rebroadcast mechanism
- This will happen under two conditions: 1) The node is just booting up and does not have recent coverage knowledge and 2) the node does not have neighbors it can hear, and broadcasting will not harm anything and may spur additional knowledge of a nodes coverage
- Covered nodes to be added to the bloom filter on the packet are now sorted first by last_heard descending and then SNR.

**Dealing with nodes from a channel we don't have**
- Currently when the node boots, the nodes where we don't have decrypted info get tossed out. Now they get pushed to an in-memory ephemeral vector.
- This is added to the check for covered nodes. 
- TODO: when a node ages out of relevancy, it should just be removed

medentem:
I've added asymmetric link modeling to the simulator: https://github.com/meshtastic/Meshtasticator/pull/21

Please review to make sure the approach is sensible. The asymmetric links are then counted and logged to the output. I also updated the simulator to more closely approximate the latest modifications the firmware.
- When no packets have been received, the rebroadcast probability is 1 (100%).
- Nodes that would normally have been purged would be in the ephemeral vector for a period of time before being considered stale
- Only 20 nodes, sorted by signal strength, are added on each hop.

Here is a simulation with 150 nodes.

Screenshot 2025-01-06 at 7 35 56 PM
Screenshot 2025-01-06 at 7 31 47 PM

medentem:
I also added the following enhancements:

1. Movement simulation seen below, tunable by config parameters. 
2. Coverage more closely approximates real world now where coverage is only added when a packet is received from a node.
3. Coverage is no longer "perfect" as nodes move, their coverage is not immediately known to have changed.

With asymmetric links and movement, here are the results with 150 nodes, with approx 60% of them moving. Bloom marginally outperforms current method on all metrics. I think this is approximately worst case? I will repeat tests with less nodes to see how it shakes out.

moving-nodes-sim

Screenshot 2025-01-06 at 9 22 28 PM
Screenshot 2025-01-06 at 9 16 21 PM

medentem:
Repeated with 50 nodes. Bloom outperforms mostly on airtime with a 19% reduction but similar reachability.

Screenshot 2025-01-06 at 9 31 39 PM
Screenshot 2025-01-06 at 9 29 35 PM

medentem:
With 25 nodes. The layout in this example is particularly interesting. With a high number of moving nodes (keeping in mind this essentially assumes no one has position based reporting turned on), the reachability drops, but airtime is also lower. 

Screenshot 2025-01-06 at 9 34 51 PM

Screenshot 2025-01-06 at 9 35 57 PM

medentem:
Lastly, with 10 nodes. Similarly, lower reach, but higher useful packets and lower airtime. This is because nodes reentering coverage are not known immediately, and they leave too quickly. 

Screenshot 2025-01-06 at 9 43 49 PM
Screenshot 2025-01-06 at 9 43 04 PM

What I need to do the math on and retest again is what the appropriate aging out time window should be relative to the simulation time so that it approximates real world and the firmware.

medentem:
Updated so time is scaled properly for the downstream config values. With 100 nodes, 52% moving, randomly assigned walking, biking or driving speeds.

Screenshot 2025-01-06 at 10 29 26 PM
Screenshot 2025-01-06 at 10 26 12 PM

fifieldt:
Can I ask about a couple edge cases?

*Cave Exploration*
https://meshtastic.discourse.group/t/nondecrementing-router-repeater-role/9707

A "chain" topology, made up of 9+ nodes where each node can only hear the one in front of and behind them in the chain.

GUVWAF:
Nice to see support for mobility and asymmetric links in the simulator! That would be good to have as a separate PR.

I’ve added one comment to PR in the simulator; the coverage can only be updated when you receive a packet originating from a direct neighbor, not when a neighbor is relaying for another node (because we only know the original transmitter, not who is relaying).

Furthermore, I think 2dB standard deviation for link asymmetry is a bit too optimistic. For example, the maximum output power of a SX127x radio is 20dBm, while there are devices with PA capable of 30dBm. Also the SX126x with receive power boost enabled has around 3dB better sensitivity. Next to that there can be effects from the antenna or local noise figures, so the worst case link asymmetry might be as large as 15dB. Maybe a standard deviation of 4dB is more appropriate. 
Here e.g. someone recorded a 7dB difference in SNR:
asymmetric


Like fifieldt, for the NodeDB discrepancies and bloom filter false positives, I’m also more interested in sparse meshes, e.g. where for some cluster of nodes there is only one node that can serve them. In dense meshes this usually works fine, because then another node will jump in.

medentem:
As expected, the current router is only really limited by `hop_count`. If you were to increase that, the chain would likely work up to the maximum hops allowed. However, the big advantage of eliminating `hop_count` (or setting it to something much higher and not having to worry about it) is that the majority of users will have no idea how to set this appropriately, and it does not adapt well to a changing mesh topology. 

Screenshot 2025-01-07 at 2 35 14 PM

medentem:
Another set of updates... :-)

1. Firmware + Simulator -  If a node does **not** have knowledge of a **direct neighbor** (or it's knowledge has aged out), it will fall back to _likely_ (configurable) rebroadcasting. This will help with specific transient states where we know coverage is unreliable - opted out of longfast, sparse network, unestablished node, coverage aging.
2. Firmware - bloom router user pref sets hop limits to 15
3. Simulator - Updated stdev to 4 per @GUVWAF recommendation

The results are as follows: In almost every scenario (high density, low density, chains, asymmetric links), the Bloom filter outperforms the current approach, except 1; in sparse networks where a majority of the nodes are mobile, the bloom filter suffers. This is because coverage is essentially always stale, and because links are so few, that is detrimental to performance. See below. 

Screenshot 2025-01-07 at 6 53 39 PM
Screenshot 2025-01-07 at 6 52 50 PM

I will consider ways to mitigate this considering it performs well under the other conditions... 😮‍💨

medentem:
Trying to finding the threshold here... This is a sparse 5 node net w/ asymmetric links and 1 node moving. The moving node suffers because it is not around long enough for anyone to realize it can be served as it moves to the edges of the network.

I am going to try to model position updates as part of the movement simulation. I'm guessing if position updates are emitted, this problem goes away because coverage is updated as a part of that packet. 

One proposal would be to emit position data if enabled, but fallback to emitting nodeinfo (no position data) when a device moves. The concern will be "that is more traffic", but the cumulative traffic may net out to be less. I'll try to model this effect. 

Screenshot 2025-01-08 at 8 26 34 AM

Screenshot 2025-01-08 at 8 27 22 AM

medentem:
@GUVWAF - I added "smart" position broadcasting and channel utilization clamping but I am getting a little confused with the time scaling. I think I am mixing two different concepts and it's skewing results. Can you help me with that? You can see the latest in the simulator PR how it works now, but I think its not quite right....

medentem:
@GUVWAF - I have been thinking about how a nodes knowledge of coverage could be improved. I think it would be helpful to expand the `relay_node` to a full 4 bytes, and reduce the filter size slightly to accommodate. The overall header size would remain the same 32 bytes.

I think this would majorly improve this routing algorithm because our knowledge of coverage would increase a lot in these other cases. Thoughts?

medentem:
**UPDATED APPROACH**

This has been a saga of changes based on excellent input from @GUVWAF. I wanted to capture the latest approach because it has evolved quite a bit since the beginning. Here is the "How it works":

Packet Header

- The packet header was expanded by 16 bytes to accommodate two changes:
  - `relay_node` was expanded to 4 bytes from 1 byte which ensures when a packet arrives, you know exactly who gave it to you. This is **critical** to a node's knowledge of its immediate neighbors.
  - A 13 byte Bloom Filter was added (`coverage_filter`) and when I node receives a packet, that tells the receiving node "Here are the nodes that have already been covered".

How a Node's Coverage Is Maintained:
- Every packet that arrives will now have `relay_node` which may or may not differ from the `from` node. The important this is that the `relay_node` is always a direct neighbor as it is the actual transmitter of the packet a node receives.
- Theses relay node ids are added to a data structure with the last time they were heard. This is effectively the node's coverage list uncompressed. _(This still needs to be implemented in firmware and will be soon. It is already in the simulator.)_

How a Node's Coverage is Applied:
- When a packet is received the node determines how much new coverage it can add by comparing it's local coverage knowledge against the bloom filter (aka. `coverage_filter`) in the packet header.
- Due to the low bandwidth nature of the mesh, which means periodic automatic packets do not occur often, and potential for stale coverage particularly in sparse, highly dynamic meshes, we need additional mechanisms to mitigate degrading coverage knowledge. Summary of those mechanisms:
  - When a node has no knowledge of any direct neighbors, it falls back to a high probability of rebroadcasting. This is to mitigate first startup, or when we haven't heard from others in a long time there is effectively no harm of broadcasting.
  - Coverage is calculated as a **weighted ratio** of new coverage we offer (when checked against the bloom filter) divided by the total coverage we're aware of - our direct neighbors
    - The weighted ratio is a decay mechanism whereby the longer its been since we've heard a node, the less it counts towards new coverage we offer. Effectively, our confidence in it being nearby starts to shrink.
- After computing our weighted additional coverage, a scaled probability of rebroadcasting is determined based on that additional coverage

What hasn't changed:
- All rebroadcasts are still subject to the SNR-based delay mechanism. In other words, retransmissions will still get stopped if a further node rebroadcasts before our delay expires. The coverage filter is a layer on top of that.

**Main Benefits**
- `hop_limit` is no longer needed with the coverage filter in place which means the mesh is more adaptable to changes over time. Hops stop when coverage stops. 
- Airtime utilization is less because wasteful rebroadcasts are clamped by coverage
- Reachability increases because hops that would otherwise have never happened can now happen without risk of unnecessary chatter
- May be possible that device roles are less important, though more testing is necessary.

GUVWAF:
> I think this would majorly improve this routing algorithm because our knowledge of coverage would increase a lot in these other cases. Thoughts?

Yes, I think extending the `relay_node` too its full ID is something to consider for 3.0. It also has the benefit that you know who sent the “implicit ACK” on a broadcast. 

While this might not be so important now that the full `relay_node` ID is added, I don’t think we can assume that every moving node uses smart position broadcast. Many nodes don’t have GPS or don’t use it because it uses too much power. 

Also, have you every changed the hop limit for `MANAGED_FLOOD`? I’m wondering whether the low reachability in case there’s a lot of nodes is just because the hop limit is too low (see also the figure in the README of the simulator). Also strategically placed `ROUTER`s may help, but it's a bit hard to decide whether it's realistic people assign these correctly.

Next to this, I’m a bit skeptical about the hard-coded parameters like `COVERAGE_RATIO_SCALE_FACTOR`, `UNKNOWN_COVERAGE_REBROADCAST_PROBABILITY` and `RECENCY_THRESHOLD`. Are these still valid for different LoRa modem presets, different `PERIOD` or `PACKET_LENGTH`, etc.?

Looking forward to the new results.

medentem:
> While this might not be so important now that the full `relay_node` ID is added, I don’t think we can assume that every moving node uses smart position broadcast. Many nodes don’t have GPS or don’t use it because it uses too much power.
> 

I can vary this as well.

> Also, have you every changed the hop limit for `MANAGED_FLOOD`? I’m wondering whether the low reachability in case there’s a lot of nodes is just because the hop limit is too low (see also the figure in the README of the simulator). Also strategically placed `ROUTER`s may help, but it's a bit hard to decide whether it's realistic people assign these correctly.
> 

No. I have not. I will try with a higher hop limit on managed flooding, but I don't think the purpose of this test is to see if we can craft a more perfect mesh using hop limits and node roles. The reality is most meshes lack the organization and knowledge to achieve this. 

> Next to this, I’m a bit skeptical about the hard-coded parameters like `COVERAGE_RATIO_SCALE_FACTOR`, `UNKNOWN_COVERAGE_REBROADCAST_PROBABILITY` and `RECENCY_THRESHOLD`.
> 

These are necessary to make testing more convenient and those values have a material impact on the coverage based algorithm. Each represents a careful mitigation of the edge cases where coverage based routing is weak due to the lower bandwidth less frequent updates from neighboring nodes. 

The effect of COVERAGE_RATIO_SCALE_FACTOR can be seen here in the linear line in red. It is used to vary how quickly a 100% probability of rebroadcast should happen. For example, if 20% of my nodes are new for this packet, we want that to rebroadcast. If the scale factor was 1, only 100% new coverage would guarantee rebroadcast. 

image

UNKNOWN_COVERAGE_REBROADCAST_PROBABILITY is used when a node first boots, or all of its coverage has fully aged out. Effectively, the node has 0 confidence as to other nodes around it. Therefore, we should prefer that it rebroadcasts because it is an ignorant node for the time being.

RECENCY_THRESHOLD is the amount of time a neighboring node remains in a nodes coverage list. As the node nears the threshold, its contribution to the probability decays because our confidence in it still being nearby decays. Again, this is a mitigation tactic based on our imperfect means of tracking coverage.

>Are these still valid for different LoRa modem presets, different `PERIOD` or `PACKET_LENGTH`, etc.?

I will need to run more tests, but I don't think the lora modem presets or packet length should impact this because all of these deal with coverage which doesn't change based on those two things. PERIOD may matter more because it is directly tied to establishing coverage knowledge.

medentem:
Here is an interesting test that shows the effect of asymmetric links. This is with 50 nodes. The % of "I don't cover this node, but I think I can" is directly linked to the % of links that are asymmetric. 8.41 / (8.41+13.31) = 38.7%

So the impact of asymmetric links is that I receive a packet, think I can reach back out to that node, but I actually can't. So we may over broadcast due to those asymmetric links.

Screenshot 2025-01-09 at 10 26 19 PM
Screenshot 2025-01-09 at 9 40 07 PM

medentem:
50 nodes. This time with some moving. 

Screenshot 2025-01-09 at 11 04 10 PM
Screenshot 2025-01-09 at 10 56 05 PM

medentem:
Sorry for delay. Simulations are coming, but running them is becoming very time consuming. I am almost done with some changes to the batch simulator so I can run these tests with a little more scale and less manual touch.

garthvh:
It would be good to add more mobile nodes, in many meshes a high percentage of nodes are mobile (and increasing) without position knowledge.

medentem:
So I've been deep in this for 2+ weeks and I think I need to reset my brain a bit to make sense of the results - so many iterations, so many tests. 

What I think I've learned so far is... anything that could degrade moment-in-time knowledge of direct neighbors will degrade the performance of the coverage router. I am working to find the inflection point where the size of the mesh is so small that the sensitivity to coverage is too high, but it seems like < 35. 

In general, results degrade with the new layers of more "real world" simulation - which is intuitive, because things like mobile nodes w/out GPS and asymmetric links degrade any given node's ability to sense direct neighbors accurately over time.

Here are the baseline results comparing managed flood (3 hops, 16byte header) with bloom router (15 hops, 32byte header) and a few other tunable constants that affect rebroadcast probability.
Figure_1
Figure_2
Figure_3
Figure_4
Figure_5

medentem:
Adding asymmetric links (about 25% asymmetric):

Figure_1-asym
Figure_2-asym
Figure_3-asym
Figure_4-asym
Figure_5-asym

medentem:
W/ Asymmetric Links and 20% nodes moving, 50% of which have GPS reporting.

Figure_1-asym-mov
Figure_2-asym-mov
Figure_3-asym-mov
Figure_4-asym-mov
Figure_5-asym-mov

medentem:
W/ Asymmetric Links and 40% nodes moving, 30% of which have GPS reporting.

Figure_1-asym-mov-40p
Figure_2-asym-mov-40p
Figure_3-asym-mov-40p
Figure_4-asym-mov-40p
Figure_5-asym-mov-40p

medentem:
Well friends. After a lot of additional testing I have come to the conclusion that the coverage based router is not viable. Here are the main reasons:

1. It is not possible to reliably detect the conditions under which we would expect a given node's knowledge of its coverage to degrade.
2. When coverage knowledge degrades, rebroadcasting decisions are often wrong.
3. It appears the coverage filter would offer roughly the same performance as the managed flood routing without requiring hop limit. That said, it adds 16 bytes and offers no real NET benefit.
4. At extremely high mesh density (500 nodes), coverage knowledge degrades because too many nodes are added to the compressed data space, which in turn increases the false positive rate.

In short, there is a narrow sweet spot where this works, but we can't detect when the network falls outside the efficacy range of the implementation.

@GUVWAF - I hope the modifications to the simulator are still useful.

GUVWAF:
That’s really a shame. It really looked promising at some point, but if it turns out not to give significant improvements in realistic scenarios, indeed I don’t think it’s worth investigating further. It’s anyway good to have considered the option and really want to thank you for doing this so extensively.

The additions to the simulator are definitely useful. If you have time to separate those additions from the bloom filter implementation, that would be great, otherwise I will try to do it sometime.

medentem:
> The additions to the simulator are definitely useful. If you have time to separate those additions from the bloom filter implementation, that would be great, otherwise I will try to do it sometime.

I will do that. Thank you!

medentem:
Farewell bloom.

nasiralamreeki:
Thanks for all the work all devs do on this project. If we could get non-DM messages to have some sort of smarter algorithm or way of handling message broadcasts, it would be nice. In my area, the network faces congestion and issues due to the current broadcast methodology.

PR #5712, opened 2024-12-31
Run tests as part of the main CI

esev:
I've moved the test-simulator job from tests.yml into its own workflow file and included it into main_matrix.yml.

I also moved the apt package installation to a common action (setup-native) shared with build_native.yml & test_native.yml.

### DIFF
--- a/.github/actions/setup-native/action.yml
+++ b/.github/actions/setup-native/action.yml
@@ -1,0 +1,14 @@
+name: Setup native build
+description: Install libraries needed for building the Native/Portduino build
+
+runs:
+  using: composite
+  steps:
+    - name: Setup base
+      id: base
+      uses: ./.github/actions/setup-base
+
+    - name: Install libs needed for native build
+      shell: bash
+      run: |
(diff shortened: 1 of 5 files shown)


caveman99:
Can't really test this as part of the PR, we need to yolo and see if it works ;-)

esev:
Tested: [main_matrix.yml](https://github.com/esev/meshtastic-firmware/actions/runs/12564687169) & [tests.yml](https://github.com/esev/meshtastic-firmware/actions/runs/12564684357)
There are a few steps that I don't have the ability to run, but those shouldn't be changed by this PR.

esev:
Converted to Draft again. I think it is safer to have no permissions on the jobs that run the tests. I'll rework this so that the permissions are on a job that runs after the tests.

esev:
Demo runs: [main_matrix.yml](https://github.com/esev/meshtastic-firmware/actions/runs/12566590746) & [tests.yml](https://github.com/esev/meshtastic-firmware/actions/runs/12566588291)

thebentern (approved):

PR #5719, opened 2025-01-02
Disable coverage generation

esev:
I noticed that `genhtml` is failing. Let's disable that for now until I can reproduce the issue locally and provide a fix.

https://github.com/meshtastic/firmware/actions/runs/12576207177/job/35052519666

### DIFF
--- a/.github/workflows/test_native.yml
+++ b/.github/workflows/test_native.yml
@@ -140,14 +140,14 @@ jobs:
           path: code-coverage-report
           merge-multiple: true
 
-      - name: Generate Code Coverage Report
-        run: |
-          sudo apt-get install -y lcov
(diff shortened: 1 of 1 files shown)


fifieldt (approved):

esev:
I think I spotted the issue. That last `generate-reports` job looks to not be checking out the correct code for the PR. It seems to be checking out the master branch. I was only able to test in my repo on a branch, so that's likely why it worked there.

Let's proceed with this PR as-is if it looks good to you. I'll work independently on a fix.

esev:
The other issue seems to be with inconsistent working directories inside the two runners for the tests. I see one uses `/home/runner/work/` and the other uses `/home/runner/_work/`.

fifieldt:
> The other issue seems to be with inconsistent working directories inside the two runners for the tests. I see one uses `/home/runner/work/` and the other uses `/home/runner/_work/`.

Runner's working directories can be anything, set by the runner operator with the `--work` flag. So, can't assume that it's a particular name or location -- there's probably a env variable set in the github script, let me look it up.

fifieldt:
@esev , maybe $GITHUB_WORKSPACE ?

PR #5725, opened 2025-01-02
T3S3 and EoRa-S3 board definition file RAM size correction and EoRa-S3 comments

S5NC:
Correct maximum ram size for ESP32-S3FH4R2 boards (T3S3, EoRa-S3) from 320 kB to 2.5 MB
512 kB SRAM, 2 MB PSRAM, _don't count the 16 kB SRAM in RTC (?)_ `512*1024 + 2*1024*1024 = 2621440`

Went through EoRa-S3 schematic and hardware design, double checked variant.h file and added comments along the way. All was correct, no functionality changes, still need to run this on a board to test the SD card

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -124,7 +124,8 @@ lib_deps =
 
 [radiolib_base]
 lib_deps =
-	jgromes/RadioLib@7.1.0
+	; jgromes/RadioLib@7.1.0
+	https://github.com/jgromes/RadioLib.git#92b687821ff4e6c358d866f84566f66672ab02b8
 
 ; Common libs for environmental measurements in telemetry module
 ; (not included in native / portduino)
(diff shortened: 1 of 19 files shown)


thebentern (approved):

mverch67:
I don't think this is correct. AFAIK this number is used by the linker to calculate the space it has for DRAM/IRAM, but PSRAM is not equally usable.

mverch67:
> Is this Arduino "maximum_data_size" the same as PlatformIO's "maximum_ram_size"?

Yes it is. And unless there's a way to tell the linker to allocate static data into external SRAM I wouldn't mess with these values.

```
Building in release mode
Retrieving maximum program size .pio\build\esp32-s3-devkitc-1-n8r2\firmware.elf
Checking size .pio\build\esp32-s3-devkitc-1-n8r2\firmware.elf
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [=         ]  13.9% (used 45552 bytes from 327680 bytes)
Flash: [===       ]  26.9% (used 897869 bytes from 3342336 bytes)
```

PR #5735, opened 2025-01-03
meshtasticd Fedora RPM packaging with rpkg

vidplace7:
~~DRAFT until the [`Meshtastic` copr tracking project](https://pagure.io/fedora-infrastructure/issue/12346) is created.~~

Compiles correctly, both locally and in Fedora COPR.

Testing here https://copr.fedorainfracloud.org/coprs/g/meshtastic/daily/

### DIFF
--- a/meshtasticd.spec.rpkg
+++ b/meshtasticd.spec.rpkg
@@ -1,0 +1,91 @@
+# meshtasticd spec file for RPM-based distributions
+#
+# Build locally with:
+# ```
+# sudo dnf install rpkg-util
+# rpkg local
+# ```
+#
+# See:
+# - https://docs.pagure.org/rpkg-util/v3/index.html
+# - https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/
+
+Name:           meshtasticd
+# Version Ex:   2.5.19
+Version:        {{{ meshtastic_version }}}
+# Release Ex:   9127.daily.gitd7f5f620.fc41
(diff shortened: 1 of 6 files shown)


.vscode/extensions.json
```diff
@@ -2,7 +2,8 @@
     // See http://go.microsoft.com/fwlink/?LinkId=827846
     // for the documentation about the extensions.json format
     "recommendations": [
-        "platformio.platformio-ide"
+        "platformio.platformio-ide",
+        "1dot75cm.rpmspec"
```
caveman99, line 6: I don't think every developer wants to install the rpm helper into vscode :-)

meshtasticd.spec.rpkg
caveman99, line 1: Are these files required in the root directory or can they be moved to something more meaningful and out-of-the-way? ;-)

.vscode/extensions.json
vidplace7: Fair enough :+1:  Removed that part

jp-bennett (approved):

vidplace7:
This is good to go after secrets exist :+1: sent privately.

meshtasticd.spec.rpkg
vidplace7: Unfortunately the `.spec.rpkg` and `rpkg.conf` files need to be present in the root of the directory where the source is :/ I haven't been able to find a clean way around this.

vidplace7:
Went a little wild and packaged all the web components :) (orcania,yder,ulfius)

NOW it's ready :sweat_smile:

fifieldt (approved):

PR #5737, opened 2025-01-04
LS20031 setup support

fifieldt:
We currently detect but do not setup LS20031 chips. LS20031 is a MTK3339-based chip. Therefore, it should share some heritage with other MTK3333 or MTK3339 chips. Re-use the L76B commands for setup.


fifieldt:
@alexmarkley , would appreciate if you could build this with GPS_DEBUG 1 and paste logs from the early part of operations :)

caveman99:
@fifieldt once we merge the improved scancode this (and the other one) probably needs a revisit.

fifieldt:
Successfully rebased.

fifieldt:
@alexmarkley How's that new job going? :)

fifieldt:
There's nothing in this patch that would make life worse for LS20031 users, and it can potentially make life better :) I think it's safe to merge.

PR #5743, opened 2025-01-04
Multi gpiochip support for native environment

psiegl:
While tinkering with Meshtastic on the popular Radxa Zero 3E/W, trying to employ an ebyte E22-900M30S I stumbled across the SBCs GPIO partitioning.

However, let's recap what an E22-900M30S requires in regards to SPI/GPIO. Motivated by [Xiao BLE/BLE Sense + Ebyte E22-900M30S](https://github.com/meshtastic/firmware/tree/master/variants/xiao_ble), an E22-900M30S requires at least:

1. SPI CS* (NSS)
2. SPI MISO
3. SPI MOSI
4. SPI SCK
5. GPIO NRST ('reset')
6. GPIO DIO1 ('irq')
7. GPIO RXEN
8. GPIO BUSY

In case DIO2 is driving TXEN, two GPIOs can be saved, which is most often the setup seen and employed by myself.

Even though the Radxa Zero 3E/Ws size and its [40-Pin GPIO header](https://docs.radxa.com/en/zero/zero3/hardware-design/hardware-interface) provide a first impression as if the header is RPI compliant: you are to some extend mislead.
Let's start with identifying the Radxas single SPI interface (see below), which employs PIN 19, 21, 23 and 24.
Bildschirmfoto_20250104_210247
As there are just four more GPIOs required (NRST, DIO1, RXEN, BUSY), one could employ any close by GPIOs.

However, the moment one will perform a `gpioinfo`, one will detect the following (full dump: [gpioinfo.log](https://github.com/user-attachments/files/18308104/gpioinfo.log) ):
```
gpiochip0 - 32 lines:
        ... 
        line  24:     "PIN_10"       unused   input  active-high 
        line  25:      "PIN_8"       unused   input  active-high
        ...
gpiochip1 - 32 lines:
        line   0:      "PIN_3"       unused   input  active-high 
        line   1:      "PIN_5"       unused   input  active-high 
        ...
        line   4:     "PIN_37"       unused   input  active-high
        ...
gpiochip2 - 32 lines:
        ...
gpiochip3 - 32 lines:
        line   0:      unnamed       unused   input  active-high 
        line   1:     "PIN_11"       unused   input  active-high 
        line   2:     "PIN_13"       unused   input  active-high 
        line   3:     "PIN_12"       unused   input  active-high 
        line   4:     "PIN_35"       unused   input  active-high 
        line   5:     "PIN_40"       unused   input  active-high 
        line   6:     "PIN_38"       unused   input  active-high 
        line   7:     "PIN_36"       unused   input  active-high 
        line   8:     "PIN_15"       unused   input  active-high 
        line   9:     "PIN_16"       unused   input  active-high 
        line  10:     "PIN_18"       unused   input  active-high 
        line  11:     "PIN_29"       unused   input  active-high 
        line  12:     "PIN_31"       unused   input  active-high 
        ...
        line  17:     "PIN_22"       unused  output  active-high 
        line  18:     "PIN_32"       unused   input  active-high 
        line  19:     "PIN_33"       unused   input  active-high 
        line  20:      "PIN_7"       unused   input  active-high 
        ...
gpiochip4 - 32 lines:
        ...
        line  10:     "PIN_27"       unused  output  active-high 
        line  11:     "PIN_28"       unused   input  active-high 
        ...
        line  18:     "PIN_23"       unused   input  active-high   <- CLK
        line  19:     "PIN_19"       unused   input  active-high   <- MOSI
        line  20:      unnamed       unused   input  active-high 
        line  21:     "PIN_21"       unused   input  active-high   <- MISO
        line  22:     "PIN_24"       unused  output  active-high   <- CS0
        line  23:      unnamed       unused   input  active-high 
        line  24:      unnamed       unused  output  active-high 
        line  25:     "PIN_26"       unused   input  active-high 
        ...
gpiochip5 - 3 lines:
        ...
```
Note, I annotated already the PINs that enable the sole SPI interface on gpiochip4. No matter which GPIO PINs were chosen previsouly, it is by no means possible to optimise the handling of BUSY, DIO1, RXEN and NRST by the same gpiochip that already handles SPI, as gpiochip4 just handles 3 further GPIOs. Especially, as meshtastic wants to control the CS*, the current config format is limited to specifying the gpiochip that handles CS*.

Now, when specifying a `config.yaml`, today it can solely support a mapping of PIN name and number, while gpiochip can solely be specified once globally. E.g.:
```
Lora:
  Module: sx1262
  DIO2_AS_RF_SWITCH: true
  DIO3_TCXO_VOLTAGE: 1.8

  # NSS     PIN_24 -> chip 4, line 22
  CS: 24
  # SCK     PIN_23 -> chip 4, line 18
  SCK: 23
  # BUSY    PIN_29 -> chip 3, line 11 !
  Busy: 29
  # MOSI    PIN_19 -> chip 4, line 19
  MOSI: 19
  # MISO    PIN_21 -> chip 4, line 21
  MISO: 21
  # NRST    PIN_27 -> chip 4, line 10
  Reset: 27
  # DIO1    PIN_28 -> chip 4, line 11
  IRQ: 28

  # RXEN    PIN_22 -> chip 3, line 17 !
  RXen: 22
  # TXEN
  TXen: RADIOLIB_NC

  gpiochip: 4
  spidev: spidev3.0

Logging:
  LogLevel: trace
  AsciiLogs: true

General:
  MACAddressSource: wlan0
```
The PIN number in case of native is anyway a bit odd, as it does not really mean PIN no. but rather line. Thus, if there is a mismatch between both, one has to use line in Linux (i.e. Meshtastic native environment).

With this patch, I propose to extend this format to allow for each PIN to specify its gpiochip and line. The old format is kept for backwards compatibility, while the current gpiochip can be considered as 'fallback' or 'globally gpiochip'. Thus, with this given patch, the above format is supported as well as the below:
```
Lora:
  Module: sx1262
  DIO2_AS_RF_SWITCH: true
  DIO3_TCXO_VOLTAGE: 1.8

  # NSS     PIN_24 -> chip 4, line 22
  CS: 
    pin: 24
    gpiochip: 4
    line: 22
  # SCK     PIN_23 -> chip 4, line 18
  SCK: 
    pin: 23
    gpiochip: 4
    line: 18
  # BUSY    PIN_29 -> chip 3, line 11 !
  Busy:
    pin: 29
    gpiochip: 3
    line: 11
  # MOSI    PIN_19 -> chip 4, line 19
  MOSI: 
    pin: 19
    gpiochip: 4
    line: 19
  # MISO    PIN_21 -> chip 4, line 21
  MISO: 
    pin: 21
    gpiochip: 4
    line: 21
  # NRST    PIN_27 -> chip 4, line 10
  Reset: 
    pin: 27
    gpiochip: 4
    line: 10
  # DIO1    PIN_28 -> chip 4, line 11
  IRQ: 
    pin: 28
    gpiochip: 4
    line: 11

  # RXEN    PIN_22 -> chip 3, line 17 !
  RXen: 
    pin: 22
    gpiochip: 3
    line: 17
  # TXEN
  TXen: RADIOLIB_NC

  gpiochip: 4
  spidev: spidev3.0

Logging:
  LogLevel: trace
  AsciiLogs: true

General:
  MACAddressSource: wlan0
```
You can notice, that TXen still employs the old format, which the YAML parser handles fine.

But how does that work, and why is it just a matter of parsing a YAML and enabling it? Already today, the portduino framework supports the handling of gpiochip and line per PIN. In particular the functions `LinuxGPIOPin(pin_size_t n, const char *chipLabel, const int linuxPinNum, const char *portduinoPinName);` (https://github.com/meshtastic/framework-portduino/blob/master/cores/portduino/linux/gpio/LinuxGPIOPin.h#L40), where `n` is employed for the PIN, `linuxPinNum` used for the `line` within Linux, while `chipLabel` obviously is meant for the gpiochip name. The last parameter `portduinoPinName` is an artificial name to make portduino happy, which could be `n` as string.
The magic of setting portduino properly up in regards to the mapping is subsequently taken care of in portduinos `void gpioBind(GPIOPinIf *p)` (https://github.com/meshtastic/framework-portduino/blob/master/cores/portduino/PortduinoGPIO.cpp#L38). Luckily, this is handled already well in `src/platform/portduino/PortduinoGlue.cpp` -> `int initGPIOPin(int pinNum, const std::string gpioChipName)`.

Thus it is just a matter of enhancing `src/platform/portduino/PortduinoGlue.cpp` -> `int initGPIOPin(int pinNum, const std::string gpioChipName)`. As one can see, I extended `initGPIOPin` with the parameter `line` and reworked the YAML parsing, while keeping the backwards compatibility. To ensure, that all affected spots were detected, the individual enum such as `cs` was reworked to `cs_pin`, `cs_line` and `cs_gpiochip` to follow the current implementation pattern.

In light of Meshtastic, the patch and the new config format mean, that Meshtastic internally uses the PIN no., which must not be the true line number. While portduino was made aware of the mapping of PIN no. to gpiochip and line, and thus can perform a seemless mapping. For the new config, it is solely crucial that the PIN no. is unique, otherwise the mapping won't work properly in portduino. Meanwhile, the PIN no. could be even artificially chosen.

Last, I reworked code that was multiple times copied into structs and foreachs.

With this, not only Radxa Zero 3E/W can perform a native meshtastic with the above (2nd) `config.yaml`, but one could also employ long term multiple gpiochips of other sorts for a chosen LoRa chip.


psiegl:
Signed the CLA

src/main.cpp
```diff
@@ -825,116 +825,62 @@ void setup()
...
+    const struct { configNames cfgName;
+                   std::string strName;
+                   bool ch341halsupport;
+    } loraModules[] = {
+      { use_rf95, "RF95", false },
+      { use_sx1262, "sx1262", true },
+      { use_sx1268, "sx1268", false },
+      { use_sx1280, "sx1280", false },
+      { use_lr1110, "lr1110", false },
+      { use_lr1120, "lr1120", false },
```
caveman99, line 837: I know this was not in the old code, but for the sake of completeness, could you add a handler for the LLCC68 here? this way we have a 1:1 mapping of all supported radiochips.

psiegl: Added LLCC68 support.

psiegl:
Sadly I can't get the build green right now. Seems some like to push to main broken code/tests :-( PRs would be better, instead of pushing to main^^

jp-bennett:
> Sadly I can't get the build green right now. Seems some like to push to main broken code/tests :-( PRs would be better, instead of pushing to main^^

Sorry about that. It's Github CI work. Pushing to master is the only way to test it.

The PR here looks great. I'll test drive it on my various rigs as soon as I can.

psiegl:
> > Sadly I can't get the build green right now. Seems some like to push to main broken code/tests :-( PRs would be better, instead of pushing to main^^
> 
> Sorry about that. It's Github CI work. Pushing to master is the only way to test it.
> 
> The PR here looks great. I'll test drive it on my various rigs as soon as I can.

any update?

src/main.cpp
```diff
@@ -826,116 +826,63 @@ void setup()
 #endif
 
 #ifdef ARCH_PORTDUINO
-    if (settingsMap[use_sx1262]) {
-        if (!rIf) {
-            LOG_DEBUG("Activate sx1262 radio on SPI port %s", settingsStrings[spidev].c_str());
-            if (settingsStrings[spidev] == "ch341") {
+    const struct { configNames cfgName;
+                   std::string strName;
+                   bool ch341halsupport;
```
jp-bennett, line 831: There's nothing inherently preventing the other chips from having ch341 hal support, just not any available hardware yet. That part can be removed to simplify this code a bit.

src/platform/portduino/PortduinoGlue.cpp
```diff
@@ -376,42 +360,65 @@ bool loadConfig(const char *configPath)
...
+                { sx126x_ant_sw_pin, sx126x_ant_sw_gpiochip, sx126x_ant_sw_line, "SX126X_ANT_SW" },
+            };
+            for (auto& pinMap : pinMappings) {
+                if (yamlConfig["Lora"][pinMap.strName].IsMap()
+                    && (yamlConfig["Lora"][pinMap.strName]["pin"]
+                      || yamlConfig["Lora"][pinMap.strName]["line"]
+                      || yamlConfig["Lora"][pinMap.strName]["gpiochip"])) {
+                    settingsMap[pinMap.pin] = yamlConfig["Lora"][pinMap.strName]["pin"].as<int>(RADIOLIB_NC);
+                    settingsMap[pinMap.line] = yamlConfig["Lora"][pinMap.strName]["line"].as<int>(settingsMap[pinMap.pin]);
+                    settingsMap[pinMap.gpiochip] = yamlConfig["Lora"][pinMap.strName]["gpiochip"].as<int>(defaultGpioChip);
```
jp-bennett, line 413: It would be great to include an example in the default config.yaml of this new approach.

jp-bennett (changes requested):

jp-bennett:
I like it! There's a minor change I think makes sense, and then there needs to be some sort of example in the pre-supplied config.yaml file to users know how to use the new approach.

src/main.cpp
psiegl: Fixed this part. I think in the long run, the CH341 SPI check needs to reworked anyway. What it truly tries to do: check for an user-space driver, which in this case is CH341. I am playing currently a bit with the [MCP2210](https://www.microchip.com/en-us/product/mcp2210), which is a bit more expensive but more adjustable. However, in case on wants to use the MCP2210 in the userspace, the CH341 flag won't be functioning that well anymore.

src/platform/portduino/PortduinoGlue.cpp
psiegl: Added the example to the `bin/config-dist.yaml`

psiegl:
> I like it! There's a minor change I think makes sense, and then there needs to be some sort of example in the pre-supplied config.yaml file to users know how to use the new approach.

done :-)

jp-bennett (approved):

PR #5749, opened 2025-01-05
Improve slot time calculation and adjust contention window size accordingly

GUVWAF:
Since the initial slot time calculation, the default number of symbols used for Channel Activity Detection (CAD) in RadioLib has changed. Therefore, our slot time is actually quite some lower, namely 28ms versus 77ms as calculated now on LongFast, which corresponds to the measurements I did recently (~26ms from `isChannelActive()` until after `startTransmit()`). To avoid this being changed again in the future, I now explicitly set the number of CAD symbols. For SX127x this is not possible, and the CAD duration is known (max. 2.25 symbols). For SX1280, there is a different calculation and the minimum recommended number of symbols is 4, so I applied this calculation when using 2.4GHz LoRa.

A reduced slot time means we can use more slots for the same delay, which should lower the chance of collisions. The slot time determines the maximum delay a node uses before rebroadcasting, depending on SNR. To make sure this is still roughly the same across firmware versions, I adjusted `CWmin` and `CWmax`, and I also decreased `SNR_MAX` to 10dB since I believe that’s closer to the actual maximum SNR. The plot below shows the max. delay versus SNR on LongFast for the different firmware versions (in 2.5 we fixed the slot time calculation #4469). 
image
Even though the adjustments make sure the curve is close to the <=2.4 values, I think it’s good to merge this **only when we release a new minor version**.

### DIFF
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -256,10 +256,17 @@ template <typename T> void LR11x0Interface<T>::startReceive()
 template <typename T> bool LR11x0Interface<T>::isChannelActive()
 {
     // check if we can detect a LoRa preamble on the current channel
+    ChannelScanConfig_t cfg = {.cad = {.symNum = NUM_SYM_CAD,
+                                       .detPeak = RADIOLIB_LR11X0_CAD_PARAM_DEFAULT,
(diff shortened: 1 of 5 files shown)


thebentern (approved):

caveman99 (approved):

thebentern:
@GUVWAF when you get a chance, reopen this one. I accidentally merged prematurely and reverted upstream

PR #5772, opened 2025-01-06
[MERGE ONLY FOR 2.6] Improve slot time calculation and adjust contention window size accordingly

GUVWAF:
Re-try for #5749. Only merge when releasing 2.6.


thebentern:
Yes. I was actually going to ask you if we should go ahead and merge the Next-hop routing into there as well

GUVWAF:
Sounds good to me.

PR #5773, opened 2025-01-06
fix esp32c3 clock speed

macvenez:
ESP32C3 max clock speed is 160MHz, but in firmware we were setting it to 240MHz, I don't think it may cause any issues but it should be safer to set it to the right clock speed

### DIFF
--- a/src/sleep.cpp
+++ b/src/sleep.cpp
@@ -71,8 +71,8 @@ void setCPUFast(bool on)
          * (Added: Dec 23, 2021 by Jm Casler)
          */
 #ifndef CONFIG_IDF_TARGET_ESP32C3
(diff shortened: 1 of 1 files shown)


src/sleep.cpp
```diff
@@ -71,8 +71,8 @@ void setCPUFast(bool on)
          * (Added: Dec 23, 2021 by Jm Casler)
          */
 #ifndef CONFIG_IDF_TARGET_ESP32C3
```
GUVWAF, line 73: This `#ifndef` means it already shouldn't set this for ESP32C3.

caveman99:
This will reduce the clockspeed for any other target to 160 MHz. Which is not desired behaviour.

PR #5774, opened 2025-01-06
Increase esp32c3 stability over wifi

macvenez:
I had many issues with wifi stability on my Heltec HT-CT62 (esp32c3 based).
Initially I thought it was related to hardware design but I checked multiple times and it was following design recommendations.

I also tried adding `WiFi.setTxPower(WIFI_POWER_8_5dBm);`  which was needed due to a hardware issues in earlier esp32c3 models but did no difference.

In the end I tried with `WiFi.useStaticBuffers(true);` and it was finally working as it should, no more wifi disconnections or intermittent loading.

I would be happy if someone with same issues could try this changes and report if working but that's been game changing for me

### DIFF
--- a/src/mesh/wifi/WiFiAPClient.cpp
+++ b/src/mesh/wifi/WiFiAPClient.cpp
@@ -143,6 +143,9 @@ static int32_t reconnectWiFi()
         delay(5000);
 
         if (!WiFi.isConnected()) {
+            WiFi.mode(WIFI_MODE_NULL);
+            WiFi.useStaticBuffers(true);
+            WiFi.mode(WIFI_STA);
             WiFi.begin(wifiName, wifiPsw);
         }
         isReconnecting = false;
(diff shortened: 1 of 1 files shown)


src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -143,6 +143,9 @@ static int32_t reconnectWiFi()
         delay(5000);
 
         if (!WiFi.isConnected()) {
+            WiFi.mode(WIFI_MODE_NULL);
+            WiFi.useStaticBuffers(true);
```
GUVWAF, line 147: This is not a known method for RP2040.

GUVWAF:
Interesting, not sure if it's desirable to set this also for all other ESP32, but if it increases stability for those also, maybe we should.

src/mesh/wifi/WiFiAPClient.cpp
macvenez: I missed it (again) but I think I should add a ifdef for esp32c3 boards only. I never played enough with esp32s3 to tell if it's a issue there too...

caveman99 (approved):

thebentern (approved):

PR #5779, opened 2025-01-07
Hello world support for UDP broadcasts over the LAN on ESP32

thebentern:
Couple of remarks:
I have this hardcoded to the address `IPAddress(224, 0, 0, 69)`, which seems to be fine (for now). I have also pegged the port to `4403` because we know this port already on the TCP api side of things.

I would like to test this out over something like a Tailscale VPN to see if it can work over such network boundaries potentially. I'm not too well versed in lower level socket communications, so we'll see. I'd also like to make a client proxy mode for this too, so that we can enable BLE and serial only devices to play as well.
image

The easiest way to test this locally is to join a couple of ESP32 nodes to your Wifi and set `lora.tx_enabled` to false

### DIFF
--- a/arch/esp32/esp32.ini
+++ b/arch/esp32/esp32.ini
@@ -37,6 +37,7 @@ build_flags =
   -DLIBPAX_ARDUINO
   -DLIBPAX_WIFI
   -DLIBPAX_BLE
+  -DHAS_UDP_MULTICAST=1
   ;-DDEBUG_HEAP
 
 lib_deps =
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -274,6 +274,12 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
             abortSendAndNak(encodeResult, p);
             return encodeResult; // FIXME - this isn't a valid ErrorCode
         }
+#if HAS_UDP_MULTICAST
+        if (udpThread && isFromUs(p) &&
+            config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST) {
+            udpThread->onSend(const_cast<meshtastic_MeshPacket *>(p));
+        }
+#endif
 #if !MESHTASTIC_EXCLUDE_MQTT
         // Only publish to MQTT if we're the original transmitter of the packet
         if (moduleConfig.mqtt.enabled && isFromUs(p) && mqtt) {
@@ -624,6 +630,12 @@ void Router::handleReceived(meshtastic_MeshPacket *p, RxSource src)
         // After potentially altering it, publish received message to MQTT if we're not the original transmitter of the packet
         if ((decoded || p_encrypted->pki_encrypted) && moduleConfig.mqtt.enabled && !isFromUs(p) && mqtt)
             mqtt->onSend(*p_encrypted, *p, p->channel);
+#endif
+#if HAS_UDP_MULTICAST
+        if ((decoded || p_encrypted->pki_encrypted) && !isFromUs(p) && udpThread &&
+            config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST) {
+            udpThread->onSend(const_cast<meshtastic_MeshPacket *>(p_encrypted));
+        }
 #endif
     }
 
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -275,8 +275,7 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
             return encodeResult; // FIXME - this isn't a valid ErrorCode
         }
 #if HAS_UDP_MULTICAST
-        if (udpThread && isFromUs(p) &&
-            config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST) {
+        if (udpThread && config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST) {
             udpThread->onSend(const_cast<meshtastic_MeshPacket *>(p));
         }
 #endif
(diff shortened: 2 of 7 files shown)


arch/esp32/esp32.ini
```diff
@@ -37,6 +37,7 @@ build_flags =
   -DLIBPAX_ARDUINO
   -DLIBPAX_WIFI
   -DLIBPAX_BLE
+  -DHAS_UDP_MULTICAST=1
```
caveman99, line 40: absolutely ok for testing, later i would always enable that if we have an IP interface (wifi, eth)

src/mesh/udp/UdpMulticastThread.h
```diff
@@ -0,0 +1,70 @@
...
+        size_t encodedLength = pb_encode_to_bytes(buffer, sizeof(buffer), &meshtastic_MeshPacket_msg, mp);
+        udp.broadcastTo(buffer, encodedLength, UDP_MULTICAST_DEFAUL_PORT);
+        return true;
+    }
+
+  protected:
+    int32_t runOnce() override
+    {
+        canSleep = true;
+        // TODO: Might consider a heartbeat for discovery or keep alive?
```
caveman99, line 62: I would align this with the store&forward heartbeat mechanism. Could also include a display symbol for the available transports in the long term.

caveman99 (approved):

src/mesh/Router.cpp
```diff
@@ -624,6 +630,12 @@ void Router::handleReceived(meshtastic_MeshPacket *p, RxSource src)
         // After potentially altering it, publish received message to MQTT if we're not the original transmitter of the packet
         if ((decoded || p_encrypted->pki_encrypted) && moduleConfig.mqtt.enabled && !isFromUs(p) && mqtt)
             mqtt->onSend(*p_encrypted, *p, p->channel);
+#endif
+#if HAS_UDP_MULTICAST
+        if ((decoded || p_encrypted->pki_encrypted) && !isFromUs(p) && udpThread &&
```
GUVWAF, line 635: I think this logic should be different than for MQTT. The reason for MQTT it is also sending everything it receives is because MQTT is used not only for transport (bridging) but also as client. So a packet directed towards you, which is not rebroadcasted, also ends up on the MQTT server. For UDP I think we should only use `onSend()` when the Router sends, and thus remove the `isFromUs()` there.

GUVWAF:
Very cool. Eventually we could use this also for control packets like Request-To-Send to better coordinate between multiple nodes without LoRa overhead.

There's some tricky things here since we have only one `Router`, so the logic for, e.g., retransmissions or (refraining from) rebroadcasting is influenced by all protocols, even if for pure LoRa transport you didn't receive any implicit ACK (same is true for MQTT now).

src/mesh/Router.cpp
thebentern: I was thinking about this as well. It makes sense to remove the isFromUs check

GUVWAF: I meant to remove `isFromUs(p)` from `Router::send()` and remove the complete block out of `Router::handleReceived()` (since packets received via UDP already arrive here). This way it only sends via UDP if it would via LoRa.
However, if you want it to always rebroadcast over UDP (without decrementing the hop limit, before adding yourself to the traceroute, etc.) then we could do it the same as for MQTT. So depends on whether we see this as a normal hop or more a virtual one.

Talie5in:
Looks like this should be able to work with the RAK13800 ETH Board as well?

https://github.com/RAKWireless/RAK13800-W5100S/blob/main/examples/RAK13800_Ethernet_UDP_W5100S/RAK13800_Ethernet_UDP_W5100S.ino

If so, and how i am reading this PR is... ability to say join, 1-2 or 3 nodes on the same network and allow them to send the packets to/from each other to pass messages around the network? Is that the intention or just one of?

GUVWAF (approved):

src/mesh/udp/UdpMulticastThread.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+#if HAS_UDP_MULTICAST
+#include "configuration.h"
+#include "main.h"
+#include "mesh/Router.h"
+
+#include <AsyncUDP.h>
+#include <WiFi.h>
+
+#define UDP_MULTICAST_DEFAUL_PORT 4403 // Default port for UDP multicast is same as TCP api server
```
kahara, line 10: `UDP_MULTICAST_DEFAULT_PORT`, maybe

T-X:
Many thanks for working on and enabling this!

If you're still looking for a way to test this over a VPN, let me know. At Freifunk we have (typically WLAN based) mesh nodes (usually) running the multicast group-aware B.A.T.M.A.N. Advanced mesh protocol. This protocol will ensure to only forward multicast packets to nodes which have indicated interest in it (or not forward a packet at all, if there is no interested node). And the firmware supports VPNs, too, and you can use multicast over the VPN as well.

The caveat however is that this Freifunk firmware only supports IPv6 multicast / MLD at the moment. (but in theory, if I read #7951 correctly it should be possible to enable IPv6 multicast on a meshtastic ESP32 device?)

T-X:
Also, IPv6 multicast with a group-aware/MLD snooping network support would allow forwarding a packet by taking the meshtastic topic into account. For instance by hashing the topic to a (mostly) unique IPv6 multicast address, thanks to the much greater address space with IPv6. This would again reduce the overall overhead on such an IP/ethernet/Freifunk/batman-adv mesh network, by only forwarding packets to other nodes which are interested in it.

T-X:
(We could also add an exception to the firewall in our [Freifunk Gluon](https://github.com/freifunk-gluon/gluon/) firmware, to allow these specific IPv4 multicast packets through. But they would then just be flooded through our entire Freifunk mesh network, without optimized paths / without IGMP snooping. But maybe that's fine for now as meshtastic shouldn't generate too much traffic anyway?)
(Edit: e.g. here: https://github.com/freifunk-gluon/gluon/tree/main/package/gluon-ebtables-filter-multicast/luasrc/lib/gluon/ebtables)

PR #5785, opened 2025-01-07
meshtasticd-debian: Fix Beta promotions, deb versioning compliance, add OBS

vidplace7:
Fix version numbers to be compliant with debian `3.0 (native)` format.
These version compliance issues manifest as warnings in Launchpad PPAs, but are fatal in OpenSUSE OBS.

Adds initial (but not automated) support for OBS build system. Should generate source debs that are compatible with OBS, but does not *yet* automatically upload them.

Testing here https://build.opensuse.org/project/show/home:vidplace7

### DIFF
--- a/.github/workflows/nightly_debian.yml
+++ b/.github/workflows/nightly_debian.yml
@@ -1,0 +1,37 @@
+name: Nightly Debian Packaging
+on:
+  schedule:
+    - cron: 0 9 * * *
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths:
+      - debian/**
+      - .github/workflows/nightly_debian.yml
+      - .github/workflows/build_debian_src.yml
+      - .github/workflows/package_ppa.yml
+      - .github/workflows/package_obs.yml
+
+permissions:
+  contents: write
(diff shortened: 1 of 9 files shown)


.github/workflows/release_channels.yml
```diff
@@ -2,7 +2,7 @@ name: Trigger release workflows upon Publish
 
 on:
   release:
-    types: [published]
+    types: [published, released]
```
vidplace7, line 5: Should correctly capture Beta "promotions", as `released` is triggered when an existing build is changed from a pre-release to a release.

.github/workflows/nightly_debian.yml
```diff
@@ -0,0 +1,37 @@
+name: Nightly Debian Packaging
+on:
+  schedule:
+    - cron: 0 9 * * *
+  workflow_dispatch: {}
+  push:
+    branches:
+      - master
+    paths:
+      - debian/**
```
vidplace7, line 10: This will run nightly *and* whenever a change is made to the debian packaging itself.

fifieldt (approved):

PR #5790, opened 2025-01-08
Initial commit of a fuzzer for Meshtastic.

esev:
This PR implements a fuzzer using the ClusterFuzzLite framework. More background about the steps required to generate a fuzzer can be found in the [Build Integration Guide for ClusterFuzzLite](https://google.github.io/clusterfuzzlite/build-integration/).

Some things to note:
* ClusterFuzzLite requires the binary to be build with specific compiler options and use the clang/llvm tool chain. I've implemented this using the `buildroot` variant, which provides a convenient way to override the build tools used for PlatformIO.
* A ` FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` CPP define is used for enabling features needed for fuzzing. I've tried to make as few changes as necessary to keep the code mostly the same while running under the fuzzer.
* `.__cxa_exception_type()` is not implemented in clang. I've implemented the solution in https://stackoverflow.com/a/47164539/4613163 within PortduinoGlue.cpp.
* clang also produces an error when a type is being narrowed (a uint64 being assigned to a uint32). A `static_cast<uint32_t>(value)` satisfies this error.
* I've only gotten the fuzzer working with the Address [Sanitizer](https://google.github.io/clusterfuzzlite/overview/#sanitizers). I've also tried the Memory Sanitizer, but it'll require recompiling a few more dependencies using clang and the fuzzer provided build flags.  

I've included a README and tried to document things so it is clear how the fuzzer works and what all the files are used for. If anything is unclear, please ask and I'll improve the comments/documentation in this PR.


fifieldt (approved):

fifieldt:
This is advanced stuff, we're very lucky to have you. Thanks for the huge effort.

esev:
Thank you @fifieldt ! To be fair this was a lot of trial and error to get it going too. :)

I added a limit on the TypedQueue class used by the native/portduino build. Using 0 still allows the original behavior of a limitless queue. Let me know if you'd prefer that change to be guarded by `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` instead.

fifieldt (approved):

esev:
Still to-do:
- The fuzzer doesn't shut down cleanly on exit. I need to figure out how to make that work.
- Needs [Github actions](https://google.github.io/clusterfuzzlite/running-clusterfuzzlite/github-actions/) to run the fuzzer on PRs and on a schedule for longer runs.

The fuzzer is functional for anyone who wants to run it locally. It just can't be run via an action until it shuts down cleanly. If this PR looks good I think it can be merged. I'll keep working on this and either add more commits to this PR or send new PRs with the to-do items.

esev:
#5807 should resolve `Have the fuzzer shut down cleanly on exit without crashing`

PR #5797, opened 2025-01-09
Add GPS capability to RAK2560 (RAKwireless WisMesh Hub)

beegee-tokyo:
The RAK2560 WisMesh Hub can support GPS, but the serial port is Serial2, not Serial1.

This merge request adds a #ifdef to gps.cpp to select the correct serial port.

Even without a GPS module on board (it is optional), with the GPS support enabled, it is possible to set a "fake" location, which can be helpful for the building of coverage maps.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -35,7 +35,11 @@ template <typename T, std::size_t N> std::size_t array_count(const T (&)[N])
 }
 
 #if defined(NRF52840_XXAA) || defined(NRF52833_XXAA) || defined(ARCH_ESP32) || defined(ARCH_PORTDUINO)
+#if defined(RAK2560)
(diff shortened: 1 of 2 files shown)


thebentern (approved):

thebentern:
> Even without a GPS module on board (it is optional), with the GPS support enabled, it is possible to set a "fake" location, which can be helpful for the building of coverage maps.

You can also set a fixed position in the python CLI and the apps.

beegee-tokyo:
> > Even without a GPS module on board (it is optional), with the GPS support enabled, it is possible to set a "fake" location, which can be helpful for the building of coverage maps.
> 
> You can also set a fixed position in the python CLI and the apps.

I tried that, but with the GPS.cpp missing, it does not report location, even if I set a fixed location.

PR #5801, opened 2025-01-09
Added illuminance sensors value on the environmental sensors screen (lux and white_lux)

isseysandei:
Added lux and white_lux to the environment sensors screen


isseysandei:
Hmmm what's happening here? Why is it failing now?
Also, can someone help me fixing the "buildtool didn't sign" warning?

fifieldt:
The formatting in src/modules/Telemetry/EnvironmentTelemetry.cpp
Doesn't match what is expected by our linting tool. If you install trunk in vscode it should take care of it.

isseysandei:
> The formatting in src/modules/Telemetry/EnvironmentTelemetry.cpp Doesn't match what is expected by our linting tool. If you install trunk in vscode it should take care of it.

I'm having problems with trunk, it outputs this error:
`trunk fmt --all --filter clang-format
Checking 50% [==========================================================================================>                                                                                          ]  2/4  2.7s 
 ↳ trunk
   ↳ Scanning the filesystem ⠸
✖ Unable to download plugin https://github.com/trunk-io/plugins/archive/v1.2.1.zip: string pointer is null`

I seen this same issue around some forums, but zero answer... Any idea to solve it? thanks!

Issue #5404, opened 2024-11-20
[Feature Request]: Don't uplink decrypted precise location to MQTT


juliatuttle:
### Platform

Cross-Platform

### Description

As part of tightening handling of precise location as PII, when uplinking a position packet to MQTT:
- If it's decrypted (protobuf or JSON), blur the location to imprecise before uplinking.
- If it's encrypted with a non-default key, consider it secured in transit between users of that key, and uplink it intact.
- If it's encrypted with a default key, it's not effectively secured, but it can't be blurred as-is, so you'd need to either drop it entirely, decrypt/blur/uplink it decrypted, or decrypt/blur/encrypt/uplink it encrypted. The last is the most complex, but would cause the least change to the current behavior.

It's worth considering whether this behavior should differ when uplinking to private brokers; my instinct is that it should apply to private brokers as well, since part of the concern that inspired this bug was that data can be relayed between (some) brokers.

thebentern:
We already do this on the official broker. I think the better approach here is to hard enforce the firmware not uplinking of MQTT packets without the OkToMQTT bit flag, under explicit consent. This will both give users the choice to opt-in and the freedom to continue to uplink precise data to private brokers as desired. Removing this functionality would be a very controversial change.

juliatuttle:
My understanding from chatting with Garth was that relying on "OK to MQTT" as consent for the app/node to transmit precise location on default keys and for any other user to relay that precise location to MQTT was insufficient, for two reasons:
1. Practically, users are apparently "shocked" to see their precise location visible to anyone on the Internet and collected by third-party maps. This isn't surprising, as users have to build a mental model of how the GPS, channel, precision, and MQTT settings interact before they can understand the impact of their choices!
2. Legally, Meshtastic is responsible for safeguarding the PII they collect in their app, and for ensuring anyone they send that data to will safeguard it in turn. Part of that safeguarding is allowing users to delete collected Pii, which is entirely infeasible when it's visible in plaintext to anyone nearby. This is unfortunate given Meshtastic's open nature, but to many users, it's a discrete product that should follow their privacy rights/expectations.

thebentern:
> Legally, Meshtastic is responsible for safeguarding the PII they collect in their app, and for ensuring anyone they send that data to will safeguard it in turn.

[...]

PR #5805, opened 2025-01-10
fix(mqtt): prioritize respecting ok-to-mqtt

ctso:
If I'm reading this code correctly, if an MQTT server resolves to an RFC1918 address the packet will always be forwarded to the broker, even if the "OK to MQTT" flag is set to `false`.  

This smells like a bit of a privacy concern.  If users have explicitly opted out of having their packets published to an MQTT broker, it shouldn't matter if that server is resolving to an RFC1918 address or not.  It's very trivial to run an MQTT broker that is publicly accessible, but configure your client to post to this via a simple TCP proxy that suddenly violates the "OK to MQTT" flag that other clients may have specified.

It's possible this logic exists for some other reason that I don't understand, so happy to close this if I'm wrong here.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -45,8 +45,6 @@ constexpr int reconnectMax = 5;
 // FIXME - this size calculation is super sloppy, but it will go away once we dynamically alloc meshpackets
 static uint8_t bytes[meshtastic_MqttClientProxyMessage_size + 30]; // 12 for channel name and 16 for nodeid
 
-static bool isMqttServerAddressPrivate = false;
-
 inline void onReceiveProto(char *topic, byte *payload, size_t length)
 {
     const DecodedServiceEnvelope e(payload, length);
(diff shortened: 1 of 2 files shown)


Talie5in:
@ctso I feel like this comment on another issue is relevant by @thebentern 

https://github.com/meshtastic/firmware/issues/5404#issuecomment-2489883180

And they are right, there are forks out there already that already ignore the OktoMqtt flag, or reversed the NeighbourInfo change so it still goes out the main public channels as another example.

Ultimately though, I think the only change here I think possibly one of the ANDs should probably be an OR,  or checked separately to if its a RFC1918 address or not.

thebentern:
@Talie5in is correct. The local address broker being the exception to the mqtt uplink was the concession by popular demand. This may change in a later version ultimately, but for now, this is by design. And as was mentioned, the bit remains a polite request to not uplink which does not prevent modified firmwares from trivially uplinking to even public brokers

ctso:
Thanks, I may follow up with another PR here to update the docs a bit to call this out.  I realize they currently say that this is best effort, but it may be nice to also call out the fact that the MQTT-publishing behavior is different when your broker is on an RFC1918 address.  This is valuable to understand for operators of MQTT devices.  I only realized this trying to troubleshoot why some of my messages were not making it up to an MQTT broker :)

PR #5808, opened 2025-01-10
Heltec Wireless Stick Lite V1/V2 support

DarkZeros:
I only have the V2.1 version, not sure if the HW is same with V1, or V2.
Given the few resources I could find I think it is.

ADC / Battery measure and TX/RX are working for my boards

The board uses ESP32-PICO-D4, and has a SX1272 Radio.

I do not know the official way to add HW support, I could have added support as UNKNOWN, which is better than nothing, but I also did the changes to the protobuf, not sure if it is the right way to do.

I can update my PR if I get some advice on it.

### DIFF
--- a/variants/heltec_wsl_v2.1/platformio.ini
+++ b/variants/heltec_wsl_v2.1/platformio.ini
@@ -1,0 +1,7 @@
+[env:heltec-wsl-v2_1]
+extends = esp32_base
+board = heltec_wireless_stick_lite
+board_level = extra
+build_flags =
+  ${esp32_base.build_flags} -D PRIVATE_HW -I variants/heltec_wsl_v2.1
+  -DGPS_POWER_TOGGLE ; comment this line to disable triple press function on the user button to turn off gps entirely.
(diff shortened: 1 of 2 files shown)


caveman99:
I am hesitant to add these legacy devices. They have been superseeded by V3 which is acively sold.

thebentern:
> Thanks for your submission and work, however I am hesitant to add these legacy devices. They have been superseeded by V3 which is acively sold.

I agree. I'm okay with adding the variants, but this would not get it's own HW_MODEL as it is legacy. It should be marked as DIY hardware model and `board_level = extra` in platformio.ini so that folks can compile it for themselves, but it won't be packaged in the releases

DarkZeros:
> > Thanks for your submission and work, however I am hesitant to add these legacy devices. They have been superseeded by V3 which is acively sold.
> 
> I agree. I'm okay with adding the variants, but this would not get it's own HW_MODEL as it is legacy. It should be marked as DIY hardware model and `board_level = extra` in platformio.ini so that folks can compile it for themselves, but it won't be packaged in the releases

Fully agree, being an old model is good to merge the support files to build in case someone has one around. But not make it into the official HW numbers. I am updating the PR to reflect that.

fifieldt (approved):

fifieldt:
Thanks so much for being a great first time contributor!

PR #5821, opened 2025-01-11
Support for Polish fonts on E-Ink devices, Polish fonts retouch, fixed Czech/Slovak OLED/E-Ink double space bug

And137:
Tested and working.

CS OLED before:
image

and after:
image


And137:
I only forgot to comment the -D OLED_CS=1 flag on platformio.ini, so i's better to leave the original platformio.ini from the master branch.

And137:
image

kyberpunk:
Hi @And137 , thanks for fixing that, I did the fix also in this MR #5695, so I just rebase it.

PR #5822, opened 2025-01-12
Cache Python & PlatformIO dependencies

esev:
There is one cache for Python/pip files and a cache per architecture for the PlatformIO downloads. This saves a few minutes for the CI workflow when the caches are present.

The caches are invalidated on changes to the .github/actions or any of the PlatformIO .ini files. The total cache size is 2.2G.

### DIFF
--- a/.github/actions/build-variant/action.yml
+++ b/.github/actions/build-variant/action.yml
@@ -68,6 +68,12 @@ runs:
           sed -i '/DDEBUG_HEAP/d' ${INI_FILE}
         done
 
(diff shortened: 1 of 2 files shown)


thebentern (approved):

esev:
That's weird. Why are these all failing? I don't even see my changes in the logs for this PR (which is expected).

esev:
I suspect we're hitting the issue described [here](https://github.com/actions/runner/issues/2816). There are multiple checkouts in this composite. Maybe the workflow file in the "Post" step has picked-up my change from this PR and is out-of-sync with the master branch?

PR #5824, opened 2025-01-12
Revert "Cache Python & PlatformIO dependencies"

esev:
Reverts meshtastic/firmware#5822

Still fails on the master branch.


fifieldt (approved):

esev:
I hadn't noticed that custom runners were being used. The caching in the setup-python step doesn't seem compatible.

esev:
Instead of reverting, the issue might be resolved by https://github.com/meshtastic/firmware/pull/5824

PR #5827, opened 2025-01-12
Bugfix for low-priority packet replacement when TX queue is full

### LINKED ISSUE #5826: [Bug]: ROUTER_LATE on LongFast - Packets sometimes dont TX
### Category

Other

### Hardware

Rak4631, Linux Native

### Firmware Version

2.5.18.89ebafc

### Description

Seeing issues where a Packet wont TX in ROUTER_LATE under LongFast if a second or additional duplicate packet comes in and it pushed the packet wait out, we don't eventuate to a "[RadioIf] Started Tx"

If however, we don't see a duplicate packet, it will queue for TX.

[DoesTX.txt](https://github.com/user-attachments/files/18387933/DoesTX.txt)
[NeverTXd.txt](https://github.com/user-attachments/files/18387934/NeverTXd.txt)

Started discussing this with @erayd in Discord, Also tested under the 2.5.15-Router_late release from Erayd's original source tree/branch.

### Relevant log output

_No response_

erayd:
Add the intended replacement packet (not the packet that was just removed, which was clearly a typo) to the TX queue after dropping a lower-priority packet. Fixes a use-after-free.

Also corrects the function comment, and adds some debug logging when packets are dropped due to a full TX queue.

Fixes #5826.


fifieldt (approved):

erayd:
Somehow `replaceLowerPriorityPacket` is getting called twice for the same packets. I do not yet know why. That should never happen...

Screenshot 2025-01-12 at 17 46 04

src/mesh/MeshPacketQueue.cpp
```diff
@@ -139,8 +147,11 @@ bool MeshPacketQueue::replaceLowerPriorityPacket(meshtastic_MeshPacket *p)
         for (; refPacket->tx_after && it != queue.begin(); refPacket = *--it)
             ;
         if (!refPacket->tx_after && refPacket->priority < p->priority) {
+            LOG_WARN("Dropping non-late packet 0x%08x to make room in the TX queue for higher-priority packet 0x%08x",
+                     refPacket->id, p->id);
             packetPool.release(refPacket);
```
esev, line 152: There is no call to erase the refPacket from the queue. If the queue was full, this will loop back into replace lower priority packet.

src/mesh/MeshPacketQueue.cpp
```diff
@@ -124,9 +131,10 @@ bool MeshPacketQueue::replaceLowerPriorityPacket(meshtastic_MeshPacket *p)
     // Check if the packet at the back has a lower priority than the new packet
     auto &backPacket = queue.back();
```
esev, line 132: I suggest removing the reference here. Maybe auto *backPacket to make it clearer what the type is.
Calling pop_back while holding a reference might not always be safe.

src/mesh/MeshPacketQueue.cpp
erayd: Thank you. That explains the loop we were seeing earlier 🙂

esev: No worries! It took me a few times to spot it as well. I kept assuming packetPool.release was taking care removing it.

erayd: Resolved in [11dc10b](https://github.com/meshtastic/firmware/pull/5827/commits/11dc10b1e0219a55d33e9650666f1eeaba31de79).

Talie5in:
[TXFillsUp-71ef9c9.txt](https://github.com/user-attachments/files/18389221/TXFillsUp-71ef9c9.txt)

Logs testing after the last commit (from priority-dropfix-2.5.18)

erayd:
Looks like there's also an issue where, if `setTransmitDelay()` is triggered too frequently, it can blow out the TX time on some of the packets to a ridiculous extent. I will put a cap on how many ms in the future that delay can be in order to prevent this. Am thinking that it should not be allowed to be delayed further than 2x CW.

src/mesh/RadioLibInterface.cpp
```diff
@@ -297,8 +298,8 @@ void RadioLibInterface::setTransmitDelay()
     if (p->tx_after) {
         unsigned long add_delay = p->rx_rssi ? getTxDelayMsecWeighted(p->rx_snr) : getTxDelayMsec();
         unsigned long now = millis();
-        p->tx_after = max(p->tx_after + add_delay, now + add_delay);
-        notifyLater(now - p->tx_after, TRANSMIT_DELAY_COMPLETED, false);
+        p->tx_after = min(max(p->tx_after + add_delay, now + add_delay), now + 2 * getTxDelayMsecWeightedWorst(p->rx_snr));
+        notifyLater(p->tx_after - now, TRANSMIT_DELAY_COMPLETED, false);
```
GUVWAF, line 302: Interesting, that `now - p->tx_after` is clearly wrong and could indeed have caused very long delays. 
It shouldn't get higher than `now + 2 * getTxDelayMsecWeightedWorst(p->rx_snr)`, since `tx_after` has been set to `millis() + getTxDelayMsecWeightedWorst()` before, but adding this check can't hurt.

Talie5in:
Adding latest log, might of mostly solved the issue.

[TXFillsUp-71ef9c9_info-LAST.txt](https://github.com/user-attachments/files/18389757/TXFillsUp-71ef9c9_info-LAST.txt)

src/mesh/RadioLibInterface.cpp
erayd: > It shouldn't get higher than `now + 2 * getTxDelayMsecWeightedWorst(p->rx_snr)`, since `tx_after` has been set to `millis() + getTxDelayMsecWeightedWorst()` before, but adding this check can't hurt.

It was blowing out, because packet RX storms were interrupting the delay and causing this line to keep adding more delay than the elapsed time since the last time the delay was increased, thus blowing out `tx_after` to ridiculous levels. This change caps the possible blowout to `now + 2 * getTxDelayMsecWeightedWorst(p->rx_snr)` in order to prevent it getting silly.

GUVWAF: Good point, makes sense.

GUVWAF (approved):

Talie5in:
Running for over 30 minutes on current testfix build that @erayd  and I have been working on, significantly improved, noisy/busy network aside.

PR #5832, opened 2025-01-12
Added illuminance sensors to the node's environmental sensor page

isseysandei:
Added illuminance sensors (lux and white_lux) to the environmental sensors page.
Added missing the white_lux to the debug output


fifieldt (approved):
Thanks!

PR #5844, opened 2025-01-13
Update RAK2560 code

caveman99:
* tx remote power telemetry
* remove uf2 script, pio run does that inline
* move sensor module to correct position
* disable LED and Accelerometer code


caveman99:
To be tested, WIP code.

caveman99:
Works for me (tm) :-)

thebentern (approved):

PR #5847, opened 2025-01-13
Environmental sensors page scrolling

isseysandei:
First iteration of the enviromental sensors creen scrolling.
When more than 3 sensors are present it shifts them up and down every 5 secs, so you can read them.
VID20250113194133-ezgif com-crop (1)

### DIFF
--- a/src/modules/Telemetry/EnvironmentTelemetry.cpp
+++ b/src/modules/Telemetry/EnvironmentTelemetry.cpp
@@ -222,7 +222,11 @@ void EnvironmentTelemetryModule::drawFrame(OLEDDisplay *display, OLEDDisplayUiSt
     }
 
     // Display "Env. From: ..." on its own
-    display->drawString(x, y, "Env. From: " + String(lastSender) + "(" + String(agoSecs) + "s)");
+    display->drawString(x, y, "Env. From: " + String(lastSender) + " (" + String(agoSecs) + "s)");
+
+    // Prepare sensor data strings
+    String sensorData[10];
+    int sensorCount = 0;
 
     if (lastMeasurement.variant.environment_metrics.has_temperature ||
         lastMeasurement.variant.environment_metrics.has_relative_humidity) {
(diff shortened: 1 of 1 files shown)


fifieldt (approved):

fifieldt:
Thanks for taking ownership of this :)

src/modules/Telemetry/EnvironmentTelemetry.cpp
```diff
@@ -232,48 +236,73 @@ void EnvironmentTelemetryModule::drawFrame(OLEDDisplay *display, OLEDDisplayUiSt
...
-                            "Water Level: " + String(lastMeasurement.variant.environment_metrics.distance, 0) + "mm");
+    static int scrollOffset = 0;
+    static bool scrollingDown = true;
+    static uint32_t lastScrollTime = millis();
 
-    if (lastMeasurement.variant.environment_metrics.weight != 0)
-        display->drawString(x, y += _fontHeight(FONT_SMALL),
-                            "Weight: " + String(lastMeasurement.variant.environment_metrics.weight, 0) + "kg");
+    // Draw up to 3 sensor data lines
+    int linesToShow = min(3, sensorCount);
```
todd-herbert, line 282: Pit of a posthumous review, sorry, but how would you feel if instead of fixing this at three, we calculated the number of lines based on display height?

fifieldt: Ah, that is a good point.

todd-herbert: I'll just quickly double check that change on a few devices and throw a PR together.

isseysandei: > Bit of a posthumous review, sorry about that.. How would you feel if instead of fixing this at three, we calculated the number of lines based on display height?

I didnt think about that, good idea!

PR #5848, opened 2025-01-13
Slight rework of CH341 HAL

psiegl:
I noticed, that the CH341 HAL could be a bit cleaned up. First, use the constructor / deconstructor of C++ to initiate/cleanup pinedio. Constructor throws error, in case not possible, which is then subsequently catched. Second, reworked the establishment of a MAC address for the CH341 HAL. The `getMacAddr` employs its parameter as an output. Thus, definining a similar variable that gets set upfront, it not useful for understanding the code.

### DIFF
--- a/src/platform/portduino/PortduinoGlue.cpp
+++ b/src/platform/portduino/PortduinoGlue.cpp
@@ -194,6 +194,7 @@ void portduinoSetup()
         }
     }
     // if we're using a usermode driver, we need to initialize it here, to get a serial number back for mac address
+    uint8_t dmac[6] = {0};
     if (settingsStrings[spidev] == "ch341") {
         try {
             ch341Hal =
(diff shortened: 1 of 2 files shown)


jp-bennett:
Looks good. I'll test it on my devices to confirm there's no surprises.

jp-bennett:
This is not calculating MAC Addresses the same way as the old code. It's a surprisingly tricky thing to get this just right. I've run into similar issues.

``libusb: warning [libusb_get_string_descriptor_ascii] suspicious bLength 18 for string descriptor (read 127)
Serial 10000002
Serial length 8
settingsStrings length 0
MAC Address from CH341: 12f4f0efb6
settingsStrings length a
MAC Address: 12f4f0efb6`` 
vs
``libusb: warning [libusb_get_string_descriptor_ascii] suspicious bLength 18 for string descriptor (read 127)
Serial 10000002
*** Blank MAC Address not allowed!``

Debugger shows that after the ch341 driver inits:
settingsStrings[mac_address] =
"9120F04F0EFB6"

psiegl:
Interesting, seems my CH341 doesn't have any serial that is why I didn't see any issue during rework:

```
Bus 001 Device 021: ID 1a86:5512 QinHeng Electronics CH341 in EPP/MEM/I2C mode, EPP/I2C adapter
Negotiated speed: Full Speed (12Mbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass          255 Vendor Specific Class
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         2 
  bMaxPacketSize0         8
  idVendor           0x1a86 QinHeng Electronics
  idProduct          0x5512 CH341 in EPP/MEM/I2C mode, EPP/I2C adapter
  bcdDevice            3.04
  iManufacturer           0 
  iProduct                2 USB UART-LPT
  iSerial                 0
```

psiegl:
@jp-bennett I reverted the part, which reads the serial. Can you retry?

jp-bennett (approved):

PR #5852, opened 2025-01-14
Small Fix: Release_Channels permissions

vidplace7:
Fixes error found with the Release_Channels workflow after Alpha 2.5.19 today

Also reworked the copr webhook after some local postman testing

### DIFF
--- a/.github/workflows/release_channels.yml
+++ b/.github/workflows/release_channels.yml
@@ -4,7 +4,9 @@ on:
   release:
     types: [published, released]
 
-permissions: read-all
+permissions:
+  contents: write
+  packages: write
 
 jobs:
   package-ppa:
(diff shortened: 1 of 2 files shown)


.github/workflows/release_channels.yml
```diff
@@ -4,7 +4,9 @@ on:
   release:
     types: [published, released]
 
-permissions: read-all
```
fifieldt, line 7: sure you don't still need some read permissions somewhere too?

vidplace7: The read permissions appear to be implicity?
My other parent release workflow (which is very similar) uses 
```yaml
permissions:
  contents: write
  packages: write
```
https://github.com/meshtastic/firmware/blob/master/.github/workflows/daily_packaging.yml#L18-L20

fifieldt (approved):

Issue #5854, opened 2025-01-14
[Bug]: proto.db doubling on NRF52


todd-herbert:
### Category

Other

### Hardware

Rak4631

### Firmware Version

2.5.19.d5cd6f8

### Description

In at least some situations, the size of proto.db seems to double when written to flash.

#### Affected:
* RAK4631
* Heltec MeshNode T114

#### Not affected:
* Heltec VM-E213

### Steps to reproduce:
* Erase with `nrf_erase2.uf2` via web flasher
* Flash 2.5.19 alpha from web flasher
* Pair BLE with Android app 2.5.16

*Then, repeat:*
* Wait for system to boot fully (~80 seconds, for GPS scan)
* Modify longname via Android app
* Wait for device to reboot

Full log (multiple boots): [t114_dbproto_growing.txt](https://github.com/user-attachments/files/18410080/t114_dbproto_growing.txt)

Note: nodedb contained no other mesh nodes during these tests

### Relevant log output

```Shell
ERROR | 11:59:01 217 [Router] Error: can't encode protobuf io error
ERROR | 11:59:01 217 [Router] Can't write prefs!
ERROR | 11:59:01 217 [Router] Failed to save to disk, retrying
DEBUG | 11:59:03 219 [Router] Opening /prefs/db.proto, fullAtomic=0
INFO  | 11:59:03 219 [Router] Save /prefs/db.proto
ERROR | 11:59:05 221 [Router] NOTE! Record critical error 12 at src/mesh/NodeDB.cpp:1220

...

DEBUG | ??:??:?? 1 Filesystem files:
DEBUG | ??:??:?? 1  prefs (directory)
DEBUG | ??:??:?? 1    config.proto (162 Bytes)
DEBUG | ??:??:?? 1    channels.proto (57 Bytes)
DEBUG | ??:??:?? 1    db.proto (13721 Bytes)

...

INFO  | ??:??:?? 1 Loaded saved devicestate version 23, with nodecount: 160
WARN  | ??:??:?? 1 Node count 160 exceeds MAX_NUM_NODES 80, truncating
```

dahanc:
> Does FILE_O_WRITE always cause it to append?!

Wow, that definitely violates the principle of least astonishment. But yeah, looks like that's intentional: https://github.com/adafruit/Adafruit_nRF52_Arduino/issues/234

PR #5856, opened 2025-01-14
Prevent NRF52 files appending

todd-herbert:
Possible fix for #5854


todd-herbert:
Possible issues with this one after clearing nodedb. Proceed with caution.

esev:
Just curious, what were the issues? I've been looking at the code in https://github.com/geeksville/Adafruit_nRF52_Arduino/blob/master/libraries/Bluefruit52Lib/src/utility/bonding.cpp and this seems to be one way they handle it.

esev:
Never mind. If the data size is smaller this wouldn't truncate the file. I see the issue.

esev:
If you first seek, and then truncate, does that solve the issue?

Issue #5839, opened 2025-01-13
[Bug]: Critical fault #12 flash filesystem corruption and format on nrf52 platform


dahanc:
### Category

Other

### Hardware

Heltec Mesh Node T114

### Firmware Version

2.5.13.1a06f88

### Description

Back in November 2024, my T114 showed a Critical fault #&NoBreak;12. I rebooted it and it seemed to work OK, but a few days later, it got into a boot loop. The serial debug output was:
```
INFO  | ??:??:?? 6 Adding node to database with 1 nodes and 147308 bytes free!
DEBUG | ??:??:?? 6 Expanding short PSK #1
INFO  | ??:??:?? 6 Wanted region 1, using US
INFO  | ??:??:?? 6 Saving /prefs/db.proto
lfs warn:314: No more free space 224
ERROR | ??:??:?? 6 Error: can't encode protobuf io error
ERROR | ??:??:?? 6 LFS assert: head >= 2 && head <= lfs->cfg->block_count
ERROR | ??:??:?? 6 LFS assert: head >= 2 && head <= lfs->cfg->block_count
ERROR | ??:??:?? 6 LFS assert: block < lfs->cfg
```
Then about a week later, another of my T114s also got into a boot loop. This one doesn't have a screen, so I don't know if it also had a Critical fault #&NoBreak;12. I erased the flash, installed 2.5.13.1a06f88, connected the USB port to a PC, and logged all of the serial output, so in case it happened again, I could see what happened right before the first crash and reboot. After about a month, it rebooted:

```
INFO  | 20:29:31 2159141 [Router] Received nodeinfo from=0xa20a11c8, id=0x7d3b35e6, portnum=4, payloadlen=44
INFO  | 20:29:31 2159141 [Router] Node database full with 80 nodes and 109096 bytes free. Erasing oldest entry
INFO  | 20:29:31 2159141 [Router] Adding node to database with 80 nodes and 109096 bytes free!
DEBUG | 20:29:31 2159141 [Router] old user /, channel=0
DEBUG | 20:29:31 2159141 [Router] Update changed=1 user MKTX G2 Gateway/MKG3, channel=0
DEBUG | 20:29:31 2159141 [Router] State: ON
DEBUG | 20:29:31 2159141 [Router] Node status update: 22 online, 80 total
INFO  | 20:29:31 2159141 [Router] Save /prefs/db.proto
lfs debug:617: Bad block at 126
lfs debug:617: Bad block at 180
[...]

PR #5859, opened 2025-01-15
Partial workaround for BLE-related corruption on shutdown / reboot

todd-herbert:
This helps prevent one specific case, where BLE disconnection during shutdown / reboot can cause NRF52 flash corruption.

*This does not address the corruption which can occur when a phone moves out of range of a node.* See https://github.com/meshtastic/firmware/issues/5839#issuecomment-2592411905

Because BLE seems to vary from phone to phone, it would be good to have this one smoke-tested by a couple of people.


thebentern (approved):

thebentern:
Since you have tested this on Android, I will test this on iOS.

PR #5864, opened 2025-01-16
COPR: Switch from hook to `copr_cli`

vidplace7:
The COPR webhook refuses to work correctly... despite returning `200`...

This PR switches to a copr-cli based solution.

Do not merge until `secrets.COPR_API_CONFIG` exists.


vidplace7:
Secret should be present now. Ready for merge 👍

PR #5881, opened 2025-01-18
No focus on new messages if auto-carousel is off

isseysandei:
Often, I need to monitor a specific screen on my node, so I disable the auto-screen carousel. However, every time a new message arrives, it gets focused, forcing me to cycle back manually. To avoid this, I implemented a check: if the carousel is disabled, it no longer focuses on new messages.


src/graphics/Screen.cpp
```diff
@@ -2661,15 +2661,18 @@ int Screen::handleStatusUpdate(const meshtastic::Status *arg)
...
+// If auto carousel is disabled, the focus doesn't chage on the incoming message
 {
-    if (showingNormalScreen) {
-        // Outgoing message
-        if (packet->from == 0)
-            setFrames(FOCUS_PRESERVE); // Return to same frame (quietly hiding the rx text message frame)
+    if (config.display.auto_screen_carousel_secs != 0) {
+        if (showingNormalScreen) {
+            // Outgoing message
+            if (packet->from == 0)
```
thebentern, line 2669: Can we make this a ternary or something to clean up the nesting? 3 if statements down feels too nesty

isseysandei:
I still need to build and test it on the node, but something like this should work fine, hopefully.

thebentern (approved):

isseysandei:
I suspect that the raspbian checks are broken... Queued for almost 5 hours...

HarukiToreda:
This messes up all standalone nodes that rely on the message screen to know there was a message received like the Tdeck. Before when a message was received it would turn the screen on and focus on the message screen, now I have to manually cycle to see and guess if I receive a message. This need to be rethought

PR #5908, opened 2025-01-22
Update variant.h

mechbox:
Xiao board: Adding UART GPS (only RX) through Xiao PIN6


caveman99:
> Xiao board: Adding UART GPS (only RX) through Xiao PIN6

This board does not have a fixed GPS chip. Hogging a line for GPS is unfriendly to people that may use this GPIO for something else. you can define the GPS Pins through the settings for the board.

PR #5916, opened 2025-01-23
NRF52 - Remove file totally before opening write

thebentern:
Given the discoveries in #5839, I believe we should abandon the seek(0) / truncate approach to opening files to write, as this  has issues in the current NRF52 framework version of LittleFS.
The more crude approach of removing the file from the filesystem prior to opening for write seems to be the only path forward, unless I am overlooking an alternative approach.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -1105,6 +1105,7 @@ bool NodeDB::saveProto(const char *filename, size_t protoSize, const pb_msgdesc_
     bool okay = false;
     auto f = SafeFile(filename, fullAtomic);
 
+#ifdef FSCom
     LOG_INFO("Save %s", filename);
     pb_ostream_t stream = {&writecb, static_cast<Print *>(&f), protoSize};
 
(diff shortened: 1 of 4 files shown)


esev (approved):

esev:
I've reached the same conclusion. Thanks for this.

@todd-herbert let's try to get this one merged too https://github.com/meshtastic/Adafruit_nRF52_Arduino/pull/1. I think this PR and yours should resolve the issues folks are seeing.

Note: I'm away from my test node till Monday, so I can't test. But peeking in on the logs shows it is still running after removing 85de193 & 1c0f43c.

src/mesh/NodeDB.cpp
todd-herbert, line 1: I

todd-herbert (changes requested):
Definitely agree! Just one line that might want attention first though.

todd-herbert:
> let's try to get this one merged too https://github.com/meshtastic/Adafruit_nRF52_Arduino/pull/1. I think this PR and yours should resolve the issues folks are seeing.

I'll bring that one forward to match the upstream PR on the Adafruit repo and then yeah lets go ahead with it.

todd-herbert:
Give me just a few minutes for a quick test run on real hardware 🙏

todd-herbert (approved):
Seems to all be running well here! (Tested on T114)
* .proto files not doubling in size
* BLE disconnect still safe with https://github.com/meshtastic/Adafruit_nRF52_Arduino/pull/1
* recovered from corruption induced by power removal during write

fifieldt (approved):
makes sense.

PR #5931, opened 2025-01-25
Space out periodic broadcasts of modules automatically

GUVWAF:
Some modules have by default the same broadcast interval (or multiples of each other) and are also started at the same time, resulting in them being sent quickly after each other, thereby making delivery harder considering the packets are being propagated through the mesh also.
I’ve seen a workaround of @b8b8 to set the intervals to, e.g., 3600, 3615, 3630 seconds, etc. With this PR, we space periodic broadcasts of modules by 15 seconds by letting them start 15 seconds after each other (which was manually done for NodeInfo, DeviceTelemetry and Position already).

### DIFF
--- a/src/mesh/MeshModule.cpp
+++ b/src/mesh/MeshModule.cpp
@@ -10,6 +10,7 @@
 std::vector<MeshModule *> *MeshModule::modules;
 
 const meshtastic_MeshPacket *MeshModule::currentRequest;
+uint8_t MeshModule::numPeriodicModules = 0;
 
 /**
  * If any of the current chain of modules has already sent a reply, it will be here.  This is useful to allow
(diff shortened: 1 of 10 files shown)


thebentern:
Really great simple approach. This should effectively spread out a lot of airtime by staggering the intervalled broadcasts. 

I was initially concerned with the usage in runOnce() for some of the modules, but looking at the context of it, you're using it inside of the if (firstTime) block, which should prevent inadvertently "stacking" the incremented intervals. I wonder if we should place a warning comment about this to prevent that issue from future consumers? 😅 
Another related thought I'd like to pick your brain about as a long term idea is potentially rescheduling some of the broadcast intervals based on a trend of observed periods of dead space in the ch. utilization.

thebentern (approved):

GUVWAF:
>  I wonder if we should place a warning comment about this to prevent that issue from future consumers?

Good point, added a warning for the function description.

> Another related thought I'd like to pick your brain about as a long term idea is potentially rescheduling some of the broadcast intervals based on a trend of observed periods of dead space in the ch. utilization.

That's an interesting thought. I'm wondering if the dead spaces are common in larger meshes (where periodic broadcasts start becoming a problem), or it's already spread out quite well.
To take this into account we'd need to start storing more channel utilization history on-device though.

b8b8:
We actually switched to primes greater than 10 seconds apart.
image

PR #5938, opened 2025-01-25
No focus on new messages if carousel off [revised]

isseysandei:
Trying to find a better implementation than #5936 , in order to avoid problems to standalone devices.
In this first attemp I just changed the 0 inside the if condition to a specific delay.
Any suggestion to improve this janky modification is open, tks!


isseysandei:
nope, by doing this you just disable the focus, but the carousel is still on....

PR #5954, opened 2025-01-28
Portduino: Allow limiting TX Power from yaml

### LINKED ISSUE #5857: [Feature Request]: add a setting in config.yaml for power limits
### Platform

Linux Native

### Description

An additional set of power limits in the config.yaml files would be along the lines of "I have a 33db radio, limit tx to 8" or "I have a 30db radio, my maximum eirp is 27, limit tx to 17" or something.

These would be separate config files to the radio setup, so the original radio configs would be universal and international.

vidplace7:
Limit output power via meshtasticd YAML config, similar to how it's handled in the `variant` files for typical nodes.
Especially useful for `E22-900M33S` (2W) modules that like to explode at 22db input power.

Resolves #5857


vidplace7:
Works locally. Could probably be handled in a cleaner way, but the `lr1110 / lr1120` throws a wrench in things.
Open to suggestions!

caveman99 (approved):

PR #5973, opened 2025-02-01
Sync with TFT branch - variant files

fifieldt:
This brings the platformio and variant.h files for variants supporting MUI up to spec with TFT branch.


mverch67:
looks good to me

thebentern:
Closing, since we integrated downstream

PR #5978, opened 2025-02-02
Remove unused usages of #include <iostream> to save Flash

Stary2001:
Saves ~100KB on wio-e5, which was previously at 99% Flash usage.
Probably saves some Flash on other platforms, too.


GUVWAF (approved):
Very nice find, thanks!

fifieldt (approved):

mrpatrick1991:
thank you so much this is amazing

PR #6001, opened 2025-02-06
Change nonce logging to DEBUG instead of Info

lizTheDeveloper:
This logging statement is leaking the nonce to stdout, if your logs are routed somewhere, this logs the nonce every time, leading to replay attack surface area being higher. Changed to debug level.


fifieldt (approved):
Thank you!

PR #6020, opened 2025-02-09
meshtasticd 2.6: Add X11 480x480 preset

vidplace7:
Add GUI preset for X11 with 480x480 viewport for `meshtasticd` 2.6.

This will be consumed by the FlatPak preview builds of meshtasticd (added to config.d by default)


mverch67 (commented):
Does it conflict with this one?
https://github.com/meshtastic/firmware/blob/master/bin/config.d/display-x11.yaml

markbirss:
> Does it conflict with this one? https://github.com/meshtastic/firmware/blob/master/bin/config.d/display-x11.yaml

Same thing just different folder, he created it first, i will revert

markbirss:
> This will be consumed by the FlatPak preview builds of meshtasticd

i tested it with 2.6 also (pio run -e native-tft)

PR #6025, opened 2025-02-10
meshtasticd flatpak: Include pio deps with release

vidplace7:
Package platformio dependencies as a zip file included with the Release.

This will be consumed by the Meshtastic FlatHub flatpak, which cannot have internet access at build-time.


vidplace7:
Left `native-tft` targets commented out for now, to be un-commented when `2.6` is merged to master.

thebentern (approved):

PR #6031, opened 2025-02-11
Added custom OCV array values for T1000-E to improve percentage estimates

nwilde1590:
Added OCV_ARRAY values for the T1000-E since the battery used by Seeed has a pretty sharp drop-off when it starts to die. This typically causes the device to drop from ~50% to 0% far faster than it should. The updated values are based on the discharge rate of my own T1000-E and confirmed to approximately match the discharge curve published by Seeed for this specific device. The new array leads to much more reasonable percentage values that decrease roughly linearly as the device (in my own environment at least) is under a fairly constant load. 

Before modification:
IMG_3513

After modification:
IMG_3514

Currently the arrays are divided by battery type. For turnkey devices that come with a battery, it would be nice to see them added as needed based on their respective performance.


fifieldt (approved):
Thank you very much for a highly welcome first contribution.

PR #6035, opened 2025-02-11
Seeed Indicator V2: set TCXO to 2.4V

mverch67:
request from Seeed developer


mverch67:
uuuuuppps... something got wrong here

Issue #6041, opened 2025-02-12
[Bug]: The SPI1 defined for RAK4631 but non-existent


Woutvstk:
### Category

Hardware Compatibility

### Hardware

Rak4631

### Firmware Version

/

### Description

In the variant.h file of the rak4631 board there seem to be wrong SPI definitions:

```
// rak4631/variant.h

#define SPI_INTERFACES_COUNT 2

#define PIN_SPI_MISO (45)
#define PIN_SPI_MOSI (44)
#define PIN_SPI_SCK (43)

#define PIN_SPI1_MISO (29) // (0 + 29)
#define PIN_SPI1_MOSI (30) // (0 + 30)
#define PIN_SPI1_SCK (3)   // (0 + 3)
```
In the documentation, I can't find any mention of a second SPI interface. (There is a QSPI but that is handled separately in the variant file)
I also can't find any mention of SPI pins 43, 44, 45 in the docs.
The pins 3, 29 and 30 are the only real SPI pins

Am I overlooking something?

### Relevant log output

```Shell

```

GUVWAF:
Fixed by #6042.

PR #6042, opened 2025-02-12
Rak4631 remove spi1

Woutvstk:
**Hardware:**

RAK4631

**Issue:**

In the variant file of rak4631 there seem to be wrong SPI definitions
See issue #6041

### DIFF
--- a/src/detect/einkScan.h
+++ b/src/detect/einkScan.h
@@ -6,28 +6,28 @@
 
 void d_writeCommand(uint8_t c)
 {
-    SPI1.beginTransaction(SPISettings(4000000, MSBFIRST, SPI_MODE0));
+    SPI.beginTransaction(SPISettings(4000000, MSBFIRST, SPI_MODE0));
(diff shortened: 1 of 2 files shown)


GUVWAF (approved):
Seems to correspond to the datasheet and the reference Arduino core from RAK Wireless: https://github.com/RAKWireless/RAK-nRF52-Arduino/blob/master/variants/WisCore_RAK4631_Board/variant.h

GUVWAF:
@Woutvstk This broke the detection of the LoRa radio on RAK4631. It seems to use an internal SPI, which do use the pins as described in the variant: 

https://github.com/meshtastic/firmware/blob/fe1ced748046f9a2f0c4e544177ac15bd51f609f/variants/rak4631/variant.h#L176-L180

I think we should either revert this, but #6080 seems to depend on it. 
Another solution is to define `LORA_MOSI`, etc. specifically, but e.g. the epaper variant of the RAK4631 still has two SPI interfaces defined.

PR #6046, opened 2025-02-13
MUI: BT programming mode

mverch67:
allow BT connection with disabled MUI:
- NodeDB: had to shift the force overwrite of BT config to the init with defaults section
- PacketAPI: notify client about BT enabled and stop processing toPhone queue until next reboot


mverch67:
uuuhhh what's that again???!
mama mia

thebentern (approved):

PR #6047, opened 2025-02-13
Licensed usage compliance

thebentern:
This closes all of the loopholes of amateur radio licensed operation allowing both the legacy admin channel and any pre-shared encryption keys on channels.
The current behavior allows users to "get around" the `setHamMode` removal of the default encryption key and be in potential violation of licensed amateur radio usage.
Untested so far (Will test soon)

### DIFF
--- a/src/mesh/Channels.cpp
+++ b/src/mesh/Channels.cpp
@@ -93,6 +93,35 @@ void Channels::initDefaultLoraConfig()
 #endif
 }
 
+bool Channels::ensureLicensedOperation()
+{
+    if (!owner.is_licensed) {
+        return false;
+    }
+    bool hasEncryptionOrAdmin = false;
+    for (uint8_t i = 0; i < MAX_NUM_CHANNELS; i++) {
+        auto channel = channels.getByIndex(i);
+        if (!channel.has_settings) {
+            continue;
+        }
+        auto &channelSettings = channel.settings;
+        if (strcasecmp(channelSettings.name, Channels::adminChannel) == 0) {
+            channel.role = meshtastic_Channel_Role_DISABLED;
+            channelSettings.psk.bytes[0] = 0;
+            channelSettings.psk.size = 0;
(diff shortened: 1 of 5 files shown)


garthvh (commented):
Looks like it still needs to prevent empty keys when licensed mode is off

GUVWAF:
What about letting it check whether the source node has `user.is_licensed` set (which is stored in `UserLite`) in order to determine whether it can rebroadcast?

thebentern:
> What about letting it check whether the source node has `user.is_licensed` set (which is stored in `UserLite`) in order to determine whether it can rebroadcast?

I don't hate that idea. I think it makes sense from the ham side, since you have required 10 minute broadcasts of NodeInfo on licensed mode. There's a slim window of potentially missing some rebroadcasts. As is, we would also prevent rebroadcasting of encrypted content with this PR.

For normal ISM users, we could just rebroadcast everything by default until we see a User that claims is_licensed and then prevent processing those packets.

GUVWAF:
> For normal ISM users, we could just rebroadcast everything by default until we see a User that claims is_licensed and then prevent processing those packets?

Yes, I think that's a reasonable approach.

thebentern:
Sounds good. I'll make a follow-up PR

argilo:
My local amateur radio club uses Meshtastic on the 70cm band (EU_433) and we had been using the default LongFast channel. (As far as I understand, this means that we were using the default PSK, which is OK in Canada since our amateur radio regulations allow encryption as long as the key is published in the public domain.)

With the 2.6 firmware, the PSK is always removed when using the "Licensed amateur radio" mode, which unfortunately means that nodes running 2.6 cannot communicate with nodes running 2.5 (until we reconfigure those nodes to remove the PSK).

Running without encryption probably makes more sense for amateur radio use, but I wanted to note this issue because I expect that other radio amateurs may run into trouble when updating to 2.6.

PR #6054, opened 2025-02-14
Increased message font to medium and enabled trackball for selecting node/channel selection.

tomhanax:
This change covers two things:

1. Increased message font to medium on larger displays, namely T-Deck and Chatter 2.0 (based on discussions on https://github.com/meshtastic/firmware/issues/2896). The font will be medium for incoming message and also while composing new message using `CannedMessageModule`. There may be some theoretical truncation on very long messages, but in practical life this is not the issue.
2. On T-Deck, trackball can be used for selecting node/channel selection. You can use UP and DOWN direction and switching between message, node and channel. LEFT / RIGHT cycles values. The changes were minor but the effect is IMHO pretty usable, even more now when we have real private messages (using public/private keys)

----
PS: This is my first PR, so sorry if something is not in order, I like to learn in time.


caveman99 (approved):

mverch67:
In general I would try to avoid to splatter the code with #if DEVICE_XY as this creates a maintenance hell. If a new device is added with same feature then dozens of places in dozens of files need to be changed which could potentially break that existing code for other devices.
Instead I suggest to use same macro mechanism but add a generic feature define, e.g. USE_MSG_FONT_MEDIUM which can be universally used by any device, i.e. if a new device is added just define USE_MSG_FONT_MEDIUM in it's variant and good is. No changes in dozens of files required.

tomhanax:
> Instead I suggest to use same macro mechanism but add a generic feature define, e.g. USE_MSG_FONT_MEDIUM
I understand what you're trying to say, and since I'm new here, I'll ask: Am I expected to redo this pull request according to this suggestion, or is it just good advice for the future?

mverch67:
> > Instead I suggest to use same macro mechanism but add a generic feature define, e.g. USE_MSG_FONT_MEDIUM
> 
> I understand what you're trying to say, and since I'm new here, I'll ask: Am I expected to redo this pull request according to this suggestion, or is it just good advice for the future?

I think there's still a change to change it :)
Just do the change, commit and push.

It would even be better and more elegant to query the actual screen height and based on that height use medium or small font. Then we could totally get rid of the define.

tomhanax:
@mverch67, I found the constant VIEW_320x240 defined in several configuration files, but I can't find any usage for it. I'm not sure if it's a good approach, and whether this constant is a good idea or what it is used for. But I'm thinking I could use it instead of defining another constant, but that would mean that all such devices would need larger text.

Do you think it's a good idea? Or would it be better to actually ask about the height of the display and decide based on that?

mverch67:
> Do you think it's a good idea? Or would it be better to actually ask about the height of the display and decide based on that?

It's not a good idea to use compiler macros. It also won't work with resizable UIs. As I said before display width and height must be taken into account. And when using lvgl this should be done via styles.

tomhanax:
Since we already have MUI, how should we complete this pull request? Will the original user interface still be actively developed, or is there a plan for some kind of black-and-white lightweight MUI version? I must admit that I am beginner in this C++ world and I don't have enough time to redo it for dynamic display height detection; I currently have no idea what all that entails. At least I have updated it to a new constant. 

Výber_064

Basically, I plan to define it for Chatter 2 and T-Deck. Is it supposed to be like this?

Výber_065

Is my PR acceptable this way?

mverch67:
> Is my PR acceptable this way?

that looks better, please commit and push these changes.
What devices did you use for tests? Are devices with 128x64 OLED still working as expected?

tomhanax:
> What devices did you use for tests? Are devices with 128x64 OLED still working as expected?
I tested T-Deck, Chatter 2, Heltec V3, all OK.

variants/t-deck/platformio.ini
```diff
@@ -67,6 +67,7 @@ build_flags =
 ;	-D USE_DOUBLE_BUFFER
   -D USE_PACKET_API
   -D MAP_FULL_REDRAW
+  -D USE_MSG_FONT_MEDIUM
```
mverch67, line 70: This is the wrong place. It should go into env:t-deck above. Didn't you say you have tested it?

mverch67 (changes requested):

variants/t-deck/platformio.ini
tomhanax: I've tested original approach with `defined(T_DECK) || defined(CHATTER_2)`, not the current commit. My fault.

mverch67 (approved):

PR #6064, opened 2025-02-15
Increase `MAX_THREADS` for InkHUD variants with WiFi

GUVWAF:
I got into a reboot loop with the InkHUD and WiFi enabled, since then we’re above `MAX_THREADS`. I increased it to 40 for InkHUD variants that have WiFi, just like is done for the T-Deck already.

@todd-herbert very nice work by the way!


thebentern (approved):

todd-herbert:
Oh thank you! I was worried about hitting that limit, so it's reassuring to hear that it was already bumped for the T-Deck.

PR #6065, opened 2025-02-16
assigning SDA/SCL so it actually works 8|

porkcube:
Turns out... [setting SDA/SCL to pin -1 makes things crashy crashy](https://discord.com/channels/867578229534359593/938197090935398451/1340338466277294241) on the DIY Xiao nRF52840 + Wio SX1262:
```
INFO  | ??:??:?? 1 

//\ E S H T /\ S T / C

DEBUG | ??:??:?? 1 Filesystem files:
DEBUG | ??:??:?? 1 Power::lipoInit lipo sensor is not ready yet
DEBUG | ??:??:?? 1 Use analog input 32 for battery level
```
(node freezes)

Removing SERIAL1_RX/TX and using Pins 6/7 confirmed working at the cost of Serial Port -  I will now go back to hiding in the corner of shame for an appropriate amount of time.

### DIFF
--- a/variants/diy/seeed-xiao-nrf52840-wio-sx1262/variant.h
+++ b/variants/diy/seeed-xiao-nrf52840-wio-sx1262/variant.h
@@ -84,17 +84,15 @@ static const uint8_t A5 = PIN_A5;
 #define PIN_NFC2 (31)
 
 // RX and TX pins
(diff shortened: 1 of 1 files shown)


adriannow:
Also tested it with updated configuration and can confirm - now works perfectly.

PR #6068, opened 2025-02-16
Prevent licensed users from rebroadcasting unlicensed traffic

thebentern:
As is (on 2.6), licensed users have rebroadcast_mode set to LOCAL_ONLY and this requires that all channels have no encryption keys, so we are effectively preventing encrypted traffic from traversing the mesh. We currently rebroadcast any plaintext traffic in good faith.
This PR takes the distinction a step further by asserting that if we have an unecrypted packet and we know that the user does not claim to be licensed, we should not rebroadcast that packet.

### DIFF
--- a/src/modules/RoutingModule.cpp
+++ b/src/modules/RoutingModule.cpp
@@ -20,6 +20,12 @@ bool RoutingModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mesh
         if ((nodeDB->getMeshNode(mp.from) == NULL || !nodeDB->getMeshNode(mp.from)->has_user) &&
             (nodeDB->getMeshNode(mp.to) == NULL || !nodeDB->getMeshNode(mp.to)->has_user))
             return false;
+    } else if (owner.is_licensed && nodeDB->getMeshNode(mp.from) != NULL && nodeDB->getMeshNode(mp.from)->has_user &&
+               !nodeDB->getMeshNode(mp.from)->user.is_licensed) {
+        // Don't let licensed users to rebroadcast packets from unlicensed users
+        // If we know they are in-fact unlicensed
+        LOG_DEBUG("Packet from unlicensed user, ignoring packet");
+        return false;
     }
 
     printPacket("Routing sniffing", &mp);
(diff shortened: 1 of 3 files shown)


src/modules/RoutingModule.cpp
```diff
@@ -20,6 +20,12 @@ bool RoutingModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mesh
         if ((nodeDB->getMeshNode(mp.from) == NULL || !nodeDB->getMeshNode(mp.from)->has_user) &&
             (nodeDB->getMeshNode(mp.to) == NULL || !nodeDB->getMeshNode(mp.to)->has_user))
             return false;
+    } else if (owner.is_licensed && nodeDB->getMeshNode(mp.from) != NULL && nodeDB->getMeshNode(mp.from)->has_user &&
+               !nodeDB->getMeshNode(mp.from)->user.is_licensed) {
+        // Don't let licensed users to rebroadcast packets from unlicensed users
+        // If we know they are in-fact unlicensed
+        LOG_DEBUG("Packet from unlicensed user, ignoring packet");
+        return false;
```
GUVWAF, line 28: I notice now that these `if`-statements also cause it to not be sent to the client (line 32), while we only want to prevent rebroadcasting.

src/modules/RoutingModule.cpp
```diff
@@ -20,6 +20,12 @@ bool RoutingModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mesh
         if ((nodeDB->getMeshNode(mp.from) == NULL || !nodeDB->getMeshNode(mp.from)->has_user) &&
             (nodeDB->getMeshNode(mp.to) == NULL || !nodeDB->getMeshNode(mp.to)->has_user))
             return false;
+    } else if (owner.is_licensed && nodeDB->getMeshNode(mp.from) != NULL && nodeDB->getMeshNode(mp.from)->has_user &&
```
GUVWAF, line 23: Minor thing, but it seems rather inefficient to call `getMeshNode()` three times for the same node, especially with large NodeDBs. Probably better to define a `fromNode`. Same optimization can be done in `perhapsDecode()`.

thebentern: Good point. I'll make a specific method to check I think.

src/modules/RoutingModule.cpp
thebentern: I believe that's actually for the best, as the licensed operator would potentially receive traffic that they would / should not respond to.

GUVWAF (approved):

Issue #5880, opened 2025-01-18
Battery Voltage for Seeed Xiao ESP32S3 Monitor


nis0521:
### Request Details

Problem: Seeed Xiao Meshtastic boards don’t have a native way of measuring battery voltage when powered by batteries. 

Solution: Because the the ESP32S3 super small boards have been out for a while, there is a workaround where you use 2 resisters from the battery terminal to a a2d pin on the board and uses a few line of code. It’s actually documented on the official product page 

https://wiki.seeedstudio.com/XIAO_ESP32C3_Getting_Started/

Issue: In order to make this work those lines of code need to make it into the firmware. I don’t know how I would go about doing this or this can be worked by the devs. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

caveman99:
The modification you quote is for the C3 board, not the S3 board. While there is a similar mod for the S3 board we won't add that to the stock firmware, as all of the available GPIO pins re tied up already. For measurements i recommend adding an INA2xx probe.

nis0521:
Okay thanks for looking into this. 

I think the main attraction for this board is the size. Adding to the board kind of defeats the purpose of the board. It’d be great if this reconsidered maybe when you guys have more bandwidth:(

Thinking about making a few remote gadgets. Not having battery monitoring is a big issue :(

nis0521:
@caveman99 so went and got some INA219 boards and an android to get this to work. Doesn't seem to be working Voltage and  Current comes in on Channel 3 and not on the Battery which is just at 0%. Channel seems to be correct at 0x40 (64). 

Any way you can reopen this as an active issue? Another dude seems like he has a workaround: 

https://github.com/meshtastic/firmware/issues/5248

Issue #5248, opened 2024-11-04
[Bug]: RP2040-Lora ignoring power INA configuration


SavageU:
### Category

Other

### Hardware

Other

### Firmware Version

2.5.7.f77c87d

### Description

RP2040-LORA-HF device is ignoring request to set INA219 as main battery device.
Continuously reports as PWD status at 0.0V despite being configured to read INA.

Occurring on 2.5.7 and 2.5.8 (downgraded back to 2.5.7 so I could increase cadence of power metrics transmit while testing.)

INA is being correctly recognised at 0x40 address on boot:

```
INFO  | ??:??:?? 3 Scanning for i2c devices...
DEBUG | ??:??:?? 3 Scanning for I2C devices on port 1
DEBUG | ??:??:?? 3 I2C device found at address 0x40
DEBUG | ??:??:?? 3 Wire.available() = 2
DEBUG | ??:??:?? 3 Register MFG_UID: 0x2000
INFO  | ??:??:?? 3 INA219 sensor found at address 0x40
DEBUG | ??:??:?? 3 I2C device found at address 0x76
DEBUG | ??:??:?? 3 Wire.available() = 1
DEBUG | ??:??:?? 3 Wire.available() = 1
INFO  | ??:??:?? 3 BMP-280 sensor found at address 0x76
INFO  | ??:??:?? 3 2 I2C devices found
DEBUG | ??:??:?? 3 acc_info = 0
DEBUG | ??:??:?? 3 found i2c sensor meshtastic_TelemetrySensorType_BMP280
DEBUG | ??:??:?? 3 found i2c sensor meshtastic_TelemetrySensorType_INA219
INFO  | ??:??:?? 3 S:B:30,2.5.7.f77c87d
INFO  | ??:??:?? 3 Build timestamp: 1729118132

```
However, the telemetry/update module is still reporting:

```
INFO  | ??:??:?? 48 [DeviceTelemetryModule] (Sending): air_util_tx=0.077111, channel_utilization=15.256667, battery_level=101, voltage=0.000000, uptime=48
```

When a remote request is placed from another node the same results occur:

```
meshtastic -s com11 --dest [redacted] --request-telemetry
Connected to radio
Sending device_metrics telemetry request to [redacted] on channelIndex:0 (this could take a while)
Telemetry received:
Battery level: 101.00%
Voltage: 0.00 V
Total channel utilization: 10.90%
Transmit air utilization: 0.06%
Uptime: 5 s
```

[...]

Issue #3409, opened 2024-03-14
[Feature Request]: Autodetect ina sensors


garthvh:
### Platform

NRF52, ESP32

### Description

It would be good if we could get the power sensors in line with the other I2C sensors and auto detect them instead of requiring users to enter a I2C address.

fifieldt:
Looks like the detection code is in src/detect/ScanI2CTwoWire.cpp for INA260, INA219, and INA3221

So to solve this it looks like src/Power.cpp should be modified to use the detected values rather than requiring config.power.device_battery_ina_address

caveman99:
this has been implemented along the way

PR #6070, opened 2025-02-16
Expose INA219 measurement as battery voltage for Seeed Xiao ESP32S3

syssi:
The Meshtastic firmware lacks the ability to publish INA power monitor measurements as battery voltage readings when `BATTERY_PIN` is not defined. This limitation affects users who want to monitor battery levels using INA sensors.

Additionally, when a valid GPIO/A0 pin is assigned as `BATTERY_PIN`, users can measure analog voltage directly through that pin without requiring firmware recompilation. This feature provides more flexibility for hardware configurations.

This improvement would benefit users by:
1. Enabling battery voltage monitoring through INA sensors even without a dedicated battery pin
2. Allowing dynamic analog voltage measurements through GPIO pins
3. Reducing the need for custom firmware builds

Related issues:

https://github.com/meshtastic/firmware/issues/5248
https://github.com/meshtastic/firmware/issues/5880

### DIFF
--- a/variants/seeed_xiao_s3/variant.h
+++ b/variants/seeed_xiao_s3/variant.h
@@ -36,7 +36,7 @@ L76K GPS Module Information : https://www.seeedstudio.com/L76K-GNSS-Module-for-S
 #define BUTTON_PIN 21 // This is the Program Button
 #define BUTTON_NEED_PULLUP
 
-#define BATTERY_PIN 1
+#define BATTERY_PIN -1
 #define ADC_CHANNEL ADC1_GPIO1_CHANNEL
 #define BATTERY_SENSE_RESOLUTION_BITS 12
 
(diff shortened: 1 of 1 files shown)


fifieldt:
What we need is this: https://github.com/meshtastic/firmware/issues/3409

syssi:
I agree auto-detection would be nice. IMO this is out of scope of this pull request.

caveman99:
> @caveman99 Some variants.h are using `BATTERY_PIN -1`. Is this an option?

Yes that would be an option, but every toolchain behaves quite differently. Did you actually test the mod?

syssi:
Yes. The `seeed-xiao-s3` build works fine with `BATTERY_PIN -1`: 

```
INFO  | ??:??:?? 2 [Power] Init sensor: INA219
INFO  | ??:??:?? 2 [Power] Opened INA219 sensor on i2c bus
DEBUG | ??:??:?? 2 [Power] Using INA on I2C addr 0x40 for charging detection
DEBUG | ??:??:?? 2 [Power] Battery: usbPower=0, isCharging=0, batMv=4052, batPct=89
```

caveman99 (approved):

And137:
Now my analog measurements on A0 stopped working. Even when I use the -BATTERY_PIN=1 flag, there’s a constant potential around 3,25V on A0 that makes the proper measurement impossible.

PR #6071, opened 2025-02-16
enable LittleFS and encryption on STM32 platform (wio-e5)

dchokola:
- Drop FreeRTOS dependency
  - LittleFS was using FreeRTOS locking functions and malloc()/free()
- LittleFS works
  - properly unlock/lock the FLASH for erasing/writing; fix computation of block sizes
  - mutex functions no-opped
  - malloc()/free() implementations from stdlib
- PKI works
  - depend on latest Crypto library from @caveman99 to make use of STM32 RNG


thebentern (approved):

variants/wio-e5/platformio.ini
```diff
@@ -1,39 +1,27 @@
 [env:wio-e5]
 extends = stm32_base
 board = lora_e5_dev_board
-board_upload.maximum_size = 233472 ; reserve the last 28KB for filesystem
+board_upload.maximum_size = 233472 ; reserve the last 128KB for filesystem
```
markbirss, line 4: This is actually only 28kb ( (262144-233472) / 1024) = 28kb

markbirss:
@dchokola tested ok wio-e5, also able with some modification to get rak3172 working

variants/wio-e5/platformio.ini
caveman99: yes, the 28kb is the correct value. the nRF52840 uses the same flash size.

src/platform/stm32wl/LittleFS.cpp
```diff
@@ -23,41 +23,46 @@
...
-#define STM32WL_SECTOR_SIZE 0x800 /* 2K */
-#define STM32WL_SECTOR_COUNT 14
+#define STM32WL_PAGE_SIZE  (FLASH_PAGE_SIZE)
+#define STM32WL_PAGE_COUNT (FLASH_PAGE_NB)
+#define STM32WL_FLASH_BASE (FLASH_BASE)
 
-#define LFS_FLASH_TOTAL_SIZE (STM32WL_SECTOR_COUNT * STM32WL_SECTOR_SIZE)
-#define LFS_BLOCK_SIZE 128
+/* FLASH_SIZE from stm32wle5xx.h will read the actual FLASH size from the chip */
+/* use the last 1/4 of the FLASH */
```
caveman99, line 39: i'd rather stick to the 28kb proposed. In line with the nRF52840 flash space, which is plenty on there. Using 1/4th of the flash would waste 36kb

src/platform/stm32wl/STM32_LittleFS.h
```diff
@@ -76,17 +76,13 @@ class STM32_LittleFS
      * code. User should not call these directly
      *------------------------------------------------------------------*/
     lfs_t *_getFS(void) { return &_lfs; }
-    void _lockFS(void) { xSemaphoreTake(_mutex, portMAX_DELAY); }
-    void _unlockFS(void) { xSemaphoreGive(_mutex); }
+    void _lockFS(void) { /* no-op */ }
+    void _unlockFS(void) { /* no-op */ }
```
caveman99, line 80: before accepting this, can we make sure the Flash SPI is not shared with the Radio SPI? Indeed FreeRTOS was only included for the locking, not actual threading.

variants/wio-e5/variant.h
```diff
@@ -18,4 +18,9 @@ Do not expect a working Meshtastic device with this target.
 #define LED_PIN PB5
 #define LED_STATE_ON 1
 
+#if (defined(LED_BUILTIN) && LED_BUILTIN == PNUM_NOT_DEFINED)
+#   undef LED_BUILTIN
+#   define LED_BUILTIN (LED_PIN)
+#endif
```
caveman99, line 24: that looks weird with the space between # and macro

caveman99 (changes requested):
Please stick to the original proposal of 28kB filesystem size at the end of the flash area

variants/wio-e5/variant.h
dchokola: I like to indent nested preprocessor directives for readability. The spaces between # and the keyword are a relic of what I learned when first learning C, which in turn is a relic of pre-ANSI C compilers (The first big open-source project I contributed to aimed to be as broadly compatible as possible—something that matters less for microcontroller code.) and poor syntax highlighting. There's a [stackoverflow thread](https://stackoverflow.com/questions/789073/indenting-defines) on the topic. I'll stick to whatever the Meshtastic devs prefer here.

fifieldt: Thanks for the reply! Yes, it does seem to differ depending on the codebase :) 

Personally, I am quite new to C++ (I learnt it for this project!). The way I thought about it was to make the preprocessor directives stand out a lot as a way to discourage their use. In my experience in the Meshtastic project they tend to be used for options, of which we have too many, and device-specific code, which we tend to try and keep in the variant folder.

Just sharing for interest, as someone on the other end of the knowledge spectrum.

caveman99: I'd prefer to keep them together and left-aligend as we do in the rest of the code. also trunk will probably change that anyway :-)

src/platform/stm32wl/LittleFS.cpp
dchokola: See https://github.com/dchokola/meshtastic-firmware/commit/f2bff947bb97b1668b277ea20bfb19cad0be41aa

variants/wio-e5/variant.h
dchokola: I have made the nested preprocessor directives left-aligned/unindented.

caveman99 (approved):

PR #6076, opened 2025-02-17
Validate MQTT config by testing a connection

esev:
Check that a connection can be established with a MQTT server before accepting a configuration change.

This is a follow-on to #6066.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -41,7 +41,6 @@ MQTT *mqtt;
 namespace
 {
 constexpr int reconnectMax = 5;
-constexpr uint16_t mqttPort = 1883;
 
 // FIXME - this size calculation is super sloppy, but it will go away once we dynamically alloc meshpackets
 static uint8_t bytes[meshtastic_MqttClientProxyMessage_size + 30]; // 12 for channel name and 16 for nodeid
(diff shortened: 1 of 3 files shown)


fifieldt:
Q) What about the case where someone is configuring a device for use elsewhere, such as a remote site or a restricted network? They may not be able to have the same network setup or access the MQTT server from where configuration takes place.

esev:
> Q) What about the case where someone is configuring a device for use elsewhere, such as a remote site or a restricted network? They may not be able to have the same network setup or access the MQTT server from where configuration takes place.

If the MQTT server can't be reached while on the local network then this will fail, unfortunately. As long as MQTT isn't enabled though, all the settings can be saved. Once the node was in place all that would be needed would be to toggle the MQTT enabled setting via an admin command.

If the network is different (different WiFi SSID/password) this check is disabled by `if (isConnectedToNetwork())` and then no validation is done and any values can be saved. So an alternate way to configure an unreachable MQTT server could be to disable WiFi and then configure MQTT and finally re-enable WiFi.

Using either of the two ways above will work in this scenario. It isn't as straightforward as it would have been without this PR though. This PR is more targeting the case of mistyped values, and providing some quick feedback to make the user aware that the chosen settings are invalid. Otherwise, without looking at the logs, there is no feedback about invalid settings.

thebentern (approved):

tache:
@esev Ran across this PR when I was searching through the code for the reason why I could not set my mqtt settings. The settings would just fail in the web app and be ignored on the command line. There was no information presented to why this was failing and took two days to figure out. Turning off my node's wifi then saving the mqtt configuration then turning on wifi finally resolved my issue. 
My mqtt server is on my local network, but on another vlan from the meshtastic nodes. The nodes can reach the server/port used for mqtt.

PR #6091, opened 2025-02-19
Trunk: Annotate PRs and Auto-Upgrade

vidplace7:
An attempt at making Trunk less painful in the PR process.

Per the Trunk docs, this should allow inline annotations for PRs, which should making formatting issues more apparent without just saying "run `trunk fmt`".
https://github.com/trunk-io/trunk-action/blob/v1/readme.md#getting-inline-annotations-for-fork-prs

Also adds an upgrade feature, similar to dependabot, which will submit PRs to upgrade Trunk components (in `.trunk`)
https://github.com/trunk-io/trunk-action/blob/v1/readme.md#automatic-upgrades


.github/workflows/nightly.yml
```diff
@@ -4,16 +4,32 @@ on:
...
-    name: Trunk Check Upload
+    name: Trunk Check and Upload
     runs-on: ubuntu-latest
 
     steps:
       - name: Checkout
         uses: actions/checkout@v4
 
       - name: Trunk Check
-        uses: trunk-io/trunk-action@782e83f803ca6e369f035d64c6ba2768174ba61b
```
vidplace7, line 17: @thebentern you explicitly pinned this to a hash years ago, I figure this isn't needed anymore? (v1 should be fine)

fifieldt (approved):
Let's give it a shot!

.github/workflows/nightly.yml
thebentern: I don't even remember _why_ that was done so I say let's move forward for science 😅

thebentern (approved):

PR #6096, opened 2025-02-19
Backup / migrate pub private keys when upgrading to new files in 2.6

thebentern:
Hopefully prevent some headache for users having backup and restore their keys manually

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -972,6 +972,8 @@ void NodeDB::loadFromDisk()
     // disk we will still factoryReset to restore things.
     devicestate.version = 0;
 
+    meshtastic_Config_SecurityConfig backupSecurity = meshtastic_Config_SecurityConfig_init_zero;
+
 #ifdef ARCH_ESP32
     spiLock->lock();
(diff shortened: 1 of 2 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -982,9 +984,18 @@ void NodeDB::loadFromDisk()
...
+        if (loadProto(configFileName, meshtastic_LocalConfig_size, sizeof(meshtastic_LocalConfig), &meshtastic_LocalConfig_msg,
+                      &config) == LoadFileResult::LOAD_SUCCESS &&
+            config.has_security && config.security.private_key.size > 0) {
+            LOG_DEBUG("Saving backup of security config and keys");
+            backupSecurity = config.security;
+        }
+        spiLock->lock();
         rmDir("/prefs");
+        spiLock->unlock();
     }
```
fifieldt, line 998: If FSCom.exists(legacyPrefFileName) and that branch is skipped --> spi won't be unlocked?

thebentern: Good call. I just ran into that condition locally here on second boot 😆 
Good news is that config restore seemed to work great

fifieldt (approved):

PR #6104, opened 2025-02-20
meshtasticd deb: Build armv6-compatible binary

vidplace7:
This should allow us to compile Raspbian `armhf` binaries that are compatible with the PiZero/Pi1 in our current armv7l build environments. Inspiration for this was taken from Raspbian's own packaging.

https://www.valvers.com/open-software/raspberry-pi/bare-metal-programming-in-c-part-1/#rpi1-compiler-flags

This mess is a side-effect of Debian 12 drawing the line at armv7 and calling it `armhf` while Raspbian 12 supports armv6 + hard-float and also calls it "armhf".

Building test builds here https://build.opensuse.org/package/show/home:vidplace7:meshtastic:daily/meshtasticd

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,6 +3,7 @@ Section: misc
 Priority: optional
 Maintainer: Austin Lane <vidplace7@gmail.com>
 Build-Depends: debhelper-compat (= 13),
+               lsb-release,
                tar,
                gzip,
                platformio,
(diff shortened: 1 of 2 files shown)


vidplace7:
Confirmed that this change fixes sefaults on Pi Zero (armv6) :heart:

fifieldt (approved):
If it's good enough for bitflip it's good enough for me!

battlehax:
Yes, tested and working on both `Raspberry Pi Zero Rev 1.3` and a `Raspberry Pi 2 Model B Rev 1.1` here!

thebentern (approved):

PR #6109, opened 2025-02-20
Ignore and disallow multi-hop traceroutes destined to broadcast address

### LINKED ISSUE #6108: [Bug]: Traceroutes to `!ffffffff` shouldn't be allowed
### Category

Other

### Hardware

Not Applicable

### Firmware Version

2.5

### Description

Relaying this report from discord:

https://discord.com/channels/867578229534359593/919642584480112750/1342192903073824809

When tracerouting to `!ffffffff`, firmware doesn't prevent it (or refuse to respond to it), which is generally a problematic thing to allow. I suspect we should make the traceroute module reject these.

### Relevant log output

```Shell

```

GUVWAF:
Fixes #6108.

### DIFF
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -643,6 +643,11 @@ bool PhoneAPI::handleToRadioPacket(meshtastic_MeshPacket &p)
         meshtastic_QueueStatus qs = router->getQueueStatus();
         service->sendQueueStatusToPhone(qs, 0, p.id);
         return false;
+    } else if (p.decoded.portnum == meshtastic_PortNum_TRACEROUTE_APP && isBroadcast(p.to) && p.hop_limit > 0) {
+        sendNotification(meshtastic_LogRecord_Level_WARNING, p.id, "Multi-hop traceroute to broadcast address is not allowed");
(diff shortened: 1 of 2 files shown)


esev:
Ignore this comment if this is too much work, I can change my code to iteratively generate traceroutes: I used this as a convenient way to ask my 0 hop neighbors for their receive SnR values. Could it still be allowed for direct (0 hop) neighbors?

thebentern (approved):

thebentern:
> Ignore this comment if this is too much work, I can change my code to iteratively generate traceroutes: I used this as a convenient way to ask my 0 hop neighbors for their receive SnR values. Could it still be allowed for direct (0 hop) neighbors?

That is interesting!

GUVWAF:
> Could it still be allowed for direct (0 hop) neighbors?

Yes, should be possible. Broadcast is actually more efficient than multiple sequentially, although the chance of success it likely higher with the latter.

I can change it to allow 0-hop traceroutes if that's considered as a reasonable middleground.

GUVWAF:
Now that I think of it; maybe it's better to also disallow the request (and send a client notification)?

GUVWAF:
Implemented the above with the last two commits.

PR #6110, opened 2025-02-20
Perhaps fix TXCO reports on pro-micro

thebentern:


fifieldt (approved):
I don't think this can hurt ....

PR #6112, opened 2025-02-21
Add support for REYAX RYS352x

fifieldt:
Thanks to the work by @lukejenkins, we can see the detection message for this affordable AG3352-based chip.

Adds detection code and sets it to use common AG3352 configuration.


caveman99:
@fifieldt once we merge the improved scancode this (and the other one) probably needs a revisit.

fifieldt:
Not needed anymore - turns out I added this already in the big GPS probe patch.

PR #6114, opened 2025-02-21
Only call GPS Probe commands once per family

fifieldt:
In the GPS probe code we write commands on the serial line and
 determine which GPS we have based on the result.

GPS units in the same family sometimes use the same command,
 but return different results (eg AG3335 and AG3332 both use $PAIR021*39).
Currently we run the command once per GPS. Instead we should run each command only once per family, record the result, and select the GNSS MODEL
 based on the result, which is what this patch does.

Before the change, we put 12 commands on the serial bus. Now we only put 6.

This should markedly improve the speed and reliability of GPS detection.

Fixes https://github.com/meshtastic/firmware/issues/5193

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -1,3 +1,7 @@
+#include <cstring> // Include for strstr
+#include <string>
+#include <vector>
+
 #include "configuration.h"
 #if !MESHTASTIC_EXCLUDE_GPS
 #include "Default.h"
@@ -1117,7 +1121,7 @@ int GPS::prepareDeepSleep(void *unused)
 }
 
 static const char *PROBE_MESSAGE = "Trying %s (%s)...";
-static const char *DETECTED_MESSAGE = "%s detected, using %s Module";
+static const char *DETECTED_MESSAGE = "%s detected";
 
 #define PROBE_SIMPLE(CHIP, TOWRITE, RESPONSE, DRIVER, TIMEOUT, ...)                                                              \
     do {                                                                                                                         \
@@ -1125,8 +1129,19 @@ static const char *DETECTED_MESSAGE = "%s detected, using %s Module";
         clearBuffer();                                                                                                           \
         _serial_gps->write(TOWRITE "\r\n");                                                                                      \
         if (getACK(RESPONSE, TIMEOUT) == GNSS_RESPONSE_OK) {                                                                     \
-            LOG_INFO(DETECTED_MESSAGE, CHIP, #DRIVER);                                                                           \
+            LOG_INFO(DETECTED_MESSAGE, CHIP);                                                                                    \
             return DRIVER;                                                                                                       \
+        }                                                                                                                        \
+    } while (0)
+
+#define PROBE_FAMILY(FAMILY_NAME, COMMAND, RESPONSE_MAP, TIMEOUT)                                                                \
+    do {                                                                                                                         \
+        LOG_DEBUG(PROBE_MESSAGE, COMMAND, FAMILY_NAME);                                                                          \
+        clearBuffer();                                                                                                           \
+        _serial_gps->write(COMMAND "\r\n");                                                                                      \
+        GnssModel_t detectedDriver = getProbeResponse(TIMEOUT, RESPONSE_MAP);                                                    \
+        if (detectedDriver != GNSS_MODEL_UNKNOWN) {                                                                              \
+            return detectedDriver;                                                                                               \
         }                                                                                                                        \
     } while (0)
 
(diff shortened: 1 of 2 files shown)


fifieldt:
Converting to draft - going to test in every chip I have.

src/gps/GPS.cpp
```diff
@@ -1192,31 +1212,34 @@ GnssModel_t GPS::probe(int serialSpeed)
...
 
     // Unicore UFirebirdII Series: UC6580, UM620, UM621, UM670A, UM680A, or UM681A
-    PROBE_SIMPLE("UC6580", "$PDTINFO", "UC6580", GNSS_MODEL_UC6580, 500);
-    PROBE_SIMPLE("UM600", "$PDTINFO", "UM600", GNSS_MODEL_UC6580, 500);
-    PROBE_SIMPLE("ATGM336H", "$PCAS06,1*1A", "$GPTXT,01,01,02,HW=ATGM336H", GNSS_MODEL_ATGM336H, 500);
-    /* ATGM332D series (-11(GPS), -21(BDS), -31(GPS+BDS), -51(GPS+GLONASS), -71-0(GPS+BDS+GLONASS))
-    based on AT6558 */
-    PROBE_SIMPLE("ATGM332D", "$PCAS06,1*1A", "$GPTXT,01,01,02,HW=ATGM332D", GNSS_MODEL_ATGM336H, 500);
+    std::vector<ChipInfo> unicore = {{"UC6580", "$PDTINFO", GNSS_MODEL_UC6580}, {"UM600", "$PDTINFO", GNSS_MODEL_UC6580}};
+    PROBE_FAMILY("Unicore Family", "$PDTINFO", unicore, 500);
```
fifieldt, line 1216: I think PDTINFO is wrong here - should be the chip name.

fifieldt:
Hmm, it looks like the new T1000E GPS detection "improvement" code slows down the GPS coming up to the extent the new faster detection code here runs before it comes up!

fifieldt:
Example from T1000E

Before:

```DEBUG | ??:??:?? 7 [GPS] Probe for GPS at 115200
DEBUG | ??:??:?? 7 [GPS] Trying $PDTINFO (UC6580)...
DEBUG | ??:??:?? 8 [GPS] Trying $PDTINFO (UM600)...
DEBUG | ??:??:?? 8 [GPS] Trying $PCAS06,1*1A (ATGM336H)...
DEBUG | ??:??:?? 9 [GPS] Trying $PCAS06,1*1A (ATGM332D)...
DEBUG | ??:??:?? 9 [GPS] Trying $PAIR021*39 (AG3335)...
INFO | ??:??:?? 9 [GPS] AG3335 detected, using GNSS_MODEL_AG3335 Module```

After:
```DEBUG | ??:??:?? 7 [GPS] Trying $PDTINFO (Unicore Family)...
DEBUG | ??:??:?? 8 [GPS] Trying $PCAS06,1*1A (ATGM33xx Family)...
DEBUG | ??:??:?? 8 [GPS] Trying $PAIR021*39 (Airoha Family)...
INFO | ??:??:?? 8 [GPS] AG3335 detected
```

fifieldt:
Tested with Tbeam/UBlox 6 - works nicely

fifieldt:
Seeed Wio-WM1110
```DEBUG | ??:??:?? 5 [GPS] Probe for GPS at 9600
DEBUG | ??:??:?? 5 [GPS] Trying $PDTINFO (Unicore Family)...
DEBUG | ??:??:?? 5 [GPS] Trying $PCAS06,1*1A (ATGM33xx Family)...
DEBUG | ??:??:?? 6 [GPS] Trying $PAIR021*39 (Airoha Family)...
DEBUG | ??:??:?? 7 [GPS] Trying $PQTMVERNO*58 (LC86)...
DEBUG | ??:??:?? 8 [GPS] Trying $PCAS06,0*1B (L76K)...
INFO | ??:??:?? 8 [GPS] L76K detected```

thebentern:
> Hmm, it looks like the new T1000E GPS detection "improvement" code in GPS::setup slows down the GPS coming up to the extent the new faster detection code here runs before it comes up!

Can we set the thread to start up on a delay in the case of the T1000E perhaps? I'm trying to recall if that had other unintended consequences.

fifieldt:
> Can we set the thread to start up on a delay in the case of the T1000E perhaps? I'm trying to recall if that had other unintended consequences.

I think I was wrong with my earlier comment. What I've done is set the time for the airoha probe command to 1000 millis, rather than 500, and it seems very happy now!

fifieldt:
> add a definition for the most common GPS and baudrate

Set GPS_BAUDRATE in variant.h and you'll get the second part of that straight away :)

mverch67:
> > add a definition for the most common GPS and baudrate
> 
> Set GPS_BAUDRATE in variant.h and you'll get the second part of that straight away :)

yes, but then it does not do real probing anymore (in case another GPS is equipped), but only after removing the mentioned macros definitions in configuration.h

fifieldt:
Ok, perhaps put in a separate feature request for that and I'll take a look?

fifieldt:
Example of the speedup for this one from T-beam: 8 seconds to 4 seconds.

I don't have a T-deck, but based on above I would expect this code would detect its GPS in about 15 seconds now.

One of the other optimizations I hope to be able to catch is: most GPSes put their version info on the line when they first turn on. If it's possible to catch that we can do detection without sending a single command.

fifieldt:
> do detection without sending a single command.

hmm! that was easy. I got the heltec wireless tracker to work in this way. Now to see if that was a fluke.

fifieldt:
> Now to see if that was a fluke.

Yeah, sadly. Not ready to add this in yet :)

caveman99 (approved):

fifieldt:
Yes, ready to go, thanks!

PR #6116, opened 2025-02-21
GPS Factory Reset no longer needed.

fifieldt:
In 2020 there was a bad batch of tbeams whose ubloxen went a little loopy. We factory reset them to bring them sane again.

It's now 2025, the problem with tbeam is long fixed, and this is not necessary for any other devices we're aware of. Removing the function to make our code more maintainable.

There is an associated protobuf entry did_gps_reset that we might be able to re-purpose for something else, or remove in 3.0.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -48,8 +48,6 @@ HardwareSerial *GPS::_serial_gps = nullptr;
 
 GPS *gps = nullptr;
 
-static const char *ACK_SUCCESS_MESSAGE = "Get ack success!";
-
 static GPSUpdateScheduling scheduling;
 
 /// Multiple GPS instances might use the same serial port (in sequence), but we can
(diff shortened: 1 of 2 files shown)


thebentern:
I think this one is probably safe to put to bed as well as a regular part of the startup process. I do wonder if we should make it an admin command to be issued from the CLI or the clients in the case of troubleshooting some of GPS issue, since the alternative is flashing an alternative firmware to run the factory reset and then reflashing meshtastic.

fifieldt:
> I do wonder if we should make it an admin command to be issued from the CLI or the clients in the case of troubleshooting some of GPS issue,

I bow to those with experience :) So far I haven't yet come across an issue that I couldn't debug with GPS_DEBUG, and fix with changing relevant startup config commands for all devices of the type.

Come to think of it, in theory our startup configuration commands do a 'soft' reset. Those turn off all NMEA messages and turn on just the ones we want, as well as changing and saving relevant config.

Share the wisdom and I'll try to hack to make it reality.

thebentern:
@caveman99 what are your thoughts on ditching the factory reset entirely vs making an explicit command to issue from clients in the case of troubleshooting?

caveman99:
@thebentern i don't think we need it at all. this was meant to bring T-Echo (not t-beam) L76K chips out of the 4800 baud limbo we put them in deliberately. these bad firmware versions were "years ago".

If anyone needs to factory reset them, there's a sketch on the lilygo website we can adapt for universal use, pretty much like we do with the nrf flash erase.

caveman99 (approved):

thebentern:
Kill it and grill it! :D

PR #6123, opened 2025-02-22
Cast user pref strings.

Mictronics:
Fixes the following build errors when user preference strings are numerical only. e.g. short name is 4 numbers.
```
error: invalid conversion from 'int' to 'const char*' [-fpermissive]
error: initializer fails to determine size of 'slipstreamTZString'
error: array must be initialized with a brace-enclosed initializer
```


fifieldt (approved):
Good catch, thanks.

PR #6124, opened 2025-02-22
Remove the usage of "bro"

sycophantic:
Removing the usage of the term "bro" in the log message.


thebentern:
Bro?! 😄 
https://www.youtube.com/watch?v=Kvc5BuLB5b0

fifieldt (changes requested):
Thank you for sending in your first patch to Meshtastic. This line is an [Easter Egg](https://en.wikipedia.org/wiki/Easter_egg_(media)), rather than a bug :)

We appreciate you signing the CLA and taking the time to submit a patch, and would welcome further contributions from you! For this one, since it's our way of having a fun in-joke, we might prefer it to stay.

PR #6140, opened 2025-02-24
remove second initSPI() in main

Woutvstk:
A second initSPI was added in a merge from master. Caused spilock issue


fifieldt (approved):

fifieldt:
Nice catch. I found the original commit where the dupe was added and this definitely needs fixing!

PR #6153, opened 2025-02-25
More trunk junk / remove old workflows

vidplace7:
Resolves more trunk items. Removes unused GitHub Actions workflows.

Also explicitly sets Trunk workflows to `runs-on: ubuntu-24.04` so it's never scheduled on a self-hosted runner.


.github/workflows/generate-userprefs.yml
vidplace7, line 1: Unused since switching to userprefs.jsonc

fifieldt (approved):

PR #6164, opened 2025-02-26
chore: better err logging in radiolib

rostekus:
I added an additional check for specific Radiolib and included logs.

Should I also add logs in places where there are none? For example:

```cpp
    err = lora.setBandwidth(bw);
    if (err != RADIOLIB_ERR_NONE)
        RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
```

### DIFF
--- a/src/mesh/SX128xInterface.cpp
+++ b/src/mesh/SX128xInterface.cpp
@@ -131,7 +131,9 @@ template <typename T> bool SX128xInterface<T>::reconfigure()
         RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
 
     err = lora.setSyncWord(syncWord);
-    if (err != RADIOLIB_ERR_NONE)
+    if (err == RADIOLIB_ERR_WRONG_MODEM)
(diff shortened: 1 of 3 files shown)


src/mesh/SX128xInterface.cpp
```diff
@@ -131,7 +131,9 @@ template <typename T> bool SX128xInterface<T>::reconfigure()
         RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
 
     err = lora.setSyncWord(syncWord);
-    if (err != RADIOLIB_ERR_NONE)
+    if (err == RADIOLIB_ERR_WRONG_MODEM)
+        LOG_ERROR("SX128X setSyncWord: WRONG MODEM error %s%d", radioLi / bErr, err);
```
fifieldt, line 135: Syntax error here

caveman99 (changes requested):
That's an awful lot of verbiage added. Can we just output the numerical error value when something goes wrong? That would save a bit of flash space.

caveman99:
This PR seems abandoned. Please re-open if neccessary

PR #6184, opened 2025-03-01
Only request NodeInfo/Position from everyone on fresh install

GUVWAF:
Previously, we would request NodeInfo and Position from *all* nodes every time a node boots (NodeDB init calls `resetRadioConfig()` which increments `radioGeneration`). With this, we'll only do this on a fresh install (when `installDefaultConfig()` is called).


caveman99:
The initial idea about radioGeneration was to reconfigure the Lora settings without a reboot. this counter was incemented as a signal to throw away all old memory content that belonged to the old preset. I don't think it has any side effects nowadays, since we reboot after config changes anyway. Still it would be good to double check if this does not inhibit functionality in the NodeInfoModule and the PositionModule. It's not used in other places any more.

caveman99 (approved):

thebentern (approved):

PR #6189, opened 2025-03-01
Enable external (UART) GPS support on Seeed WM1110 tracker dev board

thebentern:
Prior to these changes, upon initial boot, the lack of `HAS_GPS` macro definition resulted in the gps mode being set to `NOT_PRESENT` and the Seeed studio groove connector GPS module would never be scanned for and detected. 
After these changes, it is detected as an L76K and seems to work fine.


fifieldt (approved):
Yup! That'd do it!

PR #6190, opened 2025-03-01
native: do not include `device-ui` in build sources

vidplace7:
With the inclusion of the `device-ui` submodule in `master` now, `native` debian-src builds have ballooned from 110 MB to 800 MB.
This is mostly thanks to the `lvgl` dependency included for `device-ui`.

These deps aren't needed unless you're building with the UI bundled (`native-tft`), so let's exclude them.


vidplace7:
lvgl still seems to be included -_- converted to DRAFT

vidplace7:
Taking a different approach to resolve this more cleanly.
- https://github.com/meshtastic/firmware/pull/6193

Issue #6196, opened 2025-03-02
[Feature Request]: Avoid full refresh when typing on Eink displays


HarukiToreda:
### Platform

Cross-Platform

### Description

Currently most eink nodes have a max limit for partial refresh at which after that limit is hit it will force the screen to do a Full refresh, which is no problem usually. However when typing on a keyboard, the screen will refresh after 10 characters causing missed input and interrupting the typing experience with them. I propose removing the 10 refresh limit only when on the free text screen.

https://github.com/user-attachments/assets/cc49383e-e7b5-4068-9307-237db3ae6943

fifieldt:
Fixed by https://github.com/meshtastic/firmware/pull/6201

PR #6201, opened 2025-03-02
E-ink partial refresh limitation removed for free text screen

HarukiToreda:
added an UNLIMITED_FAST flag for eink displays to allow more input temporarily while using freetext screen. All other behavior remains unchanged. This will allow keyboard devices to type without the annoying full refresh flash. As mentioned on PR https://github.com/meshtastic/firmware/issues/6196

### DIFF
--- a/src/modules/CannedMessageModule.cpp
+++ b/src/modules/CannedMessageModule.cpp
@@ -1057,6 +1057,11 @@ void CannedMessageModule::drawFrame(OLEDDisplay *display, OLEDDisplayUiState *st
         display->drawString(10 + x, 0 + y + FONT_HEIGHT_SMALL, "Canned Message\nModule disabled.");
     } else if (cannedMessageModule->runState == CANNED_MESSAGE_RUN_STATE_FREETEXT) {
         requestFocus(); // Tell Screen::setFrames to move to our module's frame
+#ifdef USE_EINK
+        EInkDynamicDisplay* einkDisplay = static_cast<EInkDynamicDisplay*>(display);
+        einkDisplay->enableUnlimitedFastMode();  // Enable unlimited fast refresh while typing
+#endif
+
 #if defined(USE_VIRTUAL_KEYBOARD)
         drawKeyboard(display, state, 0, 0);
 #else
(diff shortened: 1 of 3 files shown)


src/modules/CannedMessageModule.cpp
```diff
@@ -1057,6 +1057,11 @@ void CannedMessageModule::drawFrame(OLEDDisplay *display, OLEDDisplayUiState *st
         display->drawString(10 + x, 0 + y + FONT_HEIGHT_SMALL, "Canned Message\nModule disabled.");
     } else if (cannedMessageModule->runState == CANNED_MESSAGE_RUN_STATE_FREETEXT) {
         requestFocus(); // Tell Screen::setFrames to move to our module's frame
+#ifdef USE_EINK
+        EInkDynamicDisplay* einkDisplay = static_cast<EInkDynamicDisplay*>(display);
+        einkDisplay->enableUnlimitedFastMode();  // Enable unlimited fast refresh while typing
```
caveman99, line 1062: Wouldn't you want to disable this mode again when leaving the freetext mode?

HarukiToreda: When I was testing, once i leave the freetext mode it forces a full refresh and the unlimited refresh mode stops on all other screens. All other screens go back to normal 10 refresh limit without disabling the mode. Should I still disable it despite it working as is?

caveman99: If it stops implicitly no need to add it. I didn't know it worked that way :-)

caveman99 (approved):

caveman99:
@HarukiToreda RAK4631 EInk does not look happy.

HarukiToreda:
this change should satisfy that pesky RAK ink module

HarukiToreda:
Demo of how it behaves now

https://github.com/user-attachments/assets/2d5aa8c7-a2d5-4d43-82ba-21a4f41914a9

caveman99:
Ship it then :-)

markbirss:
@HarukiToreda the UNLIMITED_FAST needs to be set in the varaint platformio.ini ?

markbirss:
> @HarukiToreda RAK4631 EInk does not look happy.

I can see this variant is using GxEPD2 support and not the meshtastic/GxEPD2 support which includes EINK_DYNAMICDISPLAY. 

@caveman99 did you try it as is ?

@HarukiToreda and @todd-herbert if you are available some time i want to discuss fast and partial updates when using keyboard input

HarukiToreda:
> > @HarukiToreda RAK4631 EInk does not look happy.
> 
> I can see this variant is using GxEPD2 support and not the meshtastic/GxEPD2 support which includes EINK_DYNAMICDISPLAY.
> 
> @caveman99 did you try it as is ?
> 
> @HarukiToreda and @todd-herbert if you are available some time i want to discuss fast and partial updates when using keyboard input in more detail

@markbirss do you think after our conversation that this is good to go forward? or is there something else you're still testing?

markbirss:
@HarukiToreda yes, it should still be good for the variants that are using EINK_DYNAMICDISPLAY imo

caveman99:
I was at embedded world last week so now just catching up.

fifieldt (approved):

Issue #6186, opened 2025-03-01
[Task]: 2.6 device-install scripts need to be updated


thebentern:
### Category

Other

### Hardware

Not Applicable

### Firmware Version

2.6.0

### Description

The device install scripts need to be reworked to account for several changes in 2.6:

1. Each ESP32 target has its own littlefs files now, intead of per platform. e.g. littlefs-{target}...  littlefswebui-{target}...
2. If a target of `-tft` is chosen, we need use the specific partition scheme for the target: 8MB or 16MB
Partition offsets can be found here:
https://github.com/meshtastic/web-flasher/blob/1c457165d432e7996d89375c1f57d45d5f096a34/stores/firmwareStore.ts#L204
Targets partition schemes are documented in the device hardware api list:
https://github.com/meshtastic/web-flasher/blob/main/types/resources.ts#L573

### Relevant log output

```Shell

```

ThatKalle:
How do we handle the scenario where a `-ftf-` target is selected; are you able to flash the `littlefswebui-{target}` file then? Or forced to go with `littlefs-{target}`?

The web flasher does not allow you to select WebUI and MUI at the same time.

PR #6205, opened 2025-03-02
device_install changes for 2.6

gjelsoe:
For #6186 .
Added 2 new arguments --tft and -tft-16mb
Some checks are added.
Before it would try to write all files to the device, if there was more than ONE littlefs-* or littlefswebui-* in the directory.

Added OTA Offsets for 8 and 16mb (fix)
Thanks to @caveman99 for spotting it.


caveman99:
I think you can send the scripts directly to master. No need to go to 2.6 branch any more.

PR #6219, opened 2025-03-03
fix(grammar): change shutdown to shut down when used as a verb

### LINKED ISSUE #6211: Grammar error
https://github.com/meshtastic/firmware/blob/f89f916f966bc0dece39c3d0ad3324e8f326092c/src/graphics/niche/InkHUD/Applets/System/Menu/MenuApplet.cpp#L242

Suggest changing “shutdown” to “shut down” in the “Save & shutdown” interface. It must be two words to be grammatically correct as an action/verb: https://linguix.com/english/common-mistake/shutdown

Frigyes06:
fixes #6211


todd-herbert:
This is handled already by existing PR https://github.com/meshtastic/firmware/pull/6216

Issue #6213, opened 2025-03-03
[Bug]: GPIO incorrect for Lilygo Lora32 2.1 1.6.1


edwar64896:
### Category

Other

### Hardware

T-Lora v2 1.6

### Firmware Version

2.5.20 (latest beta)

### Description

have connected a switch across IO12 and GND as per doc, however this doesn't appear to have any effect as a user button. Continuity and circuit both checked. Am currently checking code for this board to see where the GPIO is configured...

PIN_BUTTON defined as '12' in Device Config (Android app)

### Relevant log output

```Shell

```

edwar64896:
3ae85e2 appears to be the commit where BUTTON_PIN was nuked. I read somewhere that for the PIN_BUTTON configuration parameter to function, BUTTON_PIN must be defined...

edwar64896:
Further investigation: as this is my first foray into this repository, not entirely sure yet exactly where buttonGpio is getting handled for the Lilygo TLora32, but certainly, putting #define BUTTON_PIN 12 back into variant.h fixes the issue. I suspect though this isn't the end of the story.

edwar64896:
ok here we are...

https://github.com/meshtastic/firmware/blob/b6562e175fe90e0eeb069d6dda9e6a70ff18c096/src/ButtonThread.cpp#L31

Suspect that there are going to be more references like this meaning that really, BUTTON_PIN is going to have to be defined in variant.h for now unless someone wants to dig through the whole codebase to fix the dependency!

ndoo:
Seems like the logic that makes the user definition not work needs to be fixed instead. Because restoring this causes electrical noise to trigger the GPIO since it's floating (as expected) since the board has no pull up/down resistors on the pin (as it was never intended to be used this way).

todd-herbert:
The preferred way for tlora_v2_1_16 to add a user button would be by [setting config.device.button_gpio](https://meshtastic.org/docs/configuration/radio/device/#gpio-for-user-button).  It looks to me though like setting this value won't have any impact if the variant doesn't originally define a `BUTTON_PIN` or `USERPREFS_BUTTON_PIN`.

Looking back through git blame, this one might be my fault. Sorry about that..
> Fix button interrupt after light sleep (https://github.com/meshtastic/firmware/pull/3587)

> * Don't assume device has a button.. 

[...]

PR #6222, opened 2025-03-04
fix for issue #6213

edwar64896:
fix for issue #6213 

feel free to update comments. 

This partially undoes a patch by @ndoo which broke access to user buttons on the Lilygo TLora v2.1 series of boards.


caveman99:
While this will fix your immediate issue, we need to go about this in a different way. We don't use definitions for something that's not attached to the board when shipped. SO we can't define this button. Defining them through protobufs/config settings is the way forward, however the current logic requires a default pin to override it. If you can change the logic for config.device.button_gpio to work even when no pin to override is defied, that would be most welcome.

PR #6229, opened 2025-03-04
Enable GPS functionality for RAK4631_eth_gw variant

mandreko:
I had recently acquired a RAK4631 and used the RAK ethernet module. Using the rak4631_eth_gw variant made it so that I could not utilize GPS and put my node on meshmap.net, for example.

This PR re-enables GPS functionality for the variant. This could be a fixed location or the GPS module from RAK. But it should at least allow the location to be updated and for MQTT gateway nodes to show up on meshmap.

I suspect that others may run into this issue once the new WisMesh device from RAK is shipped, since it uses the same variant firmware: https://store.rakwireless.com/products/wismesh-ethernet-gateway


caveman99 (approved):

mandreko:
Just for documentation sake, @caveman99 checked with the designer, and found that the GPS module can not be used simultaneously with the ethernet module. However, GPS functionality should likely not be entirely disabled as it stops fixed-location from functioning properly.

caveman99:
Size after building with GPS code enabled:
`Flash: [========  ]  76.7% (used 625000 bytes from 815104 bytes)`

I think we can justify this.

PR #6233, opened 2025-03-05
Add initial support for CrowPanel ESP32 5.79” E-paper HMI

markbirss:
https://www.elecrow.com/crowpanel-esp32-5-79-e-paper-hmi-display-with-272-792-resolution-black-white-color-driven-by-spi-interface.html

Working
* sdcard
* lora sx1262 ,i2c and buzzer  on gpio pins
* buttons
* 5.79 inch 272*792 e-ink

Breakout Pinout
image

Adds
Monospace 30 font as the display is huge and requires a larger font

image

image

image


droberin:
Tested replacing eInk controller and resolution on a CrowPanel ESP32 4.2” E-paper HMI.
Looking forward for a variant for this one too!

testing_crowpanel

markbirss:
https://github.com/meshtastic/GxEPD2/pull/10 to add these displays to mehstastic GxEPD2 branch

src/graphics/EInkDisplay2.cpp
```diff
@@ -182,6 +182,9 @@ bool EInkDisplay::connect()
         // Init GxEPD2
         adafruitDisplay->init();
         adafruitDisplay->setRotation(3);
+#if defined(CROWPANEL_ESP32S3_5_EPAPER)
+        adafruitDisplay->setRotation(0);
```
caveman99, line 186: I assume calling this twice does no harm?

src/main.cpp
```diff
@@ -282,6 +282,14 @@ void setup()
     delay(100);
 #endif
 
+#if defined(CROWPANEL_ESP32S3_5_EPAPER)
```
caveman99, line 285: We have a LOT of defines already switching GPIO's around. Can you hijack one of the existing defines for these? Would save us a few bytes of cache.

src/graphics/EInkDisplay2.cpp
markbirss: The other displays are oriented differently

src/main.cpp
markbirss: let me see

markbirss: @caveman99 ok i have removed the power enable stuff from main and used the other defined enable pins ie VEXT_ENABLE  and PIN_POWER_EN, defined in the variant

caveman99 (approved):

PR #6249, opened 2025-03-06
i2c: 0x45 can also be an SHT35, not just an OPT3001

cdanis:
I have an SHT35 sensor breakout from Seeed that defaults to address 0x45.  Without this patch, it's erroneously detected as an OPT3001.

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -135,6 +135,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define LPS22HB_ADDR 0x5C
 #define LPS22HB_ADDR_ALT 0x5D
 #define SHT31_4x_ADDR 0x44
+#define SHT31_4x_ADDR_ALT 0x45
 #define PMSA0031_ADDR 0x12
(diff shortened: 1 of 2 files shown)


fifieldt (changes requested):
Thanks for this.

The patch as it stands will break the OPT3001.

What needs to happen is to find a way to differentiate the OPT3001 and SHT35 using register value lookups.

cdanis:
> Thanks for this.
> 
> The patch as it stands will break the OPT3001.
> 
> What needs to happen is to find a way to differentiate the OPT3001 and SHT35 using register value lookups.

The pre-existing line 358 does a register lookup to check for OPT3001 by a known register value.

fifieldt (approved):

fifieldt:
So it does! It was hidden in the default view, sorry about that.

cdanis:
> So it does! It was hidden in the default view, sorry about that.

The funniest part was GitHub wouldn't let me open a comment on that line because it's not part of the patch 😅

PR #6268, opened 2025-03-08
stm32wl-fs: Align E77-MBL DevKit

vidplace7:
Change E77-MBL `build_flags` / `maximum_size` to align with other STM32 platforms in the `stm32wl-fs` branch


variants/CDEBYTE_E77-MBL/platformio.ini
markbirss, line 1: reserve the last 28KB for filesystem

vidplace7: Ahh, I see it's noted as 28KB in the RAK variant and 128KB in the WIO-E5 variant. I fixed the E5's comment as well now :+1:

PR #6311, opened 2025-03-13
Simplify MAX_NUM_NODES, increase limit on large-flash targets.

vidplace7:
This PR simplifies `MAX_NUM_NODES` by putting the main logic behind common macros, rather than strewing them through each variant. Additionally, moves larger partitionmaps from the `-tft` variants "upstream" into the parent (non-tft) env.

Also adds support for larger nodeDBs for high-flash devices like the Station G2.

Increases NodeDB Size to `200` for 8MB targets and `250` for 16MB targets.
- [x] Station G2 (16MB Flash)
- [x] Lilygo T-Watch S3 (16MB Flash)
- [x] Lilygo T-ETH Elite (16MB Flash)
- [x] Lilygo T-Beam Supreme `S3-Core` (8MB Flash)
- [x] Xiao-S3 (8MB Flash)
- [x] ESP32S3 Heltecs - v3, Tracker, Vision Master Series, Wireless Paper (8MB Flash)
- [x] (DIY) NanoVHF T-Energy-S3 + E22 (16MB Flash)
- [x] and a few others (see the changed files)

Tested / working on:
- Station G2 (250 nodeDB)
- Heltec v3 (200 nodeDB)
- Heltec Tracker (200 nodeDB)


src/mesh/mesh-pb-constants.h
vidplace7, line 1: This is the main change. Much of the rest is PlatformIO noise.

thebentern:
Potentially breaking change for these targets when we change the partition scheme I think

vidplace7:
> Potentially breaking change for these targets when we change the partition scheme I think

2.6 is already a "you must wipe your device" FS-breaking change. So I'd really like to get this merged before our first Beta if possible :pray:

src/mesh/mesh-pb-constants.h
```diff
@@ -20,8 +20,20 @@
 
 /// max number of nodes allowed in the mesh
 #ifndef MAX_NUM_NODES
+#ifdef ARCH_STM32WL
+#define MAX_NUM_NODES 10
+#elif defined(ARCH_NRF52)
+#define MAX_NUM_NODES 80
+#elif defined(HAS_PSRAM)
```
GUVWAF, line 27: IMO it's not logical to base this on the amount of PSRAM if that's not used to store the NodeDB.
E.g. if there will be a RP2040 or ESP32-C3 variant with PSRAM, it doesn't mean it can handle a larger NodeDB because those have less RAM than the ESP32-S3 variants that currently use a limit larger than 100.

vidplace7: Now basing on `ARCH_ESP32_S3` + `FLASH_MB`

arch/esp32/esp32c3.ini
```diff
@@ -4,3 +4,7 @@ custom_esp32_kind = esp32c3
 
 monitor_speed = 115200
 monitor_filters = esp32_c3_exception_decoder
+
+build_flags =
+  ${esp32_base.build_flags}
+  -D ARCH_ESP32_C3
```
caveman99, line 10: please don't do that. there are predefined macros to find out which flavour of ESP32 you are on.

variants/CDEBYTE_EoRa-S3/platformio.ini
```diff
@@ -5,4 +5,5 @@ build_flags =
   ${esp32s3_base.build_flags}
   -D CDEBYTE_EORA_S3
   -I variants/CDEBYTE_EoRa-S3
-  -D GPS_POWER_TOGGLE
\ No newline at end of file
+  -D GPS_POWER_TOGGLE
+  -D FLASH_MB=4
```
caveman99, line 9: The flash size can also be derived from the board features.

caveman99 (changes requested):
Apart from the actual partition size change the platformio noise is unneccesary.

variants/CDEBYTE_EoRa-S3/platformio.ini
vidplace7: Reworked this to use `ESP.getFlashChipSize()` :heart:

vidplace7:
Reworked to use `ESP.getFlashChipSize()` and the builtin IDF macros, as suggested by @caveman99

This allows us to greatly simplify the platformio.ini changes, now only the new partition layout is needed :heart:

thebentern (approved):

thebentern:
I also realized, we need to address the decice-install scripts as a part of this PR 😓

bin/device-install.sh
vidplace7, line 1: @ThatKalle Can you please review the changes in this file? :pray: I'm no bash expert :sweat_smile: but this should be a bit easier to maintain.

vidplace7, line 1: @ThatKalle Can you please review the changes to `device-install.sh`.
With all these new variants with large partitions, it had to be reworked a bit.

ThatKalle: I think it looks good! I do like the idea of using the groupings combined with a for loop rather than the long `if [ -n .. ] && .. && ..` akin to how the `.bat` version does it. 👍

PR #6317, opened 2025-03-14
Add RAK4631 E-PAPER Dynamic variant

markbirss:
need some more testers, but should be good to merge

@caveman99 
here are changes meant for testing with https://github.com/meshtastic/firmware/pull/6201

switches to meshtastic/GxEDP2 branch for the e-ink display
enables EINK_DYNAMICDISPLAY

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -198,11 +198,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 /* Step #1: offer chance for variant-specific defines */
 #include "variant.h"
 
-// Define BUTTON_PIN to ensure button setup is always done
-#ifndef BUTTON_PIN
-#define BUTTON_PIN (-1)
-#endif
-
 #if defined(VEXT_ENABLE) && !defined(VEXT_ON_VALUE)
 // Older variant.h files might not be defining this value, so stay with the old default
 #define VEXT_ON_VALUE LOW


src/graphics/EInkDynamicDisplay.cpp
todd-herbert, line 1: Wait, what's going on here?

src/graphics/EInkDynamicDisplay.cpp
```diff
@@ -311,8 +311,10 @@ void EInkDynamicDisplay::checkFrameMatchesPrevious()
     }
 #endif
 
-    // Not redrawn, not COSMETIC, not DEMAND_FAST
+// Not redrawn, not COSMETIC, not DEMAND_FAST
+#if defined(RAK_4631_D)
```
todd-herbert, line 315: Wait, what's going on here?

todd-herbert, line 315: What's the intention with this change?

variants/rak4631_epaper/platformio.ini
```diff
@@ -20,3 +20,32 @@ lib_deps =
 debug_tool = jlink
 ; If not set we will default to uploading over serial (first it forces bootloader entry by talking 1200bps to cdcacm)
 ;upload_protocol = jlink
+
+[env:rak4631_eink_dynamic]
+extends = nrf52840_base
+board = wiscore_rak4631
+build_flags = ${nrf52840_base.build_flags} -Ivariants/rak4631_epaper -D RAK_4631_D
```
caveman99, line 27: If i recall correctly, the RAK_4631 macro is used in a few places in the firmware. Maybe define the RAK_4631_D macro in addition to RAK_4631

markbirss: yes, for a dynamicEink variant

src/graphics/EInkDynamicDisplay.cpp
markbirss: yes, testing dynamicUpdate varaint of the RAK e-paper

markbirss: yes, it intentional, just skip for the RAK_4631_D

markbirss: RAK_4361_D is a added to platformio as a seperate variant wont affect RAK_4361

todd-herbert: I don't understand. The code you have added here will not impact RAK_4631_D. It will impact all variants *expect* RAK_4641_D.

markbirss: > I don't understand. The code you have added here will not change anything for RAK_4631_D. It will impact all variants _expect_ RAK_4641_D.

Ok, i see, i actually wanted to test without it. should have been !defined(RAK_4631_D)

todd-herbert: Line 316 (`refresh = SKIPPED`) is existing code. You have wrapped it with `#if defined(RAK_4631_D)`.

All the EInkDynamicDisplay code is guarded with `#if defined(RAK_4631_D)`. You do not need to guard this specially, it will not impact the normal RAK_4631.

The guard you have added here will break any other devices which for whatever reason are still using EInkDynamicDisplay.

markbirss: > Line 316 (`refresh = SKIPPED`) is existing code. You have wrapped it with `#if defined(RAK_4631_D)`.
> 
> All the EInkDynamicDisplay code is guarded with `#if defined(RAK_4631_D)`. You do not need to guard this specially, it will not impact the normal RAK_4631 e-ink variant.
> 
> The guard you have added here will break any other devices which for whatever reason are still using EInkDynamicDisplay.


I updated it.

markbirss: > > Ok, i see, i actually wanted to test without it. should have been !defined(RAK_4631_D)
> 
> Oh, that would mean a _lot_ of redundant display updates.
> 
> > I get stuck with input just typing 6 or 7 characters
> 
> Maybe something about the hashing algorithm? I remember it occasionally throwing up false positives for duplicate frames. I think some new code got strapped on recently which enters a special mode if a user is performing text entry. It might be easy to make use of that, to temporarily suppress the "duplicate frame" check? Only a suggestion though.

It a good suggestion!  Just need assistance with implementation

todd-herbert: I think the torch has been passed a bit in terms of maintaining this class, so I'm not particularly familiar with these code changes, but glancing at it now, it seems like [CannedMessageModule.cpp](https://github.com/meshtastic/firmware/blob/6429eca5e47aa1d438012a9886632a5cdbbec898/src/modules/CannedMessageModule.cpp#L1058-L1063) calls [`enableUnlimitedFastMode()`](https://github.com/meshtastic/firmware/blob/6429eca5e47aa1d438012a9886632a5cdbbec898/src/graphics/EInkDynamicDisplay.h#L27), which sets a new frame flag `UNLIMITED_FAST`.

I imagine you might be able to edit [`EInkDynamicDisplay::checkFrameMatchesPrevious()`](https://github.com/meshtastic/firmware/blob/6429eca5e47aa1d438012a9886632a5cdbbec898/src/graphics/EInkDynamicDisplay.cpp#L294-L302) to return early if this frame flag has been set.

```cpp
void EInkDynamicDisplay::checkFrameMatchesPrevious()
{
    // If a decision was already reached, don't run the check
    if (refresh != UNSPECIFIED)
        return;

    // If text-entry in progress, don't run the check
    if (refresh & UNLIMITED_FAST)
        return;

    // If frame is *not* a duplicate, abort the check
    if (imageHash != previousImageHash)
        return;
 ```

markbirss: Understood, yes, your assistance is appreciated. Will see.

markbirss: @todd-herbert Ok, i removed the !defined(RAK_4631_D) in order to address it after as a bug instead.

PR #6318, opened 2025-03-14
update  seeed-xiao-nrf52840-kit  board defination

Dylanliacc:
- Due to the lack of pins, we have temporarily removed the button. There are some technical solutions that can solve this problem, and we are currently exploring and researching them


vidplace7:
@Dylanliacc Can you please run `trunk fmt` on this contribution? Or allow edits from project maintainers.

PR #6327, opened 2025-03-15
RP2xx0: Add UDP Multicast support

vidplace7:
Adds support for UDP Multicast to RP2040 and RP2350 variants which have wifi onboard.

[arduino-pico](https://github.com/earlephilhower/arduino-pico/tree/master/libraries) includes AsyncUDP support, seems it's just slightly different than the ESP32 implementation (ultimately based on the same upstream work)

Depends on:
- #6333

DRAFT as I do not have a means to test it yet (need a pico lora hat). Compiles correctly at least :+1:


caveman99:
Can't you run it without a hat? the sending runs into error -4 but the UDP packet should go out regarless.

PR #6343, opened 2025-03-17
raise the multicast UDP TTL limit

Jorropo:
Since 96ba94843b25f787148e3353a7476ed175518b3b we don't spray packets to all machines on the network.

So we can allow ourself to raise the TTL limit, this allows users who run L3 IGMP Routing infrastructure to pass meshtastic frames over UDP.


thebentern:
Already upstream via another PR

PR #6344, opened 2025-03-18
Integration of the Fisheye State Routing Algorithm

JannisIThink:
pull request for adding the Fisheye State Routing (FSR) Algorithm, which offers greater efficiency compared to the currently implemented Flooding Algorithm. The adoption of the new algorithm is entirely optional, allowing users to continue using the existing Flooding Algorithm by configuring the settings accordingly. The implementation of the routing algorithm is primarily structured as a dedicated module, ensuring modularity and ease of integration.


src/mesh/generated/meshtastic/admin.pb.h
```diff
@@ -67,7 +67,9 @@ typedef enum _meshtastic_AdminMessage_ModuleConfigType {
     /* TODO: REPLACE */
     meshtastic_AdminMessage_ModuleConfigType_DETECTIONSENSOR_CONFIG = 11,
     /* TODO: REPLACE */
-    meshtastic_AdminMessage_ModuleConfigType_PAXCOUNTER_CONFIG = 12
+    meshtastic_AdminMessage_ModuleConfigType_PAXCOUNTER_CONFIG = 12,
+    /* TODO: REPLACE */
+    meshtastic_AdminMessage_ModuleConfigType_FISHEYESTATEROUTING_CONFIG = 13
 } meshtastic_AdminMessage_ModuleConfigType;
```
caveman99, line 73: All files in 'generated' are overwritten automatically. plrease remove those changes and submit a change to the protobuf repository instead

JannisIThink:
Hello, sorry for me responding that late. This was due to some time constraints on my side. However, it was brought to my attention that my initial description of the PR might have been too short and did not cover the details and the goal I was aiming for with the PR. Thus, I am writing this text to give some more information on the code and clarify its intentions. The contents of this PR originated from a project at my university. It does not aim for providing a ready-to-ship solution but focuses more on providing a kind of prototype to showcase the concept and sense its acceptance and feedback from others. Therefore, there might be some errors or breaches of conventions still remaining in the code (although I gave my best to track all of them down and fix them). The goal of the project was to incorporate a proper routing algorithm into the project. I went for the proactive Fisheye State Routing Algorithm, because it greatly improves the efficiency of the Link State Routing Algorithm and reduces the total number of send packages (which saves bandwidth). The key difference to LSR is the delayed forwarding of Link State Packages. For every incoming Link State Package, a timeout is calculated depending on the distance of the receiving node from the node the message originated from. Only after the timeout expired, the node is allowed to forward the package. This leads to very accurate information on near nodes and only vague information on distant nodes. If the timeout is interrupted by a newer Link State Package from the same node, the old one is deleted. The timeout of the new one is the minimum of the old one and the newly calculated one. This ensures that every package is forwarded eventually.
I opted to implement the FSR-Algorithm as a module. The Link State Packages are represented by the struct meshtastic_FishEyeStateRouting, which contains a neighbor array, the neighbor count, the timestamp of the creation of the Link State Package (to only replace an existing one with a newer version and not with an older one), and the nodeID of the node issuing the Link State Package. The own neighborhood is provided by the neighborInfoModule. The FSR-Module registers with the ProtobufModule to be able to receive the before-mentioned structs via the mesh. The function handleReceivedProtobuf saves the received Link State Packages in a hash table if they were created more recent than the last information received from this specific node. Also, the timeout is calculated here. Run Once handles the forwarding of packages whose timeouts expired and sends our own Link State Package into the Mesh every now and then. It is called every 300 seconds (this seemed reasonable to me) or even earlier if a timeout is known to expire earlier. If changes in our own neighborhood or in the neighborhood of other nodes (through a new Link State Package) are detected, calacNextHop is called to recalculate the Next-Hops. calcNextHop calculates the Next-Hops for all known and currently reachable nodes and saves them in the Next-Hop table. To accomplish this goal, it uses a simplified version of the Dijkstra-Algorithm.
If sendToMesh is called, the FishEyeStateRoutingModule checks whether it has a Next-Hop ready for the destination of the package. If so it is written in the decoded.destination field in the message struct. (Thanks to FUVWAF for pointing out the misusage of the to-Field). 
The Router class was modified so that for every incoming message it is checked if it has a Next-Hop written in the decoded.destination field, which is neither 0 or NodeNumBroadcast nor our own NodeID. If so, the package is forwarded to the Next-Hop (provided again by the FESRModule). 
I tested the module with some topologies in Meshtasticator and to me, it seems as if the code is capable of computing a proper Next-Hop table and delivering packages correctly. 
The code allows for the Fisheye Routing to be disabled, therefore being completely optional to use. The collaboration with already existing firmware is not possible in its current form but could be fixed by the following adjustments: always set the to field to broadcast. Write the final destination it the decoded dest field and the Next-Hop in the Next-Hop field. By letting handleReceived in Router check whether those requirements are met would allow for Fisheye State Routing to be done, but would also result in the message being broadcasted (and therefore correctly delivered) by old firmware. (Thanks again to GUVWAF for bringing the compatibility thing to my mind.).
I would be grateful for further feedback on the approach, the concrete execution, or my code in general (bc I’m a CS student aiming to improve my skills 😊)

caveman99:
Thanks for the thorough explanation. I get the concept, but i still fail to see how this is substantially different from the existing NextHop Routing algorithm already in the firmware for direct messages. Group Messages still need to use the intelligent flooding approach under your proposal.

Also as i pointed out, past experience has shown that the neighbourinfo module which you rely on has a really bad impact on the network utilization, if you run it at a frequency you'd need for dynamic networks. 

When considering routing in Meshtastic, we need to keep in mind, that this is a moving network, not fixed base stations. so you can't determine a routing info and keep using it, you have to constantly recalculate.

I'd really put up a challenge to use meshtasticator to prove your approach is substantially better than the one chosen at the moment with 2.6,

PR #6351, opened 2025-03-20
Added bounds checking to memcpy and use memory-safe strlcpy

raulperdomo:
for reading serial data in processWXSerial() function.

Now memcpy should be limited to 229 bytes even on bad input. This leaves one byte for the null termination. Secondly, I changed all strcpy to strlcpy to prevent overruns.


fifieldt:
Hi, and welcome to Meshtastic!

Are you familiar with the trunk tool? We use it for linting.

raulperdomo:
I haven't run trunk before. I am trying to get it installed now and will fix the commit

caveman99 (approved):

caveman99 (approved):

Issue #2409, opened 2023-04-03
[Feature Request]: Remote Updates via IP


Noki:
### Platform

ESP32

### Description

We installed a node at the top of the tower of a church, which has excellent area coverage, but is really hard to reach for updates. The node is connected to a wireless network so the web interface of the ESP32 is reachable via remote. It would be really helpful to enable remote updates directly via IP. Currently the only solution is to attach a computer running 24/7 only to allowing remote updates. We want to eliminate that computer because it is quite power hungry compared to the ESP32.

The update feature could probably be implemented in the same way it is done for OpenMQTTGateway 1.5.0+.

caveman99:
OTA updates through wifi can not be added due to size constraints, sorry. we spport updating through bluetooth in the future.

Noki:
@caveman99 Can you please explain the size constraints in more detail? I agree that some modules might not be able to support this feature, but different modules have different amounts of flash so at least products like the Wireless Stick Lite V3 should be capable of handling OTA updates via WiFi.

caveman99:
We currently have about 650k available for the OTA firmware, which is used up to 93%. Adding Wifi OTA capability to the existing firmware will run into this size constraint. If someone could come up with an implemntation to replace (not add to) the OTAfirmware with a Wifi OTA Version, you could select either/or by replacing the OTA firmware on the board.

caveman99:
A lot of ideas and pointers for this are in https://github.com/meshtastic/firmware-ota - the new firmware will need to support ESP32 and ESP32S3 boards.

ajmcquilkin:
Not sure if this is in scope of this ticket, but would exposing a radio via PoE make any of the challenges of this PR easier? I'm working with a user with a similar use case but instead of using an ESP32 is using a RAK connected via Ethernet

caveman99:
The Updates for RAK boards (nRF52) are even less likely to work, because DFU updates are handled by the bootloader and it doesn't know about IP at all.

ajmcquilkin:
Ah yeah that's what I figured, figured it was worth a shot. Thanks for the context!

goatzillax:
Could this be enabled on devices with sdcards by just using a web server to upload to the sdcard and then initiate the update from there?

KyleMaas:
This would be really, really useful for situations where a node is deployed in a remote area.

PR #6352, opened 2025-03-20
Support WiFi OTA

mskvortsov:
This change makes use of https://github.com/mskvortsov/firmware-ota-wifi

Besides basic features, to prevent device loss, it recovers network settings after config reset (like after a major version update).

Some testing has been done only for esp32 and esp32c3 targets, in theory should also work for c2/c6/s2/s3/p4.


mskvortsov:
https://github.com/meshtastic/firmware/issues/2409 seems to be relevant

src/platform/esp32/WiFiOTA.cpp
```diff
@@ -0,0 +1,86 @@
...
+        return false;
+    return true;
+}
+
+String getVersion()
+{
+    String version;
+    esp_app_desc_t app_desc;
+    const esp_partition_t *part = esp_partition_find_first(ESP_PARTITION_TYPE_APP, ESP_PARTITION_SUBTYPE_APP_OTA_1, NULL);
+    if (!part)
```
caveman99, line 78: You are looking at OTA_1 - the BLEOTA firmware sits in OTA_0. Wouldn't it be better to use the same partition scheme in both cases? BLE checks for the application name in addition, so if you check for BLE first, if that comes back as false it must be the Wifi version.

src/platform/esp32/main-esp32.cpp
```diff
@@ -139,12 +141,19 @@ void esp32Setup()
...
-#else
-    LOG_INFO("No OTA firmware available");
+#endif
+#if !MESHTASTIC_EXCLUDE_WIFI
+    String version = WiFiOTA::getVersion();
+    if (version.isEmpty()) {
+        LOG_INFO("No WiFi OTA firmware available");
+    } else {
+        LOG_INFO("WiFi OTA firmware version %s", version.c_str());
+    }
```
caveman99, line 155: wouldn't that print that a wifi AND ble updater was available, since the wifi check only looks for the version number? BLEOTA has a version number too :-)

caveman99 (changes requested):
Nice work! See two questions inline.

src/platform/esp32/WiFiOTA.cpp
caveman99: Sorry, OTA_1 is correct, that was my bad.

TheMalkavien:
Hi, Just tested with esp32s3, it works ! Thanks for this feature.
The full procedure (compilation, esptool, espota) is a little bit of tinkering, but it sure will be simplified in the future.

Thx again.

caveman99:
@mskvortsov would you be willing to move https://github.com/mskvortsov/firmware-ota-wifi to the meshtastic org namespace? This way it's easier to add it to our build process. You'll still have full access on it :-)

Crusher-Glockman34:
Sorry for my ignorance but which file should I push to my devices and how? I just downloaded 2.6.4.b89355f Beta and extracted it but I don't see any file ending in -ota.zip referenced by other docs I've found online. I'm using all rak4631 devices and I really want to deploy my router on the roof but not until I can do OTA updates. I've previously done this one time using the Android app nRF Mesh Flash OTA. Is this the correct method and if not what's supported? Thanks

Dude5101:
> Hi, Just tested with esp32s3, it works ! Thanks for this feature. The full procedure (compilation, esptool, espota) is a little bit of tinkering, but it sure will be simplified in the future.
> 
> Thx again.

Hello!

Please will you provide more detail on how you got this to work.

I converted the firmware .elf file to .bin, tell the node to go to OTA mode and try to upload using espota.py but I get no response error.

C:\Users\fhd08>python C:\Meshtastic\tools\espota.py --ip 192.168.68.52 --file C:\Meshtastic\firmware\firmware-heltec-wsl-v3-2.7.5.bin
Sending invitation to 192.168.68.52 ..........
15:34:28 [ERROR]: No response from the ESP

slomkowski:
> And one thing I noticed is that `meshtastic --reboot-ota` doesn't always work

@mskvortsov I'm facing the same issue, I cannot restart with `--reboot-ota`. I have the same messages as you. Did you perhaps find the cause of this?

```
$ meshtastic --host 10.8.1.106 --version
2.7.8
```

```
$ meshtastic --host 10.8.1.106 --reboot-ota
Connected to radio
INFO file:node.py rebootOTA line:661 Telling node to reboot to OTA in 10 seconds
```

Serial logs:
```
INFO  | 00:38:34 50695 [ApiServer] Force close previous TCP connection
DEBUG | 00:38:34 50695 [ApiServer] PhoneAPI::close()
INFO  | 00:38:34 50695 [ApiServer] Incoming API connection
INFO  | 00:38:34 50695 [ApiServer] Incoming wifi connection
INFO  | 00:38:34 50695 [ServerAPI] Client wants config, nonce=1194018866
DEBUG | 00:38:34 50695 [ServerAPI] Got 6 files in manifest
```

Edit: I found that with firmware version ≥ 2.7.18, the `--ota-update` command should be used; nonetheless, the restart does not occur.

PR #6358, opened 2025-03-21
Remove unnecessary null pointer checks

fifieldt:
As reported by @elfring, we had several points in our code where it was unnecessary to check pointers were non-null before deleting them.

Fixes https://github.com/meshtastic/firmware/issues/6170

### DIFF
--- a/src/AudioThread.h
+++ b/src/AudioThread.h
@@ -41,10 +41,8 @@ class AudioThread : public concurrency::OSThread
             delete i2sRtttl;
             i2sRtttl = nullptr;
         }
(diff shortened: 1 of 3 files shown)


elfring:
:thought_balloon: [Would you find the implementation of a function like “CryptoEngine::encryptAESCtr” still improvable](https://github.com/meshtastic/firmware/blob/0951fdd49b67c00c3c075b4236b12af7fb04d507/src/mesh/CryptoEngine.cpp#L224-L229 "Update candidate")?

PR #6371, opened 2025-03-22
[esp32] Define BUTTON_PIN (-1) by default, fixes #6213

### LINKED ISSUE #6213: [Bug]: GPIO incorrect for Lilygo Lora32 2.1 1.6.1

ndoo:
ESP32 - Define BUTTON_PIN (-1) by default, fixes #6213 because button setup is always done in https://github.com/meshtastic/firmware/blob/1e4a0134e6ed6d455e54cd21f64232389280781b/src/main.cpp#L375-L392 and https://github.com/meshtastic/firmware/blob/1e4a0134e6ed6d455e54cd21f64232389280781b/src/ButtonThread.cpp#L31-L58 when the user preference for button pin is configured.

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -110,6 +110,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 // Define if screen should be mirrored left to right
 // #define SCREEN_MIRROR
 
+// Define BUTTON_PIN to ensure button setup is always done
+#ifndef BUTTON_PIN
+#define BUTTON_PIN (-1)
+#endif
+
 // I2C Keyboards (M5Stack, RAK14004, T-Deck)
 #define CARDKB_ADDR 0x5F
 #define TDECK_KB_ADDR 0x55


src/platform/esp32/architecture.h
```diff
@@ -53,6 +53,8 @@
 #ifdef PIN_BUTTON_TOUCH
 #define BUTTON_PIN_TOUCH PIN_BUTTON_TOUCH
 #endif
+
+#define BUTTON_PIN (-1)
 //
```
caveman99, line 58: This woud belong in configuration.h and urgently with a check if it is not already set.

caveman99 (changes requested):

src/platform/esp32/architecture.h
ndoo: Moved, rebased commit on top of master HEAD and force-pushed my branch.

ndoo:
I only compile-tested the above change, would appreciate if anyone with access to a T-LoRa V2.1 1.6 could try it out, I will functionality-test it tomorrow otherwise (my hardware is in the office).

caveman99 (approved):

ndoo:
Hmm, is there some reason this isn’t being done in the variant files for
nrf52? Or if it is, can this be safely gated behind an #ifndef (or is it
even needed?)

PR #6372, opened 2025-03-22
Fix NodeInfo overwriting publicKey in NodeDB

dfsx1:
An attacker can send NodeInfo with a empty publicKey first, then overwrite with a new key.

https://github.com/meshtastic/firmware/blob/1e4a0134e6ed6d455e54cd21f64232389280781b/src/mesh/NodeDB.cpp#L1438-L1447

First sending a empty key bypasses `if (p.public_key.size > 0) {`, 
clearing the existing publicKey (and resetting the size to 0) for a known node.

Then a new key bypasses `if (info->user.public_key.size > 0) {`, 
and this mallicious key is stored in NodeDB.

https://github.com/meshtastic/firmware/blob/1e4a0134e6ed6d455e54cd21f64232389280781b/src/mesh/NodeDB.cpp#L1451
https://github.com/meshtastic/firmware/blob/1e4a0134e6ed6d455e54cd21f64232389280781b/src/mesh/TypeConversions.cpp#L87-L88

I tested my commit with a few nodes and this seems to succesfully prevent the exploit.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -1437,13 +1437,14 @@ bool NodeDB::updateUser(uint32_t nodeId, meshtastic_User &p, uint8_t channelInde
 #if !(MESHTASTIC_EXCLUDE_PKI)
     if (p.public_key.size > 0) {
         printBytes("Incoming Pubkey: ", p.public_key.bytes, 32);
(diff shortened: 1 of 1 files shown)


caveman99 (approved):

caveman99:
Thanks for this report. I'd appreciated it if you used the security reporting mechanism instead of posting a public pull request.

PR #6391, opened 2025-03-24
linux-native: allow multiple processes to all bind to the same multicast 2tuple

Jorropo:
Depends on https://github.com/meshtastic/platform-native/pull/9


vidplace7:
Actual change behind the layers https://github.com/meshtastic/framework-portduino/pull/52

PR #6422, opened 2025-03-27
TCA8418 initial config + basic 3x4 keypad config

Nasimovy:
pull request for adding the [TCA8418](https://www.ti.com/lit/ds/symlink/tca8418.pdf), the config now is for a 3x4 keymatrix (multitap) with (backlight)

the issue i see at this moment is that the TCA8418 supports up to 80 buttons, the config for the 3x4 keypad uses 12 buttons and 1 pin as an output for the backlight.

Is there a good way on how to select a different button mapping?

~the TCA8418 does not have a identification register and neither has the AXP192.~
The method i use now needs testing on the t-beam ~is that allowed?~

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [X] RAK WisBlock 4631 (by Nasimovy)
  - [X] Heltec Vision Master T190 (by Nasimovy)


Mictronics:
Can't pull the master into my dev_nokia5130 branch without completely messing up the code. :disappointed:

Mictronics:
> the TCA8418 does not have a identification register and neither has the AXP192.
But the APX192 has a way more registers than the TCA8418.
Not tested, but the TCA8418 may return a NACK when you set and read an out of range register. Whereas the AXP192 should ACK and return a value.

Nasimovy:
if i understand the code the proper way it does not check ACK or NACK
the code at this moment does not always search for the AXP192 and the AXP2101 only if  HAS_PMU is defined,
so that is why i am asking

Mictronics:
> if i understand the code the proper way it does not check ACK or NACK

Check how it's done for ARCH_PORTDUINO platform in ScanI2CTwoWire.cpp.

Nasimovy:
> Check how it's done for ARCH_PORTDUINO platform in ScanI2CTwoWire.cpp.

i tried this but the TCA8418 always seems to give an ACK no matter the register
so instead i try to read out register 90

Nasimovy:
oh, my t-beam should be here tomorrow, i will also test and report back, i have the one with the AXP2101

Nasimovy:
it works on the T-Beam with the APX2101
`INFO | ??:??:?? 1 AXP192/AXP2101 found at address 0x34`

Nasimovy:
@Szetya the t-lora pager also uses the TCA-8418, it needs shift key capabilities, slower multitap would be a loss on usability in my opinion

Szetya:
@Nasimovy Do you think a flag would be a good solution for the TCA8418TapMap second index setting?

Nasimovy:
Flag would work, did you make a new tapmap for the tdeck pro?

Szetya:
I moved the registers to the header file. :)
Yes, I have done it, or it's in progress.
[TCA8418Keyboard.zip](https://github.com/user-attachments/files/20413926/TCA8418Keyboard.zip)

Szetya:
do you have t-deck pro?

Nasimovy:
no, i dont have a tdeck pro, i want to buy the t lora pager, i dont really like the e-paper, the flag would stay the same however

Szetya:
I'm waiting for the pager too but it's hard to catch. Anyway I want to use LR1121 radio and I don't know if I will get 433/2.4 or if I will have to buy 868/2.4. There are more options in 433.
I'm going to use the flag. If you have any suggestions I'd be happy to try it.
I had a problem with BQ25896, its i2c cime is 0x68. I found this solution:
ScanI2CTwoWire.cpp
```cpp
            case QMI8658_ADDR:
                // First the BQ25896, if this would affect other devices it will be moved to the else branch below
                registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0x14), 1); // read REG14
                // bit 5,4,3: Device Configuration bits = 000: bq25896 (revision isnformation bits 1,0 = 10 , maybe later)
                if ((registerValue & 0b00111000) == 0b00000000) {
                    type = BQ25896;
                    logFoundDevice("BQ25896", (uint8_t)addr.address);
                    break;
                }
                registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0x0A), 1); // get ID
                if (registerValue == 0xC0) {
                    type = BQ24295;
                    logFoundDevice("BQ24295", (uint8_t)addr.address);
                    break;
                }
                registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0x0F), 1); // get ID
                if (registerValue == 0x6A) {
                    type = LSM6DS3;
                    logFoundDevice("LSM6DS3", (uint8_t)addr.address);
                } else {
                    type = QMI8658;
                    logFoundDevice("QMI8658", (uint8_t)addr.address);
                }
                break;
```

Nasimovy:
do you have a fork or something where i can look at your code?

Szetya:
I'm not a good coder, I don't use to use git, fork. I don't know how to use them either. I only have local files. I tried PR once but I didn't know what I was doing. :)

mverch67:
@Nasimovy 
As there are now three devices using the TCA8418 I've refactored your class into a generic base class and three derived classes that overload the virtual methods (pressed, released) of the base class for individual keymap processing. Also changed the visibility of the class methods accordingly to provide a clear interface to the two user groups.
Will commit the changes into #6936.

Issue #6427, opened 2025-03-28
[Feature Request]: Reduce memory footprint of nodes


jasonbcox:
### Platform

Cross-Platform

### Description

NodeInfoLite could use a bitfield to compress booleans from 3 bytes to 1. I created this issue to track some PRs along that effort and to document any other low hanging fruit that could reduce memory footprint of nodes stored in the router.

If this issue leads to successful PRs, it will help in the efforts for #6426 by freeing up room for a new cache.

Please assign this task to me. I have already started working on it.

PR #6428, opened 2025-03-28
Add a static_assert to verify assumption about NodeInfoLite size

jasonbcox:
This is my first PR. Adding this nice-to-have for multiple reasons:
1. It only affects compile time, so cost on device is free.
2. It highlights a baseline assumption that MAX_NUM_NODES is dependent on. Reductions in this assertion boundary could encourage reevaluation of memory limits such as MAX_NUM_NODES.
3. Changes to NodeInfoLite, especially those that increase its memory footprint, will force PRs to highlight this fact by modifying the value. This should be useful during code review.
4. This is a baby step toward implementing #6427. It's not strictly necessary, but a good way to show off the impact.

### DIFF
--- a/src/mesh/mesh-pb-constants.h
+++ b/src/mesh/mesh-pb-constants.h
@@ -18,6 +18,10 @@
 #define MAX_RX_TOPHONE 32
 #endif
 
+/// Verify baseline assumption of node size. If it increases, we need to reevaluate
+/// the impact of its memory footprint, notably on MAX_NUM_NODES.
+static_assert(sizeof(meshtastic_NodeInfoLite) <= 192, "NodeInfoLite size increased. Reconsider impact on MAX_NUM_NODES.");
+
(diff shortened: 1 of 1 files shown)


src/mesh/mesh-pb-constants.h
```diff
@@ -18,6 +18,9 @@
 #define MAX_RX_TOPHONE 32
 #endif
 
+/// Verify baseline assumption of node size
+static_assert(sizeof(meshtastic_NodeInfoLite) <= 192, "NodeInfoLite size is larger than expected");
```
GUVWAF, line 22: Shall we make the comment a bit more clear? Why is it expected to be 192? Maybe something like "NodeInfoLite size increased, reconsider `MAX_NUM_NODES`".

GUVWAF (approved):

caveman99 (approved):

PR #6429, opened 2025-03-28
Refactor Data bitfield into an enum

jasonbcox:
This is a little easier to read and will serve as a good example for use with other (larger) bitfields in the future. Another baby step towards #6427

A longer range benefit of using a enum is that the valid values for a field are typed. It's not really relevant for this change, but becomes relevant if/when you type the flags explicitly. This helps avoid mismatching flags and their associated bitfields.

I have not tested this on devices yet, but it does build. There should be no change in behavior.

### DIFF
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -417,8 +417,9 @@ DecodeState perhapsDecode(meshtastic_MeshPacket *p)
     if (decrypted) {
         // parsing was successful
         p->channel = chIndex; // change to store the index instead of the hash
-        if (p->decoded.has_bitfield)
-            p->decoded.want_response |= p->decoded.bitfield & BITFIELD_WANT_RESPONSE_MASK;
+        if (p->decoded.has_bitfield) {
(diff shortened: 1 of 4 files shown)


GUVWAF:
Thanks for the PR. However, I'm not sure this is a better approach. You need another if-statement to set the bool, and what if we add another value that takes up 2 bits? We currently use the same approach for the header flags, which has two 3-bit values: 
https://github.com/meshtastic/firmware/blob/1e41c994b3ec9395c1c9fb2aae25947ec6306060/src/mesh/RadioLibInterface.cpp#L451

GUVWAF:
Since it’s now almost similar to what it was before, I’m still not convinced this approach is better as we’d still need the old method for multi-bit values.
Anyway, if others agree your method is better I’m fine with it :)

caveman99 (commented):
I am impartial to the change, but don't think it's neccessarily easier to read or give us other benefits. mY gut feeling is 'if it works, don't change it, unless there's substrantial profit

PR #6434, opened 2025-03-28
Possible fix for use of pin 0 as PIN_GPS_EN

macvenez:
As discussed on Discord [here](https://discord.com/channels/867578229534359593/919642584480112750/1355219959726346313)
we are using PIN 0.00 on NRF52 for PIN_GPS_EN this causes an issue because if it's set to 0 this skips all the code that creates the gps enable pin instance.

I thought about this for a possible solution, this is working correctly on our board.
Will this be ok?
Thanks!

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -1370,7 +1370,7 @@ GPS *GPS::createGps()
 
     GpioVirtPin *virtPin = new GpioVirtPin();
     new_gps->enablePin = virtPin; // Always at least populate a virtual pin
-    if (_en_gpio) {
+    #if defined(PIN_GPS_EN)
         GpioPin *p = new GpioHwPin(_en_gpio);
 
         if (!GPS_EN_ACTIVE) { // Need to invert the pin before hardware
@@ -1382,7 +1382,7 @@ GPS *GPS::createGps()
                 virtPin,
                 p); // We just leave this created object on the heap so it can stay watching virtPin and driving en_gpio
         }
-    }
+    #endif
 
 #ifdef PIN_GPS_PPS
     // pulse per second


ponzano:
Tested and working 👌

src/gps/GPS.cpp
```diff
@@ -1369,7 +1369,7 @@ GPS *GPS::createGps()
 
     GpioVirtPin *virtPin = new GpioVirtPin();
     new_gps->enablePin = virtPin; // Always at least populate a virtual pin
-    if (_en_gpio) {
+    #if defined(PIN_GPS_EN)
```
thebentern, line 1372: Won't this skip for runtime configured pins with `config.position.gps_en_gpio`?

macvenez: Yes, that's possible, do you think we can still go with this approach or we should switch to using our device specific #define as did for `GNSS_AIROHA` in the same file?
I understand there may be devices wich use GPS but don't have EN pin connected for GPS so we must use value 0 to identify those? (The issue is that with our device 0 is the correct gpio value instead...)

ponzano: > Won't this skip for runtime configured pins with `config.position.gps_en_gpio`?

Perhaps we could do this (?)

if(_en_gpio){
 >code to execute 
}
else{
#if defined(PIN_GPS_EN)
 >same code to execute
#endif

fifieldt: Check out line 1355

fifieldt: I suggest trying to modify that area rather than changing the branching here. Try and make it a generic solution rather than one specific to a board, the airoha stuff is ... not the best :)

macvenez: Alright, I'm trying to figure it out.
The issue is this probably won't support setting PIN_GPS_EN to pin 0 by app or CLI as we need a value for boards which don't have this pin. Am I right?
Thanks

macvenez: In alternative, but I don't know if that's feasible, we could use -1 as value for boards which don't use Enable pin for GPS so we could easily distinguish between zero or positive values (correct pin to use) and -1 which would mean don't use

PR #6440, opened 2025-03-29
update to handle ws80 serial data as well

tavdog:
Small change to make the string parsing of Name = value less brittle.  Adds a function to parse a line without knowing how many spaces are after the = sign.  This allows it to also work with the ws80 serial output.

## 🤝 Attestations
- [*] I have tested that my proposed changes behave as described.
- [*] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [*] RAK WisBlock 4631

### DIFF
--- a/src/modules/SerialModule.cpp
+++ b/src/modules/SerialModule.cpp
@@ -408,6 +408,49 @@ uint32_t SerialModule::getBaudRate()
     return BAUD;
 }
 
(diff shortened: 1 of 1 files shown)


caveman99 (approved):

thebentern:
Nice!

PR #6446, opened 2025-03-29
meshtasticd: CH341 / HAT+ Auto Configuration

vidplace7:
## MeshtasticD AutoConf

Adds support for automatically loading a config for a given LoRa radio via two methods:
- CH341`PIDS` "Product String"
    - https://www.wch-ic.com/downloads/CH341DS1_PDF.html
- Raspberry Pi HAT+ "Product" field
    - https://datasheets.raspberrypi.com/hat/hat-plus-specification.pdf
    - https://github.com/raspberrypi/hats/blob/master/devicetree-guide.md
    - https://github.com/raspberrypi/utils/blob/master/eeptools/eeprom_settings.txt

Both of these methods require a programmed *EEPROM* flash present on the device.

Also adds the `--sim` arg, which forces meshtasticd into Simulator mode (ignoring all other configs).

image


src/platform/portduino/PortduinoGlue.cpp
```diff
@@ -197,6 +197,53 @@ void portduinoSetup()
             }
         }
     }
+
+    // If a LoRa radio is not configured,
+    // attempt to auto config based on Product Strings
+    if (settingsMap[has_lora] != true) {
```
jp-bennett, line 203: Want to be careful here, not to break the simulator use case. What if instead, we use a Module: auto to put the daemon in autoconfig mode.

vidplace7: > Want to be careful here, not to break the simulator use case. What if instead, we use a Module: auto to put the daemon in autoconfig mode.

I preserved that use-case here 👍If no CH341 or HAT+ is found, meshtasticd will continue to start in simulator mode.
If a CH341 / HAT+ is found but it cannot be resolved to a config, meshtasticd will intentionally halt (rather than start SIM mode).

jp-bennett: Yeah, I'm just imagining a developer trying to do something that does an automatic simulator run, and also happens to have a CH341 plugged in to the machine. But if nobody complains, I won't either.

vidplace7: > What if instead, we use a Module: auto to put the daemon in autoconfig mode.

The goal with this PR is to enable "plug and play" meshtasticd without *ever* touching a config. Changing a config to place meshtasticd in autoconfig mode is counter to that that goal. I could see doing the reverse to *force* SIM mode? Something like `Module: SIM`.

> I'm just imagining a developer trying to do something that does an automatic simulator run, and also happens to have a CH341 plugged in to the machine.

Yeah that situation wouldn't be supported under my proposed PR. Exposing a way to force simulator mode would be my preferred solution here for sure. Perhaps we could just add an arg like `--sim`? (or the aforementioned `Module: SIM")

jp-bennett: > > What if instead, we use a Module: auto to put the daemon in autoconfig mode.
> 
> The goal with this PR is to enable "plug and play" meshtasticd without _ever_ touching a config. Changing a config to place meshtasticd in autoconfig mode is counter to that that goal. I could see doing the reverse to _force_ SIM mode? Something like `Module: SIM`.
> 
> > I'm just imagining a developer trying to do something that does an automatic simulator run, and also happens to have a CH341 plugged in to the machine.
> 
> Yeah that situation wouldn't be supported under my proposed PR. Exposing a way to force simulator mode would be my preferred solution here for sure. Perhaps we could just add an arg like `--sim`? (or the aforementioned `Module: SIM")

I like either the --sim flag, or modifying the default config in our rpm/deb/etc builds to include the Module:auto line.

vidplace7: > I like either the --sim flag, or modifying the default config in our rpm/deb/etc builds to include the Module:auto line.

Modified the PR to take the approach of setting `Module: auto` in the default config and only running autoconf when Module == auto.

vidplace7: `--sim` arg has been added. This forces meshtasticd into simulator mode and does not load config files from yaml.
Additionally, `Module: sim` is now available within the config files, to force sim mode.

thebentern (approved):

PR #6462, opened 2025-03-30
Disable network config for non-eth_gateway nrf52 and non-W RP2040 targets

### LINKED ISSUE #6448: [Feature Request]: Disable Network Module on all NRF boards except for the Rak 4631
### Platform

NRF52

### Description

Network functionality is only available for the RAK 4631 NRF board

thebentern:
Closes #6448


lib/device-ui
vidplace7, line 1: 🧟 return of the submodule!

fifieldt (approved):

garthvh (approved):
Looks great

src/main.cpp
```diff
@@ -1274,6 +1274,12 @@ extern meshtastic_DeviceMetadata getDeviceMetadata()
     deviceMetadata.excluded_modules |= meshtastic_ExcludedModules_AMBIENTLIGHTING_CONFIG;
 #endif
 
+#if defined(ARCH_NRF52) && !defined(NRF52_USE_JSON) // nrf52 doesn't have network unless it's a RAK ethernet gateway currently
```
caveman99, line 1277: Right now the RAK ethernet module/mode is active in these variants: monteops_hw1, rak4631, rak4631_eth_gw, rak11310 and rak_wismeshtap

I think we can drop it from wismeshtap and rak4631 proper (which would be a separate PR and release notes), but the others have to stay. Especially RAK11310 which is the RP2040 core.

src/main.cpp
```diff
@@ -1274,6 +1274,12 @@ extern meshtastic_DeviceMetadata getDeviceMetadata()
...
+    deviceMetadata.excluded_modules |= meshtastic_ExcludedModules_NETWORK_CONFIG; // No network on nRF52
+#elif defined(ARCH_RP2040) && !HAS_WIFI
+    deviceMetadata.excluded_modules |= meshtastic_ExcludedModules_NETWORK_CONFIG; // No network on RP2040
```
caveman99, line 1280: see above. rak11310 can use the ETH module.

caveman99 (approved):

PR #6489, opened 2025-04-03
update OLED library, fix nRF build of SH1107

caveman99:


tropho23:
Works on my nrf52-diy-promicro!

PR #6497, opened 2025-04-05
Fix Ublox GPS for Heltec T114

todd-herbert:
Fixes a crash specific to the T114, which occurs when a UBlox M10 GPS module moves into the `GPS_HARDSLEEP` or `GPS_OFF` states. 

Code which specifically targeted the T114 was ending the serial port, interfering with the serial messages used to put UBlox hardware into low-power states. 

Discord user hckr2000 reports that removing this code resolves the issue. I tested this with the T114's official L76K module and didn't notice any change in function or current draw.

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
     - Heltec Mesh Node T114


fifieldt (approved):

hcker2000:
Can confirm this fixes the issue for me.

PR #6500, opened 2025-04-05
Update web, use centrally defined version

vidplace7:
Update web component from 2.5.3 to 2.6.0 for esp32.

Additionally, rework all projects that consume `web` to leverage a centrally defined release version.
When new versions of `web` are published, only `bin/web.version` must be updated.

This is *much* better for reproducibility in future builds.


vidplace7:
> Error: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')

Not really sure what's going on there... GitHub Actions changes aren't reflected until Merge, so could be related to that?

danditomaso:
@vidplace7 Thanks for taking care of this!

thebentern (approved):

PR #6507, opened 2025-04-06
Fix several features of M1 and M2 (i know what the 7 is now ...)

caveman99:


src/ButtonThread.cpp
```diff
@@ -180,32 +189,27 @@ int32_t ButtonThread::runOnce()
...
-#endif
-#if defined(ARCH_PORTDUINO)
-            if ((settingsMap.count(user) != 0 && settingsMap[user] != RADIOLIB_NC) &&
-                    (settingsMap[user] != moduleConfig.canned_message.inputbroker_pin_press) ||
-                !moduleConfig.canned_message.enabled) {
-                powerFSM.trigger(EVENT_PRESS);
+                break;
             }
+#ifdef ELECROW_ThinkNode_M1
+            sendAdHocPosition();
```
caveman99, line 195: because we had concerns - the button is not easy to press, much unlike on T-Echo

src/ButtonThread.cpp
```diff
@@ -239,17 +239,17 @@ int32_t ButtonThread::runOnce()
                         screen->forceDisplay(true); // Force a new UI frame, then force an EInk update
                 }
                 break;
-#elif defined(ELECROW_ThinkNode_M2)
+#elif defined(ELECROW_ThinkNode_M1) || defined(ELECROW_ThinkNode_M2)
             case 3:
                 LOG_INFO("3 clicks: toggle buzzer");
```
caveman99, line 244: This will eventually be moved into ALERT_APP. Local noise and light for now.

PR #6510, opened 2025-04-07
Lib Update

caveman99:
Draft because PIN display doesn't work yet.


caveman99:
@alexbegoon the fix works, thanks a lot! :-) waiting for Ci to compete, then mergin and folding back into the C6 PR. This can go in next :-)

thebentern (approved):

todd-herbert:
Hey I think this is causing some weird results just now testing with Wireless Paper. From my Android 13 phone, the node can't be seen in either the Meshtastic app or the system Bluetooth settings.

The nRF Connect app does show the node, and allows me to bond with it. After bonding this way, the Meshtastic Android app works as normal.

<details>
<summary>
Data from nRF Connect app
</summary>

Screenshot_20250418_231410_nRF Connect

Screenshot_20250418_231402_nRF Connect

</detail>

cpatulea:
FYI I tried again to make this update.

`NIMBLE_TWO` made this easier, though it still required a few code tweaks, it was not too bad: https://github.com/meshtastic/firmware/commit/a1578b47cc86abd83cbef8a2f0d824e7c00b358f

Testing:

- Heltec v3: built and tested on real device + iPhone
- All devices based on ESP32 original (8 out of 57 ESP32 devices): does *not* build. It increases the static RAM usage so much, the firmware can't link (`/home/eigma/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pio/build/m5stack-coreink/firmware-m5stack-coreink-2.7.20.a1578b4.elf section .dram0.bss will not fit in region dram0_0_seg`). I spent a while trying to solve this but it seems really difficult and I don't have experience with shrinking static RAM usage.
- All other devices (48 devices): confirmed it builds successfully but not tested.

I was doing this to find a way to help `pioarduino` migration move forward. I chatted with @phaseloop on Discord and he said we should instead invest efforts into `pioarduino` + `Arduino-BLE` (i.e. *not* `NimBLE-Arduino`). That work is ongoing here: https://github.com/meshtastic/firmware/pull/9132

so I won't be pursuing `NimBLE-Arduino` v2.x but I wanted to post my notes for posterity.

Issue #6390, opened 2025-03-24
[Bug]: bug in device-install.sh


sq5bpf:
### Category

Other

### Hardware

Heltec V3

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.6.2.31c0e8f

### Description

There is a bug with matching a portion of the filename with a construct like if [ -n "${FILENAME##*"$variant"*}" ]
This evaluates to true at least on recent Debian 12.9 which uses bash version 5.2.15(1)-release (x86_64-pc-linux-gnu). To check just put an echo statement in the if-s and see how much times they match. Also please check the log output with incorrect addresses for an 8MB heltec v3.

I'm attaching a fixed version [device-install.sh.txt](https://github.com/user-attachments/files/19438378/device-install.sh.txt)

This is the diff:

```
--- device-install.sh.orig      2025-03-20 21:59:02.000000000 +0100
+++ device-install.sh   2025-03-24 22:16:27.661217168 +0100
@@ -138,7 +138,7 @@
 
        # littlefs* offset for BigDB 8mb and OTA OFFSET.
        for variant in "${BIGDB_8MB[@]}"; do
-               if [ -n "${FILENAME##*"$variant"*}" ]; then
+               if [[ "${FILENAME}" == *"${variant}"* ]]; then
                        OFFSET=0x670000
                        OTA_OFFSET=0x340000


                fi
@@ -146,7 +146,7 @@
 
        # littlefs* offset for BigDB 16mb and OTA OFFSET.
        for variant in "${BIGDB_16MB[@]}"; do
-               if [ -n "${FILENAME##*"$variant"*}" ]; then
+               if [[ "${FILENAME}" == *"${variant}"* ]]; then
                        OFFSET=0xc90000
                        OTA_OFFSET=0x650000
                fi
@@ -155,13 +155,13 @@
        # Account for S3 board's different OTA partition
        # FIXME: Use PlatformIO info to determine MCU type, this is unmaintainable
        for variant in "${S3_VARIANTS[@]}"; do
-               if [ -n "${FILENAME##*"$variant"*}" ]; then
+               if [[ "${FILENAME}" == *"${variant}"* ]]; then
                        MCU="esp32s3"
                fi
        done
 
        if [ "$MCU" != "esp32s3" ]; then
[...]

PR #6517, opened 2025-04-08
Fix sense of device set membership checks

interfect:
The flasher Bash script uses a neat `"${FILENAME##*"$variant"*}"` Bash trick to replace any string in `FILENAME` containing the string in `variant` with an empty string. It puts this in loops over arrays to check if `FILENAME` contains any `variant` in the list.

But the test needs to be `-z` (i.e. match the empty string, which we get if we found `variant`) if we want to operate on filenames that have any items from the list. With `-n` we pass the condition if we *don't* find the variant, and then operate on filenames that *don't* have *every* variant from the list.

I tested this and I couldn't flash a T3S3 without it; it would think it was one of the `BIGDB_16MB` boards and try to flash the OTA at an address outside the flash.

With it, I can flash my T3S3 fine, but I don't actually have any of the 8MB or 16MB boards, so I haven't tested that it actually properly recognizes them.

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - LilyGo T3S3

### DIFF
--- a/bin/device-install.sh
+++ b/bin/device-install.sh
@@ -138,7 +138,7 @@ if [ -f "${FILENAME}" ] && [ -n "${FILENAME##*"update"*}" ]; then
 
 	# littlefs* offset for BigDB 8mb and OTA OFFSET.
 	for variant in "${BIGDB_8MB[@]}"; do
-		if [ -n "${FILENAME##*"$variant"*}" ]; then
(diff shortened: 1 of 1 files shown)


thebentern:
Closed in #6390 :-)

PR #6526, opened 2025-04-08
MUI: update device-ui commit reference

mverch67:
new feature: map locations filtering
bugfix: boot logo / bt logo

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -94,7 +94,7 @@ lib_deps =
 
 [device-ui_base]
 lib_deps =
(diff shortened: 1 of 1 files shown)


thebentern (approved):

vidplace7:
@thebentern -- Dependabot unfortunately doesn't support updating dependencies in platformio.ini files at all. Dependabot is really pretty basic.

We could set up renovatebot to do this pretty easily (I've set it up in a few of my own projects), probably worth doing.

PR #6560, opened 2025-04-10
Improve performance of encrypted packets with a shared secret cache

### LINKED ISSUE #6426: [Feature Request]: Improve performance of encrypted packet usage
### Platform

Cross-Platform

### Description

There appears to be a fair amount of low hanging fruit that, if my understanding is correct, may dramatically improve performance of encrypted packet usage (and by extension, battery usage). Areas of interest:
* Nodes relaying packets on encrypted channels must decrypt the entire payload. Encrypting a few bytes separately from the payload would allow relayers to check if they could decrypt the message without necessarily needing to.
* The shared secret used for encrypted channels is recomputed for every packet. Generating an ECDH secret is typically an order of magnitude (or more!) expensive than encryption/decryption. If true, caching secrets could cut CPU/battery usage during decryption by 50% or more for small packets (< 1kb). There are some hurdles to tackle like the memory usage of such a cache, but at a glance it appears achievable.

Feel free to assign this to me. I am still exploring the codebase.

jasonbcox:
Add a shared secret cache to dramatically improve performance and battery usage of encrypted communications after the first packet is sent between two nodes.

Using this additional patch https://github.com/jasonbcox/meshtastic-firmware/commit/190444fce6db6de3cc482291bc60f1b065eeab06, I logged the following on a RAK4631 when sending DMs between it and another node:
On master:
* PKC encryption/decryption: ~96ms
* Packet routing: ~107ms
* All other modules, total: <5ms
* Total packet handling time: 210 - 220ms

With this patch applied, first encrypted packet is same as above. All subsequent packets between the same two nodes perform as follows (times are roughly the same regardless of message length):
* PKC encryption/decryption: ~3ms
* Other modules (routing, etc): same as above
* Total packet handling time: ~120ms

This is a **32x performance improvement** of encrypted packet decoding at the cost of (up to) ~2.4kb of RAM.

For users that send DMs often, and assuming that every ms of compute is energy cost equivalent (its not, but we can estimate), this could provide around **45% improvement in battery usage** with respect to encrypted communications.

Unit test coverage recently improved to ensure this patch does not regress, in addition to tests added in this PR (see #6485)

Fixes #6426 

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631


jp-bennett:
I definitely approve in principle. Since it's touching the crypto code, it needs more detailed review than I can give tonight.

GUVWAF:
I really appreciate the work you put into this, but I think we should also be realistic about the actual gains of this versus the drawback. 2.4k**Bytes** of RAM is not insignificant, that’s 12 nodes in the NodeDB and ~1% of the total RAM available for nRF52.

Regarding power consumption: I think if a node receives 100 messages directed towards it per day that would be a lot already. In this case we thus save 9.3 seconds of compute per day, which is only 0.01% of the time. However, for nodes connected to a client (it’s those nodes that you’re sending a DM to) we’re not going into a real sleep mode, so we’ll just be longer in `mainDelay.delay()`. How much this saves on energy depends on how the idle state is implemented on each platform, but e.g. the Arduino core for ESP32 doesn’t use power management (`CONFIG_PM_ENABLE` is not set).

Even if the power consumption of the CPU could be completely eliminated during 0.01% of the time, it should be considered relative to the LoRa radio that is always in receiving mode consuming about ~4.2mA for SX126x. If we would be able to mitigate only one packet transmission per day, which takes ~1 second on LongFast during which it consumes ~125mA more, I’d argue that’s already more than can be gained from this.

thebentern:
> I really appreciate the work you put into this, but I think we should also be realistic about the actual gains of this versus the drawback. 2.4k**Bytes** of RAM is not insignificant, that’s 12 nodes in the NodeDB and ~1% of the total RAM available for nRF52.
> 
> Regarding power consumption: I think if a node receives 100 messages directed towards it per day that would be a lot already. In this case we thus save 9.3 seconds of compute per day, which is only 0.01% of the time. However, for nodes connected to a client (it’s those nodes that you’re sending a DM to) we’re not going into a real sleep mode, so we’ll just be longer in `mainDelay.delay()`. How much this saves on energy depends on how the idle state is implemented on each platform, but e.g. the Arduino core for ESP32 doesn’t use power management (`CONFIG_PM_ENABLE` is not set).
> 
> Even if the power consumption of the CPU could be completely eliminated during 0.01% of the time, it should be considered relative to the LoRa radio that is always in receiving mode consuming about ~4.2mA for SX126x. If we would be able to mitigate only one packet transmission per day, which takes ~1 second on LongFast during which it consumes ~125mA more, I’d argue that’s already more than can be gained from this.

Does it make sense to utilize this only on non-NRF platforms? I was actually thinking the main benefit of this would be significantly increased throughput for network connected nodes instead of improved power consumption, which as you pointed out is insignificant

jp-bennett:
> Does it make sense to utilize this only on non-NRF platforms? I was actually thinking the main benefit of this would be significantly increased throughput for network connected nodes instead of improved power consumption, which as you pointed out is insignificant

We can set the size much smaller on the NRF platform. 64 slots is quite a few. I'd set it to 8 on NRF.

Not entirely related, I do wonder if using a dynamic object like the unordered_map will lead to memory fragmentation.

GUVWAF:
> I was actually thinking the main benefit of this would be significantly increased throughput for network connected nodes instead of improved power consumption

Still I think before the decryption process becomes the bottleneck in terms of throughput you'll need to receive a lot of direct messages after each other via MQTT or UDP.

> I could reduce the default number of slots to 32 or even 24 and I think the cache would perform nearly as well for the vast majority of users.

Maybe make it dependent on `MAX_NUM_NODES`?

thebentern:
> I could not make it dependent on MAX_NUM_NODES directly since it is not a constant expression at compile time.

It's not a compile time constant on all platforms but it is static. I wonder if we can static init the length struct to  MAX_NUM_NODES / 10 ?
image

thebentern:
> I wasn't aware changes to MAX_NUM_NODES were coming down the pipe. Can you link the most relevant one?

Sorry, I meant upstream prior to your initial PR. The latest change has them reliant on some runtime checks on espressif hardware.

jp-bennett:
The downside of the stale bot is that good PRs like this one get dropped.

jp-bennett (approved):

jp-bennett:
The way this is currently written, we're only talking a max of 380ish bytes, right? Playing with this patch for some higher-throughput DM experimentation, and it seems to be working as expected so far.

src/mesh/CryptoEngine.cpp
```diff
@@ -270,6 +268,69 @@ void CryptoEngine::initNonce(uint32_t fromNode, uint64_t packetId, uint32_t extr
...
 }
+
+bool CryptoEngine::setCryptoSharedSecret(meshtastic_UserLite_public_key_t pubkey)
+{
+    // The last used timestamp is in units of ~1.165 hours, which gives us
+    // ~12.3 days before the timestamps roll over. This is ok since a periodic
+    // misfire on evicting the oldest secret has very little impact.
+    const uint8_t now = (millis() >> 22) & 0xff;
+
+    // Get a short lookup key from the pubkey
```
caveman99, line 279: Only 32 of the public key's 256 bits are ever compared. Any two keys sharing a 4-byte prefix are treated as the same peer. Grinding a key pair where the first 4 bytes are the same than the victim key is a matter of minutes these days.

src/mesh/CryptoEngine.cpp
```diff
@@ -270,6 +268,69 @@ void CryptoEngine::initNonce(uint32_t fromNode, uint64_t packetId, uint32_t extr
     if (extraNonce)
         memcpy(nonce + sizeof(uint32_t), &extraNonce, sizeof(uint32_t));
 }
+
+bool CryptoEngine::setCryptoSharedSecret(meshtastic_UserLite_public_key_t pubkey)
```
caveman99, line 272: meshtastic_NodeInfoLite_public_key_t is the right type now

caveman99 (changes requested):
Please rebase this to develop. The delta is small enough, but things have moved on considerably since this was posted.

PR #6407, opened 2025-03-27
Fix: T-Watch-S3 has 8MB Flash

vidplace7:
T-Watch-S3 has 8MB of flash, not 16!

Resolves #6406

PR #6563, opened 2025-04-11
fix: Correct underlying cause of T-Watch not functioning when set to a 16MB filesystem

Kealper:
While looking at recent commits, I spotted the problem with the LilyGo T-Watch S3 behaving as if it only had 8MB of flash when LilyGo claims it has 16MB: The original board definition from when the T-Watch was first added incorrectly set the maximum flash size to 8MB instead of 16MB. Fixing that allows the full 16MB filesystem image to upload and work successfully. This PR also reverts #6407 as those changes to bump the device down to the 8MB filesystem are no longer required.

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] LilyGo T-Watch S3

### DIFF
--- a/bin/device-install.bat
+++ b/bin/device-install.bat
@@ -17,8 +17,8 @@ SET "LOGCOUNTER=0"
 SET "S3=s3 v3 t-deck wireless-paper wireless-tracker station-g2 unphone"
 SET "C3=esp32c3"
 @REM FIXME: Determine flash size from PlatformIO variant, this is unmaintainable.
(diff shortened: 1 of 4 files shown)


fifieldt:
I vaguely recall some comments in discord that some T-Watches did have only 8MB. May I confirm that this is actually not the case and they physically all have 16MB?

Kealper:
I was actually part of that original discussion, on the tail-end of it anyways. The original discussion was sparked because the T-Watch failed to successfully flash/run after the change that increased the filesystem size to accommodate more nodes in the node DB, and someone had seen "8MB" for the PSRAM listed in the output of esptool's `chip_id` and mistakenly thought that was flash size. The original "fix" of bumping the filesystem down to the 8MB version worked because the esptool max flash size was erroneously set to 8MB in the board definition when the T-Watch was added. I guess this was never noticed until we tried to bump up the flash usage beyond that for devices that could do so.

To my knowledge, all T-Watch S3s should have 16MB of flash and 8MB of PSRAM. I have tested this on my own T-Watch successfully, whereas before this change, mine had the same issues as everyone else's.

fifieldt:
Thanks!

fifieldt (approved):

caveman99 (approved):

PR #6569, opened 2025-04-11
PMSA003I: add support for driving SET pin low while not actively taking a telemetry reading

### LINKED ISSUE #6518: [Bug]: Fan of the Particle Matter Sensor is rotating all the time
### Category

Other

### Hardware

Rak4631

- [ ] InkHUD ePaper

### Firmware Version

2.6.4.b89355f

### Description

I am using the Plantower PMSA003I (RAK12039) connected to the RAK19007/3 baseboard with the RAK4631. I've noticed that the fan runs continuously, even when it's not measuring. The rotation speed increases during measurements, but it continues to rotate slowly during the intervals between measurements. This behavior consumes a significant amount of energy when running on batteries; for instance, it uses 40% of a 4000 mAh LiPo battery overnight.

RAKwireless technical support mentioned that the fan can be powered down. Ideally, I would suggest the following cycle: the fan should operate at a low rotation speed for two minutes, followed by PM concentration measurement, and then turn off completely.

### Relevant log output

```Shell

```

vogon:
the Plantower PMSA003I sensor takes about 60-120mA of power at all times while the fan is running, as I discovered when I was building a solar-powered DIY weather station node.  now you can assign a GPIO in your variant file with `PMSA003I_ENABLE_PIN` which drives the SET pin on the sensor low to put it in standby mode.  tested on my DIY node with a fork of 2.5.20.4c97351.  

fixes #6518.

known todos:
- [x] verify that there are no regressions in `master`.
- [x] verify that it doesn't cause build failures on my Heltec V3.
- [x] work with the reporter on #6518 to help them make the appropriate changes for their node's variant file and test them.

## 🤝 Attestations
  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631
  - [x] Other (please specify below): DIY node with a Heltec Mesh Node T114 and Adafruit PMSA003I breakout board.


caveman99 (approved):

fifieldt:
Hi @vogon , thanks a lot for this patch :)

Are you familiar with the trunk linting tool? We use it to keep our codebase consistent

vogon:
@fifieldt I'm not, sorry!  would be happy to learn more and make it part of my workflow.

also, I followed up offline with someone with a RAK node and managed to get the change made to the RAK variant to allow it to work out of box.  more changes for that forthcoming.

ArgoNavi:
NFC2 is IO6 on the RAK4631, which is broken out on the RAK13002

Works with my Adafruit PMSA003I

PR #6575, opened 2025-04-13
adding support for the ST7796 + creating a new variant of the T-beam

Nasimovy:
modified the existing library for the [ST7789](https://github.com/meshtastic/st7789) to support the [ST7796](https://github.com/Nasimovy/st7796)
created a new variant of the T-beam that supports the [ST7796 shield](https://lilygo.cc/products/t-beam-shiled-v1-0?_pos=1&_psq=shiled&_ss=e&_v=1.0)

## 🤝 Attestations
  - [x] LilyGo T-Beam


Nasimovy:
converted it to draft because i seem to get SPI related errors

variants/tbeam_TFT/platformio.ini
```diff
@@ -0,0 +1,14 @@
+; The 1.0 release of the TBEAM board 
+[env:tbeam_TFT]
```
caveman99, line 2: Please rename the variant to something else. the tft suffix is only meant for devices running MUI and is parsed in some install scripts.  My suggestion would be tbeam-displayshield

variants/tbeam_TFT/platformio.ini
```diff
@@ -0,0 +1,14 @@
...
+extends = esp32_base
+board = ttgo-t-beam
+board_check = true
+lib_deps = ${esp32_base.lib_deps}
+  https://github.com/Nasimovy/st7796/archive/refs/tags/1.0.5.zip ; display addon
```
caveman99, line 7: If this works with the touchscreen, i'd recommend moving it to the meshtastic org like the 7789 code.

caveman99 (changes requested):
Looking good overall, though i'd prefer if you got the touchscreen working as well

variants/tbeam_TFT/platformio.ini
caveman99: Also, why not derive it from the base tbeam definition in the same environment, like we do for instance in t-deck?

variants/tbeam_TFT/platformio.ini
caveman99: the sensorlib is also used for t-beam supreme ... i think you can just pull it in (the latest one is buggy, use the version referenced already in here)

Nasimovy:
im trying lower SPI speeds for both the LORA module and the display because it looks like the tbeam is very EMI susceptible, or the ST7796 pumps out loads of EMI.

yesterday i had a SPI related error every ~20 minutes, also next to zero Bluetooth reception

i first want to resolve that issue before i start implementing the touchscreen

caveman99 (approved):

variants/tbeam_TFT/platformio.ini
Nasimovy: for moving it to the meshtastic org how should that be done?

Nasimovy:
touchscreen is recognized and responds to touches, swipe is also recognized.
observer is not working though.

Nasimovy:
the addition of a fixed baudrate for the GPS solved the GPS issue, i dont know and cant test if other T-beams with another variant of the Ublox module has the same baudrate

i had to update the NIMBLE because of stack smashing protect failure, i will try the latest alfa and report back if the problem exists there also.

caveman99:
Thanks for the great work so far. the GPS Speed issue and the NimBLE update should be untangled from this PR. The former can be looked into later and the latter is subject of a different PR we're debugging right now. Please only submit the ST7796 support and the touchscreen stuff.

For the move of the drive repo, i will simply fork your repo back into the org. This is more straight forward than moving it.

Nasimovy:
I wanted to revert the NIMBLE, but i ran into unrelated issues with VScode, i will split it up when i get access to my desktop.

The GPS is not getting detected on startup so i thought i should include it here, the T-beam works fine without the display code though.

Nasimovy:
code works again with NIMble reverted, only issue now is that gps does not get detected on startup

caveman99:
standard baud rate on the t-beam is 9600. I am curious what your 115200 comes from. can you run a GPS factory reset sketch on your t-beam and try again?

Nasimovy:
the gps works!, it does not get identified on startup but it puts out data and meshtastic picks that up

caveman99 (approved):

mverch67 (commented):
I think using **lovyanGFX** (TFTDisplay.cpp) would have saved some work as it natively supports the ST7796 display driver and also the CST226 touch screen driver (in its _develop_ branch). It would have allowed smoother integration into TFTDisplay.cpp and avoided some code duplication for the touch driver (e.g. onEvent) as it would have been handled automatically by class _TouchScreenImpl1_. As a benefit screen powersave functions could have been inherited as well.

src/main.cpp
```diff
@@ -1292,7 +1292,7 @@ void setup()
     PowerFSM_setup(); // we will transition to ON in a couple of seconds, FIXME, only do this for cold boots, not waking from SDS
     powerFSMthread = new PowerFSMThread();
 
-#if !HAS_TFT
+#if !HAS_TFT && !USE_ST7796
     setCPUFast(false); // 80MHz is fine for our slow peripherals
```
mverch67, line 1296: This now affects all tbeams 1.0 out there also those that are NOT equipped with this ST7796 shield (i.e. 99.9%) and as a result all these probably comsume more power than before.
Is there a reason for setting the frequency to 240MHz for the oled style UI (which is not very interactive).
Maybe you can try the setting `SCREEN_TRANSITION_FRAMERATE` instead.

mverch67 (changes requested):

Nasimovy:
> > lovyangfx makes the program to large to fit on the t-beam (iram overflow ~2700)
> 
> The current tbeam size in master is:
> 
> ```
> RAM:   [=         ]   7.4% (used 96820 bytes from 1310720 bytes)
> Flash: [========= ]  89.3% (used 2165385 bytes from 2424832 bytes)
> ```
> 
> I wouldn't expect lovyanGFX being that large (260kB)?!

the IRAM is insufficient by 2736 bytes

```
/root/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pio/build/tbeam-displayshield/firmware.elf section `.iram0.text' will not fit in region `iram0_0_seg'
/root/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: IRAM0 segment data does not fit.
/root/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: region `iram0_0_seg' overflowed by 2736 bytes
```

src/graphics/images.h
```diff
@@ -21,7 +21,8 @@ const uint8_t bluetoothConnectedIcon[36] PROGMEM = {0xfe, 0x01, 0xff, 0x03, 0x03
 #endif
 
 #if (defined(USE_EINK) || defined(ILI9341_DRIVER) || defined(ILI9342_DRIVER) || defined(ST7701_CS) || defined(ST7735_CS) ||      \
-     defined(ST7789_CS) || defined(USE_ST7789) || defined(HX8357_CS) || defined(ILI9488_CS) || ARCH_PORTDUINO) &&                \
+     defined(ST7789_CS) || defined(USE_ST7789) || defined(USE_ST7796) || defined(HX8357_CS) || defined(ILI9488_CS) ||            \
+     ARCH_PORTDUINO) &&                                                                                                          \
```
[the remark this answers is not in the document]
Nasimovy: there is no OR list after the &&

src/mesh/NodeDB.cpp
```diff
@@ -600,7 +600,7 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
     config.bluetooth.fixed_pin = defaultBLEPin;
 
 #if defined(ST7735_CS) || defined(USE_EINK) || defined(ILI9341_DRIVER) || defined(ILI9342_DRIVER) || defined(ST7789_CS) ||       \
-    defined(HX8357_CS) || defined(USE_ST7789) || defined(ILI9488_CS)
+    defined(HX8357_CS) || defined(USE_ST7789) || defined(ILI9488_CS) || defined(USE_ST7796)
```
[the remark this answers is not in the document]
Nasimovy: the get_st7789_id only gets defined/used when T114 is defined?

src/main.cpp
mverch67: Have you tested `SCREEN_TRANSITION_FRAMERATE` yet? All other TFT devices set it to 3 or 5 (default is 30!).

Nasimovy: I have not, i read over that part, i will check that for sure, tomorrow morning or after work

Nasimovy: i am testing it now, it seams that normal left right swiping is not affected but typing feels smoother i think

mverch67: The lower this value the less CPU time is used which explains why typing feels smoother. If it's too low the swiping will also get slower (e.g. only one swipe per second when value is 1).

src/input/TouchScreenCST226SE.cpp
```diff
@@ -0,0 +1,104 @@
...
+                pinMode(TOUCH_IRQ, INPUT_PULLUP);
+                attachInterrupt(
+                    TOUCH_IRQ, []() { isPressed = true; }, FALLING);
+            }
+            LOG_DEBUG("CST226SE init OK at address 0x%02X", addr);
+            inputBroker->registerSource(this);
+            return;
+        }
+    }
+    LOG_ERROR("CST226SE init failed at all known addresses");
```
mverch67: I think it's missing a call to `TouchScreenBase::init(false);` in case the initialisation went wrong.

src/input/TouchScreenCST226SE.h
```diff
@@ -0,0 +1,30 @@
...
+
+#ifdef HAS_CST226SE
+
+#include "modules/CannedMessageModule.h"
+
+#include "TouchDrvCSTXXX.hpp"
+#include "TouchScreenBase.h"
+
+class TouchScreenCST226SE : public TouchScreenBase
+{
```
mverch67: I think if you inherit from the generic class `TouchScreenImpl1` instead you can remove the duplicated method `onEvent()` as its functionality gets inherited automatically.
It should work because the overloaded method `TouchScreenCST226SE::getTouch()` is defined virtual and also the (non-virtual) `TouchScreenCST226SE::init()` method is called directly without base pointer which is ok as well.

Nasimovy:
is there still something i need to improve change or test?

mverch67:
> is there still something i need to improve change or test?

Yes, I would say so. Look, the original design of the touch screen impl class is completely generic and there's no need to duplicate this entire class and add a little bit, that's not the way object-oriented C++ is intended to be handled.

Have a look at how I implemented the new touch driver for the T-Deck Plus E-Ink screen:
https://github.com/meshtastic/firmware/pull/6936/files#diff-35aefebed15b5365cae9167c43827cf77fc61afab35cc089a3a46b26af67c3b0

You'll see that here I make fully use of the generic TouchScreenImpl1 class and it's interface. That's the way how it's supposed to be used for registering touch drivers.

Nasimovy:
it works now with the new oled ui and with the default touchscreen implementation

src/platform/extra_variants/tbeam_displayshield/variant.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+TouchDrvCSTXXX tsPanel;
+static constexpr uint8_t PossibleAddresses[2] = {CST226SE_ADDR, CST226SE_ADDR_ALT};
+volatile bool isPressed = false;
+uint8_t i2cAddress = 0;
+
+bool readTouch(int16_t *x, int16_t *y)
+{
+    int16_t x_array[1], y_array[1];
+    uint8_t touched = tsPanel.getPoint(x_array, y_array, 1);
```
mverch67, line 16: the variable _isPressed_ is nowhere used. Also, using an interrupt handler for readTouch is a bit overkill for a touch method that is polled, the touch GPIO pin could be read directly, e.g.
```C++
if (digitalRead(TOUCH_IRQ) == 0) {
    uint8_t touched = tsPanel.getPoint(x_array, y_array, 1);
    ...
}
```

Nasimovy: i removed the interrupt pin because it does not seem to work, also removed the is pressed

.vscode/settings.json
```diff
@@ -10,5 +10,17 @@
   },
   "[powershell]": {
     "editor.defaultFormatter": "ms-vscode.powershell"
+  },
```
mverch67: This file should not be committed into github, it's a local workspace file.

Nasimovy:
@mverch67 it seams that i am blind, github says there is still a requested change but i cant find it

mverch67:
> @mverch67 it seams that i am blind, github says there is still a requested change but i cant find it

maybe use from terminal:
```
git diff
git restore <file>
git checkout T-beam-display-no-touch -- .vscode/settings.json
```

mverch67 (approved):

Nasimovy:
hello, could somebody that has a device with a CST328 test if the detection is working properly?, the else case is the CST328, it detects the CST226SE

mverch67:
The T-Deck Pro has a CST328.

Actually the scan result does not matter as the touch controller is hard-coded for the device variant.

Nasimovy:
Hey, @mverch67 , @caveman99 , @thebentern , @vidplace7 , sorry for the ping in the first place.

I know it has been a long road for me to get it working properly as in no code duplication etc...

But i think the final version is good/proper, could you guys take a look at it and hopefully merge it?

kind regards, Wout
i don't want to sound pushy but i get or know that it for some would look like that.

mverch67:
> But i think the final version is good/proper, could you guys take a look at it and hopefully merge it?

[PR7613](https://github.com/meshtastic/firmware/pull/7613) also integrates the ST7796 but via TFTDisplay as most other TFT displays do. It would be more homogeneous to also adopt this way for the tbeam display otherwise the code gets more and more confusing (e.g. ST7796_CS vs. USE_ST7796)...

Nasimovy:
>  integrates the ST7796 but via TFTDisplay as most other TFT displays do ST7796_CS vs. USE_ST7796

i understand that code readability is important, but USE_ST7789 is also used with the same driver type as my use case of the USE_ST7796, the mui or the lovyangfx driver is to large to use on the esp32 i know that @Szetya got it working with [lovyangfx](https://github.com/meshtastic/firmware/issues/6914#issuecomment-3109454329) but he had to exclude default included features.
So i don't really know the way forward.
An if T-beam case perhaps or a other way

thanks mverch67 for your input

Szetya:
> > integrates the ST7796 but via TFTDisplay as most other TFT displays do ST7796_CS vs. USE_ST7796
> 
> i understand that code readability is important, but USE_ST7789 is also used with the same driver type as my use case of the USE_ST7796, the mui or the lovyangfx driver is to large to use on the esp32 i know that @Szetya got it working with [lovyangfx](https://github.com/meshtastic/firmware/issues/6914#issuecomment-3109454329) but he had to exclude default included features. So i don't really know the way forward. An if T-beam case perhaps or a other way
> 
> thanks mverch67 for your input

USE_DISPLAYTYPE has its own driver. 
And for DISPLAYTYPE_CS, lovyangfx.
This is exactly why the USE_DISPLAYTYPE macro was introduced.

src/graphics/Screen.cpp
```diff
@@ -588,7 +629,7 @@ void Screen::setup()
             touchScreenImpl1->init();
         }
     }
-#elif HAS_TOUCHSCREEN && !defined(USE_EINK)
+#elif HAS_TOUCHSCREEN && !defined(USE_EINK) && !HAS_CST226SE
```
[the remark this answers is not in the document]
Nasimovy: requires multiple changes, should this be done, in a other way that's more convenient / readable?

src/detect/ScanI2CTwoWire.cpp
```diff
@@ -478,8 +489,12 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
 #endif
 
             case MLX90614_ADDR_DEF:
-                registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0x0e), 1);
-                if (registerValue == 0x5a) {
+                // Do we have the MLX90614 or the MPR121KB or the CST226SE
+                registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0x06), 1);
+                if (registerValue == 0xAB) {
```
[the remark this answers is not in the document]
Nasimovy: does this increase readability or should i add a comment on how or what the hex code means?

src/detect/ScanI2CTwoWire.cpp
```diff
@@ -468,7 +468,18 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
                 SCAN_SIMPLE_CASE(DFROBOT_RAIN_ADDR, DFROBOT_RAIN, "DFRobot Rain Gauge", (uint8_t)addr.address);
                 SCAN_SIMPLE_CASE(LTR390UV_ADDR, LTR390UV, "LTR390UV", (uint8_t)addr.address);
                 SCAN_SIMPLE_CASE(PCT2075_ADDR, PCT2075, "PCT2075", (uint8_t)addr.address);
-                SCAN_SIMPLE_CASE(CST328_ADDR, CST328, "CST328", (uint8_t)addr.address);
+            case CST328_ADDR:
+                // Do we have the CST328 or the CST226SE
+                registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0xAB), 8);
+                if (registerValue == 0x3232365345) {
```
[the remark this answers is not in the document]
Nasimovy: does this increase readability or should i add a comment on how or what the hex code means?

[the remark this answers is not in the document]
mverch67: `getRegisterValue()` returns a uint16_t. You cannot compare it with a 40bit value...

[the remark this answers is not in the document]
Nasimovy: that was a dumb mistake on my end, it worked still because the CST226SE has 2 possible addresses and my touchscreen uses the alt one

Nasimovy:
code is still working on my t-beam

Xaositek:
Testing in the develop branch, this has broken all aspects of the TFT_MESH color being applied and is reverting all screens back to the default Green.

Nasimovy:
I will look into it and report what I can find 
Thanks for letting me know

Xaositek:
> I will look into it and report what I can find 
> Thanks for letting me know

I have fixed it in develop already - https://github.com/meshtastic/firmware/pull/8796

PR #6584, opened 2025-04-14
Update platformio.ini to exclude unused modules from t1000-e

### LINKED ISSUE #6447: [Feature Request]: Disable Canned Messages, Display, Serial, Detection Sensor and Network Modules on T-1000E
### Platform

NRF52

### Description

Add Canned Messages, Display, Serial, detection sensor and the network modules as an excluded modules for the T-1000E to simplify app settings as they are not usable.

benkyd:
Resolves #6447 

## 🤝 Attestations


benkyd:
@garthvh @fifieldt I am sure this is the right change to be making however on the iPhone app the options are not greyed out. But on the android app, instead of being greyed out the options disabled are not present. Please confirm before merging.

GUVWAF:
You'll need to add `NO_EXT_GPIO` and it seems the excluded modules tag needs some changes. For example it should be `|| MESHTASTIC_EXCLUDE_CANNEDMESSAGES` here:
https://github.com/meshtastic/firmware/blob/28e62e53e5dd83888ecf1bf5bd288c3684fcde62/src/main.cpp#L1313

And I think here we'll need to add  `|| MESHTASTIC_EXCLUDE_SERIAL` also:
https://github.com/meshtastic/firmware/blob/28e62e53e5dd83888ecf1bf5bd288c3684fcde62/src/main.cpp#L1325

benkyd:
> You'll need to add `NO_EXT_GPIO` and it seems the excluded modules tag needs some changes. For example it should be `|| MESHTASTIC_EXCLUDE_CANNEDMESSAGES` here:
> 
> https://github.com/meshtastic/firmware/blob/28e62e53e5dd83888ecf1bf5bd288c3684fcde62/src/main.cpp#L1313
> 
> And I think here we'll need to add `|| MESHTASTIC_EXCLUDE_SERIAL` also:
> 
> https://github.com/meshtastic/firmware/blob/28e62e53e5dd83888ecf1bf5bd288c3684fcde62/src/main.cpp#L1325

Thanks for the comments. I was thinking this about the tags although they seemed like the correct way to implement such a feature. I will include proper tagging in this pull request too.

thebentern:
We should leave serial out of this because of this mode (which utilizes USB):
https://meshtastic.org/docs/configuration/module/serial/#override-console-serial-port

benkyd:
> We should leave serial out of this because of this mode (which utilizes USB): https://meshtastic.org/docs/configuration/module/serial/#override-console-serial-port

Agree, i thought it weird to see it in the issue

todd-herbert:
> You'll need to add `NO_EXT_GPIO` and it seems the excluded modules tag needs some changes. For example it should be `|| MESHTASTIC_EXCLUDE_CANNEDMESSAGES` here:
> 
> https://github.com/meshtastic/firmware/blob/28e62e53e5dd83888ecf1bf5bd288c3684fcde62/src/main.cpp#L1313
> 
> And I think here we'll need to add `|| MESHTASTIC_EXCLUDE_SERIAL` also:
> 
> https://github.com/meshtastic/firmware/blob/28e62e53e5dd83888ecf1bf5bd288c3684fcde62/src/main.cpp#L1325

This one is my fault.. sorry!
<https://github.com/meshtastic/firmware/pull/6272#discussion_r1986206223>

fifieldt (approved):

GUVWAF (changes requested):
Hmm, wait we have to be careful. When using `NO_EXT_GPIO` it also excludes the External Notification module in the tag.

The macro to exclude the Canned Messages module is now called `MESHTASTIC_EXCLUDE_CANNEDMESSAGES`, not `MESHTASTIC_EXCLUDE_CANNEDMSG`. If we'd go with @todd-herbert's approach, we'd need to replace all these with the `MESHTASTIC_INCLUDE_CANNEDMESSAGES` macro. However, I wonder if we really want to check for `(!HAS_SCREEN || NO_EXT_GPIO)` or just only explicitly exclude the module as the e-ink variants have `HAS_SCREEN 0` and you don't need external GPIOs if you have buttons already.

benkyd:
@todd-herbert 
> I was hoping to have that explicit opt-in for canned messages, to implementing them in InkHUD in the near future, even though those builds are `HAS_SCREEN=0` and often `NO_EXT_GPIO`

Yeah I caught that nowhere else uses opt-in for canned-messages so it was always opt-out. This way makes the change on the app that was intended - but good catch @GUVWAF. That should be cannedmessages not msg. That is fixed in 0ee20bb5. 

IMO swapping to opt-in instead of opt-out is a much larger change for a seperate PR if leads decide that is the best course.

As it is in the code it currently works as intended other than the screen as it seems that there currently isn't functionality to disable the screen code from appearing in the app, even though the opt-out currently stops it from being compiled. IMO there should be  some consolidation in how core and non-core modules behave when they are/ are not compiled. but for now this correction achieves the point of this PR. I can open a seperate code-cleanliness issue regarding consistency of opt-in and out messaging across the code.

todd-herbert:
> Hmm, wait we have to be careful. When using `NO_EXT_GPIO` it also excludes the External Notification module in the tag.
> 
> The macro to exclude the Canned Messages module is now called `MESHTASTIC_EXCLUDE_CANNEDMESSAGES`, not `MESHTASTIC_EXCLUDE_CANNEDMSG`. If we'd go with @todd-herbert's approach, we'd need to replace all these with the `MESHTASTIC_INCLUDE_CANNEDMESSAGES` macro. However, I wonder if we really want to check for `(!HAS_SCREEN || NO_EXT_GPIO)` or just only explicitly exclude the module as the e-ink variants have `HAS_SCREEN 0` and you don't need external GPIOs if you have buttons already.

To be clear, I wasn't envisioning replacing `MESHTASTIC_EXCLUDE_CANNEDMESSAGES` with `MESHTASTIC_INCLUDE_CANNEDMESSAGES` everywhere, just a one-off use of the INCLUDE one in that specific line.

The original intention there was to keep the code agnostic; to allow any of the weird cases (like InkHUD) to just drop the INCLUDE macro into their build_flags. 

I'm not at all attached to doing it that way though, it was just my first thought. Whatever suits! 👍

GUVWAF (approved):
> I can definitely understand the desire to not use a weird mix of exclude and include though.

Yeah, I would vote for not having both an INCLUDE and EXCLUDE as EXCLUDE=0 should already imply INCLUDE=1 in my view. 

Looks good now, only that you will not have the canned message settings available for devices that have `HAS_SCREEN=0` or `NO_EXT_GPIO=1`. So once the InkHUD has this feature, it might need something like
`(!HAS_SCREEN && !USE_EINK) || (NO_EXT_GPIO && !BUTTON_PIN)` to accommodate for this without needing the explicit INCLUDE macro. 

@benkyd Thanks for your first PR!

benkyd:
> > I can definitely understand the desire to not use a weird mix of exclude and include though.
> 
> 
> 
> Yeah, I would vote for not having both an INCLUDE and EXCLUDE as EXCLUDE=0 should already imply INCLUDE=1 in my view. 
> 
> 
> 
> Looks good now, only that you will not have the canned message settings available for devices that have `HAS_SCREEN=0` or `NO_EXT_GPIO=1`. So once the InkHUD has this feature, it might need something like
> 
> `(!HAS_SCREEN && !USE_EINK) || (NO_EXT_GPIO && !BUTTON_PIN)` to accommodate for this without needing the explicit INCLUDE macro. 
> 
> 
> 
> @benkyd Thanks for your first PR!

Sounds good, needs consideration for future projects for sure. looking forward to the merge.

PR #6591, opened 2025-04-15
Make startup screen show the short ID

Heltec-Aaron-Lee:
**Modify**
Show the device's short ID on the boot screen.

**Fucntion**
If the user has multiple devices around, this small change will help the user quickly distinguish between the devices.

image


todd-herbert:
I don't think this is necessary, as the short name is shown on the powered-off screen. InkHUD requires a proper shutdown to save its data, so users will always see the powered-off screen.

InkHUD's powered-off screen


Admittedly, the short name is not displayed *immediately* after flashing, on the *very first* boot, but I don't feel that this is an issue, because the user will need to connect a client app to complete device set-up, and this process will display the device's name to them.

Heltec-Aaron-Lee:
For manufacturers, I think this function is very necessary, because it will help to distinguish devices during mass production testing quickly.

Or change your thinking... Can you consider adding the display of short ID somewhere else?

todd-herbert:
Do you think these changes would suit your needs? I haven't tested it yet myself, but I will do that now to check that the design seems okay.

```cpp
InkHUD::LogoApplet::LogoApplet() : concurrency::OSThread("LogoApplet")
{
    OSThread::setIntervalFromNow(8 * 1000UL);
    OSThread::enabled = true;

    fontTitle = fontSmall;

    // During onboarding, show the default short name as well as the version string
    // This behavior assists manufacturers during mass production, and should not be modified without good reason
    if (settings->tips.safeShutdownSeen) {
        textLeft = xstr(APP_VERSION_SHORT);
        textLeft = owner.short_name;
        textTitle = "Meshtastic"
    } else {
        textLeft = "";
        textRight = "";
        textTitle = xstr(APP_VERSION_SHORT);
    }

    bringToForeground();
    // This is then drawn with a FULL refresh by Renderer::begin
}
```

todd-herbert:
Forgive me, I didn't appreciate that it's an important feature during manufacturing and testing. I was only considering end users at first.

Is this result suitable? When Meshtastic is newly installed, the left image is shown at boot. Once the user selects "Save and Shut Down", the right image is used instead.

Initial Boot | Normal Boot 
---|---
initial startup | normal startup

```cpp
InkHUD::LogoApplet::LogoApplet() : concurrency::OSThread("LogoApplet")
{
    OSThread::setIntervalFromNow(8 * 1000UL);
    OSThread::enabled = true;

    // During onboarding, show the default short name as well as the version string
    // This behavior assists manufacturers during mass production, and should not be modified without good reason
    if (!settings->tips.safeShutdownSeen) {
        fontTitle = fontLarge;
        textLeft = xstr(APP_VERSION_SHORT);
        textRight = owner.short_name;
        textTitle = "Meshtastic";
    } else {
        fontTitle = fontSmall;
        textLeft = "";
        textRight = "";
        textTitle = xstr(APP_VERSION_SHORT);
    }

    bringToForeground();
    // This is then drawn with a FULL refresh by Renderer::begin
}
```

Heltec-Aaron-Lee:
As I tested at my side, this function is pretty good for us, I will merge this part into firmware.

todd-herbert (approved):

PR #6597, opened 2025-04-16
add support for Seeed solar panel

Dylanliacc:
# Add support for seeed solar panel
## Description
Basic board defination for the Seeed Studio Seeed Solar Node

It defines the digital and analog pin mappings, communication interfaces, and peripherals like LoRa, GNSS, LEDs, and battery voltage monitoring. The configuration also sets up the pins for the onboard QSPI flash and includes necessary power management functions for the device.

Additionally, a 5W solar panel is incorporated into the power management setup, providing an energy source for the device.The maximum charging current can be 1A.
### SenseCAP Solar Node P1:
* Features:
1. All-in-one solar energy solution.
2. Directly powered by the solar panel.
3. Embedded XIAO nRF52840 module.
* Power:
1. Type-C connection for charging.
2. Solar panel (5W) for energy generation.

### SenseCAP Solar Node P1-Pro:
* Features:
1. All-in-one solar energy solution.
2. Directly powered by the solar panel.
3. Embedded XIAO nRF52840 module.
* Power:
1. Type-C connection for charging.
2. Equipped with an 18650 battery for additional power storage.
3. XIAO L76K GPS module for positioning functionality.

### Notes
Both variants are designed for long-term outdoor use, offering solar-powered operation and integration with Meshtastic for long-range communication. The 5W solar panel is integrated to ensure continuous power for remote deployments. The main difference between the two variants lies in the P1-Pro having additional battery storage and a GPS module.
There is no need to distinguish the model on the web flasher, just display Seeed Solar Node, same for register system.
## Checklist
 Pin configuration for all peripherals is correct.
 Power management setup has been validated.
 Communication interfaces (I2C, SPI) are initialized correctly.
 GPS L76KB module is configured and tested.
 5W solar panel integration verified.


lib/device-ui
mverch67, line 1: this does not seems correct. This directory has been removed in 2.6.0. 
Maybe you forgot to delete it from your folder after git pull.

caveman99 (approved):

fifieldt:
Hmmm, it looks like the protobufs were deleted rather than removed :) Probably need to re-add them at the right version.

fifieldt:
成功!

fifieldt (approved):

PR #6630, opened 2025-04-19
[WIP] Set GPS to stationary mode when fixed position is enabled

### LINKED ISSUE #6627: [Feature Request]: Stationary GPS mode when fixed position is enabled
### Platform

Cross-Platform

### Description

Some GPS chips have a "stationary" mode, which assumes acceleration and velocity are zero. This mode provides more accurate readings and improves power efficiency. When fixed pos and GPS are enabled, the GPS should start using the stationary position mode if it has it.

fifieldt:
untested WIP - this code may never run, need to check where GPS setup bails out due to fixed_position config.

As reported by @RCGV1, some GPS chips have a "stationary" mode, which assumes acceleration and velocity are zero. For fixed nodes, taking advantage of this configuration means potentially less power usage, and a more accurate position that doesn't jitter its way around the map. (See also: https://github.com/meshtastic/firmware/issues/3984 )

This patch makes changes to the setup code for the the GNSS_MODEL_MTK, GNSS_MODEL_MTK_L76B, GNSS_MODEL_ATGM336H, GNSS_MODEL_AG3335, GNSS_MODEL_AG3352 to apply that approach.

The aim would be to enable this mode when users have a GPS enabled and select "Fixed Position" without specifying coordinates. UX TBD!

Fixes https://github.com/meshtastic/firmware/issues/6627

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -547,8 +547,13 @@ bool GPS::setup()
             // only ask for RMC and GGA
             _serial_gps->write("$PCAS03,1,0,0,0,1,0,0,0,0,0,,,0,0*02\r\n");
             delay(250);
-            // Switch to Vehicle Mode, since SoftRF enables Aviation < 2g
-            _serial_gps->write("$PCAS11,3*1E\r\n");
+            if (config.position.fixed_position == true) {
+                // Set to stationary mode, which reduces jitter and may save power
+                _serial_gps->write("$PCAS11,1*1C\r\n");
+            } else {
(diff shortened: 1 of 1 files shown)


garthvh:
Fixed position does not generally use the gps at all, I would suggest the stationary mode should not need any UX at all and instead not use power or generated positions all over the place when the gps detects it is stationary.

PR #6631, opened 2025-04-19
docker alpine: Add config templates

vidplace7:
Add config templates for Docker Alpine images, just like the Debian images we provide.

Thanks @ianmcorvidae for pointing this out.


ianmcorvidae (approved):
Looks right to me, at least!

Issue #6598, opened 2025-04-16
[Bug]: NicheGraphics configuration clearing function is missing.


Quency-D:
### Category

Other

### Hardware

Heltec Vision Master E213

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [x] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.6.6

### Description

When using an Android app for factory reset, NicheGraphics related information was not restored to the factory configuration.
From the code, it can be seen that there is only clearing processing related to nodeDB.

```c
    case meshtastic_AdminMessage_factory_reset_device_tag: {
        disableBluetooth();
        LOG_INFO("Initiate full factory reset");
        nodeDB->factoryReset(true);
        reboot(DEFAULT_REBOOT_SECONDS);
        break;
    }
    case meshtastic_AdminMessage_nodedb_reset_tag: {
        disableBluetooth();
        LOG_INFO("Initiate node-db reset");
        nodeDB->resetNodes();
        reboot(DEFAULT_REBOOT_SECONDS);
        break;
    }

### Relevant log output

```Shell

```

PR #6637, opened 2025-04-20
Restore InkHUD to defaults on factory reset

todd-herbert:
Resolves https://github.com/meshtastic/firmware/issues/6598
With this PR, a factory reset now restores InkHUD to its default state.


fifieldt:
yay!
I know nothing about this stuff, but just thought I'd ask - any new SPI lock or unlock required?

todd-herbert:
Oh, quite possibly! I don't think it's implemented at all for the InkHUD stuff actually. So far, all of the displays use a dedicated SPI bus, and none of them use an SD card, so no problems have come up, but I'll definitely look into incorporating it in the near future.

todd-herbert:
Actually I should probably ask: will we need to lock the bus at all? InkHUD only does Arduino Threads time-sharing, so there's no multi-core stuff to account for. Any information here would be helpful!

GUVWAF:
For single-core ArduinoThread multitasking only, I believe indeed we wouldn't need to lock.

todd-herbert:
Oh, looks like the CI build isn't happy about one of the `#include` statements. One moment!

fifieldt (approved):

PR #6650, opened 2025-04-22
Added flight mode GPS build flag

RCGV1:
I have implemented a build flag for flight mode on the ATGM336H and the ublox line of gps modules. I have flown these modules before on high altitude balloons (up to 30km) and position worked perfectly, I copied the exact same flight mode code I used on these flights and added a build flag to optionally use them. I compiled and ran on a heltec v3 to ensure no obvious issues occurred but I have not actually tested if flight mode turns on (it should)

## 🤝 Attestations
  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


src/gps/cas.h
```diff
@@ -61,3 +61,26 @@ static const uint8_t _message_CAS_CFG_NAVX_CONF[] = {
     0x00, 0x00, 0x00, 0x00, // Time Accuracy Max
     0x00, 0x00, 0x00, 0x00  // Static Hold Threshold
 };
+
```
fifieldt, line 64: Since this is almost identical to the one before, you might want to consider changing just the mode value in the setup. 

See https://github.com/meshtastic/firmware/pull/6630/files#diff-c9e9d981789c05f2633c8bc91527d5c33f714ea43d106684250598ec0d4487ccL587 where the vehicle mode is changed to stationary.

thebentern:
Looks good, but there is a protobufs ref change that tagged along as well.

RCGV1:
@thebentern can we open this again

RCGV1:
@fifieldt

fifieldt:
Needs: https://github.com/meshtastic/protobufs/pull/749

PR #6653, opened 2025-04-22
Fix preamble detected IRQ flag

GUVWAF:
As reported by Discord user johntitor, #4621 caused the `PREAMBLE_DETECTED` IRQ flag not to be set anymore. The generic IRQ flags are an enum representing a bit position and hence must be shifted, instead of being the mask already. We got "lucky" that bit 2 is actually `RX_DONE`, which was set already.

Since those are now generic flags, I added the definition in RadioLibInterface. Confirmed that SX126x radio now occasionally shows "Ignore false preamble detection" again.
LR11x0 has all flags enabled by default, so it doesn't actually affect that radio.

## 🤝 Attestations
  - [X] LilyGo T-Echo

### DIFF
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -247,8 +247,8 @@ template <typename T> void LR11x0Interface<T>::startReceive()
     lora.setPreambleLength(preambleLength); // Solve RX ack fail after direct message sent.  Not sure why this is needed.
 
     // We use a 16 bit preamble so this should save some power by letting radio sit in standby mostly.
-    // Furthermore, we need the PREAMBLE_DETECTED and HEADER_VALID IRQ flag to detect whether we are actively receiving
-    int err = lora.startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF, RADIOLIB_IRQ_RX_DEFAULT_FLAGS, RADIOLIB_IRQ_RX_DEFAULT_MASK, 0);
+    int err =
+        lora.startReceive(RADIOLIB_LR11X0_RX_TIMEOUT_INF, MESHTASTIC_RADIOLIB_IRQ_RX_FLAGS, RADIOLIB_IRQ_RX_DEFAULT_MASK, 0);
     assert(err == RADIOLIB_ERR_NONE);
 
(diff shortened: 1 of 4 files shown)


thebentern (approved):

mangelajo:
This doesn't seem to fix it for me on T1000E + Lilygo T-NODE , at least on the EU_868 band.

mangelajo:
> This doesn't seem to fix #4775 for me on T1000E + Lilygo T-BEAM , at least on the EU_868 band.

I have also verified that it works with a Heltec T114, but not with the lilygo based ones
image

In that traceroute 8bce is the T1000-E , Aj0_Movil is a T-TBEAM, Ester is the T-114

so T1000-E to T-BEAM works, T-BEAM to T1000-E doesn't. I am going to double check that this commit is on the master branch I compiled.

mangelajo:
Confirmed, it was compiled with the patch in 
```
➜  firmware git:(master) ✗ git log  | grep "Fix preamble detected IRQ flag" -B 4
commit b1e35cd8b36a57098184491dc3841d83ff65ec76
Author: GUVWAF <78759985+GUVWAF@users.noreply.github.com>
Date:   Tue Apr 22 21:21:29 2025 +0200

    Fix preamble detected IRQ flag (#6653)
```

And the device reports version: "v2.6.9.beba1b48"

GUVWAF:
Yes, this PR is not a solution to #4775, I believe that issue can only be solved by changing the sync word, which is a breaking change. It's expected that indeed the LR11xx radios can send towards SX127x radios, but not vice versa.

This PR only affects channel activity detection.

GUVWAF:
@spydercs Not never, but only once there is a breaking change to the firmware, i.e. Meshtastic 3.0.

mangelajo:
@GUVWAF should then we may be keep #4775 still open for 3.0 tracking purposes, the issue was closed from this PR, but if it's not a fix, I guess it may remain open?

GUVWAF:
Yes, good point, I've re-opened it.

serialrf433:
https://github.com/meshtastic/firmware/issues/4775 is not really open again. Its telling currently: This conversation has been locked and limited to collaborators.

PR #6661, opened 2025-04-24
Support sub-directories when serving the web client from ESP32

todd-herbert:
Proof of concept workaround for the limitation that esp32_https_server doesn't traverse sub-directories.
@caveman99 is intending to address this one I believe. Submitting in case this draft can be a useful starting point.

Not required as a fix, as the web client's directory structure is now flattened again.

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [x] LilyGo T-Beam *(v1.2)*
  - [ ] RAK WisBlock 4631


todd-herbert:
@danditomaso is quite keen to explore this option further. It sounds like flattening the web client is turning into more of a hassle than initially suspected.

todd-herbert:
Sounds like the whole library might be swapped out anyway, which would hopefully resolve all this.
https://github.com/meshtastic/firmware/issues/6666#issuecomment-2850881021

todd-herbert:
Decision reached that this will be resolved by flattening web ui directory structure instead.

PR #6673, opened 2025-04-25
Step one of Linux Sensor support

jp-bennett:
Very WIP. Attempting to make firmware compile with modules/Telemetry/* included in the Native build.


markbirss (approved):
Was able to get readings for BME280 and BMP280
image
image

PR #6685, opened 2025-04-29
udp-multicast: remove the thread from the multicast thread API

Jorropo:
The whole API is parallel & asynchronous we don't need to start a thread ourself,
the implementation probably does when we call start listening already.


thebentern:
I think this _should_ work.

fifieldt (approved):
Given the stateless nature of UDP, this makes logical sense to me.

PR #6695, opened 2025-04-30
Add 100 msecond delay in tft_task_handler when deviceScreen is null

jp-bennett:
This fixes the 100% device usage bug observed on Native Linux


mverch67:
Great find. 
Other option would be to not create the task when there's no screen. E.g. I'd prefer to move the task creation for Linux from it's main.cpp into tftSetup.cpp so that it's in one place.

image

mverch67 (approved):
Ready to merge

thebentern (approved):

PR #6714, opened 2025-05-01
Update ICM 20948 to latest release, move out of extra

jp-bennett:
My fix for PORTDUINO landed in the upstream 20948 library, and they already bumped the version. So update here.


fifieldt (approved):

thebentern:
We probably need to go ahead and add renovate annotations for these

PR #6720, opened 2025-05-02
Update template for event userprefs

vidplace7:
Small change to the defaults in the userprefs.jsonc template.

- ~Change preset from ShortFast to ShortTurbo.~
- Populate common names / PSKs for secondary channels.
  - Please suggest better names :joy:


fifieldt:
is it OK to default to a preset that isn't legal everywhere?

(no opinion, just asking in case you do have one)

vidplace7:
> is it OK to default to a preset that isn't legal everywhere?
> 
> (no opinion, just asking in case you do have one)

Very good point :+1: I've set the preset back to Short_Fast in the example config here.

fifieldt (approved):

thebentern (approved):

PR #6735, opened 2025-05-05
Protect T-Echo's touch button against phantom presses in OLED UI

todd-herbert:
By request of a Discord user, this PR extends the #6590 fix from InkHUD to the OLED UI.
Ignores touch button presses which occur during LoRa transmission.

> Hi, I just read through the changes in src/mesh/RadioLibInterface.h and wonder: does  that only disable the backlight, or the entire touch button, during LoRa TX ? The latter would be really handy, since some T-Echo (one out of my two) have this bug where an ACK can get missing when a (TX-enabled) touch button activation  activates the E-Ink display (full) REFRESH which hampers the next TX that would ACK the next message after  (if I remember the causality correctly)

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - LilyGo T-Echo (OLED UI)


fifieldt (approved):

thebentern:
Can we macro guard this fix to the T-Echo? I'm worried this could become an annoying problem on super busy networks like events.

todd-herbert:
>  I'm worried this could become an annoying problem on super busy networks like events.

I'll absolutely add any guards you think would be helpful, but I'm not quite sure where you mean.. sorry..
I think I must be missing something obvious here, can you point me at the problem area?

caveman99:
> I'll absolutely add any guards you're after, but I'm not quite sure where you mean.. sorry.. I think I must be missing something obvious here, can you point me at the problem area?

I think the idea is to add a macro guard to the sourcecode so the change only applies to actual T-Echo devices where the cross-talk is a problem. 

```
#ifdef TTGO_T_ECHO
[...]
#endif
```

todd-herbert:
I'm still thinking I might be missing something obvious because I don't see anywhere that macro guards would make much difference. I'll add them anyway though when I'm back at my computer in a bit.

thebentern (approved):

PR #6759, opened 2025-05-07
Add client notification before role based power saving (sleep)

thebentern:


fifieldt:
The millisecond vs seconds looks good.

PR #6763, opened 2025-05-08
Update Seeed Solar Node

rcarteraz:
lowercase


vidplace7 (changes requested):
You should rename the json file in `boards/`, and the directory in `variants/` as well :pray:

thebentern (approved):

caveman99:
@rcarteraz your patch changes the board name, not the env name. If you change the board name, the json file needs to be named accordingly, this is matched 1:1. The directory name is a 'nice to have' to stay in line.

PR #6792, opened 2025-05-11
Only send nodes on want_config of 69421

### LINKED ISSUE #4564: [Feature Request]: Admin message to request a refresh of the nodedb
### Platform

### Description

Now that we have this in place https://github.com/meshtastic/firmware/issues/3740

We need to add a admin message that will refresh the node db, so that apps can separate connecting and the nodedb refresh, which takes a long time with 100 nodes.

thebentern:
Closes #4564

@garthvh you should be able to just issue a special want_config of 69421 and receive just the nodeinfo

### DIFF
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -54,7 +54,13 @@ void PhoneAPI::handleStartConfig()
     }
 
     // even if we were already connected - restart our state machine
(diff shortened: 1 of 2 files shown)


garthvh (commented):
Awesome, no protobufs required!

garthvh (approved):

PR #6798, opened 2025-05-12
Add DIY PCB

MiszczNadMiszcze:
DIY PCB for ProMicro nRF52840 and Heltec HT-RA62
https://pvv.pl/kiler

### DIFF
--- a/variants/rak_killer_v2/platformio.ini
+++ b/variants/rak_killer_v2/platformio.ini
@@ -1,0 +1,16 @@
+; SuperMini nRF52840 + Heltec HT-RA62 + DIY Breakout PCB
+[env:rak_killer_v2]
+extends = nrf52840_base
+board = promicro-nrf52840
+; board_level = extra
+debug_tool = jlink
+
+# add -DCFG_SYSVIEW if you want to use the Segger systemview tool for OS profiling.
+build_flags = ${nrf52840_base.build_flags}
+        -I variants/rak_killer_v2
+        -D PRIVATE_HW
+        -L "${platformio.libdeps_dir}/${this.__env__}/bsec2/src/cortex-m4/fpv4-sp-d16-hard"
+
+build_src_filter = ${nrf52_base.build_src_filter} +<../variants/rak_killer_v2> 
+lib_deps = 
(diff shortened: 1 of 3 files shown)


caveman99:
Two Observations:

- we already have a ton of Promicro based DIY designs. Please describe how your design is unique and better than the other designs we have. Are you intending to support this variant or is this a one shot?
- Please don't mention other vendors in your device name, especially in a derogatory manner. Nobody is getting killed around here.

MiszczNadMiszcze:
> Two Observations:
> 
>     * we already have a ton of Promicro based DIY designs. Please describe how your design is unique and better than the other designs we have. Are you intending to support this variant or is this a one shot?
> 
>     * Please don't mention other vendors in your device name, especially in a derogatory manner. Nobody is getting killed around here.

It is not better or worse, it is used to broaden the meshtastic network. Developed for seven months with small amendments, the last received a built -in watchdog which is tested and works.

The name rak_killer refers not to the killer but the conqueror. Mainly when it comes to price.

caveman99:
Sorry, i compared your definition file with the stock promicro one, and it's basically the same with some in assigments swapped around. We value your contribution, but also need to keep the firmware repository manageable. Also i don't see any watchdog code here, and even if there was, we only include this kind of thing reluctantly, rather fix the underlying bugs to make the watchdog superfluous.

MiszczNadMiszcze:
> Sorry, i compared your definition file with the stock promicro one, and it's basically the same with some pin assigments swapped around. We value your contribution, but also need to keep the firmware repository manageable. Also i don't see any watchdog code here, and even if there was, we only include this kind of thing reluctantly, rather fix the underlying bugs to make the watchdog superfluous.

I haven't seen people with such an exuberant ego.
Instead of developing meshtastich with your decisions.
Do you think I will give you everything on a tray?
WDT works, you can check
https://github.com/Uroboros67/rak_killer/blob/49acdcaa0d111f5797c80c960346030fbb8975f5/ver_2/firmware-2.6.7.2d6181f_WDT.uf2
NRF52 has problems with and without WDT there is no chance to act, I see that we must act in our own group :)

PR #6803, opened 2025-05-12
Add suppport for Quectel L80

fifieldt:
Another PMTK family chip, requires only a modification to the probe code.


randybb:
Many thanks, it is finally detecting L80-R (I know, it is an ancient one, but for a fixed node it is ok):
```
INFO  | 10:10:44 3 [GPS] L80-R detected
```

Sirtxm:
Can i use GPS files in ArduinoIDE?

PR #6813, opened 2025-05-14
Added new map report opt-in for compliance and limit map report (and default) to one hour

thebentern:


Code1110:
This should be reflected in the docs as well. https://meshtastic.org/docs/configuration/module/mqtt/ still says the default is 900 seconds and doesn't mention that the minimum is now 3600 seconds. Entering a value smaller than 3600 seconds in the UI changes the value silently to 3600 seconds and is therefore confusing.

cpatulea:
FYI some potentially related user confusion here: https://github.com/meshtastic/firmware/issues/9017

PR #6819, opened 2025-05-15
Fix run out of heap on ESP32

nikito7:
## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

ESP32 devices.

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -232,7 +232,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #endif
 
 #ifndef MINIMUM_SAFE_FREE_HEAP
+#ifdef ESP32
 #define MINIMUM_SAFE_FREE_HEAP 15000
+#else
+#define MINIMUM_SAFE_FREE_HEAP 1500
+#endif
 #endif
 
 #ifndef WIRE_INTERFACES_COUNT
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -232,7 +232,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #endif
 
 #ifndef MINIMUM_SAFE_FREE_HEAP
-#define MINIMUM_SAFE_FREE_HEAP 1500
+#define MINIMUM_SAFE_FREE_HEAP 15000
 #endif
 
 #ifndef WIRE_INTERFACES_COUNT


nikito7:
It may only be used here:

```
// returns true if the maximum number of nodes is reached or we are running low on memory
bool NodeDB::isFull()
{
    return (numMeshNodes >= MAX_NUM_NODES) || (memGet.getFreeHeap() < MINIMUM_SAFE_FREE_HEAP);
}
```
https://github.com/meshtastic/firmware/blob/066609a7189163f607fee5966f223526c30bd7d1/src/mesh/NodeDB.cpp#L1638

If 1500 default it's all free heap, it's bad.
ESP32 below 15kb could crash.

This it's only for nodes. I don't know if impact anywhere.

Because my true solution for my problem was:

```
  -D MESHTASTIC_EXCLUDE_BLUETOOTH=1
  -D MESHTASTIC_EXCLUDE_AUDIO=1
  -D MESHTASTIC_EXCLUDE_PAXCOUNTER=1
  -D MESHTASTIC_EXCLUDE_RANGETEST=1
  -D MESHTASTIC_EXCLUDE_ATAK=1
  -D MINIMUM_SAFE_FREE_HEAP=25000
```

My current it's about 90kB now.
With default firmware it's always below 20kB.
And wifi hangs or esp32 reboot.

Screenshot_20250515_083505_Home Assistant

mverch67:
I don't think the proposed solution is reasonable. Instead of changing the global value (which is also used for e.g. nrf devices that do not have much memory) you can limit the use to specific devices in the platformio.ini files.
Also, this value has been used for months or even years without any major problems and now you suggest the ten-fold value which seems quite exaggerated. Did you try other values like 2000 or 2500, 3000, and so on?

nikito7:
This PR it's mainly a proof of concept. 
If you don't agree, it's ok.
I don't have the perfect solution.
The main problem it's Bluetooth itself.

If someone have issues, adapt this for your needs:

```
[han32opt]
build_flags =
  -D PRIVATE_HW
  -D DEFAULT_REBOOT_SECONDS=3
  -D MESHTASTIC_EXCLUDE_PAXCOUNTER=1
  -D MESHTASTIC_EXCLUDE_RANGETEST=1
  -D MESHTASTIC_EXCLUDE_ATAK=1
  -D MESHTASTIC_EXCLUDE_STOREFORWARD=1

[env:han32gw-no-ble]
extends = esp32_base
board = esp32dev
board_check = true
build_flags = 
  ${esp32_base.build_flags}
  ${han32opt.build_flags}
  -I ../variants/han32gw
  -D MESHTASTIC_EXCLUDE_BLUETOOTH=1
  -D MINIMUM_SAFE_FREE_HEAP=25000
  -D MAX_NUM_NODES=100
upload_speed = 460800

[env:han32gw-ble]
extends = esp32_base
board = esp32dev
board_check = true
build_flags = 
  ${esp32_base.build_flags}
  ${han32opt.build_flags}
  -I ../variants/han32gw
  -D MESHTASTIC_EXCLUDE_BLUETOOTH=0
  -D HAS_BLUETOOTH=1
  -D MINIMUM_SAFE_FREE_HEAP=15000
  -D MAX_NUM_NODES=100
upload_speed = 460800

; EOF
```

nikito7:
A last comment. My ble firmware stills hangs wifi connection after a while. I need to reboot Acess Point.

Without BLE stack it's perfect.

PR #6822, opened 2025-05-15
Portduino: Don't filter the traditional UI

vidplace7:
Portduino: Do not filter out (exclude) the traditional UI on Linux.

Display selection should be left to the user in the yaml, not taken away at build time.

DRAFT compiles correctly but untested on real hardware.


jp-bennett:
A much bigger version of this is #6838

PR #6827, opened 2025-05-16
Fix for ICM-20948 not initializing

HarukiToreda:
Problem:
The ICM-20948 Accelerometer was failing to initialize on boot on the Heltec T114, T3S3, and Heltec V3 but worked fine on Rak4631

image

Cause:
The code was trying to figure out if the board supports a second I2C port (Wire1), but it was checking the wrong way. That caused it to pick the wrong I2C port, so the sensor couldn’t communicate properly on some boards.

Fix:
I changed the code to correctly detect which I2C port the board supports and use the one the sensor is actually connected to. After this change, the accelerometer initializes properly on all boards.

Tested on all Heltec T114, T3S3, and Heltec V3 and RAK4631

### DIFF
--- a/src/motion/ICM20948Sensor.cpp
+++ b/src/motion/ICM20948Sensor.cpp
@@ -190,12 +190,18 @@ bool ICM20948Singleton::init(ScanI2C::FoundDevice device)
 #endif
 
 // startup
-#ifdef Wire1
-    ICM_20948_Status_e status =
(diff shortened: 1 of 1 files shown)


tropho23:
Tested on RAK461, T3S3, T114, and works!

fifieldt (approved):

PR #6838, opened 2025-05-17
Unify the native display config between legacy display and MUI

jp-bennett:
WIP. I specifically don't understand why the OffsetRotate needs to be different between MUI and the legacy display.

### DIFF
--- a/src/graphics/TFTDisplay.cpp
+++ b/src/graphics/TFTDisplay.cpp
@@ -666,6 +666,8 @@ class LGFX : public lgfx::LGFX_Device
   public:
     LGFX(void)
     {
+        int _width = 0;
+        int _height = 0;
         if (settingsMap[displayPanel] == st7789)
             _panel_instance = new lgfx::Panel_ST7789;
         else if (settingsMap[displayPanel] == st7735)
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -769,6 +761,12 @@ void setup()
     // but we need to do this after main cpu init (esp32setup), because we need the random seed set
     nodeDB = new NodeDB;
 
+#if HAS_TFT
+    if (config.display.displaymode == meshtastic_Config_DisplayConfig_DisplayMode_COLOR) {
+        tftSetup();
```
mverch67, line 766: I saw that you moved the `tftSetup() `further down from line 542 by 200 lines which delays the initialization and prolongs the boot process. I actually moved it up a while ago (after sorting out the SPI issues) so that just after switching the device on the screen turn on as well...

jp-bennett: > I saw that you moved the `tftSetup() `further down from line 542 by 200 lines which delays the initialization and prolongs the boot process. I actually moved it up a while ago (after sorting out the SPI issues) so that just after switching the device on the screen turn on as well...

Yeah, unfortunately it now *must* happen after nodeDB initialization, as that's where the config values get loaded.

mverch67:
I've adapted, built, and tested the other TFT targets:
- Indicator
- PICOmputer

Also, MUI can revert to former UI.

src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -124,10 +124,14 @@ static void onNetworkConnected()
...
 
 #if defined(ARCH_ESP32) && !MESHTASTIC_EXCLUDE_WEBSERVER
-        initWebServer();
+        if (config.display.displaymode != meshtastic_Config_DisplayConfig_DisplayMode_COLOR) {
+            initWebServer();
+        }
 #endif
 #if !MESHTASTIC_EXCLUDE_SOCKETAPI
-        initApiServer();
+        if (config.display.displaymode != meshtastic_Config_DisplayConfig_DisplayMode_COLOR) {
```
mverch67, line 132: I'm wondering about this line. Would it still be possible to
a) use MQTT
b) attach remote MUI to _meshtasticd_ when set to COLOR, but no X11/FB Panel defined in config.yaml

Same above for webserver.

mverch67 (changes requested):

PR #6842, opened 2025-05-19
Unify tft

jp-bennett:
This is a WIP effort to unify the builds for MUI targets, so the use of the MUI can be a runtime config option, currently config.display.displaymode == meshtastic_Config_DisplayConfig_DisplayMode_COLOR. It's currently the native-tft target and the t-deck, but likely still contains crashes and potential bugs.


jp-bennett:
Closing in favor of the other draft PR... that i forgot I had already opened.

PR #6862, opened 2025-05-21
Added map report precision bounds

### LINKED ISSUE #6860: [Compliance]: Disallow precise position MapReports
**Platform:** Cross-Platform

**Description:**
Update the use of precise position MapReports. This change is required to improve privacy and security for all users by ensuring that precise location data is not shared unencrypted over mqtt.

thebentern:
Closes #6860


garthvh (changes requested):
I think 12-15 is a better range 16 is .2 miles and the CCPA 1850 feet minimum is .35 miles.

12 is 3.6 miles, 11 is almost 8 which is too big from a UX perspective in a lot of places

garthvh (approved):

PR #6872, opened 2025-05-23
Added full support for LTR390UV readings of UV and Lux

### LINKED ISSUE #5980: [Bug]: The LTR390V sensor is not identified
### Category

Hardware Compatibility, Other

### Hardware

Rak4631

### Firmware Version

2.5.20.4c97351

### Description

The LTR390V sensor is not identified. Protobuf supports it, but I2C does not identify it.
Its address is found at 0x53, see the log excerpt.

### Relevant log output

```Shell
INFO  | ??:??:?? 1 MAX17048 found at address 0x36
INFO  | ??:??:?? 1 SSD1306 found at address 0x3c
INFO  | ??:??:?? 1 SSD1306 found at address 0x3c
DEBUG | ??:??:?? 1 0x4 subtype probed in 2 tries 
INFO  | ??:??:?? 1 Device found at address 0x53 was not able to be enumerated
INFO  | ??:??:?? 1 BME680 found at address 0x77
INFO  | ??:??:?? 1 3 I2C devices found
```

dmarman:
This PR fixes #5980 and gives full support at the firmware level for reading UV and Lux values with the sensor LTR390UV. 
The hardware sensor has been tested and the values compared with online weather stations in sunny Barcelona.

The different UI clients must update to show the UV reading. Currently with Android app, UV values will not show up however Lux readings do appear.

Due to the nature of the sensor. Lux and UV readings must be done in two passes. So for the clients, UV and Lux values are updated every two minutes.

## 🤝 Attestations
  - [X] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


src/modules/Telemetry/EnvironmentTelemetry.cpp
```diff
@@ -52,6 +52,13 @@ BMP280Sensor bmp280Sensor;
 NullSensor bme280Sensor;
 #endif
 
+#if __has_include(<Adafruit_BMP280.h>)
```
caveman99, line 55: That looks like a copy/paste error to me?

src/modules/Telemetry/Sensor/LTR390UVSensor.cpp
```diff
@@ -0,0 +1,69 @@
...
+    ltr390uv.setResolution(LTR390_RESOLUTION_20BIT); // Datasheet default
+
+    return initI2CSensor();
+}
+
+void LTR390UVSensor::setup() {}
+
+bool LTR390UVSensor::getMetrics(meshtastic_Telemetry *measurement)
+{
+    measurement->variant.environment_metrics.has_lux = true;
```
caveman99, line 32: getMetrics is only called Once per interval. Since you are adding an IF statement you will get either the lux or the UV value. Not both. You either need to add a loop to this function or make sure you only transmit the valid part of the payload and not set both. Also constantly changing the parameters seems expensive to me. is this really neccessary? at least the resolution seems to stay the same.

caveman99 (changes requested):
please review the logic flow as detailed.

caveman99:
Also could you sign the CLA?

src/modules/Telemetry/EnvironmentTelemetry.cpp
dmarman: ups, I'll fix it.

src/modules/Telemetry/Sensor/LTR390UVSensor.cpp
dmarman: I'll remove the resolution settings as it does not change. The problem with this sensor is that it needs different gain settings for the UV and the Lux readings (based on sunlight testing). And those readings cannot be done simultaneously. 

Also, the UV reading with the characterized settings on the datasheet takes 400ms and Lux 25ms. So doing those readings in a single interval will take longer than 425ms, which I thought it is too much but maybe it is ok?
If doing like in the PR, those 425ms are not expended in the function call, but in between the 60 second interval. The sensor ADC stores the value in its register and then we can read every time a fresh new reading without waiting.

I had to set both values because if we don't do that, the app will not show both readings at the same time. 

Another option is to lower the measurement rate and resolution to get measurements of 2 * 25ms = 50ms on each interval call, maybe that's also too expensive? We should consider that those faster settings are not characterized in the datasheet, and we will have to come up with a magic number without knowing if we are loosing precision.

A secondsoption is to only read UV and forget about Lux as this is the only UV sensor compatible...

caveman99: I'd suggst adding caching for the 'other' value and always populate both. Then alternate reading the values every 60 seconds. Network transmits are not every 60 seconds, but much less frequently anyway, so we can make sure these go out with the 2 most current values.

src/modules/Telemetry/Sensor/LTR390UVSensor.h
```diff
@@ -0,0 +1,25 @@
...
+
+class LTR390UVSensor : public TelemetrySensor
+{
+  private:
+    Adafruit_LTR390 ltr390uv = Adafruit_LTR390();
+    float lastLuxReading = 0;
+    float lastUVReading = 0;
+
+  protected:
+    virtual void setup() override;
```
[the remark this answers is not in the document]
caveman99: That's by design, don't complain :-)

src/modules/Telemetry/Sensor/LTR390UVSensor.cpp
caveman99: Also remember if your phone is not connected, these 60 seconds reads are not happening.

caveman99: You could alternatively implement asynchronous sampling onf the values like we do for the BME680.

dmarman:
Guys sorry I don't know what the hell I am doing with git that I am messing everything up. I signed the CLA but my email is not showing up in the commit I don't know why. I followed the vague instructions from the CLA comment, but it didn't work.

caveman99:
> Guys sorry I don't know what the hell I am doing with git that I am messing everything up. I signed the CLA but my email is not showing up in the commit I don't know why. I followed the vague instructions from the CLA comment, but it didn't work.

It's fine, one of the nicks signed, that's enogh. thanks :-)

caveman99:
Though it looks like your latest commit added some dupes, if that's not intentional.

src/modules/Telemetry/Sensor/LTR390UVSensor.cpp
caveman99: That looks most excellent to me now. And by the way, ignore the remaining Copilot stuff... he's a bit of a Monk...

src/modules/Telemetry/Sensor/LTR390UVSensor.cpp
```diff
@@ -0,0 +1,64 @@
...
+            measurement->variant.environment_metrics.lux = lastLuxReading;
+            measurement->variant.environment_metrics.uv_lux = lastUVReading;
+
+            ltr390uv.setGain(
+                LTR390_GAIN_1); // x1 gain will already max out the sensor at direct sunlight, so no need to increase it
+            ltr390uv.setMode(LTR390_MODE_ALS);
+        }
+    }
+
+    return true;
```
caveman99, line 62: Copilot had a point there about the variable scope. Please move setting the 'has' fields to true to the conditionals. Also GetMetrics should only return TRUE if there was an actual reading.

caveman99 (changes requested):

src/modules/Telemetry/Sensor/LTR390UVSensor.cpp
```diff
@@ -56,9 +59,11 @@ bool LTR390UVSensor::getMetrics(meshtastic_Telemetry *measurement)
             ltr390uv.setGain(
                 LTR390_GAIN_1); // x1 gain will already max out the sensor at direct sunlight, so no need to increase it
             ltr390uv.setMode(LTR390_MODE_ALS);
+
+            return true;
         }
     }
 
-    return true;
+    return false;
```
caveman99, line 67: can we set has_lux and has_v_lux to false here too? i think we're good after that change.

caveman99 (changes requested):
finish line here :-)

thebentern (approved):

ArgoNavi:
Bit late but would it be possible to detect if another light sensor is connected and use that for lux instead of the LTR390UV?

I plan to use both a LTR390UV and TSL2591 for sun measurements.

PR #6879, opened 2025-05-24
graphics: Add GDEY0213B74 E-Ink display driver

chihosin:
Implement the GDEY0213B74 driver with configuration methods for scanning, waveform, update sequence, and polling for refresh completion. This driver supports both fast and full update types for the 2.13 inch E-Ink display.

Signed-off-by: ChihoSin chihosin@icloud.com

  - [x] RAK WisBlock 4631 with E-Ink Display GDEY0213B74, build with InkHUD

### DIFF
--- a/src/graphics/niche/Drivers/EInk/GDEY0213B74.cpp
+++ b/src/graphics/niche/Drivers/EInk/GDEY0213B74.cpp
@@ -1,0 +1,61 @@
+#include "./GDEY0213B74.h"
+
+#ifdef MESHTASTIC_INCLUDE_NICHE_GRAPHICS
+
+using namespace NicheGraphics::Drivers;
+
+// Map the display controller IC's output to the connected panel
+void GDEY0213B74::configScanning()
+{
+    // "Driver output control"
+    sendCommand(0x01);
+    sendData(0xF9);
+    sendData(0x00);
+    sendData(0x00);
+
+    // To-do: delete this method?
+    // Values set here might be redundant: F9, 00, 00 seems to be default
+}
+
+// Specify which information is used to control the sequence of voltages applied to move the pixels
+// - For this display, configUpdateSequence() specifies that a suitable LUT will be loaded from
+//   the controller IC's OTP memory, when the update procedure begins.
+void GDEY0213B74::configWaveform()
+{
+    sendCommand(0x3C); // Border waveform:
+    sendData(0x05);    // Screen border should follow LUT1 waveform (actively drive pixels white)
+
+    sendCommand(0x18); // Temperature sensor:
+    sendData(0x80);    // Use internal temperature sensor to select an appropriate refresh waveform
+}
+
+void GDEY0213B74::configUpdateSequence()
+{
+    switch (updateType) {
+    case FAST:
+        sendCommand(0x22); // Set "update sequence"
+        sendData(0xFF);    // Will load LUT from OTP memory, Display mode 2 "differential refresh"
+        break;
+
+    case FULL:
+    default:
+        sendCommand(0x22); // Set "update sequence"
+        sendData(0xF7);    // Will load LUT from OTP memory
+        break;
+    }
+}
+
+// Once the refresh operation has been started,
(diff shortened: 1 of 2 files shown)


src/graphics/niche/Drivers/EInk/GDEY0213B74.h
```diff
@@ -0,0 +1,42 @@
+/*
+
+E-Ink display driver
+    - GDEY0213B74
+    - Manufacturer: Goodisplay
+    - Size: 2.13 inch
+    - Resolution: 250px x 122px
+    - Flex connector marking: FPC-B001
```
todd-herbert, line 8: Can I just confirm whether this is the correct marking? Trying my hardest to keep these accurate to help users identify which driver they need.

src/graphics/niche/Drivers/EInk/GDEY0213B74.cpp
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Once the refresh operation has been started,
+// begin periodically polling the display to check for completion, using the normal Meshtastic threading code
+// Only used when refresh is "async"
+void GDEY0213B74::detachFromUpdate()
+{
+    switch (updateType) {
+    case FAST:
+        return beginPolling(50, 500); // At least 500ms for fast refresh
```
todd-herbert, line 55: Are these values optimized, or just reasonable defaults? No problem either way, just good to know for future.

todd-herbert (commented):
Thanks for adding this! Great to see some more displays supported.

src/graphics/niche/Drivers/EInk/GDEY0213B74.h
chihosin: Yes, this is the correct model information. The screen I used for development came from https://www.good-display.com/product/391.html

src/graphics/niche/Drivers/EInk/GDEY0213B74.cpp
chihosin: Only reasonable default values are used. There are many different versions of the B74 E-Ink variant on the market. Development was based on the driver from GoodDisplay: https://v4.cecdn.yun300.cn/100001_1909185148/AU-GDEY0213B74.rar

src/graphics/niche/Drivers/EInk/GDEY0213B74.h
todd-herbert: With FPC-B001?

src/graphics/niche/Drivers/EInk/GDEY0213B74.cpp
todd-herbert: That's fair enough. Probably it's tricky to optimize too much for this display anyway as I imagine different refresh waveforms are used depending on temperature.

src/graphics/niche/Drivers/EInk/GDEY0213B74.h
chihosin: It seems they used different marked in different batches, but their FPC specifications are consistent. Perhaps declaring the FPC marked in the documentation isn't a good idea.
image

todd-herbert: Thank you for checking for me! The result was definitely interesting. 

The flex connector marking is not a perfect identifier, but sometimes it's a very helpful clue for individual users building DIY hardware. 

You make a good point though: it's probably important for us to be more clear in the documentation that the connector marking is not a guaranteed id.

todd-herbert (approved):

src/graphics/niche/Drivers/EInk/GDEY0213B74.cpp
todd-herbert: Only meant that it seems this display model will automatically pick the correct waveform from OTP using the internal temperature sensor, and that the exact duration of this will vary, so there's no point for us to anticipate precisely how long we need to wait before we check for completion.

PR #6892, opened 2025-05-26
Key verification implementation

jp-bennett:
This adds the KeyVerification module, which gives us a few interesting things.
The first is an interesting security primitive I'm tongue-in-cheekily calling MOTP, Mesh-based One Time Password. This is a 6-digit number that isn't ever sent over the mesh, but is verified over the mesh using SHA256 hashing. I can write out an extensive description of what exactly it does and why if desired.

The second is the actual key verification. In this scheme, NodeA initiates a verification, sending a request over PKC to NodeB. NodeB generates the 6-digit MOTP and displays it on screen, and sends it in a clientNotification message. This number is supplied to NodeA via an admin message. This MOTP exchange is rather bulletproof for proving the nodes have the correct keys for each other, but an additional validation string is then displayed on both devices, and a final admin message confirms. This validation string is 8 characters of bastardized base64.

The last thing of note, which is primarily introduced in the  protobuf pr #690, is the addition of a payload_variant inside clientNotification messages. The nice thing here is that the oneof keyword makes these optional, as they can always be None. This should be rather backwards and forwards compatible. But this gives us a way to push messages to the phone that includes more than just a text string, like I'm doing here. It also gives us a future path towards i18n support for clientNotification messages.

### DIFF
--- a/src/modules/KeyVerificationModule.cpp
+++ b/src/modules/KeyVerificationModule.cpp
@@ -1,3 +1,4 @@
+#if !MESHTASTIC_EXCLUDE_PKI
 #include "KeyVerificationModule.h"
 #include "MeshService.h"
 #include "RTC.h"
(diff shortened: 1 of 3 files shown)


thebentern:
> Good job copilot thanks for saving us from Verificatino

Premature internationalization

PR #6893, opened 2025-05-26
The ESP32 Bluetooth service has been enhanced to support the output of waypoints in GPWPL format.

lcflyr:
I have already tested it with Heltec Wireless Tracker V1.1


src/mesh/NodeDB.cpp
```diff
@@ -1479,6 +1479,8 @@ void NodeDB::updatePosition(uint32_t nodeId, const meshtastic_Position &p, RxSou
     info->has_position = true;
     updateGUIforNode = info;
     notifyObservers(true); // Force an update whether or not our node counts have changed
+
```
fifieldt, line 1482: o
src/mesh/NodeDB.cpp: In member function 'void NodeDB::updatePosition(uint32_t, const meshtastic_Position&, RxSource)':
src/mesh/NodeDB.cpp:1483:5: error: 'nimbleBluetooth' was not declared in this scope

lcflyr:
> This is really neat. A couple of thoughts: I would like this to be opt-in via the bluetooth config, as it generates a bunch of extra traffic over the PhoneAPI. I would also like to have an NRF52 implementation as well, so that we don't tie this to a single platform.
> 
> @lcflyr are there any specific apps or use cases you created this for? I'm curious because I've looked at location standard GATT stuff recently.

I'm a beginner in embedded systems. It might take some time for me to achieve the function you mentioned.
Many APRS map software support the GPWPL waypoint format, allowing users to view the positions of other nodes without being limited to the Meshtastic app.

src/nimble/NimbleBluetooth.cpp
```diff
@@ -299,6 +306,65 @@ void NimbleBluetooth::sendLog(const uint8_t *logMessage, size_t length)
...
+    double minutes = (value - degrees) * 60.0;
+    
+    if (isLatitude) {
+        snprintf(buffer, bufferSize, "%02d%07.4f", degrees, minutes);
+    } else {
+        snprintf(buffer, bufferSize, "%03d%07.4f", degrees, minutes);
+    }
+}
+
+void NimbleBluetooth::Send_GPWPL(uint32_t node, char* name, int32_t latitude_i, int32_t longitude_i)
```
caveman99, line 326: Please re-use and possibly improve the code in `/src/gps/NMEAWPL.*` because the same functionality exists on the serial port.

fifieldt:
@lcflyr thanks so much for sharing this, very cool :)

Issue #6844, opened 2025-05-19
[Bug]:


egzumer:
### Category

Other

### Hardware

Heltec Mesh Node T114

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.6.7

### Description

Hi. This report won't be as structured as you may want, sorry for that. I have nothing more than educated guesses, but this seems to be a real issue, so I want to provide some feedback.

We have a mesh of ~16 nodes.  I have a node on a roof and a node in home. Most of the nodes in the mesh are closer to each other and I'm a bit further away. There is a roof node further away that I have a good connection with. We used to set my roof node and the other one to routers. This way we had a pretty reliable connection. Then came 2.6.

The next hop routing seems to not work very well. My home node only sees the roof node. When I try to send trace routes to another node that has good signal I loose around 50% or sometimes even 100% responses, same thing with direct messages, confirmations get lost. I tried client, router, router_late on the roof, doesn't matter. When I send messages directly from the roof node 100% always gets through. When I went to the roof with my home node and sent from there, the traceroutes were going out through the roof node and coming back directly, again close to 100% were coming back.
I switched back to 2.5 in router_late mode on the roof node (home node still on 2.6) and now I get 100% responses when I send from home.
This test was conducted on a good signal short path. But the same thing happens on the whole mesh around me. Since 2.6 came out the mesh performance for traceroutes and direct messages worsened.

I had similar issues when the roof node was set in client mode. It looks like next hop routing effectively sets its hops the same way as client routing. We have some roof nodes on the mesh, but most are nodes in buildings. I presume routing gets stuck between buildings. What 2.6 does it effectively prevents me from taking advantage of router/router_late setting.

### Relevant log output

```Shell

```

canadamadman:
This realy dosnt seem like bug  seems more like your not useing it corectly. Not supposed to be usieng reouter on your home. At all. Ever. Its for mountains. And everyone had moved to medium fast now.

egzumer:
It seems that with a mountain there would be exactly the same issue.

GUVWAF:
[...]

PR #6904, opened 2025-05-27
Don't cancel sending ReTx for relayer if we're ROUTER(_LATE)/REPEATER

GUVWAF:
Try-fix for #6844. When stopping retransmissions, which was previously only done on packets originating from ourselves, for `ROUTER(_LATE)` and `REPEATER`, we should not actually cancel a packet we're relaying.

@egzumer Can you try if this fixes the issue in your set-up?


thebentern (approved):

egzumer:
I can confirm it works as it should now. I received 3 duplicates of the same packet, scheduled TX was not cancelled, one packet was relayed.

shalberd:
@GUVWAF will this make it into a beta instead of alpha firmware, too, soon?
I have read through the issue and the fix and it seems that, since direct messaging has been introduced in 2.6, this is a bug fix in routers that should make it into more stable beta type firmware soon. What do you think?

GUVWAF:
@shalberd It is in 2.6.10 which I expect to become stable soon. 
This fix mostly involves the `ROUTER_LATE` role, because the canceling happened when it was waiting to relay.

shalberd:
>  This fix mostly involves the ROUTER_LATE role, because the canceling happened when it was waiting to relay.

@GUVWAF does this lead to ROUTER_LATE, when doing direct message relaying, not sticking to the late rebroadcast window?

https://github.com/meshtastic/firmware/blob/9feb1d378edb2847cc12755beac9aa57a10c571f/src/mesh/FloodingRouter.cpp#L58

Does the ROUTER_LATE still wait to relay now?
Or is the ROUTER_LATE, in terms of being less agressive when relaying messages in terms of letting clients-to-clients transmit first, only of use and used in Flooding Routing?

The reason I am asking: We currently have a situation where we have high mountain nodes on firmware 2.6.11, each mountain in line of sight of the other, called NFN* in node name, and we see traceroute and direct message problems from areas in far away regions with clients:
https://github.com/orgs/meshtastic/discussions/24#discussioncomment-14096482
as hopping to the comment does not seem to work, see the description:
Screenshot 2025-08-17 at 20 41 22

From the first hop on, even though two clients can see each other very strongly, our mountain routers take over, with ROUTER to ROUTER to ROUTER using up of hops. How can we remedy this? Would ROUTER_LATE be of use, in terms of programmatically, using a slight TX delay or late rebroadcast window ... in this context of direct messaging, too?

If yes, then we would try to go the way of setting all mountain nodes to ROUTER_LATE. We need to achieve that, while providing extended coverage for flooding routing and direct messages, our mountain routers do not take over a direct message from one client to another client in 0 hop range right away. I hope I express my wish and problem clearly ...

GUVWAF:
> Does this lead to ROUTER_LATE, when doing direct message relaying, not sticking to the late rebroadcast window?

Indeed, `ROUTER_LATE` only has effect on broadcast (channel) messages. 
When all nodes are updated to at least 2.6, when next-hops are assigned, there should be no competition between nodes, since only one will relay. Are all these `ROUTER` nodes actually on 2.6 or higher?

shalberd:
> Indeed, ROUTER_LATE only has effect on broadcast (channel) messages.

I see, interesting. Yes, all nodes, including the clients, are on at least 2.6, with most being at 2.6.11.
Maybe a kind of delayed, clampToLateBroadcastWindow, would be good for ROUTER_LATE as well, in NextHopRouting.

> When all nodes are updated to at least 2.6, when next-hops are assigned, there should be no competition between nodes, since only one will relay.

I think our problem is that our mountain nodes of type ROUTER are so well-placed, we are talking 1500 - 2000m altitude, that they grab packets from nodes as far away as 50 miles and never let a client in that location 60 miles away from the mountain router do its job and relay / do the next hop to the next client in 0 hop distance, with green very good signal quality, from the client perspective.

I guess the problem is, that the "only one will relay / next hop" is the mentioned extemely well-placed router node on the mountain. How can we allow client to client path to happen as an option when using our Routers on mountaintops, and many clients on rooftops in a "great plain" 60 miles away?

What about traceroute? What explains asymetric paths client to client, with client in "the great plains" to mountain router, then to another mountain router, then to another mountain router .... of to 5 router hops in between? While the response / route back to the originating client node "in the great plain" goes directly back from the target client node?

GUVWAF:
> I guess the problem is, that the "one one will relay / next hop" is the mentioned extemely well-placed router node on the mountain.

Yes, this seems to be what is happening. Note that the next-hops are assigned based on the managed flooding approach, so indeed then `ROUTER`s will get priority. 

> How can we allow client to client path to happen as an option when using our Routers on mountaintops, and many clients on rooftops in a "great plain" 60 miles away?

In your case it might actually just help to put (some) of the mountain nodes just in `CLIENT` mode.

> What explains asymetric paths client to client, with client in "the great plains" to mountain router, then to another mountain router, when to another mountain router .... of to 5 router hops in between? While the response / route back to the originating client node "in the great plain" goes directly back from the target client node?

That might just be because the way towards the target client node (LagMesh) cannot be reached directly, while the other way around it can (due to different hardware or surrounding obstacles). The other `ROUTER`s might still try to rebroadcast, but the client already got the response and will display the first route that arrives.

GUVWAF:
> Maybe a kind of delayed, clampToLateBroadcastWindow, would be good for ROUTER_LATE as well, in NextHopRouting.

Ah, sorry, I was wrong before. This is also applied in NextHopRouting, as the function `perhapsCancelDupe()` (where `clampToLateBroadcastWindow()` is called) is also called by the NextHopRouter. 
I believe the issue here is that next-hops are assigned based on the managed flooding approach, so your mountain `ROUTER`s will indeed get priority. And, if multiple mountain routers can hear the packet, they will all try to send quickly, increasing the chance of collisions.

shalberd:
>  Note that the next-hops are assigned based on the managed flooding approach, so indeed then ROUTERs will get priority

Yes, that is what is happening. Good to know, in terms of understanding, that the next-hops are assigned based in managed flooding.

> In your case it might actually just help to put (some) of the mountain nodes just in CLIENT mode.

Okay, but just putting them in ROUTER_LATE mode would not make a difference, would it?
What we do agree on: too many ROUTER device role nodes on mountains can be counterproductive.

> That might just be because the way towards the target client node (LagMesh) cannot be reached directly

The poster said the rooftop node was in "good signal quality" range to LagMesh target client node. Could be that hos mobile node was not in direct range of LagMesh client target node, though.

shalberd:
> clampToLateBroadcastWindow() is called) is also called by the NextHopRouter

ok, good, that means ROUTER_LATE can help us in the situation we are facing. In a sense, ROUTER_LATE being the better, more considerate ROUTER ...

> I believe the issue here is that next-hops are assigned based on the managed flooding approach, so your mountain ROUTERs will indeed get priority.

ok, makes sense, yes

> And, if multiple mountain routers can hear the packet, they will all try to send quickly, increasing the chance of collisions

yes, that explains the "tour of Switzerland" route the packet is taking, from router to router :-)

GUVWAF:
> ok, good, that means ROUTER_LATE can help us in the situation we are facing. In a sense, ROUTER_LATE being the better, more considerate ROUTER ...

Yes, `ROUTER_LATE` is also worth a try, if these mountain nodes have good coverage for "black spots".

> Could be that hos mobile node was not in direct range of LagMesh client target node, though.

That seems to be the case. If LagMesh would have received it, the traceroute would have shown the short route. The packet may arrive in multiple ways, but it will always show the first packet that arrived.

shalberd:
> Yes, ROUTER_LATE is also worth a try, if these mountain nodes have good coverage for "black spots".

They do, the line of sight analyses with the usual tools (https://www.heywhatsthat.com/, https://site.meshtastic.org/) show the ROUTER nodes all being extremely well-placed, covering all areas very well. I think it is just, that they are too agressive.
We need to make sure they give Clients a chance to "be the first next hop" first, and, since the ROUTERS can all see each other, but add reach in different directions for extended coverage, routers listening to other routers in the form of ROUTER_LATE can be the way to solve our problem of the "packet tour of Switzerland".

We are blessed with a very good topography here in terms of omni-directionally-accessible, super-elevated mountain nodes, but it really seems like ROUTER_LATE is the way forward.

Thank you for your insights.

GUVWAF:
> We are blessed with a very good topography here in terms of omni-directionally-accessible, super-elevated mountain nodes, but it really seems like ROUTER_LATE is the way forward.

I think so too. Switzerland is indeed a nice experimental playground for this. Good luck and happy to hear the results!

PR #6920, opened 2025-05-30
Do position broadcast on secondary channel if disabled on primary

jp-bennett:
Only compile tested for the moment.

### DIFF
--- a/src/modules/PositionModule.cpp
+++ b/src/modules/PositionModule.cpp
@@ -328,7 +328,15 @@ void PositionModule::sendOurPosition()
 
     // If we changed channels, ask everyone else for their latest info
     LOG_INFO("Send pos@%x:6 to mesh (wantReplies=%d)", localPosition.timestamp, requestReplies);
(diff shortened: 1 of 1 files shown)


thebentern (approved):

src/modules/PositionModule.cpp
```diff
@@ -328,7 +328,15 @@ void PositionModule::sendOurPosition()
 
     // If we changed channels, ask everyone else for their latest info
     LOG_INFO("Send pos@%x:6 to mesh (wantReplies=%d)", localPosition.timestamp, requestReplies);
-    sendOurPosition(NODENUM_BROADCAST, requestReplies);
+    for (uint8_t channelNum = 0; channelNum > 8; channelNum++) {
+        if (channels.getByIndex(channelNum).settings.has_module_settings) {
+            if (channels.getByIndex(channelNum).settings.module_settings.position_precision != 0) {
```
thebentern, line 333: Maybe consolidate these ifs

fifieldt (approved):
Sane

PR #6923, opened 2025-05-30
Ignore NodeInfos from Users with forged MAC addresses

thebentern:
I believe we should ignore Users which have a mismatch of mac address and nodenum.


GUVWAF:
One caveat here is when a node generates a random NodeNum when it detects a duplicate one on the mesh.

thebentern:
> One caveat here is when a node generates a random NodeNum when it detects a duplicate one on the mesh.

I had this thought as well. What do you think is a good mitigation strategy? I'm not sure there's a great experience to be had for duplicates on the mesh

GUVWAF:
Maybe a client notification (only once upon new discovery)? Then you can always ignore/delete the node yourself.

At this point I'm fairly certain there are nodes around where this has happened, especially since RAK sent out batches of nodes with the same NodeNum.

Issue #6854, opened 2025-05-20
[Board]: T-Deck pro (PCM Audio & A7682E 4G)


Szetya:
### SOC

ESP32

### Lora IC

SX1262

### Product Link

https://lilygo.cc/products/t-deck-pro

### Description

Sales have started although it is a bit slow.
I managed to get a PCM Audio version. I would appreciate support. First round of basic peripherals needed.
https://github.com/Xinyuan-LilyGO/T-Deck-Pro/tree/master

Thanks in advance!
I would appreciate your help with EPD. @caveman99  @todd-herbert

Szetya:
Okay.
- Display works with ZinggJM library even if I get warnings. :)
- LoRa radio works.
- GPS works.
- I2C scan finds devices (unknown I2C: 0x1A Touch CST328 and 0x28 Gyroscope BHI260AP) 
0x55 on the target is BQ27220 and it is incorrectly recognized as a T-Deck keyboard.
Keyboard is TCA8418, but meshtastic TCA8418Keyboard.cpp is unusable. :(
Battery voltage not measured by GPIO PIN. 

Extra lines in the setup() of the main.cpp file:
```cpp
#if defined(T_DECK_PRO)
    pinMode(BOARD_LORA_EN, OUTPUT);
    digitalWrite(BOARD_LORA_EN, HIGH);
    pinMode(BOARD_1V8_EN, OUTPUT);
    digitalWrite(BOARD_1V8_EN, HIGH);
    pinMode(BOARD_6609_EN, OUTPUT);
    digitalWrite(BOARD_6609_EN, HIGH);
    // T-Deck pro has all three SPI peripherals (E-ink, SD, LoRa) attached to the same SPI bus
    // We need to initialize all CS pins in advance otherwise there will be SPI communication issues
    // e.g. when detecting the SD card
    pinMode(LORA_CS, OUTPUT);
    digitalWrite(LORA_CS, HIGH);
    pinMode(SDCARD_CS, OUTPUT);
    digitalWrite(SDCARD_CS, HIGH);
    pinMode(PIN_EINK_CS, OUTPUT);
    digitalWrite(PIN_EINK_CS, HIGH);
    delay(100);
#endif
```
log:
```plaintext
//\ E S H T /\ S T / C

INFO  | ??:??:?? 2 Booted, wake cause 0 (boot count 1), reset_reason=reset
DEBUG | ??:??:?? 2 Filesystem files (20480/3538944 Bytes):
[...]

PR #6936, opened 2025-06-01
T-Deck Pro support

mverch67:
**Implementation status**

*   [x] E-Ink display GxEPD2_310_GDEQ031T10
*   [x] Touch screen CST328
*   [x] LoRa Radio SX1262/SX1268
*   [x] TCA8418 keyboard
*   [x] GPS M10Q
*   [x] SD card
*   [ ] BQ27220 battery chip
*   [ ] BQ25896 battery chip
*   [ ] LTR-553ALS light sensor
*   [ ] BHI260AP gyroscope

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


mverch67:
CannedMessage module function (keys) are missing currently.

Szetya:
The BQ27220 does not provide useful information.
```txt
OperationConfigA Values:0x0
External Thermistor Selected: NO
BATT_GD Pin Polarity High: NO
BATT_GD Function Enabled: NO
Can Enter SLEEP State: NO
slpwakechg Function Enabled: NO
Write Temperature Function Enabled: NO
Battery Insertion Detection Enabled: NO
Battery Insertion Pin Pull - Up Enabled: NO
Pin Function Code (PFC) Mode: 0
Wake - Up Function Enabled: NO
Wake - Up Threshold 1: 0
Wake - Up Threshold 0: 0

OperationConfigB Values:0x0
Default Seal Option Enabled: NO
Non - Removable Option Set: NO
INT_BREM Function Enabled: NO
INT_BATL Function Enabled: NO
INT_STATE Function Enabled: NO
INT_OCV Function Enabled: NO
INT_OT Function Enabled: NO
INT_POL Function Enabled (High - Level Polarity): NO
INT_FOCV Function Enabled: NO

Standard query:
        - AtRate:0 mA
        - AtRateTimeToEmpty:0 minutes
        - Temperature:-273.15 ℃
        - BatteryVoltage:0 mV
        - InstantaneousCurrent:0 mAh
        - RemainingCapacity:0 mAh
        - FullChargeCapacity:0 mAh
        - DesignCapacity:0 mAh
        - TimeToEmpty:0 minutes
        - TimeToFull:0 minutes
        - StandbyCurrent:0 mA
        - StandbyTimeToEmpty:0 minutes
        - MaxLoadCurrent:0 mA
        - MaxLoadTimeToEmpty:0 minute
        - RawCoulombCount:0 mAh
        - AveragePower:0 mW
        - InternalTemperature:-273.15 ℃
        - CycleCount:0
        - StateOfCharge:0 %
        - StateOfHealth:0 %
        - RequestChargingVoltage:0 mV
        - RequestChargingCurrent:0 mA
        - BTPDischargeSet:0 mAh
        - BTPChargeSet:0 mAh

Operation Status:
        - getIsConfigUpdateMode:NO
        - getIsBtpThresholdExceeded:NO
        - getIsCapacityAccumulationThrottled:NO
        - getIsInitializationComplete:NO
        - getIsDischargeCycleCompliant:NO
        - getIsBatteryVoltageBelowEdv2:NO
        - getSecurityAccessLevel:0
        - getIsCalibrationModeEnabled:NO

Battery Status:
===============================================
```

Szetya:
TCA8418 treatment is a bit better but still needs some work.
There is a modifierFlag variable for the modifier keys, so the TCA8418TapMap can have 7 columns.
Shift: modifierFlag bit 1
alt: modifierFlag bit 2
Sym: modifierFlag bit 3
[src_input.zip](https://github.com/user-attachments/files/20541474/src_input.zip)
The formatting needs a little tweaking. . :)

kstrauser:
I can confirm that this is somewhat working today, as-is. Keyboard input is very slow and the shift keys don't work (but you can double-tap a letter to make it caps). Backspace and enter work. You can send and receive messages through the radio.

It's not polished yet, but it does actually function as a standalone device with keyboard, screen, radio, and buzzer notifications.

This is awesome. My little TDP is alive and usable for the first time now! Thank you both for getting it to this point!

Szetya:
> I can confirm that this is somewhat working today, as-is. Keyboard input is very slow and the shift keys don't work (but you can double-tap a letter to make it caps). Backspace and enter work. You can send and receive messages through the radio.
> 
> It's not polished yet, but it does actually function as a standalone device with keyboard, screen, radio, and buzzer notifications.
> 
> This is awesome. My little TDP is alive and usable for the first time now! Thank you both for getting it to this point!

Try the contents of the zip file. The code is a bit messy but it works better. Shift, alt, Sym too.

adrianTNT:
These firmware updates are for the T-Deck Pro "4G" version or the "voice" version? Or is it irrelevant?
Thank you.

kstrauser:
@Szetya Yep, that's a nice improvement.

@adrianTNT I'm using it on the 4G.

Szetya:
@adrianTNT It is irrelevant at the moment because neither option is supported. It will probably have to be split later as it uses common PINs.

Has anyone disassembled it yet? 😉 LTR553ALS location found? I ordered from aliexpress. Also plan on installing a BME280 or BME688 once I take the time.

mverch67:
@Nasimovy
You may try if the refactored class is still working with your keyboard device 😅

Szetya:
> @Nasimovy You may try if the refactored class is still working with your keyboard device 😅

WAO! I'll look at this tomorrow on t-deck pro. The value of modifierFalg can be up to 7 at the moment (shift + alt + sym) so the tapMap array can be up to 7+1 columns. There are plenty of possibilities to handle special events or use special characters. And the two shifts can be handled separately and then 15+1 combinations are possible for each key (which is a lot) 😂
(The t-lora pager is coming soon. Version 868MHz is already available on the official site (ordered today))

mverch67:
> WAO! I'll look at this tomorrow on t-deck pro. The value of modifierFalg can be up to 7 at the moment (shift + alt + sym) so the tapMap array can be up to 7+1 columns.

I tried to use the same keys as on the T-Deck with quite similar behavior. The only difference is that instead of ALT+C - key now can use ALT+key directly.

Szetya:
TCA8418 is capable of requesting 3 buttons at once (under certain matrix rules) which is not currently used and I don't think it is needed. Currently shift increases the modifierFalg value by 1, alt by 2 and sym by 4. Before the _TCA8418_MULTI_TAP_THRESHOLD time expires, the modifier keys can be pressed in sequence to use different combinations e.g. shift & alt & key. Which gives the index of the third key of tapMap. 
It would be a good idea to have some information about the state of the modifierFalg on the display but the top line of the display is full.

mverch67:
> TCA8418 is capable of requesting 3 buttons at once (under certain matrix rules) which is not currently used and I don't think it is needed. Currently shift increases the modifierFalg value by 1, alt by 2 and sym by 4. Before the _TCA8418_MULTI_TAP_THRESHOLD time expires, the modifier keys can be pressed in sequence to use different combinations e.g. shift & alt & key. Which gives the index of the third key of tapMap. It would be a good idea to have some information about the state of the modifierFalg on the display but the top line of the display is full.

I've modified all this already, i.e. implemented the multi-key press and removed the long press which we don't use on T-Deck. I guess nobody wants to press more than one modifier key in sequence which would also simplify the logic and the programming.
What is currently not working is that the shift key can be held while typing a series of capital alphabetic characters. Looks like an issue with the modifier state.
From what you are suggesting I assume you've never worked with a T-Deck, right?

Szetya:
> > TCA8418 is capable of requesting 3 buttons at once (under certain matrix rules) which is not currently used and I don't think it is needed. Currently shift increases the modifierFalg value by 1, alt by 2 and sym by 4. Before the _TCA8418_MULTI_TAP_THRESHOLD time expires, the modifier keys can be pressed in sequence to use different combinations e.g. shift & alt & key. Which gives the index of the third key of tapMap. It would be a good idea to have some information about the state of the modifierFalg on the display but the top line of the display is full.
> 
> I've modified all this already, i.e. implemented the multi-key press and removed the long press which we don't use on T-Deck. I guess nobody wants to press more than one modifier key in sequence which would also simplify the logic and the programming. What is currently not working is that the shift key can be held while typing a series of capital alphabetic characters. Looks like an issue with the modifier state. From what you are suggesting I assume you've never worked with a T-Deck, right?

For me, it was not uncomfortable to press several buttons in a row during my tests. The beauty of it all is that you can approach the same thing in different ways. 😉
I have no problem with your solution, that's not why I wrote it. I haven't tried your version yet (I'll take the time this afternoon).
I have a first generation t-deck but I don't use it.

Szetya:
> > > TCA8418 is capable of requesting 3 buttons at once (under certain matrix rules) which is not currently used and I don't think it is needed. Currently shift increases the modifierFalg value by 1, alt by 2 and sym by 4. Before the _TCA8418_MULTI_TAP_THRESHOLD time expires, the modifier keys can be pressed in sequence to use different combinations e.g. shift & alt & key. Which gives the index of the third key of tapMap. It would be a good idea to have some information about the state of the modifierFalg on the display but the top line of the display is full.
> > 
> > 
> > I've modified all this already, i.e. implemented the multi-key press and removed the long press which we don't use on T-Deck. I guess nobody wants to press more than one modifier key in sequence which would also simplify the logic and the programming. What is currently not working is that the shift key can be held while typing a series of capital alphabetic characters. Looks like an issue with the modifier state. From what you are suggesting I assume you've never worked with a T-Deck, right?
> 
> For me, it was not uncomfortable to press several buttons in a row during my tests. The beauty of it all is that you can approach the same thing in different ways. 😉 I have no problem with your solution, that's not why I wrote it. I haven't tried your version yet (I'll take the time this afternoon). I have a first generation t-deck but I don't use it.

If you take a look at the meshtastic map of Hungary, you will understand why you don't need a qwerty device. It's currently a hobby for some crazy people. 😂 No good infrastructure and no good coverage. Also 433MHz and 868MHz devices are very mixed. Chances are high that messages will not reach their destination.

mverch67:
@kstrauser @Szetya 
Do you also have issues fully charging the T-Deck Pro? Sometimes it charges, sometimes not e.g. when shutdown. After 10h charging still at 80%. Weird. I've opened an issue [here](https://github.com/Xinyuan-LilyGO/T-Deck-Pro/issues/6).

Szetya:
> @kstrauser @Szetya Do you also have issues fully charging the T-Deck Pro? Sometimes it charges, sometimes not e.g. when shutdown. After 10h charging still at 80%. Weird. I've opened an issue [here](https://github.com/Xinyuan-LilyGO/T-Deck-Pro/issues/6).

It doesn't charge for me above 80-85% either. There was a photo here that I printed extra information on the display just to see what was happening (picture here: https://github.com/meshtastic/firmware/issues/6854#issuecomment-2907956446 the photo is in the state of discharging)
When the charge reached about 4.2V the charge was switched off, you could see it on the display 0mA & Not Charging
When I unplugged the charging cable it dropped back to the known 85% and showed about 4V.

Gruppio:
Hello! I've installed this version but after a while the keyboard has stopped working 🤔 Now or it ignores key press or types random stuff autonomously

Gruppio:
I've tried to erase and reflash but the issue is still present

Szetya:
@mverch67 Add the #define I2C_NO_RESCAN macro to the variant.h file. Thank you!

ReconPirate:
@Szetya Have tried Adding the #define I2C_NO_RESCAN macro to the variant.h file,  And unfortunately ghost writer is still at work typing on me keyboard and randomly restarting on the table next to me.  And as added oddness, i have never had it respond to me actually typing on the keyboard ever.  (Clean build, only the macro above added)

Szetya:
> @Szetya Have tried Adding the #define I2C_NO_RESCAN macro to the variant.h file, And unfortunately ghost writer is still at work typing on me keyboard and randomly restarting on the table next to me. And as added oddness, i have never had it respond to me actually typing on the keyboard ever. (Clean build, only the macro above added)

Unfortunately, I experienced it yesterday. 🙁
And I had the I2C_NO_RESCAN macro defined from the beginning. 
I'm at a loss, maybe I should use a complete directive to disable the TDECK_KB lookup for T_DECK_PRO and use it in the I2C scanner. The BQ27220 does not provide any useful information, all values are OFF or 00. My first solution was something similar in my trials. It seemed stable then. 
I am sure it is the BQ27220 that is causing it.

Nasimovy:
> @Nasimovy You may try if the refactored class is still working with your keyboard device 😅

it works fine with my keyboard, you made some nice improvements

Szetya:
> @Szetya Have tried Adding the #define I2C_NO_RESCAN macro to the variant.h file, And unfortunately ghost writer is still at work typing on me keyboard and randomly restarting on the table next to me. And as added oddness, i have never had it respond to me actually typing on the keyboard ever. (Clean build, only the macro above added)

@ReconPirate @mverch67 
https://github.com/meshtastic/firmware/blob/8f9e569825b673a21c644d3b398a8afea1ac222e/src/detect/ScanI2C.cpp#L32-L36
Maybe the order can solve the problem if the TCA8418KB comes first?!

jstockdale:
Hi there!

I installed this and it mostly works great. One question: How do you navigate between nodes and channels when sending a message?

So if you hit alt + t it selects Broadcast, and a second alt+t will select the Primary channel, but I can't figure out how to navigate to other nodes or other channels. Once the Primary channel is selected in the alt + t menu, I've tried pressing nearly every button and also swiping on the screen but nothing seems to make it change to the Secondary channel I have programmed.

Other than that, this is working great. Thanks so much for this awesome effort!! :+1: 

-John

Hylian:
Great work! I successfully ran this patch on my T-Deck Pro (Voice, US band) and communicated with another node. 

I know this is just an initial patch, but some observations so far:
- Modifier keys are not held across multiple key presses.
- Screen refresh sometimes misses the latest input, showing the previous item or character until the next input.
- The `Alt + C` bindings inherited from T-Deck don't feel great and a new command mapping should probably be created. Arrow keys in particular are important now.
- Screensaver / screen lock UX could be reconsidered in the context of e-ink.
- UI in general should take better advantage of the additional vertical height.
- Channel selector UX when sending messages is very finicky, and I accidentally transmitted on channel 0 several times when attempting to send a test message on a private channel.

I'll try and see where I can contribute once initial board support is stabilized.

jstockdale:
Latest changes fix all the issues I described. Thank you all! Can confirm most of the relatively minor bugs reported by Hylian. But my device is up and running great for the most part! Huzzah!! 🥳

jstockdale:
My unit seems to not always be charging properly however. It's been plugged in all night and it is at 40%. Before I rebooted it, it was showing 0%. I'm wondering if it stopped being able to talk to the power management chip after being on for a while?

After I rebooted it, it's been fine. Showing ... *checks again* ... 48% now.

t-deck-pro_power-issue

reticulatingsplines:
Hi all,

Thanks for your work on this, @Szetya. I was also having a multitude of keyboard problems (ignored inputs, ghost inputs) but the change you suggested above seems to have solved the problem (requoted below):
> https://github.com/meshtastic/firmware/blob/8f9e569825b673a21c644d3b398a8afea1ac222e/src/detect/ScanI2C.cpp#L32-L36
> 

Apologies if there is a more useful place to share these, but here are some high-resolution pictures of the circuitboard. Note to others re disassembly: There are two layers to the rear half of the case: an outer rear shell that attaches to the front shell via plastic clips and an inner rear shell that is attached to the front shell using 6 small screws. These screws screw into threaded inserts that have been set into plastic posts on the front shell that also fit into notches on the edges of the PCB, holding it in place. There are 6 main retention tabs for the outer shell, one at each corner as well as one on each side between the keyboard and the buttons/sim slot. There are a couple smaller tabs for alignment that don't need to be unhooked. The side tabs are especially fragile; you may have more luck than me (I broke one despite being pretty careful) if you start with the 4 corner tabs using an especially thin spudger. Removing the SIM card tray can provide a good point to start.

When you open the unit, be aware that the plastic button cap for the reset/boot buttons isn't attached to anything and will just pop right out. There are two small pieces of clear plastic that are stuck between the plastic button cap and the tactile switches on the PCB which I guess was Lilygo's way of bridging the distance and/or providing a better tactile feel to the buttons. Pay attention to where these fall out if you want to keep them, or else remember you'll need to stick something in those two spaces again if you want the side buttons to work well again after reassembly.

Missing from the pictures is a piece of black fabric tape that I removed to show where the keyboard and screen ribbon cables connect to the PCB. The battery is also wrapped in a similar black fabric tape as you can clearly see from the outside of an unmodified unit. 

IMG_4864
IMG_4865
IMG_4866
IMG_4867
IMG_4868
IMG_4869
IMG_4870
IMG_4871
IMG_4872
IMG_4873
IMG_4874
IMG_4875
IMG_4876
IMG_4877
IMG_4878

reticulatingsplines:
Other notes I forgot to put in the previous message:

The inner rear case has a cutout that's just slightly smaller than the battery, perhaps to make room for whatever the bullseye printed on the outer case was meant to represent (wireless charger? NFC reader? Location to aim hammer?). The battery in my unit is a 305070 LiPo from Tewaycell with a marked rating of 1400mAh and a mfg date from April 2025, which, hey, not bad. There is a real 10k thermistor on the protection board attached to the battery (as opposed to just a 10k resistor, which I've certainly found before) so that's not bad either.

As you can see without even opening the case, the WiFi, LoRa, and (presumably only on the 4G units?) GSM antennas are thin flex-pcb stickers that are attached to the inner rear case. These actually connect to the rest of their circuits via metal tabs that make contact with spring-loaded terminals on the PCB, which I was a little surprised by - notes taken from Blackberry et al. These spring-loaded terminals on the PCB are located near U.FL connectors that have tiny coax cables connected to them that pass their signal to another U.FL connector near the relevant IC. The onboard "chip"- or "tile"-style GPS antenna is attached via a standard U.FL connector as well. This should allow easy antenna customization for any of the 4 wireless communication protocols.

As you can see from the picture of the keyboard side of the PCB, there's still a bare footprint where Lilygo intended to put an ambient light sensor (LTR-553ALS-01 from Liteon). Liteon discontinued this sensor, which is Lilygo's explanation for not populating it on the board. Given that Lilygo's schematic shows 2 unpopulated resistor footprints, 2 resistors, and 3 capacitors total between the microphone and the light sensor, and that's exactly what we see on the back despite the missing sensor, everything's already in place if you get your hands on an LTR-553ALS-01. TMD27721 from OSRAM also looks to be a good replacement - same pinout, same dimensions, same operating voltage.   I can't find its I2C slave address on quick glance though, so if you go that route, make sure there's no collision between the other ICs that share the same I2C bus.

There's a tactile switch near the A7682E (GSM/LTE modem); based on the schematic and on Simcom's datasheet, pressing this pulls the A7682E "BOOT_CFG" pin low, and doing this while powering up the modem and then toggling the modem's reset pin, which is connected to GPIO9 of the ESP32, places the modem in "USB Download" mode, which it appears allows you to flash the modem's firmware from a computer via USB.  The modem's USB interface is available through the test points (and it appears the test points only) next to the modem on the PCB.

Szetya:
I have not yet had the courage to open it. This means that the LTR553ALS has its place, and there is the possibility of soldering a thin wire to a potentially usable environmental sensor. The I2C for the T-deck V1 is easy to use for the pull-up resistor. 
I ordered this:
https://a.aliexpress.com/_EIE1CZY

mverch67:
@reticulatingsplines 
This thing looks like a battery... right? Does T-Deck Pro has a RTC or what is this thing used for?

image

jstockdale:
@mverch67 that would be the backup battery for the onboard GPS. It's labeled VRTC in the schematic, on page 3. It's only connected to V_BCKP on the GPS chip. There is no separate RTC on the board.

image

Hope that's helpful. Cheers! 😃 

_Also fun tip for anyone doing board rework, modifications, repair, etc. remove this little guy or shield it with kapton tape or other heat protection first. Hot air will ruin these little batteries._

jstockdale:
I'm not sure what happened, but my device hangs on boot now. Anyone else experiencing this?

```
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x15 (USB_UART_CHIP_RESET),boot:0x18 (SPI_FAST_FLASH_BOOT)
Saved PC:0x42153bd6
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce3808,len:0x4bc
load:0x403c9700,len:0xbd8
load:0x403cc700,len:0x2a0c
entry 0x403c98d0
E (362) esp_core_dump_flash: No core dump partition found!
E (362) esp_core_dump_flash: No core dump partition found!
[  1246][I][esp32-hal-psram.c:96] psramInit(): PSRAM enabled
��@INFO  | ??:??:?? 1 

//\ E S H T /\ S T / C

INFO  | ??:??:?? 1 Booted, wake cause 0 (boot count 1), reset_reason=reset
DEBUG | ??:??:?? 1 Filesystem files (20480/1048576 Bytes):
DEBUG | ??:??:?? 1  /prefs/channels.proto (57 Bytes)
DEBUG | ??:??:?? 1  /prefs/device.proto (82 Bytes)
DEBUG | ??:??:?? 1  /prefs/nodes.proto (545 Bytes)
[  1307][I][esp32-hal-i2c.c:75] i2cInit(): Initialising I2C Master: sda=13 scl=14 freq=100000
DEBUG | ??:??:?? 1 Power::lipoInit lipo sensor is not ready yet
[  1318][I][XPowersCommon.tpp:453] begin(): SDA:13 SCL:14
[  1323][W][Wire.cpp:301] begin(): Bus already started in Master Mode.
INFO  | ??:??:?? 1 PPM BQ25896 init succeeded
DEBUG | ??:??:?? 1 Power::lipoChargerInit lipo sensor is ready
[  1341][W][Wire.cpp:301] begin(): Bus already started in Master Mode.
INFO  | ??:??:?? 1 CST328 found at address 0x1a
INFO  | ??:??:?? 1 LTR553ALS found at address 0x23
INFO  | ??:??:?? 1 BHI260AP found at address 0x28
INFO  | ??:??:?? 1 TCA8418 found at address 0x34
INFO  | ??:??:?? 1 BQ27220 found at address 0x55
INFO  | ??:??:?? 1 BQ25896 found at address 0x6b
INFO  | ??:??:?? 1 6 I2C devices found
DEBUG | ??:??:?? 1 acc_info = 0
DEBUG | ??:??:?? 1 SD_MMC Card Type: 
DEBUG | ??:??:?? 1 SDHC
DEBUG | ??:??:?? 1 SD Card Size: 30436 MB
DEBUG | ??:??:?? 1 Total space: 30424 MB
DEBUG | ??:??:?? 1 Used space: 25 MB
INFO  | ??:??:?? 1 S:B:102,2.7.1.f2c2f0423
DEBUG | ??:??:?? 1 Total heap: 294120
DEBUG | ??:??:?? 1 Free heap: 257776
DEBUG | ??:??:?? 1 Total PSRAM: 8386263
DEBUG | ??:??:?? 1 Free PSRAM: 8359327
DEBUG | ??:??:?? 1 NVS: UsedEntries 6, FreeEntries 624, AllEntries 630, NameSpaces 2
DEBUG | ??:??:?? 1 Setup Preferences in Flash Storage
DEBUG | ??:??:?? 1 Number of Device Reboots: 2
ESP_ERROR_CHECK_WITHOUT_ABORT failed: esp_err_t 0x105 (ESP_ERR_NOT_FOUND) at 0x403815bf
file: "src/platform/esp32/BleOta.cpp" line 16
func: static const esp_partition_t* BleOta::findEspOtaAppPartition()
expression: esp_ota_get_partition_description(part, &app_desc)
ESP_ERROR_CHECK_WITHOUT_ABORT failed: esp_err_t 0x102 (ESP_ERR_INVALID_ARG) at 0x403815bf
file: "src/platform/esp32/BleOta.cpp" line 30
func: static String BleOta::getOtaAppVersion()
expression: esp_ota_get_partition_description(part, &app_desc)
INFO  | ??:??:?? 1 No BLE OTA firmware available
INFO  | ??:??:?? 1 No WiFi OTA firmware available
INFO  | ??:??:?? 1 Init NodeDB
[  1530][E][vfs_api.cpp:105] open(): /littlefs/static/static does not exist, no permits for creation
[  1548][E][vfs_api.cpp:105] open(): /littlefs/prefs/db.proto does not exist, no permits for creation
INFO  | ??:??:?? 1 Load /prefs/nodes.proto
INFO  | ??:??:?? 1 Loaded /prefs/nodes.proto successfully
INFO  | ??:??:?? 1 Loaded saved nodedatabase version 24, with nodes count: 1
INFO  | ??:??:?? 1 Load /prefs/device.proto
INFO  | ??:??:?? 1 Loaded /prefs/device.proto successfully
INFO  | ??:??:?? 1 Loaded saved devicestate version 24
[  1607][E][vfs_api.cpp:105] open(): /littlefs/prefs/config.proto does not exist, no permits for creation
ERROR | ??:??:?? 1 Could not open / read /prefs/config.proto
INFO  | ??:??:?? 1 Install default LocalConfig
INFO  | ??:??:?? 1 Set default channel and radio preferences!
DEBUG | ??:??:?? 1 Expand short PSK #1
INFO  | ??:??:?? 1 Wanted region 0, using UNSET
[  1625][E][vfs_api.cpp:105] open(): /littlefs/prefs/module.proto does not exist, no permits for creation
ERROR | ??:??:?? 1 Could not open / read /prefs/module.proto
INFO  | ??:??:?? 1 Install default ModuleConfig
INFO  | ??:??:?? 1 Load /prefs/channels.proto
INFO  | ??:??:?? 1 Loaded /prefs/channels.proto successfully
INFO  | ??:??:?? 1 Loaded saved channelFile version 24
[  1659][E][vfs_api.cpp:105] open(): /littlefs/prefs/uiconfig.proto does not exist, no permits for creation
ERROR | ??:??:?? 1 Could not open / read /prefs/uiconfig.proto
DEBUG | ??:??:?? 1 cleanupMeshDB purged 0 entries
DEBUG | ??:??:?? 1 Use nodenum 0x2f931c38 
DEBUG | ??:??:?? 1 Number of Device Reboots: 2
DEBUG | ??:??:?? 1 Expand short PSK #1
INFO  | ??:??:?? 1 Wanted region 0, using UNSET
DEBUG | ??:??:?? 1 Coerce telemetry to min of 30 minutes on defaults
DEBUG | ??:??:?? 1 Save to disk 4
DEBUG | ??:??:?? 1 Opening /prefs/device.proto, fullAtomic=1
INFO  | ??:??:?? 1 Save /prefs/device.proto
WARN  | ??:??:?? 1 Packet History - Invalid size -1, using default 500
DEBUG | ??:??:?? 1 SPI.begin(SCK=36, MISO=47, MOSI=33, NSS=3)
DEBUG | ??:??:?? 1 Use compiled/slipstreamed tzplaceholder                                         
DEBUG | ??:??:?? 1 Set Timezone to GMT0
DEBUG | ??:??:?? 1 Read RTC time as 23
INFO  | ??:??:?? 1 GPS power state move from OFF to ACTIVE
DEBUG | ??:??:?? 1 Use GPIO44 for GPS RX
DEBUG | ??:??:?? 1 Use GPIO43 for GPS TX
DEBUG | ??:??:?? 1 Start multicast thread
DEBUG | ??:??:?? 1 NeighborInfoModule is disabled
DEBUG | ??:??:?? 1 Trackball GPIO initialized (255, 255, 255, 255, 255)
[  1792][E][vfs_api.cpp:105] open(): /littlefs/prefs/cannedConf.proto does not exist, no permits for creation
ERROR | ??:??:?? 1 Could not open / read /prefs/cannedConf.proto
INFO  | ??:??:?? 1 CannedMessageModule is enabled
[  1808][E][vfs_api.cpp:105] open(): /littlefs/prefs/ringtone.proto does not exist, no permits for creation
ERROR | ??:??:?? 1 Could not open / read /prefs/ringtone.proto
INFO  | ??:??:?? 1 Init External Notification Module
INFO  | ??:??:?? 1 Use Pin 2 for vibra motor
INFO  | ??:??:?? 1 Do EInk init
INFO  | ??:??:?? 1 Turn on screen
DEBUG | ??:??:?? 1 Update E-Paper
Busy Timeout!
_PowerOn : 10000047
```

EDIT: Correction– it hangs on every epaper write. Ooof. I wonder if I broke my screen?
Screenshot 2025-06-29 at 21 49 21

lewisxhe:
@jstockdale Did you open it? Did you touch the screen cable?

lewisxhe:
> @kstrauser @Szetya Do you also have issues fully charging the T-Deck Pro? Sometimes it charges, sometimes not e.g. when shutdown. After 10h charging still at 80%. Weird. I've opened an issue [here](https://github.com/Xinyuan-LilyGO/T-Deck-Pro/issues/6).

Where does your battery level come from? Do you get the battery level by reading the BQ27220?

jstockdale:
@lewisxhe I opened my unit after noticing this issue, and checked the ribbon cables. Both the touch and display ribbon cables were well seated and not showing any signs of damage.

I *disconnected* the battery and waited about ten seconds. Then reconnected the battery. That fixed the screen being unresponsive.

I'm currently in touch with LilyGo to see if we can come up with a reason why this happened. It sounds like the display does not get reset along with the main esp32s3 when you press the reset button.

LilyGo advises they will have an updated version of some support libraries for this device in the next few days.

I'll post here when I hear more. Cheers!

lewisxhe:
Yes, that was my reply. The problem should be that the screen cannot be reset and is stuck.
Because the screen is not connected to the RST pin, you must reset the screen by powering off and then powering on.
The only way to solve this problem is to trigger the PMU to shut down by some method, and then turn on the PMU again to achieve the effect of resetting the screen.

karman85:
Hello. I'm having trouble charging the T-Deck-Pro. I received 3 days ago and I've only been able to charge it when battery goes fully depleted. 

Right now, for example, I have it connected to the PC, battery shows 70% but no charging at all:

INFO  | ??:??:?? 2 PPM BQ25896 init succeeded
DEBUG | ??:??:?? 2 Power::lipoChargerInit lipo sensor is ready
INFO  | ??:??:?? 2 Scan for i2c devices
[  2874][W][Wire.cpp:301] begin(): Bus already started in Master Mode.
DEBUG | ??:??:?? 2 Scan for I2C devices on port 1
INFO  | ??:??:?? 2 CST328 found at address 0x1a
INFO  | ??:??:?? 2 BHI260AP found at address 0x28
INFO  | ??:??:?? 2 TCA8418 found at address 0x34
INFO  | ??:??:?? 2 BQ27220 found at address 0x55
INFO  | ??:??:?? 2 BQ25896 found at address 0x6b
INFO  | ??:??:?? 2 5 I2C devices found
INFO  | ??:??:?? 3 S:B:102,2.7.1.47f9cc3a
[...]
DEBUG | 20:02:31 86 [Power] Battery: usbPower=0, isCharging=0, batMv=3791, batPct=70

Multiple USB Cables and USB ports have been tested good delivering power. T-Deck-Pro is not only not charging but discharging even when plugged with the serial port opened.

mverch67:
> Hello. I'm having trouble charging the T-Deck-Pro. I received 3 days ago and I've only been able to charge it when battery goes fully depleted.

Welcome to the club.

karman85:
> > Hello. I'm having trouble charging the T-Deck-Pro. I received 3 days ago and I've only been able to charge it when battery goes fully depleted.
> 
> Welcome to the club.

Sorry. Didn’t want to disturb. I fully read the whole thread and didn’t found anything about this issue and saw the check on the BQ25896 implementation so I though I was the only one having the trouble. 

Just shout me out if I can be of any help :)

jstockdale:
I'm also seeing charging issues here, and am trying to figure out if there's any rhyme or reason to it all.

As far as I can tell, my unit charges reliably in certain states. But then in other states it completely refuses to charge at all.

This morning it wasn't charging, and some combination of unplugging the battery, plugging it back in, getting the device to reset, and then as soon as the charging indicator showed up I put it back in deep sleep / "shutting down" mode by holding the top rocker (power/boot) button got it sorted.

I noticed that immediately after starting charging from depletion, it trickle charges at ~.300 mW (~60-63 mA?), and only once it gets to a minimum voltage does it start charging at the full ~4W (~910 mA). Once in the ~4W charging mode it seems to charge to completion for me ... I've gone from 0% to ~60% without it dropping back below ~4W.

I'm also seeing a max draw of just north of 900mA (I need to pull out my more accurate power meter to really be sure, but on a unit accurate to about ~5% I'm seeing 930mA as max draw) with gps on and my unit not sleeping. I think the common limit (and this is definitely true for Macbook Pro models) is 900mA.

I wonder if the 800mA limit is excluding device draw? I haven't specifically tested the t-deck pro yet, but if I recall correctly, a t-deck plus with esp32s3, with wifi, gps, and keyboard burns around ~130mA when it's not sleeping. If the power limit isn't including the device draw it needs to be ~770mA or slightly under to not hit the 900mA port limit.

I'll keep troubleshooting and also may pull up the datasheets for the BQ25896 if I have time.

Thanks @mverch67 for working on this. 🙏 👍 Sent a small sponsorship your way just now to help show my appreciation. Cheers!

jstockdale:
@lewisxhe is there a correction curve on the battery power level at all? I'm seeing the device charge from 0% to 40% almost immediately (under 5 minutes), and then it's taken another 30 minutes for my unit to go from 40% to 64%. The whole time the draw at the port has been ~4.0-4.32W (870-930mA). Is improving charge capacity estimation also part of the calibration process you mentioned about the power management chip? Thanks!

Update: 1 h and 42m later we're at 90%. Stable at ~4.32-4.41W (.93A current draw at 4.75V).

lewisxhe:
There are currently no battery curves and no calibration has been done.

karman85:
Mine has gone wild. battery voltage reading is 4.13v and can see 1.4W draining in the external charger. However the gauge is reporting 55% battery:

DEBUG | 09:49:13 8054 [Power] Battery: usbPower=1, isCharging=1, batMv=4130, batPct=55

mverch67:
> Mine has gone wild. battery voltage reading is 4.13v and can see 1.4W draining in the external charger. However the gauge is reporting 55% battery:

All these values look reasonable to me. Though the voltage during charging should be higher, but maybe you are using an old version of the firmware.

karman85:
> > Mine has gone wild. battery voltage reading is 4.13v and can see 1.4W draining in the external charger. However the gauge is reporting 55% battery:
> 
> All these values look reasonable to me. Though the voltage during charging should be higher, but maybe you are using an old version of the firmware.

Same firmware as this comment (2.7.1.47f9cc3a): 
https://github.com/meshtastic/firmware/pull/6936#issuecomment-3033152674 
But back then 3.7v would read 70% :S

Szetya:
Hi @mverch67 !

Please try this fork or pull: https://github.com/Szetya/BQ27220
I made some modifications in power.cpp. I am not sure if it works 100%. You can now set the battery capacity so it will display 100%.
https://github.com/meshtastic/firmware/blob/7a1295e976bf2be24e0ac239e74e3ebc017f56df/src/Power.cpp#L1307-L1320 >>

```cpp
        if (bq == nullptr) {
            bq = new BQ27220;
            bool result = bq->init(&Wire, BQ27220_I2C_SDA, BQ27220_I2C_SCL);
            if (result) {
                int currentCap = bq->getFullChargeCap();
                if (currentCap != 1400) {
                    if (bq->isSealed()) {
                        bq->unseal();
                    }
                    bool success = bq->setDesignCapacity(1400);
                    if (success) {
                        LOG_WARN("Design capacity updated to 1400");
                    }
                    bq->seal(); // optional, if you want to lock it back
                } else {
                    LOG_WARN("Design capacity already set to 1400, skipping update.");
                }
                return true;
            } else {
                LOG_WARN("BQ27220 init failed");
                delete bq;
                bq = nullptr;
                return false;
            }
        }
        return false;
    }
```

We need to somehow reduce the write to the BQ27220 EEPROM so that it doesn't write out the 1400mAh capacity every time. This would be done with the seal(), unseal() and isSealed() functions.

Szetya:
> Hi @mverch67 !
> 
> Please try this fork or pull: https://github.com/Szetya/BQ27220 I made some modifications in power.cpp. I am not sure if it works 100%. You can now set the battery capacity so it will display 100%.
> 
> https://github.com/meshtastic/firmware/blob/7a1295e976bf2be24e0ac239e74e3ebc017f56df/src/Power.cpp#L1307-L1320
> 
> >>
> ```c++
>         if (bq == nullptr) {
>             bq = new BQ27220;
>             bool result = bq->init(&Wire, BQ27220_I2C_SDA, BQ27220_I2C_SCL);
>             if (result) {
>                 int currentCap = bq->getFullChargeCap();
>                 if (currentCap != 1400) {
>                     if (bq->isSealed()) {
>                         bq->unseal();
>                     }
>                     bool success = bq->setDesignCapacity(1400);
>                     if (success) {
>                         LOG_WARN("Design capacity updated to 1400");
>                     }
>                     bq->seal(); // optional, if you want to lock it back
>                 } else {
>                     LOG_WARN("Design capacity already set to 1400, skipping update.");
>                 }
>                 return true;
>             } else {
>                 LOG_WARN("BQ27220 init failed");
>                 delete bq;
>                 bq = nullptr;
>                 return false;
>             }
>         }
>         return false;
>     }
> ```
> 
> We need to somehow reduce the write to the BQ27220 EEPROM so that it doesn't write out the 1400mAh capacity every time. This would be done with the seal(), unseal() and isSealed() functions.

image
From 100% in 8 hours 24 minutes to 67%.

lewisxhe:
I need to clarify that LilyGo does not calibrate and write the design capacity to BQ27220 when it leaves the factory, so you need to write it yourself.

Szetya:
> I need to clarify that LilyGo does not calibrate and write the design capacity to BQ27220 when it leaves the factory, so you need to write it yourself.

Does this mean that entering my designcapacity is the right way to go?

I mean if I change the designcapacity to 1400-1500mAh then the charge can reach 100%.. the battery charge icon then changes to a plug icon which I understand means that it switches to USB power. 
(I haven't looked at the log or measurement)
In my experience, the runtime is over 10 hours (which is realistic and acceptable given a hungry esp32 and battery capacity) 
During discharge, somewhere below 50%, the charge value drops steeply.
I used the library mentioned in my https://github.com/meshtastic/firmware/pull/6936#issuecomment-3038339687 post.

lewisxhe:
Right! Currently, T-Deck-Pro has a nominal 1500mA battery, so it is more reasonable to reduce the design capacity to 1400mA. I agree with your setting.

mverch67:
@lewisxhe 
> I need to clarify that LilyGo does not calibrate and write the design capacity to BQ27220 when it leaves the factory, so you need to write it yourself.

Why isn't the code to do initial setting of the design capacity put into the firmware that is flashed at the factory? Then all devices that are shipped from the factory have the proper settings. I think it's better and more flexible to have a separate tool to program the battery design capacity than to put it into every single application that is to be run on the T-Deck Pro. Think of the use case that people want to replace the stock battery with another (larger) one. In that case the application (which assumes a fixed known design capacity) won't work properly with that device...

mverch67:
> In my experience, the runtime is over 10 hours (which is realistic and acceptable given a hungry esp32 and battery capacity)

I did run the T-Deck Pro exactly for **48 hours** with one battery charge in mode `client mute` in a busy network with 250 nodes. With the current factory settings that don't account the proper battery design capacity the percentage was already at 0% at 3.58V even though the device ran for another 10 hours.

Szetya:
> > In my experience, the runtime is over 10 hours (which is realistic and acceptable given a hungry esp32 and battery capacity)
> 
> I did run the T-Deck Pro exactly for **48 hours** with one battery charge in mode `client mute` in a busy network with 250 nodes. With the current factory settings that don't account the proper battery design capacity the percentage was already at 0% at 3.58V even though the device ran for another 10 hours.

We do not have 250 nodes in the country. 😂 In fact I'm happy to see 1-2 nodes directly. 
(Although not related to the topic but since it is on the subject. Unfortunately I have noticed that many people are discouraged by the lack of coverage (here in Hungary) . We have 2-3 or maybe 4 routers in very good and good positions. They are the backbone of the whole network, almost everyone is connected through them.) 
48+ hours seems very good. I have never tried any other mode than client.

factory-kirk:
At what point do we call this good enough to merge, then start adding incremental PRs to smooth out the remaining issues?

(I say "we" as if I wrote a single line of this, which I did not. 😀)

mverch67:
> At what point do we call this good enough to merge, then start adding incremental PRs to smooth out the remaining issues?

I would say when the biggest show stoppers are gone. Otherwise we'll get a sh*t storm of bug reports for the things we know already.

mverch67:
@mr-tbot 

If you've ever flashed one of the previous firmware versions you need to either remove the battery or let it _completely_ run empty (until it does not wake up anymore), otherwise the PMM will not be reprogrammed correctly. That's why you saw that device is not charging.

benjamind:
I've been experimenting with this build too. So far battery charging appears to be working after I let it run down completely, but can't really comment on expected battery life yet.

I have however been struggling to get the device to broadcast messages a lot, seems to have very poor signal. It also appears to be sending hi/bye messages on sleep and wake?

There's also an off-by-one error on the eink refresh with it not triggering an update on quite a few of the screens  including when typing in free text messages it always misses the last pressed character in the update. Similarly disconnecting charging cable does not update until the next input.

Good to see things making progress though!

mverch67:
@mr-tbot 
> I would expect from an e-ink device - my t-deck plus lasts considerably longer... at least in my case so far. 

This is probably because you are running the T-Deck Pro with BT on but T-Deck Plus with BT off? If so I suggest to disable BT on the Pro and then compare the runtimes.
Also device role should be same (ideally client mute if you are indoors).

Mirarak:
In the LilyGo T-Deck main menu it is not visible where the device is connected, the name of the connected device could be added to the fifth icon from the top, where it is - signal status. Thank you. Mirarak

mverch67:
> In the LilyGo T-Deck main menu it is not visible where the device is connected, the name of the connected device could be added to the fifth icon from the top, where it is - signal status. Thank you. Mirarak

Sorry, you're wrong here. This PR is for T-Deck Pro...

Mirarak:
know, but I can't create a new request, I'm not a contributor :) sorry Mirarak
______________________________________________________________
> Od: "Manuel" ***@***.***>
> Komu: "meshtastic/firmware" ***@***.***>
> Datum: 16.07.2025 16:06
> Předmět: Re: [meshtastic/firmware] T-Deck Pro support (PR #6936)
>
mverch67 left a comment (meshtastic/firmware#6936) <https://github.com/meshtastic/firmware/pull/6936#issuecomment-3078777653>In the LilyGo T-Deck main menu it is not visible where the device is connected, the name of the connected device could be added to the fifth icon from the top, where it is - signal status. Thank you. Mirarak
Sorry, you're wrong here. This PR is for T-Deck Pro...

tajo48:
Compiled and tested, works great!
Yet I'm wondering if the standard ui is the best choice and even if it is the navigation is not so great tbh

src/main.cpp
```diff
@@ -335,6 +335,15 @@ void setup()
...
     delay(100);
+#elif defined(T_DECK_PRO)
+    pinMode(LORA_EN, OUTPUT);
+    digitalWrite(LORA_EN, HIGH);
+    pinMode(LORA_CS, OUTPUT);
+    digitalWrite(LORA_CS, HIGH);
+    pinMode(SDCARD_CS, OUTPUT);
+    digitalWrite(SDCARD_CS, HIGH);
+    pinMode(PIN_EINK_CS, OUTPUT);
+    digitalWrite(PIN_EINK_CS, HIGH);
```
caveman99, line 346: is the delay(100) not needed here?

caveman99 (approved):

caveman99:
maybe move the variant dir to the new structure before merging?

src/main.cpp
mverch67: no, not needed. It's for the T-Deck with the esp32-c3 co-processor to start up. The T-Deck Pro has a TCA8418 keypad driver to control the keyboard.

Szetya:
@mverch67 
e.kbchar same in the following lines:
https://github.com/meshtastic/firmware/blob/df8b629c2c4c43b5939e7153a427100c1b769a42/src/input/kbI2cBase.cpp#L308-L315

firstel:
I flashed my T-Deck Pro with the latest alpha.   First I tried a full erase with the UI, and upon rebooting the keyboard led came on and started to flash, which is the first sign of life I have seen on this device since I let the battery run down months ago.  But the display never changed, just flashing led.  Then I tried again without the Mesh UI, and the device rebooting with the vibration motor going non-stop.   Encouraging that this has an effect on the device, but not working for me yey.  Any advice?

mverch67:
> I flashed my T-Deck Pro with the latest alpha. First I tried a full erase with the UI, and upon rebooting the keyboard led came on and started to flash, which is the first sign of life I have seen on this device since I let the battery run down months ago. But the display never changed, just flashing led. Then I tried again without the Mesh UI, and the device rebooting with the vibration motor going non-stop. Encouraging that this has an effect on the device, but not working for me yey. Any advice?

Did you put the device into DFU mode before flashing?

firstel:
Yes.  I held the top button then the bottom while holding the top, released the bottom and then the top in that order.

Sent from my mobile device. Please excuse any typos.

PR #6938, opened 2025-06-01
Add scaffolding to read EEPROM data from ch341

jp-bennett:
We import these functions from IMSProg (stripped down a bit, removing what isn't needed). They are licensed GPLv3, so should be no problem on the licensing front. This does allow reading from an attached EEPROM. May or may not be the option we choose for the long-term solution.


fifieldt (approved):

fifieldt:
@jp-bennett , did you want this in? Or still working out the long-term direction?

jp-bennett:
> @jp-bennett , did you want this in? Or still working out the long-term direction?

Still waiting to get hardware that would actually use it.

jp-bennett:
Looking like we don't need this code in Meshtasticd, so closing for now.

PR #6939, opened 2025-06-01
Convert arduino strings

thebentern:


jp-bennett:
Worked on and pulled manually.

PR #6940, opened 2025-06-02
Fixed issue using INA sensors to read voltage of node.

wel97459:
This fixes the issue of the INA sensor not being shown as the nodes voltage and battery percentage.

Tested on RP2040-lora.
Changes should fix the same problems on other devices.
Code was not changed in away that would break things.


wel97459:
What needs to be done?
I'm new to to the repo.

thebentern:
Hi,

Circling back to this one. Can you address the merge conflicts?

Thanks

wel97459:
I resolved the conflicts.

wel97459:
Been running this on my RP2040-lora boards for a month now. It's work as expected.
Also test it on a Heltech v2 board and it worked the same.

fifieldt:
Very sorry for the slow review here. With changes to hardware we normally get out our own sensors to give it a test, but it's been so busy recently :(

PR #6945, opened 2025-06-03
Fix cppcheck warnings

jp-bennett:
Remove unused variables and fix multiple cppcheck issues

### DIFF
--- a/src/graphics/draw/NodeListRenderer.cpp
+++ b/src/graphics/draw/NodeListRenderer.cpp
@@ -74,16 +74,6 @@ const char *getSafeNodeName(meshtastic_NodeInfoLite *node)
     return nodeName;
 }
 
-uint32_t sinceLastSeen(meshtastic_NodeInfoLite *node)
-{
-    uint32_t now = getTime();
-    uint32_t last_seen = node->last_heard;
-    if (last_seen == 0 || now < last_seen) {
-        return UINT32_MAX;
-    }
-    return now - last_seen;
-}
-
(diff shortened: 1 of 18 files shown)


src/graphics/draw/NodeListRenderer.cpp
```diff
@@ -139,7 +129,7 @@ void retrieveAndSortNodes(std::vector<NodeEntry> &nodeList)
         bool aFav = a.node->is_favorite;
         bool bFav = b.node->is_favorite;
         if (aFav != bFav)
```
fifieldt, line 131: so if one node is a favourite (not neither, not both), then return the favourite status of node a?

jp-bennett: It's a sort function. So if they're not both favorite, and the first one is, then sort the first one first. I believe it's equivalent to the old code.

fifieldt: You're right, it is equivalent.

prior:
A favourite, B not favourite --> sort A first (1 > 0 --> 1 "A")
A not favourite, B favourite --> sort B first (0 > 1 --> 0 "B")
A favourite, B favourite --> continue sorting
A not favourite, B not favourite --> continue sorting

now:
A favourite, B not favourite --> return "1" A
A not favourite, B favourite --> return "0" B
A favourite, B favourite --> continue sorting
A not favourite, B not favourite --> continue sorting

thebentern (approved):

PR #6951, opened 2025-06-03
Generate keys when Lora Region is set

jp-bennett:


src/modules/AdminModule.cpp
```diff
@@ -660,7 +660,23 @@ void AdminModule::handleSetConfig(const meshtastic_Config &c)
 #endif
         config.lora = c.payload_variant.lora;
         // If we're setting region for the first time, init the region
-        if (isRegionUnset && config.lora.region > meshtastic_Config_LoRaConfig_RegionCode_UNSET) {
+        if (!owner.is_licensed && isRegionUnset && config.lora.region > meshtastic_Config_LoRaConfig_RegionCode_UNSET) {
```
ianmcorvidae: I think adding a check for licensed here will cause issues for licensed users, since the stuff below the added code still needs to run for licensed nodes I think (setting tx enabled, `initRegion()` etc). Probably could add a nested if, I suppose.

PR #6962, opened 2025-06-05
upgrade Seeed wio trackerL1 device  peripheral interaction.

Dylanliacc:
# upgrade Seeed wio trackerL1 device  peripheral interaction.
To make better use of the logic of Canned messages, an interface has been opened for Canned messages module to facilitate the trackerball module to monitor the current state of Canned messages. The final implemented effect is that users can enter Canned messages by moving the joystick up and down, and can smoothly exit by pressing the menu button. After exiting, the Canned message module automatically close.

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [x] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


src/input/TrackballInterruptImpl1.cpp
```diff
@@ -52,3 +57,22 @@ void TrackballInterruptImpl1::handleIntPressed()
 {
     trackballInterruptImpl1->intPressHandler();
 }
+
+void TrackballInterruptImpl1::handleMenuPressed()
+{
+    bool activateMenuAction = false;
+
+    if (cannedMessageModule && cannedMessageModule->isUIVisibleAndInterceptingInput()) {
```
fifieldt, line 67: Looks like this isn't available in some variants:


src/input/TrackballInterruptImpl1.cpp: In static member function 'static void TrackballInterruptImpl1::handleMenuPressed()':
src/input/TrackballInterruptImpl1.cpp:65:9: error: 'cannedMessageModule' was not declared in this scope
   65 |     if (cannedMessageModule && cannedMessageModule->isUIVisibleAndInterceptingInput()) {

Dylanliacc: yes i saw that ,but not sure why i can build success for some of failed vairents

Dylanliacc: still tring to locat the problem

src/input/TrackballInterruptImpl1.cpp
```diff
@@ -2,6 +2,8 @@
 #include "InputBroker.h"
 #include "configuration.h"
 
+#include "main.h"
+#include "modules/CannedMessageModule.h"
```
thebentern, line 7:
```
#if !MESHTASTIC_EXCLUDE_CANNEDMESSAGES
#include "modules/CannedMessageModule.h"
#endif
```

src/input/TrackballInterruptImpl1.cpp
thebentern:
```
#if !MESHTASTIC_EXCLUDE_CANNEDMESSAGES
if (cannedMessageModule...)
{
}
#endif
```

jp-bennett:
The code this addresses is under a re-write in the unify-tft branch, so let's hold off on merging this here. I think this change is actually already implemented there, as the extra button is used as a cancel button.

Dylanliacc:
> The code this addresses is under a re-write in the unify-tft branch, so let's hold off on merging this here. I think this change is actually already implemented there, as the extra button is used as a cancel button.

based on base UI  close this pr

PR #6983, opened 2025-06-07
Replace blocking delay for wifi reconnect with non-blocking to keep button/display interactivity

mattster98:
## 🙏 Thank you for sending in a pull request, here's some tips to get started!

Full disclosure - I used copilot to author this change.  Annoying thing I've been hoping to solve and it appeared quick/easy.

## 🤝 Attestations
  - [X] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


fifieldt (approved):
This seems like a good improvement, thanks for talking to the bots about it :)

fifieldt:
Hi @mattster98 , are you familar with trunk? We use it for linting. If you use vscode you can install the plugin and it'll automatically fix that check that's failing

mattster98:
I am not but I'll take a look.  I'd like to figure it out as there are a couple other changes I'd like to make as well.

src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -474,4 +491,4 @@ uint8_t getWifiDisconnectReason()
 {
     return wifiDisconnectReason;
 }
-#endif
\ No newline at end of file
+
```
GUVWAF: Looks like this `#endif` ended up in line 224, but I believe it should stay here.

src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -153,24 +157,38 @@ static int32_t reconnectWiFi()
...
-#endif
+    #endif
         LOG_INFO("Reconnecting to WiFi access point %s", wifiName);
 
-        delay(5000);
+        // Start the non-blocking wait for 5 seconds
+        wifiReconnectStartMillis = millis();
+        wifiReconnectPending = true;
+        // Do not attempt to connect yet, wait for the next invocation
+        return 100; // Schedule next check soon
```
GUVWAF: Let's make this 5000, we don't need to check 50 times if 5 seconds have already passed.

src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -153,24 +157,38 @@ static int32_t reconnectWiFi()
...
+            #ifdef CONFIG_IDF_TARGET_ESP32C3
+                WiFi.mode(WIFI_MODE_NULL);
+                WiFi.useStaticBuffers(true);
+                WiFi.mode(WIFI_STA);
+            #endif
+                WiFi.begin(wifiName, wifiPsw);
+            }
+            isReconnecting = false;
+            wifiReconnectPending = false;
+            return 300000; // Resume normal interval after reconnect
```
GUVWAF: I think it's better not to return here, as it should e.g. execute `onNetworkConnected()` after this, and not only after 5 min.

mattster98:
Sorry, because I don't know what I'm doing I have made some other changes in the github web interface I thought would remain local to my fork for testing and it has attached them here.  I got vscode going and did my best to straighten it out and address the changes.  Please take another look.  Thank you for your patience with me.

GUVWAF:
Looks good to me now.
Can you revert the changes to the yml files?

mattster98:
> Looks good to me now. Can you revert the changes to the yml files?

Done - starting to get the hang of this.  Learning the hard way unfortunately.

PR #7000, opened 2025-06-09
Add InkHUD driver for WeAct Studio 1.54" display module

todd-herbert:
Display model is `ZJY200200_0154DAAMFGN`. Currently identical to `GDEY0154D67`, and using same code base, but recognized as separate in case the drivers do diverge in future.

Optimizations made to the existing `GDEY0154D67` driver, and tested on affected devices. 

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - NRF52 Pro-micro DIY (custom build)
    - LilyGo T-Echo
    -  Elecrow ThinkNode M1

### DIFF
--- a/src/graphics/niche/Drivers/EInk/GDEY0154D67.cpp
+++ b/src/graphics/niche/Drivers/EInk/GDEY0154D67.cpp
@@ -39,10 +39,10 @@ void GDEY0154D67::detachFromUpdate()
 {
     switch (updateType) {
     case FAST:
-        return beginPolling(50, 500); // At least 500ms for fast refresh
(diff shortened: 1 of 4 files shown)


todd-herbert:
Just want to confirm the default value for the number of gate lines, one moment

todd-herbert:
The verdict: in theory it shouldn't matter either way, but lets continue explicitly specifying the number of gate lines (vertical resolution) because it's a good habit.

PR #7003, opened 2025-06-10
Warn users about low entropy keys

jp-bennett:


src/mesh/NodeDB.cpp
```diff
@@ -1556,8 +1562,20 @@ bool NodeDB::updateUser(uint32_t nodeId, meshtastic_User &p, uint8_t channelInde
     }
 
 #if !(MESHTASTIC_EXCLUDE_PKI)
-    if (p.public_key.size > 0) {
+    if (p.public_key.size == 32) {
         printBytes("Incoming Pubkey: ", p.public_key.bytes, 32);
+
+        // Alert the user if a remote node is advertising public key that matches our own
+        if (owner.public_key.size == 32 && memcmp(p.public_key.bytes, owner.public_key.bytes, 32) == 0) {
+            char warning[] = "Remote device %s has advertised your public key. This may indicate a low-entropy key. You may need "
```
danditomaso, line 1570: Would stronger language be helpful here? "You may...." could become "You will...."

thebentern (approved):

src/mesh/NodeDB.cpp
jp-bennett: I know a few users duplicate their own keys on purpose, so not every case is a problem. But point taken. I may beef this up in the future.

formtapez:
There is a small problem with that:
When you change the name of your node, the warning immediately pops up - because a "new node" (yourself) is now using your public key.
That should be fixed.

PR #7008, opened 2025-06-10
Enable telemetry and I2C sensors on STM32WL (except accelerometers)

Stary2001:
This adds support for I2C sensors to the STM32WL target. It seems to just work (...if you add the right sensor library into lib_deps in your variant, for now).
Accelerometers are still disabled, I don't have any spare to test.
Also, the power sensor libraries are explicitly linked in because there's some hard dependencies in the power code when you turn on sensors.. I'm not sure how to properly untangle those yet, probably for a future PR.

This also needed some obligatory flash space wrangling:
Everything off: 220712
* disable timezone + RTC::getTZOffset change to eliminate mktime and co if it's set: 227860 -> 220712 (-7148 bytes)
* RadioInterface pow() replacement: 224816 -> 220712 (-4104 bytes)
  * It uses only pow(2, n) and pulls in a bunch of floating point code for no reason.
* Exclude audio+atak+wifi(?): Too big with everything enabled! `region 'FLASH' overflowed by 1964 bytes.`

  - [x] Wio E5 dev board

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -15,11 +15,7 @@
 // Calculate 2^n without calling pow()
 uint32_t pow_of_2(uint32_t n)
 {
-    if (n == 0) {
-        return 1;
-    } else {
-        return 1 << (n - 1);
-    }
+    return 1 << n;
 }
 
 #define RDEF(name, freq_start, freq_end, duty_cycle, spacing, power_limit, audio_permitted, frequency_switching, wide_lora)      \


arch/stm32/stm32.ini
```diff
@@ -39,9 +42,21 @@ debug_tool = stlink
 lib_deps =
   ${env.lib_deps}
   ${radiolib_base.lib_deps}
+  # This is bad, but currently they're required by Power.cpp
```
caveman99: If they are general dependencies, move them out of the telemetry dependancies instead of listing them twice. You could make a new include target.

variants/wio-e5/platformio.ini
```diff
@@ -8,20 +8,17 @@ build_flags =
   -DSERIAL_UART_INSTANCE=1
   -DPIN_SERIAL_RX=PB7
   -DPIN_SERIAL_TX=PB6
+  -DPIN_WIRE_SDA=PA15
+  -DPIN_WIRE_SCL=PB15
   -DHAL_DAC_MODULE_ONLY
```
caveman99, line 13: if you tell the whole arch it has i2c, the pin defines need to be made for all 3 targets. or do the other toolchains define those internally?

caveman99 (changes requested):

variants/wio-e5/platformio.ini
Stary2001: Currently all the pin mapping comes from here AFAIK: https://github.com/stm32duino/Arduino_Core_STM32/blob/main/variants/STM32WLxx/WL54JCI_WL55JCI_WLE4J(8-B-C)I_WLE5J(8-B-C)I/variant_generic.h
which will define this to be on PA9/10 by default
I'll check what pins it's meant to be on the RAK

arch/stm32/stm32.ini
Stary2001: I've made these properly optional now so I can just remove them from here.

variants/wio-e5/platformio.ini
Stary2001: https://docs.rakwireless.com/product-categories/wisduo/rak3172-module/datasheet/
UART is on the default PA2/PA3:
1	PA3/UART2_RX	I	Reserved - UART2/LPUART1 Interface (AT Commands and FW Update)
2	PA2/UART2_TX	O	Reserved - UART2/LPUART1 Interface (AT Commands and FW Update)
and I2C1 is on PA11/PA12:
9	PA12/I2C_SCL	I/O	GPIO and I2C (SCL)
10	PA11/I2C_SDA	I/O	GPIO and I2C (SDA)
The RAK3372 also brings out a 2nd I2C: https://docs.rakwireless.com/Product-Categories/WisBlock/RAK3372/Datasheet/ on PA9/10 which ~~goes to the seperate connector on the baseboards that have it I think.~~ goes nowhere.

Stary2001: The EBYTE dev board has no defined IO, so that should be OK for all 3 platforms.

src/mesh/RadioInterface.cpp
```diff
@@ -12,6 +12,16 @@
 #include <pb_decode.h>
 #include <pb_encode.h>
 
+// Calculate 2^n without calling pow()
+uint32_t pow_of_2(uint32_t n)
+{
+    if (n == 0) {
+        return 1;
+    } else {
+        return 1 << (n - 1);
```
GUVWAF: This should be just `1 << n`.

Stary2001: Fixed in 6d7f953. Not sure how I thought it was anything else..

GUVWAF (approved):

thebentern (approved):

PR #7009, opened 2025-06-10
Don't use assert() with side effects in a couple more places

Stary2001:
I ran into issues with the STM32 code being compiled with -DNDEBUG, which makes asserts no-op. These were in my tree but didn't make it in with the others.
Searching for assert uses with a function call with `rg 'assert\(.*[\(].*\)'` shows there's a couple left in InkHUD, but it's probably OK.


todd-herbert:
> shows there's a couple left in InkHUD, but it's probably OK.

Yes I think they're all just sanity checking, and should be totally fine to no-op on any platform. 👍

fifieldt (approved):

PR #7034, opened 2025-06-13
PacketHistory debloat RAM allocations

Marek-mk:
### Was:
- Using unordered_set caused 40B of heap allocation with every new entry.
- There was no limit on allowed entries, which contributed to memory exhaustion and firmware crashes, especially 7-8am, and in the evening, when most users is active.

### I made:
- Single memory allocation for entire history, history requires 16B per PacketRecord
- Declared size (oldest entries removed if heavy traffic)
- Heap allocations remain only for local function variables and calls.
- No sophisticated techniques - minding it's embedded firmware.

#### I tested this on 4 heltec v.2.1 devices in city network of much more than 100 nodes visible. One source of problem eliminated.

I don't have other devices yet. Please test if possible.

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Heltec (Lora32) V2.1

### DIFF
--- a/.github/workflows/daily_packaging.yml
+++ b/.github/workflows/daily_packaging.yml
@@ -1,7 +1,7 @@
 name: Daily Packaging
 on:
   schedule:
-    - cron: 0 9 * * *
+    - cron: 0 2 * * *
   workflow_dispatch:
   push:
     branches:
--- a/variants/seeed_wio_tracker_L1/variant.cpp
+++ b/variants/seeed_wio_tracker_L1/variant.cpp
@@ -59,8 +59,8 @@ const uint32_t g_ADigitalPinMap[] = {
     // D16 - Battery voltage ADC input
     31, // D16 P0.31 VBAT_ADC
     // GROVE
-    0, // D17 P0.00 GROVESDA
-    1, // D18 P0.01 GROVESCL
+    43, // D17 P0.00 GROVESDA
+    44, // D18 P0.01 GROVESCL
 
     // FLASH
     21, // D19 P0.21 (QSPI_SCK)
--- a/variants/seeed_wio_tracker_L1/variant.h
+++ b/variants/seeed_wio_tracker_L1/variant.h
@@ -74,10 +74,12 @@
 //  Communication Interfaces
 // ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 //  I2C Configuration
-#define HAS_WIRE 1
+// #define HAS_WIRE 1
 #define PIN_WIRE_SDA D14 // P0.09
 #define PIN_WIRE_SCL D15 // P0.10
-#define WIRE_INTERFACES_COUNT 1
+#define WIRE_INTERFACES_COUNT 2
+#define PIN_WIRE1_SDA D18
+#define PIN_WIRE1_SCL D17
 #define I2C_NO_RESCAN
 
 static const uint8_t SDA = PIN_WIRE_SDA;


src/mesh/PacketHistory.cpp
```diff
@@ -7,152 +7,314 @@
...
+    uint32_t now_millis = millis(); // Should not jump with time changes
+    uint32_t rxTimeMsec = 0;
+    PacketRecord *tu = NULL; // Will insert here.
+    PacketRecord *it = NULL;
 
-    for (auto it = recentPackets.begin(); it != recentPackets.end();) {
-        if (!Throttle::isWithinTimespanMs(it->rxTimeMsec, FLOOD_EXPIRE_TIME)) {
-            it = recentPackets.erase(it); // erase returns iterator pointing to element immediately following the one erased
+    // Find a free or oldest slot in the mx_recentPackets array
+    for (it = mx_recentPackets; it < (mx_recentPackets + recentPacketsCapacity); ++it) {
```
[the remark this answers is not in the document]
Marek-mk: This is by design, to have loop go only once.

fifieldt:
Thank you very much for sharing your expertise :)

src/mesh/PacketHistory.h
```diff
@@ -1,49 +1,50 @@
...
-#include <unordered_set>
 
-/// We clear our old flood record 10 minutes after we see the last of it
-#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
-#define FLOOD_EXPIRE_TIME (5 * 1000L) // Don't allow too many packets to accumulate when fuzzing.
-#else
-#define FLOOD_EXPIRE_TIME (10 * 60 * 1000L)
-#endif
+#define PACKETHISTORY_MAX (int)(MAX_NUM_NODES * 2.0) // x2..3  Should suffice. Empirical setup. 16B per record malloc'ed.
+#define FLOOD_EXPIRE_TIME (10 * 60 * 1000L) // Is this needed anymore? If not set to 0.  No change for fuzzing. Should withstand.
```
GUVWAF: I would indeed remove this entirely.

src/mesh/PacketHistory.cpp
```diff
@@ -7,152 +7,314 @@
 #endif
 #include "Throttle.h"
 
-PacketHistory::PacketHistory()
+/** PacketHistory additional debug level selection */
+#define mx_PH_DEBUGLEVEL 0 // Set to 2 for debug messages, 1 for info messages, 0 for no debug messages
```
GUVWAF: Before merging this, I'd prefer to remove the additional debug selection for PacketHistory.

src/mesh/PacketHistory.cpp
```diff
@@ -7,152 +7,314 @@
...
+PacketHistory::PacketHistory(uint32_t size) : recentPacketsCapacity(0), mx_recentPackets(NULL) // Initialize members
+{
+    if (size < 4 || size > PACKETHISTORY_MAX) { // Copilot suggested - makes sense
+        // LOG_ERROR("PH:Invalid size %d, using default %d", size, PACKETHISTORY_MAX);
+        size = PACKETHISTORY_MAX; // Use default size if invalid
+    }
+
+    // Allocate memory for the recent packets array
+    recentPacketsCapacity = size;
+    mx_recentPackets = new PacketRecord[recentPacketsCapacity];
```
GUVWAF: Maybe rename this to just `recentPackets`?

GUVWAF (commented):
Thank you very much for this! Looks good in general, I left three small comments.

src/mesh/PacketHistory.cpp
Marek-mk: Ok. (I prefix with mx_ when mixing mine with not my code)

src/mesh/PacketHistory.cpp
Marek-mk: To remove debugs entirely from code, and leave only the LOG_..s  without mx_ prefix?

GUVWAF: Yes. Please keep only the really informative logs with the normal `LOG_`.

src/mesh/PacketHistory.h
```diff
@@ -1,49 +1,49 @@
...
 #include "NodeDB.h"
-#include <unordered_set>
 
-/// We clear our old flood record 10 minutes after we see the last of it
-#ifdef FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
-#define FLOOD_EXPIRE_TIME (5 * 1000L) // Don't allow too many packets to accumulate when fuzzing.
-#else
-#define FLOOD_EXPIRE_TIME (10 * 60 * 1000L)
-#endif
+#define PACKETHISTORY_MAX (int)(MAX_NUM_NODES * 2.0) // x2..3  Should suffice. Empirical setup. 16B per record malloc'ed.
```
GUVWAF: One other thing that comes to mind, some platforms (thinking of STM32) have a rather low `MAX_NUM_NODES`. I would prefer to have a minimum packet history of e.g. 100, so please take the max. of 100 and MAX_NUM_NODES*2.

Marek-mk: done

src/mesh/PacketHistory.cpp
Marek-mk, line 1: Thanks for cleanup

src/mesh/PacketHistory.h
GUVWAF: It should be `max()` then.

Marek-mk: ups. that was a sunny sunday today.. sorry. fixed now.
I also thought if will be feasible to somehow react if are so flooded, that oldest packet is younger than a few minutes, to maybe start dropping packets? We give no reason to exhaust ram and crash the node here, but the reason of packet history could be impeded by heavily flooding. I tried linking nodes with different radio modes here, to bridge modes during transition from MF to SF, but nodes crash because of memory allocation fail at Router::send. But this is another ghost to hunt.

GUVWAF: Yes, that's an idea but dropping packets too soon might sometimes be worse than retransmitting too often.
I think maybe we should add a warning when the history is full with less than the original 10 min. recency, and maybe track for a while if this occurs often.

Marek-mk: How would you do it? Removed code that I had earlier as an option, but now to display warning?

GUVWAF: I would not delete the packets if they exceed the 10 min., but when you have to delete one because it's full, if the oldest packet was from less than 10 min. ago, show the warning.

Marek-mk: So should I use the code parts using FLOOD_EXPIRE_TIME to just issue LOG_WARN ?

GUVWAF: Yes, I think that should be good. Just one if-statement checking if the packet that gets removed was more recent than 10 min.

Marek-mk: Ok. And I found issue with duplicates taking slots unnecessarily.
SF node at the window, city, nodedb of just 120 entries:

DEBUG | 01:27:12 4907 [Router] Node status update: 120 online, 120 total
Sometimes one less. Nodedb get full in minutes here after clearing.

DEBUG | 01:15:25 4200 [Router] Packet History - insert Reuse slot@ 122/240 age=464315
WARN  | 01:15:25 4200 [Router] Packet History - insert Reused slot age 464s < 600s RECENT_WARN_AGE

Solution I want to apply: When I get *found entry, updating it's age or hops should not search for new slot. I have a pointer to the right entry. I have no more unordered_set so trying to delete record and insert new is not optimal.
I'm working on it and need a day to test it, and now I should go to sleep ;)

thebentern: Great progress! Looking forward to merging this hopefully today when the changes are ready. :-)

Marek-mk: Day went quick working and testing :) Commited moment ago.

Marek-mk:
Nice :)

PR #7041, opened 2025-06-14
More low-entropy keys, and don't issue a false warning when changing …

jp-bennett:


fifieldt (approved):

src/mesh/NodeDB.cpp
```diff
@@ -1777,7 +1777,12 @@ bool NodeDB::checkLowEntropyPublicKey(const meshtastic_Config_SecurityConfig_pub
             memcmp(keyHash, LOW_ENTROPY_HASH12, sizeof(LOW_ENTROPY_HASH12)) == 0 ||
             memcmp(keyHash, LOW_ENTROPY_HASH13, sizeof(LOW_ENTROPY_HASH13)) == 0 ||
             memcmp(keyHash, LOW_ENTROPY_HASH14, sizeof(LOW_ENTROPY_HASH14)) == 0 ||
-            memcmp(keyHash, LOW_ENTROPY_HASH15, sizeof(LOW_ENTROPY_HASH15)) == 0) {
+            memcmp(keyHash, LOW_ENTROPY_HASH10, sizeof(LOW_ENTROPY_HASH15)) == 0 ||
```
fifieldt: Hi, is there a copy paste thing here - should these numbers match?

jp-bennett: Good catch.  Fix coming

PR #7039, opened 2025-06-14
InkHUD DIY builds for ProMicro & Heltec T114

todd-herbert:
Adds PlatformIO envs (`board_level = extra`) for DIY InkHUD builds based on either NRF52 ProMicro or Heltec MeshNode T114.

Based around the common WeAct Studio E-Paper modules (B&W):
* 1.54"
* 2.13" (2025 revision)
* 2.9"
* 4.2"

Display model must be selected by user at build-time. To simplify this, the two environments have been given custom pio build tasks.

custom pio task
```
pio run -e <variant> -t build_weact_154
```
(or specify display model by manually modifying nicheGraphics.h) 

I'll write documentation on the supported hardware and build process shortly.

## 🤝 Attestations
- [x] I have tested that my proposed changes behave as described.
- [x] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck 
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [x] Other (please specify below)
    - NRF52 ProMicro DIY
    - Heltec Mesh Node T114

PR #7045, opened 2025-06-15
Exclude InkHUD DIY build from cppcheck

todd-herbert:
https://github.com/meshtastic/firmware/pull/7039#issuecomment-2973735349

> The [check (heltec-mesh-node-t114-inkhud)](https://github.com/meshtastic/firmware/actions/runs/15657639614/job/44110857216#logs) jobs are failing after this merge @todd-herbert

___

The default build task for these variants fails intentionally:
```cpp
#error If not using a DIY preset, display model and resilience must be set manually
```

Is removing `board_check = true` from the platformio.ini file an appropriate solution here?

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/variants/heltec_mesh_node_t114-inkhud/platformio.ini
+++ b/variants/heltec_mesh_node_t114-inkhud/platformio.ini
@@ -2,7 +2,6 @@
 board_level = extra
 extends = nrf52840_base, inkhud
 board = heltec_mesh_node_t114
-board_check = true
 build_flags = 
   ${nrf52840_base.build_flags}
   ${inkhud.build_flags}


jp-bennett:
I believe there's a way to leave an inline comment for cpp check, to ignore an issue in the following line

todd-herbert:
Oh, good idea, I'll look into that.

todd-herbert:
Okay looks like we're not running cppcheck with `--inline-suppr` right now. Could either add that flag to check-all.sh or add the relevant lines to suppressions.txt, if you'd prefer that to totally removing board_check for this variant?

jp-bennett:
Let's add the inline-suppr flag. I thought it was already there, honestly. I'll go ahead and make the push.

PR #7051, opened 2025-06-16
fix seeed tracker L1 iic pin 

Dylanliacc:
fix seeed tracker L1 iic pin


jp-bennett:
Tested working, will pull manually.

jp-bennett:
Manually applied to Master

PR #7058, opened 2025-06-16
Optimize key event processing and add debounce logic.

whywilson:
Add debounce on PRESS and UP/DOWN key.

  - [x] GAT562
  - [x] Heltec T114
  - [x] Heltec WSL


src/input/UpDownInterruptBase.h
```diff
@@ -27,4 +27,10 @@ class UpDownInterruptBase : public Observable<const InputEvent *>, public concur
     input_broker_event _eventUp = INPUT_BROKER_NONE;
     input_broker_event _eventPressed = INPUT_BROKER_NONE;
     const char *_originName;
+    
+    unsigned long lastUpKeyTime = 0;
+    unsigned long lastDownKeyTime = 0;
+    unsigned long lastPressKeyTime = 0;
+    const unsigned long UPDOWN_DEBOUNCE_MS = 300;
```
jp-bennett: Can we push this value lower? I have a RAK with a rotary knob, and it feels much better with the UPDOWN_DEBOUNCE_MS set to 100. Is that too low for your encoder?

jp-bennett:
Code looks great, and seems to work. Just the one comment.

src/input/UpDownInterruptBase.h
whywilson: How about adding an condition for rotary knob? 
100 is to small for up/down encoder it always trigger twice on press and paging.

jp-bennett: > How about adding an condition for rotary knob? 100 is to small for up/down encoder it always trigger twice on press and paging.

I'm not sure if we can detect at run-time whether the input is rotary vs the up/down switch you're using.  The ideal approach may be to compile in a userprefs value, or to give your device an ENV file, and set the custom timeout there. It might make more sense to discuss in a call. If you drop an email to contact@meshtastic.com, we can set up a time.

whywilson: Please check the latest commit to this pr. 
`moduleConfig.canned_message.rotary1_enabled` can be a good condition check for the updownDebounceMs.

jp-bennett: > Please check the latest commit to this pr. `moduleConfig.canned_message.rotary1_enabled` can be a good condition check for the updownDebounceMs.

Unfortunately the rotary1_enabled setting enables the RotartEncoder driver, not this one. I think we'll have to do something like:

```
#ifndef UP_DOWN_DEBOUNCE_MS
#define UP_DOWN_DEBOUNCE_MS 100
#endif
```

And then for your device, just set the define to what makes sense

whywilson: In many cases, the up-down encoder is added by the user,s so the UP_DOWN_DEBOUNCE_MS parameter should not be written into the configuration of regular devices. What do you think?

jp-bennett: > In many cases, the up-down encoder is added by the user,s so the UP_DOWN_DEBOUNCE_MS parameter should not be written into the configuration of regular devices. What do you think?

Agreed. So by default, it should either be disabled, or set to a short value to avoid causing problems.

jp-bennett: This got closed when unify merged. I'll merge your changes manually

PR #7065, opened 2025-06-18
Log TX power after limits applyng and store it in config

mrekin:
Hi!

There are some behavior which not very clear and I want change it.

Firmware allow to set max tx power for the variant using defines like SX126X_MAX_POWER.
Thouse limits works, but has no effect / can't be checked in serial logs or node configuration
- so user can only hope he used right fw with correct limits and will not burn his hw.

This PR changes tx power logging order ( after limiting ) and store limited value in node config.

I'm not sure if I do changes in best way, because there are additional logic near TX_GAIN_LORA variable and may be storing limited value in config not best idea - **lets disccuss it**.

p/s we start buid nodes using modules with PA (ra-01s-p, ra-01sh-p) and they have limit 5dBm on PA input. In some modules like ra-01sh-p PA gain depends on source voltage (3.3v -5v).  So limiting tx max power seems to be right way to configure variants)

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Custom PCB ESP32C3 + RA-01S-P (sx1268) / ESP32C3 + RA62 (sx1262)

### DIFF
--- a/src/mesh/SX126xInterface.cpp
+++ b/src/mesh/SX126xInterface.cpp
@@ -71,10 +71,6 @@ template <typename T> bool SX126xInterface<T>::init()
 
     limitPower(SX126X_MAX_POWER);
 
-    // Set power in config
-    config.lora.tx_power = power;
-    LOG_INFO("Final Tx power: %d dBm", power);
-
     int res = lora.begin(getFreq(), bw, sf, cr, syncWord, power, preambleLength, tcxoVoltage, useRegulatorLDO);
     // \todo Display actual typename of the adapter, not just `SX126x`
     LOG_INFO("SX126x init result %d", res);


thebentern (approved):

mrekin:
Oh. Copilot give correct advice.  Let me fix)

SQ9PXB:
Hello,

I noticed that after these changes it is not possible to set the power higher than 20dBm.

For example, RA62 has 22dBm, and currently cuts to max 20dBm.

mrekin:
> Hello,
> 
> I noticed that after these changes it is not possible to set the power higher than 20dBm.
> 
> For example, RA62 has 22dBm, and currently cuts to max 20dBm.

Hi. Let me check..

mrekin:
@SQ9PXB  confirmed. Thanks!
I'll make new PR with fixes (or remove logic for saving in config( )

ndoo:
@mrekin Are you sure this wouldn't cause the saved Tx power to keep dropping on devices that do have tx gain defined?

That was the reason I did not save the Tx power after it had been clamped for regulatory gain.

mrekin:
> @mrekin Are you sure this wouldn't cause the saved Tx power to keep dropping on devices that do have tx gain defined?
> 
> That was the reason I did not save the Tx power after it had been clamped for regulatory gain.

Saving removed in #7255 at all.
There are was "hack" in initial commit to do not save if gain > 0, but i think there are must be more pretty way..

ndoo:
> > @mrekin Are you sure this wouldn't cause the saved Tx power to keep dropping on devices that do have tx gain defined?
> 
> > 
> 
> > That was the reason I did not save the Tx power after it had been clamped for regulatory gain.
> 
> 
> 
> Saving removed in #7255 at all.
> 
> There are was "hack" in initial commit to do not save if gain > 0, but i think there are must be more pretty way.. 

Thanks, I tested master HEAD and it seems to work fine on my E22-900M30S :)

PR #7075, opened 2025-06-19
Added option to invert screen on InkHUD

razurac:
This change adds the ability to invert colors on InkHUD.
Settings version had to be increased as the addition of the invertScreen setting caused other options to be falsely set/unset. 

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [X] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - [X] LilyGo T-Echo


src/graphics/niche/InkHUD/Persistence.h
todd-herbert, line 1: Possibly we could make use of the existing `config.display.displaymode` setting, checking for `meshtastic_Config_DisplayConfig_DisplayMode_INVERTED`? Saves having to add a new value / increment the InkHUD settings version.

src/graphics/niche/InkHUD/Renderer.cpp
todd-herbert, line 1: I imagine it's probably negligible, but would be good to get a benchmark to see how long inverting the buffer this way takes, just for peace of mind.

src/graphics/niche/InkHUD/Applets/System/Menu/MenuApplet.cpp
todd-herbert, line 1: If we *do* use `config.display.displaymode` instead, there might be an argument for not predominantly exposing the setting via the on-screen menu. If there is increased display wear, I'd be concerned about people enabling it without being aware of the consequences.

todd-herbert (commented):
Thanks for putting the work into this, it all looks very tidy.
 
I'd be a bit concerned about the possible impact on display health from drawing white on black like this for extended periods, especially on some of the less robust display models. I remember hearing from @markbirss that he once experimented with inverting the colors and found that the display health suffered for it.
 
Another comment also seems to suggest similar ([config.proto](https://github.com/meshtastic/protobufs/blob/c758376d04cf5d3d42de24f9388836a18bae9a76/meshtastic/config.proto#L722)):
```
/*
 * Same as TwoColor, but with inverted top bar. Not so good for Epaper displays
 */
```

If the team *are* happy to go ahead with this, "at user's risk", I won't personally object.

razurac:
@todd-herbert Thanks for the review and comments.
I'll see if I can manage to get some benchmarks done for the inversion.

Regarding tying this to `config.display.displaymode`.
Yeah good point there.

As for the general issue of display health.
I honestly haven't really put in any thoughts in that regards.
Before proceeding I'll wait on @markbirss 's opinion on this matter and  if this can be done as you said it "at own risk" or if it isn't a good idea at all.

todd-herbert:
There's definitely a lot of variation between display models. In my opinion, T-Echo probably has the most robust display of all our E-Ink devices so far. I'll give the code a run tomorrow on some of the weaker displays models and see whether they complain about it or not. It'd definitely be neat to offer this feature, even via the on-screen menu if it's tolerated well enough by the hardware.

src/graphics/niche/InkHUD/Renderer.cpp
razurac: Just did a benchmark by calling millis() before and after the for loop and calculating the difference.
It's always 1-2ms
(Could only test on T-Echo as this is the only e-paper device I got currently)

todd-herbert: I think that's probably more than acceptable. Worst case scenario right now would be a 4.2" display on a ProMicro, which has 3x the pixels of the T-Echo, but that'd still only be 6ms or so.

markbirss (approved):
Im, ok, with having such a feature. (Kindle offer this also for example)

markbirss:
> @todd-herbert Thanks for the review and comments. I'll see if I can manage to get some benchmarks done for the inversion.
> 
> Regarding tying this to `config.display.displaymode`. Yeah good point there.
> 
> As for the general issue of display health. I honestly haven't really put in any thoughts in that regards. Before proceeding I'll wait on @markbirss 's opinion on this matter and if this can be done as you said it "at own risk" or if it isn't a good idea at all.

The particular display had ghosting which was just more visible when inverted. I think a inverted option is a usability feature which is a nice to have.

razurac:
> The particular display had ghosting which was just more visible when inverted. I think a inverted option is a usability feature which is a nice to have.

Alright thanks for your insight @markbirss
What would be your opinion regarding exposing this in the InkHUD options vs. only via app, etc.. ?

todd-herbert:
Personally, at this point I'm thinking we might as well go ahead and expose the setting in the on-screen menu. Why not, hey? On that note, maybe "Invert Color" would be a more precise label than "Invert Screen"?

razurac:
Rewrite my implementation to make use of config.display.displaymode
This saves having to change InkHUD settings version and adds possibility to change setting via other means too.

Small cost of this is we need a helper bool to correctly show current state of setting in the InkHUD options.

razurac:
due to

razurac:
(Reopen due to clicking close by accident)

Rewrote my implementation to make use of config.display.displaymode
This saves having to change InkHUD settings version and adds possibility to change setting via other means too.

Small cost of this is we need a helper bool to correctly show current state of setting in the InkHUD options.
And due to having to call nodeDB->saveToDisk(), changing it takes a little bit longer than before.
But as this isn't something done every time the Options screen renders I believe it is negligible.

todd-herbert:
> Small cost of this is we need a helper bool to correctly show current state of setting in the InkHUD options.

Ah yeah I hadn't considered that issue. It'd be nice if there was a more elegant way to handle it, but nothing easy is coming to mind here either, so let's just go with it.

todd-herbert (approved):
I gave the PR a run today on one of the more fragile display models, and it didn't seem to immediately upset it, so that's reassuring. 👍

razurac:
Sounds great:)
Is there anything I have left to do on my side?

This is pretty much the first time contributing to FOSS and I have yet to learn how to do it all properly.

todd-herbert:
We're happy to have you on board! Shouldn't be anything more you need to do on your end. If the project maintainers are happy with the changes, they'll merge them at the appropriate point in the release cycle.

PR #7079, opened 2025-06-19
Fix position exchange throttling issue

jeremiah-k:
## Summary
This addresses a throttling issue that was preventing the "exchange position" feature from working properly in the Android app.

## What was happening
When a device received a position exchange request (a packet with `want_response: true`), it would sometimes refuse to respond. This happened because the reply logic was checking if ANY position packet had been sent recently, rather than checking specifically for recent replies.

## The issue
The `allocReply()` function was using `lastSentToMesh` for throttling, but this variable tracks all position packets - both regular broadcasts and replies to exchange requests. So if a device had recently sent a position broadcast, it would refuse to respond to exchange requests for 3 minutes.

## What this changes
- Adds a separate `lastSentReply` variable to track only position replies
- Updates the throttling logic to only consider previous replies, not broadcasts
- Keeps the existing 3-minute throttling window to prevent spam
- Preserves all existing behavior for position broadcasts

## Result
Position exchange should now work even if the device recently sent a position broadcast, while still preventing reply spam.

Tested and working on rak4631

Related:
https://github.com/meshtastic/Meshtastic-Android/pull/2169

### DIFF
--- a/src/modules/PositionModule.h
+++ b/src/modules/PositionModule.h
@@ -64,6 +64,7 @@ class PositionModule : public ProtobufModule<meshtastic_Position>, private concu
     bool hasQualityTimesource();
     bool hasGPS();
     uint32_t lastSentToMesh = 0; // Last time we sent our position to the mesh
+    uint32_t lastSentReply = 0;  // Last time we sent a position reply (separate from broadcasts)
 
     const uint32_t minimumTimeThreshold =
         Default::getConfiguredOrDefaultMs(config.position.broadcast_smart_minimum_interval_secs, 30);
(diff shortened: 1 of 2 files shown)


src/modules/PositionModule.h
```diff
@@ -64,6 +64,7 @@ class PositionModule : public ProtobufModule<meshtastic_Position>, private concu
     bool hasQualityTimesource();
     bool hasGPS();
     uint32_t lastSentToMesh = 0; // Last time we sent our position to the mesh
```
GUVWAF: It looks like this variable is not used anymore now.

GUVWAF (commented):
This was originally added upon request in #5914, but I indeed think it makes sense to track specifically position *requests*.

src/modules/PositionModule.h
jeremiah-k: Yeah that seems to be the case.  I removed it.

fifieldt (approved):
Thank you for the nice, clean fix.

thebentern (approved):

PR #7116, opened 2025-06-24
Do not beacon Device telemetry by default anymore

thebentern:
By popular demand, default to no device telemetry by default for meshes that end up being mostly comprised of portable devices. Users can now opt-into device telemetry reporting rather than opt out. 

A couple of notes:
- When a user sets a role to Router for instance, the role specific defaults will be applied. 
- This does not affect reporting over the Phone API every minute.


garthvh (approved):
Great idea!

fifieldt:
Good default.

I have a use case managing a "fleet" of mobile trackers, where tracking battery level is important. Would suggest adding a userPrefs.jsonc so it's easy to build firmware with a different default.

thebentern:
> Good default.
> 
> I have a use case managing a "fleet" of mobile trackers, where tracking battery level is important. Would suggest adding a userPrefs.jsonc so it's easy to build firmware with a different default.

Good idear!

src/mesh/NodeDB.cpp
```diff
@@ -864,6 +866,7 @@ void NodeDB::installRoleDefaults(meshtastic_Config_DeviceConfig_Role role)
     } else if (role == meshtastic_Config_DeviceConfig_Role_SENSOR) {
         owner.has_is_unmessagable = true;
         owner.is_unmessagable = true;
+        moduleConfig.telemetry.device_update_interval = default_telemetry_broadcast_interval_secs;
```
fifieldt, line 869: does meshtastic_Config_DeviceConfig_Role_TRACKER also need this?

thebentern: I was on the fence about it. You talked me into it. I suspect its probably important to see battery levels for tracker devices

thebentern:
Added UserPref per @fifieldt feedback

src/mesh/NodeDB.cpp
```diff
@@ -850,10 +850,12 @@ void NodeDB::installRoleDefaults(meshtastic_Config_DeviceConfig_Role role)
     if (role == meshtastic_Config_DeviceConfig_Role_ROUTER) {
         initConfigIntervals();
         initModuleConfigIntervals();
+        moduleConfig.telemetry.device_update_interval = ONE_DAY;
```
fifieldt: src/mesh/Default.h

has:

#define default_telemetry_broadcast_interval_secs IF_ROUTER(ONE_DAY / 2, 60 * 60)

not sure if you want to leave/update that logic there or put it here ...

fifieldt (approved):

Issue #6960, opened 2025-06-05
[Bug]: HTTPS SSL_new_failed on ESP32 - both in 2.6.4 and 2.6.10


pleasantone:
### Category

WiFi, Hardware Compatibility

### Hardware

T-Beam S3

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.6.10

### Description

Flash Lilygo T-Beam (ESP32) with either 2.6.4 or 2.6.10.
Include the embedded web ui.
Configure Lora region.
Configure the network.
Reboot
Never see the "setting up SSL keys message" appear after first reboot.

Attempt to go to https://<ip-address> of unit.
Never connects.
Check console logs...

Does not appear on a T-Beam Supreme with 2.6.10 or a Station G2 with 2.6.10 which are both esp32s3, while the t-beam is an esp32 iirc.

```
[HTTPS:I] New connection. Socket FID=52
[HTTPS:E] SSL_new failed. Aborting handshake. FID=52
[HTTPS:I] Connection closed. Socket FID=52
[HTTPS:I] New connection. Socket FID=52
[HTTPS:E] SSL_new failed. Aborting handshake. FID=52
[HTTPS:I] Connection closed. Socket FID=52
```

### Relevant log output

```Shell
[HTTPS:I] New connection. Socket FID=52
[HTTPS:E] SSL_new failed. Aborting handshake. FID=52
[HTTPS:I] Connection closed. Socket FID=52
[HTTPS:I] New connection. Socket FID=52
[HTTPS:E] SSL_new failed. Aborting handshake. FID=52
[HTTPS:I] Connection closed. Socket FID=52
```

petraq:
Same here LILYGO® LoRa32 V2.1_1.6 - tested FW 2.6.11 also 2.7.2

```
INFO  | ??:??:?? 1 Creating the certificate. This may take a while. Please wait
INFO  | ??:??:?? 6 Creating the certificate was successful
DEBUG | ??:??:?? 6 Created Private Key: 1190 Bytes
DEBUG | ??:??:?? 6 Created Certificate: 793 Bytes
INFO  | ??:??:?? 6 SSL Cert Ready!
```

```
[HTTPS:I] New connection. Socket FID=52
[...]

PR #7121, opened 2025-06-24
feat: Dynamic light sleep support for ESP32 platform

### LINKED ISSUE #6660: [Feature Request]: Dynamic light sleep support for ESP32 platform
### Platform

ESP32

### Description

**Platform**
ESP32

**Description**
Espressif provides advanced PM support in their ESP-IDF framework, which includes dynamic frequency scaling, dynamic light sleep (which maintains Bluetooth and WiFi connections) and other features. In the standard ESP32 Arduino PIO package, this feature is disabled and all `esp_pm_*` functions return ESP_ERR_NOT_SUPPORTED. Meshtastic has some initial implementation of DFS in `sleep.c`, but it won't work with the default ESP32 Arduino PIO package, and dynamic light-sleep is permanently disabled.

Current implementation of power-saving for ESP32 devices enables light-sleep periodically and is unsuitable for mobile devices as it disconnects device from the phone.

**Use Case**
I was able to implement dynamic light sleep feature PoC for Meshtastic firmware and I got about 30h of uptime on popular Heltec V3 platform with 1200mAh battery (H1 clone). This is HUGE improvement over standard battery life of about 8h. For most of the time, Bluetooth connection was active and I didn't have any issues communicating with other nodes. I consider this feature as stable and production ready as I'm using it with my other ESP32 projects (non-Arduino based).

**Current Workaround**
By replacing ESP-IDF in Arduino package, i was able to enable support for PM configuration and other required SDK configurations, which improve Bluetooth stability when light-sleeping. Relevant repositories are linked below. I also forked Meshtastic repository and created a dedicated branch with all changes, which enable dynamic light-sleep. Proposed changes don't break build for NRF devices (tested on heltec-t114) and also allows for legacy light-sleep functionality, when no support for PM is enabled in the ESP32 Arduino framework (i.e. default).

https://github.com/m1nl/arduino-esp32
https://github.com/m1nl/esp32-arduino-lib-builder
https://github.com/m1nl/meshtastic-firmware/tree/esp_light_sleep

**Benefits**
Meshtastic users would benefit from improved battery life on cheaper ESP32-based devices.

I wanted to open a discussion about this feature before submitting a PR, as it involves replacing the Arduino package. This change can only be applied to tested devices, so a staged roll-out is necessary - similar to the approach taken in the linked repository, which installs the modified package only for the heltec-v3 variant.

These dependencies should ideally be moved under the Meshtastic organization on GitHub, and we can add GitHub Actions to enable automated releases. Note that the default ESP-IDF version (v4.4) used by the mainline package in PlatformIO (provided by Espressif) is unsupported; the same version is currently used in the modified package. An upgrade can be considered at a later stage.

Once there's a decision on adopting the customized Arduino package, I’ll proceed with creating a PR to integrate the changes into the Meshtastic firmware.

m1nl:
fixes #6660 

I've finally implemented the Feature Request I created a few months ago. Since then, I've been testing the dynamic light-sleep functionality and haven't encountered any issues.

In the initial version, legacy light-sleep support with the default Arduino framework could lead to problems after my changes. To address this, I had to rewrite and refactor parts of the code to ensure compatibility with builds compiled using the default PlatformIO Arduino framework.

I tested the updated implementation on a Heltec V3 board in the following scenarios:
- default Arduino, bluetooth, light-sleep disabled
- default Arduino, bluetooth, light-sleep enabled
- default Arduino, bluetooth, light-sleep enabled, force disable support for EXT wake-up
- customized Arduino, bluetooth, light-sleep disabled
- customized Arduino, bluetooth, light-sleep enabled
- customized Arduino, bluetooth, light-sleep enabled, force disable support for EXT wake-up
- customized Arduino, bluetooth disabled, light-sleep enabled
- customized Arduino, WiFi enabled, MQTT enabled, light-sleep enabled

There are a few topics that need discussion:
1) Power Management (PM) Support for ESP32 requires a custom variant of the Arduino framework. I’ve forked it from Espressif’s GitHub repository and am currently hosting it under my personal GitHub account. If these changes are accepted, I’d like to move the repository under the `meshtastic` organization.
2) In my opinion, re-enabling Bluetooth on the ESP32 without restarting is not possible, so NB PowerFSM state seems redundant. I'm open to restoring it if necessary, but I’d appreciate guidance on the intended logic behind this state.
3) I updated the `platformio.ini` files for the hardware variants I was able to test. I shared customized builds with others, who confirmed they were working as expected.
4) I updated the `variant.h` files for the hardware variants which do have 32kHz oscillator, according to Espressif docs, it helps with Bluetooth stability when dynamic light-sleep is enabled

Links to repositiories mentioned in 1) :
- https://github.com/m1nl/arduino-esp32
- https://github.com/m1nl/esp32-arduino-lib-builder

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [x] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


fifieldt:
This is a ton of work, thank you very much

fifieldt:
If you have time .... I "maintain" the GPS code - just wondering what the impact might be there or whether we might need further extensions?

m1nl:
AFAIK dynamic light-sleep in ESP32 platform makes the chip light-sleep when FreeRTOS waits for an event or there is an explicit vTaskDelay call (=~ sleep in Arduino). So most of the functionality shouldn't be impacted as Arduino runs all the code in one loop, which waits different intervals depending on the type of thread (using vTaskDelay under the hood). However, for critical modules or core functionality I added an explicit call to PowerFSM.trigger to exit LS state and allow for at least 300ms of uninterrupted processing time; another approach is to use "preflightSleepObserver" which prevents entering LS state if there is some work pending (like TX queue is not empty). I took a look at GPS code and I think we should exit LS state when we change state to GPS_ACTIVE and prevent entering LS as long as the state doesn't change. I'll take a look at the code and propose some changes soon.

m1nl:
If there is a chance to have this feature merged, I'd appreciate getting some feedback on the additional repos to be hosted under `meshtastic` Github organization. As mentioned above, default Arduino framework provided by Espressif doesn't enable PM functionality at all (which is really bad IMO).

m1nl:
@fifieldt changes committed; device should not go into light-sleep when GPS position is requested (=module is in GPS_ACTIVE state). I'm unable to test as I'm on vacation right now.

m1nl:
hi,

1) Bumping my last question
I'd appreciate getting some feedback on hosting additional repos under meshtastic Github organization.  Default Arduino framework provided by Espressif doesn't enable PM functionality at all. I'm happy to move content of my repos to new ones. There are two repos to be created - arduino-esp32 and esp32-arduino-lib-builder

2) I took a look at the code and it reminded me I did change the WiFi code a bit - I'm enabling power saving by default (regardless of powersaving setting in module options); I think I should revoke that change to ensure compatibility with older versions; let me know what you think

3) I'd appreciate getting feedback on removing NB sleep state - as mentioned in the description for this PR 1) it doesn't work for ESP32 (no way to reenable BT after disabling it without reboot) 2) it doesn't make much sense for NRF platform. However I see it's used in Router so Bluetooth is disabled on NRF platform when device goes into this state - is this expected? IMO this creates some confusion as ESP32 FW behaves differently than NRF one.

src/sleep.cpp
```diff
@@ -68,30 +83,21 @@ RTC_DATA_ATTR int bootCount = 0;
  */
 void setCPUFast(bool on)
 {
-#if defined(ARCH_ESP32) && HAS_WIFI && !HAS_TFT
-
+#if defined(ARCH_ESP32) && !HAS_TFT
+#ifdef HAS_WIFI
```
metala: `#if HAS_WIFI`

metala (changes requested):

src/sleep.cpp
metala: Would it also make sense to have the `HAS_ESP32_DYNAMIC_LIGHT_SLEEP` (and HAS_ESP32_PM_SUPPORT, HAS_32768HZ) as a bool as well, instead of just checking `#ifdef` ?

It's more broad discussion about convention. This has to be decided on project level I guess.

metala:
General critique about having a lot of `assert(res == ESP_OK);`:

While I've already make a boot loop with those asserts because of a wrong setting, the more concerning part is that `assert()` signifies an unrecoverable error and this should never happen, unless the system is in unstable/undefined state.

Surely some of the calls can fail in runtime or due to a wrong setting. Should we then crash the device? There is a known issue here that some ESP32 devices corrupt their device configuration when running on solar power. This not only increases the risk, but also cause the device to become irrecoverable, because of a wrong setting.

Example:
esp_sleep_enable_timer_wakeup() can return ESP_ERR_INVALID_ARG if value is out of range. Therefore if the sleep time (`sleepLeft = config.power.ls_secs * 1000LL - sleepTime;`) gets somehow messed due to a wrong ls_secs setting, the device might be hard to restore through the radio.

I'd recommend an approach that would fail in a recoverable way.

metala:
```
/work/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pio/build/heltec-v2_1/firmware.elf section `.iram0.text' will not fit in region `iram0_0_seg'
/work/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: IRAM0 segment data does not fit.
/work/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: region `iram0_0_seg' overflowed by 776 bytes
collect2: error: ld returned 1 exit status
```

The linker fails on ESP32 (not S3).

variants/esp32s3/heltec_v3/platformio.ini
```diff
@@ -1,10 +1,12 @@
-[env:heltec-v3] 
+[env:heltec-v3]
 extends = esp32s3_base
+platform_packages =
+    platformio/framework-arduinoespressif32 @ https://github.com/m1nl/arduino-esp32/archive/refs/tags/2.0.17+5ae9873e.tar.gz ; disable WiFi IRAM optimizations in ESP-IDF
```
metala: `2.0.17` seems to be a downgrade from the current `framework-arduinoespressif32 @ 3.20017.241212+sha.dcc1105b`

m1nl: To be honest, I think Espressif’s versioning scheme is pretty misleading :) I spent several hours digging into it, so let me shed some light:

1. If you check the [release page](https://github.com/platformio/platform-espressif32/releases/) you will see that current framework is based on **Arduino v2.0.17**, not v3.2.0; so this is a first red flag saying that the semantic version doesn't align with Arduino version in use - additionally, you'll see that release page mentions that Arduino framework is based on IDF v4.4.7 (which is already out of support).
2. If you go to the PlatformIO Registry page for [framework-arduinoespressif32](https://registry.platformio.org/tools/platformio/framework-arduinoespressif32), you'll see that release tag `v3.20017.241212+sha.dcc1105b` doesn't exist in GitHub. However if you trace hash `dcc1105b`, it points to this  [commit](https://github.com/espressif/arduino-esp32/commit/dcc1105b0cf1322a437b354c336f2abf72b7e512) on branch [release/v2.x](https://github.com/espressif/arduino-esp32/compare/release/v2.x). 

What I did to create my [custom repo](https://github.com/m1nl/arduino-esp32/):
- I created a fork of [lib-builder repo](https://github.com/m1nl/esp32-arduino-lib-builder)
- Made release/v4.4 a default branch (to follow ESP-IDF version in use)
- Fixed scripts to ensure build uses compatible versions of esp-dl, esp_littlefs, esp-rainmaker, esp-dsp and tinyusb (it looks they're not really doing automated builds - without these modifications build fails as it tries to pull latest commit from branches which are no longer compatible with ESP-IDF v4.4.x)
- Changed sdkconfigs for esp32 platforms (to enable PM support)
- Built the framework (```./build.sh -i tags/v4.4.8 -A tags/2.0.17```)
- Created a fork of [arduino-esp32](https://github.com/m1nl/arduino-esp32) using branch release/v2.x as a main branch
- Copied binary artifacts from esp32-arduino-lib-builder to arduino-esp32 repository and commited them (this is wrong IMO, but this was the way Espressif was releasing the framework)
- Released [arduino-esp32](https://github.com/m1nl/arduino-esp32/releases/tag/2.0.17%2B5ae9873e)

For consistency, I stuck with the Arduino version number for naming. As for Espressif’s tag style (v3.20017.241212+sha.dcc1105b), I honestly have no idea what logic they’re following :)

I think the build process can be improved by creating a GitHub action which would combine binary output from esp32-arduino-lib-builder upon release of arduino-esp32; binary artifacts should be removed from arduino-esp32 repository.

m1nl:
> ```
> /work/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pio/build/heltec-v2_1/firmware.elf section `.iram0.text' will not fit in region `iram0_0_seg'
> /work/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: IRAM0 segment data does not fit.
> /work/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: region `iram0_0_seg' overflowed by 776 bytes
> collect2: error: ld returned 1 exit status
> ```
> 
> The linker fails on ESP32 (not S3).

This fails because the ESP32 platform has very limited IRAM available, and enabling PM capabilities pushes it beyond the limit. I’ll explore making the ESP-IDF components lighter to reduce memory usage without affecting Meshtastic functionality, but that may not be feasible. In the end, we may have to accept that dynamic light-sleep isn’t compatible with the ESP32.

In newer ESP-IDF versions (5.x compatible with Arduino 3.2.x), there’s a new sdkconfig option CONFIG_ESP_SYSTEM_ESP32_SRAM1_REGION_AS_IRAM, which makes more RAM available as IRAM, allowing the build to succeed. However, adopting this would be a breaking change for the project, since it requires moving from Arduino 2.x to 3.x. I’ve already created and tested such a build with the current Meshtastic version, but this shift is too significant to include in this PR.

m1nl:
> ```
> /work/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: .pio/build/heltec-v2_1/firmware.elf section `.iram0.text' will not fit in region `iram0_0_seg'
> /work/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: IRAM0 segment data does not fit.
> /work/.platformio/packages/toolchain-xtensa-esp32/bin/../lib/gcc/xtensa-esp32-elf/8.4.0/../../../../xtensa-esp32-elf/bin/ld: region `iram0_0_seg' overflowed by 776 bytes
> collect2: error: ld returned 1 exit status
> ```
> 
> The linker fails on ESP32 (not S3).

This fails because the ESP32 platform has very limited IRAM available, and enabling PM capabilities pushes it beyond the limit. I’ll explore making the ESP-IDF components lighter to reduce memory usage without affecting Meshtastic functionality, but that may not be feasible. In the end, we may have to accept that dynamic light-sleep isn’t compatible with the ESP32.

In newer ESP-IDF versions (5.x compatible with Arduino 3.2.x), there’s a new sdkconfig option CONFIG_ESP_SYSTEM_ESP32_SRAM1_REGION_AS_IRAM, which makes more RAM available as IRAM, allowing the build to succeed. However, adopting this would be a breaking change for the project, since it requires moving from Arduino 2.x to 3.x. I’ve already created and tested such a build with the current Meshtastic version, but this shift is too significant to include in this PR.

src/sleep.cpp
m1nl: sure, I'll check these definitions and I can update them to `bool`. I see there is inconsistency about how we use `HAS_XYZ` flags in the project. I will create a separate PR for this.

m1nl:
> General critique about having a lot of `assert(res == ESP_OK);`:
> 
> While I've already make a boot loop with those asserts because of a wrong setting, the more concerning part is that `assert()` signifies an unrecoverable error and this should never happen, unless the system is in unstable/undefined state.
> 
> Surely some of the calls can fail in runtime or due to a wrong setting. Should we then crash the device? There is a known issue here that some ESP32 devices corrupt their device configuration when running on solar power. This not only increases the risk, but also cause the device to become irrecoverable, because of a wrong setting.
> 
> Example: esp_sleep_enable_timer_wakeup() can return ESP_ERR_INVALID_ARG if value is out of range. Therefore if the sleep time (`sleepLeft = config.power.ls_secs * 1000LL - sleepTime;`) gets somehow messed due to a wrong ls_secs setting, the device might be hard to restore through the radio.
> 
> I'd recommend an approach that would fail in a recoverable way.

Understood, I'll check if there is any assertion, which relies on user settings and make them fail safe. However I'd recommend to keep assertions for all calls, which use predefined values (board configuration, etc.). IMO this is the only way to validate the changes with all devices during alpha testing phase. Let me know what you think.

akohlsmith:
I'm testing this on ESP32C3 and so far the testing is looking very good. 

My only nitpick is that I think that there should be a comment in `variants/esp32c3/heltec_esp32c3/variant.h` (and others) which mentions the `HAS_32768HZ` macro and that it should be uncommented if your board has a 32kHz crystal. One of the big issues with supporting so many platforms is that these definitions tend to get buried in code.

e.g. 

```
// uncomment this is your board has a 32kHz crystal
//#define HAS_32768HZ 1
```

src/main.cpp
```diff
@@ -1467,12 +1478,25 @@ void setup()
...
+#endif
+
     // This must be _after_ service.init because we need our preferences loaded from flash to have proper timeout values
     PowerFSM_setup(); // we will transition to ON in a couple of seconds, FIXME, only do this for cold boots, not waking from SDS
     powerFSMthread = new PowerFSMThread();
 
-#if !HAS_TFT
-    setCPUFast(false); // 80MHz is fine for our slow peripherals
+#ifndef ARCH_PORTDUINO
+    auto cpuFast = false;
```
metala, line 1023: just a nitpick, auto in C is a storage-class specifier (e.g. register, static, auto), in C++ is also the bad type inference operator. So thin in C it means it's an `int`, in C++ it's a `bool`.

gxcreator:
I was digging in same direction. Any updates here?

m1nl:
> I was digging in same direction. Any updates here?

I had to switch context to another project. But I plan to revisit this - for now the main task here is to setup Arduino framework repos with Github actions. IMO code is stable and I'm using it for several months right now.

I'd appreciate testing, especially with less common devices - you can use this branch https://github.com/m1nl/meshtastic-firmware/tree/esp_light_sleep_release_stable ; it's rebased on latest stable release (2.7.11).

gxcreator:
@m1nl I've tested that on Heltec wireless paper. For some reason it gives less runtime than original FW tweaked to 40MHz. 
500mAh battery is ~5h on original firmware + 40MHz or ~4h on this one. Scenario is node jusr attached to Wi-Fi and sitting on the desk.
Probably I am missing somethnig, will retest one more time.

m1nl:
> @m1nl I've tested that on Heltec wireless paper. For some reason it gives less runtime than original FW tweaked to 40MHz. 500mAh battery is ~5h on original firmware + 40MHz or ~4h on this one. Scenario is node jusr attached to Wi-Fi and sitting on the desk. Probably I am missing somethnig, will retest one more time.

Thanks! Can you share the changes you did to tune frequency down to 40MHz? Did you just replace 80MHz with 40MHz in the code? Did you remove the part which is forcing max frequency whenever WiFi is enabled? 

Please note you need to enable Power Saving for a node to make it light-sleep; tunables (wake up time / etc.) is ignored.

gxcreator:
@m1nl I am indeed just removed this workaround to max CPU freq when connected to WiFi and just set it to 40MHz. Will share exact change a bit later.

Hmm, I was not enabling Power Saving mode - maybe I got it wrong, but shouldn't tickless mode still function fine(i.e. putting cores to sleep on idle -  when there is no freeRTOS process running) ?

m1nl:
> @m1nl I am indeed just removed this workaround to max CPU freq when connected to WiFi and just set it to 40MHz. Will share exact change a bit later.
> 
> Hmm, I was not enabling Power Saving mode - maybe I got it wrong, but shouldn't tickless mode still function fine(i.e. putting cores to sleep on idle - when there is no freeRTOS process running) ?

I wanted to keep this feature optional; PM config is not enabled unless Power Saving is requested by the user - moreover sleep code is managing PM locks to keep app running without any sleep intervals when busy.

Please enable Power Saving and keep screen on time low, retest and share the results :)

iuvi7:
Volcano work! Any plans for implement for V4? THANKS

m1nl:
thx, you guys motivated me to rebase and update this PR this week; for now you can try to use https://github.com/m1nl/meshtastic-firmware/tree/esp_light_sleep_release_stable - it **should** support V4; I'll check and let you know

gxcreator:
Tried with power save mode, it was 10h+ instead of  ~5h on original firmware + 40MHz !

art-den:
Hi @m1nl. Thanks for your work!

I noticed some strange behavior on the tbeam. The initial data is as follows:
* BT is in use
* WiFi is disabled
* GPS is disabled
* Power saving mode is enabled
* Device supply voltage is 3.7 volts

Immediately after a reset, the device consumes 110 mA for several minutes. This high consumption either decreases on its own (to 56 mA) or if I connect to the device via Bluetooth (to 30 mA).

On original 2.6.11 it consumes 48 mA without power saving enabling

I also tested your firmware on the Heltec-V3 and it works perfectly there.

m1nl:
> Hi @m1nl. Thanks for your work!
> 
> 
> 
> I noticed some strange behavior on the tbeam. The initial data is as follows:
> 
> * BT is in use
> 
> * WiFi is disabled
> 
> * GPS is disabled
> 
> * Power saving mode is enabled
> 
> * Device supply voltage is 3.7 volts
> 
> 
> 
> Immediately after a reset, the device consumes 110 mA for several minutes. This high consumption either decreases on its own (to 56 mA) or if I connect to the device via Bluetooth (to 30 mA).
> 
> 
> 
> On original 2.6.11 it consumes 48 mA without power saving enabling
> 
> 
> 
> I also tested your firmware on the Heltec-V3 and it works perfectly there.

Thanks for testing, I can try to help if share the bootlog (serial output from the board).

art-den:
> I also tested your firmware on the Heltec-V3 and it works perfectly there.

I found strange thing for  Heltec-V3 too. It consumes 10 mA when connected with my phone via BT and 19 mA when not connected. This is strange for me.

@m1nl how better to get bootlog for sharing?

m1nl:
You can send me an email (check my GitHub profile). 

If you didn't try it before, please disable Bluetooth completely (in node settings) and check the current draw; I suspect it may be BT modem looking for your phone causing more frequent wake-ups.

iuvi7:
изображение

This one from Xiao Seed Studio quite popular as well, have few at shelf, will test it later by possibility from one 18650 :)

m1nl:
> изображение
> This one from Xiao Seed Studio quite popular as well, have few at shelf, will test it later by possibility from one 18650 :)

Please keep in mind you may need to modify platformio.ini for a given variant to customize Arduino framework (https://github.com/m1nl/meshtastic-firmware/commit/5b28e11a75ef2b65f54f9698291f2f2e5f632123)

art-den:
> You can send me an email (check my GitHub profile).

Done. Main difference in logs with and without power safe enabling are too much `[GPS]` records on power safe mode. `[GPS]` records go about minute and this minute device consumes 110 mA. Then consumption decreases to 40 mA

PS: GPS is disabled in settings

m1nl:
> > You can send me an email (check my GitHub profile).
> 
> Done. Main difference in logs with and without power safe enabling are too much `[GPS]` records on power safe mode. `[GPS]` records go about minute and this minute device consumes 110 mA. Then consumption decreases to 40 mA
> 
> PS: GPS is disabled in settings

From the logs, I can see that the system is actively searching for the GPS module but eventually stops because no GPS device is found. ("Disabled" in this context means that the system attempts to discover a GPS module but does not actually use it.)

Sleep functionality is quite limited while the device is searching for GPS, which could explain the increased power consumption. Was the GPS receiver disconnected from the board when power saving was enabled? If not, I’ll need to check whether the GPS module is being detected correctly when power saving mode is active.

art-den:
> Was the GPS receiver disconnected from the board when power saving was enabled?

On T-BEAM module GPS is always connected

phaseloop:
@art-den power difference between BT connected/disconnected state is expected as there are different frame timings involved. This is something to be improved anyway with different PR.

src/sleep.cpp
```diff
@@ -78,30 +99,10 @@ RTC_DATA_ATTR int bootCount = 0;
...
-        return;
-    }
-
-// The Heltec LORA32 V1 runs at 26 MHz base frequency and doesn't react well to switching to 80 MHz...
-#if !defined(ARDUINO_HELTEC_WIFI_LORA_32) && !defined(CONFIG_IDF_TARGET_ESP32C3)
-    setCpuFrequencyMhz(on ? 240 : 80);
-#endif
-
+#ifdef ARCH_ESP32
+    uint8_t targetFrequency = on ? 240 : 80;
```
phaseloop, line 102: why remove original code? Some CPU versions support only 160 Mhz and freq should  not be lowered if TFT is defined as it will harm UI responsiveness for MUI

phaseloop:
I have few questions:

- have you explored dynamic light sleep where entering sleep is managed by FreeRTOS scheduler based on scheduler queue and running idle task?

- does DFS in its current state interferes with regular full clock speed operations as it seems there is no code synchronization to indicate something needs full speed lock

- have you explored ModemSleep mode? According to Espressif docs, even without light sleep it's possible to go down to 20 mA power draw using combination of modem sleep settings and DFS without even officially entering sleep mode.

I suspect this whole power save architecture can be greatly simplified by not having separate sleep state and but rather sleeping in short intervals (like 100 ms) when rtos scheduler is idle and device have nothing better to do.

This is why NRF52 does not have officially a light sleep mode but works live in variable latency mode - doing internally something similar to what I described.

phaseloop:
Also - in your fork you say that this commit enables power management but it seems it added support for DSP (Digital Signal Processing):

https://github.com/espressif/arduino-esp32/commit/1685ef30f182428e0e76214083d108d123ee43b7

art-den:
> power difference between BT connected/disconnected state is expected as there are different frame timings involved

@phaseloop thanks for information

art-den:
@m1nl your firmware is available in alternative online flasher here: https://mrekin.duckdns.org/flasher/?src=m1nl (only for several devices)

Many people tested it and inform no issues

m1nl:
> I have few questions:
> 
> * have you explored dynamic light sleep where entering sleep is managed by FreeRTOS scheduler based on scheduler queue and running idle task?
> * does DFS in its current state interferes with regular full clock speed operations as it seems there is no code synchronization to indicate something needs full speed lock
> * have you explored ModemSleep mode? According to Espressif docs, even without light sleep it's possible to go down to 20 mA power draw using combination of modem sleep settings and DFS without even officially entering sleep mode.
> 
> I suspect this whole power save architecture can be greatly simplified by not having separate sleep state and but rather sleeping in short intervals (like 100 ms) when rtos scheduler is idle and device have nothing better to do.
> 
> This is why NRF52 does not have officially a light sleep mode but works live in variable latency mode - doing internally something similar to what I described.

1. Yes, that's what I'm using here (if dynamic light sleep is supported by Arduino framework)
2. There is only PM lock which prevents light-sleep. There is no PM lock for DFS, but I didn't really find any scenario when it should be used.
3. AFAIK modem-sleep is WiFi-specific and is used whenever WIFI_PS_MAX_MODEM is set; my PR sets MIN / MAX_MODEM depending on power save setting.

My PR is supposed to be compatible with both PM-enabled (customized) and regular (non-PM enabled) Arduino framework, so removing sleep states from PowerFSM is not a valid option.

m1nl:
> Also - in your fork you say that this commit enables power management but it seems it added support for DSP (Digital Signal Processing):
> 
> [espressif/arduino-esp32@1685ef3](https://github.com/espressif/arduino-esp32/commit/1685ef30f182428e0e76214083d108d123ee43b7)

I don't really know much about internals of ESP-IDF <-> Arduino build system; I just used the tools they (Espressif) provide to create a custom build of the framework; I'm in a process of moving this to GitHub workflow; in the, end linker uses only parts of the framework which are actually utilized by the app, so having more modules available in the framework doesn't change anything.

src/main.cpp
```diff
@@ -1483,12 +1494,25 @@ void setup()
                                                        1000);
     }
 
+#ifdef ARCH_ESP32
+    if (config.power.is_power_saving) {
```
h3lix1, line 1013: Any reason why we wouldn't always want to do initLightSleep and just let DFS do its thing?

m1nl: Power saving has always been an optional feature in Meshtastic, and some users may expect their node to run at full performance (i.e. when it isn’t operating on battery power).

art-den: I think power saving mode (without disabling BT or WiFi) should be turned on all the time unless it changes the behavior of the device

phaseloop: Are you on discord? I started a thread about modeling power states in "ACPI" way - this will fit nicely all talks about light sleeping, config, etc and make it more easy to manage. It's not technically - just splitting node behavior into several power states and leaving implementation to particular platform.

AFAIK many of that features discussed here are supported by FreeRTOS and ESP-IDF but we need to switch to ESP-IDF v5.5 and pioarduino. And with updated stack it theoretically supports keeping WIFI/BLE/GPIO/interrupts enabled during light sleep.

https://freertos.org/Documentation/02-Kernel/02-Kernel-features/07-Lower-power-support

So essentially instead managing whole light sleep enter/exit implemented as Meshtastic power state it is done in idle task. It just replaces sleeping 30s and waking up for 30s with sleeping for 500 ms and waking up for 500 ms.

Here is the example what ESP-IDF upgrade allows for, without even enabling light sleep:

https://github.com/meshtastic/firmware/issues/8876

m1nl: @phaseloop I open discord from time to time.

I understand your point and agree with you. Based on my previous discussions with the maintainers, given the number of devices Meshtastic currently supports, replacing the framework is not a simple option. I can confirm that IDF 5.5 with the newer Arduino works on most devices after a few tweaks. However, I did encounter a few devices (including Heltec V4) that enter a boot loop for no apparent reason - likely due to the combination of SPIRAM, LittleFS, and memory management. Devices with extensive external peripherals, like the T-Deck, also require extra care, as framework light-sleeping can cause issues (GPIO holds, etc.).

I was able to explain my changes several months ago and was very close to merging this PR, but the framework was a blocking issue. I have since moved the framework packaging to a GitHub Action (https://github.com/m1nl/esp32-arduino-lib-builder) because binary blobs in the GitHub repo were not approved - even though Espressif uses them - so I hope this will help. I also believe that combining a framework upgrade with power-save changes at the same time might not be well-received.

With that in mind, I plan to fix the outstanding issues in this PR, notify the maintainers, and provide ad-hoc support if needed. I won’t have time to keep up with all new PM-specific change requests for this PR.

However, I am happy to work on new features once this PR is merged. In summary, my approach is to take "small steps" that minimize disruption while improving functionality.

phaseloop: oh, that's good to know - I was looking into lib builder myself. Once we migrate we can just replace platform release with our custom build and that's all.

I created a separate issue to track and discuss migration: https://github.com/meshtastic/firmware/issues/8896

m1nl:
Sorry for no activity, I'll rebase today and move it from draft to final state.

m1nl:
@thebentern I spent several days debugging why this PR made SPIRAM-capable devices extremely unstable when using latest develop branch. I was able to narrow the issue down to commit 5910cc2e269cac13275d198434d34be210eb3d9d, which changes the SPIRAM allocation threshold to a very aggressive value of 256 bytes.

We need to keep in mind that SPIRAM is significantly slower than internal RAM, and moving small allocations there can easily cause subtle and hard-to-debug failures. In my case, enabling dynamic light sleep made the device completely unresponsive. With a 256-byte threshold, even very small buffers are pushed to SPIRAM, and this was causing Bluetooth to hang after sleep.

SPIRAM can absolutely be used to offload larger, memory-heavy allocations - but lowering the threshold this much is not the right approach. Based on my experience with ESP32 development, keeping it at 256 bytes will likely introduce instability in components like LittleFS and other subsystems that are already difficult to debug.

For reference:
- The default value in ESP-IDF is 4096 bytes.
- I tested 2048 bytes - which I still consider quite low - and the system appears stable.
- 1024 bytes already causes Bluetooth to break after sleep, with the device hanging.

Given this, I strongly recommend *against* keeping the 256-byte threshold.

----

One comment regarding my changes conflicting with af518fbd2b256eaebb5b23841e3ca9d8892dd791 - I have also noticed that before and I added holds for these pins in my PR some time ago. However I don't understand, why we skip these pins when holds are disabled. We need to keep in mind that GPIO holds should be active only during sleep, so *all of them* have to be disabled after sleeping to avoid issues with normal GPIO usage.

----

The PR itself is now complete, and I’ve moved it to "ready" state. I’d really appreciate getting it merged.

Two things to consider:
- my changes require ESP-IDF / Arduino with custom sdkconfig (https://github.com/m1nl/esp32-arduino-lib-builder) - would it be possible to move my repo under the meshtastic organization? I’ve added automated CI builds using GitHub Actions - everything is transparent now and no binary blobs are hosted in the repository.
- I only enabled the dynamic light sleep framework on devices I was able to physically test

I'm happy to discuss all the changes, I'm available for a longer conversation on discord, just let me now.

akohlsmith:
> @thebentern I spent several days debugging why this PR made SPIRAM-capable devices extremely unstable when using latest develop branch. I was able to narrow the issue down to commit [5910cc2](https://github.com/meshtastic/firmware/commit/5910cc2e269cac13275d198434d34be210eb3d9d), which changes the SPIRAM allocation threshold to a very aggressive value of 256 bytes.
> 
> We need to keep in mind that SPIRAM is significantly slower than internal RAM, and moving small allocations there can easily cause subtle and hard-to-debug failures. In my case, enabling dynamic light sleep made the device completely unresponsive. With a 256-byte threshold, even very small buffers are pushed to SPIRAM, and this was causing Bluetooth to hang after sleep.

Wow, this is amazing debugging. I have run into similar issues on other projects but wasn't paying close enough attention here.

Does this PR allow I2C to come up early enough to enable the use of INAxxx devices to detect battery voltage for deep sleep? This was the biggest source of frustration in my earlier attempts at enabling light sleep for ESP32 and why my PRs were so difficult to manage. The devices I design do not use an ADC channel for battery voltage at all, and I had significant hacks to do be able to read the current/voltage sensor early enough.

akohlsmith:
One other question -- my own version of the espressif-arduino enabled PM for ESP32 (ESP32C3 only) also enables DFS and enables the use of an optional external 32.768kHz crystal. I took a brief look at your branch and didn't see either of these enabled. Is there any reason not to?

thebentern:
@m1nl would your recommendation be to switch to 2048 for now?

variants/esp32s3/heltec_v3/platformio.ini
cpatulea: Sorry for the drive-by comment, I have been interested in esp32 migration (https://github.com/meshtastic/firmware/pull/9122) to pioandroid so that I can set some sdkconfig flags to solve a different problem (https://github.com/meshtastic/firmware/issues/6960).

pioarduino has a feature called 'hybrid compile' which allows to set a subset of sdkconfig options: https://deepwiki.com/pioarduino/platform-espressif32/2.9-sdk-configuration-and-build-flags

This can avoid the need to maintain a esp32-arduino-lib-builder repo, it would build arduino libs on-demand.

(though I found it might be limited and only allow control over a subset of sdk config options: https://github.com/meshtastic/firmware/issues/6960#issuecomment-3808930337)

Do you have any idea if pioarduino hybrid compile can provide the control you need for this?

I think avoiding to maintain a esp32-arduino-lib-builder could be a major benefit. I was planning on helping to port esp32 variants to pioarduino platform.

m1nl:
> @m1nl would your recommendation be to switch to 2048 for now?

I can't provide any specific reference for that particular value, but my tests show that anything less than 2048 breaks Bluetooth when dynamic light-sleep is enabled; I couldn't find any Espressif materials explaining how to optimize this value.

This is my initial implementation of keeping NodeDB in PSRAM, which significantly reduces heap usage:
https://github.com/m1nl/meshtastic-firmware/commit/41eb876085d8e24d38b899ea91d1c649d31ad7e2 ; I can work on that, but I'd like to finish this feature first.

m1nl:
> > @thebentern I spent several days debugging why this PR made SPIRAM-capable devices extremely unstable when using latest develop branch. I was able to narrow the issue down to commit [5910cc2](https://github.com/meshtastic/firmware/commit/5910cc2e269cac13275d198434d34be210eb3d9d), which changes the SPIRAM allocation threshold to a very aggressive value of 256 bytes.
> > We need to keep in mind that SPIRAM is significantly slower than internal RAM, and moving small allocations there can easily cause subtle and hard-to-debug failures. In my case, enabling dynamic light sleep made the device completely unresponsive. With a 256-byte threshold, even very small buffers are pushed to SPIRAM, and this was causing Bluetooth to hang after sleep.
> 
> Wow, this is amazing debugging. I have run into similar issues on other projects but wasn't paying close enough attention here.
> 
> Does this PR allow I2C to come up early enough to enable the use of INAxxx devices to detect battery voltage for deep sleep? This was the biggest source of frustration in my earlier attempts at enabling light sleep for ESP32 and why my PRs were so difficult to manage. The devices I design do not use an ADC channel for battery voltage at all, and I had significant hacks to do be able to read the current/voltage sensor early enough.

Hey :) I think it does - when dynamic light sleep is enabled, all scheduled tasks will take place as requested. ESP-IDF will wake-up the device when next OSThread is to be run, so routines in Power.c should run as expected.

m1nl:
> One other question -- my own version of the espressif-arduino enabled PM for ESP32 (ESP32C3 only) also enables DFS and enables the use of an optional external 32.768kHz crystal. I took a brief look at your branch and didn't see either of these enabled. Is there any reason not to?

1. I think my code does enable DFS - take a look here https://github.com/m1nl/meshtastic-firmware/blob/1df8c24c02f1f13b7a0554d02461d653b09591d1/src/sleep.cpp#L502 .
2. 32.768kHz crystal is enabled for supported devices; there is a flag in `variant.h`, which enables that feature:
```
...
#define HAS_32768HZ 1
```

I believe it may be missing from several variants as defined in repo (i.e. T-Lora Pager, no schematic is available, but it can be enabled successfully).

variants/esp32s3/heltec_v3/platformio.ini
m1nl: Hey, I think it was already discussed - afair we concluded meshtastic uses custom build scripts and this may not work properly. I haven't tested this personally and given how much time I've already spent on this PR I won't be exploring that anytime soon :)

m1nl:
@thebentern is there a chance you could help with the item below?

```
- my changes require ESP-IDF / Arduino with custom sdkconfig (https://github.com/m1nl/esp32-arduino-lib-builder) - would it be possible to move my repo under the meshtastic organization? I’ve added automated CI builds using GitHub Actions - everything is transparent now and no binary blobs are hosted in the repository.
```

I can set up the CI pipeline and migrate the repository contents. After that, I’ll update the PR so all links to the custom Arduino framework point to releases under the Meshtastic GitHub organization.

akohlsmith:
I can confirm that both 32k crystal switchover (as well as failure to switch over) and DFS are active and working with this PR:

```
DEBUG | ??:??:?? 1 32k XTAL OSC has not started up
WARN  | ??:??:?? 1 Failed to switch 32K XTAL RTC source to 32.768kHz !!!

...

DEBUG | ??:??:?? 0 Switch RTC Source to 32.768kHz succeeded, using 32k XTAL

...

INFO  | ??:??:?? 2 initLightSleep() dfsSupported=1, ret=0, set/get: min_freq: 20/20, max_freq 80/80, ls_ena: 1/1
INFO  | ??:??:?? 2 PM config enabled - min_freq_mhz=20, max_freq_mhz=80, light_sleep_enable=1
INFO  | ??:??:?? 2 PowerFSM init, has_power=1, is_power_saving=1, wake_time_ms=10000
```

I still need to test whether the INA219 will work EARLY in the boot (it is detected very early) to drop into deep sleep if the battery voltage is below the threshold, but this is excellent work!

akohlsmith:
I do not see light sleep getting enabled, only periodic AGC reset attempts and messages about gpio_hold_en being attempted on output-capable IOs:

```
[2026-02-25 16:00:41] DEBUG | ??:??:?? 104 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
[2026-02-25 16:00:42] E (106235) gpio: gpio_hold_en(641): Only output-capable GPIO support this function
[2026-02-25 16:00:42] E (106238) gpio: gpio_hold_en(641): Only output-capable GPIO support this function
[2026-02-25 16:00:42] E (106245) gpio: gpio_hold_en(641): Only output-capable GPIO support this function
[2026-02-25 16:00:42] E (106252) gpio: gpio_hold_en(641): Only output-capable GPIO support this function
[2026-02-25 16:00:42] E (106260) gpio: gpio_hold_en(641): Only output-capable GPIO support this function
[2026-02-25 16:01:41] DEBUG | ??:??:?? 164 attempting AGC reset
[2026-02-25 16:01:41] DEBUG | ??:??:?? 224 attempting AGC reset
[2026-02-25 16:02:41] DEBUG | ??:??:?? 284 attempting AGC reset
[2026-02-25 16:03:41] DEBUG | ??:??:?? 344 attempting AGC reset
```

I do have light sleep enabled in the configuration:

```
$ meshtastic --port /dev/cu.usbserial-14101 --get power
Connected to radio
power.is_power_saving: True
power.wait_bluetooth_secs: 60
power.sds_secs: 3600
power.ls_secs: 86400
power.min_wake_secs: 60
power.device_battery_ina_address: 64
Completed getting preferences
```

I also do not see the "Using INA on I2C addr 0x%x for charging detection" message on bootup, although I don't see why I shouldn't: I have an INA219, I have telemetry enabled (I see those messages in the log and being transmitted over LoRa), I'm not excluding environmental sensing and `DISABLE_INA_CHARGING_DETECTION` isn't defined anywhere.

m1nl:
1) GPIO hold errors happen because sb decided to use 255 as placeholder value for TB_* defines instead of just keeping them undefined. This is not critical.

2) You won't see device explicitly entering light sleep as it's only releasing a PM lock in light sleep PowerFSM state so FreeRTOS can put device into sleep when idle. This also allows device to keep Bluetooth connection active with very small impact on overall power consumption.

akohlsmith:
> 1. GPIO hold errors happen because sb decided to use 255 as placeholder value for TB_* defines instead of just keeping them undefined. This is not critical.
> 2. You won't see device explicitly entering light sleep as it's only releasing a PM lock in light sleep PowerFSM state so FreeRTOS can put device into sleep when idle. This also allows device to keep Bluetooth connection active with very small impact on overall power consumption.

yep I just realized that after adding some explicit debugging. I've got trivial patch to fix that since it's cluttering up the debug logs.

Any idea why the INA219 is correctly detected and the configuration does reference it correctly (`power.device_battery_ina_address = 64`) yet it's not being used for charge or battery voltage? That's the last thing I'm running against (but it was also the biggest pain in my ass when I was doing my hack-job of ESP32 light sleep)

m1nl:
Does INA219 get properly detected with stock firmware ? It may be out of scope of my changes but I can take a look later during the day.

akohlsmith:
> Does INA219 get properly detected with stock firmware ? It may be out of scope of my changes but I can take a look later during the day.

yep it's absolutely detected by your branch (and stock) without any changes whatsoever. The power telemetry is also working great:

```
[2026-02-25 16:26:58] INFO  | ??:??:?? 102 [PowerTelemetry] Send: ch1_voltage=0.000000, ch1_current=0.000000, ch2_voltage=0.000000, ch2_current=0.000000, ch3_voltage=3.668000, ch3_current=427.899994
```

akohlsmith:
it looks like the AGC reset seems to never re-enter light sleep -- every minute it attempts an AGC reset, but I never see `doLightSleep()` get called until after a packet is received:

```
[2026-02-25 16:27:00] INFO  | ??:??:?? 104 [PowerFSM] ABK doLightSleep(39146 ms)
[2026-02-25 16:27:00] DEBUG | ??:??:?? 104 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
[2026-02-25 16:27:00] DEBUG | ??:??:?? 164 attempting AGC reset
[2026-02-25 16:28:00] DEBUG | ??:??:?? 224 attempting AGC reset
[2026-02-25 16:29:00] DEBUG | ??:??:?? 284 attempting AGC reset
[2026-02-25 16:30:00] DEBUG | ??:??:?? 344 attempting AGC reset
[2026-02-25 16:31:00] DEBUG | ??:??:?? 404 attempting AGC reset
```

I'm digging into this a little more but wanted to mention it in case you might already know what's going on

m1nl:
> it looks like the AGC reset seems to never re-enter light sleep -- every minute it attempts an AGC reset, but I never see `doLightSleep()` get called until after a packet is received:
> 
> 
> 
> ```
> 
> [2026-02-25 16:27:00] INFO  | ??:??:?? 104 [PowerFSM] ABK doLightSleep(39146 ms)
> 
> [2026-02-25 16:27:00] DEBUG | ??:??:?? 104 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
> 
> [2026-02-25 16:27:00] DEBUG | ??:??:?? 164 attempting AGC reset
> 
> [2026-02-25 16:28:00] DEBUG | ??:??:?? 224 attempting AGC reset
> 
> [2026-02-25 16:29:00] DEBUG | ??:??:?? 284 attempting AGC reset
> 
> [2026-02-25 16:30:00] DEBUG | ??:??:?? 344 attempting AGC reset
> 
> [2026-02-25 16:31:00] DEBUG | ??:??:?? 404 attempting AGC reset
> 
> ```
> 
> 
> 
> I'm digging into this a little more but wanted to mention it in case you might already know what's going on

Please see my comment above (the one about PM lock and FreeRTOS, AGC resets do not matter here).

akohlsmith:
nevermind, I think I just got the point of what you said. 🤦 

If `HAS_ESP32_DYNAMIC_LIGHT_SLEEP` is defined, we aren't actually setting up any kind of light sleep, only acquiring the light sleep lock so FreeRTOS does its thing -- we still enable the wakeup sources but the actual call to `esp_light_sleep_start()` never happens since it's unnecessary.

Now the logs make more sense:

```
[2026-02-25 16:44:49] INFO  | ??:??:?? 473 [PowerFSM] ABK doLightSleep(-1 ms): enter
[2026-02-25 16:44:50] DEBUG | ??:??:?? 473 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
[2026-02-25 16:44:50] INFO  | ??:??:?? 473 [PowerFSM] ABK doLightSleep(-1 ms): exit, wake cause 4 (ESP_SLEEP_WAKEUP_TIMER).
[2026-02-25 16:44:50] DEBUG | ??:??:?? 533 attempting AGC reset
[2026-02-25 16:45:49] DEBUG | ??:??:?? 593 attempting AGC reset
```

What's interesting is that receiving a packet also wakes up with the same wakeCause -- I would have expected it to wake from GPIO.

akohlsmith:
I've also figured out the INA219 not being used for battery level -- it's the same issue as I had when I was doing this last year: the battery level stuff is VERY strongly tied to AnalogBatteryLevel, which isn't at all what I'm doing on my own hardware. 

I'll rely on the brownout tripping deep sleep instead for the time being. I'm excited to see your PR merged, this is very excellent work. Thank you!

cpatulea:
Discord thread: https://discord.com/channels/867578229534359593/1476361026801762508

tdovera:
I’ve been testing the Meshtastic develop branch firmware for several days, compiled with the only modification of using the following in the heltec v3 platformio.ini:

platform_packages =
    platformio/framework-arduinoespressif32 @ https://github.com/m1nl/arduino-esp32/archive/refs/tags/2.0.17+5ae9873e.tar.gz ; disable WiFi IRAM optimizations in ESP-IDF

With this build on the Heltec V3, the LiPo battery current consumption drops from about 100 mA to around 50 mA, and I haven’t noticed any side effects with Bluetooth or WiFi enabled. In my case, with a 1200 mAh LiPo battery, I can finally exceed 20 hours of operation while keeping Bluetooth continuously connected to the phone app.

@m1nl, you’ve done a great job reducing power consumption by enabling some PM features for the ESP32-S3.

By porting the additional dynamic light sleep patches, it’s probably to achieve even better results.

akohlsmith:
I developed a tiny node (Heltec HT-CT62 based) and with my earlier changes it would basically stay at ~11mA which is roughly the limit (SX1262 with LNA on and in RX mode is ~10mA) -- with this PR I'm measuring ~16-18mA but have to verify that it's not because of the environmental sensor, which can draw a few mA due to the heater.

This patch does enable both DFS and DLS -- I tried briefly just `#undef HAS_ESP32_DYNAMIC_LIGHT_SLEEP` in `variant.h` and rebuilding, but it seems to still be building with dynamic light sleep enabled -- will try blowing away the build and building from a clean slate next just to compare, but it should not be a big difference in overall power consumption.

akohlsmith:
I am noticing WAY more frequent wakeups from light sleep than I expected:

```
INFO  | 09:56:25 1665 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4188 pcnt 99
DEBUG | 09:56:25 1666 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
INFO  | 09:56:27 1667 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4184 pcnt 99
DEBUG | 09:56:27 1668 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
INFO  | 09:56:31 1671 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4188 pcnt 99
DEBUG | 09:56:31 1672 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
INFO  | 09:56:35 1675 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4188 pcnt 99
DEBUG | 09:56:35 1676 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
INFO  | 09:56:37 1677 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4184 pcnt 99
DEBUG | 09:56:37 1678 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
INFO  | 09:56:39 1679 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4188 pcnt 99
DEBUG | 09:56:39 1680 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
INFO  | 09:56:41 1681 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4184 pcnt 99
DEBUG | 09:56:41 1682 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
```

I thought it might be related to being connected via bluetooth but after disconnecting and rebooting, I am still seeing the same.

Why is power status waking us up every 2-3 seconds?

akohlsmith:
I also noticed if I disable bluetooth in the settings the node tends to hang after about a minute or two of booting. I have not been able to debug this further yet. Re-enabling bluetooth seems to correct the behaviour (and also seems to have reset the waking up every two seconds which makes no sense at all)

m1nl:
> I’ve been testing the Meshtastic develop branch firmware for several days, compiled with the only modification of using the following in the heltec v3 platformio.ini:
> 
> platform_packages = platformio/framework-arduinoespressif32 @ https://github.com/m1nl/arduino-esp32/archive/refs/tags/2.0.17+5ae9873e.tar.gz ; disable WiFi IRAM optimizations in ESP-IDF
> 
> With this build on the Heltec V3, the LiPo battery current consumption drops from about 100 mA to around 50 mA, and I haven’t noticed any side effects with Bluetooth or WiFi enabled. In my case, with a 1200 mAh LiPo battery, I can finally exceed 20 hours of operation while keeping Bluetooth continuously connected to the phone app.
> 
> @m1nl, you’ve done a great job reducing power consumption by enabling some PM features for the ESP32-S3.
> 
> By porting the additional dynamic light sleep patches, it’s probably to achieve even better results.

Interesting, thanks for feedback! These are the sdkconfig options I've enabled (pretty standard):
https://github.com/m1nl/esp32-arduino-lib-builder/blob/release/v4.4/configs/defconfig.common#L97

m1nl:
> I am noticing WAY more frequent wakeups from light sleep than I expected:
> 
> ```
> INFO  | 09:56:25 1665 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4188 pcnt 99
> DEBUG | 09:56:25 1666 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
> INFO  | 09:56:27 1667 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4184 pcnt 99
> DEBUG | 09:56:27 1668 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
> INFO  | 09:56:31 1671 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4188 pcnt 99
> DEBUG | 09:56:31 1672 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
> INFO  | 09:56:35 1675 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4188 pcnt 99
> DEBUG | 09:56:35 1676 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
> INFO  | 09:56:37 1677 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4184 pcnt 99
> DEBUG | 09:56:37 1678 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
> INFO  | 09:56:39 1679 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4188 pcnt 99
> DEBUG | 09:56:39 1680 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
> INFO  | 09:56:41 1681 [Power] ABK Power::readPowerStatus() hasbattery 1, powered 0, charging 0, mv 4184 pcnt 99
> DEBUG | 09:56:41 1682 [PowerFSM] Setup radio interrupt (GPIO03) with wakeup by GPIO interrupt
> ```
> 
> I thought it might be related to being connected via bluetooth but after disconnecting and rebooting, I am still seeing the same.
> 
> Why is power status waking us up every 2-3 seconds?

It does not happen with my devices. Are you using any of these interrupts below? Power thread interval should 20 seconds at minimum.

```
...
#ifdef EXT_PWR_DETECT
    attachInterrupt(
        EXT_PWR_DETECT,
        []() {
            power->setIntervalFromNow(0);
            runASAP = true;
        },
        CHANGE);
#endif
#ifdef BATTERY_CHARGING_INV
    attachInterrupt(
        BATTERY_CHARGING_INV,
        []() {
            power->setIntervalFromNow(0);
            runASAP = true;
        },
        CHANGE);
#endif
#ifdef EXT_CHRG_DETECT
    attachInterrupt(
        EXT_CHRG_DETECT,
        []() {
            power->setIntervalFromNow(0);
            runASAP = true;
            BaseType_t higherWake = 0;
        },
        CHANGE);
#endif
...
```

m1nl:
> I also noticed if I disable bluetooth in the settings the node tends to hang after about a minute or two of booting. I have not been able to debug this further yet. Re-enabling bluetooth seems to correct the behaviour (and also seems to have reset the waking up every two seconds which makes no sense at all)

Actually it makes sense - without Bluetooth enabled, sleep intervals are much longer and interrupts can be missed unless GPIOs are set to wake up device from sleep. I didn't notice any hang when Bluetooth is disabled - when does it happen? Are you able to provide fragment of logs from device around the time when it hangs?

akohlsmith:
> Actually it makes sense - without Bluetooth enabled, sleep intervals are much longer and interrupts can be missed unless GPIOs are set to wake up device from sleep. I didn't notice any hang when Bluetooth is disabled - when does it happen? Are you able to provide fragment of logs from device around the time when it hangs?

It will hang in the middle of emitting a log message to the serial port -- I haven't made any changes around enabling GPIO for wake from interrupt, but perhaps dynamic light sleep is skipping this part (I see log messages saying otherwise, but every wakeup I've documented has been from a timer, not a GPIO, even when receiving a packet).

I'll re-test of course but was surprised at this. I was relying on VERY long sleeps with my hacked up version of power saving and that was working reliably since I think the sleep interval was something on the order of 10 or even 30 minutes for housekeeping/transmitting telemetry (with GPIO interrupts waking it earlier of course).

iuvi7:
@m1nl Hi! Heltec make revision of their v4 schematic, and released v.4.3 board version

There was huge fix improving RX sensitivity (merge #[9906](https://github.com/meshtastic/firmware/pull/9906)) so on today 2.7.21 ([this](https://github.com/meshtastic/firmware/releases/tag/v2.7.21.1370b23) one) is the only one firmware that works fine on 4.3 boards and allow them TX\RX normally

If it possible to ask m1nl version with powersaving for 2.7.21 ? 🙏 

Thank you

dblaber:
I have been using this PR branches revision on my daily carry for around a month now without issue (and extremely decent battery consumption). Wondering what is remaining  to have this merged, as believe this to be extremely beneficial.

vidplace7:
This needs to be reworked now that the Arduino 3.x work has been merge (to `develop` branch).
- #9122

iuvi7:
@dblaber Totally agree, seems what @vidplace7 mentioned moves this PR even further from merging 😢

m1nl:
Hey, I will try to migrate the changes. Actually I did test them with pioarduino before. But timing is bad for me right now, need at least 2 weeks.

h3lix1:
In an attempt to make a 0.5w solar node for esp32 (I'm glutten for punishment) I've tweaked a few existing features to save power.

The expectation is that this will only work if wifi AND bluetooth is disabled, meaning the only way to manage the node is through remote admin. Even serial will require a reboot of the node. Using a [Debug Mate](https://www.seeedstudio.com/Seeed-Studio-XIAO-Debug-Mate-p-6588.html) with custom firmware to count mAh and mWh, I can get a Xiao Wio ESP32 down to a 1 watt hour per day with very busy mesh traffic. 

This might be good for very specific solar routers. I have this running on a Heltec V4.3 with a estimated 2Wh/day.

The changes are #10570 #10571 #10572 and #10582 .. 

These are the settings I'm working with on with the above changes.

```power.is_power_saving = true
network.wifi_enabled = false
bluetooth.enabled = false
display.screen_on_secs = 2147483647 # translates to 0
power.min_wake_secs = 2147483647 # translates to 0
power.wait_bluetooth_secs = 2147483647  # optional if bluetooth.enabled=false, translates to 0
power.ls_secs = 300 # Wake every once in a while for admin tasks

# Maybe if you want to disable LNA for more power savings.
lora.fem_lna_mode disable # Depends on if node has LNA
```

At the very least, #10582 should be interesting since light sleep does not advance ticks, causing incorrect (inflated) chutil.

gxcreator:
I've ported this to current develop branch and with few tweaks(fixed Wi-Fi PS which is major gain, saved ~80 mA and few light-sleep tweaks) I could reach **~20-30mA** while Wi-Fi connected on Heltec Wireless Paper - while original Meshtastic draws 160 mA! 

@m1nl  patch lowered 160 -> ~120
Screenshot_20260830_193134

Properly enabling auto light-sleep and DFS:
Screenshot_20260830_212109

Enabling Wi-fi PS:
Screenshot_20260830_222241

Some light-sleep tweaks:

Screenshot_20260831_001303


Looks very promising, almost 10x battery life! I still need to sort out light sleep control since meshtastic impl is a bit shaky and then push to another PR probably

Thanks @m1nl for initial work and inspiration!

m1nl:
@gxcreator great; I didn't have time to move it forward (after a few months of rebasing this PR); did you make use of sdkconfig via PlatformIO ? as you can see I was using a custom ESP-IDF + Arduino build and this was a blocker at the time my changes were discussed; I'd be really happy to see your code / PR and can help fixing stuff if needed

gxcreator:
> @gxcreator great; I didn't have time to move it forward (after a few months of rebasing this PR); did you make use of sdkconfig via PlatformIO ? as you can see I was using a custom ESP-IDF + Arduino build and this was a blocker at the time my changes were discussed; I'd be really happy to see your code / PR and can help fixing stuff if needed

I use just
```
custom_sdkconfig =
  ${esp32s3_base.custom_sdkconfig}
  CONFIG_PM_ENABLE=y
  CONFIG_FREERTOS_USE_TICKLESS_IDLE=y

```
in `variants/esp32s3/heltec_wireless_paper/platformio.ini`

m1nl:
> > @gxcreator great; I didn't have time to move it forward (after a few months of rebasing this PR); did you make use of sdkconfig via PlatformIO ? as you can see I was using a custom ESP-IDF + Arduino build and this was a blocker at the time my changes were discussed; I'd be really happy to see your code / PR and can help fixing stuff if needed
> 
> I use just
> 
> ```
> custom_sdkconfig =
>   ${esp32s3_base.custom_sdkconfig}
>   CONFIG_PM_ENABLE=y
>   CONFIG_FREERTOS_USE_TICKLESS_IDLE=y
> ```
> 
> in `variants/esp32s3/heltec_wireless_paper/platformio.ini`

great - looking forward to see your code then

PR #7124, opened 2025-06-25
feat(bug): cache the lines and heights to reduce the overhead

csrutil:
# 👋Summary of Changes

This patch introduces a caching mechanism for text rendering in message display, which should reduce CPU overhead when displaying the same message multiple times. The key changes are:

1. **Fixed an issue when add `-D EXCLUDE_EMOJI` flag**
  ```bash
  src/graphics/draw/MessageRenderer.cpp: In function 'void graphics::MessageRenderer::drawTextMessageFrame(OLEDDisplay*, OLEDDisplayUiState*, int16_t, int16_t)':
  src/graphics/draw/MessageRenderer.cpp:333:20: error: 'now' was not declared in this scope
     float delta = (now - lastTime) / 400.0f;
  ```
2. **Added Text Caching**
   - Implemented a simple caching system using text hashing to avoid redundant line calculation
   - Added static cache variables to store previously rendered lines and their heights

3. **Code Restructuring**
   - Split the message rendering logic into smaller, more focused functions:
     - `generateLines`: Handles text line wrapping
     - `calculateLineHeights`: Determines proper height for each line
     - `renderMessageContent`: Draws the actual content

4. **Performance Optimization**
   - Added simple cache key to avoid regenerating the same content
   - Only recalculates line data when the message content changes
   - Reduced CPU usage during repeated rendering of the same message

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [x] Heltec Wireless Tracker
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

## Videos

https://github.com/user-attachments/assets/50f51a69-4958-43b7-997c-c945f847c60b

### DIFF
--- a/src/graphics/draw/MessageRenderer.cpp
+++ b/src/graphics/draw/MessageRenderer.cpp
@@ -278,7 +278,7 @@ void drawTextMessageFrame(OLEDDisplay *display, OLEDDisplayUiState *state, int16
         // Cache miss - regenerate lines and heights
         cachedLines = generateLines(display, headerStr, messageBuf, textWidth);
         cachedHeights =
-            calculateLineHeights(display, cachedLines, emotes, numEmotes);
+            calculateLineHeights(cachedLines, emotes);
         cachedKey = currentKey;
     }
 
@@ -345,7 +345,9 @@ void drawTextMessageFrame(OLEDDisplay *display, OLEDDisplayUiState *state, int16
 
 std::vector<std::string> generateLines(OLEDDisplay *display,
                                        const char *headerStr,
-                                       const char *messageBuf, int textWidth) {
+                                       const char *messageBuf,
+                                       int textWidth)
+{
     std::vector<std::string> lines;
     lines.push_back(std::string(headerStr)); // Header line is always first
 
@@ -390,9 +392,9 @@ std::vector<std::string> generateLines(OLEDDisplay *display,
     return lines;
 }
 
-std::vector<int> calculateLineHeights(OLEDDisplay *display,
-                                      const std::vector<std::string> &lines,
-                                      const Emote *emotes, int emoteCount) {
+std::vector<int> calculateLineHeights(const std::vector<std::string>& lines,
+                                      const Emote *emotes)
+{
     std::vector<int> rowHeights;
 
     for (const auto &_line : lines) {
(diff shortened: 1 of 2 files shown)


Xaositek:
Thank you @csrutil !! Merged this with @HarukiToreda's code review.

HarukiToreda:
The countdown for how long since the message was received broke, it stays 1s. May need to pull this back since the redraws were necessary for the time to recalculate. @csrutil

csrutil:
Hey @HarukiToreda, nice catch, I can reproduce in my local. @Xaositek Here is the simple amendment for the `cachedLines`. And I think we don't need to update `cachedHeights` every time, bcz it's fixed height. So we just need to add `cachedLines[0] = std::string(headerStr);`. Do i need to create another branch to fix this issue or you can just add it?

Thank you!

```git
> $ firmware % git diff
diff --git a/src/graphics/draw/MessageRenderer.cpp b/src/graphics/draw/MessageRenderer.cpp
index 8bf9906..fb8ab3c 100644
--- a/src/graphics/draw/MessageRenderer.cpp
+++ b/src/graphics/draw/MessageRenderer.cpp
@@ -279,6 +279,15 @@ void drawTextMessageFrame(OLEDDisplay *display, OLEDDisplayUiState *state, int16
         cachedLines = generateLines(display, headerStr, messageBuf, textWidth);
         cachedHeights = calculateLineHeights(cachedLines, emotes);
         cachedKey = currentKey;
+    } else {
+        // Cache hit but update the header line with current time information
+        cachedLines[0] = std::string(headerStr);
+
+       // The header always has a fixed height since it doesn't contain emotes
+       // As per calculateLineHeights logic for lines without emotes:
+       cachedHeights[0] = FONT_HEIGHT_SMALL - 2;
+       if (cachedHeights[0] < 8)
+         cachedHeights[0] = 8; // minimum safety
     }

     // === Scrolling logic ===
@@ -436,4 +445,4 @@ void renderMessageContent(OLEDDisplay *display, const std::vector<std::string> &

 } // namespace MessageRenderer
 } // namespace graphics
```

Xaositek:
@csrutil I just updated the code and tested across devices and this seems to hold up well. Thanks for the quick update.

PR #7125, opened 2025-06-25
support seeed_wio_tracker_L1_eink 

Dylanliacc:
# support for seeed_wio_tracker_L1_eink 

## main feature
- Configured the relevant pins of the e - ink screen.
- Changed the logic for maintaining the life cycle of the e - ink screen to prevent black borders from appearing on the screen after multiple partial refreshes.

  - [ ] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


.vscode/settings.json
```diff
@@ -10,5 +10,6 @@
   },
   "[powershell]": {
     "editor.defaultFormatter": "ms-vscode.powershell"
-  }
```
mverch67: This file should not be committed to github.

PR #7143, opened 2025-06-26
Remove bundling of web-ui from ESP32 devices

thebentern:
Write-up to be released:

> In Meshtastic firmware 2.7, we have made the difficult decision to remove the ESP32 bundling of the Web UI.
> 
> When we first unveiled this feature years ago, Meshtastic had no linux native option available, and one of the value-adds of the Espressif (ESP32) based boards was to host a small Web client application on-device and serve it over the network. This worked somewhat well at the time. However, as time has passed, the firmware has taken on more flash and memory demanding features competing for these resources. The performance of the on-device WebUI has struggled on the ESP32 platform as well as new challenges to package the progressingly feature-rich WebUI onto an extremely limited flash space has become a limiting factor for both the web and firmware development team. 
> We believe that going forward, using either the hosted version on https://client.meshtastic.org, [self-hosting the web-client](https://github.com/meshtastic/web?tab=readme-ov-file#self-host), or bundling on a linux native option such as a Raspberry PI or OpenWRT based device is the better solution for operating the web client. The feature-set of the Web client have progressed recently to the point where it has become impossible to package the contents for devices with limited flash space. Iconography, device images, and internationalization have enhanced the user experience at the cost of overall size. We are also exploring options to make the web client easier to consume via a standalone instance and CLI installer.
> Additionally, we believe that this space can be reclaimed for less niche usage, for such features as increased message persistence in the new BaseUI.


bojan385:
I got into Meshatstic project purchasing and building nodes, just because I wanted to use it throught web interface.
The idea is to connect to node via bluetooth when on move, but when I am in the house to universally connect to node with ANY device in the house through network using web browser - Windows PC, Linux PC, Mac, iPhone, Android, Smart TV .... anything that have web browser, no need to install any app on device, and also use it over the internet when I am at work to monitor activity.
Now when Web UI is removed it is useless for my use case.
I am extremly disappointed with this decision, and I will probably abandon Meshtastic project because of this, and stop recomending Meshtastic to friends.
This is not any kind of threath or anything like that, just very frustrating rant.

vidplace7:
> I got into Meshatstic project purchasing and building nodes, just because I wanted to use it throught web interface. The idea is to connect to node via bluetooth when on move, but when I am in the house to universally connect to node with ANY device in the house through network using web browser - Windows PC, Linux PC, Mac, iPhone, Android, Smart TV .... anything that have web browser, no need to install any app on device, and also use it over the internet when I am at work to monitor activity. Now when Web UI is removed it is useless for my use case. I am extremly disappointed with this decision, and I will probably abandon Meshtastic project because of this, and stop recomending Meshtastic to friends. This is not any kind of threath or anything like that, just very frustrating rant.

This decision wasn't made lightly. We'd welcome you to [install meshtasticd on a Linux device](https://meshtastic.org/docs/hardware/devices/linux-native-hardware/) if the webserver feature is important to you.
It has been preserved on that platform, where resource constraints are much less of an issue.

PR #7146, opened 2025-06-26
Add customizable boot logo based on resolution

vidplace7:
Adds customizable boot logos based upon device screen resolution.

Builds upon #7142 

This platformio script hook determines the display resolution, locates the corresponding `branding/logo_<height>x<width>.png` and loads it onto `data/boot/logo.png`.

DefCon here we come :skull_and_crossbones:


vidplace7:
~~DRAFT until I figure out what to do about devices with physically-rotated displays (T-Deck)~~

bin/platformio-custom.py
```diff
@@ -131,3 +131,43 @@ def esp32_create_combined_bin(source, target, env):
...
+            if isinstance(flag, tuple) and flag[0] == "LGFX_SCREEN_HEIGHT":
+                screen_height = flag[1]
+            elif isinstance(flag, tuple) and flag[0] == "LGFX_SCREEN_WIDTH":
+                screen_width = flag[1]
+        # For rotated devices, swap width and height
+        if env.get("PIOENV") in ["t-deck-tft"]: # Other forms of rotation can be added to this condition
+            return int(screen_height), int(screen_width)
+        # Otherwise, return the width and height as is
+        else:
+            return int(screen_width), int(screen_height)
```
vidplace7: @mverch67 I'd appreciate review for this portion.
I'd like to avoid hard-coding variants here if possible. Could we base this rotation (width/height swap) upon a macro? Perhaps `LGFX_ROTATION`?

mverch67: Maybe instead add a second try in the else part with reversed width/height if the logo_\<width\>x\<height\>.png does not exist?
https://github.com/meshtastic/firmware/blob/abb8b88826d0c266ea3b00322e15e0c4f56f8978/bin/platformio-custom.py#L164

mverch67 (changes requested):
As discussed I think we definitely should _not_ use the SCREEN_ROTATE macro for this. Besides that it's now really ugly to have it moved from variant.h into platformio.ini (which nobody will understand why and which is kinda error prone) this macro is not used by MUI but now should be used together with two other macros to determine the boot logo size.... eek.
I'd rather suggest to use an either explicit (e.g. USE_BOOTLOGO=logo_320x480.png) or implicit (e.g. DISPLAY_SIZE=320x480) macro that helps to handle the size and orientation of the boot logo.

variants/mesh-tab/platformio.ini
```diff
@@ -85,6 +85,7 @@ build_flags = ${mesh_tab_xpt2046.build_flags}
   -D SPI_FREQUENCY=60000000
   -D LGFX_SCREEN_WIDTH=240
   -D LGFX_SCREEN_HEIGHT=320
+  -D DISPLAY_SIZE=240x320 ; portrait mode
```
mverch67: I would say this is most likely landscape (as T-Deck)

variants/mesh-tab/platformio.ini
```diff
@@ -97,6 +98,7 @@ build_flags = ${mesh_tab_xpt2046.build_flags}
   -D SPI_FREQUENCY=60000000 ; if image is distorted then lower to 40 MHz
   -D LGFX_SCREEN_WIDTH=240
   -D LGFX_SCREEN_HEIGHT=320
+  -D DISPLAY_SIZE=240x320 ; portrait mode
```
mverch67: same here

variants/mesh-tab/platformio.ini
```diff
@@ -133,6 +137,7 @@ build_flags = ${mesh_tab_ft5x06.build_flags}
   -D SPI_FREQUENCY=75000000 ; may go higher upto 60/80 MHz
   -D LGFX_SCREEN_WIDTH=240
   -D LGFX_SCREEN_HEIGHT=320
+  -D DISPLAY_SIZE=240x320 ; portrait mode
```
mverch67: also

variants/mesh-tab/platformio.ini
vidplace7: Fixed 👍

variants/mesh-tab/platformio.ini
vidplace7: Fixed 👍

variants/mesh-tab/platformio.ini
vidplace7: Fixed 👍

variants/elecrow_panel/platformio.ini
```diff
@@ -103,6 +104,7 @@ build_flags =
   -D SPI_FREQUENCY=60000000
   -D LGFX_SCREEN_WIDTH=320
   -D LGFX_SCREEN_HEIGHT=480
+  -D DISPLAY_SIZE=480x320 ; landscape mode
```
mverch67: This should be portrait, I just tested it with real hardware.

mverch67 (changes requested):

variants/elecrow_panel/platformio.ini
vidplace7: Fixed :+1:

mverch67 (approved):

PR #7169, opened 2025-06-29
Add brightness settings for nodes with screen

whywilson:
Add brightness (contrast) settings on SSD1306 with 4 level: Low, Medium, High, Very Hight.
Meshtiny
GAT562

  - [x] Meshtiny
  - [x] GAT562

### DIFF
--- a/src/graphics/Screen.h
+++ b/src/graphics/Screen.h
@@ -86,7 +86,11 @@ class Screen
 
 // 0 to 255, though particular variants might define different defaults
 #ifndef BRIGHTNESS_DEFAULT
+#ifdef USE_SSD1306
+#define BRIGHTNESS_DEFAULT 255
+#else
 #define BRIGHTNESS_DEFAULT 150
+#endif
 #endif
 
 // Meters to feet conversion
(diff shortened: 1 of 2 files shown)


HarukiToreda:
Before this change, the default brightness for all nodes was what is now called "Very High." Now, after a reboot, it defaults to "High" and does not save the user's selection. This forces users who were accustomed to the original "Very High" setting to manually change it every time which is not ideal. I suggest reverting the default to its previous value to avoid disrupting the experience for existing users.

whywilson:
> Before this change, the default brightness for all nodes was what is now called "Very High." Now, after a reboot, it defaults to "High" and does not save the user's selection. This forces users who were accustomed to the original "Very High" setting to manually change it every time which is not ideal. I suggest reverting the default to its previous value to avoid disrupting the experience for existing users.

Thank you. Good point.

Xaositek:
@whywilson I've currently got this menu already re-worked in the [2.7-fixes-w2](https://github.com/meshtastic/firmware/pull/7148) branch, is it possible you can PR against this branch which will allow us to review it against some other upcoming items?

whywilson:
> @whywilson I've currently got this menu already re-worked in the [2.7-fixes-w2](https://github.com/meshtastic/firmware/pull/7148) branch, is it possible you can PR against this branch which will allow us to review it against some other upcoming items?

OK. I'll look into it.

PR #7175, opened 2025-06-30
Abandon std::sort in NodeDB, and associated fixes

jp-bennett:
Fixes a few things I found while troubleshooting the std::sort problem. The one additional change I want to propose is to make NodeDB::meshNodes a private member, but that requires changes in inkhud.

This does seem to slow down the initial connection of Bluetooth just a bit, but it's exceptionally safer, so worth it IMHO. There are likely speedups that could be done to offset this.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -966,6 +966,7 @@ void NodeDB::resetNodes()
 
 void NodeDB::removeNodeByNum(NodeNum nodeNum)
 {
+    std::lock_guard<std::mutex> guard(nodeDB_mutex);
     int newPos = 0, removed = 0;
     for (int i = 0; i < numMeshNodes; i++) {
         if (meshNodes->at(i).num != nodeNum)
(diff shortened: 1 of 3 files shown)


jp-bennett:
A surprising number of targets don't support std::mutex at all, so reverted that change.

thebentern:
What about NRF52 bluetooth? Have we seen this occur on NRF devices since the 2.7 changes? I have only observed on a T-Deck. Couldn't even reproduce on a Heltec V3 connected to public msh/US for hours.

jp-bennett:
@thebentern Very possibly needed on the nrf as well. I'm not entirely sure how the read callback will behave on single core devices. There may be a possibility of deadlocks.

mverch67 (commented):
I ran that code with the address sanitizer enabled (meshtasticd) for about an hour with MQTT constantly advertising new nodes and old ones that got purged. No crash issues. Looks good to me. Though I haven't checked for memory leaks.

GUVWAF (commented):
Looks good to me in general. 

To speed up the Bluetooth connection, I think setting `runASAP` (and resetting afterwards) would help, to avoid the Thread controller going to sleep.

To be completely safe I think indeed it should be added to nRF52 as well, and then we could consider moving the small queues to the PhoneAPI, and make it a vector of `toRadio` packets.

PR #4601, opened 2024-08-31
Add CO2 sensing with SCD4X

Coloradohusky:
This PR adds support for CO2 sensing with the SCD40/SCD41 sensors. These sensors also support temperature and relative humidity.

Currently, it is using `startLowPowerPeriodicMeasurement`, which has an update interval of ~30 seconds. However, the datasheet mentions a `measure_single_shot` for lower power use, but says it is limited to SCD41 sensors only. Nonetheless, I tried it on my SCD40 sensor and it appeared to work fine. `measure_single_shot` also includes a delay of 5 seconds for each use.

Average Supply Current:
- High Performance Mode: 11 mA
- Low Power Mode: 2.8 mA
- Idle Single Shot Mode: 1.1 mA for 1 min and 0.36 mA for 5 min interval
- Power-Cycled Single Shot Mode: 0.18 mA for 10 min, 0.09 mA for 20 min, 0.03 mA for 1 hour interval

Other than that, the only differences between SCD40 and SCD41 are a wider CO2 interval and higher accuracy.

See [SCD4x Low Power Operation](https://sensirion.com/media/documents/077BC86F/62BF01B9/CD_AN_SCD4x_Low_Power_Operation_D1.pdf) and the [Datasheet](https://sensirion.com/media/documents/48C4B7FB/64C134E7/Sensirion_SCD4x_Datasheet.pdf) for more information.

Thanks

PR #7190, opened 2025-07-01
Move PMSA003I to separate class and update AQ telemetry

oscgonfer:
This Pull request aims to have a clean `AirQualityTelemetry` module, decoupled from the PMSA003I sensor integration, and update to the latest changes in `master`.

It already contains the work by @fifieldt on https://github.com/meshtastic/firmware/pull/7187. Other efforts have been done as part of other sensor integrations such as this https://github.com/meshtastic/firmware/pull/4601 for the CO2 sensor SCD4X (in this commit https://github.com/meshtastic/firmware/pull/4601/commits/318da220dc08d4bfc4984ac9ff8371fdf04d19dd) or in this other https://github.com/hafu/meshtastic-firmware/tree/add-scd30 to add a SCD30 (in this commit https://github.com/meshtastic/firmware/commit/735b4282c1db8e8ff73d7867515c806a90a0c268).

I open a separate one to be able to test this features separately, and create other branches separate to integrate those sensors (and others as part of the hackathon). The features to test are:

- [x] Check this is the right implementation of the readings for this particular PMS sensor
- [x] Check the logic to send to sleep and wake up the sensor. This is not added in other PRs and it's critical for sensors that have such high consumption as the PM sensors, which draw +200mA or more on boot
- [x] Check the `rescan` logic originally implemented as part of the `AirQualityTelemetry` class, I assume for those sensors that have a long boot time
- [x] Check that this all works with the recent screen changes

Sadly, I don't have a PMSA003I sensor with me at the moment (I have ordered one), but if anyone has one, it'd be great if you can check and/or give feedback. I think the original implementer @thebentern and @fifieldt would have a lot to say. I'll try to be responsive to be able to fix.

NOTE: Since I am making a PR from a separate org github doesn't allow for edits (see this [issue](https://github.com/orgs/community/discussions/5634)).


src/modules/Telemetry/EnvironmentTelemetry.h
```diff
@@ -62,7 +62,6 @@ class EnvironmentTelemetryModule : private concurrency::OSThread, public Protobu
     uint32_t sendToPhoneIntervalMs = SECONDS_IN_MINUTE * 1000; // Send to phone every minute
     uint32_t lastSentToMesh = 0;
     uint32_t lastSentToPhone = 0;
-    uint32_t sensor_read_error_count = 0;
```
oscgonfer, line 70: Note: this is removed because it was never used.

src/modules/Telemetry/AirQualityTelemetry.cpp
```diff
@@ -142,37 +222,23 @@ bool AirQualityTelemetryModule::handleReceivedProtobuf(const meshtastic_MeshPack
...
-
-    LOG_INFO("         | PM1.0(Environmental)=%i, PM2.5(Environmental)=%i, PM10.0(Environmental)=%i",
-             m->variant.air_quality_metrics.pm10_environmental, m->variant.air_quality_metrics.pm25_environmental,
-             m->variant.air_quality_metrics.pm100_environmental);
-
-    return true;
+    m->variant.air_quality_metrics = meshtastic_AirQualityMetrics_init_zero;
+
+    if (pmsa003iSensor.hasSensor()) {
+        // TODO - Should we check for sensor state here?
```
oscgonfer: Here I am not sure if I am approaching it correctly. If the sensor is sleeping, it should wake up first, but I don't know how to handle it. Potentially in getMetrics a check to wakeUp and return if sleeping, then after successful reading to send back to sleep?

src/modules/Telemetry/AirQualityTelemetry.cpp
```diff
@@ -42,79 +60,141 @@ int32_t AirQualityTelemetryModule::runOnce()
...
+    meshtastic_Telemetry telemetry;
+    if (!pb_decode_from_bytes(p.payload.bytes, p.payload.size, &meshtastic_Telemetry_msg, &telemetry)) {
+        display->drawString(x, currentY, "No Telemetry");
+        return;
+    }
+
+    const auto &m = telemetry.variant.air_quality_metrics;
+
+    // Check if any telemetry field has valid data
+    bool hasAny = m.has_pm10_standard || m.has_pm25_standard || m.has_pm100_standard  || m.has_pm10_environmental || m.has_pm25_environmental ||
```
oscgonfer, line 164: I'd reduce to just PM1.0, PM2.5 and PM10.0 here to avoid confusion. `environmental` ones are not commonly understood.

src/modules/Telemetry/AirQualityTelemetry.cpp
```diff
@@ -1,36 +1,54 @@
...
+
+    uint32_t result = UINT32_MAX;
+
     /*
         Uncomment the preferences below if you want to use the module
         without having to configure it from the PythonAPI or WebUI.
     */
 
     // moduleConfig.telemetry.air_quality_enabled = 1;
+    // TODO there is no config in module_config.proto for air_quality_screen_enabled. Reusing environment one, although it should have its own
```
oscgonfer: Should AQ have it's own `_screen_enabled` variable?

src/modules/Telemetry/AirQualityTelemetry.cpp
```diff
@@ -206,7 +272,14 @@ meshtastic_MeshPacket *AirQualityTelemetryModule::allocReply()
 bool AirQualityTelemetryModule::sendTelemetry(NodeNum dest, bool phoneOnly)
 {
     meshtastic_Telemetry m = meshtastic_Telemetry_init_zero;
+    m.which_variant = meshtastic_Telemetry_air_quality_metrics_tag;
+    m.time = getTime();
     if (getAirQualityTelemetry(&m)) {
+        LOG_INFO("Send: pm10_standard=%f, pm25_standard=%f, pm100_standard=%f, pm10_environmental=%f, pm100_environmental=%f",
+                 m.variant.air_quality_metrics.pm10_standard, m.variant.air_quality_metrics.pm25_standard,
+                 m.variant.air_quality_metrics.pm100_standard, m.variant.air_quality_metrics.pm10_environmental,
```
oscgonfer: Same here, I'd remove the environmental altogether (they are not useful for PM monitoring). I now notice I missed `pm25_environmental` from here.

src/modules/Telemetry/AirQualityTelemetry.cpp
thebentern: That sounds like a good approach to me

oscgonfer:
@thebentern I have fixed (I think) comments by `copilot` https://github.com/meshtastic/firmware/pull/7190#pullrequestreview-2978681410

oscgonfer:
Rebased on to master

src/modules/Telemetry/Sensor/PMSA003ISensor.cpp
```diff
@@ -0,0 +1,97 @@
...
+#ifdef PMSA003I_ENABLE_PIN
+void PMSA003ISensor::sleep() {
+    digitalWrite(PMSA003I_ENABLE_PIN, LOW);
+    state = State::IDLE;
+}
+
+uint32_t PMSA003ISensor::wakeUp() {
+    digitalWrite(PMSA003I_ENABLE_PIN, HIGH);
+    state = State::ACTIVE;
+    return PMSA003I_WARMUP_MS
```
hafu:
```suggestion
    return PMSA003I_WARMUP_MS;
```

Missing semicolon ;)

src/modules/Telemetry/Sensor/PMSA003ISensor.h
```diff
@@ -0,0 +1,52 @@
...
+class PMSA003ISensor : public TelemetrySensor
+{
+  private:
+    Adafruit_PM25AQI pmsa003i = Adafruit_PM25AQI();
+    PM25_AQI_Data pmsa003iData = {0};
+
+#ifdef PMSA003I_ENABLE_PIN
+    void sleep();
+    uint32_t wakeUp();
+#endif
```
hafu: Maybe it must be made public? When compiling for the `rak4631`, which defines `PMSA003I_ENABLE_PIN` in [`variants/rak4631/variant.h`](https://github.com/meshtastic/firmware/blob/798b1f4d861e756ecd2324083726c098cfb7325e/variants/rak4631/variant.h#L209)the build fails:

```
src/modules/Telemetry/AirQualityTelemetry.cpp: In member function 'virtual int32_t AirQualityTelemetryModule::runOnce()':
src/modules/Telemetry/AirQualityTelemetry.cpp:78:42: error: 'uint32_t PMSA003ISensor::wakeUp()' is private within this context
   78 |             return pmsa003iSensor.wakeUp();
      |                                          ^
In file included from src/modules/Telemetry/AirQualityTelemetry.cpp:21:
src/modules/Telemetry/Sensor/PMSA003ISensor.h:25:14: note: declared private here
   25 |     uint32_t wakeUp();
      |              ^~~~~~
src/modules/Telemetry/AirQualityTelemetry.cpp:98:30: error: 'void PMSA003ISensor::sleep()' is private within this context
   98 |         pmsa003iSensor.sleep();
      |                              ^
In file included from src/modules/Telemetry/AirQualityTelemetry.cpp:21:
src/modules/Telemetry/Sensor/PMSA003ISensor.h:24:10: note: declared private here
   24 |     void sleep();
      |          ^~~~~
```

Making it public solves the compiling issues. 

```patch
--- a/src/modules/Telemetry/Sensor/PMSA003ISensor.h
+++ b/src/modules/Telemetry/Sensor/PMSA003ISensor.h
@@ -20,11 +20,6 @@ class PMSA003ISensor : public TelemetrySensor
     Adafruit_PM25AQI pmsa003i = Adafruit_PM25AQI();
     PM25_AQI_Data pmsa003iData = {0};

-#ifdef PMSA003I_ENABLE_PIN
-    void sleep();
-    uint32_t wakeUp();
-#endif
-
   protected:
     virtual void setup() override;

@@ -35,6 +30,8 @@ class PMSA003ISensor : public TelemetrySensor
     };

 #ifdef PMSA003I_ENABLE_PIN
+    void sleep();
+    uint32_t wakeUp();
     // the PMSA003I sensor uses about 300mW on its own; support powering it off when it's not actively taking
     // a reading
     // put the sensor to sleep on startup
```

hafu (commented):
I saw your PR and checked out, while compiling for different targets I came across some issues.

src/modules/Telemetry/Sensor/PMSA003ISensor.cpp
oscgonfer: Thank you!

src/modules/Telemetry/Sensor/PMSA003ISensor.h
oscgonfer: Thanks for this too!

oscgonfer:
Merged master and resolved @hafu comments.

src/modules/Telemetry/AirQualityTelemetry.cpp
```diff
@@ -1,36 +1,53 @@
...
-// from the PMSA003I datasheet:
-// "Stable data should be got at least 30 seconds after the sensor wakeup
-// from the sleep mode because of the fan’s performance."
-#define PMSA003I_WARMUP_MS 30000
+#if __has_include(<Adafruit_PM25AQI.h>)
+#include "Sensor/PMSA003ISensor.h"
+PMSA003ISensor pmsa003iSensor;
+#else
+NullSensor pmsa003iSensor;
 #endif
```
hafu, line 47: Building for `wio-e5` target fails:
```
src/modules/Telemetry/AirQualityTelemetry.cpp: In member function 'void AirQualityTelemetryModule::drawFrame(OLEDDisplay*, OLEDDisplayUiState*, int16_t, int16_t)':
src/modules/Telemetry/AirQualityTelemetry.cpp:114:22: error: 'FONT_SMALL' was not declared in this scope
  114 |     display->setFont(FONT_SMALL);
      |                      ^~~~~~~~~~
src/modules/Telemetry/AirQualityTelemetry.cpp:125:27: error: 'FONT_HEIGHT_SMALL' was not declared in this scope
  125 |     const int rowHeight = FONT_HEIGHT_SMALL - 4;
      |                           ^~~~~~~~~~~~~~~~~
```

```suggestion
#include "graphics/ScreenFonts.h"

```

Defines the fonts like in [`src/modules/Telemetry/EnvironmentTelemetry.cpp`](https://github.com/meshtastic/firmware/blob/708978911ba64c97af25751e1b4927a771a5d84d/src/modules/Telemetry/EnvironmentTelemetry.cpp#L204), but could also be included before, for example after/before `graphics/images.h`

oscgonfer:
Hi @thebentern, we have now tested this PR and confirmed it works properly with @Nashui-Yan .

Here is a payload being received at our MQTT broker:

```
{ 'time': 1753439517,
  'air_quality_metrics': { 'pm10_standard': 4,
                           'pm25_standard': 4,
                           'pm100_standard': 5,
                           'pm10_environmental': 4,
                           'pm25_environmental': 4,
                           'pm100_environmental': 5,
                           'particles_03um': 708,
                           'particles_05um': 202,
                           'particles_10um': 15,
                           'particles_25um': 2,
                           'particles_50um': 2,
                           'particles_100um': 0}}
```

There are two pending details we may want to evaluate:

- [ ] We removed the `rescan` as we consistently detected the sensor. Maybe we can consider that as a TelemetryModule feature overall.
- [ ] The screen shows some data "overlapping" (PM10.0 is written on top of PM2.5) - but this is I believe a `screen` issue
- [ ] Maybe we could implement an AdminMessage to choose between pm_standard and pm_environmental. Nobody really knows what these are except probably plantower, and it probably doesn't make sense to use these three fields on the payload every time (not a blocking point for the merge)

src/modules/Telemetry/EnvironmentTelemetry.h
oscgonfer: And it could be removed in other telemetry modules I think...

oscgonfer:
Reworked based on https://github.com/meshtastic/firmware/pull/8054/changes and tested with seeed-xiao-s3.

Also adds `CAN_RECLOCK_I2C` to build flags for `esp32_common` architecture as it's the only one that read and change I2C speed on demand, which many sensors require and that are not compatible with higher I2C bus speeds needed by the oled display. This last aspect is added on the `TelemetrySensor` class, with a logic to check on each architecture.

It needs https://github.com/meshtastic/protobufs/pull/840

oscgonfer:
Not sure where this part should go. Maybe in `src/detect`?

https://github.com/fablabbcn/smartcitizen-meshtastic/blob/0d420ebc26f526b2e06e8f77abd8b8e42e4c429f/src/modules/Telemetry/Sensor/TelemetrySensor.cpp#L10-L44

thebentern:
> Not sure where this part should go. Maybe in `src/detect`?
> 
> https://github.com/fablabbcn/smartcitizen-meshtastic/blob/0d420ebc26f526b2e06e8f77abd8b8e42e4c429f/src/modules/Telemetry/Sensor/TelemetrySensor.cpp#L10-L44

Sounds good to me

oscgonfer:
All builds pass? Is this real life? :eyes:

HarukiToreda:
The BME680 sensor no longer outputs data. Could it be related to this PR. Sensor shows on boot but data never shows either on the logs, phone app, or gets sent through the mesh. I do not have other type of sensor but this may affect others with IAQ.

HarukiToreda:
ignore, looks like a different PR is responsible for it

PR #7213, opened 2025-07-03
Don't set non-existent pin on e290

jp-bennett:


src/graphics/draw/MenuHandler.cpp
```diff
@@ -633,8 +633,7 @@ void menuHandler::BrightnessPickerMenu()
 
         if (selected != 0) { // Not "Back"
                              // Apply brightness immediately
-#if defined(HELTEC_MESH_NODE_T114) || defined(HELTEC_VISION_MASTER_T190) || defined(HELTEC_VISION_MASTER_E213) ||                \
-    defined(HELTEC_VISION_MASTER_E290)
+#if defined(HELTEC_MESH_NODE_T114) || defined(HELTEC_VISION_MASTER_T190) || defined(HELTEC_VISION_MASTER_E213)
```
vidplace7:
```suggestion
#if defined(HELTEC_MESH_NODE_T114) || defined(HELTEC_VISION_MASTER_T190)
```
Vision Master E213 is also failing. Perhaps it needs to be removed here as well?

vidplace7:
`heltec-vision-master-e290` and `heltec-vision-master-e213` compile properly after this change :beers:

vidplace7 (approved):

PR #7234, opened 2025-07-05
Optimize the satellite search time after GPS hot start and reduce pow…

Quency-D:
Optimize the satellite search time after GPS hot start and reduce power consumption.

## 🤝 Attestations

  - [x] heltec_mesh_node_t114

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -54,12 +54,10 @@ HardwareSerial *GPS::_serial_gps = nullptr;
 GPS *gps = nullptr;
 
 static GPSUpdateScheduling scheduling;
-#ifdef GPS_SLEEP_DELAY_TIME
 static uint32_t gps_start_time=0;
 static bool gps_first_lock=false;
 static uint32_t gps_first_lock_time=0;
 static bool gps_need_sleep_delay=false;
-#endif
 
 /// Multiple GPS instances might use the same serial port (in sequence), but we can
 /// only init that port once.
@@ -73,6 +71,8 @@ static struct uBloxGnssModelInfo {
     uint8_t protocol_version;
 } ublox_info;
 
+#define GPS_ENABLE_SLEEP_DELAY_FIX_TIME 20  //time in seconds enable gps delay sleep. If gps lock time greater than GPS_ENABLE_SLEEP_DELAY_FIX_TIME, delay GPS_SLEEP_DELAY_TIME before gps sleep
+#define GPS_SLEEP_DELAY_TIME 20             //time in seconds wait for gps module backup info after locked before sleep
 #define GPS_SOL_EXPIRY_MS 5000 // in millis. give 1 second time to combine different sentences. NMEA Frequency isn't higher anyway
 #define NMEA_MSG_GXGSA "GNGSA" // GSA message (GPGSA, GNGSA etc)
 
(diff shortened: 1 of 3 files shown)


jp-bennett:
What do we gain by delaying GPS sleep? What Backup info are we looking for?

Quency-D:
> What do we gain by delaying GPS sleep? What Backup info are we looking for?

Delayed sleep is designed to give GPS time to save backup information. This backup information is not very clear about the specific content inside the GPS module itself. I only know that this information can be used for GPS hot start to speed up location acquisition time.

fifieldt:
If there is an issue with how we're sleeping GPS it would be nice to find a more general approach using the current framework. As we already have various metrics and timeouts in the code I would prefer to tweak those rather than add another. We probably need to understand the how much time a GPS needs to be on a bit better and do some power testing to support this work.

Quency-D:
> If there is an issue with how we're sleeping GPS it would be nice to find a more general approach using the current framework. As we already have various metrics and timeouts in the code I would prefer to tweak those rather than add another. We probably need to understand the how much time a GPS needs to be on a bit better and do some power testing to support this work.

I have only tested a few L76Ks so far, and it takes longer to save the backup information. You can try the L76K model first.

Quency-D:
If the hot start time exceeds GPS_ENABLE_SLEEP_DELAY_FIX_TIME, it means the signal is not very good. Delay will be performed at this time, and there will be no delay at other times.

RichieFrame:
I wonder if this would also fix the wildly inaccurate GPS positioning on relatively static nodes, where they will move around up to 30ish meters from the actual position everytime it gets a hot lock

Quency-D:
> I wonder if this would also fix the wildly inaccurate GPS positioning on relatively static nodes, where they will move around up to 30ish meters from the actual position everytime it gets a hot lock

It may be this issue, as incomplete information storage leads to positional deviation.

vidplace7:
@Quency-D please rebase this PR. nRF52840 variants have moved to the `variants/nrf52840` directory.

fifieldt:
@Quency-D , haven't forgotten about this :) I just asked @GPSFan on discord in the gnss channel for some advice. He's an expert at how GPS works so can probably help us.

fifieldt:
OK, so the theory is that we may indeed be turning off before ephemeris/almanac is saved. Looking at this more deeply. [Discord link](https://discord.com/channels/867578229534359593/1202456506587418664/1397404501274529894)

fifieldt:
I think based on the information from gpsfan, we can try applying this to every device. I have a range of GPS chips and devices here to test with.

fifieldt:
My testing hitlist:
- [ ] Seeed WM1110 w/ Air530
- [ ] Seeed Tracker 1000e w/ AG3335
- [ ] Heltec Wireless Tracker with UC6580
- [ ] LilyGo TBeam with Ublox 6

fifieldt:
Getting a lot of 
```WARN | 05:13:33 1334 [GPS] SOME data is TOO OLD: LOC 0, TIME 0, DATE 46022```

in the logs


```
(reader.time.age() < GPS_SOL_EXPIRY_MS) && (reader.date.age() < GPS_SOL_EXPIRY_MS))) {
        LOG_WARN("SOME data is TOO OLD: LOC %u, TIME %u, DATE %u", reader.location.age(), reader.time.age(), reader.date.age());
```

need to check that this delay is not interacting with that part of the code.

fifieldt:
This also seems related to the GPS_FIX_HOLD_TIME for the Airoha chips. Might try and combine these ...

PR #7240, opened 2025-07-05
Key verification flow on BaseUI

jp-bennett:
This PR adds the key verification process to BaseUI, making it possible to complete with no connected client.


src/graphics/draw/MenuHandler.cpp
```diff
@@ -869,6 +874,39 @@ void menuHandler::wifiToggleMenu()
...
+
+void menuHandler::keyVerificationFinalPrompt()
+{
+    char message[40] = {0};
+    memset(message, 0, sizeof(message));
+    sprintf(message, "Verification: \n");
+    keyVerificationModule->generateVerificationCode(message + 15); // send the toPhone packet
+
+    if (screen) {
+        static const char *optionsArray[] = {"REJECT", "ACCEPT"};
```
Xaositek: Should be have these as "Reject" and "Accept" to match other menus?

src/modules/KeyVerificationModule.cpp
```diff
@@ -81,21 +86,17 @@ bool KeyVerificationModule::handleReceivedProtobuf(const meshtastic_MeshPacket &
             generateVerificationCode(message + 15);
             static const char *optionsArray[] = {"ACCEPT", "REJECT"};
```
Xaositek: Should this be in other menu it is ordered "Reject" and "Accept" and here flipped. We should be consistent

Xaositek (approved):

PR #7252, opened 2025-07-07
Initial Burning Man 2025 coordinates support.

tschundler:
This cherry picks from https://github.com/exadeci/Meshtastic-device-bm

I then fixed street sizes to correctly match 2025 golden spike data. I had to drop existing display code due to significant changes in display. So currently this only displays _your own_ position.

I have many Meshtastic devices, but they're all in storage, so my only testing is with a t-lora 2.0

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other (please specify below) -> LilyGo T-Lora 2.0


src/graphics/BRC.h
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+    }
+
+    snprintf(addrStr, sizeof(addrStr), "%d:%02d & %d%s", hour, minute, int(d * unitMultiplier), unit);
+    return addrStr;
+}
+
+
+static void drawBRCAddress(OLEDDisplay *display, int16_t x, int16_t y, const GPSStatus *gps)
```
Xaositek: This doesn't appear used any longer - should we still include it?

tschundler: You're right, we should not. Removed.

Also redid the BRC.h code to not use its own static string and accept a pointer.  And operations for radial & annular address are separate so they could also be on separate lines (maybe better on epaper / mui)

I also adjusted the display code to use the address from a phone if there's no GPS or no GPS lock.

vidplace7:
@exadeci can you sign the [Contributor License Agreement](https://cla-assistant.io/meshtastic/firmware?pullRequest=7252) please so this can be merged? 🍻

tschundler:
Oh, it doesn't like my formatting. I'll use trunk.

tschundler:
Added another commit to now the address is also shown on a favorite node's screen. (this replace's the peer's uptime)

PXL_20250710_074352636 RAW-01 COVER

(I was going to send this as a separate PR but figured might as well add to this, since it isn't merged yet)

jp-bennett:
Deleted the bogus second comment from CLAassistant. I also checked on the backend and confirmed that the CLA was signed.

PR #7253, opened 2025-07-07
feat(display): enable screen wake on received messages

whywilson:
Bring the feature on enable/disable screen wakeup when message received.
Related PR on the protocol: https://github.com/meshtastic/protobufs/pull/724

  - [x] Heltec T114
  - [x] GAT562
  - [x] Meshtiny


HarukiToreda:
I definitely see there could be a use case for this. Right now, though, we're trying to keep the action menu options streamlined until we can figure out a more comprehensive approach for adding smaller, non-core features like this. My suggestion would be to hold off on changes like this for now, so we can avoid ending up with an action menu that's cluttered with lots of individual toggles that one button users would need to navigate from if they open the action menu by mistake. If power saving is the issue I would for suggest to lower the screen timeout so the screen turns off quicker.

Xaositek:
Everything here looks sane. Full transparency, we are trying to run down any outstanding bugs before merging in any new features so waiting on this just a bit before it merges in.

Xaositek:
To avoid having a lititany of functions in a single menu a single button user would have to scroll through, talked to @HarukiToreda and discussed the following:

System Action -> Notification Options -> Screen Wake & Beep Actions

We were thinking of nesting "Beep Actions" one level deeper and having one more central "Notification Options" (Notification Opts if too long) which contains the Screen Wake and Beep Actions logically grouped

whywilson:
How about list like this?

System Action
 - Notifications
   - Beep Actions
   - Other settings eg, Buzzer, Viber, LED
 - Screen Options
   - Wakeup
   - Brightness


> To avoid having a lititany of functions in a single menu a single button user would have to scroll through, talked to @HarukiToreda and discussed the following:
> 
> System Action -> Notification Options -> Screen Wake & Beep Actions
> 
> We were thinking of nesting "Beep Actions" one level deeper and having one more central "Notification Options" (Notification Opts if too long) which contains the Screen Wake and Beep Actions logically grouped

Xaositek:
I like it! We'll also want to add the "Screen Color" flag under Screen Options submenu.

I think we keep "Switch to MUI" at the top level, it's only 2-3 devices and we want to keep that available.

> How about list like this?
> 
> System Action
> 
> * Notifications
>   
>   * Beep Actions
>   * Other settings eg, Buzzer, Viber, LED
> * Screen Options
>   
>   * Wakeup
>   * Brightness
> 
> > To avoid having a lititany of functions in a single menu a single button user would have to scroll through, talked to @HarukiToreda and discussed the following:
> > System Action -> Notification Options -> Screen Wake & Beep Actions
> > We were thinking of nesting "Beep Actions" one level deeper and having one more central "Notification Options" (Notification Opts if too long) which contains the Screen Wake and Beep Actions logically grouped

whywilson:
Current menus on my local env. 

System Action
├── Notifications 
├── Screen Options 
├── Reboot 

Notifications
└── Beep Actions 

Screen Options
├── Wakeup 
├── Brightness (Only B&W)
└── Screen Color (Only HAS_TFT)

And to to check if device can switch to MUI? I don't have a device supported it yet.

Xaositek:
I can check - requires a TDeck or SenseCap Indicator

src/graphics/draw/MenuHandler.cpp
```diff
@@ -347,37 +347,25 @@ void menuHandler::homeBaseMenu()
...
+    int options = 1;
 
-    // Check if brightness is supported
-    bool hasSupportBrightness = false;
-#if defined(ST7789_CS) || defined(USE_OLED) || defined(USE_SSD1306) || defined(USE_SH1106) || defined(USE_SH1107) || HAS_TFT
-    hasSupportBrightness = true;
-#endif
+    #if HAS_TFT
+        optionsArray[options] = "Switch to MUI";
+        optionsEnumArray[options++] = MUI;
```
jp-bennett, line 357: The MUI option needs to be in the optionsNumbers enum above, and the two static arrays need to incremented for the extra option.

Xaositek:
@whywilson Overall fantastic work here! Tested on TDeck and T114 without issues.

Also saw the change to add saveUIConfig, great addition there too!

Xaositek:
Moved this PR into a new branch for continual work

PR #7258, opened 2025-07-07
Heltec Wireless Paper, VM-E213 Hardware Revisions

todd-herbert:
Adds InkHUD / BaseUI driver support for the Heltec's recent hardware revisions.
- Wireless Paper V1.1.1
- Wireless Paper V1.2
- Vision Master E213 V1.1

This new hardware will use the existing `heltec-wireless-paper` and `heltec-vision-master-e213` variants. Display model is detected at boot, and the appropriate driver selected.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - Heltec VM-E213
    - Heltec Wireless Paper (Tested by @igorrecioh, thank you!)

### DIFF
--- a/src/graphics/niche/Drivers/EInk/SSD1682.cpp
+++ b/src/graphics/niche/Drivers/EInk/SSD1682.cpp
@@ -1,0 +1,41 @@
+#include "./SSD1682.h"
+
+#ifdef MESHTASTIC_INCLUDE_NICHE_GRAPHICS
+
+using namespace NicheGraphics::Drivers;
+
+SSD1682::SSD1682(uint16_t width, uint16_t height, EInk::UpdateTypes supported, uint8_t bufferOffsetX)
+    : SSD16XX(width, height, supported, bufferOffsetX)
+{
+}
+
+// SSD1682 only accepts single-byte x and y values
+// This causes an incompatibility with the default SSD16XX::configFullscreen
+void SSD1682::configFullscreen()
+{
+    // Define the boundaries of the "fullscreen" region, for the controller IC
+    static const uint8_t sx = bufferOffsetX; // Notice the offset
+    static const uint8_t sy = 0;
+    static const uint8_t ex = bufferRowSize + bufferOffsetX - 1; // End is "max index", not "count". Minus 1 handles this
+    static const uint8_t ey = height;
+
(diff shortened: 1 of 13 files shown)


thebentern (approved):

Quency-D:
One thing to note is that the sleep mode configurations of SSD1680 and SSD1682 are different. SSD1682 cannot continue to refresh quickly after waking up from sleep.
image

todd-herbert:
Apparently I hadn't implemented deep sleep at all. I've added it now, although if similar to SSD1681, it may only save a few µA.

src/graphics/niche/Drivers/EInk/SSD1682.cpp
```diff
@@ -38,4 +38,14 @@ void SSD1682::configFullscreen()
...
 }
 
+// Enter a lower-power state
+// May only save a few µA
+// Not: this is *not* deep sleep mode 2
+// According to Heltec, SSD1682 only supports one deep sleep mode, requiring this exact sequence
+void SSD1682::deepSleep()
+{
+    sendCommand(0x10);
+    sendData(0x11);
```
Quency-D: It is calculated based on bits, so it should be 0x03 instead of 0x11.

todd-herbert: Oh yes, good catch!

todd-herbert:
It seems like SSD1682's deep sleep mode is similar to SSD1681's "deep sleep mode 2"
SSD1681 datasheet

We will not be able to make use of this display's deep sleep, as RAM data (image for differential refresh) is not preserved.

Quency-D:
> We will not be able to make use of this display's deep sleep, as RAM data (image for differential refresh) is not preserved.

Yes, there is indeed this problem, and I noticed it when testing. This part may need to be commented out.

todd-herbert:
I guess right now it's still using the default behavior and (unsuccessfully) attempting to enter deep sleep mode 1. It doesn't seem to cause any problems, but you're probably right, it'd be better to avoid sending the command altogether.

PR #7265, opened 2025-07-08
adding no-GPS target for xiao NRF52840, enabling sensor I2C on D6 and D7

JThramer:
This is a small PR that adds a variant environment for the Xiao NRF52840, allowing usage of I2C sensors via D6 & D7 via an omission of a build flag.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other (Xiao NRF52840 with I2C on `D6` & `D7`)


NomDeTom:
Is there any crossover with https://github.com/meshtastic/firmware/pull/7105 ?

@ndoo  had some ideas on this.

JThramer:
> Is there any crossover with #7105 ?
> 
> @ndoo had some ideas on this.

@NomDeTom It seems like there are some similar moves being done, but @ndoo seems to be doing some extra stuff that specifically concerns the power-amplifier they're using. Perhaps we can consolidate using `extends`.

ndoo:
I can add this into my PR, will take a look tonight.

ndoo:
@JThramer So, I think your variant could be simplified to (in addition to moving it to variants/diy/platformio.ini):

```
[env:seeed_xiao_nrf52840_kit_nogps]
extends = env:seeed_xiao_nrf52840_kit
board_level = extra
build_flags = ${env:seeed_xiao_nrf52840_kit.build_flags} -D PRIVATE_HW
build_unflags = -DGPS_L76K
```

I still think `board_level = extra` & `-D PRIVATE_HW` is the way to go.

[Hardware Model Acceptance Policy](https://meshtastic.org/docs/development/firmware/build/#hardware-model-acceptance-policy) is a bit ambiguous about what constitutes a DIY device, but the part I am specifically worried about is that if you do not include those definitions, then they will appear in the Web Flasher and GitHub release assets.

@thebentern or someone else from the Meshtastic team could probably give you a definitive direction on that, or Seeed Studio @Dylanliacc might have to do some discussions separately with the Meshtastic team on whatever agreement they have for the existing variants they officially support.

ndoo:
On the other hand, if the devs and Seeed are happy to fold it into the existing variant, then the simplified lines I provided (less the DIY related definitions) will do

thebentern:
If I'm understanding the intention correctly, I believe it can still use HW_MODEL that exists for the other kit (no need for PRIVATE_HW) as it still targets this same hardware but it would use a new variant perhaps? It would also not be delivered in the web flasher, as there needs to be a single variant delivered there.

Kucharskov:
That's quite strange. I bought a board (Xiao nRF52840 + Wio-SX1262) that looks identical to the one in the Flasher image. After installing the official software, I2C does not work there – neither the screen nor the sensor, but GPS support is enabled, which is not included in the kit by default.

I also had to remove the “GPS_L76K” flag to disable GPS and restore I2C. This option should be available immediately in Flasher so that the appropriate version can be uploaded without any hassle.

PR #7279, opened 2025-07-09
Test zero hop

OverkillFPV:
First iteration of the zero hop count router, does not include any filter at this stage


OverkillFPV:
Lol i didn't mean to do that

PR #7285, opened 2025-07-09
Stop preventing sleep on low battery for NRF52

### LINKED ISSUE #4378: [Bug]: nrf52 boards are not entering deep sleep when battery is critically low (make system_off mode work on nrf52)

Joseph-DiGiovanni:
This PR removes code that inhibits low battery sleep on NRF52-based nodes. The "freezing" noted in the [comment](https://github.com/meshtastic/firmware/commit/92fd5511ecb82bf312aeb2324b3f6411675a71d5) when it was added in 2022 is not well documented, or explained anywhere else as far as I can find. <strike>More importantly, it doesn't appear to be happening anymore.</strike> It would seem to me that its not only safe, but necessary to remove this workaround and prevent batteries from over-discharging.

EDIT: I've updated the PR to only avoid low battery sleep when necessary. See my [update comment](https://github.com/meshtastic/firmware/pull/7285#issuecomment-3076518179) for details.

Fixes #4378

## 🤝 Attestations

  - [ ] <strike>Heltec (Lora32) V3</strike> N/A
  - [ ]  <strike>LilyGo T-Deck</strike> N/A
  - [ ]  <strike>LilyGo T-Beam</strike> N/A
  - [x] RAK WisBlock 4631

I've also tested this PR cherrypicked to the latest beta release (v2.6.11)

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -871,12 +871,16 @@ void Power::readPowerStatus()
             LOG_DEBUG("Low voltage counter: %d/10", low_voltage_counter);
             if (low_voltage_counter > 10) {
 #ifdef ARCH_NRF52
-                // We can't trigger deep sleep on NRF52, it's freezing the board
-                LOG_DEBUG("Low voltage detected, but not trigger deep sleep");
-#else
-                LOG_INFO("Low voltage detected, trigger deep sleep");
-                powerFSM.trigger(EVENT_LOW_BATTERY);
+                // NRF52 doesn't wake from super deep sleep for some reason
+                // Only trigger it if Super Deep Sleep Seconds is UINT32_MAX
+                if (config.power.sds_secs != UINT32_MAX) {
+                    LOG_DEBUG("Low voltage detected, but not triggering deep sleep");
+                } else
 #endif
+                {
+                    LOG_INFO("Low voltage detected, trigger deep sleep");
+                    powerFSM.trigger(EVENT_LOW_BATTERY);
+                }
             }
         } else {
             low_voltage_counter = 0;


caveman99:
What was happening is the node would not boot back up when sufficient power was applied, especially on a slowly rising level like you would get from a solar panel. Can you confirm this is not happening any more?

Joseph-DiGiovanni:
> node would not boot back up when sufficient power was applied

@caveman99 This doesn't seem to be the expected behavior. The [code](https://github.com/meshtastic/firmware/blob/39716ed1baacbcdabd744dfbb6534a3dbbbbc495/src/PowerFSM.cpp#L68) tells the device to sleep for `config.power.sds_secs` which [defaults](https://github.com/meshtastic/firmware/blob/39716ed1baacbcdabd744dfbb6534a3dbbbbc495/src/mesh/Default.h#L18) to one day for routers and `UINT32_MAX` (effectively infinite time) otherwise.

If you've witnessed a node wake from the low battery state, it was likely a router that woke up the next day to check the battery level rather than waking _because_ of the battery level. With that in mind, I can confirm my WisBlock mini _does not_ correctly wake after `config.power.sds_secs` for unknown reasons, which is likely what they mean by "freezing".

Perhaps allowing super deep sleep when `sds_secs` is the `UINT32_MAX` default is a good compromise? That way nodes expecting to wake up regularly to recheck voltage (like routers) will not fall into this bugged infinite sleep while all other users that expect an infinite sleep will get exactly that :)

Joseph-DiGiovanni:
Now we'll only allow super deep sleep if "Super Deep Sleep Seconds" is set to the default of `UINT32_MAX`. Then it doesn't matter if the board freezes since it wouldn't be waking up for over 136 years anyway.

This avoids the known deep sleep bug on NRF52 while preserving the intended behavior for most users.

garthvh:
This was very problematic, can you verify this is not happening anymore?

Joseph-DiGiovanni:
Please read my previous comment. The latest version of this PR should make the board freezing issue irrelevant.

Kazagumo:
tested on my device,seems work

Joseph-DiGiovanni:
Can this be merged please? This simple fix is far overdue for something that could prevent battery damage. At this point I've shipped 80+ devices with this PR cherrypicked to 2.6.11 with no issue, so I'm fairly confident it works as intended.

Joseph-DiGiovanni:
@garthvh If you have an issue we can discuss it. Closing my PR without reading it is ridiculous.

jp-bennett:
I think the missing bit of information here is that RAK boards are intended to wake from deep sleep when the battery power exceeds 3.7v, regardless of what super deep sleep seconds is set to. 

See https://forum.rakwireless.com/t/how-can-i-automatically-recovery-a-wisblock-box-from-a-low-battery-scenario/11352/3 for Bernd, a RAK engineer describing exactly the bug in question.

Joseph-DiGiovanni:
@jp-bennett reading that thread it seems to imply there is a hardware issue rather than software. The current expected behaviour is to sleep X seconds regardless of what board is in use. In the case of solar nodes just set SDS seconds to any non max value (or one day as is the current default for routers) and the low battery sleep is bypassed avoiding freezing.

On a side note, it seems strange to prioritize the uptime of solar nodes with undersized batteries/panels over the battery health and safety of all nodes.

alexl83:
I'm running this changes (cherry-picked)  on 2.6.11 latest beta on my heltecs T114 and rak4631 and everything runs fine

viric:
I think that the hang you see not respecting SDS time is expected in a ROUTER node:
https://github.com/meshtastic/firmware/blob/develop/src/platform/nrf52/main-nrf52.cpp#L358

The device sleeps only if TRACKER/SENSOR. Otherwise it's powered off and will only be powered back on user button press.

viric:
@Joseph-DiGiovanni in this my PR (https://github.com/meshtastic/firmware/pull/7882) I do more than using the deep sleep, FYI.

Joseph-DiGiovanni:
Closing in favor of #7882. Thank you for the awesome work @viric

PR #7297, opened 2025-07-10
GPS for STM32WL

Stary2001:
This PR turns on GPS for STM32 - mostly just turning on some ifdefs.
It all seems to work OK after increasing the serial receive buffer size, I arbitrarily picked 256, which should fit ~3 NMEA sentences at a time.

I think it's time to turn on DEBUG_MUTE for good, though...
Before PR: `Flash: [========= ]  94.5% (used 220516 bytes from 233472 bytes)`
Before PR with DEBUG_MUTE: `Flash: [========  ]  77.1% (used 180100 bytes from 233472 bytes)`
After this PR: `Flash: [========= ] 93.5% (used 218300 bytes from 233472 bytes)`

I also took the chance to refactor the GPS serial selection so we don't have to change it if there's another STM32 variant with it on a different Serial instance - a new `GPS_SERIAL_PORT` define overrides it if present.
It would be _nice_ if someone with a RAK2560+GPS could make sure it really didn't break anything but I don't really see how.

## 🤝 Attestations
  - [x] Wio-E5 dev board


thebentern (approved):

GUVWAF:
Very nice. Only doubt I have is whether GPS should be enabled by default since most (all?) STM32WL targets don't have an on-board GPS module. Having no serial logs at all makes working with it rather hard.

Stary2001:
Yeah, IMO it makes sense to turn off GPS by default. It doesn't fit with DEBUG_MUTE off without compiling out the entire GPS module (with MESHTASTIC_EXCLUDE_GPS), though, so we'd have to specify it in every STM32 variant except those that have GPS :/

PR #7311, opened 2025-07-12
STM32: Properly ignore OneButton

vidplace7:
Properly ignore OneButton on STM32 platform.

~This saves \~15% flash utilization.~ I got excited too quickly, this was a result of `-DDEBUG_MUTE`


vidplace7:
@Stary2001

PR #7321, opened 2025-07-13
Nodelist with Playa addresses for burningmesh

tschundler:
- move the heading list on oled devices from its own page to the carousel
 - new "BRC" page in the UI, a single-column nodelist like this:
PXL_20250712_220721680 RAW-01 MP COVER

real address will have less text like "4:40&C-30" (30 ft towards the man from C)

---

this has more copy-paste than I'd normally be comfortable with, but I think more refactor to reduce duplication will make upstream merges more difficult. The refactors (eg single function to display time with units) should be in the main branch.

---

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other: t-lora v2

IMPORTANT: This is more likely to introduce bad UI than https://github.com/meshtastic/firmware/pull/7252

### DIFF
--- a/src/graphics/draw/NodeListRenderer.cpp
+++ b/src/graphics/draw/NodeListRenderer.cpp
@@ -354,10 +354,12 @@ void drawEntryBRC(OLEDDisplay *display, meshtastic_NodeInfoLite *node, int16_t x
     auto xText = x + ((isHighResolution) ? 6 : 3);
     display->drawString(xText, y, nodeName);
 
-    char buf[14] = "";
-    BRCAddress(node->position.latitude_i, node->position.longitude_i).compact(buf, 14);
-    auto nameWidth = display->getStringWidth("WWWW"); // Fixed width so they are aligned.
-    display->drawString(xText + nameWidth, y, buf);
+    if (nodeDB->hasValidPosition(node)) {
+        char buf[14] = "";
+        BRCAddress(node->position.latitude_i, node->position.longitude_i).compact(buf, 14);
+        auto nameWidth = display->getStringWidth("WWWW") - 2; // Fixed width so they are aligned.
+        display->drawString(xText + nameWidth, y, buf);
+    }
 
     if (node->is_favorite) {
         if (isHighResolution) {
@@ -626,11 +628,7 @@ void drawNodeListWithCompasses(OLEDDisplay *display, OLEDDisplayUiState *state, 
 
 void drawBRCList(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t x, int16_t y)
 {
-    auto ourNode = nodeDB->getMeshNode(nodeDB->getNodeNum());
-    double lat = DegD(ourNode->position.latitude_i);
-    double lon = DegD(ourNode->position.longitude_i);
-
-    drawNodeListScreen(display, state, x, y, "BRC", drawEntryBRC, drawLastSeenExtra, 0, lat, lon, 1);
+    drawNodeListScreen(display, state, x, y, "BRC", drawEntryBRC, drawLastSeenExtra, 0, 0, 0, 1);
 }
 
 /// Draw a series of fields in a column, wrapping to multiple columns if needed


tschundler:
...something is wrong with this that is shows commits already in the repo with different IDs

tschundler:
rebased - now the commits are correct.

tschundler:
I noticed it includes nodes with no location at 12:49&407101ft
So I should fix the list to exclude those (maybe an update tomorrow night?) 

~Please do review, but don't commit yet.~ (fixed in 17-July commits)

tschundler:
>  if we created a standalone frame for all the BRC navigational points that we called on demand.

I'm not sure what you mean. What do you mean by "frame" (page/screen something else?), "navigational points" (your friends' locations on playa?) and "on demand"

There's a new page in the list of available screens. How would you get to it other than stepping through the list of screens?

> results in menu systems unintentionally being available

Long-press opens the node favorite menu, which was intended - it's a variant on the node list, so that made sense. Or did you mean some other menu?

---

I'd be inclined to make the screen classes and not have separate places for icon vs menu vs title vs render - all just class methods, etc. It would be just a new class added to a list building up the available screens. But that's a bigger refactor for the main branch someday not by me.

Who knows where the  code will actually go. (maybe you do?) 

I suspect next year it would be likely re-integrated from scratch - that's why BRC.h/cpp is separate - I expect that will still be valid next year, but not the rest. And the rest I took the approach I did hoping it would be likely to easily merge the next months' changes without getting in the way. But I don't really know what is in the pipeline..

Xaositek:
Looking at the code, I think we're really close; I would leave the bearings where it currently is and overall just add:
`fsi.positions.nodelist_brc = numframes;
normalFrames[numframes++] = graphics::NodeListRenderer::drawBRCList;
indicatorIcons.push_back(icon_bm);`

I would also remove this line as the Node Action frame doesn't apply to the BRC bits (the contents of the menu wouldn't affect the BRC screen).
` || this->ui->getUiState()->currentFrame == framesetInfo.positions.nodelist_brc`

src/graphics/Screen.cpp
```diff
@@ -1386,7 +1390,8 @@ int Screen::handleInputEvent(const InputEvent *event)
                            this->ui->getUiState()->currentFrame == framesetInfo.positions.nodelist_hopsignal ||
                            this->ui->getUiState()->currentFrame == framesetInfo.positions.nodelist_distance ||
                            this->ui->getUiState()->currentFrame == framesetInfo.positions.nodelist_hopsignal ||
-                           this->ui->getUiState()->currentFrame == framesetInfo.positions.nodelist_bearings) {
+                           this->ui->getUiState()->currentFrame == framesetInfo.positions.nodelist_bearings ||
+                           this->ui->getUiState()->currentFrame == framesetInfo.positions.nodelist_brc) {
```
Xaositek, line 1394: I would pull out the nodelist_brc condition

src/graphics/Screen.cpp
```diff
@@ -905,7 +905,7 @@ void Screen::setFrames(FrameFocus focus)
 
 // Show detailed node views only on E-Ink builds
 #ifdef USE_EINK
-    fsi.positions.nodelist_lastheard = numframes;
+    fsi.positions.nodelist_brc = numframes;
```
Xaositek: I would undo this to leave it as the default definition

src/graphics/Screen.cpp
```diff
@@ -916,11 +916,15 @@ void Screen::setFrames(FrameFocus focus)
...
     indicatorIcons.push_back(icon_distance);
-#endif
-#if HAS_GPS
+
     fsi.positions.nodelist_bearings = numframes;
     normalFrames[numframes++] = graphics::NodeListRenderer::drawNodeListWithCompasses;
     indicatorIcons.push_back(icon_list);
+#endif
+#if HAS_GPS
+    fsi.positions.nodelist_bearings = numframes;
```
Xaositek: Change this to nodelist_brc

Xaositek (changes requested):

src/graphics/Screen.cpp
tschundler: oops, I think that was a mistake from experimenting.

src/graphics/Screen.cpp
tschundler: why? I put it here because it is a nodelist, and it makes sense to use features like setting a favorite like the other nodelists.

It's showing where others (nodes) are at the event. You might favorite some to be on the top.

Xaositek: This one is fine - upon further review we can leave it.

ews:
Hi @tschundler and @Xaositek, I've implemented the "additive" fix that was discussed. The changes are in this single commit:

https://github.com/ews/firmware/commit/9b257e8c02bee4429c8d9268a3962ed3638cdea8

I've also created a PR to your fork : https://github.com/tschundler/Meshtastic-device/pull/3

This restores the separate Bearings screen and adds the BRC screen without modifying core functionality. I am absolutely not familiar with this code, so please take a look and suggest changes if needed.

tschundler:
So you find that a better user experience?

Moving the bearings to the rotating pages was a choice for a better UX with fewer pages to step through on a 1-button interface, and less to visually scan in the tabs at the bottom of the screen.

But if others think it's really better with the extra screen to click through, let's use ews' version.

Xaositek:
> So you find that a better user experience?
> 
> Moving the bearings to the rotating pages was a choice for a better UX with fewer pages to step through on a 1-button interface, and less to visually scan in the tabs at the bottom of the screen.
> 
> But if others think it's really better with the extra screen to click through, let's use ews' version.

The goal here was to ensure the long term simplification of the BRC additions to the firmware. That way next year, if we have merge conflicts or other issues, we have decent parity with the firmware itself and only managing a couple changes directly within the BRC frame.

PR #7326, opened 2025-07-13
Adjust Seeed Wio Tracker L1 E-Ink variant

todd-herbert:
* modify display driver for better performance
* rename variant, to use the standard -inkhud suffix
* general tidying

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - Seeed Wio Tracker L1 (E-Ink)


caveman99 (changes requested):
Thanks for the PR. To stay in line with the toher boards, and selector in the webflasher, would it be possible to add a non-inkhud (i.e. baseUI) variant to this PR?

I know there never was one, but it would add the consistency.

rcarteraz:
@todd-herbert Echoing @caveman99's comment. My thinking is BaseUI should really be the default experience and InkHUD is the optional firmware available for those that want that experience.

todd-herbert:
I'm somewhat familiar with that codebase and can't say I'm personally interested in supporting it.

Xaositek:
Just popping in from the BaseUI side to offer what assistance I can! @todd-herbert What do you need from me or what can I do to help get this over the finish line?

PR #7341, opened 2025-07-14
Map report should work over devices which do not have network hardware (with client proxy)

thebentern:


garthvh:
Yay, this was confusing

PR #7357, opened 2025-07-16
To add VN_433 and VN_923 for Vietnam

IoTThinks:
This PR is to add VN_433 and VN_923 for Vietnam.

Main document: https://vanban.chinhphu.vn/default.aspx?pageid=27160&docid=204286 In Page 16: https://datafiles.chinhphu.vn/cpp/files/vbpq/2021/10/08-btttt.signed.pdf The accepted frequencies by Vietnam government is
- Item 39: 433.05 - 434.79 MHz <=25 mW ERP
- Item 45: 920 - 923 MHz <=25 mW ERP

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -170,6 +170,15 @@ const RegionInfo regions[] = {
     RDEF(KZ_433, 433.075f, 434.775f, 100, 0, 10, true, false, false), RDEF(KZ_863, 863.0f, 868.0f, 100, 0, 30, true, false, true),
 
     /*
+        Vietnam
+        Main document: https://vanban.chinhphu.vn/default.aspx?pageid=27160&docid=204286
+        Page 16: https://datafiles.chinhphu.vn/cpp/files/vbpq/2021/10/08-btttt.signed.pdf
+        Item 39: 433.05 - 434.79 MHz <=25 mW ERP
+        Item 45: 920 - 923 MHz <=25 mW ERP
+    */
+    RDEF(VN_433, 433.05f, 434.79f, 100, 0, 14, true, false, false), RDEF(VN_923, 920.0f, 923.0f, 100, 0, 14, true, false, false),
+
+    /*
        2.4 GHZ WLAN Band equivalent. Only for SX128x chips.
     */
     RDEF(LORA_24, 2400.0f, 2483.5f, 100, 0, 10, true, false, true),


IoTThinks:
This is the related PR in protobufs: https://github.com/meshtastic/protobufs/pull/730
Thanks a lot.

caveman99 (changes requested):
According to APPENDIX 19 of https://extendmax.vn/vietnam-mic-circular-08-2021-tt-btttt-list-of-equipment-exempt-from-frequency-license devices that want to use more than 10mW need to adhere to a duty cycle of 10% for base stations and 1% for clients. you can not operate meshtastic with these severe restrictions. Unfortunately we have to reject your request in the current form since it will violate local law.

fifieldt:
Original Appendix 19:
image

Translated version:

image

IoTThinks:
Let me double check and I may use another frequencies.
I will revisit this issue in a few days.
Thanks a lot.

IoTThinks:
Usually, we follow Singapore's guideline for LPWAN.
So let me compare the proposed VN_923 to SG_923.
This is to find a way and still comply to the local law.

In Meshtastic, SG_923 sets the duty cycle to 100% and power limit to 20dBm.
https://github.com/meshtastic/firmware/blob/master/src/mesh/RadioInterface.cpp#L151

image

In item 30c for LPWAN, the duty cycle for LPWAN is 1% too. Same as VN_923.

https://www.imda.gov.sg/-/media/imda/files/regulation-licensing-and-consultations/ict-standards/telecommunication-standards/radio-comms/imdatssrd.pdf

image

Meshtastic allows SG_923 to use 100% duty cycle because may be the proposer used item 30d to have no restriction to the duty cycle?
image

Then in the same way, **VN_923 can use item 44 (instead of item 45) to have 100% duty cycle like SG_923**?
https://datafiles.chinhphu.vn/cpp/files/vbpq/2021/10/08-btttt.signed.pdf
image

Item 44 in https://extendmax.vn/vietnam-mic-circular-08-2021-tt-btttt-list-of-equipment-exempt-from-frequency-license
image

What do you think?
Thanks a lot and have a nice day.

fifieldt:
Thanks for the work, I follow the logic.

I guess it depends on the definition of "Short Range Device". Based on what I've seen online, I think our devices could fit that, but I wouldn't be confident to make the call myself :) I'd like to get some more opinions.

caveman99:
The mentioned 'mitigation techniques' are not qualifierd here, but usually amount to CAD+AFA in other regulatory documents. We do CAD, but not AFA, so this usually does not apply.

IoTThinks:
If Meshtastic could have some LBT capabilities, then we could avoid a lot of regulation blockers.
Does Meshtastic have LBT (Listen before talk)?

HoangTheBoss:
Hi all, as I understand, since most roles are relaying messages it would count as a Gateway and adhere to 10% duty cycle like EU and would be fine(?)

caveman99:
A Gateway is a device that sends messages to another meduim, like from airwaves to the internet. Also LBT is CAD, but still the AFA is missing.

fifieldt:
@IoTThinks , do you have any contact with the regulator?

IoTThinks:
> @IoTThinks , do you have any contact with the regulator?

Sorry, I haven't revisited this issue recently.

IoTThinks:
I think Meshtastic may want to implement LBT and to limit the duty cycle on its own (like LoRaWAN), then we can deploy Meshtastic to Asia countries (which having similar restriction on LoRa/LoRaWAN/LPWAN).
I believe most of the users and IoT devices do not "talk" too much to violate the duty cycle.

Thanks a lot.

PR #7358, opened 2025-07-16
Small BaseUI Updates

Xaositek:
- Calculate the length of the `combinedName` string and properly use it
- Use sanitized longNameStr.c_str() in the Home frame display for longer names


Xaositek:
@jp-bennett Can you validate the change I made in NotificationRenderer.cpp is alright? `if (curSelected == static_cast<int8_t>(numDigits)) {`

jp-bennett (approved):

PR #7367, opened 2025-07-17
Add `OK_TO_MQTT` to userprefs

oscgonfer:
This is a simple PR to add `OK_TO_MQTT` to userPrefs.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -550,7 +550,6 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
     config.lora.tx_enabled =
         true; // FIXME: maybe false in the future, and setting region to enable it. (unset region forces it off)
     config.lora.override_duty_cycle = false;
-    config.lora.config_ok_to_mqtt = false;
 
 #if HAS_TFT // For the devices that support MUI, default to that
     config.display.displaymode = meshtastic_Config_DisplayConfig_DisplayMode_COLOR;
@@ -585,6 +584,11 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
     config.lora.ignore_mqtt = USERPREFS_CONFIG_LORA_IGNORE_MQTT;
 #else
     config.lora.ignore_mqtt = false;
+#endif
+#ifdef USERPREFS_CONFIG_LORA_OK_MQTT
+    config.lora.config_ok_to_mqtt = USERPREFS_CONFIG_LORA_OK_MQTT;
+#else
+    config.lora.config_ok_to_mqtt = false;
 #endif
     // Initialize admin_key_count to zero
     byte numAdminKeys = 0;
--- a/userPrefs.jsonc
+++ b/userPrefs.jsonc
@@ -18,6 +18,7 @@
   // "USERPREFS_CHANNEL_2_UPLINK_ENABLED": "false",
   // "USERPREFS_CONFIG_GPS_MODE": "meshtastic_Config_PositionConfig_GpsMode_ENABLED",
   // "USERPREFS_CONFIG_LORA_IGNORE_MQTT": "true",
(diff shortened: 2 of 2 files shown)


garthvh:
This should not be done in bulk, users need to explicitly opt into this value individually.

oscgonfer:
Hey @garthvh 

I discussed with @caveman99 and I was assuming this was possible but maybe I am wrong.

fifieldt:
@garthvh - the close here seems to be based on an assumption that people are using a phone app. My use case for this does not involve people using the phone app. Its a tracker role in a running race. We have a separate legal waiver for participants that takes care of any issues.

thebentern:
This seems risky from a compliance perspective with regard to not only our public mqtt broker, but it also enables trivial uplink without consent for any 3rd party brokers. I think this can be done in forks with the waiver for participants, but I am concerned that if this exists in the main firmware repo, in an audit it could demonstrate willing subversion of some of the privacy policy terms we are currently working on.

> @garthvh - the close here seems to be based on an assumption that people are using a phone app.

fifieldt:
We could add a check to the code so that the user pref is not applied if the public broker is in use?

thebentern:
> We could add a check to the code so that the user pref is not applied if the public broker is in use?

Based on my research, from a compliance standpoint it's actually worse if we allow implicit consent to forward the traffic to a 3rd party broker.

fifieldt:
It is not implicit consent in this case though :) It's explicit, with in-person verified sighting of government issued photo ID and signatures :) The waiver covers death and serious injury, so position privacy is a pretty minor part!

thebentern:
> It is not implicit consent in this case though :) It's explicit, with in-person verified sighting of government issued photo ID and signatures :) The waiver covers death and serious injury, so position privacy is a pretty minor part!

That is good to have such waivers in place, but that case doesn't apply to all usage necessarily. From my perspective, this should be a feature in a fork for that event instead of a userpref Meshtastic has approved, effectively transfering the liability to that party. If we approve it, we create a loophole which just adds risk for the project.

fifieldt:
From my perspective, as a volunteer for a nonprofit, I have got limited time to flash hundreds of devices (in my dreams, this maxes out at about 5000, but that's at least 3 years away). User-prefs is a feature to help me perform that process. it's an advanced feature for people in similar situation and is built with a different assumption about the level of competence of the user compared to other firmware features.

Telling such users to maintain a fork is making a statement about what Meshtastic should be used for. I personally believe there are many use cases outside the popular phone app that we should be trying to support. It seems like the compliance work (which seems rather ill-specified?) isn't taking those into account right now?

For the phone app use case, I think there's a big potential to use the market ecosystem and trademark program to demonstrate reputable/safe sources in a way that will be sufficient for compliance needs. I have some experience in this area - hit me up on discord :)

garthvh:
> @garthvh - the close here seems to be based on an assumption that people are using a phone app. My use case for this does not involve people using the phone app. Its a tracker role in a running race. We have a separate legal waiver for participants that takes care of any issues.

This is not based on the user using a phone app. Users need to opt in individually to this feature.

oscgonfer:
Hi,

Even if this seems to be a no-go because of the concerns listed above, I think there is still a use case where a single user or user group manages hundreds of devices and they agree on doing a certain approach. Having a consent mechanism for the deployment of these type of flags would be very convenient and save tons of time.

The environmental monitoring use case is a good example, where a group agrees on a certain configuration. In this particular case there is not just a user-to-user communication, but rather a network where the ownership of the devices is somewhat centralised (by a user, or user group). 

Now I wonder if a revised version of this could be accepted. For instance, a compile flag specific to a platformio board variant, or with a different consent mechanism, but still leaving the door open via userPrefs.

And, correct me if I am wrong, but I understood that the consent mechanism in `userPrefs` was implicit by the addition of a remote admin key.

PR #7370, opened 2025-07-17
Misc cppcheck fixes

jp-bennett:
Attempted to go after the low-hanging fruit suggested by cppcheck.


src/graphics/draw/ClockRenderer.cpp
```diff
@@ -366,9 +365,6 @@ void drawAnalogClockFrame(OLEDDisplay *display, OLEDDisplayUiState *state, int16
 
     // hour hand radius and y coordinate
     int16_t hourHandRadius = radius * 0.35;
-    if (isHighResolution) {
```
fifieldt, line 369: just wanted to check this one was a cppcheck since it seemed different to the others.

jp-bennett: Yes, cppcheck called out that this code was logically equivalent to just setting false.

fifieldt (approved):

Issue #7364, opened 2025-07-16
[Bug]: T114 sets incorrect date when GPS is enabled on firmware 2.7.2


AtomicJay:
### Category

Hardware Compatibility

### Hardware

Heltec Mesh Node T114

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.2.f6d3782 Alpha

### Description

When GPS is enabled, the date is set 2282 days in the past (e.g., instead of the date of writing—July 16, 2025—it shows April 17, 2019). The time of day, however, matches. When I disable GPS, the date syncs back to normal. I did not encounter such an issue on firmware 2.6.11.60ec05eb beta, which I switched back to due to this issue.

Steps to recreate:
1. Flash HelTec T114 (v2) (with GPS module) with firmware 2.7.2.f6d3782 Alpha.

2. Enable GPS.

### Relevant log output

```Shell

```

AtomicJay:
Unsure whether the date it sets to being so close to the most recent GPS week rollover (April 6-7, 2019) is a coincidence or not.

garthvh:
Probably need to wait a bit for a better lock, this is pretty common with that GPS chip

garthvh:
> Unsure whether the date it sets to being so close to the most recent GPS week rollover (April 6-7, 2019) is a coincidence or not.

Most likely the date the GPS code was written by the manufacturer

AtomicJay:
> Probably need to wait a bit for a better lock, this is pretty common with that GPS chip

I've tried leaving GPS enabled overnight with no luck.

AtomicJay:
I'll try leaving it on for a while again. Not expecting much, given my previous attempts, but I guess it's worth a try.

garthvh:
Make sure it has a clear view of the sky

jp-bennett:
Does that GPS ever get a lock? And does the date continue to be wrong if it does get a valid position?

AtomicJay:
> Does that GPS ever get a lock? And does the date continue to be wrong if it does get a valid position?

Yes. Even with over 10 sats and great positional accuracy, the date remains as it is.

AtomicJay:
![Image](https://github.com/user-attachments/assets/62144454-4f91-45f9-83ad-848700ac9f8a)

[...]

PR #7375, opened 2025-07-18
Add additional Epoch check for time set

fifieldt:
We have two perhapsSetRTC functions, which are called to set the time.

The one with 3 parameters had a helpful check to reject an invalid time, by comparing the time from the source against when the firmware was compiled. The one with 2 parameters, which is called from the GPS lookForTime did not.

As a result, certain GPS with bad time handling could set a time that was in the past.

This patch adds the same epoch check code to the other perhapsSetRTC method.

Fixes https://github.com/meshtastic/firmware/issues/7364


thebentern:
Great catch!

fifieldt:
Tested OK by a T114 user!

vidplace7:
Native tests are failing after this merge :grimacing: 
@fifieldt can you take a look?

PR #7386, opened 2025-07-19
Add Trace Route on BaseUI

whywilson:
## New Features
1. Trace Route in Favorites Action
2. Trace Route in Node List menu
3. Cooldown count down for 30s
4. Trace Route result showing

  - [x] Heltec T114
  - [x] GAT562
  - [x] Meshtiny


Xaositek (approved):

jp-bennett:
I did manage to cause a reboot by selecting a node to traceroute once. Second attempt did work. Could we show the result as a showSimpleBanner() instead of a module frame?

whywilson:
> I did manage to cause a reboot by selecting a node to traceroute once. Second attempt did work. Could we show the result as a showSimpleBanner() instead of a module frame?

Before using the new frame as Trace Route Result, I used the banner, but there were several problems and the banner display was not user-friendly.
1. Trace Route has multiple states: Tracing (static), Cooldown (dynamic), Result (static and disapper after XXs)
2. The width of the current banner is not fixed, it always gets out of the screen.

I still recommand the new frame for the Trace Route. Just like the canned message. How do you think so?

Issue #6478, opened 2025-04-01
[Board]: Seeed Xiao ESP32S3 + SX1262 for xiao (not the kit)


puflet:
### SOC

ESP32

### Lora IC

SX1262

### Product Link

https://www.seeedstudio.com/Wio-SX1262-for-XIAO-p-6379.html

### Description

Hi Meshers!
I have bought the board from the link, as I had some "Xiao ESP32-S3" laying around and decided I do not need the kit.
It turned out, that the board in question uses different pins for the LoRA chip, as it does not use Seeed's 'sense' connector, but old-school soldiered pins. I have managed to rewrite the firmware by only making changes to pins_arduino.h and variant.h files.
I think it is useless to add a new HardwareModel, but maybe a board variant, with proper explanation of the difference between the sold "Kit" version and the alternative "DIY" approach.

madeofstown:
I'm very interested in your changes. I am working on a carrier board for Xiao and a 1 watt radio.Alreasy have the nRF version working but it would save me some time if I could look at your variant files for the ESP32S3 version.

puflet:
Nothing spectacular, only rewritten some of the pins.
The boards I have used:
[SKU: 113010003 Wio-SX1262 for XIAO](https://www.seeedstudio.com/Wio-SX1262-for-XIAO-p-6379.html)
[SKU: 113991114 Seeed Studio XIAO ESP32S3](https://www.seeedstudio.com/XIAO-ESP32S3-p-5627.html)
**pins_arduino.h**
`static const uint8_t SS = 5; // was 41
`
**variant.h**
```
#define HAS_GPS 0 // (not actually required, on the test board I do not plan using GPS)
//#define USCREEN_SSD1306 // (as above)

#define LORA_CS 5 // was 41
#define LORA_RESET 3 // was 42
#define LORA_DIO1 2 // was 39

//#define LORA_DIO2 5 // was 38 (not sure about that, could not find DIO2 exposed according to Seeed's documentation)
#define SX126X_BUSY 4 // was 40
#define SX126X_RXEN 6 // was 38
```

Sabering1:
bump

Ope8888:
[...]

PR #7388, opened 2025-07-20
Add Seeed ESP32-S3 GPIO Variant (Alternative to LEGO Connector)

Sabering1:
This pull request introduces a new variant for the Seeed Studio ESP32-S3 board, functionally identical to the existing seeed_xiao_esp32s3 but using direct GPIO pin connections instead of the onboard Grove (LEGO-style) connector.
Motivation

The Grove connector on this board is mechanically fragile and prone to breakage, rendering the device unusable with Meshtastic despite the ESP32 and LoRa chips remaining functional. This variant ensures continued usability by routing the required signals via GPIO headers instead. I have 3 of these devices and on all 3 the connector snaped, our local community in Serbia also reported this issue with this seeed board.

This PR also closes: https://github.com/meshtastic/firmware/issues/6478

Mostly its just remaping pin layout to use GPIO. 

## 🤝 Attestations

Seeed ESP32-S3


Ope8888:
#6478 is different! We're using the Xiao SX1262, which is the same as the Xiao NRF52 kit...

The pins aren't the same as the Xiao ESP32 S3 kit.

variants/ seeed_s3_gpio/platformio.ini
```diff
@@ -0,0 +1,17 @@
...
+board_build.partitions = default_8MB.csv
+upload_protocol = esptool
+upload_speed = 921600
+lib_deps =
+  ${esp32s3_base.lib_deps}
+build_unflags =
+  ${esp32s3_base.build_unflags}
+  -DARDUINO_USB_MODE=1
+build_flags = 
+  ${esp32s3_base.build_flags} -DSEEED_XIAO_S3 -I variants/seeed_xiao_s3
```
caveman99: You are not hitting your own directory here with the include directive.

variants/ seeed_s3_gpio/variant.h
```diff
@@ -0,0 +1,90 @@
...
+  ▐░▌   ▐░▌  ▀▀▀▀█░█▀▀▀▀ ▐░█▀▀▀▀▀▀▀█░▌▐░█▀▀▀▀▀▀▀█░▌     ▐░█▀▀▀▀▀▀▀▀▀  ▀▀▀▀▀▀▀▀▀█░▌
+   ▐░▌ ▐░▌       ▐░▌     ▐░▌       ▐░▌▐░▌       ▐░▌     ▐░▌                    ▐░▌
+    ▐░▐░▌        ▐░▌     ▐░█▄▄▄▄▄▄▄█░▌▐░▌       ▐░▌     ▐░█▄▄▄▄▄▄▄▄▄  ▄▄▄▄▄▄▄▄▄█░▌
+     ▐░▌         ▐░▌     ▐░░░░░░░░░░░▌▐░▌       ▐░▌     ▐░░░░░░░░░░░▌▐░░░░░░░░░░░▌
+    ▐░▌░▌        ▐░▌     ▐░█▀▀▀▀▀▀▀█░▌▐░▌       ▐░▌      ▀▀▀▀▀▀▀▀▀█░▌ ▀▀▀▀▀▀▀▀▀█░▌
+   ▐░▌ ▐░▌       ▐░▌     ▐░▌       ▐░▌▐░▌       ▐░▌               ▐░▌          ▐░▌
+  ▐░▌   ▐░▌  ▄▄▄▄█░█▄▄▄▄ ▐░▌       ▐░▌▐░█▄▄▄▄▄▄▄█░▌      ▄▄▄▄▄▄▄▄▄█░▌ ▄▄▄▄▄▄▄▄▄█░▌
+ ▐░▌     ▐░▌▐░░░░░░░░░░░▌▐░▌       ▐░▌▐░░░░░░░░░░░▌     ▐░░░░░░░░░░░▌▐░░░░░░░░░░░▌
+  ▀       ▀  ▀▀▀▀▀▀▀▀▀▀▀  ▀         ▀  ▀▀▀▀▀▀▀▀▀▀▀       ▀▀▀▀▀▀▀▀▀▀▀  ▀▀▀▀▀▀▀▀▀▀▀
+*/
```
caveman99: Please remove the ASCII art

variants/ seeed_s3_gpio/platformio.ini
```diff
@@ -0,0 +1,17 @@
+[env:seeed-xiao-s3]
```
caveman99: You can not establish 2 environments with the same name. This is a duplicate, please change the env name.

variants/ seeed_s3_gpio/platformio.ini
```diff
@@ -0,0 +1,17 @@
...
+extends = esp32s3_base
+board = seeed-xiao-s3
+board_check = true
+board_build.partitions = default_8MB.csv
+upload_protocol = esptool
```
caveman99, line 6: Please add `board_level = extra` so this does not get built by default.

caveman99 (changes requested):
You have a whitespace in your directory name. Also we just reshuffled the directory structure for these variant files. please move the dir to the ESP32s3/diy subdir.

Sabering1:
> #6478 is different! We're using the Xiao SX1262, which is the same as the Xiao NRF52 kit...
> 
> The pins aren't the same as the Xiao ESP32 S3 kit.

i just noticed its two different boards that are similiar, ill fix it soon as its mostly the same with minor corrections needed

Sabering1:
Rebalanced pins, also seems that another user has already did that https://www.reddit.com/r/meshtastic/comments/1lihuas/bypass_seeed_esp32_s3_lora_kit_broken_connector/ and the firmware works

Pin layout is mosty the same except for two pins LORA_CS that goes to NSS and DO2 to RF_SW, connecting with jumper wires works fine, please review and let me know if there is anything else that needs to be done.

https://wiki.seeedstudio.com/wio_sx1262_with_xiao_esp32s3_kit/

ab0oo:
I don't think this is ready yet.
It hangs for anywhere from 30 seconds to almost 2 minutes during the I2C scan at boot:
```
INFO  | ??:??:?? 5 Scan for i2c devices
DEBUG | ??:??:?? 5 Scan for I2C devices on port 1
INFO  | ??:??:?? 117 No I2C devices found
DEBUG | ??:??:?? 117 acc_info = 0
```
The amount of time seems to vary from boot to boot.  In this case it was 112 seconds, which seems unreasonable.
It looks like SCL and SDA are mapped to what SHOULD be unused pins - SCL to D6, SDA to D5, but something is causing the scan to go sideways.
I guess one solution would be to simply disable the I2C scan, but there's certainly a chance someone is going to try to put an I2C _something_ on this stack, and then pull their hair out when it doesn't work.
I'm not sure how to proceed, other than just accept the long boot time.  Has anyone tested this with an I2C device installed?

S5NC:
If your connector is broken you may not have working interface elements of the green LED on GPIO48 and pulled-up button on GPIO21.

As LORA_RST needs to be connected via the GPIO pins, the 10k pull-up on LORA_RST will be fine.

variants/ seeed_s3_gpio/variant.h
S5NC: Was in the original! https://github.com/meshtastic/firmware/blob/3b6eefa8bbfcb2ae02e4e311b190cc2dfef9a05b/variants/esp32s3/seeed_xiao_s3/variant.h#L1-L25

S5NC:
Your changes are

```
LED_PIN 48 -> 44

GPS_RX_PIN 44 -> 39
GPS_TX_PIN 43 -> 40

LORA_CS 41 -> 4
LORA_RESET 42 -> 43
LORA_DIO1 39 -> 2
LORA_DIO2 38 -> 1

SX126X_BUSY 40 -> 3
SX126X_RXEN 38 -> 1
```
If you want to solder underneath you need
```
LORA_DIO1 39 -> 1 (and SX126X_DIO1)
LORA_BUSY 40 -> 2
LORA_RESET 42 -> 3 (and SX126X_RESET)
LORA_CS 41 -> 4 (and SX126X_CS)
SX126X_RXEN 38 -> 5

LORA_MOSI already 9
LORA_MISO already 8
LORA_SCK already 7
```
Double check with [this resource](https://wiki.seeedstudio.com/wio_sx1262_with_xiao_esp32s3_kit/#hardware-overview).
Then check for collisions.

Ignore `LORA_DIO2`, it isn't used.

PR #7393, opened 2025-07-20
Unify the shutdown proceedure

jp-bennett:
This should make the shutdown beeps work everywhere. Does not yet address longlong userbutton presses happening on release.


src/Power.cpp
```diff
@@ -681,7 +681,14 @@ bool Power::setup()
...
 {
-    LOG_INFO("Shutting Down");
+
+#if HAS_SCREEN
+    if (screen) {
+        screen->showSimpleBanner("Shutting Down...", 0); // stays on screen
+    }
+#endif
+    nodeDB->saveToDisk();
+    playShutdownMelody();
```
thebentern, line 690: IMO we should play the shutdown melody before the the savetoDisk, because the FS operations can lag. We want the audio and visual feedback to initiate at the same time

jp-bennett: Yeah, much better to let the user know.

PR #7402, opened 2025-07-21
Added LostAndFound menu item that triggers LostAndFound mode.

Paxy:
Added "LostAndFound" trigger to Home Action menu to enable switching to "LostAndFound" mode on nodes with screen. The action switch from Client mode to LostAndFound mode and from LostAndFound to Client mode.
Action is only preformed if device has GPS.

Tested on Heltec Wireless Tracker.

## 🤝 Attestations

  - [X] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/modules/SystemCommandsModule.cpp
+++ b/src/modules/SystemCommandsModule.cpp
@@ -76,15 +76,15 @@ int SystemCommandsModule::handleInputEvent(const InputEvent *event)
         return 0;
     case INPUT_BROKER_LOST_TOGGLE:
         LOG_WARN("Lost Toggle");
-        if (config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT) {
-            LOG_WARN("Changing mode to LOST_AND_FOUND");
-            config.device.role = meshtastic_Config_DeviceConfig_Role_LOST_AND_FOUND;
-            IF_SCREEN(screen->forceDisplay(); screen->showSimpleBanner("Switching to LOST!", 3000);)
-            nodeDB->saveToDisk(SEGMENT_CONFIG);
-        } else {
+        if (config.device.role == meshtastic_Config_DeviceConfig_Role_LOST_AND_FOUND) {
             LOG_WARN("Changing mode to Client");
             config.device.role = meshtastic_Config_DeviceConfig_Role_CLIENT;
-            IF_SCREEN(screen->forceDisplay(); screen->showSimpleBanner("Switching to Client!", 3000);)
+            IF_SCREEN(screen->forceDisplay(); screen->showSimpleBanner("Switching to Client!", 1000);)
+            nodeDB->saveToDisk(SEGMENT_CONFIG);
+        } else {
+            LOG_WARN("Changing mode to LOST_AND_FOUND");
+            config.device.role = meshtastic_Config_DeviceConfig_Role_LOST_AND_FOUND;
+            IF_SCREEN(screen->forceDisplay(); screen->showSimpleBanner("Switching to LOST!", 1000);)
             nodeDB->saveToDisk(SEGMENT_CONFIG);
             }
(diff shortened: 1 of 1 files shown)


Xaositek:
I think we need to account for "Client Mute" and "Client Hidden" roles also with an option to switch back to any of these three (Client, Client Mute, Client Hidden).

Paxy:
For that to be possible, we would need to record somewhere original mode, to be restored later. Any suggestion, where should we record original mode? It should persist on reboot.

Xaositek:
Given the "incoming" role could be one of various roles (Client, Client Mute, Client Hidden, Tracker, Sensor, TAK, TAK_Tracker), it may be easier to just convert to "Lost & Found" regardless.

Coming back to "Client" is okay, they can fine tune it later via client app.

Paxy:
In my first, local, version, I have implemented submenu to activate Lost mode, and return to one of modes chosen from menu. But then removed submenu. In case of emergency, this option should be as easiest as possible to activate. 
That's why simple toggle is probbably the best way (without submenu).
Also thinking about adding trigger for the same action with 8-10 button presses (using multipress function) for both devices with the screens (like Heltec Wireless Tracker) and without (for Card Tracker).

src/modules/SystemCommandsModule.cpp
```diff
@@ -74,6 +74,26 @@ int SystemCommandsModule::handleInputEvent(const InputEvent *event)
         }
 #endif
         return 0;
+    case INPUT_BROKER_LOST_TOGGLE:
+        LOG_WARN("Lost Toggle");
+#if !MESHTASTIC_EXCLUDE_GPS
+        if (gps) {
+            LOG_WARN("GPS present");
+            if (config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT) {
```
Xaositek: I think we need to remove this check - too many possibilities for incoming role to account for them here

Paxy: Did you mean to remove check is currently Client mode too? So it switch to Lost mode no matter what was original mode?

Xaositek: Given the user could have the device in one of several modes (Client, Client Mute, Client Hidden, Tracker, Sensor, TAK, TAK_Tracker) - it makes more check for the check to be "If in Lost Mode, Change Back to Client" and the else side it it to be "Switch to Lost Mode".

if (config.device.role == meshtastic_Config_DeviceConfig_Role_LOST_AND_FOUND) {
             config.device.role = meshtastic_Config_DeviceConfig_Role_CLIENT;
} else {
             config.device.role = meshtastic_Config_DeviceConfig_Role_LOST_AND_FOUND;
}

This way they can always change to Lost & Found and back regardless of role they were in.

src/modules/SystemCommandsModule.cpp
```diff
@@ -74,6 +74,21 @@ int SystemCommandsModule::handleInputEvent(const InputEvent *event)
         }
 #endif
         return 0;
+    case INPUT_BROKER_LOST_TOGGLE:
+        LOG_WARN("Lost Toggle");
+        if (config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT) {
```
Xaositek: I think the check for "config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT" should be removed - this would ensure whichever role the device is currently is is not a road block for getting into Lost & Found mode.

If you flip the condition to be "config.device.role == meshtastic_Config_DeviceConfig_Role_LOST_AND_FOUND" then you can make it be the move to client or move to lost.

Xaositek (changes requested):

src/modules/SystemCommandsModule.cpp
Paxy: Ok, flipped conditions.

PR #7407, opened 2025-07-21
nRF52840 promicro deepsleep fix with some additions, fixes #7368

### LINKED ISSUE #7368: [Bug]: nRF52 does not wake up when pressing a button
### Category

Other

### Hardware

DIY

- [ ] InkHUD ePaper

### Firmware Version

2.6.5 and later

### Description

I use custom nrf52840 devices. I can turn off any of them perfectly well with a long press on the button on any firmware version. However, waking up with a short press on the button is only possible on firmware versions 2.6.4 and below, and starting with 2.6.5, the devices do not respond to this action: a reboot is required.
The button is connected in accordance with the diagram available in the diy catalog, i.e. with a pull-up to +3.3V. In the off state, the +3.3V bus remains operational, i.e. the absence of a pressing event due to the absence of a falling edge can be excluded.
There are no reported errors in the console when shutdown procedure is performing.

### Relevant log output

```Shell

```

MagnusKos:
Fixed the previously noticed bug #7368, which concerned assemblies on ProMicro. Judging by the code, the bug may also concern commercial devices on NRF52840, but this has not been tested.

The bug was in the command to stop serial port 1 (Serial1.end()): judging by the code base, the Arduino object Serial1 is not used anywhere on nRF52 architecture (but not 100% sure), but this serial port's buses are used in the GPS control modules. Obviously, there is a conflict.

The first commit solves this problem very simply: the fact of using Serial1 is checked. If it is not used, then the end() method is not called.

Next, a reset of the GPREGRET register is additionally introduced. On the Nordic forum, you can find messages about the values ​​written to the register being incorrect if a reset was not performed before. The problem is not common, but it is better to prevent it.

An explicit connection of the SENSE signal to the user button has also been added. The signal is used in deep sleep mode and allows the device to wake up. The device wakes up without specifying a signal, but it is better to specify this explicitly to avoid difficult to track bugs.

---
UPD: Tested these commits on T-1000e. Nothing breaks, sleep and wake are ok, GPS shuts down and goes to an active state as usual.
UPD1: Fixes #7368

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
   - DIY nRF52840 ProMicro


skobkin:
Just in case, my Heltec T114 seems to not be affected by this problem (mentioned this in #7368).

skobkin:
To my previous comment.

1. T114 **without screen** wasn't waking up by user button press.
~~2. I've tested an update with this change backported by @MagnusKos.~~

~~It worked and node was waking up, but it was also waking up on USB connect and disconnect which probably shouldn't happen if I'm not mistaken. Will check later to verify if T114 with screen on vanilla firmware wakes up on USB connection state change.~~

UPD: T114 with screen on vanilla 2.7.3 wakes up on USB too.

alexl83:
I'm running this changes (cherry-picked)  on 2.6.11 latest beta on my heltecs T114 and rak4631 and everything runs fine

csrutil:
One or two weeks ago, I thought it was a compatibility issue. I managed to fix it with a simple workaround, so I’m really glad you fixed it properly. Great job!

```ruby
diff --git a/src/platform/nrf52/main-nrf52.cpp b/src/platform/nrf52/main-nrf52.cpp
index 1bf9a39fd..56faa01ed 100644
--- a/src/platform/nrf52/main-nrf52.cpp
+++ b/src/platform/nrf52/main-nrf52.cpp
@@ -285,7 +285,9 @@ void cpuDeepSleep(uint32_t msecToWake)
     // This may cause crashes as debug messages continue to flow.
     Serial.end();
 #ifdef PIN_SERIAL1_RX
+#ifndef MESH_WATCH
     Serial1.end();
+#endif
 #endif
     setBluetoothEnable(false);
```

Issue #7149, opened 2025-06-27
[Feature Request]: UDP support for RAK13800 ethernet module


kabili207:
### Platform

NRF52

### Description

I would really love to be able to use the Mesh over UDP feature with this ethernet module. I did see possible support was mentioned in a comment on [the original pull request](https://github.com/meshtastic/firmware/pull/5779#issuecomment-2577207713), but it doesn't look like it has been mentioned anywhere else.

I wasn't able to find a recent, maintained version of AsyncUDP for this module, so I get the feeling this will require a very different approach than the ESP32 and Linux native ports

rjDipcord:
This. I would assume in a a perfect implementation an ethernet module connected to a network should behave exactly as a Wi-Fi connected one does.

vidplace7:
The UDP implementation on ESP32 relies on the [`AsyncUDP`](https://github.com/espressif/arduino-esp32/tree/master/libraries/AsyncUDP) library. Unfortunately there's no such library for the nRF52 series.
TCP/IP networking is really an afterthought on the [Arduino-nRF52](https://github.com/adafruit/Adafruit_nRF52_Arduino/tree/master/libraries) platform. ☹

thebentern:
Closed with #7385

kabili207:
I would not consider this completely closed as #7385 only enables support when using the ETH GW variant, which has a significant number of modules and features disabled that make it unusable to me for my primary base node.

I moved the HAS_UDP_MULTICAST flag to the base rak4631 variant and it seems to be working quite well on my base node with seemingly no ill effects when running on another mobile RAK node that doesn't have the ethernet module. With that in mind, can this flag be enabled on the base rak4631 variant?

PR #7408, opened 2025-07-21
Enable UDP for base RAK4631. Relates to #7149

kabili207:
This enables UDP functionality for the basic RAK4631 variant. This was previously only enabled on the ETH Gateway variant via #7385, however this variant is too restrictive for several nodes in my area's mesh. Enabling this on the basic variant proved successful and did not cause any issues when flashed to nodes that did not have an ethernet module connected.

## 🤝 Attestations

  - [x] RAK WisBlock 4631


garthvh:
What is too restrictive about the existing rak firmware that supports the ethernet module?

kabili207:
The eth_gw variant disables screen, power monitoring, external notifications, pax counter, remote hardware, store & forward, canned messages, and waypoints. I don't personally use *all* of these features, but I do use some and so do others on my mesh. The eth_gw variant seems like it's intended to be an extremely minimal build, rather than something that could be used as a primary home base node like many people on my local mesh use.

Edit: direct link to the eth_gw variant: https://github.com/meshtastic/firmware/blob/master/variants/rak4631_eth_gw/platformio.ini

caveman99 (approved):
The ETH variant needs to make a lot of room for ArduinoJson/MQTT. There might be a point in the future where we remove the ETH libs from the regular 4631 firmware for size reasons and in turn spruce up the ETH variant a bit more.

For now this is good to merge.

kabili207:
Ah, that makes sense. Can't say I need the JSON stuff for my own setup, especially considering it's not part of the basic 4631 firmware, but I can see why some might like to have it.

Perhaps having a compromise variant between the basic and ETH variant could be useful, but I suppose that's a concern for the future.

Anyway, thank you!

caveman99:
The JSON will eventually be removed from the firmware completely and replaced by a plugin in the mqtt broker decoding the raw protobufs and providing compatible json off-board. - Not today, not tomorrow, and when we remove it from the firmware, the alternatives will be in place.

kabili207:
Rebased my changes for the recently merged layout change in #7396

kabili207:
Switched the merge branch from `master` to `develop` and cherrypicked my changes. Not sure when how I ended up on master originally, but I figured I should move that to the current default branch

fifieldt:
@caveman99 in light of #8130 is enabling UDP by default on the RAK4631 base still the way forward?

kabili207:
I'm holding back on resolving any of the conflicts until #8462 is sorted out, as it makes the RAK effectively useless and I haven't been able to determine if enabling UDP contributes to the problem or not.

PR #7412, opened 2025-07-21
Migrate remaining variants to new dir structure

vidplace7:
Migrates remaining variants to the new dir structure
- esp32c6
- native (portduino)
- I forgot `mesh-tab` earlier! Move that too


vidplace7:
All affected platformio environments have been compile tested. Shipit! Finally done with these :D

fifieldt (approved):

PR #7415, opened 2025-07-22
Moves the shutdown thread into the Power class, make shutdown and reboot private

jp-bennett:


vidplace7:
This change broke all `inkhud` targets :grimacing: @jp-bennett

PR #7423, opened 2025-07-22
Deprecate configs

jp-bennett:
Looking through settings for things that can be easily discarded. Possibly 3.0 work

### DIFF
--- a/src/gps/GPSUpdateScheduling.cpp
+++ b/src/gps/GPSUpdateScheduling.cpp
@@ -34,7 +34,8 @@ uint32_t GPSUpdateScheduling::msUntilNextSearch()
     uint32_t now = millis();
 
     // Target interval (seconds), between GPS updates
-    uint32_t updateInterval = Default::getConfiguredOrDefaultMs(config.position.gps_update_interval, default_gps_update_interval);
+    uint32_t updateInterval =
+        Default::getConfiguredOrDefaultMs(config.position.broadcast_smart_minimum_interval_secs, default_gps_update_interval);
 
     // Check how long until we should start searching, to hopefully hit our target interval
     uint32_t dueAtMs = searchEndedMs + updateInterval;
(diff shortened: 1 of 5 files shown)


jp-bennett:
I'm contemplating splitting this out, with the triple click and always_smart landing right away.

src/gps/GPS.cpp
```diff
@@ -1039,12 +1039,12 @@ void GPS::down()
             softsleepSupported = true;
 
         if (softsleepSupported) {
-            // How long does gps_update_interval need to be, for GPS_HARDSLEEP to become more efficient than
+            // How long does broadcast_smart_minimum_interval_secs need to be, for GPS_HARDSLEEP to become more efficient than
```
fifieldt, line 1042: The gps update interval and the smart update stuff is independent and used by people differently, this is not deprecated and shouldn't be removed.

fifieldt (changes requested):
The gps update interval and the smart update stuff is independent and used by people differently, this is not deprecated and shouldn't be removed.

fifieldt:
Lemme know if you want a run-through of these settings, how folks use them, and how they interact with our multitude of GPS chips :)

src/gps/GPS.cpp
fifieldt: It's easy to get confused between the broadcast interval and the update interval :)

jp-bennett:
@fifieldt would love to chat about it at some point. I was under the impression that we *always* used smart interval. (That was the intent the last time I touched this code, but that has been a while).

If that is the case, we should be able to derive the update interval from the smart interval. Basically, check for a new GPS position from hardware at the same rate as smart interval wants to check for a new position.

And then it seems like the maximum broadcast interval time could be a multiple of the smart interval.

fifieldt:
Yes, we should chat.

Is there a problem you're trying to solve, other than simplification?

If you look even in the last couple hours in the gnss channel, you'll see our resident GPS expert cynfab talking about altering update interval :)

capricornusx:
Am I correct in understanding that we also need to update the documentation regarding these configuration parameters, for example, on this page: https://meshtastic.org/docs/configuration/radio/position/#smart-broadcast?

PR #7424, opened 2025-07-22
Use platformio-core to build the matrix

vidplace7:
Use PlatformIO's Python API to build the matrix.

The goal with this PR was to replace the script without modifying behavior (inputs/outputs) at all.


vidplace7:
Tested in my fork :) Seems to be behaving happily

PR #7426, opened 2025-07-23
Deprecate smart gps config -- Always Smart!

jp-bennett:


src/mesh/NodeDB.cpp
```diff
@@ -948,7 +940,6 @@ void NodeDB::installRoleDefaults(meshtastic_Config_DeviceConfig_Role role)
     } else if (role == meshtastic_Config_DeviceConfig_Role_CLIENT_HIDDEN) {
         config.device.rebroadcast_mode = meshtastic_Config_DeviceConfig_RebroadcastMode_LOCAL_ONLY;
         config.device.node_info_broadcast_secs = MAX_INTERVAL;
-        config.position.position_broadcast_smart_enabled = false;
```
fifieldt, line 951: might benefit from a

broadcast_smart_minimum_interval_secs = MAX_INTERVAL to be safe? otherwise it might broadcast on movement at the defaults if there isn't a CLIENT_HIDDEN exception in all the right places elsewhere in the code

src/mesh/NodeDB.cpp
```diff
@@ -644,11 +644,6 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
         config.position.gps_mode = meshtastic_Config_PositionConfig_GpsMode_DISABLED;
 #else
     config.position.gps_mode = meshtastic_Config_PositionConfig_GpsMode_ENABLED;
-#endif
-#ifdef USERPREFS_CONFIG_SMART_POSITION_ENABLED
```
vidplace7, line 648: Can you remove `USERPREFS_CONFIG_SMART_POSITION_ENABLED` in userPrefs.jsonc as well? :pray: 

This is the only place it was used.

vidplace7 (changes requested):

src/mesh/NodeDB.cpp
jp-bennett: Sure. I was unsure if that was a generated file.

PR #7427, opened 2025-07-23
Remember destination fix

HarukiToreda:
When a new freetext or preset message is sent and a destination is selected, the next message would forget the previously selected destination and would need to be selected again. With this fix it will remember the last destination selected until changed again.


src/modules/CannedMessageModule.cpp
```diff
@@ -63,8 +66,16 @@ CannedMessageModule::CannedMessageModule()
...
+    if (!lastDestSet) {
+        dest = NODENUM_BROADCAST;
+        channel = 0;
+        lastDest = dest;
+        lastChannel = channel;
+        lastDestSet = true;
+    } else {
+        dest = lastDest;
+        channel = lastChannel;
+    }
```
jp-bennett, line 73: Doesn't this block defeat the purpose of this function, to initiate a message with a specific destination?
I suggest pulling this code into a new function, LaunchRepeatDestination() that checks for lastDestSet, and then calls LaunchWithDestination()

HarukiToreda: This should fix the reply function too. Tested replying, it is remembered on both Preset and freetext screen. If a destination is never selected or replied to, it defaults to Broadcast Default channel

jp-bennett (approved):

PR #7433, opened 2025-07-23
add WIP for Unit C6L

caveman99:


variants/esp32c6/m5stack_unitc6l/platformio.ini
```diff
@@ -0,0 +1,24 @@
...
+upload_protocol = esptool
+;upload_port = /dev/ttyACM2
+lib_deps =
+  ${esp32c6_base.lib_deps}
+  adafruit/Adafruit NeoPixel@^1.12.3
+  ponoor/PI4IOE5V6416@^0.2.0
+build_flags = 
+  ${esp32c6_base.build_flags}
+  -D PRIVATE_HW
+  -I variants/m5stack_unitc6l
```
vidplace7:
```suggestion
  -I variants/esp32c6/m5stack_unitc6l
```
:) You dropped this

caveman99: Indeed :-)

variants/esp32c6/m5stack_unitc6l/pins_arduino.h
```diff
@@ -0,0 +1,28 @@
+#ifndef Pins_Arduino_h
+#define Pins_Arduino_h
+
+#include <stdint.h>
+
+#define USB_VID 0x303a
+#define USB_PID 0x1001
```
jp-bennett: This is different from the VID and PID from m5stack:
#define USB_VID 0x2886
#define USB_PID 0x0048

Any clue why?

variants/esp32c6/m5stack_unitc6l/platformio.ini
```diff
@@ -0,0 +1,24 @@
+[env:m5stack-unitc6l]
+extends = esp32c6_base
+board = esp32-c6-devkitm-1
```
jp-bennett: Probably needs to be esp32-c6-devkitc-1

src/AmbientLightingThread.h
```diff
@@ -183,9 +183,9 @@ class AmbientLightingThread : public concurrency::OSThread
 #endif
 #endif
             pixels.show();
-            LOG_DEBUG("Init NeoPixel Ambient light w/ brightness(current)=%d, red=%d, green=%d, blue=%d",
-                      moduleConfig.ambient_lighting.current, moduleConfig.ambient_lighting.red,
-                      moduleConfig.ambient_lighting.green, moduleConfig.ambient_lighting.blue);
+            // LOG_DEBUG("Init NeoPixel Ambient light w/ brightness(current)=%d, red=%d, green=%d, blue=%d",
```
caveman99, line 186: This seems unrelated

src/main.cpp
```diff
@@ -544,6 +543,12 @@ void setup()
 #endif
 #endif
 
+#if defined(M5STACK_UNITC6L)
+    pinMode(LORA_CS, OUTPUT);
```
caveman99, line 547: If we never change that, use the existing PIN_POWER_EN macro rather than defining a new one. c6l_init should be switched to the lateInitVariant mechanism used in src/platform/extra_variants rather than being called here

variants/esp32c6/m5stack_unitc6l/variant.cpp
```diff
@@ -0,0 +1,74 @@
...
+    Wire.beginTransmission(addr);
+    Wire.write(reg);
+    Wire.write(value);
+    Wire.endTransmission();
+}
+/*******************************************************************/
+void c6l_init()
+{
+    // P7 LoRa Reset
+    // P6 RF Switch
```
caveman99, line 46: As pointed out bove, this is prime initLateVariant() stuff

src/AmbientLightingThread.h
jp-bennett: It was throwing a log message every 30 seconds, this is the first platform I've used that had a neopixel from factory. It was pretty disruptive to getting a decent log.

variants/esp32c6/m5stack_unitc6l/variant.cpp
jp-bennett: I tried. Doing this in initLateVariant() causes the LoRa radio to stop responding. Likely because the reset line gets twiddled while setting up the GPIO extender.

src/main.cpp
jp-bennett: It does get passed in to RadioLib, so should get controlled there. This is essentially the same code as runs for the T-deck/pro, t-lora-pager, and rp2040.

caveman99:
@jp-bennett left a few comments, gonna continue tomorrow with it.

PR #7441, opened 2025-07-23
Build RP2350 (Pi Pico 2)

vidplace7:
Meshtastic has supported the `rp2350` platform for a while now, let's build it in CI!

This change will result in 2 additional builds (upon PR and upon release)
- `pico2`
- `pico2w`

Additionally, adds `check` jobs for `picow` and `pico2w`, currently there are no rp2040/rp2350 boards with `board_check` defined :astonished:

### DIFF
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -30,7 +30,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


Bfreak:
Thank you for this work! Can someone please share the UF2, or push it to the web flasher? I'm trying to build in VScode myself but it's a nightmare for someone new to firmware building as RP2350 has no support in platformIO.

PR #7442, opened 2025-07-23
Fix timezone definition for UTC in TZPicker function

barbabarros:
This pull request includes a minor update to the `TZPicker` method in the `MenuHandler` class to ensure proper timezone definitions.

Timezone definition update:

* [`src/graphics/draw/MenuHandler.cpp`](diffhunk://#diff-42b779fbaf784d6b09c1770397a59db686044c519108888f421651113995f1a7L190-R190): Updated the timezone definition for UTC from `"UTC"` to `"UTC0"` to align with standard timezone formatting.
* This way, the schedule is updated as soon as it changes. There is no need to reset the device.

## 🤝 Attestations

  - [X] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


Xaositek:
Verified that UTC0 is the correct POSIX definition string.

PR #7450, opened 2025-07-24
WashTastic variant

valzzu:
Made a variant for washtasic.

uses the build flag for 30dBm E22 also uses private hadware instead of promicro


vidplace7 (approved):
Your PlatformIO file looks good 👍 (and in the correct directory ❤️)

caveman99 (approved):
So i assume correctly it uses the promicro definitions but puts a different label on the build, cause of the different radio?

valzzu:
> So i assume correctly it uses the promicro definitions but puts a different label on the build, cause of the different radio?

Pretty much yea.

Issue #3207, opened 2024-02-11
[Feature Request]: Spread telemetry/GPS updates across channels


tropho23:
### Platform

NRF52

### Description

(This feature request is *not* only for nrf52 but the platform choice does not allow me to choose 'all platforms')

Request firmware feature to spread telemetry/GPS updates across multiple channels.

Currently users are limited to using primary channel for sending telemetry and GPS data.  Ideally, selective send of this data to any channel(s) we choose but this has been limited to primary channel to avoid mesh saturation/channel utilization.

If allow sending GPS updates to any/all channels but limit update intervals per channel to respect the *total* interval update that would be rather useful.  For example:

_(With GPS update interval set to 10 minutes)_
00:00=send GPS update on channel 0/primary
00:10=send GPS update on channel 1
00:20=send GPS update on channel 2
00:30=send GPS update on channel 3
00:40=send GPS update on channel 0/primary (40 minutes since last update)
...and so on.  The idea is not exceed the GPS update interval in total, which is 10 minutes.

The example above allows users to send GPS updates via more than only primary channel, but updates respect the 10 minute interval set.  This allows channel utilization to remain the same as if GPS updates are sent only on the primary channel.

I recognize users can increase GPS update interval and flood the mesh but they can already do that so this proposal adds zero risk if implemented in the manner suggested above.

This feature cannot be depending on mobile apps to function as standalone devices should be able to use this feature.

jp-bennett:
Somewhat related is the desire to set precision levels for different channels. Possibly through a bitmask that gets AND'd with the location at https://github.com/meshtastic/firmware/blob/master/src/modules/PositionModule.cpp#L110 before sending the packet.

garthvh:
Seems really complex, uses a bunch of airtime end is of marginal benefit, still not sure why secondary channels need position. Also has to be done in 3.0.

andrekir:
we really don't want to send the same packet multiple times in different channels. this would have a huge impact in total airtime considering rebroadcasts/ACKs.

GUVWAF:
I agree with Garth and Andre. The suggestion to let it send on a different channel at each opportunity is good, but then again if you want frequent updates on one channel, you’ll need to set the updates really high. Duplicating the settings for each channel seems very cumbersome.

[...]

PR #7467, opened 2025-07-26
feat: precision-aware position filtering for multi-channel broadcasting

truongsinh:
Implements intelligent GPS position filtering that handles broadcasting the same position across multiple channels with different precision levels. Prevents high-precision position data from being overwritten by lower-precision versions of the same location when broadcasting across multiple channels.

Key Features:
- Multi-channel position broadcasting with channel-specific precision levels
- Preserves highest precision version when same position received with lower precision
- Smart movement detection works across all precision levels
- Optimized for devices broadcasting on multiple channels simultaneously

Use Cases:
- Broadcasting full precision on private/encrypted channels
- Broadcasting degraded precision on public channels for privacy
- SAR scenarios with both operational and public communication channels
- Privacy-conscious position sharing with precision control per channel

Position Update Logic:
- Same location + lower precision = NO UPDATE (preserve high precision)
- Same location + higher precision = UPDATE (upgrade precision)
- Different location = ALWAYS UPDATE (regardless of precision)
- Same precision = UPDATE (refresh timestamp)

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/mesh/generated/meshtastic/deviceonly.pb.h
+++ b/src/mesh/generated/meshtastic/deviceonly.pb.h
@@ -33,6 +33,8 @@ typedef struct _meshtastic_PositionLite {
     uint32_t time;
     /* TODO: REPLACE */
     meshtastic_Position_LocSource location_source;
+    /* Indicates the bits of precision set by the sending node */
+    uint32_t precision_bits;
 } meshtastic_PositionLite;
 
 typedef PB_BYTES_ARRAY_T(32) meshtastic_UserLite_public_key_t;
--- a/src/modules/PositionModule.cpp
+++ b/src/modules/PositionModule.cpp
@@ -41,6 +41,30 @@ PositionModule::PositionModule()
     }
 }
 
+bool PositionModule::shouldUpdatePosition(const meshtastic_PositionLite &existingPos, const meshtastic_Position &incomingPos)
+{
+    // Assume precision is 32 if not set (older firmware compatibility)
+    uint32_t existingPrecision = existingPos.precision_bits ? existingPos.precision_bits : 32;
+    uint32_t incomingPrecision = incomingPos.precision_bits ? incomingPos.precision_bits : 32;
+    
+    // Use the lower precision for position comparison
+    uint32_t minPrecision = (incomingPrecision < existingPrecision) ? incomingPrecision : existingPrecision;
+    
+    int32_t degradedExistingLat = existingPos.latitude_i & (0xFFFFFFFF << (32 - minPrecision));
+    int32_t degradedExistingLon = existingPos.longitude_i & (0xFFFFFFFF << (32 - minPrecision));
+    
+    int32_t degradedIncomingLat = incomingPos.latitude_i & (0xFFFFFFFF << (32 - minPrecision));
+    int32_t degradedIncomingLon = incomingPos.longitude_i & (0xFFFFFFFF << (32 - minPrecision));
+    
+    if (degradedExistingLat != degradedIncomingLat || degradedExistingLon != degradedIncomingLon) {
+        // Position changed - always update regardless of precision
+        return true;
+    } else {
+        // Same position - only update if precision is equal or higher
+        return incomingPrecision >= existingPrecision;
+    }
+}
+
 bool PositionModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshtastic_Position *pptr)
 {
     auto p = *pptr;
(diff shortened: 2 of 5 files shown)


src/mesh/generated/meshtastic/deviceonly.pb.h
```diff
@@ -33,6 +33,8 @@ typedef struct _meshtastic_PositionLite {
     uint32_t time;
     /* TODO: REPLACE */
     meshtastic_Position_LocSource location_source;
+    /* Indicates the bits of precision set by the sending node */
```
fifieldt, line 36: For reference, this code is automatically generated. The source is here: https://github.com/meshtastic/protobufs

truongsinh: Thanks, I'm aware. If this PR is approved, I'll make PR for that repo as well before merging this.

src/modules/PositionModule.cpp
```diff
@@ -338,7 +382,7 @@ void PositionModule::sendOurPosition()
         if (channels.getByIndex(channelNum).settings.has_module_settings &&
             channels.getByIndex(channelNum).settings.module_settings.position_precision != 0) {
             sendOurPosition(NODENUM_BROADCAST, requestReplies, channelNum);
-            return;
```
GUVWAF, line 341: We deliberately do not periodically transmit position out over multiple channels to avoid a storm of packets, only the first channel that has position enabled is used. The other channels are request-only.

GUVWAF (commented):
If one changes its position from high precision to low precision on the same channel, then we should not ignore this. This way you can never undo a precise location if you don't move.
If we do update when position and precision is the same, then this does not actually reduce traffic, right?

truongsinh:
> If one changes its position from high precision to low precision on the same channel, then we should not ignore this. This way you can never undo a precise location if you don't move.

You're technically correct. Without movement, a high-precision position can't be reverted. If we want to support reverting or degrading precision, we'll need an additional mechanism to decide when a lower-precision update should still be accepted.

> If we do update when position and precision is the same, then this does not actually reduce traffic, right?

Right, this logic operates on the recipient side, so it doesn't reduce transmission traffic. However, traffic reduction isn’t the goal of this PR. The purpose is to allow nodes to broadcast different levels of precision on different channels, and ensure those updates don’t get unintentionally overwritten by subsequent broadcasts on the public channel.

src/modules/PositionModule.cpp
truongsinh: > The other channels are request-only.

Can you help me find the doc how to do request on other channels? So far, using Android, I only see "exchange position" when viewing details of another node. I assume the precision for that "exchange position" is the setting of public channel.

> We deliberately do not periodically transmit position out over multiple channels to avoid a storm of packets, only the first channel that has position enabled is used.

Can this be made configurable, or is it a hardcoded design choice? If it’s hardcoded, I assume nodes running a patched firmware (that removes that return statement) could still broadcast position on multiple channels, correct?

GUVWAF:
> Without movement, a high-precision position can't be reverted. 

I think it would be a bit misleading if nodes store a precise fixed location forever when someone once set it like this.

> However, traffic reduction isn’t the goal of this PR. 

I was put off guard by the new comment:
``` 
// Smart position update logic with precision awareness:
// Only update if position has actually changed to avoid unnecessary mesh traffic
```

src/modules/PositionModule.cpp
GUVWAF: > I assume the precision for that "exchange position" is the setting of public channel.

Indeed you can use "exchange position" for this, but it will not necessarily use the primary channel. Position packets do not use PKI encryption, and then the channel where you most recently heard the node's NodeInfo on is used:
https://github.com/meshtastic/firmware/blob/28aeb0f09e9bb633d5ba8b6511050ac778e030c0/src/mesh/Router.cpp#L193

> Can this be made configurable, or is it a hardcoded design choice? If it’s hardcoded, I assume nodes running a patched firmware (that removes that return statement) could still broadcast position on multiple channels, correct?

Yes, it's a hardcoded design choice. I believe most users will not realize this means you're potentially duplicating traffic 7 times. It has been discussed several times, e.g.: https://github.com/meshtastic/firmware/issues/3207
With patched firmware, yes, it's possible.

axunes:
the docs talk about the precision on "a particular channel" which makes it sound like this is already implemented

PR #7471, opened 2025-07-26
Add on screen keyboard for encoder and trackball device

whywilson:
  - [x] Meshtiny
  - [x] GAT562 with up down encoder addson
  - [x] Seeed Wio Tracker L1
IMG_5542


Xaositek:
I have reviewed and I believe we need present this option like we do via other Virtual and Physical Keyboard options - there is an option similar to "New Preset Message" which is "New FreeText Message" - you can see an example in MenuHandler.cpp; lines 318-320:

    if (kb_found) {
        optionsArray[options] = "New Freetext Msg";
        optionsEnumArray[options++] = Freetext;
    }

This would ensure we are only long-term supporting a single path for Preset and FreeText

whywilson:
> I have reviewed and I believe we need present this option like we do via other Virtual and Physical Keyboard options - there is an option similar to "New Preset Message" which is "New FreeText Message" - you can see an example in MenuHandler.cpp; lines 318-320:
> 
> 
> 
>     if (kb_found) {
> 
>         optionsArray[options] = "New Freetext Msg";
> 
>         optionsEnumArray[options++] = Freetext;
> 
>     }
> 
> 
> 
> This would ensure we are only long-term supporting a single path for Preset and FreeText

Now the new Free Text menu is in the canned messages list. 
It is triggered by the trackball or encoder. 
And it ought to be different from the IIC keyboard. How do you think?

Xaositek:
I would follow the same pathways of the SEEED SenseCAP Indicator which uses Virtual Keyboard. This is also a virtual keyboard in the same way so it should be loaded in the same manner. We are aiming for a consistent user experience.

whywilson:
> I would follow the same pathways of the SEEED SenseCAP Indicator which uses Virtual Keyboard. This is also a virtual keyboard in the same way so it should be loaded in the same manner. We are aiming for a consistent user experience.

How about use a variable for **osk_found** for that?
osk_found will be true when use the encoder or the trackball?

Xaositek:
In main.cpp we have at line 578-580:
`#if defined(USE_VIRTUAL_KEYBOARD)
    kb_found = true;
#endif`

Should we add another definition line here and use kb_found? This would re-use everything.

mrekin:
Hi!
Very nice virtual kb.
Is it hard to add additional/alternative kb layout ?
Now you have hardcoded chars list (inside initializeKeyboard). That nice, but limits charset to hardcoded and will not allow to use all chars from font.
May be possible to list all chars exist in font (like 0123..ABCD..abcd.. ) - in this case additional charsets (cyrillic chars as example) can be listed and used.

whywilson:
> Hi! Very nice virtual kb. Is it hard to add additional/alternative kb layout ? Now you have hardcoded chars list (inside initializeKeyboard). That nice, but limits charset to hardcoded and will not allow to use all chars from font. May be possible to list all chars exist in font (like 0123..ABCD..abcd.. ) - in this case additional charsets (cyrillic chars as example) can be listed and used.

Thanks for your suggestion. I can try extracting the characters, but currently the supported characters are limited to a...z, A...Z, 0...9, and simple punctuation characters, with uppercase input enabled by long-pressing a key.

If a function similar to the Shift key were supported, perhaps more characters could be supported.

whywilson:
This PR will be continue on https://github.com/meshtastic/firmware/pull/7625

Issue #7480, opened 2025-07-27
[Feature Request]: Block packages with ID 258 on portnums only and/or turn pornums only into a whitelist


Ne00n:
### Platform

Cross-Platform

### Description

As discussed on Discord, we have a problem in our local mesh network, that someone keeps spamming packages with the ID 258, which isn't blocked by portnums only.

Currently when enabling portnums only, this is a blacklist, so packages such as with the ID 258 are still forwarded, I suggested to turn this blacklist into a whitelist instead, so if any other issues occur with custom ID's used by someone this would be already blocked.

Suggested by Stary on Discord was to update portnums, to block anything higher than 255.
I don't mind which solution will be implemented though, if we can drop 258 in our network it would be lit.

So we can just tell people to update their firmware and enable port nums only to get rid of that problem.
I posted the statistics we collected with graphs on discord, I can repost them here if needed.

Please lemme know if you need anything.

PR #7483, opened 2025-07-28
CORE_PORTNUMS_ONLY should not include ports > 255.

fifieldt:
As reported by @Ne00n and Stary, the core portnums only router forward limit still forwarded ports in the private application space. This patch removes ports above 255 from the core portnums, as this was unintended.

fixes https://github.com/meshtastic/firmware/issues/7480

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


thebentern:
@fifieldt based on the original intent of the mode, which is to aggressive block traffic outside of the normal ports going through a router, I propose a whitelist approach rather than the current blacklist, which just creates somewhat of a whack-a-mole situation as new problematic traffic patterns emerge.
https://github.com/meshtastic/firmware/pull/7487

PR #7487, opened 2025-07-28
Core portnums rebroadcast mode whitelist instead of blacklist

### LINKED ISSUE #7480: [Feature Request]: Block packages with ID 258 on portnums only and/or turn pornums only into a whitelist

thebentern:
Closes #7480

### DIFF
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -651,11 +651,12 @@ void Router::handleReceived(meshtastic_MeshPacket *p, RxSource src)
         shouldIgnoreNonstandardPorts = true;
 #endif
         if (shouldIgnoreNonstandardPorts && p->which_payload_variant == meshtastic_MeshPacket_decoded_tag &&
-            IS_ONE_OF(p->decoded.portnum, meshtastic_PortNum_ATAK_FORWARDER, meshtastic_PortNum_ATAK_PLUGIN,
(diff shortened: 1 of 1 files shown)


fifieldt (approved):

shalberd:
great idea, thank you. Anything that keeps traffic down, especially here in Europe, is a good thing.
With this change, mountain routers can set CORE_PORTNUMS_ONLY and have all relevant functionality.
This change now makes CORE_PORTNUMS_ONLY a real alternative to LOCAL_ONLY.

@thebentern how about a combination of CORE_PORTNUMS_ONLY and LOCAL_ONLY?

Sven

PR #7492, opened 2025-07-28
DEBUG_MUTE correctness

Stary2001:
Make all of the `#ifdef DEBUG_PORT` guards also not get compiled if `DEBUG_MUTE` is set. The logs would already get removed anyway, but this helps prevent crashes when float support is turned off.
The one in Router.cpp has side effects now (sending a client notification), so I un-guarded it.

I also added a forced assert for STM32 that when debug printing is enabled we MUST support floats, I keep ignoring my own comment.

## 🤝 Attestations
  - [x] Seeed Xiao NRF52 kit
  - [x] Custom wio-E5 board


thebentern (approved):

fifieldt (approved):
excellent work

PR #7493, opened 2025-07-29
Only send Neighbours if we have some to send.

fifieldt:
The original intent of NeighborInfo was that when a NeighbourInfo was sent all of the nodes that saw it would reply with NeighbourInfo. So, NeighbourInfo was sent even if there were no hop-zero nodes in the NodeDB.

Since 2023, when this was implemented, our understanding of running city-wide meshes has improved substantially. We have taken steps to reduce the impact of NeighborInfo over LoRa.

This change aligns with those ideas: we will now only send NeighborInfo if we have some neighbors to contribute.

The impact of this change is that a node must first see another directly connected node in another packet type before NeighborInfo is sent. Nodes will, however, receive the regular periodic broadcast of NeighborInfo, and will be able to send NeighborInfo if it has at least 1 neighbor.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/modules/NeighborInfoModule.cpp
+++ b/src/modules/NeighborInfoModule.cpp
@@ -112,8 +112,8 @@ void NeighborInfoModule::sendNeighborInfo(NodeNum dest, bool wantReplies)
     // only send neighbours if we have some to send
     if (neighborInfo.neighbors_count > 0) {
         printNeighborInfo("SENDING", &neighborInfo);
+        service->sendToMesh(p, RX_SRC_LOCAL, true);
     }
-    service->sendToMesh(p, RX_SRC_LOCAL, true);
 }
 
 /*
--- a/src/modules/NeighborInfoModule.cpp
+++ b/src/modules/NeighborInfoModule.cpp
@@ -106,12 +106,13 @@ void NeighborInfoModule::sendNeighborInfo(NodeNum dest, bool wantReplies)
     meshtastic_NeighborInfo neighborInfo = meshtastic_NeighborInfo_init_zero;
     collectNeighborInfo(&neighborInfo);
     meshtastic_MeshPacket *p = allocDataProtobuf(neighborInfo);
-    // send regardless of whether or not we have neighbors in our DB,
-    // because we want to get neighbors for the next cycle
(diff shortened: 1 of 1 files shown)


fifieldt:
@caveman99 , as the original author, does this seem sane to you?

Picked this up when we found a Heltec V3 that was stuck in a loop sending 0-neighbour neighbourinfo packets to our MQTT server. Thousands of them.

src/modules/NeighborInfoModule.cpp
```diff
@@ -106,12 +106,13 @@ void NeighborInfoModule::sendNeighborInfo(NodeNum dest, bool wantReplies)
...
-    // because we want to get neighbors for the next cycle
     p->to = dest;
     p->decoded.want_response = wantReplies;
     p->priority = meshtastic_MeshPacket_Priority_BACKGROUND;
-    printNeighborInfo("SENDING", &neighborInfo);
+    // only send neighbours if we have some to send
+    if (neighborInfo.neighbors_count > 0) {
+        printNeighborInfo("SENDING", &neighborInfo);
+    }
     service->sendToMesh(p, RX_SRC_LOCAL, true);
```
GUVWAF: `sendToMesh()` should also be in the if-clause then, but better to check it before allocating the data protobuf to avoid memory leak also.

GUVWAF (commented):
Idea sounds good to me, left a comment though.

> This means that a node with no neighbors is no longer able to trigger other nodes to broadcast NeighborInfo.

This shouldn’t be the case already. Nodes only periodically send NeighborInfo, not triggered by other nodes; `wantReplies` is always false. They will only rebroadcast other nodes’ NeighborInfo.
In one of the alpha releases some time ago there was a bug that when the broadcast interval was not set, it would broadcast it as fast as possible.

src/modules/NeighborInfoModule.cpp
fifieldt: D'oh!

PR #7504, opened 2025-07-29
Enable S&F on any channel

jacobpretorius:
Hello,

This is more of a question than an immediate PR for merging, but I thought it would be easier to show what I'm talking about.

I have a Station G2 that I want to permanently advertise on the mesh maps to help with adoption in my area. Therefore, I need channel 0 to be the default public channel on it as there's no way around that hardcoded limitation (I've tried).

Yet when I set up S&F, whenever I do testing (latest alpha firmware) with a private channel 1 shared across my devices, whenever they come back into range messages are not automatically sent to them that they missed. If I message the station node with "SF", I get back a reply: "S&F not permitted on the public channel."

That seems to render S&F unusable? Why?

By removing the checks as in my code below, I can have exactly this setup. Channel 0 reports the station on the maps, channel 1 is my private S&F shared with other devices.

Please help me understand if I'm missing something or if the current limitation with S&F is there for some other reason.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Station G2

### DIFF
--- a/readmejp.md
+++ b/readmejp.md
@@ -1,0 +1,29 @@
+## Setup env
+python3 -m venv venv     
+source venv/bin/activate   
+
+## Test on device
+
+### Station G2
+1. Enable Serial Debug:
+# Edit variants/esp32s3/station-g2/platformio.ini line 22:
+# Change: -DARDUINO_USB_MODE=0  
+# To:     -DARDUINO_USB_MODE=1
+2. Rebuild and Flash:
+pio run -e station-g2 -t upload
+3. Connect Serial:
+pio device monitor --baud 115200
+  
+
+### Heltec v3
+
+pio run -e heltec-v3 -t upload
+
+pio device monitor --baud 115200
+
+## If protobufs need updating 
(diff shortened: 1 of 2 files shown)


Nevermore-42:
Hi Jacob!

Perhaps it will be better to be able to chose on which channel you want to use the S&F module individually? 

As it is right now you either allow storing&forwarding everything (including public channels if you have them as secondary) or you are not allowed at all if your primary is public (your case).

jacobpretorius:
> Hi Jacob!
> 
> Perhaps it will be better to be able to chose on which channel you want to use the S&F module individually?
> 
> As it is right now you either allow storing&forwarding everything (including public channels if you have them as secondary) or you are not allowed at all if your primary is public (your case).

Hi Nevermore-42,

That would certainly be nicer I think. I'm not usually a C++ dev so it's probably above my skill level to implement that. Hopefully someone can implement it 👍

PR #7511, opened 2025-07-30
128row display

caveman99:


tropho23:
Tested with nrf52840; works perfectly except default orientation is upside down relative to the 128x128 OLED's PCB.  Is it possible to flip the default orientation?

caveman99:
I do think this is display specific and can be solved through the protobufs or device firmware pre-config. if we flip this in here it's contrary to the rest of the displays. This case is not the first one where a display was put in 'upside down' :-)

PR #7514, opened 2025-07-30
bugfix Syntax error: "(" unexpected in device-update.sh

mikecarper:
Use /bin/bash instead of /bin/sh and the error goes away and shellcheck is happy as well


vidplace7:
This wasn't needed in the debian packaging afaik...

PR #7516, opened 2025-07-31
Move BLE toggle menu option and add confirmation for canned messages in L1

thebentern:
Based on some feedback about how easy it is to accidentally fire off messages and disable bluetooth with Seeed L1 Tracker Pro inside of a bag, we're adding a bit of friction to the process:
1) Move bluetooth to system menu where it makes more sense
2) Add a confirmation dialog for sending canned messages for this target. Other targets can opt in to this with the `CANNED_MESSAGE_ADD_CONFIRMATION` macro

Tested on:

- [x] L1 Tracker Pro
- [x] RAK4631

### DIFF
--- a/src/graphics/draw/MenuHandler.cpp
+++ b/src/graphics/draw/MenuHandler.cpp
@@ -14,7 +14,9 @@
 #include "modules/AdminModule.h"
 #include "modules/CannedMessageModule.h"
 #include "modules/KeyVerificationModule.h"
+
 #include "modules/TraceRouteModule.h"
+#include <functional>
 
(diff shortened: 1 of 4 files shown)


thebentern:
> I noticed a small user experience issue in that when you confirm "Yes" it drops you back into the select preset menu and allows you to control scrolling through the messages before the "Message Sent" pops up. I feel like we should clear the menus and drop the user back to the original frame they were on to use up that time.
> 
>  IMG_5157.mov

Good call. I'll fix that today

Xaositek (approved):
All issues addressed - works great!

PR #7520, opened 2025-07-31
Fix Heltec T114 ADC multiplier

OnyxClawe:
The current ADC multiplier for the Heltec T114 reports 95% at a full battery charge. I've adjusted it to 4.99 to correct for this. A full charge now results in 100% battery being reported instead of 95%.
## 🤝 Attestations

    [✅] I have tested that my proposed changes behave as described.
    [✅] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
        [✅] Heltec T114 V2

### DIFF
--- a/variants/nrf52840/heltec_mesh_node_t114/variant.h
+++ b/variants/nrf52840/heltec_mesh_node_t114/variant.h
@@ -208,7 +208,7 @@ No longer populated on PCB
 #undef AREF_VOLTAGE
 #define AREF_VOLTAGE 3.0
 #define VBAT_AR_INTERNAL AR_INTERNAL_3_0
-#define ADC_MULTIPLIER (4.90F)
+#define ADC_MULTIPLIER (4.99F)
 
 #define HAS_RTC 0
 #ifdef __cplusplus


OnyxClawe:
Apologies for the lack of a label! Im very new to github and i do not see an option to add a label. This is a bugfix

thebentern:
@OnyxClawe what battery mAH rating does this correspond to?

OnyxClawe:
@thebentern 
I used a 3000mAh cell in my testing

alexl83:
I'm running this change (cherry-picked)  on 2.6.11 latest beta on my heltecs T114, everything runs fine

fifieldt (approved):

HarukiToreda:
This change is inherently incorrect. The firmware currently assumes that 4.19V = 100%, which misleads users into believing that a fully charged battery should always reach 4.19V. In reality, this is only a suggested nominal value from battery manufacturers.

Batteries with built-in over-voltage protection typically cut off charging earlier:

Many boards stop charging at 4.17V.

The T1114 is even stricter, cutting off at 4.15V.

By forcing the ADC to show “100%” at 4.19V, the reported voltage no longer reflects the true battery voltage. This is misleading and can cause downstream issues, particularly in setups such as solar power systems that rely on accurate voltage readings to trigger cutoffs or protection logic.

The previous ADC calibration, while not flawless, was based on the actual physical resistor values used on the board. Because no resistors are perfect, there was always some variance, but those readings were grounded in the real-world hardware. This made them more accurate to the true voltage than the artificial adjustment to 4.19V.

Accurate voltage reporting is far more important than presenting a cosmetically “perfect” percentage. For users who depend on precise readings for system behavior, convenience adjustments like this introduce unnecessary risk.


Here's the voltage reading with my supplying 4.19V with this PR. Shows we are at 4.24V on the display





Here's the OLDER firmware reading with my supplying 4.19V. Shows we are at 4.17 on the display






A more accurate ADC value should be 4.924 to be as close as possible to true Voltage value of the battery.

fifieldt:
@HarukiToreda , that makes sense to me. Let's revert?

HarukiToreda:
> @HarukiToreda , that makes sense to me. Let's revert?

yeah lets revert it. I'll push something to improve this after some more testing

PR #7530, opened 2025-08-01
Add some manual github actions to make building subsets easier

NomDeTom:
Added 3 new workflows:
- Build Target List - makes a list of all of the targets within each arch
- Build One Arch - builds *everything* in a particular arch
- Build One Target - accepts a build target from the action above and builds it.

I've also added another conditional to one of the workflows to stop it running automatically on a fork.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/DebugConfiguration.h
+++ b/src/DebugConfiguration.h
@@ -23,6 +23,7 @@
 #define MESHTASTIC_LOG_LEVEL_ERROR "ERROR"
 #define MESHTASTIC_LOG_LEVEL_CRIT "CRIT "
 #define MESHTASTIC_LOG_LEVEL_TRACE "TRACE"
+#define MESHTASTIC_LOG_LEVEL_HEAP "HEAP"
 
 #include "SerialConsole.h"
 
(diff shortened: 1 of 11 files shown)


NomDeTom:
@vidplace7

NomDeTom:
> Looks good, especially the changes to run only here.
> 
> I'm a bit concerned at the proliferation of additional pipelines, but manual only shouldn't pose an issue

This is intended to only really be run on a forked repo, hence all the additional guards.

fifieldt:
Thanks for this. The guards are good - having random actions run on forks is annoying :)

Q) These are nicely laid out. Is there a way they could be integrated into existing actions to avoid that little bit of duplication? (actually, is there any overlap?)

NomDeTom:
@fifieldt I did consider rolling it all into the existing builder, but the actions to generate the matrix interferes with the generation of a single build target - the final `build one target` splits the action in two at this point. The `build one arch` and `build target list` are semi-superfluous, I suppose, but they're good demos of how their relevant actions work.

fifieldt (approved):

PR #7531, opened 2025-08-01
style: apply formatting like removing executable bits when not needed…

iranzo:
Removing extra spaces and adding pre-commit config in addition to current trunk formatting.

Additionally removing execution permission for many files that doesn't needed (code, headers, etc).

Additionally use `is None` in python comparisons instead of `== None`

About this: `- Please do not reformat lines that you didn't have to change the code on`, this change precisely is fixing  in the repo trailing spaces, end of files, permissions, etc and also enabling pre-commit in trunk for future usage


fifieldt:
This is too many changes to reasonably review. Suggest breaking up the patch into smaller chunks.

PR #7565, opened 2025-08-06
Add base generic ELRS targets

Senk02:
Hey guys, girls, neither or both!

Currently, from what I have seen, the support for ELRS targets is very minimal, and I would like to change that as ELRS generally uses the same exact hardware as meshtastic.

For my first commit, I am adding all the generic targets that are available, and will add more as time goes on. I would ask that this PR not get merged just yet, and more so is used as a basis for conversation, about adding all the ELRS targets.

There are many ways to go about this; We could have a single header file with all the targets that users can select, or something along those lines... Or something cleaner can be had, if changes to the flasher are allowed. From what I understand, that is a bit more of a slippery slope. Since this has potential to replace the bandit (which is already in the flasher) with literally every single ELRS target, I think it could be a nice way to support it.

Let me know what you all think, or how we can potentially go about this. My discord is `senk0`, feel free to reach out. 

Edit: Grammar fixes

Cheers!
- Tina


Senk02:
I fully understand this is a massive commit, but here's what's going on:

I made a python script to read the https://github.com/ExpressLRS/targets/tree/master ELRS targets repo, convert them into header files, then toss them into their own chip directories with pio files and whatnot.

I also added a settable pin in main, to be able to set a second radio's CS pin high, and left it open for others to use, as to not make it target specific. 

Everything is still far from perfect, but it serves as a starting point.

Issue #3883, opened 2024-05-14
[Feature Request]: use internal ESP32, NRF52, RP temperature sensor of RTC


roha-github:
### Platform

NRF52, ESP32, RP2040

### Description

Idea: use the internal temperature sensor of the Real Time Clock (RTC) as a temperature sensor. 

RTC chips have an internal temperature sensor for ongoing calibration of the Real Time Clock. The internal sensor could be used to ( approximately) determine the temperature. 

https://www.instructables.com/Temperature-Measuring-Without-Any-Sensors-in-NRF52/
https://docs.espressif.com/projects/esp-idf/en/stable/esp32s2/api-reference/peripherals/temp_sensor.html
https://learnembeddedsystems.co.uk/using-the-rp2040-on-board-temperature-sensor

roha-github:
> You'll get a temperature that's (much) higher than the actual ambient temperature, and it's unknown how to compensate for it. I'm not sure how useful this really is.

It depends ... I have already used the RTC temperature sensor in projects. With a correction value you can adjust the measurement temperature to the little heating of the chip.

But with pure sensors, the MCU is almost always in sleep mode anyway and there is de facto no heating.  The values are then +/-1°C and a good indicator of heat.

Btw you can also use LEDs as light sensors for certain circuits.

roha-github:
@caveman99 This is about the possibility of enabling a temperature sensor WITHOUT additional hardware. I would leave it up to the user to decide exactly what this is and for what purposes it can be used.

Have you had any practical experience with the temperature sensors from RTC? What were your deviations from reality?

thebentern:
A dirt cheap supported BMP180 or SHT temp i2c sensor can be added for little cost and effort. Trying to use problematic CPU temperatures isn't something that seems like a fruitful effort in the face of relatively easy alternatives.

GUVWAF:
> What I particularly like about Meshtastic is that it's not just made for nerds. Without IT knowledge, users can buy one of the inexpensive devices and simply flash it via the web UI.

That's indeed nice. However, if a temperature sensor is included by default, those people expect it to be accurate. You can’t really say that you shouldn’t rely on it when it’s -20 degrees outside, but if it’s +20 degrees it _might_ be accurate, depending on how busy your mesh and thus processor and radio is, etc.

roha-github:
As caveman99 says, the real temperature and the RTC temperature are not directly identical. 

[...]

PR #7577, opened 2025-08-08
Add nrf internal temperature sensor

viric:
Useful as Enviroment Telemetry, in absence of a better temperture sensor.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631

### DIFF
--- a/src/modules/Telemetry/EnvironmentTelemetry.cpp
+++ b/src/modules/Telemetry/EnvironmentTelemetry.cpp
@@ -198,6 +198,11 @@ T1000xSensor t1000xSensor;
 IndicatorSensor indicatorSensor;
 #endif
 
+#ifdef ARCH_NRF52
+#include "Sensor/NRFTempSensor.h"
+NRFTempSensor nrfTempSensor;
+#endif
+
 #define FAILED_STATE_SENSOR_READ_MULTIPLIER 10
 #define DISPLAY_RECEIVEID_MEASUREMENTS_ON_SCREEN true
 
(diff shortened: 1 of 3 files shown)


caveman99:
per https://github.com/meshtastic/firmware/issues/3883 we measure ambient temperature, not the system temperature. This is out of the scope of Meshtastic firmware.

viric:
Would it help if we made it an option?
The NRF chips are used at such low power that its temperature is very close to the temperature of the enclosure.
Most mounted nodes I have seen with a temperature sensor have the sensor in the same box and exposed to the sun, which means that the measure is as bad as the cpu temperature. So I think the cpu temperature is a good representative not of ambient temperature but of the node enclosure temperature.

Many other supported sensors will be installed in the node enclosure and report not ambient temperature.

Moreover in my case I'm more interested of the enclosure temperature because of the battery charging conditions. Maybe we can have a new telemetry measurement for this, not "Environmental", to avoid confusion? In any case the measurement is very useful.

viric:
Notice that measuring well ambient temperature is very difficult and requires very good hardware (against sun influence) and installation conditions, so I don't expect any good ambient temperature measurement from meshtastic nodes reporting environment temperature. I don't see why the nrf die temperature is particularly worse than any of the other sensors supported.

And I wrote the code so if any other sensor is added it will be taken in preference over the nrf die temperature.

PR #7590, opened 2025-08-09
Add Israel region frequencies config

unconditional:
**Israel region config**

917.3 - 919.9 MHz, <25 mW erp, 1% duty cycle, 200 kHz spacing, 125 kHz bandwidth

Legal:

The Wireless Telegraph Regulations 2021, pages 19-20

https://www.gov.il/he/Departments/DynamicCollectors/gazette-official?skip=0&limit=10&BookletNum=9301&FolderType=3
https://olaw.org.il/takanot/takanot-9301.pdf

Google-translated:
en01
en02


fifieldt:
Per discussion on the protobuf, unfortunately we can't support regions with only 1% duty time :(

PR #7602, opened 2025-08-11
Test commit for XEdDSA support

jp-bennett:
Very early preview of working XEdDSA packet signing.


lkosson:
With each subsequent `NodeInfo` update requiring a proper signature, wouldn't this feature require #5007 to be deployed as well? Without it, resetting a node to factory defaults and/or regenerating a key pair will not only prevent DMs as in abovementioned discussion, but also block changes to `shortName` / `longName`, forcing user to use old (semi-broken) identity.

jp-bennett:
> With each subsequent `NodeInfo` update requiring a proper signature, wouldn't this feature require #5007 to be deployed as well? Without it, resetting a node to factory defaults and/or regenerating a key pair will not only prevent DMs as in abovementioned discussion, but also block changes to `shortName` / `longName`, forcing user to use old (semi-broken) identity.

I think we have a good opportunity to do exactly this: Make the public Key the new primary identifier.

jp-bennett:
TODO: make sure the packetID is included in the signature, and verified on receive
or use the first half of the signature as the source of the currently randomized portion of the packetID?

weebl2000:
A total of 66 bytes per message seems like way too much imho.

PR #7620, opened 2025-08-13
Fix 'buildroot' compiles (OpenWRT)

vidplace7:
`ARCH_PORTDUINO` should be defined in the shared portduino base. Buildroot (OpenWRT) compiles are currently broken :'(


vidplace7:
Confirmed this fixes compile issues with `buildroot` target by implementing it as a workaround via https://github.com/meshtastic/openwrt/pull/35

Ship it!

PR #7621, opened 2025-08-13
Docker: Update Debian images to trixie

vidplace7:
Update Docker images to Debian 13 (trixie).

DRAFT requisite upon further testing.


migillett:
Built and tested on a raspberry pi 4 using Raspbian and docker and connected to a MeshAdv Hat over SPI. Ran the built container mediumfast and longfast, accessed via mobile and `contact` TUI. Traceroutes sending as expected. Text messages sending as well. Logs look normal as well, nothing too crazy. I say we're good to move forward.

PR #7624, opened 2025-08-14
Prompt user to select destination upon launch of canned message module

### LINKED ISSUE #7615: [Feature Request]: Canned message menu selection should start on [Select Destination] instead of the first pre-set message
### Platform

Cross-Platform

### Description

Title

ford-jones:
Resolves #7615 
Ensures that when a user launches the canned message module the first option highlighted is: `[Select Destination]`.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/modules/CannedMessageModule.cpp
+++ b/src/modules/CannedMessageModule.cpp
@@ -78,16 +78,15 @@ void CannedMessageModule::LaunchWithDestination(NodeNum newDest, uint8_t newChan
     lastDestSet = true;
 
     // Rest of function unchanged...
-    // Always select the first real canned message on activation
-    int firstRealMsgIdx = 0;
+    // Upon activation, highlight "[Select Destination]"
+    int selectDestination = 0;
     for (int i = 0; i < messagesCount; ++i) {
-        if (strcmp(messages[i], "[Select Destination]") != 0 && strcmp(messages[i], "[Exit]") != 0 &&
-            strcmp(messages[i], "[---- Free Text ----]") != 0) {
-            firstRealMsgIdx = i;
+        if (strcmp(messages[i], "[Select Destination]") == 0) {
+            selectDestination = i;
             break;
         }
     }
-    currentMessageIndex = firstRealMsgIdx;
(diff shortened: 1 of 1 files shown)


Xaositek:
Willing to be wrong here - but we default the user to their primary channel if a new message and if replying it will go back to the source of the message.

Selecting the first canned message is by design.

HarukiToreda:
the reason why it defaulted to the first message, was to prioritize sending a message as quick as possible for example an SOS with a rotary encoder, whoever now that this feature is open to all devices I can see that moving down through all messages to reach the destination can be not ideal for devices with 1 button. I think it is a good thing to change it to default destination to be highlighted. Whoever this will need to be tested to ensure devices with onboard keyboards don't see any regressions since the first spot is the Freetext option that doesn't exist on normal devices.

ford-jones:
@HarukiToreda thanks for the insight!
Sounds good to me :smiley_cat: I guess I will wait until somebody with the onboard keyboard hardware is able to test this out for us then!

Xaositek (commented):
I will approve, but thinking out loud should we just make it select the first value always?

HarukiToreda:
I got a SenseCap indicator to test it.

HarukiToreda:
> I will approve, but thinking out loud should we just make it select the first value always?

not sure how selecting the first value will affect onscreen keyboards yet, testing now.

ford-jones:
@Xaositek 

> I will approve, but thinking out loud should we just make it select the first value always?

I'm not opposed to that! 
Just wondering though, for all intents and purposes might it not make sense to give the user some choice over which message occupies the first position? I.e. Allow the user to favourite a preset message?

HarukiToreda:
No regressions on touch devices. Good to push

https://github.com/user-attachments/assets/354bfac3-936f-4ca3-bdc5-f0b4074c2245

HarukiToreda (approved):

PR #7636, opened 2025-08-15
Support for T-Echo Lite, credits to @Szetya for doing all the heavy lifting!

caveman99:


thebentern (approved):

Szetya:
If you follow Twitter events, an additional back cover with a push button is planned. Hopefully with TCA8418. This will also allow sending short text messages. 
https://x.com/lilygo9/status/1942754288671023571

PR #7640, opened 2025-08-15
Remove JSON serialization from most NRF52 targets

thebentern:


variants/nrf52840/monteops_hw1/platformio.ini
```diff
@@ -9,7 +9,7 @@ build_flags = ${nrf52840_base.build_flags}
 build_src_filter = ${nrf52_base.build_src_filter} +<../variants/nrf52840/monteops_hw1> +<mesh/eth/> +<mesh/api/> +<mqtt/>
 lib_deps = 
   ${nrf52840_base.lib_deps}
-  ${networking_base.lib_deps}
+  ${nrf52_networking_base.lib_deps}
   https://github.com/RAKWireless/RAK13800-W5100S/archive/1.0.2.zip
```
caveman99: I think that should stay in, this board has an actual ETH linkup.

PR #7661, opened 2025-08-17
Loopback support in portduino with crc check to ignore own packets

jp-bennett:


src/mesh/udp/UdpMulticastHandler.h
```diff
@@ -85,12 +92,18 @@ class UdpMulticastHandler final
...
+#endif
         udp.writeTo(buffer, encodedLength, udpIpAddress, UDP_MULTICAST_DEFAUL_PORT);
         return true;
     }
 
   private:
     IPAddress udpIpAddress;
     AsyncUDP udp;
+#if ARCH_PORTDUINO
+    uint32_t last_crc;
```
Jorropo, line 106: This needs to be
```suggestion
    std::atomic<uint32_t> last_crc;
```
on portduino.
(the rest of this file is also lacking sync but that a bug)

Jorropo:
I have sent a PR here https://github.com/meshtastic/framework-portduino/pull/59
I've tested it and it works, it can't have the 2+ packets race bug.
It also remove the rare risk of CRC32 collisions.

PR #7664, opened 2025-08-18
Skip GPS baud rate probing if disabled, reduce the stucking duration

whywilson:
If the GPS module is not available at startup, the baud rate cycle will take more than 30 seconds, causing the interface to freeze. If GPS is not enabled, this check can be skipped.

  - [x] Heltec T114
  - [x] Wio Tracker L1
  - [x] Meshtiny
  - [x] GAT562

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -492,6 +492,10 @@ static const int rareSerialSpeeds[3] = {4800, 57600, GPS_BAUDRATE};
  */
 bool GPS::setup()
 {
+    // skip baud rate probing if disabled
+    if (isDisabled())
+        return true;
+
     if (!didSerialInit) {
         int msglen = 0;
         if (tx_gpio && gnssModel == GNSS_MODEL_UNKNOWN) {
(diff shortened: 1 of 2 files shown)


jp-bennett:
I believe the reason why we don't do this already is that on some hardware, the only way we have to disable the GPS is to send it a message to go to sleep. So we need to discover it at boot to tell it to sleep.

So this probably needs to check for a GPS enable line. If we have a GPIO to turn it off, then we don't need to probe.

whywilson:
> I believe the reason why we don't do this already is that on some hardware, the only way we have to disable the GPS is to send it a message to go to sleep. So we need to discover it at boot to tell it to sleep.
> 
> So this probably needs to check for a GPS enable line. If we have a GPIO to turn it off, then we don't need to probe.

What's the device handling the GPS in that way? I can try to check when I get one.

fifieldt:
Yes, another aspect to consider here is that by default GPSes tend to be configured to emit a fairly maximal set of messages. For the ones we can't power off (eg soldered to board), we can only turn down the flood if we know what GPS chip we have.

whywilson:
> Yes, another aspect to consider here is that by default GPSes tend to be configured to emit a fairly maximal set of messages. For the ones we can't power off (eg soldered to board), we can only turn down the flood if we know what GPS chip we have.

Thank you all for the reminding.

fifieldt:
Thank you for your work to optimise this!

PR #7682, opened 2025-08-19
Update Dockerfile to better string wrangle

jp-bennett:
The $(cat...) expression isn't getting evaluated in the dockerfile. Probably needs to be run inside a bash -c call.


vidplace7:
~This is probably also needed in `alpine.Dockerfile`~ ❤️ nevermind, we don't bundle web with Alpine.

jp-bennett:
Looking like this isn't a bash string replacement issue, but a missing .tar on the web repo.

PR #7691, opened 2025-08-20
Initial stab at rak6421 autoconf

jp-bennett:
When starting Meshtasticd with autoconf, the app will now look for an EEPROM, and try to detect whether there is a RAK6421 config string there, and auto-config if found. Properly formatted string:
        //      RAK6421-13300-S1:aabbcc123456:5ba85807d92138b7519cfb60460573af
        // <model string>:mac address :<16 random unique bytes in hexidecimal>


Kongduino:
Looks good. Wondering whether having a sanity check (codeword, CRC32) to make sure we are dealing with actual registration data would be safer.

PR #7625, opened 2025-08-14
Add on-screen keyboard implementation on Trackball device.

whywilson:
- [x] I have tested that my proposed changes behave as described.
- [x] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [x] Seeed Wio Tracker L1 Pro 
IMG_5762

PR #7694, opened 2025-08-20
Account for 128x128 SH1107 displays

thebentern:


whywilson:
Hi Ben, 
Could you make a test on the on-screen keyboard layout on the 128x128 display with this PR? It can be triggered in Free-Text menu. 
https://github.com/meshtastic/firmware/pull/7625

caveman99:
that looks like it is resolved in the meantine by the exact area conflicting here. Please re-open if assessment is incorrect.

PR #7717, opened 2025-08-22
Opt-in position broadcasting

### LINKED ISSUE #7692: [Feature Request]: Make Position be disabled by default
### Platform

Cross-Platform

### Description

Position broadcast should be an opt-in and not an opt-out

ford-jones:
Resolves #7692 

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


src/mesh/NodeDB.cpp
```diff
@@ -928,8 +928,8 @@ void NodeDB::installRoleDefaults(meshtastic_Config_DeviceConfig_Role role)
         owner.has_is_unmessagable = true;
         owner.is_unmessagable = true;
         config.device.node_info_broadcast_secs = ONE_DAY;
-        config.position.position_broadcast_smart_enabled = true;
-        config.position.position_broadcast_secs = 3 * 60; // Every 3 minutes
+        config.position.position_broadcast_smart_enabled = true; // Broadcast is implicit in the selection of this role
+        config.position.position_broadcast_secs = 3 * 60;        // Every 3 minutes
```
GUVWAF, line 932: `position_broadcast_secs` is set by default to `default_broadcast_secs` (every 15 min.). To completely disable sending positions, this would need to be set to `MAX_INTERVAL` also.

GUVWAF (approved):
Implementation looks good to me now.

Let's see if there's common consensus about this. If so, this requires updating the documentation also.

ford-jones:
Cheers @GUVWAF !
Should the doc updates be made by me as a part of this PR or is there some other process for producing and updating the docs?

GUVWAF:
Yes, documentation updates use the following repository: https://github.com/meshtastic/meshtastic

thebentern:
IMO this needs some discussion. Node locations are one of the big discoverability factors on local meshes, so I'm not sure we want to add friction there at a small savings of bandwidth. Some users already complain about the setup process of meshtastic being too involved. I'm not sure adding steps for position would be worth the bandwidth saved.

ford-jones:
Yes fair shout @thebentern :slightly_smiling_face: 
My gatherings from the ticket were that position broadcasting may be considered invasive by some users. Especially considering that a lot of users on the mesh are here because they favour privacy. So I understand how this may be of concern to that demographic and why making the position broadcast opt-in is of some value.

I personally don't feel strongly about this either way and can see both sides, happy to go with the flow here.

ks-hl:
Not to derail too much, but I've seen some discussion about making a setup wizard for new nodes. When you connect to a node for the first time, a dialog would pop up that walks you through "Give your device a name", "Select your region", "Select your modem preset", "Do you want to share your location with the mesh?". Can still make it so location defaults to off, but it makes it obvious that it's an option that can be enabled.

fifieldt:
Per @thebentern & @ks-hl , I think we should revisit this after the "welcome wizard" exists

fifieldt:
It doesn't look like we have consensus for this to continue as-is right now, so let's close this one to keep the PR list clean. It'll still be there, should we need it in the future :)

PR #7718, opened 2025-08-22
Log more information about ignored packet

notmasteryet:
I'm running this patch on my local node to analyse reasons for busyRx and Rx errors.

The node runs Portduino version with LR1121 chip. I did not test on other hardware.


GUVWAF (approved):
Looks good to me, can indeed help finding nodes with potential problematic hardware (no TCXO).

PR #7722, opened 2025-08-22
Pkc fix

jp-bennett:
Do PKC encryption if we can, even if the channel isn't 0.

Also, make admin module explicitly request PKC if the request came in over PKC encryption

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -505,7 +505,9 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
     if (mp.decoded.want_response && !myReply) {
         myReply = allocErrorResponse(meshtastic_Routing_Error_NONE, &mp);
     }
-
+    if (mp.pki_encrypted) {
+        myReply->pki_encrypted = true;
+    }
     return handled;
 }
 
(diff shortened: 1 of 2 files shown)


thebentern (approved):

GUVWAF:
I think the original reason for this was for packets that need to be received on a channel with a specific name (legacy admin, GPIO, serial). Maybe we should allow those also when they are received with an admin key?

thebentern:
> I think the original reason for this was for packets that need to be received on a channel with a specific name (legacy admin, GPIO, serial). Maybe we should allow those also when they are received with an admin key?

So check for those special channel names here?
https://github.com/meshtastic/firmware/pull/7722/files#diff-fb9b4a4da3229ad0f8213d3e55e97cb8068e71ba0225002aa050cf06788f6b8aL533

We've been talking about completely removing legacy admin in 2.8, but probably need to account for gpio / serial at the least.

GUVWAF:
Yes, we could also check there if the `channel` maps to any of those with the name "serial" or "gpio" indeed.

thebentern:
@GUVWAF pushed some changes to account for the special channel names.

GUVWAF (approved):
Looks good to me.

src/modules/AdminModule.cpp
```diff
@@ -505,7 +505,9 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
     if (mp.decoded.want_response && !myReply) {
         myReply = allocErrorResponse(meshtastic_Routing_Error_NONE, &mp);
     }
-
+    if (mp.pki_encrypted) {
```
metala, line 508: if (mp.pki_encrypted && myReply) {

Causes crashes sometimes.

PR #7733, opened 2025-08-23
Reduce power of EU433 to 10dBm

fifieldt:
We are currently blocked from making the breaking change to fix EU_433 channel centres until 3.0 (https://github.com/meshtastic/firmware/issues/3371 )

However, as already updated in https://github.com/meshtastic/meshtastic/pull/919 the documentation, the power limit for EU_433 is 10dBm. We can change the power limit without breaking anything, so this patch sets the power limit to match the ETSI spec without changing any other settings.


shalberd:
> We can change the power limit without breaking anything

Very good, thank you

PR #7741, opened 2025-08-25
Add RAK WisMesh Tap V2 (ESP32S3) Hardware Variant

DanielCao0:
This pull request adds support for the RAK WisMesh Tap V2 hardware variant on the ESP32S3 platform. It introduces new configuration files that define pin mappings, hardware features, and build settings for PlatformIO. The changes are grouped as follows:

**New Hardware Variant Definitions:**

* Added `variant.h` to define pin mappings and hardware features for the RAK WisMesh Tap V2, including LEDs, GPS, buttons, battery, buzzer, and SD card support.
* Added `pins_arduino.h` to specify Arduino pin constants and USB VID/PID for the new variant.

**PlatformIO Build Configuration:**

* Added a new `platformio.ini` file with build environments and flags tailored for the RAK WisMesh Tap V2, including TFT and touch panel configuration, library dependencies, and feature toggles.

ce1696ba2f230d3a2d12f70d6873f7dc


fifieldt (approved):

DanielCao0:
> Does the SD card work when running MUI?
> 
> In case not move SDCARD_USE_SPI=1 into the .ini file so that it's defined when the MUI code is compiled.

Yes, SD cards are supported. The SPI for the SD card and the screen are the same SPI.
image

PR #7753, opened 2025-08-26
Feature: Seamless Cross-Preset Communication via UDP Multicast Bridging

ViezeVingertjes:
This Pull Request introduces significant enhancements to how Meshtastic nodes handle and bridge messages across different modem presets, particularly when using UDP multicast. The core objective is to improve compatibility and enable seamless communication between nodes operating on various preset configurations, such as "LongFast" and "MediumFast," without requiring manual channel configuration matching.

Previously, nodes communicating over UDP multicast would often fail to interpret encrypted packets if their local channel configuration (including the PSK) did not exactly match the sender's. This meant that a node on "LongFast" would not be able to decode a packet from a node on "MediumFast" via UDP, even if both were using the default PSK.

This PR addresses this limitation by implementing a "fallback" decryption mechanism for UDP multicast packets. When an encrypted UDP packet arrives and cannot be decrypted with the currently active channel, the system will now attempt to decrypt it using the default PSK in conjunction with a hash derived from each known modem preset name. If a match is found, the packet is successfully decoded, and its channel is re-mapped to the receiving node's primary channel for consistent local processing and re-encoding. This allows nodes on different presets (but sharing the default PSK) to seamlessly exchange messages over UDP.


GUVWAF (commented):
Looks nice, left some comments though.
While it's only applicable for bridging when using the default encryption key and channel names, I think that's the most used case.

src/mesh/udp/UdpMulticastHandler.h
```diff
@@ -54,6 +54,7 @@ class UdpMulticastHandler final
         LOG_DEBUG("Decoding MeshPacket from UDP len=%u", packetLength);
         bool isPacketDecoded = pb_decode_from_bytes(packet.data(), packetLength, &meshtastic_MeshPacket_msg, &mp);
         if (isPacketDecoded && router && mp.which_payload_variant == meshtastic_MeshPacket_encrypted_tag) {
+            mp.transport_mechanism = meshtastic_MeshPacket_TransportMechanism_TRANSPORT_MULTICAST_UDP;
```
GUVWAF: This is already set on line 53, maybe better to remove both and set it after line 55.

src/mesh/Router.cpp
```diff
@@ -416,6 +416,27 @@ DecodeState perhapsDecode(meshtastic_MeshPacket *p)
...
+            crypto->decrypt(p->from, p->id, rawSize, bytes);
+
+            meshtastic_Data decodedtmp;
+            memset(&decodedtmp, 0, sizeof(decodedtmp));
+            if (pb_decode_from_bytes(bytes, rawSize, &meshtastic_Data_msg, &decodedtmp) &&
+                decodedtmp.portnum != meshtastic_PortNum_UNKNOWN_APP) {
+                p->decoded = decodedtmp;
+                p->which_payload_variant = meshtastic_MeshPacket_decoded_tag;
+                // Map to our primary channel index so subsequent processing and re-encoding use local preset
+                chIndex = channels.getPrimaryIndex();
```
GUVWAF: Better to look for the channel index with default encryption key, it's not necessarily the primary channel. You could use the `isDefaultChannel()` method for that.

src/mesh/Router.cpp
```diff
@@ -416,6 +416,27 @@ DecodeState perhapsDecode(meshtastic_MeshPacket *p)
             }
         }
     }
+
+    // Fallback: for UDP multicast, try default preset names with default PSK if normal channel match failed
+    if (!decrypted && p->transport_mechanism == meshtastic_MeshPacket_TransportMechanism_TRANSPORT_MULTICAST_UDP) {
```
GUVWAF, line 422: Let's guard this with an `#ifdef HAS_UDP_MULTICAST` for flash-constrained nRF devices that don't have UDP support.

src/mesh/FloodingRouter.cpp
```diff
@@ -68,6 +68,13 @@ bool FloodingRouter::isRebroadcaster()
 void FloodingRouter::perhapsRebroadcast(const meshtastic_MeshPacket *p)
 {
     if (!isToUs(p) && (p->hop_limit > 0) && !isFromUs(p)) {
+        // If this packet arrived via UDP and is still encrypted, skip rebroadcast.
```
GUVWAF: This causes packets to not be rebroadcasted when they use PKI for other nodes or channels you don't have. I believe this can be safely removed, as the Router will try to decrypt before calling the FloodingRouter; it will not end up here twice.

ViezeVingertjes:
> Looks nice, left some comments though. While it's only applicable for bridging when using the default encryption key and channel names, I think that's the most used case.

Thanks! Yes, it's indeed for the default channels, it might already be `niche` so i tried to keep it as small as possible.
I will address them once i get home; all seem fair comments.

**For context:**
Our personal use-case is that we would like to switch to MF in multiple cities, but a large portion of the users have a hard time on LF already (noise, poor location, maybe not enough nodes in their area yet etc.) and they might otherwise be completely cut-off. So we're trying to get into some hybrid state with at least the big routers and try to get some kind of smooth transition going. 

Time will tell, but first reactions on it functionally were positive in here.

src/mesh/udp/UdpMulticastHandler.h
ViezeVingertjes: Modified in: 81a4c84c2e67af25290759847e5856ca7ce20afc

src/mesh/Router.cpp
ViezeVingertjes: Modified in: 81a4c84c2e67af25290759847e5856ca7ce20afc

src/mesh/Router.cpp
ViezeVingertjes: Modified in: 81a4c84c2e67af25290759847e5856ca7ce20afc

src/mesh/FloodingRouter.cpp
ViezeVingertjes: Modified in: 81a4c84c2e67af25290759847e5856ca7ce20afc

GUVWAF (approved):

src/mesh/udp/UdpMulticastHandler.h
jp-bennett: Are we only setting the transport mechanism if we can decode? That does not seem right.

GUVWAF: "decoded" here is referring to protobuf decoding, not decrypting. 
I realize it can also be in the clause afterwards, because the packet is unused otherwise.

ViezeVingertjes: > Are we only setting the transport mechanism if we can decode? That does not seem right.

Pretty sure there was no code path that does anything with the packet, right? If it cant decode; it just returns and does nothing with it.

ViezeVingertjes: > "decoded" here is referring to protobuf decoding, not decrypting. I realize it can also be in the clause afterwards, because the packet is unused otherwise.

Well spotted; moved -> 264e574f449d95ea73db1c6ab411a406e888c9b6

jp-bennett (approved):

ViezeVingertjes:
Any chance you have the rights te rights to re-run the checks? Had to force push after switching the commits to my personal account instead of work... made it angry. @GUVWAF

GUVWAF:
@thebentern Do you think this is good to merge?

ViezeVingertjes:
> @thebentern Do you think this is good to merge?

That would be great! more and more people over here are interested in setting it up for their area, now the barrier seems to be that they have to build from source, which some can, some cant.

RichieFrame:
This commit seems to have broken the build

ViezeVingertjes:
> This commit seems to have broken the build

That's odd, i have it running on multiple devices. Ill check it out once i get into the office.

ViezeVingertjes:
```
src/DisplayFormatters.h:7:24: note: candidate: 'static const char* DisplayFormatters::getModemPresetDisplayName(meshtastic_Config_LoRaConfig_ModemPreset, bool, bool)'
    7 |     static const char *getModemPresetDisplayName(meshtastic_Config_LoRaConfig_ModemPreset preset, bool useShortName,
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~
src/DisplayFormatters.h:7:24: note:   candidate expects 3 arguments, 2 provided
```

That's weird; it fails on expecting 3 arguments, but the file clearly shows it accepts 2.

```
class DisplayFormatters
{
  public:
    static const char *getModemPresetDisplayName(meshtastic_Config_LoRaConfig_ModemPreset preset, bool useShortName);
};
```

Unclear (yet) what is going on there.

[Edit]
Locally all builds succeed just fine... which is also shown by the last report
{DD08638C-AFC8-4133-B115-048F48528136}

ViezeVingertjes:
I almost feel like there has been some force push on `develop` which explains the differences and why it successfully build and merged, as those are ran before that happened? Thats the best i can think of.

image

Any of you guys have an idea? @GUVWAF @thebentern @jp-bennett 

- All checks succeeded
- ???
- Merged succesfully to `develop` (which should have been fine, as all checks were green, no conflicts)
- Broken?

GUVWAF:
It seems to be caused already by the merge commit 09de0e3edb6d62c8e5b47030d97ca86afd8ff5d2, not this one.

RichieFrame:
This is the PR that changed getModemPresetDisplayName 2 weeks ago:

https://github.com/meshtastic/firmware/pull/7812/files

And this pull's test run showing all firmware builds failing:

https://github.com/meshtastic/firmware/actions/runs/17750552455

ViezeVingertjes:
Only thing i can think of would be that there were merge conflicts and those were accidently incorrectly resolved. As this branch was working properly on August 30th and that commit seems to be from september 2nd

caveman99:
right now a fresh checkout of develop indeed fails. Looks like your branch is out of sync.

caveman99:
@ViezeVingertjes 
grafik
Your branch is 164 commits behind. That's the reason for the desynch. If you fast forward, you'll see the errors as well.

ViezeVingertjes:
> @ViezeVingertjes grafik Your branch is 164 commits behind. That's the reason for the desynch. If you fast forward, you'll see the errors as well.

Hmm, would say it wouldn't merge if there were merge conflicts or something. But we can revert it and then reapply on latest develop too, its just a few lines of code, it has been a PR for a while.

tsuckow:
I'm confused by the need for this. 

I say that as someone looking to implement a mode where a connected device over Bluetooth can switch to seeing only encrypted packets without filtering. Promiscuous mode if you will.

My understanding was the udp protocol was basically that but I don't understand why there would be issues with channel settings on devices then. Unless udp included decrypted packets in which case channel names get messed up because you only have the channel num instead of channel hash.

Tl;Dr: does udp decrypt packets of known channels?

ViezeVingertjes:
> I'm confused by the need for this.
> 
> I say that as someone looking to implement a mode where a connected device over Bluetooth can switch to seeing only encrypted packets without filtering. Promiscuous mode if you will.
> 
> My understanding was the udp protocol was basically that but I don't understand why there would be issues with channel settings on devices then. Unless udp included decrypted packets in which case channel names get messed up because you only have the channel num instead of channel hash.
> 
> Tl;Dr: does udp decrypt packets of known channels?

It's already in use on multiple places currently to bridge multiple presets (LongFast <> MediumFast). That's its only purpose we had for it, as packets were ignored previously due to being encrypted on a different preset. This just accepts those packets, re-encrypt them to the right preset and then rebroadcasts it. That way, people on LongFast can see people on MediumFast and visa versa and public channels messages are also shared.

tsuckow:
Oh so messages on the long fast channel become messages on the medium fast channel and vice versa? 

So udp wouldn't just bridge the radios, it is also merging the primary channel, though it also sounds like it might merge all channels that are unencrypted.

PR #7755, opened 2025-08-26
Don't use pin 0 on RAK for input

jp-bennett:
The rotary encoder and updown input types are very easy to accidentally misconfigure, and soft-brick a RAK 4631 by claiming pin 0, the boot pin. This PR checks for this combination, and just refuses to init pin 0 on the RAK.

### DIFF
--- a/src/input/RotaryEncoderInterruptBase.cpp
+++ b/src/input/RotaryEncoderInterruptBase.cpp
@@ -18,14 +18,23 @@ void RotaryEncoderInterruptBase::init(
     this->_eventCcw = eventCcw;
     this->_eventPressed = eventPressed;
 
-    pinMode(pinPress, INPUT_PULLUP);
-    pinMode(this->_pinA, INPUT_PULLUP);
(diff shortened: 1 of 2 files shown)


src/input/RotaryEncoderInterruptBase.cpp
```diff
@@ -18,14 +18,23 @@ void RotaryEncoderInterruptBase::init(
...
+    bool isRAK = false;
+#ifdef RAK_4631
+    isRAK = true;
+#endif
 
-    //    attachInterrupt(pinPress, onIntPress, RISING);
-    attachInterrupt(pinPress, onIntPress, RISING);
-    attachInterrupt(this->_pinA, onIntA, CHANGE);
-    attachInterrupt(this->_pinB, onIntB, CHANGE);
+    if (!RAK_4631 || pinPress != 0) {
```
thebentern: If we just made the isRAK, should we not use that here?

jp-bennett: Derp. Fix incoming.

PR #7759, opened 2025-08-26
Update M5Stack Core / Basic Pin Defs

RicInNewMexico:
The m5stack_core variant was originally set up to use the v1 version of m5stack's original lora module (now obsolete).

The newer modules retain the same MISO/MOSI/SCK/CS pin definitions, but the pins for DIO0 (IRQ) and RESET have changed.

This PR updates the pin defs for the m5stack core (now actually referred to as the m5stack basic) to work with the currently available lora modules.

## 🤝 Attestations

- [X ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [X] Other: M5Stack Core


RicInNewMexico:
PXL_20250826_165257379

fifieldt:
Thanks @RicInNewMexico . May I ask: do you know when the v1 version was last sold / new one introduced?

Issue #7771, opened 2025-08-27
[Bug]: T-Lora Pager thinks that it's 2005


JThramer:
### Category

Hardware Compatibility

### Hardware

Other

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [x] OLED slide UI on any display

### Firmware Version

2.7.6 (main - 26c38ffc8)

### Description

T-Lora Pager thinks that it's 2005-08-27, which leads to very interesting behavior.

This seems somewhat tied to GPS, which performs pretty terribly on the BaseUI on this device. Other devices I own (including ones with no GPS at all) don't seem to have this problem.

Flagging it because it seems to lead to pretty interesting behavior in-app.

### Relevant log output

```Shell

```

JThramer:
Said interesting behavior: 

![Image](https://github.com/user-attachments/assets/89083b8a-c88d-4755-8dcc-d3d164f80de6) 
("Pager calling" was sent first)

<img width="1290" height="2796" alt="Image" src="https://github.com/user-attachments/assets/6c491884-7941-4191-9975-d81b78f1a529" />
(my pager is apparently ancient history...)

JThramer:
Also seems to pollute the NodeDB: 

![Image](https://github.com/user-attachments/assets/da6e1348-aff0-455e-b789-6ff3746ecfbe)

fifieldt:
Looking into this related one: https://github.com/meshtastic/firmware/issues/7750#issuecomment-3225948816

@JThramer most helpful would be logs - from first boot to 30+mins later. Plug in via USB go to https://flasher.meshtastic.org/ and click Open Serial monitor

JThramer:
@fifieldt No worries, i'll try to pull some logs for you after work. I typically stream the log directly through serial through `minicom` or `meshtastic --noproto`.

JThramer:
@fifieldt as requested, I left a log tail running all night: 

[tio_tty.usbmodem1101_2025-08-28T01:34:49.log](https://github.com/user-attachments/files/22028907/tio_tty.usbmodem1101_2025-08-28T01.34.49.log)

fifieldt:
Thank you very much @JThramer .

This makes it clear the time is coming from the RTC chip, not the GPS as I assumed.

[...]

Issue #7750, opened 2025-08-26
[Bug]: Rak WisMesh Tag time set to 21 May 2019


pelgraine:
### Category

Other

### Hardware

Other

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.5.ddd1499; 2.7.4c1f4f70; 2.7.3.cf574c7

### Description

![Image](https://github.com/user-attachments/assets/2ebef7b8-c044-4dfe-9831-3269d5f93bd3)
![Image](https://github.com/user-attachments/assets/8f9071e4-722d-4c87-b155-d17fd473f2a8)
![Image](https://github.com/user-attachments/assets/ce9c635c-873c-42e2-9e97-b4fc3c8b4b6e)

All new nodes seen in the Rak Wishmesh Tag node list display with the time of 21 May 2019, and also displays node first/last seen as six years ago. 

This time error also seems to be the cause of the Position Exchange Failed error that occurs when trying the Exchange Positions function in IOS, even with GPS enabled on the Tag and iPS Location Permissions set correctly.

I have reflashed with firmware-rak4631-2.6.11.60ec05e in the meantime without issue, other than it being the wrong hardware. 

### Relevant log output

```Shell

```

pelgraine:
Just to add to it, the Tag would display the 2019 date before it had even found other nodes. 

I have five other nodes on hand, including a Seeed T1000e and Heltec Meshpocket on 2.7.5 alpha firmware, that don't have this issue. Only seems to occur with the Tag.

Xaositek:
so this is more of a GPS issue with the L76K giving us the wrong information; here is my "fix" that I have to do if you powered it up inside.

1. Set GPS to disabled either through the screen or an app (iOS or Android)
2. Reboot the device and verify it now gets the proper time
3. Wait about 20 minutes
4. Take the device outside and then enable GPS again
5. Verify the clock does not revert and year is now preserved

pelgraine:
> so this is more of a GPS issue with the L76K giving us the wrong information; here is my "fix" that I have to do if you powered it up inside.
> 
> 1. Set GPS to disabled either through the screen or an app (iOS or Android)
> 2. Reboot the device and verify it now gets the proper time
> 3. Wait about 20 minutes
[...]

PR #7772, opened 2025-08-27
We don't gotTime if time is 2019.

fifieldt:
There are certain GPS chips that have a hard-coded time in firmware that they will return before lock. We set our own hard-coded time, BUILD_EPOCH, that should be newer and use the comparison to not set a bad time.

In https://github.com/meshtastic/firmware/pull/7261 we introduced the RTCSetResult and improved it in https://github.com/meshtastic/firmware/pull/7375 .

However, the original try-fix left logic in GPS.cpp that could still result in broadcasting the bad time.

Further, as part of our fix we cleared the GPS buffer if we didn't get a good time. The mesh was hurting at the time, so this was a reasonable approach. However, given time tends to come in when we're trying to get early lock, this had the potential side effect of throwing away valuable information to get position lock.

This change reverses the clearBuffer and changes the logic so if time is not set it will not be broadcast.

Fixes https://github.com/meshtastic/firmware/issues/7771 
Fixes https://github.com/meshtastic/firmware/issues/7750

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -1504,7 +1504,7 @@ static int32_t toDegInt(RawDegrees d)
  * Perform any processing that should be done only while the GPS is awake and looking for a fix.
  * Override this method to check for new locations
  *
- * @return true if we've acquired a new location
+ * @return true if we've set a new time
  */
 bool GPS::lookForTime()
 {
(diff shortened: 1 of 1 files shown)


fifieldt:
(testing)

fifieldt:
My GPS testing list:
- [x] Seeed Tracker WM1110 w/ Air530 (L76K)
- [x] Seeed 1000e w/ AG3335
- [x] LilyGo TBeam with Ublox 6

fifieldt:
No regressions on 3 common chips, marking as ready for review

b8b8:
Seems OK on T1000E after testing.

thebentern:
Looks good. Seems like a fix we'd want to target at master?

fifieldt:
> Looks good. Seems like a fix we'd want to target at master?

Probably ... but I don't know when to do develop and when to do master at this stage :) Help!

PR #7776, opened 2025-08-28
Fix webserver support for ESP32C6.

Kazagumo:
Related to https://github.com/meshtastic/firmware/pull/4692
update toolchain of esp32c6 so we can get webserver back to work.

## 🤝 Attestations

  - [x] DIY ESP32C6 board.


Kazagumo:
Current Status:
- [x] Webserver
- [ ] Bluetooth : works fine but crash while calling NimBLEDevice::deinit(),waiting for patches from upstream.
- [ ] Paxcounter : Bluetooth works but can't get wifi event.
- [ ] ADC : seems working,further test required.

PR #7781, opened 2025-08-29
Fix freetext hang

### LINKED ISSUE #7714: [Bug]: T-Deck Pro Touchscreen stops working after sending message with keyboard
### Category

Other

### Hardware

T-Deck

- [ ] InkHUD ePaper
- [x] OLED slide UI on any display

### Firmware Version

2.7.6

### Description

Built 2.7.6 firmware from source and flashed T-Deck Pro (4G version)

Description: 
Touchscreen stops working after sending a message with the keyboard. UI is completely unresponsive but does update occasionally. You can still use the device via BLE. Device needs to be reset

If you respond to the message with a preset message, it keeps working. The UI only breaks if you send a message with the keyboard.

Replication:
1. Reset device
2. receive a message
3. reply to message with "freetext" option
4. after hitting "enter" the touchscreen stops working

Workaround:
1. The only way to recover from touchscreen not working is to reboot (reset switch)

### Relevant log output

```Shell

```

thebentern:
Fixes #7714

### DIFF
--- a/bin/org.meshtastic.meshtasticd.metainfo.xml
+++ b/bin/org.meshtastic.meshtasticd.metainfo.xml
@@ -87,6 +87,9 @@
   </screenshots>
 
   <releases>
+    <release version="2.7.7" date="2025-08-28">
+      <url type="details">https://github.com/meshtastic/firmware/releases?q=tag%3Av2.7.7</url>
+    </release>
     <release version="2.7.6" date="2025-08-12">
       <url type="details">https://github.com/meshtastic/firmware/releases?q=tag%3Av2.7.6</url>
     </release>
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-meshtasticd (2.7.6.0) UNRELEASED; urgency=medium
+meshtasticd (2.7.7.0) UNRELEASED; urgency=medium
 
   [ Austin Lane ]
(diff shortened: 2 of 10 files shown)


mverch67 (approved):
Tested with native + TCA8418 keymatrix and it worked

GUVWAF:
This is caused by https://github.com/meshtastic/firmware/pull/7612. I think that is breaking more things also, e.g. [local position updates](https://github.com/meshtastic/firmware/blob/5f8503c62ddb5a1d40cdbf71960d11a08e6d5e07/src/modules/PositionModule.cpp#L70), [SerialModule echo](https://github.com/meshtastic/firmware/blob/5f8503c62ddb5a1d40cdbf71960d11a08e6d5e07/src/modules/SerialModule.cpp#L357), and possibly more.

GUVWAF:
It affects the InkHUD too, whose applets are also modules.
I think we should remove that PR altogether, as it also doesn't prevent spoofed packets to end up at the client, because the RoutingModule sends them there.

GUVWAF (approved):

Xaositek:
Tested on RAK4631 (Meshenger) and T114 (NRF TXT) and everything looks good. Tested new messages and replying to messages on both devices. Also tested Preset Messages and no regressions

PR #7787, opened 2025-08-29
BaseUI Updates

Xaositek:
- Device Role Picker from LoRa frame
- Show Role on LoRa frame
- Move Date to Clock frame
- Account for low resolution wide screen OLEDs (i.e. 256x64)


HarukiToreda (approved):
Tested on OLED, TFT and Eink, no issues. Role picker works as expected and Date does not overlap on any screen resolution. Good to approve

PR #7795, opened 2025-08-30
Metadata: battery percentage

### LINKED ISSUE #7659: [Feature Request]: include battery voltage/percent with "request metadata"
### Platform

Cross-Platform

### Description

So I can check the battery of my remote nodes without having to wait an hour for the device metrics to update

ford-jones:
Resolves #7659 
Includes the remaining battery amount in device metadata responses. 

I've made an accompanying pull request in [protobufs](https://github.com/meshtastic/protobufs/pull/760) which adds the property to the `DeviceMetadata` struct.

The github issue has a comment which also requests that environmental telemetry data be included in the response, I thought I would table it here as I think that might need some discussion.

## Testing Method:
1. Navigate to `firmware` and pull branch
2. Pull protobuf changes from [https://github.com/ford-jones/protobufs/tree/7659-voltage-in-metadata](https://github.com/ford-jones/protobufs/tree/7659-voltage-in-metadata)
3. Run `./bin/regen-protos.sh`
4. Build and flash to device
5. Navigate to `python` (cli), repeat 2 & 3
6. Run:
```
python3 ./meshtastic/__main__.py --info
```

## 🤝 Attestations

  - [x] Heltec (Lora32) V3


capricornusx:
Hi, I noticed that the proposed PR adds the remaining battery amount to the `DeviceMetadata` struct, but the protobuf files already have a [DeviceMetrics](https://github.com/meshtastic/protobufs/blob/master/meshtastic/telemetry.proto#L18) struct containing battery-related fields like `BatteryLevel`. It seems the PR might be duplicating existing functionality.

Could you please clarify if this duplication is intentional or if the PR should be adjusted to avoid overlapping with `DeviceMetrics`? Also, regarding the request to include environmental telemetry data mentioned in the GitHub issue, would it be better to discuss it separately?

garthvh:
Same comment, there are already ways to do this and it makes the device metadata packet which is used by the apps to establish admin connectivity larger.

ford-jones:
Agree with you both guys, I think we can close this. 
I was trying to close some of the low hanging fruit and missed this overlap.

Cheers.

fifieldt:
Yup, it seems like the user enjoys the request metadata functionality and wishes it was possible to send requests for other types of data. So this should be addressed in another way.

PR #7796, opened 2025-08-31
Portduino config refactor

jp-bennett:
Get rid of the raw settings maps, move everything to the portduino_config struct
Add the -y flag to print out the resulting yaml config
move all gpio to the pin/gpiochip/line scheme


jp-bennett:
Have an idea how to make GPIO handling even better. Testing.

mverch67:
Currently, when I run native-tft I get a black X11 screen:

image

Edit: this black screen may be related to my environment of the linux box... it also happens on master

mverch67:
> Edit: this black screen may be related to my environment of the linux box... it also happens on master

The black screen issue was caused by the I2C keyboard scanning (removed again [here](https://github.com/meshtastic/device-ui/commit/973f9ed4b929469bbf785b12592dfb4326ada701))

PR #7797, opened 2025-08-31
Make ExternalNotification show up in excluded_modules, more STM32 modules

Stary2001:
ExternalNotification wasn't added to excluded_modules in device metadata, which https://github.com/meshtastic/python/pull/818 relies on.
Also, enable more stuff on STM32, again.
At some point I want to clean up some of these ifdefs into something more manageable, but that'll be a future PR...

## 🤝 Attestations
  - [x] Wio-E5 board


fifieldt (approved):

Stary2001:
I've fixed rak3172 and tested on hardware.

thebentern (approved):

PR #7802, opened 2025-09-01
Add support for the RV-3028 on native Linux

jp-bennett:
Also includes a minor logic fix in readFromRTC()
Also downgrades the Linux system time to RTCQualityDevice


fifieldt (approved):
Logic checks out

thebentern (approved):

PR #7812, opened 2025-09-01
If usePreset is False, show value as Custom

Xaositek:
If usePreset is False, show on-screen value as Custom rather than relying upon modemPreset


Xaositek:
7DA41086-2D72-4DB3-96BE-82523EC216CB_1_102_o

PR #7813, opened 2025-09-01
Add RF switch settings for STM32WL variants

Stary2001:
RF switch settings per-board for STM32WL. Allows rak3172/ebyte e77 modules to work properly!

Requires the change to boards/wiscore_rak3172.json from #7797 to see any observable effect, because defining it as `PB8` did not toggle the actual pin PB8.

## 🤝 Attestations

  - [x] Wio-E5
  - [x] RAK3172


Stary2001:
ugh, every platform builds STM32WLE5JCInterface...

fifieldt (approved):

PR #7824, opened 2025-09-02
Revert "Add GAT562 Mesh Tracker Pro"

thebentern:
Reverts meshtastic/firmware#7815


thebentern:
This was not reviewed

PR #7825, opened 2025-09-02
Try-fix Cardkb detection

thebentern:


Xaositek (approved):
No regressions found in NRF TXT (Heltec T114) or Meshenger (RAK4631); additionally both devices were allowed to screen sleep (60 second timeout) and when awoken, all functionality remained present.

HarukiToreda:
Tested on Heltec V3 with cardKB. This seems to have solve the issue where ESP32 would wake from power save sleep but was no longer initializing the cardkb unless it manually rebooted. Now it wakes and works with no issues.

PR #7828, opened 2025-09-02
Add support for new ESP32 DIY variant 9m2ibr_aprs_lora_tracker

ndoo:
9M2IBR APRS LoRa Tracker: ESP32-WROOM-32 + EBYTE E22-400M30S
https://shopee.com.my/product/1095224/21692283917

Originally developed for LoRa_APRS_iGate and GPIO assignment is similar to https://github.com/richonguzman/LoRa_APRS_iGate/blob/main/variants/ESP32_DIY_1W_LoRa_Mesh_V1_2/board_pinout.h

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

https://github.com/user-attachments/assets/aafcdb88-2c0e-43a4-b520-879b256345e0


ndoo:
Pushed a commit to fix GPS Tx/Rx pin swap (the schematic named the nets from the perspective of the GPS rather than the MCU).

I have an early schematic of the device but I'm not sure I am allowed to share it as it is non-public information AFAICT.

PR #7852, opened 2025-09-04
Add a new GPS model CM121.

Quency-D:
## 🤝 Attestations

  - [x] Heltec (Lora32) V4
  - [x] Heltec t114

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -808,6 +808,14 @@ bool GPS::setup()
             } else {
                 LOG_INFO("GNSS module configuration saved!");
             }
+        } else if (gnssModel == GNSS_MODEL_CM121) {
+            // only ask for RMC and GGA
+            //enable GGA
+            _serial_gps->write("$CFGMSG,0,0,1,1*1B\r\n");
+            delay(250);
+            //enable RMC
+            _serial_gps->write("$CFGMSG,0,4,1,1*1F\r\n");
+            delay(250);
         }
         didSerialInit = true;
     }
@@ -1240,7 +1248,14 @@ GnssModel_t GPS::probe(int serialSpeed)
     _serial_gps->write("$PUBX,40,GSV,0,0,0,0,0,0*59\r\n");
     _serial_gps->write("$PUBX,40,VTG,0,0,0,0,0,0*5E\r\n");
     delay(20);
-
+    // Close NMEA sequences on CM121
+    _serial_gps->write("$CFGMSG,0,0,0,1*1A\r\n");
+    _serial_gps->write("$CFGMSG,0,1,0,1*1B\r\n");
+    _serial_gps->write("$CFGMSG,0,2,0,1*18\r\n");
+    _serial_gps->write("$CFGMSG,0,3,0,1*19\r\n");
+    _serial_gps->write("$CFGMSG,0,4,0,1*1E\r\n");
+    delay(20);
+    
     // Unicore UFirebirdII Series: UC6580, UM620, UM621, UM670A, UM680A, or UM681A
     std::vector<ChipInfo> unicore = {{"UC6580", "UC6580", GNSS_MODEL_UC6580}, {"UM600", "UM600", GNSS_MODEL_UC6580}};
     PROBE_FAMILY("Unicore Family", "$PDTINFO", unicore, 500);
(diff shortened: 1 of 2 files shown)


fifieldt:
Unicore?

src/gps/GPS.cpp
```diff
@@ -1240,7 +1248,14 @@ GnssModel_t GPS::probe(int serialSpeed)
...
+    // Close NMEA sequences on CM121
+    _serial_gps->write("$CFGMSG,0,0,0,1*1A\r\n");
+    _serial_gps->write("$CFGMSG,0,1,0,1*1B\r\n");
+    _serial_gps->write("$CFGMSG,0,2,0,1*18\r\n");
+    _serial_gps->write("$CFGMSG,0,3,0,1*19\r\n");
+    _serial_gps->write("$CFGMSG,0,4,0,1*1E\r\n");
+    delay(20);
+    
     // Unicore UFirebirdII Series: UC6580, UM620, UM621, UM670A, UM680A, or UM681A
     std::vector<ChipInfo> unicore = {{"UC6580", "UC6580", GNSS_MODEL_UC6580}, {"UM600", "UM600", GNSS_MODEL_UC6580}};
```
fifieldt: Can probably just add {"CM121", "CM121", GNSS_MODEL_CM121} here, since everything in this family uses PDTINFO. Should result in slightly faster detection :)

If the CM121 requires the checksum (for earlier Unicore chips it was optional), we can probably add the checksum for everything here.

src/gps/GPS.cpp
```diff
@@ -808,6 +808,14 @@ bool GPS::setup()
             } else {
                 LOG_INFO("GNSS module configuration saved!");
             }
+        } else if (gnssModel == GNSS_MODEL_CM121) {
+            // only ask for RMC and GGA
```
fifieldt, line 812: Might want to insert a line to enable the correct satelllite "constellations" (i.e. BDS, GPS, GLONASS etc).

Should be able to find CFGSYS in the manual

_serial_gps->write("$CFGSYS,h35155\r\n");

src/gps/GPS.cpp
Quency-D: Checksum is optional and can be left unused.

src/gps/GPS.cpp
Quency-D: The default configuration is directly used here. Do I have to reconfigure it again? Because different models may have different configurations.
image

fifieldt: If it's difficult, it's probably OK to leave it at factory defaults for now. We do set this explicitly for other chips, since there is no guarantee the user has kept the default configuration.

We do also perform some optimisations for the user for some chips based on the lora region (eg enable NAVIC in India), but that is early work.

Patch is definitely OK as is, just something to consider :)

fifieldt:
This patch is good to go.

However, since it adds extra commands to the probe code I'm just going to test the detection code on other GPS chips still performs as normal :) Should be 100% OK, just part of my regular testing for this section of code.

- [x] Seeed Tracker WM1110 w/ Air530 (L76K)
- [x] Seeed 1000e w/ AG3335
- [x] LilyGo TBeam with Ublox 6

src/gps/GPS.cpp
```diff
@@ -1240,9 +1248,16 @@ GnssModel_t GPS::probe(int serialSpeed)
     _serial_gps->write("$PUBX,40,GSV,0,0,0,0,0,0*59\r\n");
     _serial_gps->write("$PUBX,40,VTG,0,0,0,0,0,0*5E\r\n");
     delay(20);
-
-    // Unicore UFirebirdII Series: UC6580, UM620, UM621, UM670A, UM680A, or UM681A
-    std::vector<ChipInfo> unicore = {{"UC6580", "UC6580", GNSS_MODEL_UC6580}, {"UM600", "UM600", GNSS_MODEL_UC6580}};
+    // Close NMEA sequences on CM121
+    _serial_gps->write("$CFGMSG,0,0,0,1*1A\r\n");
```
fifieldt: message type 0 = GGA. We don't need to turn this one off.

src/gps/GPS.cpp
```diff
@@ -1240,9 +1248,16 @@ GnssModel_t GPS::probe(int serialSpeed)
...
+    _serial_gps->write("$CFGMSG,0,1,0,1*1B\r\n");
+    _serial_gps->write("$CFGMSG,0,2,0,1*18\r\n");
+    _serial_gps->write("$CFGMSG,0,3,0,1*19\r\n");
+    _serial_gps->write("$CFGMSG,0,4,0,1*1E\r\n");
```
fifieldt: message type 4 = RMC. We don't need to turn this one off.

fifieldt:
OK, good thing I tested-- this sorta accidentally broke the Heltec Wireless Tracker UC6580 ... since it disabled GGA and RMC and the code for that does not turn them on explicitly.

Simple fix - just turn off fewer things during probe.

src/gps/GPS.cpp
fifieldt: uploaded a patch to fix this

src/gps/GPS.cpp
fifieldt: uploaded a patch to fix this

fifieldt (approved):

fifieldt:
Thank you for your patience, the GPS detect code is a little sensitive :) I think we're good to go!

Quency-D:
> OK, good thing I tested-- this sorta accidentally broke the Heltec Wireless Tracker UC6580 ... since it disabled GGA and RMC and the code for that does not turn them on explicitly.
> 
> Simple fix - just turn off fewer things during probe.

Thank you for your patient testing. Our previous testing was not comprehensive enough.

Quency-D:
> Thank you for your patience, the GPS detect code is a little sensitive :) I think we're good to go!

Very good.

lyusupov:
@Quency-D 
is there any datasheet for the CM121 ?

PR #7864, opened 2025-09-04
Add Adaptive Polling Intervals to WebServer

capricornusx:
### 🤔 Problem

WebServer currently polls every `5ms` regardless of activity, consuming unnecessary CPU resources during idle periods.

### 🦥 Solution

Replace fixed 5ms interval with adaptive polling based on HTTP request activity:

* `50ms` - Active mode (first 5 seconds after HTTP request)
* `200ms` - Medium mode (5-30 seconds after last activity)
* `1000ms` - Idle mode (30+ seconds without requests)

### 📝 Implementation

* Added `markActivity()` calls to HTTP request handlers
* Implemented `getAdaptiveInterval()` for dynamic timing
* Preserves all existing functionality

This optimization improves battery life on devices with WiFi enabled while keeping the web interface fully responsive
when needed.

### 📊 Measurement Results
Measured on ESP32 (`esp32-pico-d4`) over 30-second intervals:

| Metric       | Original (5ms) |   	Optimized (Adaptive)    |  Improvement   |
|--------------|:--------------:|:--------------------------:|:--------------:|
| Memory Usage |     	1.5%      |            0.1%            | 155x reduction |
| CPU Usage    |    	15.5%	     |            0.1%            | 155x reduction |
| Polling Rate |   	168/sec	    |        1/sec (idle)        | 168x reduction |
| Battery Life |   	Baseline    | 	+5-10% 	When WiFi enabled |                |

🔋 Benefits up to ~10% battery life improvement

## 🤝 Attestations

    - [x] TLORA_V2_1_16

### DIFF
--- a/src/mesh/http/ContentHandler.cpp
+++ b/src/mesh/http/ContentHandler.cpp
@@ -148,6 +148,8 @@ void registerHandlers(HTTPServer *insecureServer, HTTPSServer *secureServer)
 
 void handleAPIv1FromRadio(HTTPRequest *req, HTTPResponse *res)
 {
+    if (webServerThread)
+        webServerThread->markActivity();
 
     LOG_DEBUG("webAPI handleAPIv1FromRadio");
 
@@ -380,6 +382,9 @@ void handleFsDeleteStatic(HTTPRequest *req, HTTPResponse *res)
 
 void handleStatic(HTTPRequest *req, HTTPResponse *res)
 {
+    if (webServerThread)
+        webServerThread->markActivity();
+
     // Get access to the parameters
     ResourceParameters *params = req->getParams();
 
--- a/src/mesh/http/WebServer.cpp
+++ b/src/mesh/http/WebServer.cpp
@@ -175,6 +175,25 @@ WebServerThread::WebServerThread() : concurrency::OSThread("WebServer")
     if (!config.network.wifi_enabled && !config.network.eth_enabled) {
         disable();
     }
+    lastActivityTime = millis();
+}
+
+void WebServerThread::markActivity()
+{
+    lastActivityTime = millis();
+}
+
+int32_t WebServerThread::getAdaptiveInterval()
+{
+    uint32_t timeSinceActivity = millis() - lastActivityTime;
+
+    if (timeSinceActivity < 5000) {
+        return 50;
+    } else if (timeSinceActivity < 30000) {
+        return 200;
+    } else {
+        return 1000;
+    }
 }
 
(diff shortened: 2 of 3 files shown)


capricornusx:
:heavy_check_mark:  fixed

fifieldt (approved):
this is great, thank you!

capricornusx:
up

cpatulea:
Just a side question, how do you measure CPU usage? @capricornusx

capricornusx:
@cpatulea extrapolation :)

cpatulea:
Could you please share a little more detail? I suspect that some problems on my node might be caused by CPU usage. But Meshtastic does not expose CPU usage. and everywhere I find that "measuring CPU usage on ESP32 is hard". I would really appreciate if you could explain what you did.

capricornusx:
@cpatulea

How I estimated CPU impact
- Model: CPU cost of the web thread ≈ cost_per_loop × poll_rate.
- old_interval_ms = 5 ms.
- new_avg_interval = p_active*50 + p_medium*200 + p_idle*1000 (p_* are fractions summing to 1).
- Approximate saving: CPU_saving ≈ 1 − old_interval_ms / new_avg_interval.
- Example (p_active=0.1, p_medium=0.2, p_idle=0.7): new_avg_interval = 745 ms → ~99.3% reduction in poll frequency (5 / 745 ≈ 1/149).

Practical measurement advice (ESP32):
1) Instrument the server loop timing (recommended, simple and precise):
- Measure execution time around secureServer->loop()/insecureServer->loop(), log average, then compute CPU% = loop_time_ms / poll_interval_ms.

```cpp
// cpp
uint64_t t0 = esp_timer_get_time();
secureServer->loop();
insecureServer->loop();
uint64_t t1 = esp_timer_get_time();
double loop_ms = (t1 - t0) / 1000.0;
double cpu_pct = (loop_ms / interval_ms) * 100.0;
```

2) FreeRTOS run-time stats (global view of tasks):
- Enable CONFIG_FREERTOS_GENERATE_RUN_TIME_STATS and provide a run-time counter (portGET_RUN_TIME_COUNTER_VALUE implementation).
- Call vTaskGetRunTimeStats() to get per-task CPU share.

3) External/low-level: toggle a GPIO at loop start/end and measure duty cycle with logic analyzer/oscilloscope.

- The simplified CPU_saving formula assumes loop() cost is small and constant; real savings depend on actual loop execution time and interactions with other tasks, interrupts, WiFi stack.

PR #7867, opened 2025-09-05
Show message popup over VirtualKeyboard

whywilson:
Popup and preview message when new message comes when VirtualKeybaord on screen.
Meshtiny

  - [x] [Meshtiny](https://shop.mtoolstec.com/product/meshtiny)
  - [x] [Wio Tracker L1 Pro](https://shop.mtoolstec.com/product/seeed-wio-tracker-l1-pro)
  - [x] [GAT562 Pro](https://shop.mtoolstec.com/product/gat562-mesh-tracker)
  - [x] Heltec T114

### DIFF
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -149,6 +149,11 @@ void Screen::showSimpleBanner(const char *message, uint32_t durationMs)
 // Called to trigger a banner with custom message and duration
 void Screen::showOverlayBanner(BannerOverlayOptions banner_overlay_options)
 {
+    // Don't show overlay banner if virtual keyboard is active
+    if (NotificationRenderer::current_notification_type == notificationTypeEnum::text_input) {
+        return;
+    }
+
 #ifdef USE_EINK
     EINK_ADD_FRAMEFLAG(dispdev, DEMAND_FAST); // Skip full refresh for all overlay menus
 #endif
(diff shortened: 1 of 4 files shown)


src/graphics/Screen.cpp
```diff
@@ -1400,21 +1415,28 @@ int Screen::handleTextMessage(const meshtastic_MeshPacket *packet)
...
+
+                // Only wake/force display if the configuration allows it
+                if (shouldWakeOnReceivedMessage()) {
+                    setOn(true);    // Wake up the screen first
+                    forceDisplay(); // Forces screen redraw
+                }
+            } else {
+                // Virtual keyboard is active - just mark that frames need regeneration when keyboard closes
+                // The devicestate and hasUnreadMessage are already set above, so message will appear later
+                LOG_INFO("Virtual keyboard active - deferring frame list update for new message");
```
fifieldt: This should probably be LOG_DEBUG level, rather than LOG_INFO, since the audience of the message is developers, rather than users.

Xaositek:
I am adding in @thebentern and @jp-bennett for reviewers - I understand the need, but this adds a LOT of new code to support into the project. Its a lot of redundant code and I want to think through if this is the best approach.

jp-bennett:
I think we're seeing the inherent problem with treating the on-screen keyboard as a notification.

I'd really like to see an approach that both virtual keyboards could use, maybe a module frame?

whywilson:
> I think we're seeing the inherent problem with treating the on-screen keyboard as a notification.
> 
> 
> 
> I'd really like to see an approach that both virtual keyboards could use, maybe a module frame? 

The virtual keyboard for SenseCAP indicator vs this new bring on-screen keyboard in one module?

jp-bennett:
> 
> The virtual keyboard for SenseCAP indicator vs this new bring on-screen keyboard in one module?

Probably two separate modules, but they do basically the same thing and are implemented in wildly different ways. If we can nail down what the ideal implementation looks like, we can refactor both to use that approach.

whywilson:
> > 
> > The virtual keyboard for SenseCAP indicator vs this new bring on-screen keyboard in one module?
> 
> Probably two separate modules, but they do basically the same thing and are implemented in wildly different ways. If we can nail down what the ideal implementation looks like, we can refactor both to use that approach.
> 
> 

I can try to reconstruct osk into a module first, but not only need to be modified here, but also the structure of the whole notification needs to be optimized.

whywilson:
This PR will be continued here https://github.com/meshtastic/firmware/pull/8379.

PR #7870, opened 2025-09-05
Use char buffer for probeResponse

thebentern:
While looking at some _other_ memory leak vectors, this one came up as a fairly straightforward change to get rid of another heap allocation


fifieldt:
Back when GPS probe was in its "unreliable" days, we used something similar to this. If this is the way to go (and I do not have anywhere near sufficient C++ knowledge to say one way or another), we must test this incredibly well :)

fifieldt (commented):
@thebentern - 256 may not be enough, especially for 115200 baud rate (probe runs for 500ms 
.. at 14.4 bytes per ms that could.be way over). Our buffers for most normal platforms are 4096.

thebentern:
> @thebentern - 256 may not be enough, especially for 115200 baud rate (probe runs for 500ms .. at 14.4 bytes per ms that could.be way over). Our buffers for most normal platforms are 4096.

We can absolutely extend the buffer, but I've been testing with the Wireless Tracker and it's been detecting the Unicore without fail on multiple boots.

fifieldt:
Glad to hear it! Try the T1000E - that one's sensitive.

fifieldt:
Testing hitlist:
- [ ] Seeed WM1110 w/ Air530 (L76K)
- [x] Seeed L1 Tracker Pro w/ L76K
- [x] Seeed Tracker 1000e w/ AG3335 (detected on 2nd scan) 
- [ ] LilyGo TBeam with Ublox 6
- [x] RAK WisMesh Pocket with Ublox 8

fifieldt:
Also, if possible, it's best to try and test with a GPS chip that has been factory reset (or a device that has never had Meshtastic on it before). We turn off a lot of crap once we setup a chip, so detection gets easier after we do it the first time.

thebentern:
> Also, if possible, it's best to try and test with a GPS chip that has been factory reset (or a device that has never had Meshtastic on it before). We turn off a lot of crap once we setup a chip, so detection gets easier after we do it the first time.

Good idea. I remember some of the early T-Beams ublox had a lot of garbage turned on initially.

Issue #7863, opened 2025-09-04
[Feature Request]: CLIENT_BASE mode: ROUTER for favorites, CLIENT otherwise (for attic/roof nodes!)


compumike:
### Platform

Cross-Platform

### Description

Like many Meshtastic fans, I currently have a roof/attic node (`CLIENT` mode), plus one or more mobile nodes (`CLIENT_MUTE` or `CLIENT`) which are indoors and/or much less well positioned. **The current modes and routing algorithm do not let me take advantage of my relatively strong roof/attic node.**

Due to managed flooding SNR prioritization, my mobile node’s outgoing packets are usually picked up and rebroadcasted by a relatively weak (probably indoor) neighbor node elsewhere in my town, not by my much better-positioned base station.

-----

### Proposal: `CLIENT_BASE` mode (for attic/roof nodes)

I propose a new `CLIENT_BASE` mode, to be suggested for possible use on attic/roof nodes, with the following behavior:

1. Acts as `ROUTER` (early, guaranteed rebroadcast) for any packets where either `p->from` OR `p->to` are an `is_favorite` node.
2. Acts as `CLIENT` otherwise.

(Seriously, that's it.)

-----

### Super Easy Implementation

Adding this `CLIENT_BASE` mode requires:

- no extra firmware state
- no changes to packets
- very few lines of code

In fact, it might just be one extra `else if` condition in `RadioInterface::getTxDelayMsecWeighted`: :smile:

https://github.com/meshtastic/firmware/blob/f31fd34ce049d4720a2adc69101fb9fe2e8d1b62/src/mesh/RadioInterface.cpp#L325

-----

### Actually Uses Attic/Roof Nodes

With attic/roof nodes set to `CLIENT_BASE`, those stronger nodes will actually get used by the routing! This may encourage people to actually deploy more of these stronger nodes.

-----

### Improves Packet Delivery Reliability

Currently, packets from my relatively weak indoor node are most likely getting rebroadcast by some other relatively weak indoor node.

With `CLIENT_BASE`, messages sent from a weaker/indoor/mobile device will actually take advantage of their nearby attic/roof base station, which currently does not get utilized reliably or at all. This may significantly extend the range at which messages can be reliably exchanged in a single hop.

-----

### Falls Back Perfectly

When not near a CLIENT_BASE, the mobile CLIENT/CLIENT_MUTE just works as normal.

-----

### Congestion Neutral (maybe even negative!)

[...]

PR #7873, opened 2025-09-05
Add `CLIENT_BASE` role: `ROUTER` for favorites, `CLIENT` otherwise (for attic/roof nodes!)

compumike:
Initial implementation of https://github.com/meshtastic/firmware/issues/7863 , following the general shape of @GUVWAF 's comment at: https://github.com/meshtastic/firmware/issues/7863#issuecomment-3257400576

This PR adds a `CLIENT_BASE` role. This role behaves as `ROUTER` for packets from or to any favorited node, and acts as `CLIENT` otherwise. Please see issue #7863 for more details.

This PR requires this update to protobufs: https://github.com/meshtastic/protobufs/pull/763 which defines `CLIENT_BASE = 12` in the `Config.DeviceConfig.Role` enum.

This compiles but I have NOT had the chance to test it yet. Feel free to review commit-by-commit: may be easier to follow that way. Hope it gets the conversation going! Thanks. ❤️

### DIFF
--- a/src/mesh/FloodingRouter.cpp
+++ b/src/mesh/FloodingRouter.cpp
@@ -43,11 +43,30 @@ bool FloodingRouter::shouldFilterReceived(const meshtastic_MeshPacket *p)
     return Router::shouldFilterReceived(p);
 }
 
+bool FloodingRouter::roleAllowsCancelingDupe(const meshtastic_MeshPacket *p)
+{
+    if (config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER ||
+        config.device.role == meshtastic_Config_DeviceConfig_Role_REPEATER ||
+        config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER_LATE) {
+        // ROUTER, REPEATER, ROUTER_LATE should never cancel relaying a packet (i.e. we should always rebroadcast),
+        // even if we've heard another station rebroadcast it already.
+        return false;
+    }
+
+    if (config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT_BASE) {
+        // CLIENT_BASE: if the packet is from or to a favorited node,
+        // we should act like a ROUTER and should never cancel a rebroadcast (i.e. we should always rebroadcast),
+        // even if we've heard another station rebroadcast it already.
+        return !nodeDB->isFromOrToFavoritedNode(*p);
+    }
+
+    // All other roles (such as CLIENT) should cancel a rebroadcast if they hear another station's rebroadcast.
+    return true;
+}
+
 void FloodingRouter::perhapsCancelDupe(const meshtastic_MeshPacket *p)
 {
-    if (config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER &&
-        config.device.role != meshtastic_Config_DeviceConfig_Role_REPEATER &&
-        config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER_LATE &&
+    if (roleAllowsCancelingDupe(p) &&
         p->transport_mechanism == meshtastic_MeshPacket_TransportMechanism_TRANSPORT_LORA) {
         // cancel rebroadcast of this message *if* there was already one, unless we're a router/repeater!
         // But only LoRa packets should be able to trigger this.
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -1750,6 +1750,20 @@ void NodeDB::set_favorite(bool is_favorite, uint32_t nodeId)
     }
 }
 
+bool NodeDB::isFavorite(uint32_t nodeId)
+{
+    meshtastic_NodeInfoLite *lite = getMeshNode(nodeId);
+    if (lite) {
+        return lite->is_favorite;
+    }
+    return false;
+}
+
+bool NodeDB::isFromOrToFavoritedNode(const meshtastic_MeshPacket &p)
+{
+    return isFavorite(p.from) || isFavorite(p.to);
+}
+
 void NodeDB::pause_sort(bool paused)
 {
     sortingIsPaused = paused;
--- a/src/mesh/RadioInterface.h
+++ b/src/mesh/RadioInterface.h
@@ -180,6 +180,9 @@ class RadioInterface
     /** The worst-case SNR_based packet delay */
     uint32_t getTxDelayMsecWeightedWorst(float snr);
 
+    /** Returns true if we should rebroadcast early like a ROUTER */
+    bool shouldRebroadcastEarlyLikeRouter(meshtastic_MeshPacket *p);
+
     /** The delay to use when we want to flood a message. Use a weighted scale based on SNR */
     uint32_t getTxDelayMsecWeighted(float snr, meshtastic_MeshPacket *p);
 
--- a/src/mesh/RadioInterface.h
+++ b/src/mesh/RadioInterface.h
@@ -181,7 +181,7 @@ class RadioInterface
     uint32_t getTxDelayMsecWeightedWorst(float snr);
 
     /** The delay to use when we want to flood a message. Use a weighted scale based on SNR */
-    uint32_t getTxDelayMsecWeighted(float snr);
+    uint32_t getTxDelayMsecWeighted(float snr, meshtastic_MeshPacket *p);
 
     /** If the packet is not already in the late rebroadcast window, move it there */
     virtual void clampToLateRebroadcastWindow(NodeNum from, PacketId id) { return; }
--- a/src/mesh/generated/meshtastic/config.pb.h
+++ b/src/mesh/generated/meshtastic/config.pb.h
@@ -64,7 +64,12 @@ typedef enum _meshtastic_Config_DeviceConfig_Role {
     in areas not already covered by other routers, or to bridge around problematic terrain,
     but should not be given priority over other routers in order to avoid unnecessaraily
     consuming hops. */
-    meshtastic_Config_DeviceConfig_Role_ROUTER_LATE = 11
+    meshtastic_Config_DeviceConfig_Role_ROUTER_LATE = 11,
+    /* Description: Treats packets from or to favorited nodes as ROUTER, and all other packets as CLIENT.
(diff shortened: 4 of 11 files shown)


fifieldt:
Hi @compumike , are you familiar with the trunk linting tool? We use it to keep our code consistently formatted. Usually just need to add the plugin to vscode and it will work everything out.

src/mesh/NodeDB.cpp
```diff
@@ -1750,6 +1750,20 @@ void NodeDB::set_favorite(bool is_favorite, uint32_t nodeId)
...
+    meshtastic_NodeInfoLite *lite = getMeshNode(nodeId);
+    if (lite) {
+        return lite->is_favorite;
+    }
+    return false;
+}
+
+bool NodeDB::isFromOrToFavoritedNode(const meshtastic_MeshPacket &p)
+{
+    return isFavorite(p.from) || isFavorite(p.to);
```
GUVWAF: Let's check here first if `p.to != NODENUM_BROADCAST` to avoid looping through the DB searching for a node that will not exist.

src/mesh/RadioInterface.h
```diff
@@ -180,8 +180,11 @@ class RadioInterface
     /** The worst-case SNR_based packet delay */
     uint32_t getTxDelayMsecWeightedWorst(float snr);
 
+    /** Returns true if we should rebroadcast early like a ROUTER */
+    bool shouldRebroadcastEarlyLikeRouter(meshtastic_MeshPacket *p);
+
     /** The delay to use when we want to flood a message. Use a weighted scale based on SNR */
-    uint32_t getTxDelayMsecWeighted(float snr);
+    uint32_t getTxDelayMsecWeighted(float snr, meshtastic_MeshPacket *p);
```
GUVWAF: We can remove the `snr` parameter here, and in the function get from `p`.

src/mesh/FloodingRouter.cpp
```diff
@@ -43,11 +43,30 @@ bool FloodingRouter::shouldFilterReceived(const meshtastic_MeshPacket *p)
...
+    // All other roles (such as CLIENT) should cancel a rebroadcast if they hear another station's rebroadcast.
+    return true;
+}
+
 void FloodingRouter::perhapsCancelDupe(const meshtastic_MeshPacket *p)
 {
-    if (config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER &&
-        config.device.role != meshtastic_Config_DeviceConfig_Role_REPEATER &&
-        config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER_LATE &&
+    if (roleAllowsCancelingDupe(p) &&
```
GUVWAF: Maybe put `p->transport_mechanism == meshtastic_MeshPacket_TransportMechanism_TRANSPORT_LORA` first for short-circuit evaluation, to avoid looping through the database when it's not needed.

GUVWAF (commented):
Looks good to me in general, left three comments. 
Some testing would be good.

src/mesh/generated/meshtastic/config.pb.h
```diff
@@ -64,7 +64,12 @@ typedef enum _meshtastic_Config_DeviceConfig_Role {
     in areas not already covered by other routers, or to bridge around problematic terrain,
     but should not be given priority over other routers in order to avoid unnecessaraily
     consuming hops. */
-    meshtastic_Config_DeviceConfig_Role_ROUTER_LATE = 11
+    meshtastic_Config_DeviceConfig_Role_ROUTER_LATE = 11,
```
fifieldt: these files are automatically generated, so changes to these files should be removed from the patch

compumike:
@GUVWAF and @fifieldt , thank you, I believe I've just pushed a few commits that I believe resolve all of your suggestions!

I also had the opportunity to test this locally with a few nodes by watching the serial debug logs on my `CLIENT_BASE`, sending traceroutes from a `CLIENT` to another `CLIENT`, and I can confirm that:

- when neither from or to are favorited, it has a long delay (1500+ ms) and prints `rx_snr found in packet. Setting tx delay` with no `Router:`
- when both from and to are favorited, it has a short delay and prints `rx_snr found in packet. Router: setting tx delay` with the `Router: ` part
- when only from is favorited, short and `Router:`
- when only to is favorited, short and `Router:`

This looks good! ✅ 

@fifieldt I believe the CI checks are now failing because I removed the changes to `src/mesh/generated/meshtastic/config.pb.h` as you requested. Please let me know the best way to handle that.

compumike:
@fifieldt I rebased on `develop` and it builds in CI now. 😄 

I also fixed a logic bug in `shouldWakeOnReceivedMessage` in `src/graphics/Screen.cpp`. (It wasn't actually comparing `config.device.role` to the enum value.)

compumike:
@GUVWAF and @fifieldt There are three places I've identified where there is some special behavior for `ROUTER` and I'm not sure if any behavior needs to change to handle `CLIENT_BASE`, but I wanted to run it by you.

-----

1. `shouldLoraWake`

https://github.com/compumike/meshtastic-firmware/blob/95c5a94ad37cd78ac4cb8711ad97ef4870d34332/src/sleep.cpp#L526-L527

This might need to include `CLIENT_BASE` as well, so the device wakes up fast enough? But I'm not sure if it matters because I thought `CLIENT` couldn't do deep sleep anyway.

-----

2. `MeshService::handleFromRadio`

https://github.com/compumike/meshtastic-firmware/blob/95c5a94ad37cd78ac4cb8711ad97ef4870d34332/src/mesh/MeshService.cpp#L83-L84

It seems to me that we do NOT want to add `CLIENT_BASE` to the `isPreferredRebroadcaster` list, because we want the `CLIENT_BASE` to participate in `sendOurNodeInfo` as a normal `CLIENT` would.

-----

3. `NextHopRouter::stopRetransmission`

https://github.com/compumike/meshtastic-firmware/blob/95c5a94ad37cd78ac4cb8711ad97ef4870d34332/src/mesh/NextHopRouter.cpp#L172-L175

I don't have a good understanding of `NextHopRouter`, but I think we may need to add a special case here for `CLIENT_BASE` and favorited from/to (much like the new `FloodingRouter::roleAllowsCancelingDupe`).

-----

**Summary:** I'm leaning yes to changes required on 1 and 3, and no on 2. But will wait to hear your thoughts.

compumike:
@Xaositek I created a PR for docs here: https://github.com/meshtastic/meshtastic/pull/2014

GUVWAF:
@compumike 

> I'm leaning yes to changes required on 1 and 3, and no on 2

I agree with 2 and 3, but I think that 1 is also not required. This is only for a special case for when setting these roles to low power mode: #3251.

Regarding 3: the next-hop router is used from firmware version 2.6 on for direct messages and indeed we should not cancel a packet in the queue here for `CLIENT_BASE` also.

compumike:
@GUVWAF Implemented in a commit I just pushed, https://github.com/meshtastic/firmware/pull/7873/commits/c7de22f76cabe550b33e58cd9dceb54faae2b314

Anything else?

src/graphics/Screen.cpp
```diff
@@ -1605,13 +1605,15 @@ bool shouldWakeOnReceivedMessage()
...
+    - If role is not CLIENT / CLIENT_MUTE / CLIENT_HIDDEN / CLIENT_BASE
     - If the battery level is very low
     */
     if (moduleConfig.external_notification.enabled) {
         return false;
     }
-    if (!meshtastic_Config_DeviceConfig_Role_CLIENT && !meshtastic_Config_DeviceConfig_Role_CLIENT_MUTE) {
+    if (!IS_ONE_OF(config.device.role, meshtastic_Config_DeviceConfig_Role_CLIENT,
+                   meshtastic_Config_DeviceConfig_Role_CLIENT_MUTE, meshtastic_Config_DeviceConfig_Role_CLIENT_HIDDEN,
+                   meshtastic_Config_DeviceConfig_Role_CLIENT_BASE)) {
```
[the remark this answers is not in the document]
compumike: Hi Copilot, thank you for the suggestion, but I'm pretty sure that `if (!x) { ... }` is 100% equivalent to `if (x) { /* empty */ } else { ... }` 🤣

compumike:
@thebentern Looks like the Copilot feedback has been resolved 👍

compumike:
Rebased to `develop`.

compumike:
@GUVWAF I've just fixed what I believe is a memory leak in NextHopRouter::stopRetransmission. https://github.com/meshtastic/firmware/pull/7873/commits/0fc33c352a4b5541e7054064fee9dcddfcd24fb4

Before my patch, the retransmission lifecycle of an entry `pending` looks like this:

```
// in NextHopRouter::send:
startRetransmission(packetPool.allocCopy(*p));

// in startRetransmission:
auto rec = PendingPacket(p, numReTx);
pending[id] = rec;

...

// in stopRetransmission:
auto old = findPendingPacket(key);
auto p = old->packet;
if (some conditions) packetPool.release(p);
auto numErased = pending.erase(key);
```

I believe that the `packetPool.release(p)` should happen unconditionally whenever `pending.erase(key)` is called. So with my patch, the flow is:

```
// in NextHopRouter::send:
startRetransmission(packetPool.allocCopy(*p));

// in startRetransmission:
auto rec = PendingPacket(p, numReTx);
pending[id] = rec;

...

// in stopRetransmission:
auto old = findPendingPacket(key);
auto p = old->packet;
auto numErased = pending.erase(key);
packetPool.release(p);
```

This way, a newly allocated packet copy is always allocated at the same time an entry is set in `pending`, and it's always released at the same tie the entry is erased from `pending`.

Note that this packet copy within `pending` is used ONLY within NextHopRouter. When `NextHopRouter::doRetransmissions` runs, it always creates yet another copy when it calls: `FloodingRouter::send(packetPool.allocCopy(*p.packet));` or `NextHopRouter::send(packetPool.allocCopy(*p.packet));`.

Therefore, this copy of the packet that lives in `pending` should always be cleaned up with `pending`.

Let me know what you think.

GUVWAF:
@compumike Very nice find. 

It was actually fixed back when this logic was still in the ReliableRouter: https://github.com/meshtastic/firmware/pull/5328.
It must have been get lost when it was copied to the NextHopRouter.

PR #7875, opened 2025-09-06
Phone GPS display on Position Screen for BaseUI

HarukiToreda:
This is a PR to show when a phone shares GPS location with the node so you can reliably know what coordinate is being shared with the Mesh.

https://github.com/user-attachments/assets/2f8e0bfe-9867-4ea5-b195-88ccf95ae49f


Xaositek (approved):
Tested against T114 and Android, works as expected

PR #7882, opened 2025-09-06
Allow deepsleep in rak4630 and make it restart well when power comes back

viric:
When the power goes under 1.7V in a nrf52 it gets into brownout-reset, which will make the device hang (if running) and it will never restart when power comes back. One has to switch off the power supply completely first, to get a working power-on-reset when there is enough power. This is not acceptable for unattended solar nodes, at least for me. I want the device to restart when there is enough power again.

We have to arrange two things:
* Make the nrf52 deepSleep, which means in practice "SystemOff" in nrf52 terms.
* Program the LPCOMP (low power comparator) so it wakes the device again once VBAT > 11/16 VDD (for example). VDD has to be a reference, and it is a good reference like 3V3 in RAK main boards I think. It is in RAK19007.

With these changes, I can put the device power to 2.5V, then after a while it will go deep-sleep. Then I put the power down to 1.68V (usual brown-out reset condition). Then I rise the power to 4V and the device boots up again. I do this with a lab adjustable power supply.

Without this changes, the device hangs forever once reaching VBAT 1.68V.

This covers also https://github.com/meshtastic/firmware/pull/7285 and is related to https://github.com/meshtastic/firmware/issues/4378 

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631

### DIFF
--- a/src/platform/nrf52/main-nrf52.cpp
+++ b/src/platform/nrf52/main-nrf52.cpp
@@ -13,8 +13,8 @@
 #include "PowerMon.h"
 #include "error.h"
 #include "main.h"
-#include "power.h"
 #include "meshUtils.h"
+#include "power.h"
 
 #include <hal/nrf_lpcomp.h>
 
@@ -405,7 +405,8 @@ void cpuDeepSleep(uint32_t msecToWake)
         battery_adcEnable();
 
         nrf_lpcomp_task_trigger(NRF_LPCOMP, NRF_LPCOMP_TASK_START);
-        while(!nrf_lpcomp_event_check(NRF_LPCOMP, NRF_LPCOMP_EVENT_READY));
+        while (!nrf_lpcomp_event_check(NRF_LPCOMP, NRF_LPCOMP_EVENT_READY))
+            ;
 #endif
 
         auto ok = sd_power_system_off();


viric:
I tested the change over 2.7.5 and it worked, then I rebased over develop for the PR.

viric:
Ouch, apparently when going deep sleep (systemoff) the node starts as if with blank memory, as if flash was erased, and the node has to be reconfigured. I don't know what causes this.

viric:
The bluetooth pairing and the radio config is lost, but the nodedb is retained.
Maybe the problem is related to SystemOff retaining some RAM sections by default. I will try to figure out how to disable the RAM retention completely.

But isn't program start clearing the BSS and loading the DATA section from flash at startup?

viric:
Solved by disabling ram retention. Now the node resets perfectly well after normal voltage comes back from 2.5V or 1.68V.

viric:
What I'm afraid is what happens if the DeepSleep gets enabled by low battery by a level *higher* than the 7/8*VDD LPCOMP. Because the threshold to restart the chip will not be actually passed so the device may stay in system off? I will try to test that.

viric:
Hm what yesterday worked for me it doesn't work for me at all today. The device does not wake up from sleep. I will keep on trying to find the reason.

HarukiToreda:
Since you mentioned in your last comment that it’s still not working, I’ve changed the status to Draft until you’re ready to have this tested again. Could you also let me know whether you’re using iOS or Android? That will help narrow down the issue.

viric:
@HarukiToreda  correct! I asked Nordic about the behavior I observe now, for which I don't have explanation. I cannot understand why one day all worked for me and the next day is all fuzziness.

In short, if I remove the RAM part of the code, it goes to sleep forever, never waking up. If I keep the RAM part, it reboots immediately at system off.

I will wait Nordic's answer.

I use Android but the behavior is observable regardless of the bluetooth connection. It doesn't have anything to do with the client I think.

HarukiToreda:
sounds good, keep us updated on how it goes, I ask about Android specifically since the app currently is having issues where when a node disconnects it will not auto reconnect as it should, this is a problem that has been reported and is in the works of being fixed so hope it doesn't give you a factor that may not be related to this.

viric:
@HarukiToreda Solved! I probably tested something wrong first. Now it works perfect.
It stands well in deep sleep even at 1.68V, 1.58V, ... and it reboots when the battery voltage is over 3.7V.

Now we probably need some discussion on what variants it is enabled. I only have a RAK19007 + RAK4631 and it works there, where the VBAT voltage divider is connected to the nrf52 AIN3. But I don't know if that's the same for all "RAK4630" preprocessor macro.

Any volunteer to test this with a battery?

src/platform/nrf52/main-nrf52.cpp
```diff
@@ -389,6 +391,29 @@ void cpuDeepSleep(uint32_t msecToWake)
...
+         * We have the device going deep sleep under 3.1V, which is AIN3 = 1.86V
+         * So we can wake up when VBAT>=VDD is restored to 3.3V, where AIN3 = 1.98V
+         * 1.98/3.3 = 6/10, but that's close to the VBAT divider, so we
+         * pick 6/8VDD, which means VBAT=4.1V.
+         * Reference:
+         * VDD=3.3V AIN3=5/8*VDD=2.06V VBAT=1.66*AIN3=3.41V 
+         * VDD=3.3V AIN3=11/16*VDD=2.26V VBAT=1.66*AIN3=3.76V
+         * VDD=3.3V AIN3=6/8*VDD=2.47V VBAT=1.66*AIN3=4.1V
+         */
+        c.reference = NRF_LPCOMP_REF_SUPPLY_11_16;
```
viric: This 11/16 is a good match for the voltage divider in the RAK19007. This could be made a variant.h parameter or we could make code that tries to calculate a good approximation.

viric:
> Does this impact 4631 at all?

Sure it impacts. The 4631 packages a 4630 inside.

My device is a RAK19007 with a 4631 inside, that has 4630 inside, that has
a nrf52840 inside.

HarukiToreda:
I should be able to test this tonight after work. Had to find my old rak board from my solar setup.

benjamincarmel:
As a RAK4631 user this would be beneficial to me

viric:
@HarukiToreda  I'm testing with a battery now but it takes long to discharge.

HarukiToreda:
> @HarukiToreda I'm testing with a battery now but it takes long to discharge.

I cannot replicate the brownout issue that this PR is supposed to resolve. As seen on this video, I am supplying gradually lower voltage down to 1.5V and even though I do so, when voltage is re-applied it still powers back up no problem, this is with the Stock 2.7.7 firmware. How are you re-creating the original issue? 

https://github.com/user-attachments/assets/b93b84a7-b581-489e-a1fc-ba812cfc9437

viric:
If you go too low it will be perceived as a poweroff I think.
Some people experiencing the hang:
https://devzone.nordicsemi.com/f/nordic-q-a/37058/brownout-watchdog-reset-on-voltage-drop

I could easily reproduce the issue with mine, at around 1.68V, 1,65V. You should try a slow voltage going down to 1,68V, then try to go up. Another down to 1,65V, then go up. If you go much under, it can perceive a poweroff and it will power-on reset.

Btw I'm testing this change with battery and it recovered fine from sleep mode, which started at ~3.1V. It recovered because I plugged usb; this is not a solar node.

HarukiToreda:
Sorry for the lapse in update, I was able to replicate the issue with one of my batteries that finally ran out of power after days of waiting and your fix did allow it to reboot properly again. Without it it went into a state that even the reset button would not allow reboot, had to completely disconnect the power to get it to come back.

goesta:
Great work ! The T-Echo lite (variants/nrf52840/t-echo-lite) has this issue too. Would be nice to have it included in the PR. The divider should be 100k / 100k (https://github.com/Xinyuan-LilyGO/T-Echo-Lite/blob/main/project/T-Echo-Lite_V1.0/T-Echo-Lite_V1.0.pdf)

Quency-D:
@viric  I have a question: does this commit have any issues if lpcomp is not added? Without lpcomp, it will remain asleep and cannot be woken up, as mentioned in commit #7285.

viric:
On Mon, Nov 24, 2025 at 03:35:02AM -0800, Quency-D wrote:
> Quency-D left a comment (meshtastic/firmware#7882)
> 
> @viric  I have a question: does this commit have any issues if lpcomp is not added? Without lpcomp, it will remain asleep and cannot be woken up, as mentioned in commit #7285.

Right, without the lpcomp it will not wake up again.

But notice that the PR #7285 this was already the desired behavior there
to avoid draining out the battery beyond safe voltages.

If we would not trigger the EVENT_LOW_BATTERY as before the change then
the device would keep on running until, draining the battery, a likely
hang at ~1.68V battery voltage.

Quency-D:
@viric I understand, thank you for your patient explanation.

ndoo:
@viric did you consider rolling this out to gat562 as well? It seems to have the same basic pinout and voltage divider (seeing as rak4631 variant cross-flashed to it exhibits accurate voltage readings).

Am I also right in thinking that this would also be applicable to all nRF variants? In that case, would using greater than/less than etc. comparisons in preprocessor directives help to automatically set the best value possible for all nRF variants?

Glad to test these changes on some of the other nRF-based devices I have, but wanted to get a feel of whether it’s worth pursuing/if you would like to give it a try yourself. Thanks!

ndoo:
Also... this is a pretty huge feature that did not make it to the release notes 😅

viric:
I only own one kind of device, and a friend another. No clue about gat562.

I agree that we can build some logic that figures out the best divider but
that can be any next step. We don't need comparisons in preprocessor -
they are cheap to run in runtime. We only need the right board numbers,
and then the code can figure out the best divider.

For what I remember, it was quite messy how the voltage measurement was
specified per board (pin, divider, etc.) so we would need to unify that a
bit to be able to build generic logic.

For example, for RAK4631, the battery pin is specified as to the rak4630
package (A0), not the not nrf52840 package (AIN3). Then, the
ADC_MULTIPLIER 1.73 is not the voltage divider factor in the circuit,
2.5/1.5 = 1.66. And this for every board I guess.

ndoo:
Hm, if we just reuse the macros already defined for the battery sense, we
shouldn’t have to worry about what the pin name is, right?

viric:
On Mon, Nov 24, 2025 at 09:05:42PM -0800, Andrew Yong wrote:
> ndoo left a comment (meshtastic/firmware#7882)
> 
> Hm, if we just reuse the macros already defined for the battery sense, we
> shouldn’t have to worry about what the pin name is, right?

The LPCOMP requires knowing the analog input with its own enum of nrf52
analog inputs, not from the input gpio number. We would need a conversion
from one to the other. If I find inspiration I can try to write that.

PR #7887, opened 2025-09-07
Show GPS Date properly in drawCommonHeader

Xaositek:
- Show GPS Date properly in drawCommonHeader
- Remove bad and redundant code


HarukiToreda (approved):
Tested on both Heltec V3 and T114, no regressions and code seems cleaner.

Issue #7888, opened 2025-09-07
[Bug]: 2.7.8 date 1943-11-30


Pedrom-pt:
### Category

Other

### Hardware

Seeed Card Tracker T1000-E

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.8

### Description

Hi
2.7.8 reports date as 1943-11-30

It's ok at 2.7.7. Updated by full wipe

The user @fischkind, discord have the same problem with t-beam
Android 2.6.35

![Image](https://github.com/user-attachments/assets/0cf8a41f-55fc-4872-80ff-19d9b3ffdc75)

https://github.com/user-attachments/assets/f51c2063-bcd6-4b7a-b64b-797b80e93314

### Relevant log output

```Shell

```

fifieldt:
Thanks for the report. Are you familiar with providing log files? (flasher.meshtastic.org --> Open Serial Monitor)

fifieldt:
I believe this may be related: https://github.com/meshtastic/gh-action-firmware/issues/10

PR #7894, opened 2025-09-07
Add BUILD_EPOCH to latest setup step.

fifieldt:
Previously this was in setup-base. However, setup-base is no longer used by the setup job.

Fixes https://github.com/meshtastic/gh-action-firmware/issues/10 Fixes https://github.com/meshtastic/firmware/issues/7888


fifieldt:
I am not 100% sure this is the right fix, but it should not be harmful.

Xaositek (approved):

PR #7862, opened 2025-09-04
Backmerge from Master into develop

fifieldt:

PR #7901, opened 2025-09-08
Update protobufs

fifieldt:


fifieldt:
This may not be the correct approach.

fifieldt:
Let's try it and see if we can get the backmerge working.
https://github.com/meshtastic/firmware/pull/7862

fifieldt:
Did not work, still showing conflict

fifieldt:
Didn't work since we went to the latest protobufs, not the one master had. Updating master now.

PR #7908, opened 2025-09-08
Exclude some modules to fix Sensecap Indicator TFT size

fifieldt:
Firmware no longer fits in Seeed Sensecap Indicator's flash.

Let's try excluding some modules until it does.

## 🤝 Attestations

  - [ ] Seeed Sensecap Indicator


mverch67:
This is not necessary due to https://github.com/meshtastic/firmware/pull/7860 but it's ok :)

PR #7909, opened 2025-09-08
Allow label enforcement job to run on self-hosted runners

fifieldt:
Previously, this check would only run on github-provided runners.


vidplace7 (approved):
This feels like a "try it and see" situation (revert if necessary).
I'm not sure the self-hosted runner images have the necessary nodejs dependencies to run `actions/github-script` currently.

fifieldt:
I confirmed using https://github.com/meshtastic/gh-runners/pull/9 that it does indeed work!

Issue #7911, opened 2025-09-08
[Bug]: Position broadcast interval not honored for TAK_Tracker role


RichieFrame:
### Category

Other

### Hardware

T-Echo

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.8.a0c0388

### Description

Role Tak Tracker.

Position interval is configured at 60s, Smart Position is configured off, and confirmed as disabled in settings of both Android app and web client.

When node is stationary, position is broadcast approximately every 37 to 41 seconds, when node is moving on foot approx 2m/s, position is broadcast approximately every 22 seconds, this implies that Smart Position is being used, note 41s interval in the provided log when stationary

Interestingly the behavior is near identical to another node on 2.5.22 firmware with Smart Position enabled, which is also set for a 60s broadcast interval but is broadcasting approx every 40s when stationary, and at the correct 15s interval when moving beyond the min distance when moving on foot; note, when stationary the position drift (1-2m if any) is well under the configured min distance (7m)

### Relevant log output

```Shell
DEBUG | 05:42:41 8685 [GPS] Publish pos@68be6cd0:2, hasVal=1, Sats=23, GPSlock=1
DEBUG | 05:42:41 8685 [GPS] New GPS pos@68be6cd0:3 lat=xxxx lon=xxxx alt=xxx pdop=0.84 track=342.75 speed=0.00 sats=23
DEBUG | 05:42:41 8685 [GPS] onGPSChanged() pos@68be6cd0 time=1757310161 lat=xxxx lon=-xxxx alt=xxx
INFO  | 05:42:41 8685 [GPS] updatePosition LOCAL pos@68be6cd0

INFO  | 05:42:52 8696 [GPS] updatePosition LOCAL pos@68be6cdb
INFO  | 05:43:03 8707 [GPS] updatePosition LOCAL pos@68be6ce6

INFO  | 05:43:10 8714 [Position] Send pos@68be6ce6:6 to mesh (wantReplies=0)
DEBUG | 05:43:10 8714 [Position] cancelSending id=0xbd384fa2, removed=0
DEBUG | 05:43:10 8714 [Position] Send location with precision 32
INFO  | 05:43:10 8714 [Position] Position packet: time=1757310190 lat=xxxx lon=xxxx
[...]

PR #7932, opened 2025-09-09
handleNewPosition should not do smart things if smart position is dis…

fifieldt:
…abled

PositionModule::handleNewPosition() is called from GPS::publishUpdate, and its primary function is to send positions if they have exeecded the smart position travel threshold.

However, we did all those checks and sent positions without checking whether the user had actually enabled smart position.

Fixes https://github.com/meshtastic/firmware/issues/7911

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/modules/PositionModule.cpp
+++ b/src/modules/PositionModule.cpp
@@ -492,7 +492,7 @@ void PositionModule::handleNewPosition()
     meshtastic_NodeInfoLite *node = nodeDB->getMeshNode(nodeDB->getNodeNum());
     const meshtastic_NodeInfoLite *node2 = service->refreshLocalMeshNode(); // should guarantee there is now a position
     // We limit our GPS broadcasts to a max rate
-    if (nodeDB->hasValidPosition(node2)) {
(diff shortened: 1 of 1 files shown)


fifieldt:
I'm not 100% confident in this, would appreciate review or testing :)

garthvh:
Pretty sure it was partly intentional to not waste airtime on redundant positions, but it should not be sending faster than the configured interval.

jp-bennett:
I believe we intentionally deprecated the smart position setting, to make position always smart.

thebentern:
> I believe we intentionally deprecated the smart position setting, to make position always smart.

This is correct. I do wonder if we do need carve out for TRACKER or TAK_TRACKER roles?

fifieldt:
Ah! We did too! So there's probably some treatment needed in the app based on firmware version to not have a disable toggle? At least in #7911 what the user is seeing is probably smart position settings they can't even see on the page since the toggle is off.

PR #7937, opened 2025-09-09
Add LOG_HEAP log type, and more heap debug messages

jp-bennett:
Add the LOG_HEAP log call, and Add some debugging around memory allocation and freeing

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -193,8 +193,12 @@ void MeshService::handleToRadio(meshtastic_MeshPacket &p)
                                                  // (so we update our nodedb for the local node)
 
     // Send the packet into the mesh
+    auto heapBefore = memGet.getFreeHeap();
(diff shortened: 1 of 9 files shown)


src/mesh/MeshService.cpp
```diff
@@ -193,8 +193,12 @@ void MeshService::handleToRadio(meshtastic_MeshPacket &p)
                                                  // (so we update our nodedb for the local node)
 
     // Send the packet into the mesh
+    auto heapBefore = memGet.getFreeHeap();
```
mverch67, line 196: I think this should go into a #ifdef DEBUG_HEAP as these look like costly operations. Not sure if the compiler will optimize these away when not used (they are not declared const).

mverch67: Also saw crashes on esp32 when I used these in MUI periodically. So I switched the heap icon off by default.

Issue #7944, opened 2025-09-09
[Feature Request]: Change external notification defaults


RCGV1:
### Platform

Cross-Platform

### Description

Only alert for bells by default and not on regular messages, change the alert duration to only 5 seconds instead of the excruciatingly long 60s.

RCGV1:
Or maybe just disable it for everything but dms

vidplace7:
https://github.com/meshtastic/protobufs/pull/727

Protobuf exists, needs firmware implementation

danditomaso:
My vote is disable entirely except for DMs

b8b8:
5 seconds, DMs

OnyxClawe:
Started implementing this, i'm limited in my skills but i adjusted the default nag time: [Pull request here](https://github.com/meshtastic/firmware/pull/7946)

nexpspace:
> [meshtastic/protobufs#727](https://github.com/meshtastic/protobufs/pull/727)
> 
> Protobuf exists, needs firmware implementation

I struggle to fully understand the difference between buzzer_mode and the external notification settings.
It would make more sense to me, if there is an additional button in external notification, where you can check "Alert DMs" (or a toggle between all messages and DMs).
So you could separately switch off messages, bells and DMs.

---

> Buzzer is enabled only for direct messages and alerts
> https://github.com/meshtastic/protobufs/pull/727

I am not sure if "alerts" here mean the alarm bell? There seem to be different names for the bell in the source code, sometimes it is called bell (`containsBell`), sometimes it is alert (`isAlert`)? I implemented this as if alert means bell (though in my tests this fells on the bell character, not sure yet why, not changed any code there. I tested only with the iOS bell character/alert button though).

---

I implemented it this way:
If `buzzer_mode == 4` (`DIRECT_MSG_ONLY`) and `alert_messages_buzzer` is on - then only buzz on DMs.
Is this the way you wanted it to be? Then I could create a pull request for it.

Btw: The only menu I found to change the buzzer_mode is in baseUI, neither in webclient nor in iOS I could find the option. I tested my changes with meshtastic-cli (`meshtastic --set device.buzzer_mode 4`).

PR #7946, opened 2025-09-09
Issue: #7944 External notification module: Adjusted default nag timeout to 5s (from 60s)

OnyxClawe:
As mentioned in [this issue](https://github.com/meshtastic/firmware/issues/7944). I've adjusted the default external notification nag timeout to be 5s rather than 60s.
Due to my lack of knowledge and skills i was not able to implement only notifying when receiving a DM. Help with this is greatly appreciated!

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/mesh/Default.h
+++ b/src/mesh/Default.h
@@ -27,7 +27,7 @@
 #ifdef USERPREFS_RINGTONE_NAG_SECS
 #define default_ringtone_nag_secs USERPREFS_RINGTONE_NAG_SECS
 #else
-#define default_ringtone_nag_secs 60
+#define default_ringtone_nag_secs 5
 #endif
 
 #define default_mqtt_address "mqtt.meshtastic.org"


fifieldt (approved):

src/mesh/Default.h
```diff
@@ -27,7 +27,7 @@
 #ifdef USERPREFS_RINGTONE_NAG_SECS
 #define default_ringtone_nag_secs USERPREFS_RINGTONE_NAG_SECS
 #else
-#define default_ringtone_nag_secs 60
+#define default_ringtone_nag_secs 5
```
thebentern: Can we do 15? 60 is definitely too long, but 5 seconds seems abruptly short

fifieldt:
I've changed the default nag to 15s rather than 5s, since we don't yet have consensus for 5s and 15s is still much better than 60 :) Let's merge this and keep chatting about the value.

PR #7955, opened 2025-09-10
Put guards in place around debug heap operations

thebentern:


fifieldt (approved):

mverch67 (approved):
Now the code gets a bit difficult to read :(

Why not use macros :)

E.g. turn

```
#if defined(DEBUG_HEAP)
        auto heapBefore = memGet.getFreeHeap();
#endif
        auto a = packetPool.allocCopy(*p);
#if defined(DEBUG_HEAP)
        auto heapAfter = memGet.getFreeHeap();
        LOG_HEAP("Alloc in MeshService::sendToMesh() pointer 0x%x, size: %u, free: %u", a, heapBefore - heapAfter, heapAfter);
#endif
```

into:

```
        DEBUG_HEAP_BEFORE;
        auto a = packetPool.allocCopy(*p);
        DEBUG_HEAP_AFTER("MeshService::sendToMesh()");
```

jp-bennett:
Should work as is, but I like Manuel's idea to use the macro here. The macro will need to pointer itself passed as an arg as well, but should be pretty straightforward.

PR #7957, opened 2025-09-11
Mute chats

### LINKED ISSUE #7943: [Feature Request]: Mute Notifications for Specific Chats
### Platform
Cross-Platform

### Description

It would be very useful to have the ability to mute notifications for a specific chat. For example, suppress notifications from a public channel so that new messages do not interrupt the user, while still allowing notifications for private messages.

Use case / Benefit:

Avoid distractions from high-traffic public channels.

Maintain awareness of important private messages.

Improve usability and user control over notifications.

Proposed behavior:

Option to mute a specific chat or channel.

Notifications remain active for other chats.

Ability to toggle mute on/off easily.

(Original feature request: https://github.com/meshtastic/device-ui/issues/189)

ford-jones:
Resolves #7943

Introduces mechanisms to _mute_ message notifications received over specified channels or from certain nodes. 
When a message is received from a muted node or channel, the device's buzzer (if present) should **not** be activated. Drawing of any alert UI notification to the device's screen should also be **blocked**.

The exception to this is bell alerts (when activated), which should still sound the buzzer and display a notification.

It should be noted that this behaviour is distinctly different to _ignoring_ a node - an action which blocks the message entirely. Muted messages can still be viewed in the messages menu of the device or client application, they are just delivered discreetly.

- [x] Buzzer **always** sounds when no channels or other nodes are muted
- [x] OLED popup **always** displays when no channels or other nodes are muted
- [x] When a specific channel is muted, messages received over that channel from any node **don't** invoke the buzzer 
- [x] When a specific channel is muted, messages received over that channel from any node **don't** display a popup 
- [x] When specific nodes are muted, **they do not** invoke the buzzer **while others still do** when sent over a non-muted channel
- [x] When specific nodes are muted, **they do not** display the popup **while others still do** when sent over a non-muted channel

Protobuf changes: https://github.com/meshtastic/protobufs/pull/769
Python CLI changes: https://github.com/meshtastic/python/pull/822
Docs: https://github.com/meshtastic/meshtastic/pull/2028

## 🤝 Attestations

  - [x] Heltec (Lora32) V3


ford-jones:
There are two things left to be done here:
- Block message notification on InkHUD UI
- Block message notification on Meshtastic UI

I'm currently lacking the hardware to do this, I have some on the way but it's unlikely it will be here any time soon. If this PR is approved I will open a new issue for these TODOS which someone else can take up in the meantime - otherwise, I will do it when I have the hardware.

src/mesh/generated/meshtastic/admin.pb.h
```diff
@@ -247,6 +247,10 @@ typedef struct _meshtastic_AdminMessage {
         uint32_t set_ignored_node;
```
fifieldt, line 247: the edits to automatically generated protobuf files should be removed before we can merge this one.

ford-jones: Ah cheers I missed this, will run a regen now :)

garthvh:
Channels is fine, nodes is kind of a duplication of the existing ignore functionality.

ford-jones:
Apologies, didn't mean to close/reopen.

@garthvh Happy to make changes, but I think it's worth considering that unlike ignore - the message is still delivered and can be read.
> It should be noted that this behaviour is distinctly different to ignoring a node - an action which blocks the message entirely. Muted messages can still be viewed in the messages menu of the device or client application, they are just delivered discreetly.

ford-jones:
After a discussion with @garthvh this morning; we've determined the best course of action here is to decouple muting nodes and muting channels into two seperate PR's.

This is to relieve pressure on the size constraint of the user proto - which currently hangs a big question mark over the ability to merge this.

src/modules/ExternalNotificationModule.cpp
```diff
@@ -453,12 +453,15 @@ ProcessMessage ExternalNotificationModule::handleReceived(const meshtastic_MeshP
...
             bool containsBell = false;
-            for (int i = 0; i < p.payload.size; i++) {
+            for (size_t i = 0; i < p.payload.size; i++) {
                 if (p.payload.bytes[i] == ASCII_BELL) {
                     containsBell = true;
                 }
             }
 
+            meshtastic_NodeInfoLite *sender = nodeDB->getMeshNode(mp.from);
+            meshtastic_Channel ch = channels.getByIndex(sender->channel ? sender->channel : channels.getPrimaryIndex());
```
thebentern: I'm not sure this logic is correct. Don't we want the meshtastic_MeshPacket -> p.channel?

ford-jones: Ah yes I think you're right,
I'm moments away from hopping into bed but I'll do this once I'm up in the morning :)

Thanks for pointing this out!

src/mesh/generated/meshtastic/admin.pb.h
fifieldt: New protobuf is merged

thebentern (approved):

PR #7965, opened 2025-09-12
Fix memory leak in `NRF52Bluetooth`: allocate `BluetoothStatus` on stack, not heap

compumike:
This is the same as https://github.com/meshtastic/firmware/pull/7964 but for `NRF52Bluetooth`.

Unfortunately I do not have any NRF52-based devices to test with at this moment, but since `BluetoothStatus` is shared between platforms, I again believe it should be completely safe.

### DIFF
--- a/src/platform/nrf52/NRF52Bluetooth.cpp
+++ b/src/platform/nrf52/NRF52Bluetooth.cpp
@@ -59,7 +59,8 @@ void onConnect(uint16_t conn_handle)
     LOG_INFO("BLE Connected to %s", central_name);
 
     // Notify UI (or any other interested firmware components)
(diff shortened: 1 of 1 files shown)


thebentern:
@compumike I'll get this tested on some of my devices. If you're on our discord, reach out with your handle and we'll get you contributor flaired and we can send you some NRF hardware devices to mess with if you're interested.

compumike:
> @compumike I'll get this tested on some of my devices. If you're on our discord, reach out with your handle and we'll get you contributor flaired and we can send you some NRF hardware devices to mess with if you're interested.

Thanks, I'm @compumike on the discord!

PR #7982, opened 2025-09-14
Test develop --> master

fifieldt:
we're 100 commits diverged ... how possible is this?


fifieldt:
so, to fix the merge conflict for variants/esp32s3/tlora-pager/rfswitch.h we need to open two different branches and work out which one is the good one ... both have commits yesterday & it's not immediately clear to me which one is the good one.

thebentern:
We need to slow roll merging PRs into develop for a bit. I think once we get a somewhat stablized alpha off of master, we can start cherry picking some of the bigger features upstream and eventually do a bigger pull to sync things up

fifieldt:
@thebentern , when you get a moment, could you write up a bit of a paragraph about your vision for the dual branch strategy to share with everyone :)

I originally thought part of the point of having separate 'stable' master was so we could keep a branch open for development while we work on releases. Result: confused by comments we should slow development in the development branch :)

garthvh:
The issues are in master already so every thing needs to slow down right now, stuff needs to be tested better before being merged into develop too.  Both branches need to have more testing before code is merged in.

PR #7986, opened 2025-09-14
T-Lora Pager: Fully fix rotary encoder and speaker fuzzing/popping

WillyJL:
- Rotary encoder now uses an interrupt + freertos thread + queue mechanism:
  - It sounds overengineered but hear me out
  - Reading the gpio pins in the interrupt proved unreliable and missed some inputs randomly all the time
  - The LilyGoLib factory example uses a freertos thread to poll the gpio every 2ms which is inefficient
  - This instead uses an interrupt to detect that something happened, which wakes the freertos idle background thread to poll the gpio immediately just once so it wont get missed, and dispatch the inputs to a queue
  - The usual "OSThread" mechanism polls the queue only, not gpio, every 10ms as before to send the events to GUI (sending events from the freertos thread causes crashes) (sidenote, the "OSThread" name is a bit deceiving, I spent more than I'd like to admit getting mad at why waiting for queue items in a "thread" froze the whole system, just to realize "OSThread" is actually a protothread ran in sync and yielding needs to be done manually and all system components including input, keyboard, and gui are done in this manner, and its not compatible with freertos synchronization primitives)
  - Overall, now rotary encoder is fully fixed, it responds very well to scrolling and never skips a tick (except scrolling fast, that might detect 2 ticks when 3 were scrolled for example, but it will not skip ticks out of the blue)
- **New `InputPollable` system in `InputBroker` that allows for polling inputs right after an interrupt so no events are missed**, which standardizes the process explained above so other input sources can use it (eg TCA8418 in #7989, the code for this is the same, feel free to merge either PR first)
  - Basic idea is:
  - `InputEvent` sources inherit from `InputPollable` (like they did with `Observable<const InputEvent *>`
  - Define a `pollOnce()` method on the input source class, this will be run from userspace so it can do whatever
  - Inside `pollOnce()` add events with `inputBroker->queueInputEvent(&event)` (instead of `this->notifyObservers(&event)` with previous system)
  - During setup attach an interrupt for the device and in the interrupt handler call `inputBroker->pollSoonRequestFromIsr(this)` (`this` being the input event source class which inherits from `InputPollable`)
  - With this setup, when the interrupt is hit the class will be polled using `pollOnce()` by `InputBroker` as soon as the interrupt handler exits, bypassing cooperative multitasking delay of OSThread, providing much better responsiveness and avoiding missed inputs, and avoiding unnecessary periodic polling
  - After the input events are collected with `inputBroker->queueInputEvent(&event)` in `pollOnce()`, they will be processed by GUI and other modules on the main thread when possible via `inputBroker->processInputEventQueue()` which is shared among all input event sources that use this paradigm
  - Note that this relies on FreeRTOS, so it will not work on platforms that don't use FreeRTOS
- Speaker fuzzing and random popping fixed by cutting power to the amplifier when no sound is being played
- I'm not sure how to measure battery consumption in idle / sleep, would be nice if someone could check that, I'm not entirely sure if any changes here could affect that and if it would need some changes for proper sleep

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - [x] LilyGo T-Lora Pager


src/AudioThread.h
```diff
@@ -20,6 +25,9 @@ class AudioThread : public concurrency::OSThread
 
     void beginRttl(const void *data, uint32_t len)
     {
+#ifdef USE_XL9555
```
mverch67: This should read `#ifdef T_LORA_PAGER` as `USE_XL9555` is supposed to be used by all devices that have a TCA9555 io expander. Same below (3x).

mverch67:
> I'm not sure how to measure battery consumption in idle / sleep, would be nice if someone could check that, I'm not entirely sure if any changes here could affect that and if it would need some changes for proper sleep

The measured power consumption during light sleep was 36mA. Switching off the peripherals before light sleep didn't save any power. In deep sleep it's 65μA.

QuietImCoding:
eargerly awaiting this being merged, rotary encoder as it stands is really broken

QuietImCoding:
Ah, need a tag for "hardware-support" to pass pull requirements

WillyJL:
possibly on hold, would like feedback on this https://github.com/meshtastic/firmware/pull/7989#issuecomment-3300813448 as it would provide a standard interface for other input subsystems to do what this pr does to get more reliable inputs without duplicating code and having multiple background threads

QuietImCoding:
Interesting thread, definitely like the idea of centralizing input events in a more streamlined way, should help deal with a greater variety of form factors going forward.

WillyJL:
i reworked it based on the system i made for #7989 , seems to work great and means the 2 systems will not duplicate necessary resources for thread/queue etc...

they have 1 commit in common implementing the underlying system, thats why it may look similar. merging either PR first will be fine, then can just merge develop into the other PR.

WillyJL:
not sure if i need to do something to fix CI on this and keyboard improvements PR

EDIT: ah i see its a build fail, i hadnt scrolled up enough

giannoug:
I'm so glad I came across this PR! I had actually started working on these fixes myself. I managed to get the rotary encoder working by refactoring RotaryEncoderImpl to behave like [LILYGO’s examples](https://github.com/Xinyuan-LilyGO/LilyGoLib/tree/master/src/rotary). The encoder worked fine, but navigation felt a bit laggy (yours is much smoother!). I wasn’t sure if that was a Meshtastic, display, or hardware limitation, so I stopped there. The popping noise though...

This PR works perfectly on my hardware and fixes both of the issues I was experiencing! Navigation is smooth, and there’s no popping.

I’m also interested in adding a Greek font for displaying messages and a Greek keyboard layout. Before I spend time on it and possibly hit a dead end, should I start working on it now or wait for the -tft release (or something else)? Maybe I could open a new issue so we can discuss it there?

WillyJL:
Regarding Greek font and layout, probably best to open an issue for discussion

hylaride:
FWIW, I merged this into my own repo that was synced off of today's develop branch and the scroll wheel is SO MUCH BETTER.

WillyJL:
> FWIW, I merged this into my own repo that was synced off of today's develop branch and the scroll wheel is SO MUCH BETTER.

i have my 4 currently open PRs for tlora pager and cardputer adv and the multi message storage branch from meshtastic repo merged in my fork's develop branch as ive been dogfooding them on my device, if you wanna try the keyboard improvements and rtc too

mverch67:
fixes #8258

jeremyd:
I'm having some kind of timing issues with heltecv3 and rotary encoder running the latest firmware (master branch), but it's that, it goes the wrong direction 30-40% of the time.    Any pointers on how this was tested on heltecv3?  Pins used?  I read somewhere I need to make sure the pins support interrupt.. Tried different pins and different encoder parts, and even setting a couple different deBounceMs delays in the code to no avail.  It's the GPIO encoder type.

@WillyJL

WillyJL:
@jeremyd this PR is specific to T-Lora Pager, afaik no other devices currently use the same FSM rotary encoder. also, heltecv3 has no rotary encoder, so there is nothing to test. if you added your own rotary encoder to your specific device, thats not something that someone else can test.

jeremyd:
> @jeremyd this PR is specific to T-Lora Pager, afaik no other devices currently use the same FSM rotary encoder. also, heltecv3 has no rotary encoder, so there is nothing to test. if you added your own rotary encoder to your specific device, thats not something that someone else can test.

Oh ok, sorry, I was reading it and it had the " I have tested that my proposed changes do not cause any obvious regressions on the following devices: heltecv3".  No worries, yeah I think the rotary encoder I'm trying to use, does not act the same I suppose, as what the firmware is expecting.

Issue #7985, opened 2025-09-14
[Board]: M5stack Cardputer ADV


Szetya:
### SOC

ESP32

### Lora IC

SX1262

### Product Link

https://docs.m5stack.com/en/core/Cardputer-Adv

### Description

@mverch67  Maybe a Cardputer ADV PR? ;)
You're the master. :)
At the moment, I don't know whether lovyan or meshtastic/st7789 driver would be the better choice. I've tried the second one for now.
GPS ✅️
LoRa ✅️
Keyboard ✅️

I couldn't get the sound to work. I'm happy to share a few files.
[m5stack_cardputer_adv.zip](https://github.com/user-attachments/files/22321903/m5stack_cardputer_adv.zip)

<img width="685" height="1433" alt="Image" src="https://github.com/user-attachments/assets/22876625-c012-430d-8243-2211a196dc29" />

Szetya:
Neopixel will probably not be usable. The LED PWR Enable and TFT backlight PIN are shared.

p0ns:
@Szetya can you try compiling the code without `#define SX126X_DIO3_TCXO_VOLTAGE 1.8` in `variant.h` and see it it works with the Cardputer Adv and Cap Lora module? Afaik it doesn't have a TCXO.

I have the Cardputer Adv but no Lora module so I built my own using a Ra-01SH, and I had to remove that for it to work, but I'm not sure if that goes with the Cap

Szetya:
> [@Szetya](https://github.com/Szetya) can you try compiling the code without `#define SX126X_DIO3_TCXO_VOLTAGE 1.8` in `variant.h` and see it it works with the Cap? Afaik it doesn't have a TCXO.
> 
> I have the Cardputer Adv but no Lora module so I built my own using a Ra-01SH, and I had to remove that for it to work, but I'm not sure if that goes with the Cap

I'll be home after (Z)13:30. I'll check the code with the radio you mentioned to see if it works. Maybe there will be an sx1278 at the bottom of the drawer. 😉

Szetya:
Yes. It seems to work with SX1278, which means it should also work with RA01SH.
Sending and receiving are fine, it responds to traceroute.
I made some changes to the variant.h file. The USE_RF95 macro definition was added for testing purposes.
A few relevant lines from the log:
```ini
DEBUG | ??:??:?? 2 RF95Interface(cs=5, irq=4, rst=3, busy=6)
INFO  | ??:??:?? 2 Start meshradio init
[...]

PR #7989, opened 2025-09-15
TCA8418, T-Lora Pager, T-Deck Pro: Keyboard improvements

WillyJL:
- Refactored TCA8418 and derived keyboards:
  - `KeyState` enum and field moved to `TCA8418Keyboard` class (Nokia multitap keyboard) so classes derived from `TCA8418KeyboardBase` can handle key hold state their own way
  - Pass `key` parameter to `release()` allowing T-Lora Pager and T-Deck Pro to track multiple held keys at once
  - Standardize `trigger()` implementation into `TCA8418KeyboardBase`, and fix it to process all available FIFO key events correctly
  - Remove unused/unrelated code from T-Lora Pager and T-Deck Pro keyboards (eg multitap)
  - Make event maps const so they do not take up RAM
- N-key rollover: Due to above changes, **T-Lora Pager and T-Deck Pro can now handle multiple (probably all) keys being held at once, and register them correctly in order of key release**, helps greatly with typing fast registering inputs correctly (starting pressing the next key before releasing the previous key doesn't invalidate the previous key anymore)
- **T-Lora Pager and T-Deck Pro Keyboard modifiers persist while held**, even if a non-modifier key is pressed, like you would expect from a keyboard (eg can keep holding shift to type multiple keys in caps)
  - However can still press a single modifier on its own to make it apply to the next keystroke (eg for one handed use, press modifier then press key, can click another modifier to cancel the active modifier)
- **TCA8418 now uses `KB_INT` pin to trigger on interrupts, thus no more 300ms typing delay** on T-Lora Pager and T-Deck Pro keyboards
  - I plan to bring this concept to other applicable i2c keyboards soon when this PR is merged
- **New `InputPollable` system in `InputBroker` that allows for polling inputs right after an interrupt so no events are missed**, this is what TCA8418 now uses for interrupts
  - Basic idea is:
  - `InputEvent` sources inherit from `InputPollable` (like they did with `Observable<const InputEvent *>`
  - Define a `pollOnce()` method on the input source class, this will be run from userspace so it can do whatever
  - Inside `pollOnce()` add events with `inputBroker->queueInputEvent(&event)` (instead of `this->notifyObservers(&event)` with previous system)
  - During setup attach an interrupt for the device and in the interrupt handler call `inputBroker->pollSoonRequestFromIsr(this)` (`this` being the input event source class which inherits from `InputPollable`)
  - With this setup, when the interrupt is hit the class will be polled using `pollOnce()` by `InputBroker` as soon as the interrupt handler exits, bypassing cooperative multitasking delay of OSThread, providing much better responsiveness and avoiding missed inputs, and avoiding unnecessary periodic polling
  - After the input events are collected with `inputBroker->queueInputEvent(&event)` in `pollOnce()`, they will be processed by GUI and other modules on the main thread when possible via `inputBroker->processInputEventQueue()` which is shared among all input event sources that use this paradigm
  - Note that this relies on FreeRTOS, so it will not work on platforms that don't use FreeRTOS

Unfortunately I do not have a T-Deck Pro to test, but being so similar to the T-Lora Pager keyboard I'm guessing it should be fine.

I also have no way of testing regressions on `TCA8418Keyboard`, which AFAICT is for Nokia 5130 with multitap (click same key multiple times to cycle what character it produces); since it's multitap I did not add N-key rollover to it, I tried to keep it equivalent behavior to current, but can't test for regressions from the refactoring.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - [x] T-Lora Pager

## Summary by CodeRabbit

* **New Features**
  * Added interrupt-driven keyboard input support for compatible hardware.
  * Improved keyboard responsiveness by processing queued key events promptly.
  * Added support for light-sleep wake and automatic input polling.
  * Enabled canned message functionality on the T-Deck Pro.

* **Bug Fixes**
  * Improved multitap, long-press, modifier, and held-key handling.
  * Prevented duplicate actions from repeated key events.
  * Improved backspace and character selection behavior.


Szetya:
Hi @WillyJL! 

The "cardputer adv" keyboard is also TCA8418. There is no PR yet.
I shared the related files here
https://github.com/meshtastic/firmware/issues/7985#issue-3415275208
Could you later complete the development on this keyboard file as well?
Thanks! 😇

WillyJL:
> Hi @WillyJL! 
> 
> 
> 
> The "cardputer adv" keyboard is also TCA8418. There is no PR yet.
> 
> I shared the related files here
> 
> https://github.com/meshtastic/firmware/issues/7985#issue-3415275208
> 
> Could you later complete the development on this keyboard file as well?
> 
> Thanks! 😇

Of course! That was the next thing I wanted to try and figure out, since a YouTuber and friend of mine got a cardputer adv and noted how typing felt quite sluggish.

How should we go about it? Make a PR for cardputer adv first and once that's merged I update this PR for it? Or merge this first, and I'll provide the updated versions of the relevant files for cardputer adv for when that PR is opened?

Szetya:
> Of course! That was the next thing I wanted to try and figure out, since a YouTuber and friend of mine got a cardputer adv and noted how typing felt quite sluggish.
> 
> How should we go about it? Make a PR for cardputer adv first and once that's merged I update this PR for it? Or merge this first, and I'll provide the updated versions of the relevant files for cardputer adv for when that PR is opened?

I am not a PR expert. I don't want to start being one either. I'm afraid I would do more harm than good. :)
It's too complicated for me, and I'm afraid I'd mess something up and then freeze up.
I am happy to share similar things, the first steps. Sometimes I come up with ideas myself.
If what I have shared is useful, feel free to use it for PR.
I don't think the cardputer is in progress yet. 
Since the last modification to the TCA8418 keyboard was made by you, it will be easier for you to merge it.

WillyJL:
@Szetya just to confirm, are you the person that uploaded "Meshtastic For Cardputer-Adv" to M5Burner? i could not find a source for that firmware on M5Burner, and since it seems to work mostly well it would make sense to go off from that code to get it merged into here. if it is you, and its based on the files you shared in https://github.com/meshtastic/firmware/issues/7985#issue-3415275208, then i can open a PR for you with you as commit author, and based on whatever PR gets merged first i will update the other.

Szetya:
> @Szetya just to confirm, are you the person that uploaded "Meshtastic For Cardputer-Adv" to M5Burner? i could not find a source for that firmware on M5Burner, and since it seems to work mostly well it would make sense to go off from that code to get it merged into here. if it is you, and its based on the files you shared in [#7985 (comment)](https://github.com/meshtastic/firmware/issues/7985#issue-3415275208), then i can open a PR for you with you as commit author, and based on whatever PR gets merged first i will update the other.

No, I didn't upload it.
Since M5burner binary is not official, I would be happy if it were available here later.
Oh, I just noticed that some uppercase characters in the files are lowercase (e.g., n). 🙁

WillyJL:
> No, I didn't upload it.

interesting. so it would seem this "ozbk" guy on M5Burner is in breach of Meshtastic's GPLv3 license.

Szetya:
> > No, I didn't upload it.
> 
> interesting. so it would seem this "ozbk" guy on M5Burner is in breach of Meshtastic's GPLv3 license.

I tried searching for the user on GitHub. I was unsuccessful, or perhaps I wasn't persistent enough. 😉
However, I found them on the m5 community page.

mverch67 (commented):
I tested with a T-Deck Pro. The key combinations work except backspace: Typing backspace 2x within a short time reacts like escape and goes back to the main screen. Typing it slowly with a larger pause deletes the previously typed characters as expected.

Edit: I noticed when typing quickly a longer text several typed characters are missing. I believe this wasn't the case before this PR.

WillyJL:
Interesting, I can't replicate either of those on T-Lora Pager. I had that behavior, where pressing backspace multiple times quickly caused it to switch view (not quit the text input, but switch frame/tab in the ui to the clock one I think, and from there pressing backspace again kept deleting characters in the text input view even though I couldn't see it until all text was deleted where the next backspace removed the text input view from the tab list/switcher), before this PR, but not at all with these changes. Also, typing fast on it before I missed many inputs because I tend to click the next key before releasing the previous one, which with this PR is not a problem and registers correctly so I can type reliably much faster with this PR.

I'll reread the T-Deck Pro code in case I missed something but other than that debugging will be difficult for me as I don't have the device

WillyJL:
> > > No, I didn't upload it.
> 
> > 
> 
> > interesting. so it would seem this "ozbk" guy on M5Burner is in breach of Meshtastic's GPLv3 license.
> 
> 
> 
> I tried searching for the user on GitHub. I was unsuccessful, or perhaps I wasn't persistent enough. 😉
> 
> However, I found them on the m5 community page. 

I asked in the Meshtastic discord, someone asked then in the M5Stack discord, they said:

> There is no open source code yet, we are still optimizing

Which seems scummy to me. Anyway, I have built the code you shared in your issue and had my friend sn0ren test it with his cardputer adv (I don't have one), he said it works so I'll open a PR for it shortly with you as commit author so there's proper support without some third party gatekept build

Szetya:
> I tested with a T-Deck Pro. The key combinations work except backspace: Typing backspace 2x within a short time reacts like escape and goes back to the main screen. Typing it slowly with a larger pause deletes the previously typed characters as expected.
> 

I experienced this behavior with the backspace key just yesterday when using the original files (on the cardputer). I can repeat it over and over again. At first, I thought it was a feature.

Szetya:
> > > > No, I didn't upload it.
> > 
> > 
> > > 
> > 
> > 
> > > interesting. so it would seem this "ozbk" guy on M5Burner is in breach of Meshtastic's GPLv3 license.
> > 
> > 
> > I tried searching for the user on GitHub. I was unsuccessful, or perhaps I wasn't persistent enough. 😉
> > However, I found them on the m5 community page.
> 
> I asked in the Meshtastic discord, someone asked then in the M5Stack discord, they said:
> 
> > There is no open source code yet, we are still optimizing
> 
> Which seems scummy to me. Anyway, I have built the code you shared in your issue and had my friend sn0ren test it with his cardputer adv (I don't have one), he said it works so I'll open a PR for it shortly with you as commit author so there's proper support without some third party gatekept build

We need to agree on which display driver to use. I don't want to decide that myself.
I couldn't fix the sound, so there will definitely be more work to do on that.
@mverch67 You have experience with sound. Maybe you could help. 
Thank you for everything! 🤝

mverch67:
@WillyJL: this is 2.7.7 alpha compared to the PR branch, **quick typing test on the T-Deck Pro**. Some characters get eaten.

2.7.7

https://github.com/user-attachments/assets/dbf06ec3-c8fd-4e6d-a53d-2e4c40fa0631


PR:

https://github.com/user-attachments/assets/004d341b-01b1-44dc-a76c-6ef879be57a8

WillyJL:
@mverch67 thanks for the videos. The fact the screen refreshes slower makes me think it might be related. The TCA8418 has a 10 event FIFO, so I'm guessing the issue here is that there are more than 10 key press/releases before the main loop polls it again due to the gui sleeping (I'm assuming) longer to not redraw as fast on the epaper display, and it would make sense since it's at about 4-5 typed keys (so ~10 key events) not being drawn yet that it starts losing the last key not yet registered.

If my conclusion is correct, I see 2 ways to handle it. One is similar to the other PR I made, have a freertos thread that gets woken up by the interrupt and submit into queue, so the inputs don't depend on the main loop. Other is changing how display delays drawing, if my assumption of it sleeping longer is the case, to not sleep but instead just no-op and keep track of the last time to drew to the screen, so it can draw less often without sleeping on the main loop.

mverch67:
I have added the two logs. Indeed, there is a screen refresh pushing in when the characters get lost:
[2-7-7.txt](https://github.com/user-attachments/files/22363666/2-7-7.txt)
[PR.txt](https://github.com/user-attachments/files/22363671/PR.txt)

Edit: log not related to the videos... the text displayed now on the PR version is: "_this is ck t test_"

WillyJL:
> If my conclusion is correct, I see 2 ways to handle it. One is similar to the other PR I made, have a freertos thread that gets woken up by the interrupt and submit into queue, so the inputs don't depend on the main loop.

im thinking of ways to implement this fix, one way of course would be to do essentially the same thing i did for  the rotary encoder PR, but that means 2 background threads doing nothing until an interrupt happens...

i was thinking maybe its worth standardizing this? eg move this freertos thread to input broker and have an interrupt-safe method in input broker to give it a class with its own method to run ASAP in the freertos thread that would then queue events into a input broker queue and input broker would process them on the main thread in loop().

for example:
- interface `InputPollable` with virtual `pollOnce()` method
- `KbI2cBase` inherits from `InputPollable` and in its `pollOnce()` method it does the usual `TCAKeyboard.trigger()` but instead of `this->notifyObservers(&e)` it would do `inputBroker->queueEvent(&e)`
- in the interrupt handler put `inputBroker->pollSoonFromIsr(this)` (`this` being `KbI2cBase`, since it inherits `InputPollable`)
- `inputBroker->pollSoonFromIsr(InputPollable*)` will add this instance to a queue and wake up a freertos thread internal to input broker so as soon as ISR context is exited, this thread runs to poll these inputs
- somewhere in `loop()` add code to process events that were queued with `inputBroker->queueEvent(&e)` on the freertos thread

(names arent great can always change them, but i hope you get the idea)

this way gpio can be polled as soon as something happens to register the events so they are not lost (like they are here on tdeck pro), while handling them later when system is ready, and other input systems can use this model too without spawning each their own freertos thread and queue... @mverch67 if you too think this is a good idea i will implement it and update #7986 to use it too

one concern i have tho is im not sure what the state of freertos is on nrf52 based devices, is it possible to use freertos on them?

mverch67:
@WillyJL 
Your suggestions with the inputBroker / InputPollable sounds interesting. As with the queue the actual event time gets lost I assume that events need to be tagged with a time stamp so that long press / double press could be distinguished?

Regarding nrf52 I found [this](https://community.platformio.org/t/cannot-open-files-freertos-h-task-h-queue-h-timers-h-from-timers-c-in-platformio/40512/16) from last year which indicates that freeRTOS is available for the nordic platform. But care must be taken about the code size as nrf52 firmware is at its flash size limits.

WillyJL:
> As with the queue the actual event time gets lost I assume that events need to be tagged with a time stamp so that long press / double press could be distinguished?

i dont think it should: the way inputs reach the UI is the same as before, they are processed on the main loop, so for example if a screen refresh takes long the inputs will still be processed after they happen. this system however makes sure the inputs are recorded when they happen, even if the system is not ready to process them, so they do not get lost. AFAICT things like long press are handled in input sources, like TCA8418Keyboard class, which will emit different input events based on timing. this does not change, actually it would mean they are more responsive because it does not poll inputs with 300ms interval, but as soon a the hardware reports that an event has happened, so the logic in TCA8418Keyboard class to calculate time between inputs will be even more precise now. a concrete example of this is what you mentioned with pressing backspace 2 times quickly, for me on develop branch it often switches tab/frame instead of deleting 2 characters, while with this PR it always registers as deleting 2 characters, my guess is because it records the inputs right as they happen, not delayed by polling, thus the input timing is more accurate, even if they are processed by the UI later anyway.

as for FreeRTOS on nrf52, i noticed `src/freertosinc.h` which includes FreeRTOS headers if FreeRTOS is available for the given platform and sets the `HAS_FREE_RTOS` define, which is the case for esp32, nrf52 and rp2040. i added appropriate `#if` checks for `HAS_FREE_RTOS` and `MESHTASTIC_EXCLUDE_INPUTBROKER` so this new helper is only compiled in where appropriate. i tried building targets tlora-pager (esp32s3), rak4631 (nrf52840), rak3172 (stm32), rak11310 (rp2040) and pico2 (rp2350), they all compile fine (besides current issues with building on develop branch), and it works very well on T-Lora Pager. i will push shortly when develop branch is fixed up so i can merge it here.

unfortunately i do not have any other device on different architectures so i cannot test those without freertos besides "it compiles", but i assume its fine since i dont think there is any overlap between devices without freertos and devices with these input modules that im porting to a interrupt + freertos task + queue approach.

WillyJL:
besides some targets still failing to build in develop branch, this PR should be ready.

@mverch67 could you please test if the situation on T-Deck Pro is improved since your last test?

mverch67:
> @mverch67 could you please test if the situation on T-Deck Pro is improved since your last test?

Hi @WillyJL, I tested this branch with T-Deck Pro but noticed that the keyboard is not working at all. Then I defined the KB_INT in variant.h:
image
and now it works, no more eaten keys! Great work!
 

BTW: do you think it should also work without defining KB_INT? Maybe not all TCA8418 devices have a keyboard IRQ?

WillyJL:
You're absolutely right, I forgot to check whether some TCA8418 based keyboards don't have the interrupt pin. I will double check and also add back the code for polling in case some of them don't have interrupts, inside an ifndef

WillyJL:
that should do it. i tested on t-lora pager by removing KB_INT and indeed it still works, just slower. only other devices using TCA8418 currently are cardputer adv in #8010 which has KB_INT, and nokia 5130 in TCA8418Keyboard.cpp which i couldnt find much information on, so probably good enough like this.

EDIT: updated OP with a breakdown of InputPollable paradigm in case others want a short explanation of how to use it

WillyJL:
@mverch67 is it possible to reopen this? Or should I make a new PR?

WillyJL:
*sigh*

WillyJL:
solved conflicts and merged develop. tested on t-lora pager, still works great.

PR #8016, opened 2025-09-17
Allow Left / Right Events for selection and improve encoder responsives

Links2004:
Currently the rotery encoder skips and is very slow, this comes from the fact that the code only allows a event every 50ms.

This PR changes the debounce time to 5ms and allows the navigation of the selectors when the encoder is set to emit LEFT / RIGHT events.

Which as a result make the encoder way nicer to use.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


thebentern:
There is some parallel work going on to look into interrupts instead of increasing polling frequency over here #7986

Links2004:
The mentioned PR (#7986) only changes code for the `T_LORA_PAGER` that was using polling before (see ifdef at the top of the changed files).

The generic `RotaryEncoderInterruptBase.cpp` file that is changed in this PR uses interrupts already not polling.

the `setIntervalFromNow` in this case only defines how fast the "main" loop get the info that there is a "event" 
`runOnce` sets the interval back to `INT32_MAX`.

50ms is simply too slow for a rotery encoder, you easily can move a knop faster than that.
think of the value as input lag and an event limiter.

50ms means that the one rotation step is executed after 50ms.
Any second rotation in that time frame makes things worse since any new IRQ will reset the timer and will still result in only one event.
which means that you can rotate for a long time and only 50ms after you stop one event is fired.
It's like playing a game with at minimum 50ms input lag and max ~15 actions per second.

TLDR: A value of 50ms results in a knop that will skip a lot of steps and a UI that feels extremely slow and laggy.

5ms feels good and results in a responsive UI, feel free to test this yourself by setting the rotary to LEFT / RIGHT events, the difference is extremely noticeable.


It's not perfect the lower the better but at some point you can get bad inputs from cheap hardware.
dont get me wrong, you can solve this in hardware with a small cap ~2nF is more than enough, but all guides and the official docs simply connect a rotary encoder with out caps or pullups, and I did not want to break hardware that users build with this change.

with a good hardware setup 0ms is possible, which makes the experience even better ;)
Have added a define to allow configuring during compile if needed.

One thing to keep in mind is that the `OSThread` is Cooperative multitasking, so a value of 0ms does not mean that it happens instantaneously, its can be more, same for a value of 50ms the code only guarantees that at least 50ms are gone by before `runOnce` is called, but I can be 100ms or more too.

WillyJL:
@Links2004 Still, this could be improved. I'm working on a way that multiple input subsystems can request to be polled as soon as possible after an interrupt, which would bypass the delay imposed by cooperative multitasking with OSThread. See https://github.com/meshtastic/firmware/pull/7989#issuecomment-3300813448 for a breakdown. I'm currently working on implementing it in that PR, will then update #7986 to use the same system, and could possibly rework this rotary encoder too.

PR #8036, opened 2025-09-19
Add another seeed_xiao_nrf52840_kit build environment for I2C pinout

NomDeTom:
I have added another build environment for the `seeed_xiao_nrf52840_kit_i2c` at `board_level=extra`, which uses the base `seeed_xiao_nrf52840_kit` with the build flags set to use the i2c pins instead of GPS. This enables easy access for build_one_target users to compile the firmware on their own github or machine.

## 🤝 Attestations

- [tick] I have tested that my proposed changes behave as described.
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [tick] Other (please specify below)
A Discord user with this device has confirmed it uses the i2c pins as defined.


smurfy:
Is there a possibility to have this as an official alternative build on the web-flasher? 
It works great and i find adding sensors more useful than gps (which could be shared from phone).

Javito68:
> seeed_xiao_nrf52840_kit_i2c

Hi NomdeTom, where can I get the pre-compiled firmware with I2C support for: seeed_xiao_nrf52840_kit? Thanks.

PR #8047, opened 2025-09-19
C6l fixes

jp-bennett:
Found a BLE crash upon deinit, where the nimble driver seems to internally dereference a null pointer after deinit is called.

Also fixes a GPS crash during probing. My guess is that the c6l fails to put the UART into the 4800 or 57600 baudrates, and calling _serial_gps->available() in this state causes a crash.

### DIFF
--- a/src/nimble/NimbleBluetooth.cpp
+++ b/src/nimble/NimbleBluetooth.cpp
@@ -231,6 +231,8 @@ class NimbleBluetoothServerCallback : public NimBLEServerCallbacks
     {
         LOG_INFO("BLE disconnect");
 #endif
+        if (ble->isDeInit)
+            return;
 
         meshtastic::BluetoothStatus newStatus(meshtastic::BluetoothStatus::ConnectionState::DISCONNECTED);
         bluetoothStatus->updateStatus(&newStatus);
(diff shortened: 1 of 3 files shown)


jp-bennett:
@Links2004 GitHub wouldn't let me request your review, but the BLE change is directly related to your work with Nimble 2.3.6, and I'd love your thoughts

src/nimble/NimbleBluetooth.cpp
```diff
@@ -231,6 +231,8 @@ class NimbleBluetoothServerCallback : public NimBLEServerCallbacks
     {
         LOG_INFO("BLE disconnect");
 #endif
+        if (ble->isDeInit)
```
thebentern, line 235: Failing on some of the non-C6 ESP32 builds
```
src/nimble/NimbleBluetooth.cpp: In member function 'virtual void NimbleBluetoothServerCallback::onDisconnect(NimBLEServer*, ble_gap_conn_desc*)':
src/nimble/NimbleBluetooth.cpp:234:13: error: 'ble' was not declared in this scope
         if (ble->isDeInit)
             ^~~
src/nimble/NimbleBluetooth.cpp:234:13: note: suggested alternative: 'byte'
         if (ble->isDeInit)
             ^~~
             byte
```

PR #8053, opened 2025-09-20
Make sure next-hop is only set when they received us directly

GUVWAF:
While simulating some more scenarios with asymmetric links, it became clear that sometimes we’re assigning next-hops to nodes that we cannot directly reach, even though they relayed the packet and ACK to us. This is caused when we can hear them, but they have received the packet via another node.
Example scenario where node 2 can reach node 0, but not vice versa:
image

Hence, this PR ensures we only account a node as relayer when it relays **with 1 hop less than we did**.
For this, we need to store the hop limit with which we relay it, so this is added to the PacketHistory. This causes it to consume 20 bytes per entry due to padding, so I also increased `NUM_RELAYERS` to 6 to make full use of the allocated memory.
Furthermore, we need to make sure the FloodingRouter also is calling `wasSeenRecently()` when sending out a packet, hence not `Router::send()` is called, but rather its own implementation of `send()`.


src/mesh/PacketHistory.cpp
```diff
@@ -66,7 +66,13 @@ bool PacketHistory::wasSeenRecently(const meshtastic_MeshPacket *p, bool withUpd
...
     r.sender = getFrom(p); // If 0 then use our ID
     r.next_hop = p->next_hop;
-    r.relayed_by[0] = p->relay_node;
+    bool weWillRelay = false;
+    uint8_t ourRelayID = nodeDB->getLastByteOfNodeNum(nodeDB->getNodeNum());
+    if (p->relay_node == ourRelayID) { // If the relay_node is us, store it
+        weWillRelay = true;
+        r.ourTxHopLimit = p->hop_limit;
+        r.relayed_by[0] = p->relay_node;
+    }
```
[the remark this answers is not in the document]
GUVWAF: "Initialize the record to zero" is used before.

src/mesh/FloodingRouter.cpp
```diff
@@ -102,8 +102,7 @@ void FloodingRouter::perhapsRebroadcast(const meshtastic_MeshPacket *p)
 
                 LOG_INFO("Rebroadcast received floodmsg");
                 // Note: we are careful to resend using the original senders node id
-                // We are careful not to call our hooked version of send() - because we don't want to check this again
```
erayd, line 146: Is this concern no longer valid? There don't seem to be corresponding changes to a local `send()`.

GUVWAF: I've gone through the history and it was added here: https://github.com/meshtastic/firmware/commit/ea24394110759a35f1aa141ea6809e74f4f677ab

It seems "this" was referring to `wasSeenRecently()`, which makes sense, because originally we didn't need to check this again. However, now we need to add our hop limit to the record, so it should actually be called again.

erayd: Apparently I did not read the summary properly... looks like you already explained this there! Sorry for the spam...

GUVWAF: No worries, I appreciate the critical eyes :)

thebentern:
Tagging @compumike as well since I can't add non-org reviewers

compumike:
OK, I think I understand this and think it's a pretty important fix!

My understanding is:

## Goal

The goal is to reduce false-positive `next_hop` entries, either pointing at relays we can't reach bidirectionally, or relays that we can reach but can't themselves reach the destination node.

False-positive `next_hop` entries are a bad mesh experience, leading to delays before falling back to flooding, and reduced overall reliability.

## Asymmetric Connectivity

Because of asymmetric connectivity, we ideally want to make sure that we have bidirectional connectivity to any potential relay node we consider as a `next_hop`. But especially we would like to increase our certainty that the `next_hop` can hear us.

We do this by rejecting possible `next_hop` updates that are from relays that couldn't have possibly been downstream of us.

## Improvement in this PR

The `p->hop_limit == found->ourTxHopLimit - 1` heuristic makes sure they didn't hear it from some other way upstream or downstream, but it still allows the possibility that they heard it from another relay at the same hop level as we are.

## Change in `PacketHistory::wasSeenRecently`

The change "Only update the relayer if it heard us directly (meaning hopLimit is decreased by 1)" in `PacketHistory::wasSeenRecently` has to do with `p` as the **original request** (not the ACK).

This change means that if:

- We already relayed the original (via flooding is fine for this), AND
- We now hear someone else relaying (because they put their own `relay_node` on the packet), AND
- We hear them with a hop limit exactly one less than we relayed with,

then we ASSUME they (`p->relay_node`) heard it from us, so we add `p->relay_node` at the head of `r.relayed_by`.

Otherwise, we don't make that assumption, and we don't add `p->relay_node` to the `r.relayed_by` list.

This makes sense and is more conservative than the previous behavior.

**Important change: the `relayed_by` list now represents only "potential nodes that we heard relaying our relay (or another relay at the same hop as us)".**

## Change in `NextHopRouter::sniffReceived`

The change in `NextHopRouter::sniffReceived` only has to do with `p` as the **ACK packet** (not the original request).

This change allows two different conditions to cause an update for `next_hop` for node `origTx`, where `origTx` is the sender of the `ACK` (the recipient of the original request). We update `next_hop` if either:

- We already heard `p->relay_node` as a relayer on the original request (`wasAlreadyRelayer`), and we relayed the original request (`weWereRelayer`) (probably to `p->relayNode`), OR
- We didn't hear any other relayers (`weWereSoleRelayer`) and we got the ACK directly.

## Questions

1. What keeps the `relayed_by` list from having duplicate entries when we sniff retries from the same `relay_node`? (I think we want to dedup.)

2. If `startIdx = 0` don't we want to copy the entire `relayed_by` list from `found`? (Currently skips last entry; I think we want to copy all.)

3. Do we add ourselves to the relayed_by only AFTER the packet actually goes out on the radio? Or is it possibly still in our txQueue? Making sure that `relayed_by` can only add things AFTER we transmit would improve the ability of this heuristic to reject duplicate paths.

4. The `weWereSoleRelayer` check might mean something different now because of the changes to `PacketHistory::wasSeenRecently`. Since `relayed_by` now only stores relays we heard that are one hop down from us, the `weWereSoleRelayer` doesn't protect against the case where we can hear the ACKing node, but it can't hear us, **even if we sniffed another relay at the same hop level**. In this case, we could hear the ACK and think we can point `next_hop` directly to the destination, incorrectly. Not sure this is easily solvable, but I do think the change in the meaning of `relayed_by` may make this check as currently written be much less reliable?

(May need to store two `relayed_by` lists: one for those that might have heard it from us [as implemented in this PR], and another that's for any one we heard that might be received by the destination, regardless of hop level. But maybe not worth doing right now.)

5. Might be outside of the scope of this PR, but there's still the potential to incorrectly set `next_hop` when there are asymmetric connections. Consider:

```
A ↔ B ↔ C ↔ D
↕       ↓
EEEEEEEEE
```
In this diagram, I mean:
- there's a chain of bidirectional connectivity A-B-C-D
- bidirectional connectivity A-E
- E can hear C, but C can not hear E

```
A sends a DM to D. (Initially assume all `next_hop` are blank.)

PKT_1:
A transmits P_ORIG(id=123 from=A to=D want_ack=true relay_node=A hop_start=3 hop_limit=3)
A's PacketHistory = (A, 123, relayed_by=[A], ourTxHopLimit=3)

B and E both hear A.
B's PacketHistory = (A, 123, relayed_by=[], ourTxHopLimit=0)
E's PacketHistory = (A, 123, relayed_by=[], ourTxHopLimit=0)

PKT_2: (suppose E transmits first)
E transmits P_ORIG(id=123 from=A to=D want_ack=true relay_node=E hop_start=3 hop_limit=2)
E's PacketHistory = (A, 123, relayed_by=[E], ourTxHopLimit=2)
A hears this and suppresses its retries. A adds E to relayed_by:
A's PacketHistory = (A, 123, relayed_by=[E, A], ourTxHopLimit=3)


PKT_3: (B doesn't hear E and transmits next:)
B transmits P_ORIG(id=123 from=A to=D want_ack=true relay_node=B hop_start=3 hop_limit=2)
B's PacketHistory = (A, 123, relayed_by=[B], ourTxHopLimit=2)
A hears this and adds B to relayed_by:
A's PacketHistory = (A, 123, relayed_by=[B, E, A])
C hears B:
C's PacketHistory = (A, 123, relayed_by=[], ourTxHopLimit=0)

PKT_4:
C transmits P_ORIG(id=123 from=A to=D want_ack=true relay_node=C hop_start=3 hop_limit=1)
C's PacketHistory = (A, 123, relayed_by=[C], ourTxHopLimit=1)
B hears this and adds C to relayed_by:
B's PacketHistory = (A, 123, relayed_by=[C, B], ourTxHopLimit=2)
E hears C and adds C to relayed_by, since it passes the hop_limit == ourTxHopLimit - 1 test. (This is where things get asymmetric / weird.)
E's PacketHistory = (A, 123, relayed_by=[C, E], ourTxHopLimit=2)

D hears C. Yay. DM received! Time to ack.

PKT_5:
D transmits P_ACK(id=789 request_id=123 from=D to=A want_ack=false relay_node=D hop_start=3 hop_limit=3)
D's PacketHistory = (D, 789, relayed_by=[D], ourTxHopLimit=3)

C hears D:
C's NextHopRouter::sniffReceived looks up (A, 123). Has wasAlreadyRelayer=false, weWereRelayer=true, weWereSoleRelayer=true. The (p->hop_start != 0 && p->hop_start == p->hop_limit && weWereSoleRelayer) condition succeeds.
C sets D.next_hop = D
C's PacketHistory = (A, 123, relayed_by=[C], ourTxHopLimit=1), (D, 789, relayed_by=[], ourTxHopLimit=0)

PKT_6:
C transmits P_ACK(id=789 request_id=123 from=D to=A want_ack=false relay_node=C hop_start=3 hop_limit=2)
C's PacketHistory = (A, 123, relayed_by=[C], ourTxHopLimit=1), (D, 789, relayed_by=[C], ourTxHopLimit=2)

B hears C.
B's NextHopRouter::sniffReceived looks up (A, 123). Has wasAlreadyRelayer=true, weWereRelayer=true, weWereSoleRelayer=false. The (weWereRelayer && wasAlreadyRelayer) condition succeeds.
B sets D.next_hop = C
B's PacketHistory = (A, 123, relayed_by=[C, B], ourTxHopLimit=2), (D, 789, relayed_by=[], ourTxHopLimit=0)

E also hears C.
E's NextHopRouter::sniffReceived looks up (A, 123). Has wasAlreadyRelayer=true, weWereRelayer=true, weWereSoleRelayer=false. The (weWereRelayer && wasAlreadyRelayer) condition succeeds.
E sets D.next_hop = C <--- this entry is incorrect and will cause problems!
E's PacketHistory = (A, 123, relayed_by=[C, E], ourTxHopLimit=2), (D, 789, relayed_by=[], ourTxHopLimit=0)

PKT_7: (suppose E goes first)
E transmits P_ACK(id=789 request_id=123 from=D to=A want_ack=false relay_node=E hop_start=3 hop_limit=1)
A hears E.
A's PacketHistory = (A, 123, relayed_by=[B, E, A]), (D, 789, relayed_by=[])
A's NextHopRouter::sniffReceived looks up (A, 123). Has wasAlreadyRelayer=true, weWereRelayer=true, weWereSoleRelayer=false. The (weWereRelayer && wasAlreadyRelayer) condition succeeds.
A sets D.next_hop = E <-- this entry is incorrect and will cause problems!

PKT_8: (B doesn't know about E's transmission)
B transmits P_ACK(id=789 request_id=123 from=D to=A want_ack=false relay_node=B hop_start=3 hop_limit=1)
A hears B.
A's NextHopRouter::shouldFilterReceived calls wasSeenRecently and it returns true, so sniffReceived never runs.

Final next_hop states:
C: D.next_hop = D
B: D.next_hop = C
E: D.next_hop = C (incorrect, will cause problems)
A: D.next_hop = E (incorrect, will cause problems)
```

## Overall

Definitely a good thing! I think it will reduce a lot of incorrect `next_hop` assignments.

Have to think about it some more, but I think lots of these weird cases would go away if packets had a `prev_relay_node` byte. That byte could even be used when the destination originates its ACK. In the example above, D would transmit its ACK with: from=D to=A relay_node=D prev_relay_node=C hop_start=3 hop_limit=3. And from that, C could more confidently assign its next_hop.

GUVWAF:
Thanks for the extensive review, @compumike.

1. Indeed, we may want to add this, although it’s uncommon for a node to relay twice (only for fallback to flooding or the original transmitter). This will already increase the number of relayers from 3 to 6, so that also helps.

2. Good point! It should be `NUM_RELAYERS – startIdx`.

3. It’s already added before, but when we cancel a transmission, we remove ourselves as relayer:
https://github.com/meshtastic/firmware/blob/47a82bdb98c46841847be0034818ea3d8da279f1/src/mesh/Router.cpp#L324-L325

4.  Indeed. I wonder if we should add relayers then if it’s one hop less **or the same hop**. A node can also be our next-hop if it hears the same transmissions as us, but can transmit further. What do you think?

5. Your example is correct and will lead to an issue indeed. Without having the full path each message took, I believe we will end up with these kind of issues. However, I believe this is a rather uncommon case and there is always still fallback to flooding, which would resolve this with just one retransmission.

thebentern:
This has some merge conflicts now based on the upstream changes for the other routing PRs

GUVWAF:
I'll try to resolve them together with the other comments later today.

GUVWAF:
I resolved the conflicts (by reusing the `hop_limit` value and separating it into two times 3 bits), and addressed comments 1, 2 and 4 (3 was already solved).

shalberd:
Hi @GUVWAF and @compumike thank you very much for your hard work on making next hop routing even better.

By the way, for mountain nodes >1500m widely visible in terms of radio coverage, changing them to ROUTER_LATE really helped out traffic here in Switzerland. I also encouraged people to use CLIENT when in lateral valleys not accessible from all 4 directions, even from higher locations up on each V-valley side.

GUVWAF:
You're welcome and thanks for the feedback, that's great to hear!

PR #8054, opened 2025-09-20
rework sensor instantiation to saves memory by removing the static allocation

Links2004:
rework I2C sensor init.
    
The goal is to only instantiate sensors that are pressend to save memory.

Side effacts:
  - easier sensor integration (less C&P code, sensor only needs to be "registerd" at one place)
  - nodeTelemetrySensorsMap can be removed when all devices are migrated to save more RAM

## 🤝 Attestations

  - [x] Heltec (Lora32) V2.1
  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -562,6 +562,7 @@ class AnalogBatteryLevel : public HasBatteryLevel
                    config.power.device_battery_ina_address) {
             if (!ina226Sensor.isInitialized())
                 return ina226Sensor.runOnce() > 0;
+            return ina226Sensor.isRunning();
         } else if (nodeTelemetrySensorsMap[meshtastic_TelemetrySensorType_INA260].first ==
                    config.power.device_battery_ina_address) {
             if (!ina260Sensor.isInitialized())
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -647,23 +647,24 @@ void RadioInterface::limitPower(int8_t loraMaxPower)
     }
 
 #ifndef NUM_PA_POINTS
-    if (TX_GAIN_LORA > 0) {
+    if (TX_GAIN_LORA > 0 && !devicestate.owner.is_licensed) {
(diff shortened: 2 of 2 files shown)


mverch67:
A much larger portion of static allocation could be saved if all the sensor instances are declared as pointer and allocated in heap space during [init](https://github.com/meshtastic/firmware/blob/b49496d99d20f6c351387b91bf303d89914c9b55/src/modules/Telemetry/EnvironmentTelemetry.cpp#L246). Then no changes in this BSEC sensor class are required,  e.g. see here:
https://github.com/meshtastic/firmware/blob/b49496d99d20f6c351387b91bf303d89914c9b55/src/modules/Telemetry/EnvironmentTelemetry.cpp#L72

Links2004:
true, had a look at EnvironmentTelemetry.cpp and it looks like there is a lot more possibility for optimizion.
In my opinion there is a loot to gain there, the current desigen looks very inefficient over all.

looked a bit deeper and I think the I2C device handling overall needs some work.

Links2004:
have working code that elimenates the need for `nodeTelemetrySensorsMap` and only creates the sensor instances when the I2C scan found a device.

moved over 4 sensors to show case the idea, but want some feedback before doing all of them.
moving this 4 sensors over saved already 1152 bytes of static allocations.

I dont see that we get memory fragmentation problem with this code, since the only cases where we delete / free something is when the sensor failed it initialisation.

tested this with a BMP-280 connected to a Heltec V2.1

please review and give feedback if this code is something that can be merged (will migrate the other sensors then).

Links2004:
did a second batch of sensors

RAM -816
Flash -916

thebentern:
Tagging @oscgonfer for visibility

Links2004:
The PR now covers all sensors that where allocated staticly in `EnvironmentTelemetry`.

Total saved:

RAM -2160
Flash -4332

we will save CPU cycles too since we do a lot less `hasSensor` calls.

some sensors like the INA ones are part of `PowerTelemetry` and `EnvironmentTelemetry` not sure why but I did not touch them.

Ready for review.

PR #8059, opened 2025-09-20
Device Telemetry opt in

thebentern:
@garthvh 
@jamesarich I'm going to hold off on pushing this into a release until we have this plumbed into the config in the app store and playstore releases. I don't want to make folks opt in with no way to do it. 😅 

- [x] iOS implemented
- [x] Android implemented
- [ ] Web implemented
- [ ] Python implemented


RCGV1 (approved):

cwb:
How does one turn this back on in 2.7.13?

```
~ ❯❯❯ meshtastic -t meshtastic.local --set telemetry.device_telemetry_enabled true
Connected to radio
LocalConfig and LocalModuleConfig do not have attribute telemetry.device_telemetry_enabled.
Choices are...
```

Also no luck in the web UI or iOS app...

ianmcorvidae:
I know I talked to you in discord as well, but for anyone who comes across this here and wonders the answer: the above command works with the CLI version 2.7.4 or later. Other client apps will also be picking up the option over time, of course.

jpryor223:
not finding the the Web implemented yet, not finding an issue or PR if one needs to be created; ok with the iOS and Android implemented on the main telemetry config, however seeing this setting will also need to be implemented in the Remote Administration telemetry config, can only see the interval right now

Issue #8029, opened 2025-09-18
Wait until after GPS lock hold before updating position, if we can.


fifieldt:
After the [recent patch](https://github.com/meshtastic/firmware/pull/7801) , we hold lock for a bit before updating the position. The positions that come in after the hold are genuinely better positions than what we've been doing before. However, they only come 20 seconds after we've got lock. 

So, at the moment, to avoid badness while making an improvement what the code is set up to do is update the local position twice - first, when lock is obtained, and second, after the hold expires. Since local position updates are asynch to mesh broadcasts, it means in many cases the good position will be broadcast, but not always.

Instead, if the settings allow, we should skip that first local position update.

"If the settings allow" is the meat if this issue. If we look at the broadcast/smart update/gps update intervals and they are all well above 20 seconds, it's likely safe. However, on the other end, we don't want to break GPS updates for the people who leave GPS permanently on (gps_interval <10secs). We want to avoid potential loops where any of the broadcast/update/interval settings are set to exactly or less than the 20 second hold time. All doable, just needs some attention to detail.

compumike:
This would be a good improvement, and I am guessing it would likely fix some of the GPS weirdness like I posted here: https://discord.com/channels/867578229534359593/1202456506587418664/1417561253412344000 with a really bad position followed by a reasonable position within the same minute.

I'm not super familiar with the GPS power management code, but you might instead be able to do this with a counter that tracks the number of position updates received from the GPS since last setting the power to `ACTIVE`.

Json-181:
To add to this, could we also add a toggle to easily purge position data from nodes that don't have gps, and are broadcasting the phone's shared location? 
For example, sometimes you want to share a location via the phone share, but then later would like the location on the node to not be broadcasted. But once that node has a location its more or less there now.

fifieldt:
@Json-181 probably need a different issue to track that.

fifieldt:
@compumike , yup, this will fix that :)

Imagine, that prior to 3 weeks ago every position in Meshtastic was potentially that bad!

fifieldt:
OK, so here's how I think this is going to pan out:

gps_update_interval | Behaviour
-- | --
1-10 secs | No hold logic. Always update the local position. Accuracy comes from the fact the GPS is always on (see GPS::down() )
11-19 secs | Hold for (gps_update_interval – 1). Update local position after the hold. This gives the best possible chance for a high quality lock, given our limited hold time.
[...]

PR #8064, opened 2025-09-21
[WIP] Wait until after GPS lock hold before updating position, if we can.

fifieldt:
After the recent patch, we hold lock for a bit before updating the position. The positions that come in after the hold are genuinely better positions
 than what we've been doing before. However, they only come 20 seconds
 after we've got lock.

Previously, we would update the local position as soon as we got a lock as well as at the end of the hold, since a hold was not always possible. With this patch, if the settings allow, we should skip that first local position update.

Fixes https://github.com/meshtastic/firmware/issues/8029

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


fifieldt:
(untested, and some code can be optimised, back on this in a couple days)

src/gps/GPS.cpp
```diff
@@ -1118,55 +1141,71 @@ int32_t GPS::runOnce()
...
     bool tooLong = scheduling.searchedTooLong();
     if (tooLong)
         LOG_WARN("Couldn't publish a valid location: didn't get a GPS lock in time");
 
-    // Once we get a location we no longer desperately want an update
-    if ((gotLoc && gotTime) || tooLong) {
-        if (tooLong && !gotLoc) {
-            // we didn't get a location during this ack window, therefore declare loss of lock
-            if (hasValidLocation) {
-                LOG_DEBUG("hasValidLocation FALLING EDGE");
```
fifieldt, line 1145: Bug - need to remember to handle falling edge.

src/gps/GPS.cpp
```diff
@@ -1118,55 +1142,81 @@ int32_t GPS::runOnce()
...
-            shouldPublish = true; // publish our update at the end of the lock hold
-            publishUpdate();
-            down();
+    }
+
+    // Hold has expired , Search time has expired, we got a time only, or we never needed to hold.
+    if (shouldPublish || tooLong || millis() > fixHoldEnds) {
+        if (gotTime && hasValidLocation) {
+            shouldPublish = true;
+        }
```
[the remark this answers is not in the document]
fifieldt: publishUpdate() checks for shouldPublish when it is called, it won't publish if the flag is not set.

src/gps/GPS.cpp
```diff
@@ -1118,55 +1142,81 @@ int32_t GPS::runOnce()
...
     if (gotLoc && prev_fixQual == 0) { // just got a lock after turning back on.
-        fixHoldEnds = millis() + 20000;
-        shouldPublish = true; // Publish immediately, since next publish is at end of hold
+        if (updateInterval <= 10 * 1000UL) {
+            shouldPublish = true;
+        } else {
+            // Hold for up to 20secs after getting a lock to download ephemeris etc
+            uint32_t holdTime = updateInterval - 1000;
+            if (holdTime > 20000)
+                holdTime = 20000;
```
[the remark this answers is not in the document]
fifieldt: will do!

fifieldt:
Definitely bugs in this one ...
```
INFO | ??:??:?? 91 [GPS] GPS power state move from HARDSLEEP to HARDSLEEP
DEBUG | ??:??:?? 96 [GPS] Took 10s to get lock
DEBUG | ??:??:?? 96 [GPS] Predict 31s to get next lock
DEBUG | ??:??:?? 96 [GPS] 28s until next search
```

fifieldt:
Testing with T1000E

gps_update_interval | Behaviour | Test
-- | -- | --
1-10 secs | No hold logic. Always update the local position. Accuracy comes from the fact the GPS is always on (see GPS::down() ) | OK
11-19 secs | Hold for (gps_update_interval – 10). Update local position after the hold. This gives the best possible chance for a high quality lock, given our limited hold time. | OK
20 secs | As per 11-19 seconds (so hold for 10 secs) | OK
21-40 secs | Hold for 20 seconds, and then update the local position. |OK
41+ secs | As per 21-40 seconds. | OK


Ended up changing the hold time for lower gps_update_intervals to gps_update_intervals - 10sec to avoid GPS being stuck on all the time for those.

fifieldt:
Confirmed tooLong still works correctly as before:

```
WARN | 02:46:54 908 [GPS] Couldn't publish a valid location: didn't get a GPS lock in time

DEBUG | 02:46:54 908 [GPS] Took 900s to get lock

DEBUG | 02:46:54 908 [GPS] Predict 0s to get next lock

DEBUG | 02:46:54 908 [GPS] 29s until next search

INFO | 02:46:54 908 [GPS] GPS power state move from ACTIVE to HARDSLEEP
```

src/gps/GPS.cpp
```diff
@@ -1113,55 +1136,81 @@ int32_t GPS::runOnce()
...
+        if (gotLoc) {
+#ifdef GPS_DEBUG
+            if (!hasValidLocation) { // declare that we have location ASAP
+                LOG_DEBUG("hasValidLocation RISING EDGE");
+            }
+#endif
+            if (updateInterval <= GPS_UPDATE_ALWAYS_ON_THRESHOLD_MS) {
+                hasValidLocation = true;
+                shouldPublish = true;
+            } else if (!hasValidLocation || prev_fixQual == 0 || fixHoldEnds < millis()) {
```
fifieldt: there's a race condition here - we can potentially get here just as the hold ends and then get a new position and hold again.

fifieldt: need a new way to check for IDLE --> ACTIVE :(

fifieldt: might be a simple fix - trying to add GPS_THREAD_INTERVAL :)

fifieldt: works!

b8b8:
GPS update: 8sec hold no hold - confirmed
GPS update: 25sec hold for 15sec - confirmed
GPS update: 35sec hold for 20 sec - confirmed
GPS update: 120sec hold for 20 sec - confirmed (hmm i thought that should be higher)

fifieldt:
Thanks, @b8b8 . 20s is the max hold time.

fifieldt:
Confirmed with b8b8's testing that the map view is now a lot smoother, positions are more accurate etc.

fifieldt:
Ran a loop inside another loop to see if the parallel paths worked out OK.  It did, the paths were clear and distinct, with no overlap. This is good position data :)


Screenshot_20251007-152134

PR #8088, opened 2025-09-23
Mute: Nodes (Draft 1)

### LINKED ISSUE #7943: [Feature Request]: Mute Notifications for Specific Chats

ford-jones:
Resolves #7943

Introduces mechanisms to mute message notifications received from certain nodes.
When a message is received from a muted node, the device's buzzer (if present) should not be activated. Drawing of any alert UI notification to the device's screen should also be blocked.

The exception to this is bell alerts (when activated), which should still sound the buzzer and display a notification.

It should be noted that this behaviour is distinctly different to ignoring a node - an action which blocks the message entirely. Muted messages can still be viewed in the messages menu of the device or client application, they are just delivered discreetly - similar to muting a channel.

- [x] Buzzer always sounds when no nodes are muted
- [x] OLED popup always displays when no nodes are muted
- [x] When a specific node is muted, messages received from that node don't invoke the buzzer or a popup

Protobuf changes: https://github.com/meshtastic/protobufs/pull/779
Python CLI Changes: https://github.com/meshtastic/python/pull/822
Docs: https://github.com/meshtastic/meshtastic/pull/2038

### 🤝 Attestations
  - [x] Heltec (Lora32) V3
  - [x] LilyGo T-Deck

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -665,7 +665,8 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
 
 #if defined(ST7735_CS) || defined(USE_EINK) || defined(ILI9341_DRIVER) || defined(ILI9342_DRIVER) || defined(ST7789_CS) ||       \
     defined(HX8357_CS) || defined(USE_ST7789) || defined(ILI9488_CS) || defined(ST7796_CS) || defined(USE_SPISSD1306) ||         \
-    defined(USE_ST7796) || defined(HACKADAY_COMMUNICATOR) bool hasScreen = true;
+    defined(USE_ST7796) || defined(HACKADAY_COMMUNICATOR)
+    bool hasScreen = true;
 #ifdef HELTEC_MESH_NODE_T114
     uint32_t st7789_id = get_st7789_id(ST7789_NSS, ST7789_SCK, ST7789_SDA, ST7789_RS, ST7789_RESET);
     if (st7789_id == 0xFFFFFF) {


ford-jones:
@garthvh Here's the one for nodes, should we revisit this in future.
Will leave it in draft for now.

ford-jones:
@garthvh the `is_muted` flag has now been added to `NodeInfo`, let me know if this checks out :+1:

src/mesh/generated/meshtastic/admin.pb.h
```diff
@@ -249,6 +249,10 @@ typedef struct _meshtastic_AdminMessage {
         uint32_t set_ignored_node;
```
fifieldt, line 259: If the protobufs have been updated, we shouldn't see the changes in here. Someone probably needs to do a protobuf merge from master or something ...

ford-jones: Hey! Not sure I'm understanding correctly. 
These protobuf additions (`set_muted_node` and `remove_muted_node`) were added by me and pertain to this pull request. This stuff here is the result of running `regen-protos.sh`.

Is that what you're referring to?

The protobuf changes are here: https://github.com/meshtastic/protobufs/pull/779

fifieldt: generally the protobuf changes are merged into the repo separately through an automated process, so individual patches shouldn't need to carry the changes...

ford-jones: I hear ya, that's why I opened the proto PR. I'm just a little confused as this file we're discussing is one of the `generated/meshtastic/*.pb.h`. Should these not be checked in? Otherwise this wouldn't build or pass the CI.

Happy to jump in a call to get on the same page when you're next available :slightly_smiling_face: 
Sorry if I'm totally misunderstanding.

fifieldt: New, you're right - since the protobuf was released they should be here, and so these changes would just magically disappear . But it's not, so the automation is not working.

ford-jones:
@thebentern here's the PR we were chattin about on the phone today :slightly_smiling_face:

ford-jones:
@fifieldt let me know if this is looking better to you, I just resolved a conflict in the protobuf submodule's ref's.

PR #8097, opened 2025-09-23
Add moar nodedb nodes on esp32 w/ psram to 800, stored messages to 200

h3lix1:
WARNING: This was vibe coded from gpt5-codex-high. It seems to work, and looking at it initially, it isn't bad. (Trust, this is better than most of my other c++ work)

I have tested this over the last day. I have found bugs with counters, but those are squished.

Note, "online" counter updates after it receives time from NTP/GPS, and will initially show a higher number on boot.

There is a bubble sort used for nodedb. It's completing normally within 3-4ms, but sometimes jumps to 11ms. This seems OK, but willing to accept advice here.

Node Hot/Cold Split

ESP32‑S3 builds now keep the 196 B meshtastic_NodeInfoLite payload in PSRAM using a custom allocator that calls heap_caps_malloc(MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT) (src/mesh/NodeDB.h:20, src/mesh/NodeDB.cpp:73).
DRAM carries only the latency-critical fields in a NodeHotEntry cache (~20 B per node: num, last_heard, snr, channel/flags) alongside dirtiness bits for sync-on-demand (src/mesh/NodeDB.h:33, src/mesh/NodeDB.cpp:78).
Sorting, routing, favorite flips, online counts, and next-hop decisions run entirely out of that hot cache, so the usual packet/route/UI fast paths stay in internal RAM.

Memory Footprint per Node (bytes)

  | DRAM | PSRAM | Total
-- | -- | -- | --
Existing build | 196 | 0 | 196
New split | 20 | 196 | 216
Net change | -176 | +196 | +20

- For 800 nodes on Station G2 (4/8MB PSRAM) the split frees roughly 140 KB of internal heap (800 × 176 B) while PSRAM absorbs ~157 KB of cold data.

Capacity & Secondary Effects:

- MAX_NUM_NODES is a constant 800 as long as psram size is > 4, otherwise the old flash-based limits apply (src/mesh/mesh-pb-constants.h:54).

- The packet history ring still targets max(MAX_NUM_NODES*2, …) entries (src/mesh/PacketHistory.cpp:11), so doubling the node ceiling means the history structure grows accordingly—keep an eye on overall PSRAM consumption if future caps rise again.

Serialization & Cold Access:

- NodeDB save/load moves through PSRAM: hot nodes are copied into a temporary vector before protobuf encoding, then cleared back out after disk writes (src/mesh/NodeDB.cpp:1322, src/mesh/NodeDB.cpp:1414).
- GUI detail panes, phone syncs, and other “profile” views touch PSRAM when they dereference cold fields. These flows happen far less frequently, so the added latency is acceptable.

Runtime Behavior:

- Fast paths (unchanged timing): neighbor sorting, routing decisions, getMeshNodeChannel, set_favorite, online counts, packet next-hop updates (src/mesh/NodeDB.cpp:1750, src/mesh/NodeDB.cpp:1939, src/mesh/NodeDB.cpp:2156).
- Potentially slower paths: full NodeInfo dumps to the phone, detail panels that copy cold payloads, database saves—each now copies between DRAM and PSRAM but only on demand.

Large Mesh Readiness:

- Telemetry and UI counters are widened to 16 bit, keeping online/total counts accurate past 255 nodes (src/NodeStatus.h:16, src/mesh/ProtobufModule.h:16).
- InkHUD map passes now iterate with size_t, so they handle the full PSRAM-backed node list without truncation (src/graphics/niche/InkHUD/Applets/Bases/Map/MapApplet.cpp:156).
- The result is a Station G2-class ESP32‑S3 node that can track ~800 peers, with ~20 B/node of hot metadata in DRAM and the 196 B/node cold payload safely in PSRAM.

Also recently added:

### PSRAM-aware message/node expansion

- Gate the “moar messages / moar nodes” knobs behind `has_psram()` so we only scale up when ≥2 MB of PSRAM is available.
- Move the MeshPacket pool into a PSRAM-backed allocator on ESP32-S3; if allocation fails we fall back to heap so radios keep working.
- Bump the BLE message queue to 200 entries (356 B each → ~84.7 KB in PSRAM plus ~0.8 KB of DRAM for the pointer ring) while keeping a runtime limit that collapses back to 32 messages if PSRAM isn’t present.
- Reuse the same helper for NodeDB sizing, so nodes stay capped on low-memory boards without more `ESP.getPsramSize()` calls.

## 🤝 Attestations

  - [X] Heltec (Lora32) V4
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [X] Seeed Studio T-1000E tracker card (Does not add/subtract functionality)
Station G2
LILYGO Pager


src/mesh/NodeDB.cpp
```diff
@@ -1024,6 +1186,29 @@ void NodeDB::clearLocalPosition()
...
 void NodeDB::cleanupMeshDB()
 {
+#if defined(CONFIG_IDF_TARGET_ESP32S3)
+    refreshHotCache();
+    int newPos = 0, removed = 0;
+    for (int i = 0; i < numMeshNodes; i++) {
+        auto &node = psramMeshNodes[i];
+        if (node.has_user) {
+            if (node.user.public_key.size > 0) {
+                if (memfll(node.user.public_key.bytes, 0, node.user.public_key.size)) {
```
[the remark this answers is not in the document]
h3lix1: meshUtils.h declares this..

src/mesh/NodeDB.cpp
```diff
@@ -1873,6 +2218,59 @@ bool NodeDB::isFull()
...
+            int oldestBoringIndex = -1;
+            for (int i = 1; i < numMeshNodes; i++) {
+                const NodeHotEntry &hot = hotNodes[i];
+                if (!(hot.flags & HOT_FLAG_IS_FAVORITE) && !(hot.flags & HOT_FLAG_IS_IGNORED) &&
+                    !(hot.bitfield & NODEINFO_BITFIELD_IS_KEY_MANUALLY_VERIFIED_MASK) && hot.last_heard < oldest) {
+                    oldest = hot.last_heard;
+                    oldestIndex = i;
+                }
+                const auto &node = psramMeshNodes[i];
+                if (!(hot.flags & HOT_FLAG_IS_FAVORITE) && !(hot.flags & HOT_FLAG_IS_IGNORED) && node.user.public_key.size == 0 &&
```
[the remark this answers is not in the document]
h3lix1: Renamed

NomDeTom:
Is this extensible to the extra qspi flash on the xiao NRF52?

https://wiki.seeedstudio.com/xiao-ble-qspi-flash-usage/

h3lix1:
@NomDeTom I'm not sure we want to slow the nrf52 down any more than it is already.

NomDeTom:
>  I'm not sure we want to slow the nrf52 down any more than it is already.

I was just thinking of things like nodeDB rolling attacks could be resisted more easily by increasing the size. I'm not sure this would slow it down particularly.

h3lix1:
I currently lack the skill required to make this work for nrf52. Placing in draft for now until someone more talented than I am can make this work.

h3lix1:
I currently have a $50 bounty out for anybody better than me who can do this for nrf52 nodes with flash.  In the meantime, can we get this in at least for the ESP32s out there?

garthvh:
Can't rush this in, 200 nodes is already problematic and entirely vibe coded solutions are generally buggy, get some people testing builds for this.

h3lix1:
More testing complete on this MR and changes since the first revision

- Store 100 messages (up from 32) for BOARD_MAX_RX_TOPHONE that will be stored in PSRAM
- Storing 3000 nodes in NodeDB. This might be problematic for bluetooth LE nodes
- Wifi is very fast and is able to load 700 nodes in less than a second

Comparing to the development branch, even with 3000 nodes, it is saving about 16% heap memory. All memory is allocated ahead of time.
| Phase   | Field       |   Non-PSRAM (bytes / MB)  |     PSRAM (bytes / MB)    |  Difference (bytes / MB)  | % diff (PSRAM vs Non) |
| :------ | :---------- | :-----------------------: | :-----------------------: | :-----------------------: | :-------------------: |
| Running | Free heap   |  176,700 bytes / 0.17 MB  |  204,648 bytes / 0.20 MB  |  +27,948 bytes / +0.03 MB |        +15.82%        |
| Running | Free PSRAM  | 2,035,403 bytes / 1.94 MB | 1,278,319 bytes / 1.22 MB | −757,084 bytes / −0.72 MB |        −37.20%        |

- Additional log message during bootup
`INFO  | ??:??:?? 2 NodeDB PSRAM backing at 0x3de00800 (DRAM) capacity 3000 entries (~588000 bytes)`

Testing on a production router has proven successful with no reboots and 661 nodes currently.

So far this change has been tested successfully on the following platforms:
[x] Heltec (Lora32) V4
[x] Seeed Studio T-1000E tracker card (Does not add/subtract functionality)
[x] Station G2
[x] T-Lora Pager
[x] T-Beam-S3Core
[] T-Deck (awaiting delivery)

With 2MB of PSRAM this will use 37%. With 8MB of most ESP32-S3 nodes.

Next is to move PacketRecord to PSRAM for a savings of about 120KB with NUM_MAX_NODES == 3000, making the ring 6000 entries. For now it fits in DRAM.

Moving back to draft for now, but this is looking very good.

garthvh:
200 nodes is slow over WiFi

h3lix1:
@garthvh For me it's very fast with the Lora V4 and Xiao Wio. Bluetooth is a different beast and can take a few minutes to get 400 nodes. It seems like a lazy population might be better for bluetooth nodes if the client side can support that. (multiple queues possibly?) . I guess we can just limit this to wifi enabled nodes, or only send the most recently heard 100 nodes if on bluetooth to the client.

garthvh:
Needs to be compatible with the 90% of people using Bluetooth, TCP is also pretty slow 800-3000 nodes seems really optimistic in real world use.

h3lix1:
This MR solves the problem of not having a large enough nodedb. I don't see 3000 nodes as being too much of a problem as the memory is all pre-allocated and leaves enough for everything else, but doing the large dump of the DB when connecting over bluetooth is a problem. I'm not sure why you're finding the wifi to be slow, as I can download the DB very quickly, but maybe my wifi is special.

In my previous message I'm trying to provide solutions to the large DB problem. We can have the node download the last 100 heard, and then do a fair share queue between node info updates and other incoming updates. Other options include doing a comparison of blocks of node IDs and share the ones that are missing. Any other thoughts?

The bay mesh currently cycles through 358 nodes every 3 hours, 500 every day, and currently up to 716 total over the last 8 days. I am guessing this will be towards 900 or 1000 at the end of this year, 2k at the end of next year. Add in some events, and 3k doesn't seem unreasonable as a goal to expand to.

I like this change, and I think it is the absolute best way to increase node counts for nodes with PSRAM while decreasing heap utilization. The problem is the communication between phone and mesh device also needs a refresh to support large data dumps.

I see this as the beginning of downloading much larger objects over time without needing a phone always attached. This, plus the ability for reliable message delivery, makes for the ability to transfer images, or other binary data, without impacting realtime communications. Large NodeDB just happens to be the first use-case that requires some kind of fair share mechanism.

NomDeTom:
@garthvh I thought there were recent optimisations to the app code, to bring the nodeDB over after initial handshake? If this is a way to slow the nodeDB rolling in a big mesh, it seems useful, if not advisable.

garthvh:
It was removed from Android  because of issues with the legacy connection process and is newish on iOS. This needs to be isolated to infrastructure roles initially and 3000 is just too high. What is the problem being solved here? For the client apps this creates a ton of issues to manage.

thebentern:
@h3lix1 I think what I would be interested in is how to gate this to infrastructure only roles like Router / Router Late, since those are not accessed as much on client apps, which as was pointed out becomes a headache on initial connection.

h3lix1:
@thebentern I don't think restricting this to router roles makes much sense, as the benefits for clients are great as well. As mentioned above, bay mesh nodedb expires clients faster than the 3 hour default nodeinfo cycle, causing issues for next-hop routing, keys, local node info, etc. This is also rather bad for MUI devices since more nodes will start showing up as unknown that should be known.

Unless we plan to move how things like encryption and keys stored/retired, a bad actor can inject the wrong key for any random user simply because the user rolled off the DB.

I don't know how to do this for the nrf52 nodes using a flash-based database. I'm willing to give anybody a bounty to get this working for those nodes as well as it's important, I just don't know how to handle all the gotchas. I.e. Flash corruption, write wear leveling, node performance, etc.

If the issue is the initial connection, let's find a way to support lazy loading or simply limit the DB dump size. (Or give clients the option how many to initially load?) But I feel this is very needed for client and router roles alike. If nothing else, it saves a significant amount of heap, even with 3000 nodes.

h3lix1:
With the latest bluetooth enhancements, NodeDB downloads much faster. Depending on the node, it downloads 200 nodes in 6-7 seconds in 2.7.13, compared to 18 seconds for 2.7.11.

src/DebugConfiguration.h
```diff
@@ -53,12 +53,42 @@ extern MemGet memGet;
 #define LOG_TRACE(...) SEGGER_RTT_printf(0, __VA_ARGS__)
 #else
 #if defined(DEBUG_PORT) && !defined(DEBUG_MUTE)
-#define LOG_DEBUG(...) DEBUG_PORT.log(MESHTASTIC_LOG_LEVEL_DEBUG, __VA_ARGS__)
```
fifieldt, line 56: Are the changes to this file necessary?

src/main.cpp
```diff
@@ -295,7 +295,7 @@ void printInfo()
 {
     LOG_INFO("S:B:%d,%s,%s,%s", HW_VENDOR, optstr(APP_VERSION), optstr(APP_ENV), optstr(APP_REPO));
 }
-#ifndef PIO_UNIT_TESTING
+#if !defined(PIO_UNIT_TESTING) || !(PIO_UNIT_TESTING)
```
fifieldt, line 302: Is this change needed for this patch?

src/mesh/NodeDB.cpp
```diff
@@ -24,6 +24,7 @@
 #include "modules/NeighborInfoModule.h"
```
fifieldt, line 24: suggest - consider the nodedb refactor work here: https://github.com/meshtastic/firmware/pull/8372 and see if there's a way to move in that kind of direction if possible, since nodedb is a large file :)

src/NodeStatus.h
```diff
@@ -14,16 +14,16 @@ class NodeStatus : public Status
     CallbackObserver<NodeStatus, const NodeStatus *> statusObserver =
         CallbackObserver<NodeStatus, const NodeStatus *>(this, &NodeStatus::updateStatus);
 
-    uint8_t numOnline = 0;
```
fifieldt: since uint8_t is such a severe limit, should we consider separately moving to uint16_t for these kind of things even before this patch is merged? Would help reduce the size of this PR, making it easier to review.

src/DebugConfiguration.h
h3lix1: They were, but I can take them out if it's a problem. During testing, the console would panic because consoleInit() failed due to memory constraints and caused a null pointer crash on startup. It was a pain to troubleshoot, and caused an issue where the 1200 baud firmware update would also fail.

This was a guard in case other tools allocated too much PSRAM/DRAM and caused a crash. It seemed to also make my USB console more stable somehow, but that is completely empirical not tested. Overall, it just seemed like a Good Thing.

src/NodeStatus.h
h3lix1: Placed this in part of its own change #8478, as a dependency for this change

src/main.cpp
h3lix1: Removed, not necessary for this patch

src/DebugConfiguration.h
fifieldt: Good to know. Perhaps put that in as a separate patch? again, the motivation is just to reduce the size of this one and make it more accessible to reviewers.

src/libtinylsm/README.md
```diff
@@ -0,0 +1,681 @@
+# Tiny-LSM for Meshtastic NodeDB
```
fifieldt: Is there an existing LSM library we can use instead?

This adds a huge amount of code that we would have to maintain.

h3lix1:
As expected, trying to get this to work with nrf52 is a massive lesson in frustration. The diff 26daa3d works well (great, even) for ESP32 without having to deal with trying to write a database for for a platform that can't handle it. Revertring this back to esp32-specific with psram and maybe someone will want it.

PR #8130, opened 2025-09-27
Revert cross-preset default-key bridging with UDP and disable UDP by default

thebentern:
It quickly became apparent based on conversations with users on the discord that this feature is extremely controversial and needs to be removed.
Reverts #7753

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -701,7 +701,7 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
 #ifdef USERPREFS_NETWORK_ENABLED_PROTOCOLS
     config.network.enabled_protocols = USERPREFS_NETWORK_ENABLED_PROTOCOLS;
 #else
-    config.network.enabled_protocols = 1;
+    config.network.enabled_protocols = 0;
 #endif
 #endif
 
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -483,35 +483,6 @@ DecodeState perhapsDecode(meshtastic_MeshPacket *p)
         }
     }
 
-#if HAS_UDP_MULTICAST
-    // Fallback: for UDP multicast, try default preset names with default PSK if normal channel match failed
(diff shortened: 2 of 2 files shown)


fifieldt (approved):
Alas, I've also seen these conversations. We need to talk more before we can have nice things :)

GUVWAF:
That PR was merely such that *if* you have nodes on two different presets with the default key, you will be able to **decrypt** those packets. Bridging happens automatically as UDP is a common medium (except when you set the rebroadcast mode to `LOCAL_ONLY`). So this doesn't actually decrease channel utilization, it merely makes the packets visible. 

IMO we should reconsider whether UDP should be enabled by default.

thebentern:
> That PR was merely such that _if_ you have nodes on two different presets with the default key, you will be able to **decrypt** those packets. Bridging happens automatically as UDP is a common medium (except when you set the rebroadcast mode to `LOCAL_ONLY`). So this doesn't actually decrease channel utilization, it merely makes the packets visible.
> 
> IMO we should reconsider whether UDP should be enabled by default.

Indeed. I added "default-key" to the title and removed the enabled by default.
We should also consider controls at individual channel levels perhaps re-using uplink / downlink when UDP is enabled.

GUVWAF:
Hmm, but why do we need to revert the decryption logic if it still creates the same traffic, just not visible?

Maybe it makes sense to put it back after #8134?

thebentern:
> Hmm, but why do we need to revert the decryption logic if it still creates the same traffic, just not visible?
> 
> Maybe it makes sense to put it back after #8134?

Yeah I think so. I'll cherry pick that change over.

garthvh:
Sending faster preset data to slower ones is bad.

GUVWAF:
Yes, it's definitely not a long-term solution, but mostly for meshes where the majority already moved to a faster preset and are trying to get the remaining nodes over. My point is, however, that currently people may already bridge it, but it happens to be invisible because of non-matching channel hashes.

VideoFX:
I would like to be able to bridge LF and MF for the purpose of migration.

PR #8132, opened 2025-09-27
Add WiFi/BT toggle for screenless devices

lkosson:
With this change, long-pressing a button on headless devices (such as Seeed Xiao ESP32) toggles between Wi-Fi and Bluetooth.

The goal is to provide an easy way to recover from misconfigured or outdated Wi-Fi settings, when serial USB connection to the device is not available. ESP32-based modules can't have BT and Wi-Fi enabled at the same time, and require a reboot to switch between them. This PR binds previously unused (on devices without screen) long-press event to perform such switch.

## 🤝 Attestations

  - [X] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [X] Other (Seeed Xiao ESP32-S3)


fifieldt:
I'll let people more familiar with the button code comment, however one item: you probably want to limit this to ESP32 devices. It doesn't have meaning for eg NRF52, which don't have wifi.

fifieldt:
Hi @lkosson , do you need help updating this patch?

PR #8134, opened 2025-09-27
Re-use channel Uplink / Downlink for UDP controls

thebentern:
For consideration


jp-bennett (approved):

jp-bennett:
Only thing to consider is whether we would eventually want to control the UDP uplink separately from the MQTT.

thebentern:
Agreed. I think in the long term we can split this control out into it's own levers

> Only thing to consider is whether we would eventually want to control the UDP uplink separately from the MQTT.

thebentern:
@GUVWAF added the default key fallback to this one

src/mesh/Router.cpp
```diff
@@ -355,13 +355,21 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
...
         packetPool.release(p_decoded);
     }
-
 #if HAS_UDP_MULTICAST
-    if (udpHandler && config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST) {
-        udpHandler->onSend(const_cast<meshtastic_MeshPacket *>(p));
+    // For packets that are already encrypted, we need to determine the channel from packet info
+    else if (udpHandler && config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST &&
+             isFromUs(p)) {
+        udpHandler->onSend(const_cast<meshtastic_MeshPacket *>(p), p->channel);
```
GUVWAF: Here `p->channel` is not an index, but still the hash since it's encrypted, so looking up the channel in `onSend()` will fail.
To support PKI, we probably need to allow up-/downlinking a hash of 0x00 if there is at least one channel with up-/downlink enabled just like we do for MQTT.

src/mesh/Router.cpp
```diff
@@ -355,13 +355,21 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
         if (moduleConfig.mqtt.enabled && isFromUs(p) && mqtt) {
             mqtt->onSend(*p, *p_decoded, chIndex);
         }
+#endif
+#if HAS_UDP_MULTICAST
+        // Only publish to UDP if we're the original transmitter of the packet
+        if (udpHandler && config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST &&
+            isFromUs(p)) {
```
GUVWAF: `isFromUs(p)` is new here and causes packets from other nodes not to be broadcasted via UDP. 
For MQTT this is here because there is also an `mqtt->onSend()` in `Router::handleReceived()` to allow for messages you decide not to rebroadcast to also end up on MQTT. If we want that for UDP as well, we'd need it there also, and guard it there with `!isFromUs()`.
If we only want to broadcast packets we'd also broadcast over LoRa, then we'd need to keep it as is and remove the `isFromUs(p)` here.

GUVWAF (changes requested):

src/mesh/udp/UdpMulticastHandler.h
```diff
@@ -53,6 +54,13 @@ class UdpMulticastHandler final
         LOG_DEBUG("Decoding MeshPacket from UDP len=%u", packetLength);
         bool isPacketDecoded = pb_decode_from_bytes(packet.data(), packetLength, &meshtastic_MeshPacket_msg, &mp);
         if (isPacketDecoded && router && mp.which_payload_variant == meshtastic_MeshPacket_encrypted_tag) {
+            // Check if downlink is enabled for this channel
+            auto &ch = channels.getByIndex(mp.channel);
```
GUVWAF: Also here `mp->channel` not an index yet.

src/mesh/Router.cpp
```diff
@@ -355,13 +355,55 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
...
 #endif
         packetPool.release(p_decoded);
     }
-
 #if HAS_UDP_MULTICAST
-    if (udpHandler && config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST) {
-        udpHandler->onSend(const_cast<meshtastic_MeshPacket *>(p));
+    // For packets that are already encrypted, we need to determine the channel from packet info
+    else if (udpHandler && config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST &&
+             isFromUs(p)) {
```
GUVWAF, line 392: `isFromUs(p)` is here still.

src/mesh/Router.cpp
```diff
@@ -355,13 +355,55 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
...
+        // Check if any channel has uplink enabled (for PKI support)
+        bool uplinkEnabled = false;
```
GUVWAF, line 394: Maybe instead of this bool we can use an `int8_t` to store the channel index where uplink is enabled, and -1 if none was found such that we can eliminate the same loop hereafter.

HansvanMeer:
We are using this on the moment to connect MF en LF.
It was running fine on 2.7.10. But on version 2.7.11 is isn't working anymore.

Talie5in:
> We are using this on the moment to connect MF en LF. It was running fine on 2.7.10. But on version 2.7.11 is isn't working anymore.

As per the notes on the 2.7.11 release
https://github.com/meshtastic/firmware/releases/tag/v2.7.11.ee68575

> [!IMPORTANT]
> Due to feedback received from users, UDP has been disabled by default and the default PSK decryption bridging logic has been removed for now until channel level controls can be added to a future release.

2.7.10 will be the release you can use to continue the bridging at this time.

jamesarich:
Digging up this grave - as I'm now exploring further transports past mqtt and udp. Is this an option we're still interested in pursuing?

I think it does make sense to include the up/down link configuration option per transport, but adds additional ux complexity to sort out.

PR #8137, opened 2025-09-27
Heltec MeshPocket: Adjust Battery urve

shortwavesurfer2009:
### ❌ (Please delete all these tips and replace them with your text) ❌

  to say "Hey, I think this idea X should be implemented and I'm starting work on it. My general plan is Y, any feedback
- Please do not submit untested code.
- If you do not have the affected hardware to test your code changes adequately against regressions, please indicate this, so that contributors and commnunity members can help test your changes.
- If your PR gets accepted you can request a "Contributor" role in the Meshtastic Discord

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


shortwavesurfer2009:
Sorry, if I'm supposed to add a label to this, I don't know how.

HarukiToreda:
when you say more accurate, what do you mean? Are you modifying it to reach 4.20V or are you modifying it to reach the true battery value which you can measure with a multimeter?

shortwavesurfer2009:
> when you say more accurate, what do you mean? Are you modifying it to reach 4.20V or are you modifying it to reach the true battery value which you can measure with a multimeter?

I am modifying it such that when the app shows that the charge is 0%, the node is near to death.

HarukiToreda:
> > when you say more accurate, what do you mean? Are you modifying it to reach 4.20V or are you modifying it to reach the true battery value which you can measure with a multimeter?
> 
> I am modifying it such that when the app shows that the charge is 0%, the node is near to death.

That’s not really answering the question I asked. When you fully charge the node with those values applied, what percentage and voltage level are you actually seeing?

The default values set by Heltec are based on the actual resistor tolerances used on the board, and even if you have the same model, those tolerances can vary from unit to unit. That’s why I’m asking what the goal of your adjustment is.

If the goal is to make the battery show 100% at full charge, that’s not quite right. Meshtastic considers a battery “fully charged” at around 4.19 V, but the actual charging IC stops charging once the battery hits about 4.15 V. The purpose of the ADC calibration is not to force the reading to say 100%, but to show the true voltage of the battery as accurately as possible.

I tested your values on my MeshPocket, and at full charge it showed 4.07 V at 92%, which is incorrect. That’s why it’s important to calibrate based on real voltage, not just the percentage you want to see.

shortwavesurfer2009:
> > > when you say more accurate, what do you mean? Are you modifying it to reach 4.20V or are you modifying it to reach the true battery value which you can measure with a multimeter?
> > 
> > 
> > I am modifying it such that when the app shows that the charge is 0%, the node is near to death.
> 
> That’s not really answering the question I asked. When you fully charge the node with those values applied, what percentage and voltage level are you actually seeing?
> 
> The default values set by Heltec are based on the actual resistor tolerances used on the board, and even if you have the same model, those tolerances can vary from unit to unit. That’s why I’m asking what the goal of your adjustment is.
> 
> If the goal is to make the battery show 100% at full charge, that’s not quite right. Meshtastic considers a battery “fully charged” at around 4.19 V, but the actual charging IC stops charging once the battery hits about 4.15 V. The purpose of the ADC calibration is not to force the reading to say 100%, but to show the true voltage of the battery as accurately as possible.
> 
> I tested your values on my MeshPocket, and at full charge it showed 4.07 V at 92%, which is incorrect. That’s why it’s important to calibrate based on real voltage, not just the percentage you want to see.

Okay, maybe it is a particular unit problem then, because if I charge it with the settings applied, it ends up at 4.24 volts and says PWD for 6 days on battery before it finally goes low enough to show 100%

If I run it with the default settings, it tells me that at full charge it is 4.48 volts.

HarukiToreda:
I have 3 more Meshpockets, once I charge them I will test to see how much the difference is between them.

HarukiToreda:
okay the last 3 I tested which are pretty new displayed 4.39V, 4.34 and 4.37V with the default ADC. With your modification they go to roughly 4.15V - 4.16V which is ideal since it's closer to the max voltage they can charge. I can only assume my oldest one may have been a early release when tolerances were much higher, I think I can safely say that's good proof this change can benefit the newer meshpockets.

HarukiToreda (approved):

mrekin:
@HarukiToreda @shortwavesurfer2009 
 Hi!
Which meshpocker version (5k or 10k) you use in your tests? And which fw variant?
5k and 10k have different battery curves in sources (looks like 5k has lihv battery and 10k regular lipo battery)
- but both varians use the same ADC params..

HarukiToreda:
> @HarukiToreda @shortwavesurfer2009 Hi! Which meshpocker version (5k or 10k) you use in your tests? And which fw variant? 5k and 10k have different battery curves in sources (looks like 5k has lihv battery and 10k regular lipo battery)
> 
> * but both varians use the same ADC params..

I used the 5000mAh one

mrekin:
> used the 5000mAh one

and *5000* variant/target?

HarukiToreda:
yes. 5000 variant

mrekin:
My 5000 pocker shows now (with patch applied) at full charge 4.19V and 87%.
I try to understood why battery curves different, is there problem with curve for 5000 variant or battery truely charges to 4.35V (both 5r and 10k have label 3.87V on the device)

HarukiToreda:
Not sure why it changed for you, change of the ADC multiplier should not affect the different curves in any way. Try opening a bug report so this can logged on a new PR. Refer this one as a possible reason for the change.

mrekin:
> Not sure why it changed for you

But if you use 5000 variant - you must have similar values as I understood, while it was build with 
-D HELTEC_MESH_POCKET_BATTERY_5000
Curves for 5 and 10 differs since first commit (and official flasher provide now only 10k variant for users).

PR #8144, opened 2025-09-28
Finish deprecating the Repeater role behavior

thebentern:
@NomDeTom


src/modules/AdminModule.cpp
GUVWAF, line 629: Should we fall back to `CLIENT` when someone tries to set `REPEATER` still? And maybe in general if it's still set to `REPEATER`, convert it to `CLIENT`?

thebentern: Good call. I think we did something similar for Router_client, so we can do the same here.

shalberd:
@thebentern Does this deprecation solve the issue of older, persisting nodes with device role Repeater sticking around?
I guess there is no way to keep older devices, be it with old firmware or in general role "Repeater" to still grab and forward traffic, right?
We see "Meshtastic ffff" in our traceroutes and our assumption is that it is a repeater ...

PR #8148, opened 2025-09-28
Bug / Send upgraded (duplicate) packets to phone if the queue removal failed.

h3lix1:
- Prevent duplicate delivery of hop-limit “upgraded” packets by marking them as seen when we couldn’t replace a queued TX copy, keeping the phone replay queue free of higher-hop duplicates (src/mesh/FloodingRouter.cpp:40-79).
- Apply the same safeguard in next-hop routing so routed traffic doesn’t leak upgraded duplicates to the phone (src/mesh/NextHopRouter.cpp:52-100).
- Solves #8149 

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Station G2
Heltec Lora32 V4


h3lix1:
Follow up to #5534 found a bug that will cause duplicates to phones in a case where the packet couldn't be removed from the queue.

h3lix1:
@GUVWAF @thebentern Found this one today. The change is simple if the queue replacement fails, re-set seenRecently for the rest of the flow.

PR #8166, opened 2025-09-30
Update MQTT root on lora change

### LINKED ISSUE #8139: [Bug]: Default MQTT root topic does not change when LoRa region is changed
### Category

Other

### Hardware

Not Applicable

- [ ] InkHUD ePaper

### Firmware Version

Master

### Description

When changing region, the root topic needs to be manually deleted to get reset to the appropriate root topic for the new region.

Without this manual step, packets get cross-uplinked to the previous region’s root topic. 

In Southeast Asia, SG_923/MY_919/TH share land borders and many commute frequently across them. 

Having to remember to manually clear the MQTT setting gets tedious pretty quickly. Do consider also having the nodedb cleared when changing regions as the previous region’s nodes all remain behind too. 

### Relevant log output

```Shell

```

ford-jones:
Resolves #8139 

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -770,6 +770,15 @@ void AdminModule::handleSetConfig(const meshtastic_Config &c)
                 changes = SEGMENT_CONFIG | SEGMENT_MODULECONFIG;
             }
         }
+        if (config.lora.region != myRegion->code) {
+            //  Region has changed so check whether there is a regulatory one we should be using instead.
+            //  Additionally as a side-effect, assume a new value under myRegion
+            initRegion();
+
+            //  subscribe to appropriate MQTT root topic for this region
+            sprintf(moduleConfig.mqtt.root, "%s/%s", default_mqtt_root, myRegion->name);
+            changes = SEGMENT_CONFIG | SEGMENT_MODULECONFIG;
+        }
         break;
     case meshtastic_Config_bluetooth_tag:
         LOG_INFO("Set config: Bluetooth");


ford-jones:
@ndoo suggested that we simultaneously clear the nodedb. 
What do others think? :thinking:

ndoo:
I think one regression here is that if the user has a custom MQTT topic (perhaps some specific implementation on their own MQTT server), it would get reset by this change. May be inconveniencing that group of users for the convenience of this group of users. 

What I assume would fix it is to generate the root topic with the old region and do a string comparison against the new generated root topic

ford-jones:
I was considering checking `USERPREFS_MQTT_ROOT_TOPIC` as done in NodeDB.cpp here for that reason if you think it would suffice?

ford-jones:
@ndoo this should do it ^ :slightly_smiling_face: 
Still need to apply this logic to the menu handler then will pull this out of draft

ndoo:
> @ndoo this should do it ^ :slightly_smiling_face: 
> 
> Still need to apply this logic to the menu handler then will pull this out of draft

Thanks!

Hmm, I may be wrong here, but IIRC even the default generated topic will get saved to userPrefs so this block of code will functionally always enter the first if condition, hence necessitating doing an actual string comparison to determine if the topic is at default. 

I.e. the userpref will only be empty if region is UNSET

ndoo:
Sorry, I would love to test but am away from a device for 2-3 more days due to a work trip.

ford-jones:
Ah, you're 100% correct :heavy_check_mark: 
I've just gotten back home to test this and you're right about the preference being overwritten, I'll push something else up shortly :satellite:

ndoo:
No worries, thanks for testing it.

PR #8173, opened 2025-09-30
T-Rex Support

jp-bennett:


jp-bennett:
Copilot's comments were nitpicks, but still valid. Fixed.

src/gps/RTC.cpp
```diff
@@ -109,6 +109,25 @@ RTCSetResult readFromRTC()
         }
         return RTCSetResultSuccess;
     }
+#elif defined(RX8130CE_RTC)
```
fifieldt, line 112: Is a BUILD_EPOCH check needed here?

jp-bennett: Yes, it probably should be checked..

vidplace7:
:t-rex:

PR #8182, opened 2025-10-02
Multi message storage

HarukiToreda:
### Overview  
This PR introduces **multi-message storage and threaded conversation view** support to BaseUI, improving how messages are displayed, stored, and preserved across sessions.  

### Features  

- **Message Storage**
  - Stores the last **20 messages** in RAM.  
  - Persists the last 20 messages to flash on properly initiated reboot/shutdown  
    (not hard power cut or reset button press).  
  - Automatically restores saved messages from flash into RAM on boot.  
  - Added debug logs showing the **exact space used when saving to flash** and the  
    **space restored on boot**.  

- **Conversation Threads**
  - Messages are displayed **newest to oldest** with timestamps in an auto-scrollable thread view.  
  - Sent messages are labeled as **“Me”**.  
  - Incoming messages are automatically sorted into the correct thread (DM or Channel broadcast).  
  - Auto-focuses on the last active conversation when a new message arrives.  
  - Uses boot time counter if no RTC is available, and once valid time is obtained,  
    messages are updated and written with the corrected timestamp for persistence.  
  - **Conversation Menu Options:**
    - Switch between conversation views (**Channels, DMs, or All**).  
    - **Reply** directly to the currently focused conversation.  
    - **Dismiss Last**: remove the oldest message in the current conversation.  
    - **Dismiss All**: clear all messages (when in *View All*).  

- **Message Status Indicators (for sent messages)**
  - ✅ **Checkmark** = confirmed ACK.  
  - ❌ **X** = failed or timed-out.  
  - ⚠️ **Exclamation mark** = relayed but no ACK.  

- **Client Integration**
  - Messages created on other clients (phone/web) are captured and added into the correct thread.  

### Additional Changes  
- Removed legacy message handling from **Screen.cpp**.  
- Cleaned up **CannedMessageModule** by removing legacy message sending and temporary message logic.  
- Fixed **Emote screen rendering** issues.  
- Added autofocus to message screen after sending a new message.  
- Favorite Node screens jump to conversation with action menu option.
- Nodelist manual up and down scrolling
- Message manual up and down scrolling.
- Rework isHighResolution to be an enum called ScreenResolution (https://github.com/meshtastic/firmware/pull/8182/changes/a0b4e8270eb597331fdeb85a41b5b695c0c8baca)
- Reduce usage of Unit C6L macro guards (https://github.com/meshtastic/firmware/pull/8182/changes/6dfa5329a3d5488cac219063a856d38c7dff76b3)


Xaositek (approved):

src/modules/CannedMessageModule.cpp
```diff
@@ -195,13 +200,13 @@ void CannedMessageModule::drawHeader(OLEDDisplay *display, int16_t x, int16_t y,
...
         if (this->dest == NODENUM_BROADCAST) {
-            display->drawStringf(x, y, buffer, "To: Broadcast@%s", channels.getName(this->channel));
+            display->drawStringf(x, y, buffer, "To: #%s", channels.getName(this->channel));
         } else {
             display->drawStringf(x, y, buffer, "To: %s", getNodeName(this->dest));
         }
     } else {
         if (this->dest == NODENUM_BROADCAST) {
-            display->drawStringf(x, y, buffer, "To: Broadc@%.5s", channels.getName(this->channel));
+            display->drawStringf(x, y, buffer, "To: #%.5s", channels.getName(this->channel));
```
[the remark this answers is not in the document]
HarukiToreda: We are no longer labeling channels as @, now we use #. The other will be used for nodes when needed.

src/modules/TextMessageModule.cpp
```diff
@@ -15,14 +17,21 @@ ProcessMessage TextMessageModule::handleReceived(const meshtastic_MeshPacket &mp
 #endif
 
     // We only store/display messages destined for us.
-    // Keep a copy of the most recent text message.
     devicestate.rx_text_message = mp;
     devicestate.has_rx_text_message = true;
+#if HAS_SCREEN
```
mverch67, line 24: I think when MUI is active on the device (config.display.displaymode == meshtastic_Config_DisplayConfig_DisplayMode_COLOR) it should not do any of this message storing as MUI is already handling this.
For the other standalone scenario with a connectd MUI device over serial I guess it would make sense to have means to enable/disable the message store at runtime.

HarukiToreda: good point

HarukiToreda:
big change, we switched from dynamic std::string storage to fixed-size char[] to reduce flash build size

bobricius:
Is it possible, for example, to add UP-DOWN arrows for devices that have a keyboard to switch between messages or scroll through a list?
Thank you very much for this feature, it brings standalone devices up to the level of those with the new UI.

Xaositek:
> Is it possible, for example, to add UP-DOWN arrows for devices that have a keyboard to switch between messages or scroll through a list? Thank you very much for this feature, it brings standalone devices up to the level of those with the new UI.

This will be reviewed and performed in a separate initiative and update to BaseUI

ford-jones:
This is great!

> ⚠️ Exclamation mark = relayed but no ACK.

May I suggest the usage of a different emoji for this particular delivery state? maybe :signal_strength: ?
Failure to retrieve an ack from the final destination is very common and is no indicator of whether it actually _reached_ them or not. 

I think the exclamation mark is a bit loaded, i.e. it _seems like a bad thing has happened._

HarukiToreda:
yeah unfortunately the ack part of multimessage is a side adjustment, it will need to be expanded to be more advanced in the future, for example, if you receive an ack from the first node, we currently stop listening to any others, what happens if that ack came from one of your other home nodes? it gives you a sense that something happened when it could mean nothing. Maybe a message telling you how many nodes first relayed your message in the future could be a possible expansion, but that PR should not focus on that, that seems like sometihng that should be addressed later instead.

this is what it currently looks like

benjamincarmel:
Is any testing for this being done on the T-Pager?

HarukiToreda:
> Is any testing for this being done on the T-Pager?

it has been fully tested and working on all devices with screens, T-Pager included.

src/modules/TextMessageModule.cpp
Xaositek: Pushed a fix for this

lbibass:
tested this on my Wio Tracker L1 and it worked great! Fantastic work!

PR #8187, opened 2025-10-02
Added support for SugarCube device

igorka48:
This PR adds support for the SugarCube device in Meshtastic.
The SugarCube is a small ESP32-based board, originally designed for other purposes, but with this integration it can now be repurposed as a fully functional Meshtastic node.

With these changes:

Users can flash Meshtastic firmware directly onto a SugarCube.

A device that might otherwise be unused can be turned into a working mesh node.

Motivation

Many people already have SugarCube boards on hand. This update extends the usability of existing hardware, making it easier to grow the Meshtastic network without buying new devices.


HarukiToreda:
Could you share a link with more information about this device? I wasn’t able to find any details on it or where it is being sold. Also, does this board come with LoRa built in, or is it something that requires a DIY modification?

igorka48:
> Could you share a link with more information about this device? I wasn’t able to find any details on it or where it is being sold. Also, does this board come with LoRa built in, or is it something that requires a DIY modification?

Added link and photo to description. 

This device was very popular in Ukraine.
Tens of thousands were sold, and an unknown number were assembled by volunteers.

It is based on the Lilygo TTGO LoRa32 V2.1_1.6 915Mhz.

Now it is rapidly losing popularity for its original purpose and can be converted into a Meshtastic network in case of global power outages.

mverch67:
Except the button pin all the definitions are identical to [tlora_v2_1_16](https://github.com/meshtastic/firmware/tree/develop/variants/esp32/tlora_v2_1_16).

In principle you could achieve the same with a minimal addition to above platformio.ini:
```
[env:sugarcube]
extends = tlora-v2-1-1_6
build_flags = ${tlora-v2-1-1_6.build_flags} -DBUTTON_PIN=0
```

igorka48:
> Except the button pin all the definitions are identical to [tlora_v2_1_16](https://github.com/meshtastic/firmware/tree/develop/variants/esp32/tlora_v2_1_16).
> 
> In principle you could achieve the same with a minimal addition to above platformio.ini:
> 
> ```
> [env:sugarcube]
> extends = tlora-v2-1-1_6
> build_flags = ${tlora-v2-1-1_6.build_flags} -DBUTTON_PIN=0
> ```

Button and LED. 
SugarCube has led pin paired with buzzer. And using LED create noise. So, I've disabled a led usage

variants/esp32/sugarcube/platformio.ini
```diff
@@ -0,0 +1,8 @@
+[env:sugarcube]
+extends = esp32_base
+board = ttgo-lora32-v21
```
vidplace7:
```suggestion
board = ttgo-lora32-v21
board_level = extra
```
`board_level` must be set to `extra` :+1:

vidplace7 (changes requested):

igorka48:
> Except the button pin all the definitions are identical to [tlora_v2_1_16](https://github.com/meshtastic/firmware/tree/develop/variants/esp32/tlora_v2_1_16).
> 
> In principle you could achieve the same with a minimal addition to above platformio.ini:
> 
> ```
> [env:sugarcube]
> extends = tlora-v2-1-1_6
> build_flags = ${tlora-v2-1-1_6.build_flags} -DBUTTON_PIN=0
> ```

Very simple and clean approach. 

`
> [env:sugarcube]
> extends = env:tlora-v2-1-1_6
> board_level = extra
> build_flags = 
>   ${env:tlora-v2-1-1_6.build_flags} 
>   -DBUTTON_PIN=0 
>   -DPIN_BUZZER=25
>   -ULED_PIN
`

But I'm strugling with #undef LED_PIN
`  -ULED_PIN`
doesn't work

mverch67:
> But I'm strugling with #undef LED_PIN
> ` -ULED_PIN`
> doesn't work

What you could do is put this into _platformio.ini_
`
build_flags = ${tlora-v2-1-1_6.build_flags} -DBUTTON_PIN=0 -DLED_PIN=-1`

and in _variant.h_ you check `LED_PIN`
```c++
#if defined(LED_PIN) && LED_PIN == -1
#undef LED_PIN
#else
#define LED_PIN 25
#endif
```

mverch67:
Actually the idea was to put this into the single file `variants/esp32/tlora_v2_1_16/platformio.ini` without the new extra directory. This makes sense because SugarCube is derived from tlora v2 so it fits best directly within its platformio.ini file. Having a dependency between the two variant folders as it is currently committed feels a bit awkward as it never has been done this way and would be challenging once there are changes in tlora v2 (which could then possibly break the SugarCube firmware).

fifieldt:
@igorka48 , could you put the contents of ‎variants/esp32/sugarcube/platformio.ini‎ into variants/esp32/tlora_v2_1_16/platformio.ini

It should work just as well, and be a bit cleaner for the maintainers.

mverch67 (approved):

fifieldt (approved):

PR #8194, opened 2025-10-03
STM32: EByte E77-MBL serial pin update

vidplace7:
Fix EByte E77-MBL Serial pin definitions to match the datasheet (identical to wio-e5).

This fixes current build failures with this target
```
firmware/src/mesh/Channels.cpp:330: undefined reference to `Serial2'
```


vidplace7:
@Stary2001

fifieldt (approved):

Stary2001:
Sorry for not getting to this until now, my weekend was a bit more eventful than planned!
The schematic from https://www.cdebyte.com/products/E77-900MBL-01/4#Downloads says that the USB to UART converter is connected to PA2/3 (which can be one of lpuart1, uart2), so this won't work as is.

PR #8200, opened 2025-10-04
wip

ammarkkhan:
"Add MQTT policy override fields to ChannelSettings"

"Extend ChannelSettings proto with mqtt_allow and mqtt_ignore enums"

"Protobuf schema update: per-channel MQTT control"


thebentern:
This change would require actual protobufs support

PR #8202, opened 2025-10-04
Clear out user.id except for sending to phone

### LINKED ISSUE #8170: [Bug]: Firmware allows the user-id to be set to an arbitrary string
### Category

Other

### Hardware

Not Applicable

- [ ] InkHUD ePaper

### Firmware Version

2.7.3

### Description

The firmware allows the user-id to be set via protobufs to any arbitrary string.

The convention is that the user-id is set to the hex value of the node-id.

Should we check if the user-id is a valid hex string?

### Relevant log output

```Shell
How it can be set in current firmware:

id: "!kronos"
long_name: "kronos"
short_name: "BB07"

How it is conventionally set:

Import Time
10:31:48 AM - 09-30-2025
packet
from: 1128103100
to: 4294967295
channel: 24
decoded {
  portnum: NODEINFO_APP
  bitfield: 0
}
id: 3323496629
rx_time: 1759253508
rx_snr: -16.25
hop_limit: 5
rx_rssi: -107
hop_start: 7
relay_node: 12
transport_mechanism: TRANSPORT_LORA
payload
id: "!433d7cbc"
long_name: "kronos"
short_name: "BB07"
macaddr: "H\312C=|\274"
hw_model: HELTEC_WIRELESS_TRACKER
public_key: "\333\030z\300x|\037\222\351P \210N\317\372:\270\302\346\236\225\266\335\314y\242\271\007\002\251\317\006"
is_unmessagable: false
```

thebentern:
Closes #8170 

Avoid storing user.id as anything but a null terminated empty string and send to client as a generated value

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1601,7 +1601,8 @@ void loop()
 
     service->loop();
 #if !MESHTASTIC_EXCLUDE_INPUTBROKER && defined(HAS_FREE_RTOS)
-    inputBroker->processInputEventQueue();
+    if (inputBroker)
+        inputBroker->processInputEventQueue();
 #endif
 #if defined(LGFX_SDL)
     if (screen) {


src/mesh/NodeDB.cpp
```diff
@@ -56,10 +56,6 @@
 #include <WiFiOTA.h>
 #endif
 
-// stringify
-#define xstr(s) str(s)
-#define str(s) #s
-
```
jp-bennett, line 62: Did this get moved somewhere? This is needed to use the build epoch as a filename.

jp-bennett: Ah, it's already in configuration.h.

h3lix1:
#8228 found with this..

PR #8217, opened 2025-10-05
allows WiFi to attempt reconnect more than once as defined, default 6 attempts.

sandeepkumar0153:
simple reconnect attempts to connect a predefined number of times to the WiFi network. 
there is an issue with raspberry pico w, where it may not connect in the first attempt after a restart.
works for all other devices as well.

  - [x] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [x] LilyGo T-Beam
  - [x] RAK WisBlock 4631
  - [x] Nrf52840 diy
  - [x] Rpi pico w

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -66,6 +66,20 @@ inline void onReceiveProto(char *topic, byte *payload, size_t length)
           (strcmp(e.channel_id, channels.getGlobalId(ch.index)) == 0 && ch.settings.downlink_enabled))) {
         return;
     }
+
+    bool anyChannelHasDownlink = false;
+    size_t numChan = channels.getNumChannels();
+    for (size_t i = 0; i < numChan; ++i) {
+        const auto &c = channels.getByIndex(i);
+        if (c.settings.downlink_enabled) {
+            anyChannelHasDownlink = true;
+            break;
+        }
+    }
+
+    if (strcmp(e.channel_id, "PKI") == 0 && !anyChannelHasDownlink) {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


sandeepkumar0153:
refactoring code slightly, will update once done.

sandeepkumar0153:
updated the code, added some debug log info content.
tested on pico w and nrf52840 diy. 
1. no wifi AP max reconnect attempt scenario.
2. wifi ap connect after 3 attempts caused by wifi router reboot.

sandeepkumar0153:
reverted variant.h to default config for two devices to match meshtastic repo.

src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -49,6 +49,15 @@ char ourHost[16];
 // To replace blocking wifi connect delay with a non-blocking sleep
 static unsigned long wifiReconnectStartMillis = 0;
 static bool wifiReconnectPending = false;
+static unsigned char wifiConnectAttempts = 0;
+
+#ifndef MaxWifiConnectionAttempts
+#define  MaxWifiConnectionAttempts 6
```
thebentern: Macros should be CAPS_LOCK

src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -49,6 +49,15 @@ char ourHost[16];
...
+#endif
+
+#ifndef wifiReconnectGap
+#define  wifiReconnectGap 10000
```
thebentern: Macros should be CAPS_LOCK

PR #8223, opened 2025-10-05
update build-* to also output binaries to *-update.bin/uf2

jp-bennett:
This is in preparation for supporting the factory reset firmware for non-esp32 targets.

### DIFF
--- a/bin/build-nrf52.sh
+++ b/bin/build-nrf52.sh
@@ -47,6 +47,8 @@ else
 	cp bin/*.uf2 $OUTDIR
 fi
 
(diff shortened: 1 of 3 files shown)


fifieldt (approved):

vidplace7:
Conflicts with #8248 -- chat with me @jp-bennett 

Under the new naming scheme after that PR is merged, we should instead name the new factory reset .uf2 to `$basename.factory.uf2`, while leaving the "update" uf2 untouched (avoiding rename).

jp-bennett:
@vidplace7 Sure, if #8248 tests out and gets merged, I'll be glad to close this one and follow that naming scheme

PR #8239, opened 2025-10-07
Actions: Simplify matrices, cleanup build_one_* (#8218)

fifieldt:
cherry pick from develop


fifieldt:
Checked and this is indeed the same. Giving it a shot.

PR #8240, opened 2025-10-07
Do slightly better at threading the search for GPS hardware

jp-bennett:
There's room to do more of this, and eliminate more delay()s. Seems like a decent start.


jp-bennett:
I'm testing this on a thinknode M5 with the L76k and the R1-neo  with the ATGM336H. So far both seem to be behaving as expected.

PR #8253, opened 2025-10-08
Actions: CI docker with a fancy matrix

vidplace7:
- Simplify matrix for docker builds
- Rip out all the docker builds in `build_one_*` -- they aren't published anywhere or used for testing, it's just a waste of electricity! :zap: 

Tested in my fork :+1:

### DIFF
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -27,7 +27,6 @@ on:
 
 jobs:
   setup:
-    if: github.repository == 'meshtastic/firmware'
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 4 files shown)


.github/workflows/main_matrix.yml
```diff
@@ -27,7 +27,6 @@ on:
 
 jobs:
   setup:
-    if: github.repository == 'meshtastic/firmware'
```
fifieldt, line 30: Personally, I find the workflows from upstream running on my fork annoying ... disabling them is a hassle. Others may share this view, or not ...

.github/workflows/main_matrix.yml
```diff
@@ -42,10 +41,6 @@ jobs:
           python-version: 3.x
           cache: pip
       - run: pip install -U platformio
-      - name: Uncomment build epoch
```
fifieldt, line 45: please do make sure the build epoch is properly set :)

.github/workflows/main_matrix.yml
vidplace7: That leaves us with no way to test in forks at all :sweat_smile:
We have already gated the check/gather_artifact/release pieces so that they will be skipped on forks.

.github/workflows/main_matrix.yml
vidplace7: This was moved to `platformio-custom.py` a while ago and never got cleaned up here :) Just playing janitor

PR #8263, opened 2025-10-08
Decrease nodedb load delays over bluetooth, include prefetch, remove 20ms polling loop

h3lix1:
- Refactored PhoneAPI node delivery so we keep a small prefetched queue of NodeInfo entries, clearing and refilling it across sessions; BLE reads now pop from this cache, immediately re-populate, and continue logging each node while moving on to the next.
- New prefetch path and consumption behaviour, and replaced  meshtastic_NodeInfo_init_default assignment with value-initialisation (sic.)
- Switched the FromRadio characteristic to support notifications, moved the read callback off the old 20 ms polling loop (!!)

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Heltec (Lora32) V4


src/nimble/NimbleBluetooth.cpp
```diff
@@ -353,15 +364,18 @@ void NimbleBluetooth::setupService()
     // Define the characteristics that the app is looking for
     if (config.bluetooth.mode == meshtastic_Config_BluetoothConfig_PairingMode_NO_PIN) {
         ToRadioCharacteristic = bleService->createCharacteristic(TORADIO_UUID, NIMBLE_PROPERTY::WRITE);
-        FromRadioCharacteristic = bleService->createCharacteristic(FROMRADIO_UUID, NIMBLE_PROPERTY::READ);
+        // Allow phones that subscribe to notifications to stream data without polling reads.
+        FromRadioCharacteristic =
+            bleService->createCharacteristic(FROMRADIO_UUID, NIMBLE_PROPERTY::READ | NIMBLE_PROPERTY::NOTIFY);
```
thebentern, line 369: I'm not sure why we need notify on fromRadio. We use the fromNum characteristic to notify clients there is a new fromRadio to Read. This actually might require some definition changes to take place on some clients

h3lix1: Require? Nah.

Giving options.. For when clients can support streaming. It fails successfully otherwise.

h3lix1:
No longer needed :)

PR #8281, opened 2025-10-08
NimBLE speedup

thebentern:
This speeds up the process quite a bit when syncing a large NodeDB. @jp-bennett I wanted to get your input because I think the delay(20) state polling patch had your name on it

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -120,7 +120,7 @@ lib_deps =
 [device-ui_base]
 lib_deps =
 	# renovate: datasource=git-refs depName=meshtastic/device-ui packageName=https://github.com/meshtastic/device-ui gitBranch=master
-	https://github.com/meshtastic/device-ui/archive/e564d78ae1a7e9a225aaf4a73b1cb84c549f510f.zip
+	https://github.com/meshtastic/device-ui/archive/6d8cc228298a1ecd9913aed757187e9527c1facc.zip
 
 ; Common libs for environmental measurements in telemetry module
 [environmental_base]
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -33,6 +33,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "pcf8563.h"
 #endif
 
+/* Offer chance for variant-specific defines */
+#include "variant.h"
+
 // -----------------------------------------------------------------------------
 // Version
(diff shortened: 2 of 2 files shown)


jp-bennett:
Seems to make sense. Just need to make sure we test it out well with both iOS and Android. Seems like the original code ended up the way it did to make both platforms happy.

compumike:
While I haven't tested the changes here, and think a NimBLE speedup would be great, my understanding was that the goal of the original code's `phoneWants` / `hasChecked` signaling and polling was a way to avoid calling `getFromRadio` from within the `onRead` callback. And this may have been important because the `onRead` callback is being run in a separate FreeRTOS task (and possibly concurrently on a different core) from the main application loop. Am I missing something?

thebentern:
> While I haven't tested the changes here, and think a NimBLE speedup would be great, my understanding was that the goal of the original code's `phoneWants` / `hasChecked` signaling and polling was a way to avoid calling `getFromRadio` from within the `onRead` callback. And this may have been important because the `onRead` callback is being run in a separate FreeRTOS task (and possibly concurrently on a different core) from the main application loop. Am I missing something?

Shouldn't the mutex lock on nimble prevent that condition from occuring?

compumike:
@thebentern The `nimble_mutex` prevents the `NimbleBluetoothFromRadioCallback::onRead` and `BluetoothPhoneAPI::runOnce` from running at the same time, but that's it.

- `BluetoothPhoneAPI::runOnce` runs in the main loop, in the main FreeRTOS task.
- `NimbleBluetoothFromRadioCallback::onRead` runs in NimBLE's FreeRTOS task, which may be on a separate core.

If `onRead` now directly calls `getFromRadio`, that means the `getFromRadio` call is happening in the NimBLE FreeRTOS task, not the main task.

This means that `PhoneAPI::getFromRadio` can now run concurrently with all the other main application code!

All of the nice assumptions that the rest of the codebase makes about the code running mostly linearly are now in question. Is all of `PhoneAPI::getFromRadio`, and any calls it makes (such as `PhoneAPI::available`, `NodeDB::readNextMeshNode`, `onNowHasData`, any other downstream calls) safe for concurrent access?

Maybe they are, I don't know. It's just a big change in assumptions from when `getFromRadio` was only called from within the `runOnce`.

PR #7697, opened 2025-08-20
Update platformio/espressif32 to v6.12.0

vidplace7:
Closes #7809 

Update platformio/espressif32 to v6.12.0
https://github.com/platformio/platform-espressif32/releases/tag/v6.12.0

Also adds the `intelhex` dependency if needed to workaround a PlatformIO bug with fetching this automatically.
See: https://github.com/platformio/platform-espressif32/issues/1632

PR #8285, opened 2025-10-09
Reset ESP32S3 after flash using watchdog

macegr:
ESP32S3-based devices with direct USB connections currently do not automatically reset when esptool flashing has completed. This requires users to press a physical reset button or power-cycle a device before the new firmware will run, which can pose usability or confusion issues for installations where such actions are not straightforward or expected.

**Dependency:** Once https://github.com/meshtastic/firmware/pull/7697 is merged, the newer version of esptool included will support a watchdog-based reset method. This has been confirmed to work on a Heltec WiFi LoRa 32 V4.

### DIFF
--- a/arch/esp32/esp32s3.ini
+++ b/arch/esp32/esp32s3.ini
@@ -3,3 +3,4 @@ extends = esp32_base
 custom_esp32_kind = esp32s3
 
 monitor_speed = 115200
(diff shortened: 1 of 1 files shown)


macegr:
Turns out #7697 resolves this issue automatically due to upstream detection in esptool!

Issue #8226, opened 2025-10-06
[Feature Request]: When NodeDB is reset, keep favorite nodes


pedjas:
### Platform

Cross-Platform

### Description

When user resets NodeDB, information on all nodes is reset. There is need to keep some nodes intact, like favorite nodes. The most often reason for resetting is to clean up cluttered node list of unknown and not needed nodes. But there are always nodes that should be preserved - ones we do have interest of.

It would be good to have choice to preserve favorite nodes from deletion.

radonslide:
Good idea. 
With the Client_Base role, it makes sense to be able to keep the favorites that have been set, if possible.

shalberd:
>  With the Client_Base role, it makes sense to be able to keep the favorites that have been set, if possible.

not just with the client_base role, but also with Router and Router_Late new ability to not decrease hop count between two nodes Client_Base/Router/Router_Late when marked as favorite from_node. New in 2.7.11 https://github.com/meshtastic/firmware/pull/7992
By the way, in alpine conditions, I'd only do this for nodes at each end of a long valley at best, or a small additional hop around a sharp bend, to increase range, but definitely not for omni-directionally-accessible  Client_Base/Router/Router_Late nodes.

Though I have decided to just print out those bi-directional favorites and re-import them quickly via QR code (via Android app) upon factory reset when doing a clean node upgrade from 2.6 to 2.7.

jkrauska:
I'd like the ability to export/import nodeDBs and/or single node imports.  this would let me pre-seed nodes when swapping radio hardware.

shalberd:
> export/import nodeDBs

yeah, that'd be nice

> and/or single node imports

@jkrauska on the Android App 2.7.1 and since firmware 2.6.11, I can already per-node say "Share contact" under Actions of that node. That generates a QR code that I print out / make a sceenshot of, and then via Android app connected to  a new node, I can add these CQ code export nodes via the Smartphone camera and the android app from the printout.
Not electronically, but working for pre-seeding nodes quickly node per node, especially for nodes far away or not sending their node info very often.

![Image](https://github.com/user-attachments/assets/38735ed1-d05d-446d-9778-50889284fbab)

jkrauska:
> > export/import nodeDBs
> 
> yeah, that'd be nice
> 
> > and/or single node imports
> 
[...]

PR #8292, opened 2025-10-10
Persist favourites on NodeDB reset

ford-jones:
Resolves: #8226
Introduces the option to preserve favourited nodes following a reset of the NodeDB. 

Does this by default for nodes set to `ROUTER`, `ROUTER_LATE` or `CLIENT_BASE`.

Protobuf changes: https://github.com/meshtastic/protobufs/pull/791

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [x] RAK WisBlock 4631
  - [x] Seeed XIAO NRF52840 / wio-sx-1262


shalberd:
> Does this by default for nodes set to ROUTER, ROUTER_LATE or CLIENT_BASE.

I think having this option across the board, e.g. also for Client and Client_Mute Nodes, would be nice.

The reason there is different on e.h. nrf52 based devices: favoriting certain other mesh nodes on my client or client_mute type node keeps them from being moved off the node DB when there are a large number of nodes, say 300, as we have in Switzerland. Giving them a persistency during operations that I would like to maintain, at least optionally, after a node DB reset, or even a factory reset, for that matter.
So it'd be great if I could keep my favorites as well on Client and Client_Mute device role nodes.
Currently, I am exporting my favorite nodes node by node via a QR code and then re-import them (Android, 2.7 Firmware, 2.7.1 app version).

Thank you for working on this feature.

src/mesh/generated/meshtastic/config.pb.h
```diff
@@ -723,7 +725,7 @@ extern "C" {
 
 /* Initializer values for message structs */
 #define meshtastic_Config_init_default           {0, {meshtastic_Config_DeviceConfig_init_default}}
-#define meshtastic_Config_DeviceConfig_init_default {_meshtastic_Config_DeviceConfig_Role_MIN, 0, 0, 0, _meshtastic_Config_DeviceConfig_RebroadcastMode_MIN, 0, 0, 0, 0, "", 0, _meshtastic_Config_DeviceConfig_BuzzerMode_MIN}
+#define meshtastic_Config_DeviceConfig_init_default {_meshtastic_Config_DeviceConfig_Role_MIN, 0, 0, 0, _meshtastic_Config_DeviceConfig_RebroadcastMode_MIN, 0, 0, 0, 0, "", 0, _meshtastic_Config_DeviceConfig_BuzzerMode_MIN, 0}
```
shalberd: 0 means preserve favorites false, correct?

So later on, e.g. in the Web client or the Android and IOS apps, there could be a checkmark type UI element that allows setting preserve favorites to true / 1, I think.

shalberd:
persisting / preserving nodes set to be ignored / on the ignore list for rebroadcast would be great feature, too, especially for ROUTER, ROUTER_LATE, CLIENT_BASE, but that is something for a different story :-)

src/mesh/generated/meshtastic/config.pb.h
ford-jones: yes :+1:

src/mesh/generated/meshtastic/config.pb.h
```diff
@@ -357,6 +357,8 @@ typedef struct _meshtastic_Config_DeviceConfig {
     /* Controls buzzer behavior for audio feedback
  Defaults to ENABLED */
     meshtastic_Config_DeviceConfig_BuzzerMode buzzer_mode;
+    /* Tell the node to persist favourited nodes following a NodeDB reset. */
+    bool preserve_favourites;
```
shalberd: small nit, no biggie, as I am in Europe, too:
I think elsewhere, the spelling is favorites, American/Canadian English.

Or maybe just American English variety, in any case, it seems to be "favorites" elsewhere in the codebase.
I have no favourite way ;-)

ford-jones:
> I think having this option across the board, e.g. also for Client and Client_Mute Nodes, would be nice.

@shalberd If you check out the protobuf, I created an option for doing precicely that. I.e. `config.device.preserve_favourites`. This would allow you to persist them through a NodeDB reset even as a `CLIENT` or any other role. 

I like @thebentern's idea though, having this as an additional optional parameter for `reset-nodedb`.

shalberd:
> I like @thebentern's idea though, having this as an additional optional parameter for reset-nodedb.

Yes, me, too

src/mesh/NodeDB.cpp
```diff
@@ -402,6 +402,11 @@ NodeDB::NodeDB()
         config.position.gps_mode = meshtastic_Config_PositionConfig_GpsMode_ENABLED;
         config.position.gps_enabled = 0;
     }
+    if (config.device.preserve_favourites ||
```
shalberd: think of removing the device role qualifiers here

shalberd: ah, I did not refresh page, all good, cool, thank you. You can close this comment, or not, confused ...
still seeing
```

IS_ONE_OF(config.device.role, meshtastic_Config_DeviceConfig_Role_CLIENT_BASE, meshtastic_Config_DeviceConfig_Role_ROUTER,
                  meshtastic_Config_DeviceConfig_Role_ROUTER_LATE))
```

shalberd: I see, either the config / script setting is set OR device is of type CLIENT_BASE, ROUTER, ROUTER_LATE, 
meaning if of those device types, favorites are preserved regardless of config.device.preserve_favourite setting.
We'd need to make sure we document this somewhere, later on :-) 
How about making this depend just on the config setting?

ford-jones: Yes, I'm thinking a refactor which checks the role prior to this, which would then change the value of the config to the appropriate default if one of those roles which exercise special treatment over favourites is enabled. e.g. for a router, set `config.device.preserve_favourite` to `true` if it is not already.

This would then open the door to an inverse / opt-out setting for those roles too.

thebentern:
@ford-jones protobufs upstream in develop should be up to date. Just need to fix the conflict on the ref and we'll be g2g
image

PR #8302, opened 2025-10-11
Gps reset detect

jp-bennett:
A couple logging improvements, and then changing the GPS reset timing to happen during probing, right after the serial speed is changed.

The theory is that changing the serial speed may send odd characters up the serial line, and confuse the GPS.

Additionally, this GPS has been observed to output ``$PAIR021,AG3335M_V2.6.0.AG3335_20221230`` spontaneously after a reset, so we can identify the chip based on that output.

I've further tested on the Heltec tracker and the Heltec v4, and added GPS reset model detection for those.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -1275,20 +1275,20 @@ GnssModel_t GPS::probe(int serialSpeed)
         memset(&ublox_info, 0, sizeof(ublox_info));
         delay(100);
 
-#ifdef PIN_GPS_RESET
+#if defined(PIN_GPS_RESET) && PIN_GPS_RESET != -1
         digitalWrite(PIN_GPS_RESET, GPS_RESET_MODE); // assert for 10ms
         delay(10);
         digitalWrite(PIN_GPS_RESET, !GPS_RESET_MODE);
 
         // attempt to detect the chip based on boot messages
-        std::vector<ChipInfo> airoha = {
+        std::vector<ChipInfo> passive_detect = {
             {"AG3335", "$PAIR021,AG3335", GNSS_MODEL_AG3335},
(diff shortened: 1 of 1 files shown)


fifieldt:
Good news - the NMEA setup (GSA and GSV configuration) will run after the detection, regardless of how it's done within that method - see line 656

fifieldt:
TODO: check every variant that sets GPS_RESET to see if we need to change the ifdef

fifieldt:
Have not forgotten about this excellent work, trying to find time to do some testing.

jp-bennett:
We could merge a version that's restricted to the tracker, as that's the device that really has the problem.

fifieldt:
Works for me!

src/gps/GPS.cpp
```diff
@@ -1275,6 +1275,24 @@ GnssModel_t GPS::probe(int serialSpeed)
         memset(&ublox_info, 0, sizeof(ublox_info));
         delay(100);
 
+#ifdef PIN_GPS_RESET
```
fifieldt: suggest && PIN_GPS_RESET != -1

fifieldt:
Variants with PIN_GPS_RESET:

board_level = pr, for testing:
- [x] heltec_mesh_node_t114
- [ ] heltec_mesh_solar
- [x] heltec_v4
- [x] heltec_wireless_tracker
- [ ] heltec_wireless_tracker_v2
- [ ] rak4631
- [x] tracker-t1000-e


board_level = extra - assuming fine.
- gat562_mesh_trial_tracker
- heltec_capsule_sensor_v3
- heltec_wireless_tracker_V1_0
- monteops_hw1
- rak2560
- rak3401_1watt
- rak_wismeshtap
- rpipico-slowclock
- trackerd
- tracksenger
- wio-t1000-s

fifieldt (approved):

PR #8316, opened 2025-10-12
Fix muted protobuf compile errors

thebentern:


src/graphics/Screen.cpp
```diff
@@ -1490,7 +1490,7 @@ int Screen::handleTextMessage(const meshtastic_MeshPacket *packet)
                     strcpy(banner, "Alert Received");
                 }
                 screen->showSimpleBanner(banner, 3000);
-            } else if (!channel.settings.mute) {
+            } else if (!channel.settings.has_module_settings || !channel.settings.module_settings.is_muted) {
```
Xaositek, line 1493: Should we setting the conditional on if has_module_settings exists? I feel like that part isn't relevant to this check.

thebentern: If we don't have module_settings, we can't check for is_muted (or assume mute), and this would cause a core panic.

PR #8322, opened 2025-10-12
Fix erroneous limiting of power in Ham Mode

fifieldt:
Ham Mode ignores region regulatory limits, so regardless of whether we set a single TX_GAIN_LORA or an array with a non-linear PA, we shouldn't limit the power.

### DIFF
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -72,6 +72,7 @@ void PhoneAPI::handleStartConfig()
 
     LOG_INFO("Start API client config");
     nodeInfoForPhone.num = 0; // Don't keep returning old nodeinfos
+    nodeInfoQueue.clear();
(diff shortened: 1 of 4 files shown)


fifieldt:
I don't have a ham license, so can't test this properly :)

Marx1:
This does not fix the main issue in the old patch. We should not artificially limit the power. There are instances where greater than 11db is needed. 

Not all of the power amps are the same, so some radios need 13 or 14 db setting to achieve 1w, or we need 2or3db more to counter filter loss.

https://github.com/meshtastic/firmware/pull/8273#issuecomment-3395564154 is a great option while maintaining the ability to adjust power above the normal setting in the spec sheet.

fifieldt:
@Marx1 , I've left an update on the other thread :) This bugfix is still valid though, I believe.

PR #8323, opened 2025-10-13
Fix build-one-target

NomDeTom:
Fix the fix. Repair the repair. Probably needs @vidplace7 to check it over.

### DIFF
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -27,7 +27,6 @@ on:
 
 jobs:
   setup:
-    if: github.repository == 'meshtastic/firmware'
     strategy:
       fail-fast: true
       matrix:
@@ -58,7 +57,6 @@ jobs:
       check: ${{ steps.jsonStep.outputs.check }}
 
   version:
-    if: github.repository == 'meshtastic/firmware'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
@@ -91,7 +89,6 @@ jobs:
         run: bin/check-all.sh ${{ matrix.check.board }}
 
   build:
-    if: github.repository == 'meshtastic/firmware'
     needs: [setup, version]
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/build_one_target.yml
```diff
@@ -25,7 +25,22 @@ on:
...
-permissions: read-all
+permissions:
+  actions: read
+  attestations: read
+  checks: write
+  contents: write
+  deployments: read
+  id-token: write
+  issues: read
+  models: read
```
fifieldt: We don't need the AI models stuff for this, do we? trunk is complaining about this line

NomDeTom: Any permissions you don't mention get set to `none` automatically. I don't know what's needed exactly, but its basically demanding that the action has all the same permissions as the things it is calling, regardless of whether they will be called or not.

NomDeTom:
@vidplace7 I'm about to bring this up to date with the main branch, and poke it again.

NomDeTom:
@vidplace7 I removed native builds and added back as many of the conditional skips as I could.

Question: is it worth changing the matrix of arches to `          - all`?

.github/workflows/main_matrix.yml
```diff
@@ -359,7 +363,7 @@ jobs:
 
   publish-firmware:
     runs-on: ubuntu-24.04
-    if: ${{ github.event_name == 'workflow_dispatch' }}
+    if: ${{ github.event_name == 'workflow_dispatch' && github.repository == 'meshtastic/firmware'}}
```
vidplace7:
```suggestion
    if: ${{ github.event_name == 'workflow_dispatch' && github.repository == 'meshtastic/firmware' }}
```
You missed this ` ` 🤪

.github/workflows/main_matrix.yml
vidplace7, line 1: Please don't gate in here 😅 we need to be able to test forks

PR #8332, opened 2025-10-13
Revert "Fix Station G2 Lora Power Settings"

thebentern:
Reverts meshtastic/firmware#8273 until we have an additional lora param to handle line loss gain compensation


serialrf433:
@thebentern @fifieldt 
How many hours writing text should i spend to revert this revert?
https://github.com/meshtastic/firmware/issues/8445#issuecomment-3591919685

PR #8333, opened 2025-10-13
Update stale_bot.yml

NomDeTom:
Extend stale period to 60 days, and added a message on stale marking.

I checked nothing. This is a YOLO.

### DIFF
--- a/.github/workflows/stale_bot.yml
+++ b/.github/workflows/stale_bot.yml
@@ -19,7 +19,8 @@ jobs:
       - name: Stale PR+Issues
         uses: actions/stale@v10.1.0
         with:
-          days-before-stale: 60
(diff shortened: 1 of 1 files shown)


thebentern:
Can we keep the 45 days? The fact that we have almost 200 open issues tells a pretty stark story about our ability to manage the flood of issues we get on a day to day basis.

korbinianbauer:
> Can we keep the 45 days? The fact that we have almost 200 open issues tells a pretty stark story about our ability to manage the flood of issues we get on a day to day basis.

I mean very few of those are actually Issues, most of them are Feature Requests, and most of those should spend a couple rounds of scrutiny in the Discussions section first.

NomDeTom:
> I mean very few of those are actually Issues, most of them are Feature Requests, and most of those should spend a couple rounds of scrutiny in the Discussions section first.

They're only cycled out when there's no movement for 45 +7 days. Plenty of time for discussion, I think.

PR #8342, opened 2025-10-14
Allow vibra or buzzer only notifications to obey cutoff

Xaositek:
- Allow vibra or buzzer only notifications to obey cutoff


Xaositek:
@caveman99 or @thebentern  Can you sanity check this. The issue we're attempting to resolve is that if Buzzer Actions are set to disabled (i.e. no audio is playing); the vibra / buzzer runs forever. This removes the isRtttlPlaying check to allow it to always turn off

PR #8349, opened 2025-10-15
Add LR1121 support for T-Beam SUPREME ESP32-S3

mariusfaber98:
The LR1121 LoRa chip was not supported, but I got the T-Beam SUPREME ESP32-S3 with LR1121 as a gift.
So I wrote this fix so I can actually use it with Meshtastic and everyone else can also.

This PR adds LR1121 support for the tbeam-s3-core. I tried it on my T-Beam SUPREME ESP32-S3 and it works.
I only sent a message and did not do extensive testing, but I think everything should work fine.
Pretty small fix to make it work. (Thanks Claude)

I hope I did everything right - pretty new to embedded development and Meshtastic.

If you are curious, the T-Beam SUPREME ESP32-S3 with LR1121 is from AliExpress:
https://de.aliexpress.com/item/1005005418286231.html

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other (T-Beam SUPREME ESP32-S3 with LR1121)

### DIFF
--- a/variants/esp32s3/tbeam-s3-core/rfswitch.h
+++ b/variants/esp32s3/tbeam-s3-core/rfswitch.h
@@ -1,0 +1,11 @@
+#include "RadioLib.h"
+
+static const uint32_t rfswitch_dio_pins[] = {RADIOLIB_LR11X0_DIO5, RADIOLIB_LR11X0_DIO6, RADIOLIB_NC, RADIOLIB_NC, RADIOLIB_NC};
+
+static const Module::RfSwitchMode_t rfswitch_table[] = {
+    // mode                  DIO5  DIO6
+    {LR11x0::MODE_STBY, {LOW, LOW}},  {LR11x0::MODE_RX, {HIGH, LOW}},
+    {LR11x0::MODE_TX, {LOW, HIGH}},   {LR11x0::MODE_TX_HP, {LOW, HIGH}},
+    {LR11x0::MODE_TX_HF, {LOW, LOW}}, {LR11x0::MODE_GNSS, {LOW, LOW}},
+    {LR11x0::MODE_WIFI, {LOW, LOW}},  END_OF_MODE_TABLE,
+};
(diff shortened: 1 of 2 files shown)


lyusupov:
TCXO voltage is not valid 

image

<br>
<br>

[**HPD-16E datasheet**](https://github.com/Xinyuan-LilyGO/T-Display-S3-Pro-MVSRLora/blob/main/information/HPD16E_Series_V0.1_20230913.pdf) says that

image


By the way, are you aware that Semtech LR11xx has a specific incompatibility with Meshtastic issue ?
https://github.com/meshtastic/firmware/issues/4775

This issue affects operation of Meshtastic nodes based on Semtech SX1276 against nodes based on Semtech LR11xx - they don't see each other because Meshtastic uses non-standard LoRa synword byte....

LilyGO sells the LR1121 Supreme with SoftRF because the SoftRF project uses LoRa _(in FANET radio protocol only)_ with different syncword value which is not affected by the SX1276 compatibility issue mentioned above.

The issue https://github.com/meshtastic/firmware/issues/4775 is postponed until Meshtastic release 3.x with no specific ETA yet. It means that use of LR11xx radio for the purpose of Meshtastic is very limited at this time...

Seeed Studio had made T1000-E tracker for Meshtastic with LR1110 in year 2024. However, all further Seeed products for Meshtastic are based on SX1262 again.

mariusfaber98:
@lyusupov Thanks for the helpful comment. I've updated the LR11X0_DIO3_TCXO_VOLTAGE to 3.0 should be right i think.

I wasn't aware there was an issue with the LoRa syncword, but I think basic support is better than no support. What do you think about adding basic support now as unofficial/experimental support, and then it can become fully officially supported with 3.0?

fifieldt (approved):

PR #8356, opened 2025-10-15
Add Tanmatsu variant

renzenicolai:
This pull request adds support for [Tanmatsu](https://docs.tanmatsu.cloud/) by adding a variant for the board.

Early revisions of the Tanmatsu hardware use XTAL clock source while newer revisions use TCXO, hence the need for TCXO_OPTIONAL. There are two board variants, one with SX1262 based LoRa module for 868/915 MHz bands and one with SX1268 based LoRa module for 433MHz band.

## 🤝 Attestations

- [x] I have tested that my proposed changes do not cause any obvious regressions (it only adds a variant)


variants/esp32c6/tanmatsu/platformio.ini
```diff
@@ -0,0 +1,10 @@
+[env:tanmatsu]
+extends = esp32c6_base
+board = esp32-c6-devkitm-1
+board_level = pr
```
fifieldt: board_level = extra

is where we start :)

renzenicolai:
We don't have a frontend application for the ESP32-P4 on the board that works with up-to-date Meshtastic yet, but the contents of this PR work seem to work as intended.

The outdated frontend code reboots every minute or so, but worked fine with 2.6.7... You can find the code for that part of the porting effort here: https://github.com/Orange-Murker/konsool-meshtastic-ui

renzenicolai:
We've updated the frontend app to work with lastest meshtastic. Please review & merge this PR.

src/mesh/NodeDB.cpp
```diff
@@ -906,6 +906,15 @@ void NodeDB::installDefaultModuleConfig()
     moduleConfig.ambient_lighting.green = (myNodeInfo.my_node_num & 0x00FF00) >> 8;
     moduleConfig.ambient_lighting.blue = myNodeInfo.my_node_num & 0x0000FF;
 
+#if defined(HAS_SERIAL) && HAS_SERIAL
```
fifieldt, line 909: These changes would affect more than just this variant. Suggest merging just the variant and leaving this out, or trying to use existing code.

renzenicolai: It currently won't as none of the other variants define HAS_SERIAL. It simply adds a way for a variant to specify default values for the serial interface, which is needed on devices with a separate frontend microcontroller.

PR #8370, opened 2025-10-16
Remove the up/down shortcut to launch canned messages

jp-bennett:
By request, remove the up/down shortcut launching canned messages.


src/modules/CannedMessageModule.cpp
```diff
@@ -360,16 +360,7 @@ int CannedMessageModule::handleInputEvent(const InputEvent *event)
         return handleEmotePickerInput(event);
 
     case CANNED_MESSAGE_RUN_STATE_INACTIVE:
-        if (isSelect) {
```
thebentern, line 430: Do we know what exactly this was doing?

HarukiToreda: I believe while in the module screen on an inactive state, it was preventing the select button from mistakenly moving you out of the screen since select is also used to pull action menu. It was intercepting it so it did nothing.

src/modules/CannedMessageModule.cpp
HarukiToreda, line 1: do we still need the checks on line 56? we are now defaulting cannedmessage as enabled, so there's no need for the checks for them right?

src/modules/CannedMessageModule.cpp
jp-bennett: Basically correct. It was a short circuit to make sure the select button didn't do anything, and that event got passed into the screen handler.

src/modules/CannedMessageModule.cpp
jp-bennett: I'm blind, which checks?

HarukiToreda: CannedMessageModule::CannedMessageModule()
    : SinglePortModule("canned", meshtastic_PortNum_TEXT_MESSAGE_APP), concurrency::OSThread("CannedMessage")
{
    this->loadProtoForModule();
    if ((this->splitConfiguredMessages() <= 0) && (cardkb_found.address == 0x00) && !INPUTBROKER_MATRIX_TYPE &&
        !CANNED_MESSAGE_MODULE_ENABLE) {
        LOG_INFO("CannedMessageModule: No messages are configured. Module is disabled");
        this->runState = CANNED_MESSAGE_RUN_STATE_DISABLED;
        disable();
    } else {
        LOG_INFO("CannedMessageModule is enabled");
        moduleConfig.canned_message.enabled = true;
        this->inputObserver.observe(inputBroker);
    }
}

PR #8372, opened 2025-10-16
Refactor nodedb

jp-bennett:
Beginning the task of isolating the internals of NodeDB from its public functions. The eventual goal here is to add a mutex at the beginning of each public function, so we don't ever trounce our nodeDB from the API thread.


fifieldt (approved):
Read all the code.

This is good stuff!

rcd:
Not sure if you are interested, but I did play with a modification to NodeDB to keep the meshNodes vector sorted by nodeNum so that is could be accessed via binary search O(log n) instead of the linear search that it is currently using. The biggest impact is in packet parsing and routing decisions.

The current code keeps things sorted by most recently active, so if there are only a small number of active nodes, the linear search isn't really that bad. However once the mesh grows to more than a few active nodes the overhead becomes significant for these low power boards.

The primary reason for the current code's sort order is for display. The nodes are resorted every 5 seconds, regardless of if a user is viewing the node list or not. My code keeps a separate pointer vector sorted by the display order. The new vector is resorted only as needed for display, and then it is only sorted when something has changed, or every 5 seconds. There is no overhead for resorting if the display access methods are not called.

I'm currently running this on a heltec-v3, a t114 board, and a tbeam with no observable issues. I'm accessing from iOS and Mac apps. I have not tested other interfaces.

e535d7ecdd3527324cc31b612ccc673a7744cbb5

jp-bennett:
This is really stale, and some of the ideas have landed via other PRs, so closing.

PR #8373, opened 2025-10-16
Force CannedMessages to another node to be a PKI DM

jp-bennett:
Tighten up the CannedMessage security slightly by only listing nodes with valid public keys, and then explicitly setting it to be PKI.

I don't believe there is a vulnerability in the way this was being done, but this is slightly more hardened as a result of a question that was raised.


HarukiToreda:
just tested it. Works perfectly

PR #8390, opened 2025-10-18
Initial Thinknode-m3 support

jp-bennett:
Initial support for the Thinknode m3 tracker.

At the very least, needs the hardware ID from the protobufs added before this is ready to merge.


fifieldt:
Hi @jp-bennett , I re-targeted this to develop to avoid it going into master during out beta feature freeze.

Issue #8350, opened 2025-10-15
[Feature Request]: CSV Export doesn't have RSSI


uecken:
### Platform

ESP32

### Description

Please add RSSI in csv export file for coverage analysis.

fifieldt:
@ford-jones you Looked at this file recently didn't you?

ford-jones:
Hey, yeah I did some work in the module cleaning up it's results as they would just hang around in the filesystem.

Happy to take a look into this soon if nobody else beats me to it 🙂

ford-jones:
@fifieldt is there anything else we want to put in here while we're at it? e.g. SNR?

fifieldt:
If it's cheap, go nuts!

PR #8395, opened 2025-10-18
Include RSSI in rangetest csv

ford-jones:
Resolves: #8350 

Writes rx RSSI values alongside rangetest results to `rangetest.csv`.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3

### DIFF
--- a/src/modules/RangeTestModule.cpp
+++ b/src/modules/RangeTestModule.cpp
@@ -235,8 +235,8 @@ bool RangeTestModuleRadio::appendFile(const meshtastic_MeshPacket &mp)
         }
 
         // Print the CSV header
-        if (fileToWrite.println("time,from,sender name,sender lat,sender long,rx lat,rx long,rx elevation,rx rssi,rx "
-                                "snr,distance,hop limit,payload")) {
+        if (fileToWrite.println("time,from,sender name,sender lat,sender long,rx lat,rx long,rx elevation,rx "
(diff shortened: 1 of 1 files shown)


ford-jones:
Hey @fifieldt, I just realised I haven't actually got two ESP32's handy. I've only got the 1 heltec, the rest is all nRF...
Would you be able to give this a try and fill out the attestation ?

Otherwise, I can test it next week when I see my buddy who also has a heltec v3.

src/modules/RangeTestModule.cpp
```diff
@@ -234,8 +235,8 @@ bool RangeTestModuleRadio::appendFile(const meshtastic_MeshPacket &mp)
         }
 
         // Print the CSV header
-        if (fileToWrite.println(
-                "time,from,sender name,sender lat,sender long,rx lat,rx long,rx elevation,rx snr,distance,hop limit,payload")) {
+        if (fileToWrite.println("time,from,sender name,sender lat,sender long,rx lat,rx long,rx elevation,rx rssi,rx "
```
fifieldt: This is minor, but I feel like we should keep the ordering for existing fields in case any users are relying on that order being the same. Suggestion would be to put rx_rssi on the end as the last field.

ford-jones: Sure thing, makes sense :slightly_smiling_face:

PR #8385, opened 2025-10-18
Fix NimbleBluetooth reliability and performance

compumike:
As discussed on Discord https://discord.com/channels/867578229534359593/1039674919346249728/1427833207528755270 and elsewhere: fix NimbleBluetooth so that all `getFromRadio` happens in the main FreeRTOS task, and we don't have to worry about adding mutexes everywhere else in the codebase forever! 😂 

For details about how it works, see the large comment block near the start of `NimbleBluetooth.cpp`.

Loading 199 nodeInfos across BLE from Heltec v3 to iPhone 13 mini took `93980-87537 =` **6.443 seconds**:

```
INFO  | 22:58:01 87 [NimbleBluetooth] BLE requestHighThroughputConnection
INFO  | 22:58:01 87 [NimbleBluetooth] Client only wants node info, skipping other config
DEBUG | 22:58:01 87 [NimbleBluetooth] Got 6 files in manifest
INFO  | 22:58:01 87 [NimbleBluetooth] Start API client config millis=87537
DEBUG | 22:58:01 87 [NimbleBluetooth] Send My NodeInfo
DEBUG | 22:58:02 88 [NimbleBluetooth] nodeinfo: 20/199
DEBUG | 22:58:03 88 [NimbleBluetooth] nodeinfo: 40/199
DEBUG | 22:58:03 89 [NimbleBluetooth] nodeinfo: 60/199
DEBUG | 22:58:04 90 [NimbleBluetooth] nodeinfo: 80/199
DEBUG | 22:58:05 90 [NimbleBluetooth] nodeinfo: 100/199
DEBUG | 22:58:05 91 [NimbleBluetooth] nodeinfo: 120/199
DEBUG | 22:58:06 91 [NimbleBluetooth] nodeinfo: 140/199
DEBUG | 22:58:06 92 [NimbleBluetooth] nodeinfo: 160/199
DEBUG | 22:58:07 93 [NimbleBluetooth] nodeinfo: 180/199
DEBUG | 22:58:08 93 [NimbleBluetooth] Done sending 199 of 199 nodeinfos millis=93980
DEBUG | 22:58:08 93 [NimbleBluetooth] FromRadio=STATE_SEND_FILEMANIFEST
INFO  | 22:58:08 93 [NimbleBluetooth] Config Send Complete millis=93989
INFO  | 22:58:08 94 [NimbleBluetooth] BLE onConfigComplete
INFO  | 22:58:08 94 [NimbleBluetooth] BLE requestLowerPowerConnection
[...]

PR #7880, opened 2025-09-06
 NimbleBLE update to 2.3.6

Links2004:
This PR update NimbleBLE to 2.3.6 and does the needed migrations.

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [x] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [x] Other (please specify below)
         ESP32-S3 + RF95
         ESP32-C3 + RF95

PR #8404, opened 2025-10-19
Fix NimbleBluetooth: process fromPhoneQueue (phone->radio) before toPhoneQueue (radio->phone)

compumike:
Fix for the issue in https://github.com/meshtastic/firmware/pull/8385#issuecomment-3419723563 where commit https://github.com/meshtastic/firmware/commit/099ab06b9243d2427f3c406c275041b8f196aeb5 breaks the iOS client (especially if trying to connect within the first 10 seconds after boot, it seems).

As @garthvh pointed out, the clients expect to do a write-then-read when sending wantConfig. The read should block until the write is processed. So this switches the processing order and means that `onRead` (`toPhoneQueue`) will block until all `onWrite` (`fromPhoneQueue`) packets have been processed!

This fixes the iOS client, and simplifies the NimbleBluetooth code significantly! 👏 

I've only tested on NimBLE 1. (I'm still fighting to get `NIBMLE_TWO` running locally...)

Would appreciate testing especially from:
- android app
- NIMBLE_TWO

## 🤝 Attestations

  - [X] Heltec (Lora32) V3

### DIFF
--- a/src/graphics/draw/UIRenderer.cpp
+++ b/src/graphics/draw/UIRenderer.cpp
@@ -563,6 +563,7 @@ void UIRenderer::drawDeviceFocused(OLEDDisplay *display, OLEDDisplayUiState *sta
     display->setTextAlignment(TEXT_ALIGN_LEFT);
     display->setFont(FONT_SMALL);
     int line = 1;
+    meshtastic_NodeInfoLite *ourNode = nodeDB->getMeshNode(nodeDB->getNodeNum());
 
     // === Header ===
 #if defined(M5STACK_UNITC6L)
@@ -740,7 +741,6 @@ void UIRenderer::drawDeviceFocused(OLEDDisplay *display, OLEDDisplayUiState *sta
     int yOffset = (isHighResolution) ? 0 : 5;
     std::string longNameStr;
 
-    meshtastic_NodeInfoLite *ourNode = nodeDB->getMeshNode(nodeDB->getNodeNum());
     if (ourNode && ourNode->has_user && strlen(ourNode->user.long_name) > 0) {
         longNameStr = sanitizeString(ourNode->user.long_name);
     }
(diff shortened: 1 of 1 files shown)


compumike:
OK, I was able to test this PR successfully with `NIMBLE_TWO` with a Heltec V3. All good with iOS client 👍 

Here is what I had to do to get `NIMBLE_TWO` working on the `heltec-v3` platform: https://gist.github.com/compumike/75b87fd7e2f2399198c270be42c337b4 -- thanks @Links2004 

Without the changes to `NimbleBluetooth.cpp` proposed in #7880 https://github.com/meshtastic/firmware/pull/7880/files#diff-228810058710e46189de098dbd92bba5cd367ad8b49b20d2efc2b202b258b766 I can't get it to compile: `src/nimble/NimbleBluetooth.cpp:783:30: error: 'getClientByHandle' is not a member of 'NimBLEDevice'`. So I don't know how y'all are managing to test it on `NIMBLE_TWO` without these changes!

compumike:
#8404 (This PR) works for me **without** `NIMBLE_TWO`, but fails to build with `NIMBLE_TWO` with `src/nimble/NimbleBluetooth.cpp:783:30: error: 'getClientByHandle' is not a member of 'NimBLEDevice'`.

#8405 (This PR + one extra commit) works for me **with OR without** `NIMBLE_TWO`. The one extra commit is https://github.com/meshtastic/firmware/pull/8405/commits/524d2874ec267b7a245a794ba6b9d54434e88b29 which are the changes from @Links2004 #7880 to `NimbleBluetooth.cpp`.

-----

I think this PR (#8404) is safe and good to merge, and fixes the iOS client issue, and should be compatible with the Android one too. <-- if possible, I wouldn't wait on merging this one, because the current behavior is broken on iOS. 🙏 

I'm less sure about #8405 because I don't have a C6L to test with. Maybe @Links2004 can take a peek too?

compumike:
**Update:** This PR works as-is both **with OR without** `NIMBLE_TWO` on my `heltec-v3` with the iOS app. (Never mind the above comments about #8405; that can be dealt with separately in #7880!)

I made a more minimal guide to using `NIMBLE_TWO` with `heltec-v3`: https://gist.github.com/compumike/64cd0384b66c540364254679e2c55fc7 . Compared to my [earlier version](https://gist.github.com/compumike/75b87fd7e2f2399198c270be42c337b4), by skipping `-DCONFIG_BT_NIMBLE_ROLE_CENTRAL_DISABLED`, the rest of `NimbleBluetooth.cpp` seems to work fine for now!

-----

I have tested this PR #8404 with:

- heltec-v3
  - (default) without `NIMBLE_TWO`
    - ✅ iOS client BLE: works great
    - ✅ web client BLE: works great
  - `-DNIMBLE_TWO` per [minimal gist](https://gist.github.com/compumike/64cd0384b66c540364254679e2c55fc7) 
    - ✅ iOS client BLE: works great
    - ✅ web client BLE: works great
- m5stack_unitc6l
  -  ✅ builds cleanly (I do not have hardware to test with 🤷 )

HarukiToreda:
tested on a Heltec V3 on powersave mode, node reconnects perfectly every time bluetooth wakes up from sleep with no issues.

PR #8427, opened 2025-10-23
Move airtime calculation to when Tx is complete

GUVWAF:
Try-fix for huge airtime calculations and missing Tx done interrupts.

With the new airtime calculation method we’re doing an SPI call to the radio while it is transmitting. Maybe it doesn’t like that and causes it to miss the interrupt also.


fifieldt (approved):
Seems worth a shot

Issue #8082, opened 2025-09-22
[Bug]: Heltec Wireless Paper with InkHud rebooting constantly


derfloh:
### Category

Other

### Hardware

Heltec Wireless Paper

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [x] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.10

### Description

A while ago I've bought a Heltec Wireless Paper and it rebooted/crashed way too often (reliably every 3-5 Minutes, but sometimes it worked for hours) (this was with 2.7.5 IIRC). I recently updated to 2.7.9 to see if the situation has improved, but it didn't. I did some testing/debugging with 2.7.10 but it also showed the same symptoms. 

Crash report looks like:

```
DEBUG | 11:23:19 310 [Router] Module 'RecentsListApplet' wantsPacket=1
DEBUG | 11:23:19 310 [Router] Module 'RecentsListApplet' considered
DEBUG | 11:23:19 310 [Renderer] FULL-update debt:0.933333
DEBUG | 11:23:19 310 [Renderer] UNSPECIFIED: using FAST
INFO  | 11:23:19 310 [Renderer] Drawing card for node 0xb50443e0
INFO  | 11:23:19 310 [Renderer]  nodeDB Size=200
INFO  | 11:23:19 310 [Renderer]  Printing hops away
INFO  | 11:23:19 310 [Renderer]  Node has 3 hops away
INFO  | 11:23:19 310 [Renderer]  Node Field HasHopsAway=1
INFO  | 11:23:19 310 [Renderer]  Node 3
INFO  | 11:23:19 310 [Renderer]  Printing hops away value 3
INFO  | 11:23:19 310 [Renderer]  Printed hops away
INFO  | 11:23:19 310 [Renderer]  Printed hatch
INFO  | 11:23:19 310 [Renderer] Drawing card for node 0x6f2531d2
INFO  | 11:23:19 310 [Renderer]  nodeDB Size=200
INFO  | 11:23:19 310 [Renderer]  Printing hops away
INFO  | 11:23:19 310 [Renderer]  Node has 1 hops away
INFO  | 11:23:19 310 [Renderer]  Node Field HasHopsAway=1
INFO  | 11:23:19 310 [Renderer]  Node 1
INFO  | 11:23:19 310 [Renderer]  Printing hops away value 1
[...]

PR #8428, opened 2025-10-23
InkHUD crash fix when nodes get deleted from NodeDB

HarukiToreda:
This fix resolves a crash that could happen when the NodeDB reached its limit and automatically purged older nodes. When that happened, the Node List screen was still trying to draw information for a node that no longer existed, leading to a reboot. The applet now safely skips any deleted or missing nodes, preventing crashes during automatic purges. This crash could also be seen if a node was deleted while on the nodelist applet.

Should fix issue https://github.com/meshtastic/firmware/issues/8082


fifieldt:
This should go into master too, for the beta.

PR #8434, opened 2025-10-23
Add build support for TenStar Esp32-C3 SuperMini

kzsolt3141:
Tried to build the **esp32c3_super_mini** platform for my ESP32C3-SuperMini. The defined pinout was not compatible with my version.
Added new variant, with the correct pinout.
Tested with my own ESP32C3-SuperMini, LLCC68 LoRa, SSD1306, works correctly.

### DIFF
--- a/variants/esp32c3/diy/esp32c3_super_mini_tenstar/variant.h
+++ b/variants/esp32c3/diy/esp32c3_super_mini_tenstar/variant.h
@@ -1,0 +1,63 @@
+#ifndef _VARIANT_ESP32C3_SUPER_MINI_TENSTAR_
+#define _VARIANT_ESP32C3_SUPER_MINI_TENSTAR_
+
+/*----------------------------------------------------------------------------
+ *        Headers
+ *----------------------------------------------------------------------------*/
+
+#ifdef __cplusplus
+extern "C" {
+#endif // __cplusplus
+
+// I2C (Wire) & OLED
+#define WIRE_INTERFACES_COUNT (1)
+#define I2C_SDA (8)
+#define I2C_SCL (9)
+
+#define USE_SSD1306
+
+// GPS
+#undef GPS_RX_PIN
+#undef GPS_TX_PIN
+#define GPS_RX_PIN (20)
+#define GPS_TX_PIN (21)
+
+// Button, active LOW
+#define BUTTON_PIN (10)
+
+// LoRa
+#define USE_LLCC68
+// TODO: why enable SX126x if not used ?
+// Note: without these, LoRa will not work correctly
+#define USE_SX1262
+// #define USE_RF95
+#define USE_SX1268
+
+#define LORA_DIO0 RADIOLIB_NC
+#define LORA_RESET (0)
+#define LORA_DIO1 (1)
+#define LORA_RXEN (2)
+#define LORA_BUSY (3)
+#define LORA_SCK (4)
+#define LORA_MISO (5)
+#define LORA_MOSI (6)
+#define LORA_CS (7)
+
+#define SX126X_CS LORA_CS
+#define SX126X_DIO1 LORA_DIO1
+#define SX126X_BUSY LORA_BUSY
+#define SX126X_RESET LORA_RESET
+#define SX126X_RXEN LORA_RXEN
+
+#define SX126X_DIO3_TCXO_VOLTAGE (1.8)
+#define TCXO_OPTIONAL // make it so that the firmware can try both TCXO and XTAL
+
+#ifdef __cplusplus
+}
+#endif
+
+/*----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


variants/esp32c3/diy/esp32c3_super_mini_tenstar/variant.h
```diff
@@ -0,0 +1,63 @@
...
+#define USE_SSD1306
+
+// GPS
+#undef GPS_RX_PIN
+#undef GPS_TX_PIN
+#define GPS_RX_PIN (20)
+#define GPS_TX_PIN (21)
+
+// Button, active LOW
+#define BUTTON_PIN (10)
```
NomDeTom, line 26: I traced the boot button through to pin 9 on my boards, and the very-difficult-to-read schematic on the Tenstar Robot store Aliexpress page also seems to indicate that it is linked to GPIO9.

variants/esp32c3/diy/esp32c3_super_mini_tenstar/variant.h
```diff
@@ -0,0 +1,63 @@
...
+
+// LoRa
```
NomDeTom, line 28: Can you expand on why you chose these pins?

For reference, the ones in the original super_mini were chosen because they matched the existing Heltec HT-CT62 internal pins, and were therefore known to be usable.

NomDeTom (commented):
Generally I can see what you're trying to do here - using the named pins for their described purpose rather than using the existing pinout. However, I've left some notes for you.

If you're going to raise a new variant, perhaps consider trying to match the ESP32S3 zero, which has an almost-identical-but-mirrored pinout, and could therefore share boards (with adjustment for buttons).

variants/esp32c3/diy/esp32c3_super_mini_tenstar/variant.h
kzsolt3141: I tried PIN8 and PIN9 for button but in both cases the I2C got stuck: display, sensors, nothing communicates. So I moved to an other pin, in this case PIN10. Maybe I should change the comments: "Connect Button to this pin, active LOW"

variants/esp32c3/diy/esp32c3_super_mini_tenstar/variant.h
kzsolt3141: In my understanding on this board, UART, SPI, I2C are fixed, so only pins 0, 1, 2, 3 and 10 are available. Did I missed something?  
Or should I rearrange these four values?
#define LORA_RESET (0)
#define LORA_DIO1 (1)
#define LORA_RXEN (2)
#define LORA_BUSY (3)

image

variants/esp32c3/diy/esp32c3_super_mini_tenstar/variant.h
NomDeTom: See my comment below. Io9 is a strapping pin, and also connected to a button, so using it for anything other than a user button doesn't make sense.

variants/esp32c3/diy/esp32c3_super_mini_tenstar/variant.h
NomDeTom: Ah, on the esp32c3 they're not fixed functions.  I'm not sure why they've used them in this way.

I'd strongly suggest rethinking the use of io9 for anything other than the button that it is connected to.

NomDeTom:
@kzsolt3141 have you tried the other pinout for the supermini? You said that it's not compatible, but if you try it you may find that it works.

variants/esp32c3/diy/esp32c3_super_mini_tenstar/variant.h
mrekin: > I traced the boot button through to pin 9 on my boards,

Just to confirm - Esp32C3 datasheet says that GPIO9 used for bootmode selection, so it have to be wired to boot btn )

mrekin:
> other pinout

I use different pinout https://github.com/mrekin/MeshtasticCustomBoards/blob/main/firmware/variants/esp32c3/diy/diy_nano_c3/variant.h
(at least 2 working implementations with e22 30S and 22s)
Variant was created for existing hw, so I may be not optimal (especially with rxen txen).

@kzsolt3141 
This PR don't contians SX126X_DIO2_AS_RF_SWITCH, does it works without TXEN pin definition?

kzsolt3141:
I tried again the existing esp32c3-supermini platform with success. Initially I was thinking that my board does not support other than dedicated pins for i2c and spi, but I was wrong. So i will  close this PR, ant thank you all for the constructive comments!

fifieldt:
Thanks a ton for being willing to work upstream, @kzsolt3141 :)

Issue #6158, opened 2025-02-26
[Feature Request]: Add API Port to Native config.yaml


pdxlocations:
### Platform

Linux Native

### Description

For running multiple instances of meshtasticd, it could be useful to allow configuration of the API port in config.yaml

PaulRoze:
Hey, looking for my first contribution here. I think PR #8435 already solved this one (merged Jan 17). APIPort works from config.yaml now — parsed in PortduinoGlue.cpp, passed to initApiServer. Example is in bin/config-dist.yaml line 218, commented out.

Priority: CLI -p first, then YAML APIPort, then default 4403.

Looks like this can be closed.

PR #8435, opened 2025-10-23
Add support for setting API port from the config file

tedwardd:
## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - Native/Linux
    
## Summary
Adds support for setting API port via config.yaml. This addressed #6158.

## Behavior
Evaluates `["General"]["APIPort"]` and confirms it's a valid port number _only if_ the port was not defined at runtime.
In order of evaluation priority:
1. Runtime argument
2. Config option
3. Default 4403

## Testing
Tested on Linux using sim mode and confirmed that the port is set according to the expected behavior above


thebentern:
@jp-bennett

Issue #8417, opened 2025-10-21
[Feature Request]: Activate Secondary I2C bus on Heltec V4


b8b8:
### Platform

ESP32

### Description

It appears there was initial work done to have a secondary i2c bus added to pins 3+4 on the Heltec V4 however currently only the one used for the screen (17+18) is active. While the pins at 17+18 are available, they are in a relatively awkward position at the end of the board.

<img width="582" height="502" alt="Image" src="https://github.com/user-attachments/assets/e2d2fc70-1a92-4259-a165-6b6d578b0e38" />

<img width="688" height="310" alt="Image" src="https://github.com/user-attachments/assets/c8f9b215-245f-46c2-a1ec-7572291d7952" />

<img width="617" height="337" alt="Image" src="https://github.com/user-attachments/assets/4a71ecd7-9bbc-4b23-ade5-371368a68f67" />

fifieldt:
In case someone implementing this needs a reference,m, here's how it looks for another device that has 2 I2C 

<img width="1014" height="510" alt="Image" src="https://github.com/user-attachments/assets/cc9c574b-0739-4fbe-846a-7080a3f4252f" />

HarukiToreda:
if there were a secondary, it should match the V3 node which was 41 and 42, since this is meant to be a drop in replacement for it, changing the pins around will work against that.

b8b8:
> if there were a secondary, it should match the V3 node which was 41 and 42, since this is meant to be a drop in replacement for it, changing the pins around will work against that.

Sure, no problem with that. That makes sense.

Bajotumn:
> if there were a secondary, it should match the V3 node which was 41 and 42, since this is meant to be a drop in replacement for it, changing the pins around will work against that.

41 and 42 are routed to the GPS socket on the back of the board though. I'd second 3 and 4 for this use case.

<img width="426" height="125" alt="Image" src="https://github.com/user-attachments/assets/3d619e00-6232-4b94-b11a-f047811deda4" />

<img width="342" height="459" alt="Image" src="https://github.com/user-attachments/assets/86ca2b75-b4c1-4904-9428-37859d951897" />

b8b8:
> > if there were a secondary, it should match the V3 node which was 41 and 42, since this is meant to be a drop in replacement for it, changing the pins around will work against that.
> 
> 41 and 42 are routed to the GPS socket on the back of the board though. I'd second 3 and 4 for this use case.
[...]

PR #8447, opened 2025-10-25
Registers GPIO3 and GPIO4 for secondary I2C bus

Bajotumn:
Resolves meshtastic/firmware#8417

Given pins 41 and 42 are used for GPS on this version it would seem using those pins, though drop in compatible with the V3 board, would render the GNSS socket unusable unless one used the inconvenient OLED pins on the forward edge of the board.

I will test this on a Heltec v3, RAK 4631, and a T-Beam as I have those, but a regression seems quite unlikely given this is a simple `variant.h` update.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/variants/esp32s3/heltec_v4/variant.h
+++ b/variants/esp32s3/heltec_v4/variant.h
@@ -6,6 +6,10 @@
 #define I2C_SDA 17 // I2C pins for this board
 #define I2C_SCL 18
 
+// Enable secondary bus for external periherals
+#define I2C_SDA1 SDA
+#define I2C_SCL1 SCL
+
 #define VEXT_ENABLE 36 // active low, powers the oled display and the lora antenna boost
 #define BUTTON_PIN 0
 


Bajotumn:
> This probably needs WIRE_INTERFACES_COUNT to be set to 2

It looks like this is implicitly set in `configuration.h` when `I2C_SDA1` is defined.
https://github.com/meshtastic/firmware/blob/1ac2382d7cad544c5685aa50a18096d1fe20aadb/src/configuration.h#L303-L311

Happy to make this change if we think it should be explicitly defined though.

Bajotumn:
And here's the testing setup

image

image

b8b8:
Nice, thanks for adding this

HarukiToreda:
I would suggest reviewing this one again, Heltec recently introduced this Touch expansion carrier and it is meant to fully integrate to the existing assigned pins. So it's best to first see what the i2c pin is used for it to avoid conflicts in plans already in the works


https://heltec.org/project/wifi-lora-32-v4-expansion-housing/

variants/esp32s3/heltec_v4/variant.h
```diff
@@ -6,6 +6,10 @@
 #define I2C_SDA 17 // I2C pins for this board
 #define I2C_SCL 18
 
+// Enable secondary bus for external periherals
+#define I2C_SDA1 SDA
```
HarukiToreda: must make sure there's no conflict with the heltec expansion housing

b8b8:
image
image

image


they appear to be setup for sensor use and pass through to the back as expected (but reversed????)

Bajotumn:
> image image image
> they appear to be setup for sensor use and pass through to the back as expected (but reversed????)

Very interesting. Good catch on the pass through to the external sockets.

I have one of these coming in and can verify once it arrives if `SDA` and `SCL` need swapping. It does look like the [official Heltec](https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/0d05479b43af8e4588a07fcca6e05e541b5cb86e/variants/heltec_wifi_lora_32_V4/pins_arduino.h) code still has them defined as I do here. Time will tell.

b8b8:
> > image image image
> > they appear to be setup for sensor use and pass through to the back as expected (but reversed????)
> 
> Very interesting. Good catch on the pass through to the external sockets.
> 
> I have one of these coming in and can verify once it arrives if `SDA` and `SCL` need swapping. It does look like the [official Heltec](https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/0d05479b43af8e4588a07fcca6e05e541b5cb86e/variants/heltec_wifi_lora_32_V4/pins_arduino.h) code still has them defined as I do here. Time will tell.

Could just be a documentation error and nothing needs to change then. I dont think there is any sensor actually hooked up to the add-on board this would affect.

HarukiToreda:
yeah probably doesn't need to change, but in the word of caution, it's good to make sure with the actual carrier, harder to change it once people have already soldered stuff to it and suddenly it switched from an update..

Bajotumn:
> yeah probably doesn't need to change, but in the word of caution, it's good to make sure with the actual carrier, harder to change it once people have already soldered stuff to it and suddenly it switched from an update..

So far, hopefully that's just me! 😀

HarukiToreda:
looks like this was done on this PR https://github.com/meshtastic/firmware/pull/8539

b8b8:
image

 lol they did swap the pins.
 ok time to switch my cables, easy fix.

HarukiToreda:
I think it's safe to say this PR can be closed since they got their people already on the case

Issue #8409, opened 2025-10-19
[Board]: XIAO Wio-SX1262 Header Board (NOT Kit)


Brandon-T:
### SOC

ESP32

### Lora IC

Semtech SX1262

### Product Link

https://www.seeedstudio.com/Wio-SX1262-for-XIAO-p-6379.html

### Description

When installing Meshtastic on the SeeedStudio XIAO ESP32-S3 with the above Wio-SX1262, it will say SX1262 Init Failed: -2 in the console.

This board (Header): https://www.seeedstudio.com/Wio-SX1262-for-XIAO-p-6379.html
is different from (Kit):
https://www.seeedstudio.com/Wio-SX1262-with-XIAO-ESP32S3-p-5982.html

The pin-out is (Wio-SX1262-for-XIAO-p-6379):
```c++
// XIAO S3 LORA module
#define USE_SX1262

#define LORA_MISO 8
#define LORA_SCK 7
#define LORA_MOSI 9
#define LORA_CS 5

#define LORA_RESET 3
#define LORA_DIO1 2

#define LORA_DIO2 6

#ifdef USE_SX1262
#define SX126X_CS LORA_CS
#define SX126X_DIO1 LORA_DIO1
#define SX126X_BUSY 4
#define SX126X_RESET LORA_RESET

//  DIO2 controlls an antenna switch and the TCXO voltage is controlled by DIO3
#define SX126X_DIO2_AS_RF_SWITCH
#define SX126X_RXEN 6
#define SX126X_TXEN RADIOLIB_NC
#define SX126X_DIO3_TCXO_VOLTAGE 1.8
#endif
```

Result:
```
INFO | ??:??:?? 119 CannedMessageModule is enabled
INFO | ??:??:?? 119 External Notification Module Disabled
DEBUG | ??:??:?? 119 SX126xInterface(cs=5, irq=2, rst=3, busy=4)
DEBUG | ??:??:?? 119 SX126X_DIO3_TCXO_VOLTAGE defined, using DIO3 as TCXO reference voltage at 1.800000 V
INFO | ??:??:?? 119 Start meshradio init
INFO | ??:??:?? 119 Radio freq=906.875, config.lora.frequency_offset=0.000
INFO | ??:??:?? 119 Set radio: region=US, name=LongFast, config=0, ch=19, power=30
INFO | ??:??:?? 119 myRegion->freqStart -> myRegion->freqEnd: 902.000000 -> 928.000000 (26.000000 MHz)
[...]

PR #8448, opened 2025-10-25
Add support for SeeedStudio's XIAO ESP32-S3 Wio-SX1262 Header

Brandon-T:
Fixed: https://github.com/meshtastic/firmware/issues/8409

image

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [x] LilyGo T-Beam
  - [x] RAK WisBlock 4631


variants/esp32s3/diy/seeed-xiao_esp32s3_diy_header/variant.h
NomDeTom, line 1: I would suggest removing the ASCII artwork. It's cute when the OEM does it. It's annoying if it's everywhere.

NomDeTom:
This looks good, and it's good that you've tested it. Putting it in DIY as a separate variant is the right thing to do.

@fifieldt does this need to be private_hw or continue under the HWID for the xiao S3?

variants/esp32s3/diy/seeed-xiao_esp32s3_diy_header/variant.h
Brandon-T: Fixed. Also changed to `-D PRIVATE_HW`

Compiled with `~/.platformio/penv/bin/pio run -e seeed-xiao-s3-diy-header` which created `~/Desktop/mestastic_firmware/.pio/build/seeed-xiao-s3-diy-header/firmware.factory.bin`

Flashed with: `~/.platformio/penv/bin/pio run -e seeed-xiao-s3-diy-header -t upload`

and tested. Also tried it with Visual Studio Code to build and flash as well, just in case.

Sabering1:
any updates?

NomDeTom:
You will need to update the GPS pin definitions in line with https://github.com/meshtastic/firmware/pull/8772

Also, you have overlapping I2C pin definitions with the radio pins. I suggest omitting them, or pushing them to unused internal pins, or putting them on the BTB header.

PR #8457, opened 2025-10-26
Another SHT41 detected as SHT31

andrewdarkstar:
Added serial 0x109f. SHT41 module bought at aliexpress.com

## 🤝 Attestations

  - [x] Heltec (Lora32) V3

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -8,15 +8,15 @@ plugins:
       uri: https://github.com/trunk-io/plugins
 lint:
   enabled:
-    - checkov@3.2.483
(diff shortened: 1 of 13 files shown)


dem1tris:
The serial numbers for SHT4x are declared as unique by the datasheet, so the detection logic is wrong, and such fixing attempts seem to be useless. I'll try to implement general fix

dem1tris:
See https://github.com/meshtastic/firmware/pull/8525

fifieldt:
Thanks for sending this in, as @dem1tris noted a more general fix was implemented.

PR #8467, opened 2025-10-27
PIN_PWR_DELAY_MS --> PERIPHERAL_WARMUP_MS

fifieldt:
It turns out we had two methods for delaying startup while peripherals warmed up. They were invented within months of each other and just missed the chance to merge.

Let's delete PIN_PWR_DELAY_MS and use PERIPHERAL_WARMUP_MS, since it's most common and earlier in the sequence.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1186,11 +1186,6 @@ void setup()
 #endif
 #endif
 
-#ifdef PIN_PWR_DELAY_MS
-    // This may be required to give the peripherals time to power up.
-    delay(PIN_PWR_DELAY_MS);
-#endif
-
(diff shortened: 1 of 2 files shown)


HarukiToreda:
tested on Heltec V3 with Cardkb and BME680 since this was one of the reasons why it was introduced. All peripherals started with no issues both from soft reboot and cold start boot using a power switch. Same test was made on the T114 with a CardKB, BME680 and IMU. No regressions found, safe to merge.

fifieldt:
This has waited 6 weeks for review , and an expert confirmed it did no harm, going to merge ...

PR #8478, opened 2025-10-28
Refactor num uint16

h3lix1:
## What changed
- Bigger node counts: NodeStatus and ProtobufModule::numOnlineNodes now use uint16_t (was uint8_t).
- Map/UI fixes: Iteration/counting in the map applet uses wider types to avoid overflow.

## Motivation
- To support #8097 when adding more nodes to the DB

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Heltec (Lora32) V4


fifieldt:
Thank you! I think we can all agree that 255 is quite a severe limit, and this is a logical step in breaking free of those shackles :)

PR #8486, opened 2025-10-29
Support overriding GPS serial pins on all architectures

Stary2001:
We only support overriding serial pins on ESP32. It turns out we can do this on the other architectures, too.
~~Opening as draft until it's tested better.~~

## 🤝 Attestations 
  - [x] RAK4631 (by GlockTober in discord)


fifieldt (approved):

NomDeTom:
@fifieldt can you mark this as bugfix? It used to work. Not sure when it stopped.

PR #8492, opened 2025-10-30
Better implementation of ExternalNotificationModule::stopNow

Xaositek:
- Better implementation of ExternalNotificationModule::stopNow
- Improved ExternalNotificationModule::stopNow with logging from runOne version
- Fixed isRtttlPlaying checks so it works effectively


Xaositek:
Tested on Meshenger (one of the original devices to show problems), NRF TXT, and HEL TXT.

PR #8502, opened 2025-10-31
fix(misc): Fix few compilation/check warnings

gxcreator:
Small fixes to eliminate compiler warnings, mainly removing dead code.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Tested on Heltec Wireless Paper (ESP32S3)

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -376,7 +376,7 @@ class AnalogBatteryLevel : public HasBatteryLevel
         // ADC2 wifi bug workaround not required, breaks compile
         // On ESP32S3, ADC2 can take turns with Wifi (?)
 
-        int32_t adc_buf;
+        int adc_buf;
         esp_err_t read_result;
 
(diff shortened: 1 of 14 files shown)


Xaositek (changes requested):
This moves around a lot of code just for the sake of moving things around. Too many things are blindly moved without consideration in BaseUI for me to be comfortable with it. I do not agree.

Xaositek:
Some of this is also available already in https://github.com/meshtastic/firmware/pull/8182

HarukiToreda:
most of these changes are already addressed on https://github.com/meshtastic/firmware/pull/8182 which redesigns a huge chunk of the code already. I'd suggest only merging this after that PR is merged

PR #8506, opened 2025-11-01
Skip setting up Lora GPIO lines when using a ch341 radio on native

jp-bennett:
Got a report that we were unintentionally talking to system GPIO lines when using ch341-based hardware, and the usermode driver. Easy solution here is to just skip initializing the Lora GPIO lines when using ch341.


jp-bennett:
@thebentern this would be ideal to sneak into 2.7.13

PR #8510, opened 2025-11-01
Detect two more variants of SHT4x

sandos:
A trivial change to add two more register values for SHT4x sensors.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3


sandos:
I am _not_ sure that these values are correct, or in particular, unique for SHT4x sensors. This could be aliexpress mis-labelling things?

fifieldt:
Hi @sandos , thanks a lot for looking into this.

At this stage we're almost certain this field which we originally thought was a model number is a serial number. Would you be confident in switching the code to check another register the SHT4x has instead? Idea would be to find one that SHT4 has that SHT3 doesn't or doesn't respond to in the same way ... and just see that the SHT4 responds.

sandos:
I only have 5 of these SHT4x sensors, so would not really know how to differentiate, but I can have a look around at least.

dem1tris:
See https://github.com/meshtastic/firmware/pull/8525

fifieldt:
Thanks for sending this in, as @dem1tris noted a more general fix was implemented.

PR #8515, opened 2025-11-02
Add GPS Profile support.

fifieldt:
This patch makes major changes to address the current complexity of GPS Configuration.

Introduces **GpsProfile**, a new protobuf we use to configure a pre-tuned selection of GPS settings. The aim is to replace the current 5 configuration options with a single user-friendly label that is tailored to real-world Meshtastic use cases.

Based on a users selection of pedestrian, vehicle, airborne, or fixed position, we apply different settings to GPS Update Interval, Position Broadcast Interval, Smart Broadcast Min Interval, Smart Broadcast Min Distance, and Position Flags in firmware.
This should allow the apps to drastically simplify the position configuration shown to users.

We aim take advantage of the profile to improve the accuracy of our position information, such as by discarding erroneous positions based on physically impossible movement. In this patch we use advanced features in GPS hardware specific to each profile to do this. In a follow-up patch we'll extend smartPosition to handle cases in software where GPS hardware support for this is not available.

More information about the selections for each use case is available in this [google sheet](https://docs.google.com/spreadsheets/d/1-f9z5zx2VCYqE6ivYXm-XqtLeVZSGMKpC3G3wma6-vw/edit?gid=633840990#gid=633840990)

PositionFlags has long been seen as a complicated and confusing burden in the apps. We determined that very few users use them, with the majority happy with the default set, and some of the 'advanced' flags don't even work.

After the protobuf declaration, this patch is the next step to deprecate PositionFlags,
 which will now be selected in firmware based on the GPSProfile.

Depends on: https://github.com/meshtastic/protobufs/pull/749

Co-Authored-By: @RCGV1 (#6650)

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


RCGV1:
I tested this feature path on a real LILYGO T-Beam S3 Core (u-blox M10) with a rebased protobuf branch and a small corrective firmware patch.

Validated end-to-end:
- BLE OTA streamed the 2,358,000-byte test image successfully with ACK pacing and terminal success.
- Sending gps_profile=VEHICLE through the admin/protobuf path resulted in the expected live position config: GPS update 30s, position broadcast 900s, smart minimum interval 120s, smart minimum distance 500m, smart enabled, flags 937.
- Sending MANUAL afterward preserved the existing user-tuned position settings.
- The bench node was restored to normal firmware and its original config after testing.

I do not think this PR is merge-ready as submitted:
1. AdminModule::handleSetConfig() assigns config.position but does not apply the selected profile timer/distance/flag policy. A client selecting a profile therefore updates the enum without receiving the advertised profile settings.
2. Several GNSS branches treat MANUAL/unset as pedestrian, which conflicts with the protobuf definition that MANUAL leaves choices to the user.
3. The T-Beam u-blox M10 path has no GPS-profile dynamic-model configuration.
4. The protobuf PR needs a rebase onto current protobuf master before firmware can safely consume it; pinning its older head removed newer wire definitions needed by current firmware.

I have a tested correction locally that centralizes profile application, preserves MANUAL, and maps u-blox M10 models (fixed=2, pedestrian=3, vehicle=4, airborne <2g=7).

PR #8517, opened 2025-11-02
Rename screen options to display options and add units chooser

jp-bennett:
Tested so far on the t-lora Pager. Slight menu re-arrange, and add the units chooser.


src/graphics/draw/MenuHandler.cpp
```diff
@@ -1578,6 +1585,30 @@ void menuHandler::FrameToggles_menu()
...
+            config.display.units = meshtastic_Config_DisplayConfig_DisplayUnits_METRIC;
+            service->reloadConfig(SEGMENT_CONFIG);
+        } else if (selected == ImperialUnits) {
+            config.display.units = meshtastic_Config_DisplayConfig_DisplayUnits_IMPERIAL;
+            service->reloadConfig(SEGMENT_CONFIG);
+        } else {
+            menuHandler::menuQueue = menuHandler::screen_options_menu;
+            screen->runNow();
+        }
+    };
```
Xaositek, line 1612: Should we add the selection for the existing selection here?

PR #8523, opened 2025-11-03
Feat/6704 neighbor info on demand

### LINKED ISSUE #6704: [Feature Request]: Pull Neighbor list remote
### Platform

NRF52

### Description

Hello.

Would it be possible to make a way to pull a neighbor list from a remote node.
I would like to see what my remote nodes can pick up on hop 0, but i don´t know any way to do that.
I tried the neighbor via lora option, but i don´t see any way to view the information.

Cheers
Brian.

DaneEvans:
Closes #6704 
Allows you to requestneighbours from a node. 

 - Uses the 'want response' field 
- Node needs to have neighbourInfo turned on, but can have the interval set to 0. 
- Does not need to have sendViaLora == true. 

Added limits to reduce retriggers: 
- Limit responses to one every 3 minutes. 
- added ignore for a dummy neighbour used solely for triggering this feature : 
```
DEBUG | ??:??:?? 464 [Router] Packet contains 1 neighbors
DEBUG | ??:??:?? 464 [Router] Neighbor 0: node_id=0x0, snr=0.00
DEBUG | ??:??:?? 464 [Router] Ignoring dummy neighbor info packet (single neighbor with nodeId 0, snr 0)
INFO  | ??:??:?? 464 [Router] NeighborInfoRequested.
```

### DIFF
--- a/bin/org.meshtastic.meshtasticd.metainfo.xml
+++ b/bin/org.meshtastic.meshtasticd.metainfo.xml
@@ -87,6 +87,9 @@
   </screenshots>
 
   <releases>
+    <release version="2.7.15" date="2025-11-13">
+      <url type="details">https://github.com/meshtastic/firmware/releases?q=tag%3Av2.7.15</url>
+    </release>
     <release version="2.7.14" date="2025-11-03">
       <url type="details">https://github.com/meshtastic/firmware/releases?q=tag%3Av2.7.14</url>
     </release>
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+meshtasticd (2.7.15.0) unstable; urgency=medium
+
+  * Version 2.7.15
+
+ -- GitHub Actions <github-actions[bot]@users.noreply.github.com>  Thu, 13 Nov 2025 12:31:57 +0000
+
 meshtasticd (2.7.14.0) unstable; urgency=medium
 
   * Version 2.7.14
--- a/version.properties
+++ b/version.properties
@@ -1,4 +1,4 @@
 [VERSION]  
 major = 2
 minor = 7
-build = 14
+build = 15


DaneEvans:
@thebentern 
I can't add labels - it's passing everything else. 

Also weigh in on whether you need me to change anything.

src/modules/NeighborInfoModule.cpp
```diff
@@ -138,13 +138,19 @@ Pass it to an upper client; do not persist this data on the mesh
...
 {
+    LOG_DEBUG("NeighborInfo: handleRecievedProtobuf");
     if (np) {
         printNeighborInfo("RECEIVED", np);
         updateNeighbors(mp, np);
     } else if (mp.hop_start != 0 && mp.hop_start == mp.hop_limit) {
         // If the hopLimit is the same as hopStart, then it is a neighbor
         getOrCreateNeighbor(mp.from, mp.from, 0, mp.rx_snr); // Set the broadcast interval to 0, as we don't know it
     }
+    if (mp.decoded.want_response && np->neighbors_count == 0) { // want response, and is otherwise empty.
```
GUVWAF: `want_response` packets are handled in `callModules()`. The cleanest way to implement this is by overriding `allocReply()`. See e.g. the NodeInfoModule for an example. In this way, the correct `request_id` is also set, which the recipient uses to link to the request.

GUVWAF: I would also not use a 0 neighbor count for this, because it lets this node think the requester has 0 neighbors. For other requests, e.g. position, we send the requester's position with it as well.

DaneEvans: Done. 
Needs clean up and some further testing, but it seems to work.

And I'll need to change the way I trigger it to avoid the empty array

DaneEvans:
got to love trunk. Possibly the worst autoformatter available. 

This is rediculous, the vsc extension runs for over an hour without doing anything. 
CLI doesn't include clang-format
It doesn't tell you what's actually wrong
And manually running clang-format apparently doesn't match what it actually wants

DaneEvans:
@GUVWAF I'm happy again. 

No issues found in testing on my network.

GUVWAF (approved):
Thanks for the updates and the perseverance with Trunk. Looks good to me now.

PR #8524, opened 2025-11-03
Fixed the issue where T-Echo did not completely shut down peripherals…

lewisxhe:
After the T-Echo enters power-off mode, you need to set the display screen as the input and PPS as the input. This can greatly reduce current consumption and turn off the GPS, PPS, and LED lights


lewisxhe:
@caveman99 Do you have any questions about this PR? It's been running for quite some time now.

thebentern (approved):

PR #8530, opened 2025-11-03
Add initial Nix shell

agustinmista:
Hey! :hand: 

This PR introduces a minimal Nix shell to setup the build environment needed by this project. This is especially useful for NixOS users, as it avoids issues caused by pip trying to install some dependencies (namely `google`) into the read-only Nix store. Instead, both PlatformIO and Python are set to store data locally under `.platformio` and `.python3`, respectively.

I tested this by entering the shell (via direnv) on my NixOS system (x86_64) without neither PlatformIO nor Python globally installed, and running `pio run -e heltec-v3` succesfully.

For the record, I might have missed some implicit dependencies that weren't needed to build the firmware of my device, but those can be easily added later if someone else runs into issues related to that.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -49,7 +49,7 @@
               # Set up PlatformIO to use a local core directory.
               export PLATFORMIO_CORE_DIR=$PWD/.platformio
               # Tell pip to put packages into $PIP_PREFIX instead of the usual
-              # location. This is especialy necessary under NixOS to avoid having
+              # location. This is especially necessary under NixOS to avoid having
               # pip trying to write to the read-only Nix store. For more info,
               # see https://wiki.nixos.org/wiki/Python
               export PIP_PREFIX=$PWD/.python3


agustinmista:
Hey @vidplace7!

This PR seems to have gone stale. Is there any chance we could revive it?

vidplace7:
@drupol do you mind reviewing this? 🙏

drupol (commented):
Let me come back with a simpler alternative.

drupol:
Find my proposal at https://github.com/meshtastic/firmware/pull/9130

I believe we shouldn't only rely on an experimental features (flakes) by default and propose a way to get it working with Nix (legacy) by default.

@agustinmista Feel free to let me know what you think of this approach as well.

agustinmista:
Thanks a lot @drupol!

I have been primarily using flakes for a while now but, indeed, having support for legacy `nix-shell` is a good idea :+1: 

I can integrate your changes into this PR and add you as a co-author via [`Co-authored-by`](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors). Does that work for you?

drupol:
Thanks for checking, no worries at all 🙂

I’m not particularly attached to my PR, so feel free to copy whatever you find useful. There’s absolutely no need to add me as a co-author.

What matters to me here is mostly the long-term picture:

* long-term reliability
* reducing the number of inputs in `flake.nix`
* avoiding `numtide/flake-utils`
* ensuring legacy `nix-shell` continues to work

I’m perfectly fine with relying on flakes for dependency locking; this just feels like a reasonable middle ground that doesn’t penalise developers who want a quick, working Nix-based development environment without fully committing to flakes.

Thanks again for the discussion and for moving this forward.

agustinmista:
Thanks again for taking the time to look into this. I agree with all your points.

I have now integrated your changes almost verbatim into this PR. Let me know if there's anything else needed on my end before getting this merged :smile_cat: 

Happy new year! :fireworks: :tada:

flake.nix
```diff
@@ -0,0 +1,66 @@
...
+
+    # Shim to make flake.nix work with stable Nix.
+    flake-compat = {
+      url = "github:NixOS/flake-compat";
+      flake = false;
+    };
+  };
+
+  outputs =
+    inputs@{ nixpkgs, ... }:
```
drupol:
```suggestion
    inputs:
```

flake.nix
```diff
@@ -0,0 +1,66 @@
...
+    let
+      lib = inputs.nixpkgs.lib;
+
+      forAllSystems =
+        fn:
+        lib.genAttrs lib.systems.flakeExposed (
+          system:
+          fn {
+            pkgs = import nixpkgs {
```
drupol:
```suggestion
            pkgs = import inputs.nixpkgs {
```

flake.nix
agustinmista: Fixed in 35ba1649544a2930d679bc998396d7ce7b84d337

flake.nix
agustinmista: Fixed in 35ba1649544a2930d679bc998396d7ce7b84d337

drupol (approved):

agustinmista:
Hey @vidplace7!

I think this PR should be ready to go, but please let me know if there's anything else missing on my end :smile:

Issue #8514, opened 2025-11-01
[Feature Request]: `CLIENT_BASE`: make it harder to "accidentally" favorite other nodes


compumike:
### Platform

Cross-Platform

### Description

Favorite nodes have a special meaning in `CLIENT_BASE` which affects routing for the local mesh. 

The goal of this feature request is to make a few small changes across firmware and clients so that it's much less likely that anyone will unintentionally/accidentally add nodes as favorites from a `CLIENT_BASE`.

(See discussion https://github.com/meshtastic/firmware/discussions/8367 for an example.)

I've implemented for firmware and iOS (see PR linked below), but I need help especially  from Android/web client developers to implement these small changes. 🙏  Let's use this issue as a central point for coordination of the concept.

-----

## The plan: reducing "accidental" favorites from `CLIENT_BASE`

`CLIENT_BASE` should require explicit add-to-favorites. Five small proposed changes:

If `CLIENT_BASE`:

1. (firmware) Don't auto-favorite from firmware.
2. (client apps) Don't auto-favorite from client apps.
3. (client apps) Show confirmation / warning dialog in client apps when explicitly adding a favorite.
4. (firmware) When configuring a device into `CLIENT_BASE`, set `is_favorite = false` for all nodes in nodeDB.
    - ⚠️  This one is controversial. It was not part of my original feature request. I'm happy to go either way on including it or not. Whichever is preferred by the project owners! 
    - It could also be deferred to later discussion so that the other four changes aren't blocked. From the perspective of the client apps, declining or deferring this item # 4 would only change the contents of warning message shown for #5.
5. (client apps) Show confirmation / warning dialog when reconfiguring a device into `CLIENT_BASE`.
With these changes, it should be much harder for anyone to unknowingly/accidentally misuse the power of `CLIENT_BASE`. 📶 

-----

## Proposed confirmation / warning messages

Please see the screenshots in https://github.com/meshtastic/Meshtastic-Apple/pull/1493 . I think a similar style and wording should be possible to implement in the other clients.

#### Clicking "Add to favorites" on a node:

Proposed text:

> Client Base should only favorite other nodes you control. Improper use will hurt your local mesh.

<img width="491" height="362" alt="Screenshot 2025-11-01 at 5 14 23 PM" src="https://github.com/user-attachments/assets/54710619-e60e-49cf-8be7-9700d6ab60c6" />

#### In device configuration, changing role to "Client Base":

Proposed text:

[...]

PR #8534, opened 2025-11-03
NodeDB: `clearAllFavorites()` when switching role to `CLIENT_BASE`

compumike:
One part of the larger feature request **[Feature Request]: CLIENT_BASE: make it harder to "accidentally" favorite other nodes** https://github.com/meshtastic/firmware/issues/8514#issuecomment-3482176166

Tested and working here, switching roles back and forth.

## 🤝 Attestations

  - [X] Heltec (Lora32) V3

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -1793,8 +1793,10 @@ void NodeDB::clearAllFavorites()
             num_changed++;
         }
     }
-    sortMeshDB();
-    saveNodeDatabaseToDisk();
+    if (num_changed > 0) {
+        sortMeshDB();
+        saveNodeDatabaseToDisk();
+    }
     LOG_INFO("clearAllFavorites: removed %d favorites", num_changed);
 }
 


fifieldt (approved):

src/mesh/NodeDB.cpp
```diff
@@ -958,6 +958,11 @@ void NodeDB::installRoleDefaults(meshtastic_Config_DeviceConfig_Role role)
         moduleConfig.telemetry.environment_update_interval = MAX_INTERVAL;
         moduleConfig.telemetry.air_quality_interval = MAX_INTERVAL;
         moduleConfig.telemetry.health_update_interval = MAX_INTERVAL;
+    } else if (role == meshtastic_Config_DeviceConfig_Role_CLIENT_BASE) {
```
shalberd, line 961:
```
 } else if (IS_ONE_OF(role, meshtastic_Config_DeviceConfig_Role_ROUTER,
                       meshtastic_Config_DeviceConfig_Role_ROUTER_LATE, meshtastic_Config_DeviceConfig_Role_CLIENT_BASE)) {
```

for also covering 0-cost hop feature, which is also based on favorites 

https://github.com/meshtastic/firmware/pull/7992/files#diff-fb9b4a4da3229ad0f8213d3e55e97cb8068e71ba0225002aa050cf06788f6b8aR107

compumike: As I haven't had time to thoroughly study the 0-cost hop feature, I don't feel comfortable putting my name on that change right now. I'd like to keep this PR minimal. Could we defer your question to a later PR?

shalberd: @h3lix1 what Mike is doing here makes a lot of sense, keeping the favorite list initially clean on role switch. Can you have a look at this and let us know whether this would be good for your case of 0 cost hop and favorites set, too?

compumike:
Marking as draft so this doesn't get merged until there's some sort of consensus one way or another whether to include this feature, per comment at https://github.com/meshtastic/firmware/issues/8514#issuecomment-3486801211

PR #8539, opened 2025-11-04
Add the Heltec v4 expansion box.

Quency-D:
1. Add the Heltec v4 expansion box.It is a TFT screen. 
2. https://heltec.org/project/wifi-lora-32-v4-expansion-housing/
3. For the OLED version, a second I2C1 is added as an external sensor interface.

  - [x] Heltec (Lora32) V3
  - [x] Heltec (Lora32) V4

### DIFF
--- a/src/SerialConsole.cpp
+++ b/src/SerialConsole.cpp
@@ -50,6 +50,7 @@ void consolePrintf(const char *format, ...)
 
 SerialConsole::SerialConsole() : StreamAPI(&Port), RedirectablePrint(&Port), concurrency::OSThread("SerialConsole")
 {
+    api_type = TYPE_SERIAL;
     assert(!console);
     console = this;
     canWrite = false; // We don't send packets to our port until it has talked to us first
--- a/src/graphics/SharedUIDisplay.cpp
+++ b/src/graphics/SharedUIDisplay.cpp
@@ -1,6 +1,8 @@
 #include "configuration.h"
 #if HAS_SCREEN
+#include "MeshService.h"
 #include "RTC.h"
+#include "draw/NodeListRenderer.h"
 #include "graphics/ScreenFonts.h"
 #include "graphics/SharedUIDisplay.h"
 #include "graphics/draw/UIRenderer.h"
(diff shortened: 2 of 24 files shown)


variants/esp32s3/heltec_v4/platformio.ini
```diff
@@ -7,3 +7,107 @@ build_flags =
   ${esp32s3_base.build_flags}
   -D HELTEC_V4
   -I variants/esp32s3/heltec_v4
+lib_deps = 
+  ${esp32s3_base.lib_deps}
+
+
+[env:heltec-v4-oled]
```
thebentern: This should be heltec-v4, as it will break the existing web-flasher and install scripts.

mverch67:
You could put
`#define CHSC6X_ADDR 0x2E`
into _configuraton.h_ and 
`SCAN_SIMPLE_CASE(CHSC6X_ADDR, CHSC6X, "CHSC6X", (uint8_t)addr.address);`
into _ScanI2CTwoWire.cpp_ and 
`CHSC6X`
into _ScanI2C.h_ to have the touch screen detection reported in the debug logs.

src/graphics/TFTDisplay.cpp
```diff
@@ -422,7 +422,54 @@ static LGFX *tft = nullptr;
...
+namespace lgfx
+{
+ inline namespace v1
+ {
+class TOUCH_CHSC6X : public ITouch
+{
+public:
+    TOUCH_CHSC6X(void)
+    {
+      _cfg.i2c_addr = 0x2E;
```
mverch67: `TOUCH_SLAVE_ADDRESS` as defined in platformio.ini?

HarukiToreda:
I would advice to also update the documentation to add the V4 with the exact i2c Pins this PR is introducing
https://meshtastic.org/docs/hardware/devices/heltec-automation/lora32/

Quency-D:
> You could put `#define CHSC6X_ADDR 0x2E` into _configuraton.h_ and `SCAN_SIMPLE_CASE(CHSC6X_ADDR, CHSC6X, "CHSC6X", (uint8_t)addr.address);` into _ScanI2CTwoWire.cpp_ and `CHSC6X` into _ScanI2C.h_ to have the touch screen detection reported in the debug logs.

I2C scanning functionality has been added.

variants/esp32s3/heltec_v4/platformio.ini
```diff
@@ -7,3 +7,109 @@ build_flags =
...
+  -D SCREEN_ROTATE
+  -D SCREEN_TRANSITION_FRAMERATE=5
+  -D BRIGHTNESS_DEFAULT=130 ; Medium Low Brightness
+  -D HAS_TOUCHSCREEN=1
+  -D TOUCH_I2C_PORT=0
+  -D TOUCH_SLAVE_ADDRESS=0x2E
+  -D SCREEN_TOUCH_INT=TOUCH_INT_PIN
+  -D SCREEN_TOUCH_RST=TOUCH_RST_PIN
+
+  -D USE_POWERSAVE
```
Quency-D: I've noticed that when using MUI, it resets every time the sleep time is up. Is this normal? If so, resetting every two minutes isn't a good experience, and I'd like to adjust it to an hour or longer.

mverch67: What do you mean by reset? The reason for e.g. 120s is that after this time the esp32 is woken up from light sleep and checks for tasks to do, e.g. sending GPS position or telemetry data and goes back to sleep after a few secs (`minimum wakeup time`). If you set sleep time to 1h it will void all other timers that may be set below 1h, e.g. GPS or device telemetry.

Quency-D: Resetting, like pressing the reset button, starts execution from the first line of the main function. Downloading the latest firmware to t-deck also starts execution from the first line of main, instead of continuing from after the sleep function.

Quency-D: Oh, thank you for your patient explanation. I misunderstood. The time cannot be set too long.

scotthmurray:
@Quency-D Hey, I just wanted to say *thank you* for this contribution to the latest release! It has made my Heltec much more usable. (I think this PR is what improved support for the display, which now shows proper emoji characters instead of garbage/unknown characters.) Thanks! 👏

ndoo:
@Quency-D @thebentern It looks like because the custom_meshtastic_* envs only get applied to env:heltec-v4 but not env:heltec-v4-tft, binaries don't get built for the expansion box product. 

Not sure if this is intentional or not, but it would be nice to get binaries for the TFT version too. 

Do let me know if I should file an issue about the above.

thebentern:
> Not sure if this is intentional or not, but it would be nice to get binaries for the TFT version too.
> 
> Do let me know if I should file an issue about the above.

It was not intentional, so please file an issue. :-)

ndoo:
Done at #9311

PR #8485, opened 2025-10-29
nrf52: add watchdog

SebKuzminsky:
This PR depends on the recently merged <https://github.com/meshtastic/Adafruit_nRF52_Arduino/pull/2>.

This PR is in good shape and is ready for review. I've been running it here locally on a rak4631 and it seems good.

There are some parts of this PR i'm uncertain about and would especially like feedback/guidance on:
* I don't love the `#include <nrfx_wdt.c>` in `main-nrf52.c`. Is there a better way to optionally compile/link C code from packages pulled in as dependencies, as `framework-arduinoadafruitnrf52` is?
* It feels like the `NRFX_*` architecture configs should go somewhere else, not at the top of `main-nrf52.c`.  `src/platform/nrf52/architecture.h` maybe?
* The slightly convoluted split initialization of the wdt (init in Setup, enable in Loop) is awkward, but i have a reason for it. The nrfx wdt API has a notion of channels, whereby different threads or execution contexts can use the wdt hardware independently of each other, by allocating their own channels and then petting that channel. These allocations have to happen after `nrfx_wdt_init()` and before `nrfx_wdt_enable()`. I'd like some day to have an architecture-agnostic way to let each thread have its own watchdog channel, which (on nrf52 at least) would require delaying enabling the wdt until all the thread setup is done. The split-setup convolution is in preparation for that future.

# 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [x] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)

PR #8554, opened 2025-11-05
Revert "nrf52: add watchdog"

Xaositek:
Reverts meshtastic/firmware#8485

Reverting this based upon discussions in Contributor Lounge and having three known good nRF52 radios (two T114, one Wio Tracker L1) suddenly start showing Critical Error 12 and 13 errors.

Examples:
https://discord.com/channels/867578229534359593/931765413350613052/1435481221420552293
https://discord.com/channels/867578229534359593/931765413350613052/1435487335340048416

### DIFF
--- a/arch/nrf52/nrf52.ini
+++ b/arch/nrf52/nrf52.ini
@@ -7,7 +7,7 @@ extends = arduino_base
 platform_packages =
   ; our custom Git version until they merge our PR
   # TODO renovate
-  platformio/framework-arduinoadafruitnrf52 @ https://github.com/meshtastic/Adafruit_nRF52_Arduino#c770c8a16a351b55b86e347a3d9d7b74ad0bbf39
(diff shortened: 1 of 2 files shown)


thebentern:
@SebKuzminsky we're reverting this for now until it's more apparent what is going on

SebKuzminsky:
Thanks for the heads up! I don't normally follow the meshtastic discord. I'll see if I can repro here and come up with a solution.

shalberd:
> reverting this based upon discussions in Contributor Lounge and having three known good nRF52 radios

Oh, ok, that means I should possibly be careful on RAK 4631 as well, eh? Not install 2.7.13, I mean. I think I'll try 2.7.13 still.

thebentern:
> > reverting this based upon discussions in Contributor Lounge and having three known good nRF52 radios
> 
> Oh, ok, that means I should possibly be careful on RAK 4631 as well, eh? Not install 2.7.13, I mean. I think I'll try 2.7.13 still.

This never made it into a release. It was merged into develop for a number days before the revert.

PR #8558, opened 2025-11-05
Don't Favorite Nodes if our Role is CLIENT_BASE

Xaositek:
- Don't Favorite Nodes if our Role is CLIENT_BASE


shalberd:
mmh, this only does not auto-favorite when sending canned message DMs.
So this the the only time in direct messaging this behavior happens? Not when sending any direct message to others from a node CLIENT_BASE?

I am not sure, but I think, generically, another PR also solves this issue:
https://github.com/meshtastic/firmware/pull/8495

PR #8560, opened 2025-11-05
Store hop/mqtt/transport mechanism info in S&F

weebl2000:
Before this, all messages received when enabling S&F server would return:

`Hops away: -1`

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [x] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Heltec wireless tracker v1.2


GUVWAF (approved):
Looks good to me. 
At some point we'd probably be better off by just storing the full MeshPacket.

PR #8562, opened 2025-11-05
Reject legacy text message DMs

jp-bennett:
I'm not thrilled to be doing this now, in a feature freeze. But I thought this had already been taken care of. If we get a message that should have been an PKI DM, reject it.


garthvh:
So devices with a yellow lock are essentially unmessagable now?

Issue #8555, opened 2025-11-05
[Improvements]: Add additional bits in the NodeDB for managed and "preferred routing" nodes


NomDeTom:
Client_base currently uses `favorite` nodes to manage their routing choices. An additional bit to mark them separately than favourites for messaging purposes would dis-ambiguate this.

Currently there is no list of nodes that a particular node may manage. Adding another bit for this would enable quick checking of nodes which are known to be previously managed.

shalberd:
Same for the 0-cost hop feature, by the way. 
https://github.com/meshtastic/firmware/pull/7992#
@h3lix1 There, too, an additional bit / field name separate from is_favorite would be of tremendous utility.

NomDeTom:
Not dead, just sleeping

NomDeTom:
Still not dead

PR #8567, opened 2025-11-06
CLIENT_BASE: Act like ROUTER_LATE for fav'd nodes, instead of like ROUTER

korbinianbauer:
As requested by @NomDeTom on discord in #transport-layer

> Hey, could you do a favour here and make the changes to firmware to make client_base act like router_late for favourite nodes, please?
> We're discussing it on the ⁠contributor-hangout , and it's agreed that this is the correct thing to do.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [X] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [X] RAK WisBlock 4631


src/mesh/FloodingRouter.cpp
```diff
@@ -124,6 +124,9 @@ void FloodingRouter::perhapsCancelDupe(const meshtastic_MeshPacket *p)
     if (config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER_LATE && iface) {
         iface->clampToLateRebroadcastWindow(getFrom(p), p->id);
     }
+    if (config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT_BASE && iface) {
```
GUVWAF: We should also check for `isFromOrToFavoritedNode()` here then.

GUVWAF:
@compumike Would like to know your thoughts on this also.

src/mesh/FloodingRouter.cpp
korbinianbauer: > We should also check for isFromOrToFavoritedNode() here then.

I thought about adding it for clarity, but it would be checking the same thing twice.

We don't get here if `roleAllowsCancelingDupe()` returns true, and that already depends on `isFromOrToFavoritedNode()`

https://github.com/meshtastic/firmware/blob/36c217857038f69d6a178419d91b5867e270aea6/src/mesh/FloodingRouter.cpp#L105-L110

I believe checking again would set us up for strange bugs in the future, even if it seems more clear now.

I could add a comment for clarity instead?

Alternatively, what would probably be the cleanest way is to introduce something like `roleCancelsOnDupe(p)` and `roleClampsToLateWindowOnDupe(p)` and fix the multiple responsibility/vagueness problem that `roleAllowsCancelingDupe()` has right now. But that'd be a bigger diff.

compumike:
@GUVWAF I didn't hear / don't see any record of the `#contributor-hangout` conversation, so I don't have context of what problem this change is trying to solve.

But in general I'm okay with this change in behavior! 👍 

Net effects of moving `CLIENT_BASE` to act in the `ROUTER_LATE` window:

- 🔴 higher latency
- 🔴 higher total airtime / channel congestion (because a message may be rebroadcasted first by a `CLIENT` and then later by the `CLIENT_BASE`)
- 🟢 slightly higher reliability of message deliverability
- 🟢 better resistance to misconfiguration (unintended favorites)

Which is a real tradeoff, but also is a totally reasonable set of tradeoffs to make!

src/mesh/FloodingRouter.cpp
compumike: I agree that we should also check `isFromOrToFavoritedNode()` (because otherwise this will unintentionally move ordinary `CLIENT`-style rebroadcasts to be later)

korbinianbauer: @compumike 

If we hit perhapsCancelDupe(), CLIENT-style rebroadcast is already off the table, no?

NomDeTom:
@korbinianbauer are you able to address the comments above?

geirgp:
@compumike you're right about the tradeoffs. To add some context on a problem this issue is helping to solve:

In sparse meshes with a "rooftop base + intermediate relay + distant node" topology, the current `CLIENT_BASE` (acting as `ROUTER` for favorites) can actually **reduce** message propagation to favorited nodes:

1. Indoor node sends message
2. `CLIENT_BASE` roof node rebroadcasts early (ROUTER priority)
3. Critical intermediate `CLIENT` node (not favorited, ~1km away) hears the `CLIENT_BASE` rebroadcast
4. Intermediate node **cancels its own rebroadcast** (managed flooding: "someone already rebroadcast this")
5. Message never reaches the wider mesh because the critical hop was suppressed

In my testing, `ROUTER_LATE` on the roof consistently solved this, while `CLIENT_BASE` performed *worse* than having no roof node at all in some cases.

However, `ROUTER_LATE` on the rooftop is [advised against by official docs](https://meshtastic.org/blog/demystifying-router-late/) because it rebroadcasts every packet it hears, significantly increasing airtime usage and potentially degrading the mesh. But since it obviously works well for this specific topology, having it replace `ROUTER` behavior in `CLIENT_BASE` (for favorited nodes only) would be ideal - giving us the late rebroadcast timing without the airtime concerns of a full `ROUTER_LATE` deployment.

korbinianbauer:
> @korbinianbauer are you able to address the comments above?

Full agree on the comments by @compumike from my side.

I also think the trade-off is absolutely worth it. The latency and ChUtil downsides are something that can and should be adressed by A) using an appropriate SF and B) throttling traffic at the origin, not by waiving reliability.

src/mesh/FloodingRouter.cpp
NomDeTom: @korbinianbauer can you take a look at this suggestion?

compumike: @korbinianbauer Ah, I believe I see what you mean! Sounds like the case of the duplicate packet arriving over LoRa is already okay. (However, I'm imagining that with some of the other transport layer discussions going on, it's possible that in the future, a CLIENT won't always cancel upon hearing a dupe. So a comment, at the very least, would be a pointer that a check might be helpful there in the future.)

However, if a `CLIENT_BASE` sees a duplicate packet arrive over non-LoRa (i.e. MQTT), then with the current PR the `CLIENT_BASE` will always move that packet from the normal CLIENT window to the ROUTER_LATE window, even if it's not to/from a favorited node. So this behavior is not 100% consistent with the "behaves like CLIENT for non-favorited nodes".  If you were to add the check for `isFromOrToFavoritedNode()` in here, then it will keep that 100% consistency, as far as I can tell.

With the goal of 100% consistency with the statement "CLIENT_BASE behaves exactly like CLIENT for non-favorited nodes", I'd probably still suggest adding the extra `isFromOrToFavoritedNode()` check here, even though it burns some extra CPU time for a CLIENT_BASE. However, I admit this is really an edge case (dupe received over MQTT) and the consequences don't seem too bad (only moves it to ROUTER_LATE window).

korbinianbauer: @compumike Got you. I have to admit I didn't think too hard about MQTT and saving that extra check is not a hill I need to die on :)

korbinianbauer (commented):
Added extra check for fav'd source for consistency

compumike (approved):

compumike:
This PR looks good and I think it can be merged 👍 

---

Reviewing the larger section of `ROUTER_LATE` behavior, I have a question for @GUVWAF :

#### Should `ROUTER_LATE` rebroadcast late always, or only after receiving dupe?

With regards to @geirgp 's "critical intermediate node" example:

My understanding is that `ROUTER_LATE` (and after this PR, `CLIENT_BASE`) will initially receive a packet and schedule a rebroadcast during the normal `CLIENT` window. They'll only move the rebroadcast to the later `clampToLateRebroadcastWindow` once they hear that packet a _second_ time! (Because `clampToLateRebroadcastWindow` is only called from `perhapsCancelDupe`.)

This means there's still a _random race_ between the `ROUTER_LATE` roof node and the `CLIENT` intermediate node. ~50% of the time, the roof node will still win, suppressing the intermediate node.

Instead, what if `ROUTER_LATE` (and `CLIENT_BASE`, when from/to favorite) _always_ enqueued their rebroadcasts with `clampToLateRebroadcastWindow`, even when they've only heard a packet once?

Slightly higher latency, but this would make it so that in this example, the intermediate node would never be suppressed by the roof node.

I can imagine:

- `bool RadioInterface::shouldRebroadcastLateLikeRouterLate(meshtastic_MeshPacket *p)`
- modify `RadioInterface::getTxDelayMsecWeighted` to check it and add delay accordingly

(I don't think this question should delay merging this PR.)

korbinianbauer:
@compumike 

> Slightly higher latency, but this would make it so that in this example, the intermediate node would never be suppressed by the roof node.

Note that the "Late" Window is only guaranteed to be later than the Client window at similar SNR.

Even if Router_Late was to **always** rebroadcast in the late window, there is a good chance it will still supress a Client.

See Point 4 in #8431

compumike:
> @compumike 
> 
> 
> 
> > Slightly higher latency, but this would make it so that in this example, the intermediate node would never be suppressed by the roof node.
> 
> 
> 
> Note that the "Late" Window is only guaranteed to be later than the Client window at similar SNR.
> 
> 
> 
> Even if Router_Late was to **always** rebroadcast in the late window, there is a good chance it will still supress a Client.
> 
> 
> 
> See Point 4 in #8431 

Ah, thanks @korbinianbauer  -- I had not seen that in #8431!

aerodan:
The following scenario still exists. Though the effect is now downgraded to an accidentally favorited node acting like router_late instead of router, it's still an unwanted outcome and a mesh nuisance.

1) User configures Client_Base hoping to have a local or onsite router in their control. 

2) The same user sends a DM to a nearby direct node. That node is auto-favorited and now acts as router_late despite not being qualified to do so. 

Fsvorites are a poor way to define a router, as users commonly see it as "these are my friends". This still appears to have mesh health considerations.

NomDeTom:
> The following scenario still exists. Though the effect is now downgraded to an accidentally favorited node acting like router_late instead of router, it's still an unwanted outcome and a mesh nuisance.
> 
> 1. User configures Client_Base hoping to have a local or onsite router in their control.
> 2. The same user sends a DM to a nearby direct node. That node is auto-favorited and now acts as router_late despite not being qualified to do so.

Not quite. DMs no longer auto-favourite on the client apps. This was an app-based behaviour. The alternative to this current pr was to pull the feature, which is currently a new favourite (pun intended) role.

> Fsvorites are a poor way to define a router, as users commonly see it as "these are my friends". This still appears to have mesh health considerations.

Yes, we needed a quick and easy way to get started rather than bamboozle users with another nother tag or list.

aerodan:
> > The following scenario still exists. Though the effect is now downgraded to an accidentally favorited node acting like router_late instead of router, it's still an unwanted outcome and a mesh nuisance.
> > 
> > 1. User configures Client_Base hoping to have a local or onsite router in their control.
> > 2. The same user sends a DM to a nearby direct node. That node is auto-favorited and now acts as router_late despite not being qualified to do so.
> 
> Not quite. DMs no longer auto-favourite on the client apps. This was an app-based behaviour. The alternative to this current pr was to pull the feature, which is currently a new favourite (pun intended) role.
> 
> > Fsvorites are a poor way to define a router, as users commonly see it as "these are my friends". This still appears to have mesh health considerations.
> 
> Yes, we needed a quick and easy way to get started rather than bamboozle users with another nother tag or list.

Would it be a quick and dirty at-least-for-now fix to have a small table of a static nodes (max of lets say five) stored on the router, that are tagged as My_Base_Clients? Then the act of Favoriting is no longer burdened with special powers. Only those nodes as statically set by the Client_Base administrator will ever be used as "My_Base_Clients" and thus routed for... A command-line only option can set them for the time being. Not elegant, but also not as destructive and much more intentional. The log could spell out those nodes each boot, and perhaps later UI built where those hosts can be shown with an icon in the UI, selected from the node list etc. So: 1_ User enables Client_Base and become a potential router admin. 2_ New router admin sets up to five nodes as My_Base_Clients using their hardware node identifiers via command line. 3_ Favoriting nodes now has nothing to do with routing. The log reminds Client_Base administrators that those nodes are set on each boot. This would carry the feature until a protocol method of handing it is invented, and probably has UI object reuse.

NomDeTom:
See https://github.com/meshtastic/firmware/issues/8555

shalberd:
Thank you all for the hard and thoughtful work on this. I am watching your efforts a lot and like how you are willing to alter the behavior of features. In our case here in CH, moving favorites-handling to the late rebroadcast window really helps avoid some of the, probably due to our node density and topography, unwanted side effects of previous early reboradcast window. Thank you, truly.
Also, great to see you are thinking of using a field separate from is_favorite in future potentially for 0-cost-hop and preferred-client_base nodes.

Hamberthm:
Hi! As far as I understand, this changes apply only to DMs to and from a fav node.
I`ve been losing messages on LongFast that won´t reach my indoor node, because my roof node won´t rebroadcast them for some reason. This makes the function of CLIENT_BASE only valid and useful for DMs.

I propose that in CLIENT_BASE ALL broadcast channel packets should be rebroadcasted at least ONE MORE TIME, so to ensure proper reach of LongFast messages and the like to the indoor nodes.

For packets that the algorithm decided should not be rebroadcasted, I propose to do so BUT after subtracting the remaining hops on the packet (making it 0) so it won´t congest the mesh but ensure reach to the indoor nodes.

korbinianbauer:
@Hamberthm 

This PR only changes the timing, not if a packet is rebroadcast or not.
Proposals along these lines are out there though, maybe look into the discussion page about that.

erayd:
> I propose that in CLIENT_BASE ALL broadcast channel packets should be rebroadcasted at least ONE MORE TIME, so to ensure proper reach of LongFast messages and the like to the indoor nodes.

This would cause congestion problems on many meshes. It's not feasible without significant throttling, and if you're throttling traffic you end up with exactly the same issue (i.e. some broadcast traffic not being rebroadcast). There simply isn't enough bandwidth available to have everybody's roof node rebroadcast every packet.

> For packets that the algorithm decided should not be rebroadcasted, I propose to do so BUT after subtracting the remaining hops on the packet (making it 0) so it won´t congest the mesh but ensure reach to the indoor nodes.

Not only would this still congest the mesh (sending a packet as zero-hop still requires just as much airtime to transmit), it actually *breaks* the mesh by preventing any node that hears it from rebroadcasting. Remember that the nodes inside your house are not the only ones that can hear your CLIENT_BASE; even a very weak signal can still make it halfway across a city fairly easily.

Hamberthm:
> > I propose that in CLIENT_BASE ALL broadcast channel packets should be rebroadcasted at least ONE MORE TIME, so to ensure proper reach of LongFast messages and the like to the indoor nodes.
> 
> This would cause congestion problems on many meshes. It's not feasible without significant throttling, and if you're throttling traffic you end up with exactly the same issue (i.e. some broadcast traffic not being rebroadcast). There simply isn't enough bandwidth available to have everybody's roof node rebroadcast every packet.
> 
> > For packets that the algorithm decided should not be rebroadcasted, I propose to do so BUT after subtracting the remaining hops on the packet (making it 0) so it won´t congest the mesh but ensure reach to the indoor nodes.
> 
> Not only would this still congest the mesh (sending a packet as zero-hop still requires just as much airtime to transmit), it actually *breaks* the mesh by preventing any node that hears it from rebroadcasting. Remember that the nodes inside your house are not the only ones that can hear your CLIENT_BASE; even a very weak signal can still make it halfway across a city fairly easily.

Thanks for the detailed answer, got it. The complexity of this system amazes me everytime. 

I'm thinking on alternatives like the indoor node sending some type of hash to the roof node to check if it has the same last messages, kind of an S&F server thing. But better I shut up and let the people who know to work on it. 

I hope some workaround can be found, as community messages (LongFast) are important, even more on emergency situations.

shalberd:
> thinking on alternatives

Hi Humberto

personally, I would go with a solution directly connecting to the rooftop node via a special, high-quality bluetooth antenna or even WLAN, depending on the board. That is, not using the indoor node at all. Advantage: that way, you can just configure the rooftop node as CLIENT

For example, I am using this one https://www.data-alliance.net/antenna-2-4ghz-omnidirectional-2-5dbi-weatherproof-rp-sma-thru-hole-mount-wifi-bluetooth/  really just an example, there are so many great antennas for Bluetooth. 50-100m range should be possible, being close to the window, for example, with your smartphone.

and it has much, much better bluetooth range, even when tucked inside the plastic case, not on the outside.

erayd:
> I'm thinking on alternatives like the indoor node sending some type of hash to the roof node to check if it has the same last messages, kind of an S&F server thing.

You're describing a feature I've been working on since September 😉. It's not intended for roof nodes though - if you use it that way, it'll throttle fairly aggressively. 

Remember that the whole point of CLIENT_BASE is to *not* relay all broadcast traffic, to prevent the kind of congestion that's seen when people put routers on their roofs. The default LONG_FAST modulation has a data rate equal to roughly 2% of an old dialup modem, and that bandwidth is *shared* between every node in the same area. The unfortunate consequence of such limited bandwidth is the need to ration it quite aggressively in order to keep everything working effectively, as without such rationing it is very quickly exhausted.

> I hope some workaround can be found, as community messages (LongFast) are important, even more on emergency situations.

The workaround that actually scales well is to place the node you want to receive on somewhere that can hear other mesh nodes without needing a roof node to relay to it. Having your phone connected directly to a client node on your roof achieve this quite effectively. I do appreciate that it's cumbersome though.

Hamberthm:
> > thinking on alternatives
> 
> Hi Humberto
> 
> personally, I would go with a solution directly connecting to the rooftop node via a special, high-quality bluetooth antenna or even WLAN, depending on the board. That is, not using the indoor node at all. Advantage: that way, you can just configure the rooftop node as CLIENT
> 
> For example, I am using this one https://www.data-alliance.net/antenna-2-4ghz-omnidirectional-2-5dbi-weatherproof-rp-sma-thru-hole-mount-wifi-bluetooth/  really just an example, there are so many great antennas for Bluetooth. 50-100m range should be possible, being close to the window, for example, with your smartphone.
> 
> and it has much, much better bluetooth range, even when tucked inside the plastic case, not on the outside.
> 
> 
> 

Indeed, I designed my roof node to be in reach of bluetooth for most of the house, sacrificing some perfomance by using coax to lower the node 2m into a place where I can connect to it. I just discovered by accident that some messages weren't reaching my portable node inside the house. So, even when this is not a problem for me, I was just thinking about people living in apartment buildings and the like who may have roof nodes to serve a bunch of lower nodes. 

This may or may not be a real problem, I just got worried by it :)

shalberd:
> I was just thinking about people living in apartment buildings and the like who may have roof nodes to serve a bunch of lower nodes.

Here is Switzerland, we try to encourage people to use CLIENT_MUTE in the city and then to place their own apartment / client_mute node on the side of the apartment building close to the window that faces a big mountain with a CLIENT node :-) No need for using a rooftop node that way to connect to the mesh.

Thanks to both of you for an interesting discussion, I can relate to what you are worried about / what you mentioned,.

PR #8576, opened 2025-11-06
Add API types, state, and log message in Debug screen. Added persistent "Connected" icon

jp-bennett:
This adds some simulated type reflection to PhoneAPI, then adds code to track which API endpoint is connected, in MeshService. And finally adds a line to the debug screen, to reflect whether the API is connected, and what type.

### DIFF
--- a/src/graphics/draw/DebugRenderer.cpp
+++ b/src/graphics/draw/DebugRenderer.cpp
@@ -646,10 +646,9 @@ void drawSystemScreen(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t x
     int textWidth = display->getStringWidth(appversionstr);
     int nameX = (SCREEN_WIDTH - textWidth) / 2;
 
-    display->drawString(nameX, getTextPositions(display)[line], appversionstr);
-#if !defined(M5STACK_UNITC6L)
-    if (SCREEN_HEIGHT > 64 || (SCREEN_HEIGHT <= 64 && line < 4)) { // Only show uptime if the screen can show it
-        line += 1;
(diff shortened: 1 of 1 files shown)


Xaositek (approved):
BaseUI pieces look good!

src/graphics/draw/DebugRenderer.cpp
```diff
@@ -660,8 +661,26 @@ void drawSystemScreen(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t x
             snprintf(uptimeStr, sizeof(uptimeStr), " Uptime: %um", mins);
         textWidth = display->getStringWidth(uptimeStr);
         nameX = (SCREEN_WIDTH - textWidth) / 2;
-        display->drawString(nameX, getTextPositions(display)[line], uptimeStr);
+        display->drawString(nameX, getTextPositions(display)[line++], uptimeStr);
     }
+    // line++;
+    char api_state[50] = {0};
+    if (service->api_state == service->STATE_BLE)
+        snprintf(api_state, sizeof(api_state), "API State: BLE Connected");
```
thebentern: Can we truncate these to API: BLE Connected? I feel like it's pretty intuitive we're exposing the state.

thebentern (approved):

src/graphics/draw/DebugRenderer.cpp
jp-bennett: I've tickled the wording a bit locally. Jason is talking about an icon instead of text, too.

jp-bennett:
Looks like the reset code is pretty goobered, will fix when back at a desktop

PR #8578, opened 2025-11-07
Don't ack messages when mqtt client proxy is on but only uplink

RCGV1:
When sending messages from a client using MQTT client proxy enabled all messages will instantly ack even if only uplink is enabled. I modified the firmware to discard packets being sent to the firmware if down link is off. 

This is a longstanding bug that prevented messaging from being useful while being connected to MQTT with uplink on.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -59,7 +59,13 @@ inline void onReceiveProto(char *topic, byte *payload, size_t length)
         LOG_ERROR("Invalid MQTT service envelope, topic %s, len %u!", topic, length);
         return;
     }
+
     const meshtastic_Channel &ch = channels.getByName(e.channel_id);
+    // Find channel by channel_id and check downlink_enabled
+    if (!(strcmp(e.channel_id, "PKI") == 0 ||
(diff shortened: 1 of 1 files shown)


RCGV1:
IMG_5105

GUVWAF (approved):

RCGV1:
@GUVWAF Just realized I didn't account for dms, should we also throw out dms from this function. So you also don't get implicit acks for dms.

GUVWAF:
Yes, would probably be good. 
But to be honest this is actually an issue with the client proxy. With MQTT on the device it will not subscribe to channels it hasn't downlink configured for.

RCGV1:
I think it would be more difficult to implement this on android and iOS as the client proxy is pretty blind to the raw packets and what channels the packets are coming in on.

RCGV1:
@GUVWAF I think the issue with dropping packets if its a dm is that I dont know when we downlink dms. Is it when primary channel downlink is on?

RCGV1:
Since if I blindly turned it off this would disable over mqtt dms

GUVWAF:
It's when there is at least one channel with downlink enabled.

PR #8584, opened 2025-11-07
modules: fix broadcasting neighbor info

### LINKED ISSUE #8581: [Bug]: NeighborInfoApp stopped broadcasting over LoRa
### Category

Other

### Hardware

Not Applicable

- [ ] InkHUD ePaper

### Firmware Version

2.7.10, 2.7.11, 2.7.13

### Description

I'm maintaining a node monitoring dashboard that relies on nodes broadcasting their neighbor info over lora.

We have around 300 active nodes in the area. I am running 6 nodes myself with different hardware (RAK, LILYGO, Raspberry Pi,...).

Around the time of the release of 2.7.10 or 2.7.11 I noticed neighbor info to gradually disappear from the mesh. This includes the nodes I maintain. 

I have primary and secondary channels, I have reduced the broadcast interval to 7200 seconds, and I enabled the LoRa checkbox.

But it wouldn't get broadcasted and I see less and less neighbor info on our mesh. I'm at this point fairly confident that the neighbor info app module might be bugged.

Any idea what change has caused this?

### Relevant log output

```Shell

```

l5yth:
- ~~NeighborInfoModule now schedules its first run 35 s after boot as documented, instead of waiting for the full configured interval.~~ Scratched that.
- Neighbor broadcasts are emitted every cycle regardless of the currently known neighbor count, with an inline comment explaining why we keep sending empty payloads to bootstrap future discovery. This reverts #7493
- Neighbor cleanup is more forgiving: entries fall back to min_node_info_broadcast_secs when no interval is known and are only purged after 4× that duration, preventing premature eviction when RX is spotty.
- getOrCreateNeighbor() now normalizes zero-interval updates to min_node_info_broadcast_secs, ensuring we never store a 0 s interval that would trigger immediate cleanup, while still honoring explicit intervals from peers.
- attempt to fix #8581 

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

~~These changes are untested as I want to gather feedback on the proposed changes first.~~ I'm currently testing this on my hardware. :hourglass_flowing_sand:

### DIFF
--- a/src/modules/NeighborInfoModule.cpp
+++ b/src/modules/NeighborInfoModule.cpp
@@ -34,7 +34,7 @@ void NeighborInfoModule::printNodeDBNeighbors()
     }
 }
 
-/* Send our initial owner announcement 35 seconds after we start (to give network time to setup) */
+/* Neighbor info broadcasts start after the configured interval so we honor user pacing */
 NeighborInfoModule::NeighborInfoModule()
     : ProtobufModule("neighborinfo", meshtastic_PortNum_NEIGHBORINFO_APP, &meshtastic_NeighborInfo_msg),
       concurrency::OSThread("NeighborInfo")
@@ -44,7 +44,8 @@ NeighborInfoModule::NeighborInfoModule()
 
     if (moduleConfig.neighbor_info.enabled) {
         isPromiscuous = true; // Update neighbors from all packets
-        setIntervalFromNow(35 * 1000); // Send our initial broadcast roughly 35 seconds after boot
+        setIntervalFromNow(Default::getConfiguredOrDefaultMs(moduleConfig.neighbor_info.update_interval,
+                                                             default_telemetry_broadcast_interval_secs));
     } else {
         LOG_DEBUG("NeighborInfoModule is disabled");
         disable();


src/modules/NeighborInfoModule.cpp
```diff
@@ -44,8 +44,7 @@ NeighborInfoModule::NeighborInfoModule()
 
     if (moduleConfig.neighbor_info.enabled) {
         isPromiscuous = true; // Update neighbors from all packets
-        setIntervalFromNow(Default::getConfiguredOrDefaultMs(moduleConfig.neighbor_info.update_interval,
-                                                             default_telemetry_broadcast_interval_secs));
+        setIntervalFromNow(35 * 1000); // Send our initial broadcast roughly 35 seconds after boot
```
thebentern: I don't think this is a good idea. This is likely to fill the neighborinfo with innacurate stale entries in highly mobile meshes. Part of the rationale of waiting for the longer interval is to allow for a fresh perspective on who the node's neighbors are.

korbinianbauer: In the spirit of this PR, would it be the right thing to send an empty neighbor info 35s after boot, though?

l5yth: Yeah, on a second thought, I will probably reset this back to the configured default until we have a better idea how to do this. The other changes in this PR are more relevant for fixing the underlying issue.

l5yth:
Testing of this patch was inconclusive, so I will close this for now.

DaneEvans:
Testing on the 'request' functionality seemed similar to this - before I realised that a few were on default primary channels. 
After fixing that, the lists got populated as I expected.

PR #8587, opened 2025-11-07
Refactor RadioLibInterface to support multiple instances

Andrik45719:
These changes make it possible to determine which LoRa module triggered the interrupt by reading the GPIO state.
This allows for creating multiple instances of RadioLibInterface.

https://github.com/meshtastic/firmware/discussions/3831

## 🤝 Attestations

  - [x] Heltec (Lora32) V4

### DIFF
--- a/src/mesh/RadioLibInterface.cpp
+++ b/src/mesh/RadioLibInterface.cpp
@@ -39,7 +39,7 @@ RadioLibInterface::RadioLibInterface(LockingArduinoHal *hal, RADIOLIB_PIN_TYPE c
                                      RADIOLIB_PIN_TYPE busy, PhysicalLayer *_iface)
     : NotifiedWorkerThread("RadioIf"), module(hal, cs, irq, rst, busy), iface(_iface)
 {
-    instance = this;
+    instances.push_back(this);
 #if defined(ARCH_STM32WL) && defined(USE_SX1262)
     module.setCb_digitalWrite(stm32wl_emulate_digitalWrite);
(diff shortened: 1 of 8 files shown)


src/mesh/RadioLibInterface.cpp
```diff
@@ -78,7 +84,7 @@ void INTERRUPT_ATTR RadioLibInterface::isrTxLevel0()
 
 /** Our ISR code currently needs this to find our active instance
  */
-RadioLibInterface *RadioLibInterface::instance;
+std::vector<RadioLibInterface*> RadioLibInterface::instances;
```
mverch67, line 87: Because this is a static variable you have to explicitly remove instances that have been pushed into this vector when the instance (i.e. an instance of a derived class of RadioLibInterface) is deleted as in [main.cpp](https://github.com/meshtastic/firmware/blob/77e0a24838ffab446742af88799d7bc9cc7110af/src/main.cpp#L1263) where several radio chip types are probed.

Andrik45719: I currently have two radio interfaces running. The variable must be static, as it needs to be accessed from an interrupt.
The problem is that there is not enough IRAM memory on the ESP32, I will solve it.

mverch67: Still, you need to remove instances from the vector in the dtor, otherwise the vector will contain pointers to deleted objects.

Andrik45719: Hi, I’ve opened a new PR, could you check it?
https://github.com/meshtastic/firmware/pull/8607

PR #8596, opened 2025-11-08
Add Promiscuous mode for serial

tsuckow:
Adds a mode for the serial protocol to get packets unaltered (encrypted) FromRadio and allow arbitrary packets to be sent ToRadio.

This is useful for creating virtual nodes or bridging traffic between frequencies.

Protobuf PR https://github.com/meshtastic/protobufs/pull/804

## 🤝 Attestations

  - https://github.com/tsuckow/meshtastic_bridge
  - [X] Heltec T114

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -187,7 +187,10 @@ void MeshService::handleToRadio(meshtastic_MeshPacket &p)
         return;
     }
 #endif
-    p.from = 0;                          // We don't let clients assign nodenums to their sent messages
+    // Allow any encrypted packet as-is in promiscuous mode
+    if (!(serialPromiscuousEnabled && p.which_payload_variant == meshtastic_MeshPacket_encrypted_tag)) {
+        p.from = 0; // We don't let clients assign nodenums to their sent messages
+    }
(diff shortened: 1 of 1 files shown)


tsuckow:
May cause problems with next hop routing since ToRadio will have the relay_node overwritten and FromRadio is before routing is handled. Not a huge issue, but need to think if this is something better solved in firmware or should be responsibility of software interfacing with it

thebentern:
So this setting would not be a persisted config? It seems like an admin message rather than a ToRadio flag would be a more consistent approach for turning this on

tsuckow:
But wouldn't an admin message allow it to be sent over Lora? I don't believe we should be able to send over Lora because this is changing the behavior of the serial connection for the duration of the connection. If something else connects over serial the state resets.

tsuckow:
> May cause problems with next hop routing since ToRadio will have the relay_node overwritten and FromRadio is before routing is handled. Not a huge issue, but need to think if this is something better solved in firmware or should be responsibility of software interfacing with it

I'm still investigating some strange behavior with pki encrypted messages. I can't tell if the packet is being altered in some problematic way or if I've actually found a bug in the firmware that's unrelated to this change.

PR #8612, opened 2025-11-11
Fix multicast address

### LINKED ISSUE #7674: [Bug]: Mesh-UDP using reserved multicast address
### Category

WiFi

### Hardware

Heltec V3

- [x] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper

### Firmware Version

2.7.5

### Description

Mesh to UDP is using address of 224.0.0.69 which is a reserved multicast address, its reserved for internal IGMP protocols like mDNS, HSRP, etc.

The suggestion would be to use 239.0.0.0-239.255.255.255 (excluding 239.128.0.0/16) range so that it can be routed in and out of the wireless. 

Can we change it to 239.0.0.69? It'd make the world of difference.

Source, I'm a Cisco CCIE Network engineer. Using Cisco Access Points and switching.  Multicast flooding was enabled. I even tested it on a GL-iNet travel router and it dropped the multicast traffic. (Wired->Sta)

All wireless meshtastic radios were able to get the traffic injected into the wired network, but it wasn't allowed to be routed out b/c of the 224.0.0.69 multicast address. Moving this to 239.0.0.69 would permit two ESP32 to connect to the wireless and broadcast to each-other.

**REF:**
https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml

The range of addresses between 224.0.0.0 and 224.0.0.255, inclusive,
is reserved for the use of routing protocols and other low-level
topology discovery or maintenance protocols, such as gateway discovery
and group membership reporting.  Multicast routers should not forward
any multicast datagram with destination addresses in this range,
regardless of its TTL.

**REF:**
224.0.0.69 reserved for all-AODV-RPL-nodes

Image

Wired capture of UDP packets:
They get replicated within the same vlan (as expected for 224.0.0.0/24 traffic) but drops at the access points.
Image

### Relevant log output

```Shell

```

sjm42:
Fixes #7674 

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default multicast network address (from 224.0.0.69 to 239.0.0.69) to improve reliable message routing.
  * Multicast listening and sending now use the corrected group address for consistent communication.


Jorropo:
It "works" but this is very break everything and see if anyone cares kind of approach.

I think a better solution is to instantiate two `UdpMulticastHandler`, one with the old IP and one with the new.
Add docs saying the old IP is deprecated and remove it in 2~3 versions.

Edit: should be <10 SLOC change (make the IP an argument to `UdpMulticastHandler` and create two of them rather than one).

sjm42:
Would that solution mean just sending all the udp packets to two multicast addresses, the old and new one,
until the old one is finally deprecated? That should work, given that the packet rate is quite small even in the "worst" case with meshtastic. Any LAN should be easily able to handle that.

Jorropo:
This is exactly what I have in mind :slightly_smiling_face:

sjm42:
Well after looking at the code including `main.cc` and because `udpIpAddress` is a private field of the class `UdpMulticastHandler`,
I would rather suggest not changing the caller side at all but just rename the current `udpIpAddress`
to something like `udpIpAddressDeprecated` and creating a new `udpIpAddress` with the more correct address
in the constructor. Also just add another `udp.writeTo()` line to handle both destinations.

How would that sound? I can make a PR suggestion of this approach.

Edit: I realized I did not check how the udp receiving side is arranged. With repeating the udp.listenMulticast() call we would get everything as doubled on the receiving side, sigh. What would be best approach for receiving? For now on, use the old udp address for receiving, and "halfway" into old address deprecation, move on and only listen for the new address? And then finally, remove the deprecated sender address also?

Jorropo:
You will need a second listener too.

It seems weird to me to do it here but I doubt I care.
Arguably this is a good candidate to here or even lower (add a function that takes multiple IPs and send the packet to each IP) since Linux has more efficient APIs to send an identical packet to multiple destinations, altho, altho this run at kbps at most so doesn't matter.

sjm42:
If the code starts to send multicast to addresses a+b and also the receiving side would be listening to a+b and both nodes are running the new firmware, every packet would be received as duplicated.

One stupid "workaround" might be to first listen for both addresses and after first packet is seen on either address,
mark the other address internally to be ignored in the future. I am not familiar enough with the internals to be able to just shut down the other unnecessary listener afterwards.

Another option would be making the multicast address configurable and optional. The current default would be the old address naturally, but could be changed. Probably that would be way too much hassle to introduce something new to be configured, idk.

Jorropo:
> and both nodes are running the new firmware, every packet would be received as duplicated

You can give duplicated packets to the router and it will handle them.
I remember this used to cause issues but I am pretty sure Jonathan fixed it.

> One stupid "workaround" might be to first listen for both addresses and after first packet is seen on either address,
mark the other address internally to be ignored in the future.

We don't want that, it is not gonna help us transition since unless you restart all your nodes with the new code they are still all gonna use the old multicast address.
You can do some filtering on a per packet basis, keep a buffer of like 10 packets hashes and don't send packets to the router if we already seen them.

But turns out the router already implement this logic so ... I would try letting both listeners callback into the router with the same packets and letting it dedup.

sjm42:
Okay, if it is not harmful to send duplicates (other than that statistics get distorted?),
this should be quite simple. Just listen to two multicast addresses and only use the existing `onReceive()` handler.

On the sending side, just send the packet once for each destination address.
Is it safe to call `udp.listenMulticast()` twice and both multicast addresses are then subscribed to?

If that is yes, then this should work?

[https://github.com/meshtastic/firmware/compare/develop...sjm42:meshtastic-firmware:multicast-fix-take2](https://github.com/meshtastic/firmware/compare/develop...sjm42:meshtastic-firmware:multicast-fix-take2)

edit: oh, and also I am assuming that the `udp.writeTo()` is not modifying the send buffer at all. Is this correct?
The code is just sending exactly the same contents from the same buffer into two different destination multicast addresses.

Jorropo:
> Is it safe to call udp.listenMulticast() twice and both multicast addresses are then subscribed to?
> If that is yes, then this should work?

On linux I wrote the code so it should work.
For ESP32 idk try it, I hope so it's not like 2 is a big number.

> edit: oh, and also I am assuming that the udp.writeTo() is not modifying the send buffer at all. Is this correct?
> The code is just sending exactly the same contents from the same buffer into two different destination multicast addresses.

It would be beyond insane if writeTo mutate the passed-in buffer.
writeTo is inspired from POSIX semantics and they clearly state functions like write, send, writeTo cannot modify the input buffer.

sjm42:
Well if my proposed changes make sense, I would make a new pull request or what? :)

Jorropo:
@sjm42 usually you would force push this branch to avoid sending more notifications to people watching this repo but both work.

f4dvk:
I would like to test it because the address 224.0.0.69 does not work with my wifi access point.

Thanks,
Stephane.

Xaositek (approved):
This seems like a good change for 2.8, fixed merge conflict due to code changes. Still targeting 239 over 224.

Xaositek:
Just some extra words of thought here - overall this change is low effort, high reward - we are putting ourselves in a great place long-term for compatibility, I don't feel like a migration effort for this is necessary. We make a clean break and everyone wins. Have a banner in the release notes stating this address is changing for the good, if folks ultimately need the old address, they can tweak a local build.

jp-bennett (approved):
Might as well break it for 2.8, as there's already interop issues between 2.7 and 2.8

PR #8633, opened 2025-11-13
Don't broadcast the MAC address by default

vd994:
Fixes #8336.

- Add config.device.send_mac_address in NodeDB.cpp
- Randomize the nodenum on first boot
- A factory reset generates a new nodenum.

This is a proof of concept. The config variable needs to be added in protobufs in order for the build to be successful.

This significantly improves privacy and slightly improves security.
Simply removing the option to share the MAC address, or making it a variant flag, is an acceptable alternative.
Further possible step: remove macaddr from nodeinfo

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "protobufs"]
 	path = protobufs
-	url = https://github.com/meshtastic/protobufs.git
+	url = https://github.com/vd994/protobufs.git
+	branch = no-mac-address
 [submodule "meshtestic"]
 	path = meshtestic
 	url = https://github.com/meshtastic/meshTestic
--- a/src/mesh/generated/meshtastic/config.pb.h
+++ b/src/mesh/generated/meshtastic/config.pb.h
@@ -357,6 +357,7 @@ typedef struct _meshtastic_Config_DeviceConfig {
     /* Controls buzzer behavior for audio feedback
  Defaults to ENABLED */
     meshtastic_Config_DeviceConfig_BuzzerMode buzzer_mode;
+    bool send_mac_address;
 } meshtastic_Config_DeviceConfig;
 
 /* Position Config */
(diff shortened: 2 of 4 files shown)


fifieldt:
Note for reviewers: this is considered a breaking change and requires community discussion.

thebentern:
Thank you for the PR, but we will be addressing this in 2.8 and have already started on an option to derive the nodenum from the pubkey rather than the mac address.

vd994:
> While it improves the security, reduces the easy to use factor. Users will have a "new node" on each reset / flash when erasing. There will be a lot of old nodes.

Agreed, however it could be a option when resetting instead. Old nodes which get deleted eventually are worth having for people's privacy.

> Also opens up a new attack vector: "This is ME, i just resetted my node. Trust me." Just needs to copy the long+short name. I would suggest to use a hash function to generate a nodenum from the mac. So the mac would be still hidden, but the node id would remain the same.

Yes, but not one which Meshtastic defends against currently. Correct me if I'm wrong, but anyone can already change their MAC address and/or NodeID to an existing user's and say, "I reset my node/key." Checking keys is the only way to reasonably verify someone's identity.

> I would suggest to use a hash function to generate a nodenum from the mac. So the mac would be still hidden, but the node id would remain the same.

This wouldn't prevent tracking over time. People should at least have the option to randomize the NodeID.

vd994:
> Thank you for the PR, but we will be addressing this in 2.8 and have already started on an option to derive the nodenum from the pubkey rather than the mac address.

Thank you for your work on that. Should I propose the MAC address parts of this on the XEdDSA branch instead? NodeID randomization is just a way to make things work until private key based NodeIDs are implemented.

PR #8636, opened 2025-11-14
Always Zero-Hop MQTT traffic

vidplace7:
*Always* zero hop packets that originate from MQTT (Downlink).

This allows users to continue to use MQTT *on node*, while leaving the precious airwaves alone.

Changing this behavior *should* be extremely high friction (modify and compile your own firmware, not a standard config) as it's risk for abuse (intentional or accidental) is extremely high.

The Meshtastic Public MQTT, as well as many private-mesh-group MQTT servers are Zero-Hopped for this reason (high abuse potential). Let's end this madness everywhere.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -108,7 +108,7 @@ inline void onReceiveProto(char *topic, byte *payload, size_t length)
     p->to = e.packet->to;
     p->id = e.packet->id;
     p->channel = e.packet->channel;
-    p->hop_limit = e.packet->hop_limit;
+    p->hop_limit = 0; // Always zero-hop MQTT traffic
     p->hop_start = e.packet->hop_start;
     p->want_ack = e.packet->want_ack;
     p->via_mqtt = true; // Mark that the packet was received via MQTT


Xaositek:
This has moved into a space where **users have zero control** and anyone who needs to actually allow hopped MQTT where an MQTT bridge is necessary is **FORCED** to build firmware.

wehooper4:
> This has moved into a space where **users have zero control** and anyone who needs to actually allow hopped MQTT where an MQTT bridge is necessary is **FORCED** to build firmware.

This is how it really needs to be.

I've seen this on two very congested networks in the US South East where people plugged into private MQTT's and were flooding packets out without care of the impact on overall congestion.

In the use cases where there is a good use case, those users are sophisticated enough to do custom firmware builds.

vidplace7:
> This has moved into a space where **users have zero control** and anyone who needs to actually allow hopped MQTT where an MQTT bridge is necessary is **FORCED** to build firmware.

We're saying the same thing. Framing it differently; This *enforces* "I know what I am doing, even if it's a bad idea". Taking easy control of this from the common user is a feature not a bug.

Xaositek:
Sensible defaults and user toggles are the best course of action. Allow users and administrators to perform these actions via CLI toggles **and** apps to decrease user friction

wehooper4:
> Sensible defaults and user toggles are the best course of action. Allow users and administrators to perform these actions via CLI toggles **and** apps to decrease user friction

The issue is the friction needs to be extremely high. Even a "I'm not a jerk" warning or whatever wont prevent people turning in on in congested meshes as they think they are doing a cool good thing.

Have it as a build-time only flag that one can set to turn it off is the only way that would prevent this traffic from jamming up routers in the European and metro US meshes.

Hoser01:
I agree with Xaositek’s assessment, and I want to add some operational context from our regional network.

In deployments like ours, node owners, especially those who operate routers, repeaters, or other backbone infrastructure, intentionally control whether their hardware will pass MQTT traffic. That control is part of how we design and manage the network. Removing or restricting that ability in stock firmware would force groups like ours to either redesign our networks or maintain custom firmware builds, and that is not sustainable long term.

Our mesh relies heavily on MQTT bridged/rf traffic to reach areas where we do not have internet connected nodes. Much of our regional traffic originates on MQTT and is delivered over RF through these infrastructure nodes. We have a main private channel, almost all traffic is MQTT sourced. We keep the primary channel RF only unless a remote user needs to send telemetry. This configuration has been stable and reliable for nearly 60 active users. Getting everyone to change device profiles or channel settings would create significant operational issues.

We also maintain a private RF only channel for sensitive or isolated operations, and this depends on the ability to choose, per node and per channel, what traffic should be bridged.

In summary, the ability to decide whether a node participates in MQTT passthrough is a fundamental decision for the network operator. Removing this flexibility would negatively affect established regional networks that rely on mixed RF and MQTT infrastructure. Preserving the current behavior, while improving safeguards or defaults if needed, keeps compatibility intact without forcing community groups to manage their own firmware builds.

Cheers!

wehooper4:
The issues is the meshes with 100's of active users (Europe, US cities).

Just blocking it on the routers isn't enough, as user nodes trying to send it still eat router RX airtime and create packet collisions even if they are blocked at the router.

People enable MQTT downlink because they think they are doing a good thing, even is it's very detrimental to the mesh. No amount of warning is going to prevent the RF impact from it.

Hoser01:
I understand the intent behind this PR, but right now it is doing the bluntest possible thing by replacing p->hop_limit = e.packet->hop_limit with p->hop_limit = 0, which forces every MQTT originated packet to be zero hop with no option for configuration. Zero hop as the default is absolutely reasonable and it directly addresses the very real problem of casual users unintentionally flooding public meshes with downlink traffic. That part makes sense.

The concern is that some networks intentionally rely on limited hop MQTT for linking segments, running backbone paths, or supporting automated services. Removing all configurability forces those deployments into custom firmware and creates fragmentation within the ecosystem. A more balanced solution would preserve the zero hop default while still providing a deliberate, high friction override for advanced administrators who need it.

One approach would be adding explicit configuration options such as network.mqtt_downlink_max_hops and network.mqtt_downlink_allow_override, both defaulting to zero and false. These options could be available only in the CLI or in an advanced section of the UI so they are not casually changed. Another protective layer would be allowing hopped MQTT only on specific channels that have downlink enabled while leaving all other channels strictly zero hop. This maintains safety for typical users while still offering an official, controlled path for experienced operators who have legitimate use cases for limited hop MQTT without requiring forked firmware.

BurdMahn:
Is there a way to make this a setting on the node side rather than the mqtt side that defaults to zero hop but can be changed? As in a separate setting than the downlink option. That way, operators who want to use mqtt as a crutch for not having real rf coverage can enable it on individual nodes, but the average user who just connects to a private mqtt server that doesnt zero hop packets without knowing all the details doesnt accidentally spam the airwaves?

Jckf:
Adding my vote: This behavior _must_ be configurable without requiring custom firmware. We rely on MQTT to RF propagation for parts of our mesh as well.

Nullvoid3771:
I’m highly opposed to this. Just use a broker that can inject zero hops packet like with meshmonitors built in broker.

b8b8:
https://github.com/meshtastic/firmware/issues/10397

[Feature Request]: Alternate default frequency when ham mode or MQTT is enabled

The point is people running custom MQTT with zerohop turned OFF, do not flood the LoRa network with traffic. Which is what is happening. Custom MQTT is destroying local LoRa meshes. Its reasonable that people with OKTOMQTT enabled are willing to carry this extra traffic and participate together.

What's happening now is the custom MQTTs are "stealing" the connectivity of the larger LoRa mesh without thought, and even if the pure LoRa users have IGNOREMQTT enabled, they are still subject to the constant flood of MQTT traffic (causing high channel utlization) of another nearby node with custom MQTT enabled.

This still allows people to run custom MQTT with all likeminded people running nodes. Nothing is taken away from them, they still can hop over their own LoRa. They just dont interfere with the pure LoRa nodes in the same region.

Nullvoid3771:
#10397 is a consideration, I appreciate that his solution is furthering opt in. However I don’t know if that’s the best solution. 

Here's what I think would be better long term. 

# Hybrid MQTT Infrastructure for Meshtastic

## Purpose

Provide a hybrid infrastructure based on MQTT that combines cellular-like infrastructure with ad hoc RF meshing. The design allows traditional Meshtastic hop-based networking while supporting MQTT backhaul between repeater sites. It should also provide the flexibility to use multiple HaLow or LoRa radios as fallback links or to bridge traffic between different presets, allowing traffic from channels such as LongFast to be carried over an RF backhaul such as ShortTurbo while remaining contained within its original encrypted channel.

## Reason

Traditional MQTT brokers create unnecessary congestion by distributing large amounts of MQTT traffic to radios that do not benefit from it. This increases radio subscriptions, inflates smaller nodeDBs, raises channel utilization, and allows traffic from distant locations to propagate into local meshes where it provides little value while consuming airtime.

## Summary

The goal is a cellular-like network of repeater stations that serve nearby ad hoc routers, route_late nodes, and client_base devices while maintaining synchronization through an MQTT backhaul.

These repeater stations would:

* Synchronize with one another over MQTT.
* Continue serving nearby mobile radios over RF.
* Provide additional fallback paths when normal RF hops are unavailable.
* Support alternate transport methods such as HaLow or additional LoRa radios.
* Scale through filtering and traffic management mechanisms.

Traffic carried over MQTT remains encrypted within Meshtastic channels. Links between MQTT brokers and repeater sites can also remain encrypted.

MQTT connectivity does not need to rely on the public Internet. Backhaul links could operate over HaLow, additional LoRa radios, Ethernet, Wi-Fi, VPNs, or any other available transport. Direct site-to-site TCP is optional rather than required.

The objective is improved stability, redundancy, and a different approach to using MQTT that minimizes its current negative effects.

---

# Detailed Explanation

Rather than relying solely on standard MQTT brokers, provide users with clear deployment tools similar to the Meshtastic boilerplate or MeshMonitor while expanding those tools to support optional zero-hop injection.

Standard Mosquitto deployments are not well suited to this purpose without additional functionality.

## NodeDB Management

The MQTT broker should process and filter node information before forwarding it to radios.

Doing so minimizes nodeDB growth on ESP32 and nRF52 devices and allows filtering before radios consume memory.

For example:

* Apply a configurable radius filter for nodes reporting position.
* Automatically ignore nodes beyond a configured distance such as 50 km.
* Allow favorited nodes to bypass filtering.
* Prevent ignored nodes from being rebroadcast.

MeshMonitor already implements much of this functionality.

## Education and Default MQTT Usage

Continue promoting zero-hop injection as the primary use case on Meshtastic's public MQTT broker since it effectively prevents flooding the RF mesh.

For deployments that intentionally use MQTT over RF to build hybrid infrastructure, MQTT access should generally be limited to repeater stations or local loopback services such as a Raspberry Pi. This follows the same philosophy as Router roles by placing infrastructure only in appropriate locations rather than throughout the mesh.

## Hybrid Repeater Stations

Hybrid repeater stations would use MQTT over RF to synchronize multiple repeater sites.

Their purpose is to:

* Keep repeater stations synchronized.
* Uplink and downlink traffic between sites.
* Provide a backbone for larger coverage areas.
* Reduce unnecessary RF hops.

Backhaul connectivity could use:

* Ethernet
* Wi-Fi
* VPN tunnels
* HaLow
* Additional LoRa radios
* Other available transports

Public Internet connectivity is optional rather than required.

If the backhaul fails, well-planned repeater stations should ideally continue communicating directly over RF whenever possible.

This provides redundancy using alternate frequencies or alternate radio technologies.

A local Raspberry Pi could also:

* Run MeshMonitor.
* Manage multiple radios.
* Provide more granular filtering.
* Support remote management.
* Optionally exchange keys locally.
* Copy NodeInfo and related information between local radio instances.

## Client Operation

Any radio served by these hybrid stations would continue to transmit and receive MQTT traffic over RF when okay_to_mqtt is enabled.

Users who do not want traffic originating from these repeater systems or regional MQTT deployments can continue using ignore_mqtt.

MeshMonitor radius filtering can automatically remove distant nodes that do not improve local communications.

Automation can also suspend synchronization if channel utilization exceeds configured thresholds to avoid contributing additional congestion.

Future features such as the Traffic Management module planned for Meshtastic 2.8 could further improve positional traffic handling within hybrid infrastructure.

## Regional MQTT vs. Hybrid Infrastructure

Regional MQTT deployments where many users subscribe to a common broker and indiscriminately downlink traffic present several problems:

* Large radio subscription counts.
* Excessive nodeDB growth.
* Topic leakage.
* Traffic unrelated to local areas.
* Increased channel utilization.

This proposal is not intended to expand that model.

Instead, privately managed MQTT backhaul between repeater stations provides a controlled method for increasing stability, reducing unnecessary RF hops, and supporting a cellular-like infrastructure while preserving normal ad hoc RF meshing as the primary communication method or fallback.

This proposal does not replace Meshtastic's existing routing, packet handling, or duplicate suppression. Existing Router, Router_Late, and Client roles continue to function normally, and Meshtastic's packet deduplication remains responsible for preventing duplicate retransmissions. The hybrid infrastructure extends the transport available to infrastructure nodes rather than introducing a separate routing protocol.

## Scaling

Repeater stations could be organized into clusters.

Each cluster would synchronize internally through MQTT while only overlapping neighboring clusters at RF boundaries.

This approach is most appropriate for:

* Dense urban areas.
* Bridging towns separated by terrain.
* Rural regions where direct RF coverage is limited by distance or geography.

## Improving MQTT Rather Than Rejecting It

Traditional MQTT brokers were not designed for this type of deployment.

Rather than abandoning MQTT entirely, expanding broker functionality and improving operator education would better support these hybrid deployments.

MQTT has the potential to:

* Connect geographically separated radios.
* Scale communication over long distances.
* Filter node information before it reaches radios.
* Provide flexible routing between infrastructure nodes.
* Support multiple transport technologies.

These capabilities require better tooling in brokers, clients, and firmware.

## RF Isolation

Pure RF isolation creates independent pockets of coverage that make roaming between cities and different meshes more difficult.

It also encourages infrastructure to be managed under a single operator's policies.

Meshtastic is fundamentally an unmanaged ad hoc mesh.

Individual operators should remain free to transmit useful packets even if doing so increases overall channel utilization. When congestion becomes excessive, operators can choose to remain on the frequency or migrate elsewhere.

Likewise, operators of managed infrastructure should be free to decide which traffic they rebroadcast.

Hybrid infrastructure provides another option rather than replacing traditional RF meshing.

If roaming extends beyond reasonable limits, neighboring repeater clusters can ignore one another's MQTT synchronization, allowing communication to naturally return to standard RF hopping, which remains limited to seven hops.

## MeshMonitor

MeshMonitor is referenced throughout this proposal because it currently implements several capabilities relevant to this design, including:

* Node filtering.
* Zero-hop injection.
* Traffic management.
* Automation.
* Infrastructure control.

These existing capabilities demonstrate functionality that could benefit hybrid infrastructure.

## Conclusion

Some operators prefer keeping Meshtastic entirely RF-based without TCP involvement.

This proposal is not intended to replace RF networking, but to provide an additional deterministic layer that improves stability and redundancy where appropriate.

Meshtastic currently depends on RF propagation and a maximum of seven hops for end-to-end delivery.

A hybrid MQTT infrastructure can provide another transport layer that increases packet survivability under varying conditions while reducing unnecessary RF hops and maintaining compatibility with traditional ad hoc meshing.

PR #8643, opened 2025-11-15
Populate syslog timestamp field when time is available

### LINKED ISSUE #7609: [Feature Request]: Set syslog timestamp field when possible
### Platform

Cross-Platform

### Description

At current remote syslog messages are sent in the RFC5424 format while setting the `TIMESTAMP` field to `NILVALUE` aka `-`: https://github.com/meshtastic/firmware/blob/master/src/DebugConfiguration.cpp#L179 (the last `-` there).
While this is standards compliant, turns out some syslog implementations assume the timestamp will always be populated. (For example, the Rust crate syslog_loose as used by Vector.)

Ideally all syslog parsers claiming RFC5324 compatibility would accept the syslog messages with nil timestamps and substitute time of receive, but since that's very much not true today, it would help if timestamps were included when available.

jhayden02:
Fixes #7609

Changed syslog messages from always using NILVALUE for the TIMESTAMP field. Now uses RFC5424 UTC timestamp format when a valid time is available from RTC. Falls back to NILVALUE if no valid time is available.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


jhayden02:
Why was this closed? Was there changes that needed to be made?

PR #8658, opened 2025-11-17
mqtt: do not try to send packets when it disconnected

omgbebebe:
While testing our network's performance during internet outages, we encountered some strange behavior. When the connection to the mqtt server is temporary unstable or completely disconnected, messages are either lost or delayed in their transmission over the air. These changes eliminate situations where attempts are made to send messages while the connection to the mqtt server is down.

To reproduce such behavior you can enable the mqtt module and enable uplink/downlink on the primary channel, then block mqtt traffic for the node on the internet gateway via iptables or something like that.

- [X ] I have tested that my proposed changes behave as described.
  - [ X] Heltec (Lora32) V3
  - [ X] Heltec (Lora32) V4

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -51,6 +51,7 @@ constexpr int reconnectMax = 5;
 static uint8_t bytes[meshtastic_MqttClientProxyMessage_size + 30]; // 12 for channel name and 16 for nodeid
 
 static bool isMqttServerAddressPrivate = false;
+static bool isConnected = false;
 
 inline void onReceiveProto(char *topic, byte *payload, size_t length)
 {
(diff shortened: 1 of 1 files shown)


src/mqtt/MQTT.cpp
```diff
@@ -688,7 +692,7 @@ void MQTT::publishNodeInfo()
 }
 void MQTT::publishQueuedMessages()
 {
-    if (mqttQueue.isEmpty())
+    if (mqttQueue.isEmpty() || !isConnected)
```
thebentern, line 695: Realizing after the merge, we probably want to not check the isConnected if mqtt.proxy_to_client_enabled is true

thebentern: Fixed in ef4cb2abfb6cc23dacb860f342e7a8c677387a25

Issue #8649, opened 2025-11-16
[Bug]: Do not scale telemetry interval less than configured


RCGV1:
### Category

Other

### Hardware

Not Applicable

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7

### Description

Nodes should not sen telemetry more often than configured even if they detect less nodes, sometimes a node could be in a bad spot where it sees not many nodes but it can get out well.

https://github.com/meshtastic/firmware/issues/8619

### Relevant log output

```Shell

```

Issue #8619, opened 2025-11-12
[Bug]: Configured Telemetry Interval is not correct


markusgritsch:
### Category

Other

### Hardware

Heltec Mesh Node T114

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.13

### Description

I have configured the Telemetry Power Metrics to be sent out every 30 minutes:

<img width="750" height="1334" alt="Image" src="https://github.com/user-attachments/assets/5e63835b-77a3-410e-91a9-c6683bba64d3" />

However, on a second node, the telemetry is received every 18 minutes:

<img width="750" height="1334" alt="Image" src="https://github.com/user-attachments/assets/abdc93a9-c9de-40c7-a68e-8745a19b34bb" />

I suppose the interval is stored as 30 * 60 = 1800 seconds, but then wrongly interpreted as 1800 / 100 = 18 minutes.

### Relevant log output

```Shell

```

korbinianbauer:
The interval scales with the number of online nodes.

I assume there's few online nodes in your list, it will scale down to 60% of the configured interval minimum.

0.6 * 30 = 18

Not very intuitive, but works as expected

b8b8:
Interval scales up and **down** based on number of "online" nodes.

<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/f8955384-f1f8-4046-afab-6b5f2ecd663d" />

markusgritsch:
Thank you for the quick and detailed reply!

I have to agree, it is very unintuitive and unexpected. If the node decides itself how long the interval should be, wouldn't it be better to have no configurable interval at all instead of scaling a specified intervall at will?

I mean if it is scaled to protect from too much traffic, it makes little sense to me to scale back long and short configured intervals by the same factor.

korbinianbauer:
Agreed. That's the kind of artefact you get when trying to stay backwards compatible across multiple platforms (firmware, APIs, Apps), unfortunately

markusgritsch:
Ok, so I close this issue since it is a feature :)

bplein:
I'd say it's not a feature, it's "by design". This behavior is not apparent from the client apps nor from the name of the settings as output in the exported backups. 

[...]

PR #8664, opened 2025-11-18
Don't scale up the frequency of telemetry sending

RCGV1:
Previously if the num online nodes was lower than a certain amount our node would send telemetry at a higher interval than configured, this leads to congestion in scenarios where a node has a bidirectional mesh connection where it doesn't see many nodes but transmits well. It is also not typical for a setting to be used more like a suggestion, the broadcast interval should be a minimum interval.

Related issues:
https://github.com/meshtastic/firmware/issues/8649
https://github.com/meshtastic/firmware/issues/8619


fifieldt:
This patch makes sense. It might be worth checking what the firmware implemented hard minimum limit is and making the minimum a bit more flexible in small networks? Just to make sure there isn't someone out there relying on this we're cutting off.

thebentern:
The original purpose of scaling down the coefficient was to not hamstring small off-grid and backpacking networks into the same coarse broadcast interval constraints as large public ones

RCGV1:
I think we should let the user have more decision over the interval especially now that you have to manually enable it.

RCGV1:
I don't think that for a hiking network you would need to know clients battery percent less than every 30 min

b8b8:
> I don't think that for a hiking network you would need to know clients battery percent less than every 30 min

The telemetry is still subject to the 25% channel utilization hard throttle imposed on client role. 
Not everyone's use case is the same as yours. Some people also want more environmental telemetry, not just device telemetry.

RichieFrame:
I believe this will also fix issue #7911

b8b8:
> I don't think that for a hiking network you would need to know clients battery percent less than every 30 min

How about starting the scaling at 20 nodes instead of 40, and reducing the minimum interval to 15min.

crhode:
HELTEC_MESH_NODE_T114

There are seven radios "online" in my mesh.

Firmware 2.7.17.9058cce Alpha

I am under the impression that this firmware contains a patch for this bug [8664](https://github.com/meshtastic/firmware/pull/8664).  I try to set telemetry interval to one hour:

```
meshtastic \
   --set telemetry.device_update_interval 3600 \
   --set telemetry.environment_update_interval 3600 \
   --set telemetry.power_measurement_enabled false \
   --set telemetry.environment_measurement_enabled false \
   --set telemetry.air_quality_enabled false
```

... but, using the **python** API, I see telemetry from me every minute on the serial port.  

Q1. Is this truly broadcast?  

Q2. Why so often?

Q3.  Isn't this patch supposed to allow a minimum interval larger than one minute?

Here's what I keep telling myself:

```
# Actual TEL interval is always at least 60s but otherwise scales
# according to this formula:
#
# Actual = Spec * (1 + ((online_nodes - 40) * 0.075))
#
# ... according to https://github.com/meshtastic/firmware/issues/8619
# in Nov 2025 for Meshtastic 2.7.  (I expect this has been fixed!)
#
# ... so Spec is meaningless until online_nodes exceeds about 26,
# which is weird, and even then Spec would have to be at least 40hrs
# for Actual to reach longer than 1min.  It's too strange.  As a
# practical matter, you can't set TEL interval on a lightly loaded
# network.
```

... but I think this is the way it "used to be."  

Q4.  How has this changed for 2.7.17.9058cce Alpha?

fifieldt:
@crhode - direct connections (serial, Bluetooth, http, UDP) different to Lora.

crhode:
> @crhode - direct connections (serial, Bluetooth, http, UDP) different to Lora.

I  think you're right!  Ne-ver Mi-nd.  I'm not seeing any telemetry from my "other" radio, so I must have changed the interval in it more or less successfully.  There's just not a lot of Meshtastic "folklore" accessible by newbies.

fifieldt:
For sure! If you get a moment, please don't hesitate to throw in a docs patch - we're pretty thin on contributors there and could use your experience.

PR #8666, opened 2025-11-18
mqtt: obey own ok2mqtt flag

omgbebebe:
When the MQTT module is enabled and the channel has an MQTT uplink enabled, all packets from this node are sent to the MQTT topic, regardless of whether the ok_to_mqtt flag is set.
This change is intended to correct this behavior and ensure that the ok_to_mqtt flag is respected for packets sent from the node.
This behavior is useful when you want to act as an MQTT gateway for other nodes but don't want to send your own packets to the internet.

- [X ] I have tested that my proposed changes behave as described.
  - [X] Heltec (Lora32) V4

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -746,10 +746,16 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp_encrypted, const meshtastic_Me
         // For uplinking other's packets, check if it's not OK to MQTT or if it's an older packet without the bitfield
         bool dontUplink = !mp_decoded.decoded.has_bitfield || !(mp_decoded.decoded.bitfield & BITFIELD_OK_TO_MQTT_MASK);
         // check for the lowest bit of the data bitfield set false, and the use of one of the default keys.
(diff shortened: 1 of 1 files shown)


thebentern:
I think there is a misunderstanding here. lora.ok_to_mqtt only controls the lora transport of mqtt originated (or traversed) packets. It should not influence uplink _to_ mqtt of packets

omgbebebe:
> I think there is a misunderstanding here. lora.ok_to_mqtt only controls the lora transport of mqtt originated (or traversed) packets. It should not influence uplink _to_ mqtt of packets

Okay. But how do I configure my stationary node to act as an MQTT gateway for any other nodes, but I don't want to send packets from this stationary node to MQTT at all?
This requirement was the main reason for this PR.

omgbebebe:
Okay. Let's close it for now. I don't think I fully understand the purpose of this option. In my case, without the patch, it literally does nothing. But I can imagine a case where a node sends its packets to some private MQTT server and wants to collect packets from other nodes on that MQTT server, but doesn't want its packets published to other MQTT servers. Maybe this makes sense in some cases. The documentation says that disabling ok_to_mqtt should prevent nodes from publishing any packets to MQTT (I believe this applies to any node, including the source node). I don't know.

Issue #8468, opened 2025-10-27
[Bug]: Heltec Mesh Node T114 lockups on 2.7.10


markusgritsch:
### Category

Other

### Hardware

Heltec Mesh Node T114

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.10.94d4bdf

### Description

I'm running a rooftop node using the Heltec T114. When using firmwares 2.7.10, 2.7.11, 2.7.12 (cannot say it started with 2.7.10, was present maybe earlier), the node locks up with the green LED lit on high traffic/lots of nodes. Everything runs fine when I use the same node in a remote location where only a few nodes are visible.

I went back to 2.6.11 and it is running fine for several days now also on my rooftop. This hanging of the T114 with a lit green LED seems to be a recurring issue. I found similar reports: https://www.reddit.com/r/meshtastic/comments/1jaqpdy/heltec_t114_v2_locking_up_after_a_few_days_of/

(Running 2.7.10 on a Heltec V3 shows now problems. Runs stable for days in busy locations.)

### Relevant log output

```Shell

```

Xaositek:
Can you provide some more details around your device and mesh? Some of these questions are in conflict but trying to isolate any unique information

- Which LoRa Frequency and Channel
- What is the role of your device
- How it is powered
- How many active nodes on the mesh
- How many total nodes in your NodeDB?
- Is it connected to MQTT via a phone/tablet?
- Is Bluetooth enabled/disabled?

korbinianbauer:
When you tested the behaviour in low vs. high traffic area, was your phone within BLE range in one of these spots and not the other?

markusgritsch:
>     * Which LoRa Frequency and Channel

EU868, so 869.525 MHz

>     * What is the role of your device

I took the box to a hilltop where I configured it as ROUTER. Had a lockup. Took it down to investigate. Mounted on the rooftop for testing as normal CLIENT, also lockup.

>     * How it is powered

5000 mAh LiIon battery and solar panel. Battery never lower than 90%.

![Image](https://github.com/user-attachments/assets/4a3d1c4d-e009-44a7-b6f0-55b5df475c6f)

>     * How many active nodes on the mesh

Hard to say, maybe a few dozen but definitely <100.

[...]

PR #8670, opened 2025-11-18
nrf52 watchdog (attempt #2)

SebKuzminsky:
This PR differs from the disastrous #8485 in two ways:
* WDT now only runs while the MCU runs; the WDT is paused when the MCU is in Sleep or Halt.
* WDT timeout is now 90 seconds instead of 2 seconds. This matches the ESP32 WDT timeout. This should fix the spurious watchdog bite during flashing.

I would appreciate a sanity test from @Xaositek and @HarukiToreda

### DIFF
--- a/src/platform/nrf52/main-nrf52.cpp
+++ b/src/platform/nrf52/main-nrf52.cpp
@@ -5,6 +5,7 @@
 #include <SPI.h>
 #include <Wire.h>
 
+#define APP_WATCHDOG_SECS 90
 #define NRFX_WDT_ENABLED 1
 #define NRFX_WDT0_ENABLED 1
 #define NRFX_WDT_CONFIG_NO_IRQ 1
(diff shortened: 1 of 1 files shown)


SebKuzminsky:
@markusgritsch can you see if this resolves #8468 for you?
(And by "resolve" i mean it reboots by itself after 90 seconds instead of hanging forever when it has a problem.)

HarukiToreda:
awesome, i'll set a couple to test out

Xaositek:
@thebentern Let's merge this so we can test it a bit more; I do like the rework here, great job @SebKuzminsky

HarukiToreda:
I've been running it with no issues at all

Xaositek:
Following the merge I've tested so far my Wio Tracker L1 Pro and Heltec T114; was able to change settings successfully.

SebKuzminsky:
Yay, thanks folks! :)

FFAMax:
On the NRF52840 Seeed XIAO, the system is functioning as expected. I am wondering if it is possible to log the reboot reason during boot, specifically if there is a mechanism in the hardware that preserves the trigger for the reboot.

HarukiToreda:
I'm getting T114 nodes that are off(deep sleep) and they will randomly reboot and power on by themselves. It's very random and sometimes it can take a full day or two to do this.

markusgritsch:
> I'm getting T114 nodes that are off(deep sleep) and they will randomly reboot and power on by themselves. It's very random and sometimes it can take a full day or two to do this.

I have two T114 running 2.7.16 powered by a battery, which never sporadically reboot. They also never turn on themselves from deep sleep. Do you have anything connected to the USB connector? My devices turn themselves on from deep sleep if I connect them to power on the USB port.

Xaositek:
@SebKuzminsky between @HarukiToreda's comment and now https://github.com/meshtastic/firmware/issues/8959, do we need to investigate?

SebKuzminsky:
I think it's unlikely that @HarukiToreda's comment indicates a problem with the new watchdog, given the large and variable time. But it's not impossible.

I like @FFAMax's suggestion to somehow log/record the reboot reason, but i don't know how to do that.

#8959 is reported against 2.7.15, which does not include the nrf watchdog (but does include some other nrf deep sleep changes), so i think this also does not suggest a wdt problem. It's possible the reported version number is wrong and it's really somewhere between 2.7.15 and 2.7.16, which could potentially include the nrf watchdog, i'll ask for more info in that issue.

PR #8674, opened 2025-11-18
Simplify CannedMessageModule::drawHeader

jasonbcox:
Deduplicates some of the implementation and saves ~112 bytes of flash memory. Behavior is unchanged.

It builds locally and existing tests pass but there is no coverage for this change. I don't have screen hardware necessary to test for regressions. Help here is appreciated.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/modules/CannedMessageModule.cpp
+++ b/src/modules/CannedMessageModule.cpp
@@ -191,20 +191,17 @@ int CannedMessageModule::splitConfiguredMessages()
 
     return this->messagesCount;
 }
+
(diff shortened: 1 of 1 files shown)


Xaositek:
This all gets gutted with https://github.com/meshtastic/firmware/pull/8182; I don't think we'd want to merge that with Multi Message around the corner.

jasonbcox:
Makes sense. I will revisit after that lands and reopen if there's anything left to gain.

PR #8681, opened 2025-11-19
Add a new frequency range for EU SRD area with adaptive DC

caveman99:
We might wanna consider switching telemetry and maybe even position off for this band to operate with the reduced duty cycle.

### DIFF
--- a/src/DisplayFormatters.cpp
+++ b/src/DisplayFormatters.cpp
@@ -34,6 +34,9 @@ const char *DisplayFormatters::getModemPresetDisplayName(meshtastic_Config_LoRaC
     case meshtastic_Config_LoRaConfig_ModemPreset_LONG_MODERATE:
         return useShortName ? "LongM" : "LongMod";
         break;
+    case meshtastic_Config_LoRaConfig_ModemPreset_LITE_FAST:
+        return useShortName ? "LiteF" : "LiteFast";
+        break;
     default:
         return useShortName ? "Custom" : "Invalid";
         break;
--- a/src/airtime.cpp
+++ b/src/airtime.cpp
@@ -133,11 +133,12 @@ bool AirTime::isTxAllowedChannelUtil(bool polite)
 
 bool AirTime::isTxAllowedAirUtil()
 {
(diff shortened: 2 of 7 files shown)


caveman99:
this requires a) https://github.com/meshtastic/protobufs/pull/814 and b) further discussion

phaseloop:
Is this even legal according to the linked ETSI document? 

```
The present document applies to RFID interrogators and tags operating together as a system. 
```

AFAIK Meshtastic node would not even pass "battery assisted tag" definition, not to mention interrogator role.


Edit: EU direction is specific too:
https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=CELEX:32019D1345

phaseloop:
Leaving this as separate comment for visibility:

of you look at section 47b in EU directive table we may be able to use new channels there in 868 band (pending bandwidth discussion):

- meshtastic is a data network (duh!)
- each node in mesh is an access point (non forwarding nodes will have lower duty cycle which does not hurt anyone)
- adaptive power control is something we can explore, we can lower transmit power a bit if neighbouring nodes have strong rssi for example.

caveman99:
The german regulatory body sees the definition of an access point decisivly as a fixed (immovable) device, and that is in line with the ETSI definition context. So routers may use 10% and clients only 2,5%. The PR addresses this by introducing a dynamic duty cycle tied to the device role.

phaseloop:
Makes sense - but I don't think this should be tied to device role as meshtastic routing and timings have nothing to do with node being fixed or not - most fixed nodes use CLIENT role. We already have a flag for "unattended" device and we should probably have one indicating that device is fixed.

This attribute could actually also help routing protocol anyway long term so for example fixed node will have priority as nexthop instead of a device that may not be in range in next 10 mins.

caveman99:
Superseded by #10120 (firmware) and meshtastic/protobufs#824, which landed equivalent functionality: the EU_866 region, role-adaptive duty cycle, and the Lite/Narrow presets are now in develop. Merging current develop into this branch yields an identical tree, so there is nothing left to contribute here. Closing in favor of those PRs.

PR #8705, opened 2025-11-21
Add Thinknode M6

caveman99:


lyusupov:
@caveman99 

image


<br>
<br>

Is **RTC INT** input pin missing ?

caveman99:
@lyusupov Meshtastic does not use that pin, although it gets defined in a few of the variants. RTC_INT is connected to P0.04 in this design.

PR #8716, opened 2025-11-22
R1 Neo - Added OCV_ARRAY from measured discharge curve testing + update ADC multiplier

simon-muzi:
The ADC resistor divider ratio is 0.6 -> multiplier should be 1/0.6 ~=1.667

We data logged a full discharge curve at constant 100mA draw over 15hours to get a realistic voltage curve for battery SoC measurements.


simon-muzi:
```
#elif defined(R1_NEO)
#define OCV_ARRAY 4330, 4292, 4254, 4216, 4178, 4140, 4102, 4064, 4026, 3988, 3950
```
Will need to remove the old overwrite in `power.h`

jp-bennett (approved):

Issue #8711, opened 2025-11-22
[Feature Request]: Add to telemetry/Device health: Battery state, Charging state, Line power


FFAMax:
### Platform

Cross-Platform

### Description

Hello, Team!

Now 101% used to show device powered up from AC/USB/Line power.
Seems like need a little bit change protocol and App UI to match reality.
Need add a few more statuses:
Battery state: 
* Connected
* Disconnected
* Unknown

Battery charging state:
* Idle
* Charging
* Discharging
* Unknown

Line power:
* Connected
* Disconnected
* Unknown

Since I start working on https://github.com/meshtastic/firmware/issues/8614 it has become clear that we need more options to accurately report device metrics.

Thanks!

Issue #7522, opened 2025-07-31
[Bug]: SHT40 and SHT41 Temp Sensor is not usable, Output is shown as "NaN"


Stephan-2025:
### Category

Hardware Compatibility

### Hardware

Heltec Mesh Node T114

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.3

### Description

The current Meshtastic version 2.7.3 recognizes both the SHT40 and SHT41 sensors, but the values aren't read/displayed correctly. They're only displayed as "NaN" (see screenshot). I ordered and tested both types, as they're also on the compatibility list on the Meshtastic website. Unfortunately, it doesn't seem to work anymore (with the latest firmware). I don't know if it used to work before.

![Image](https://github.com/user-attachments/assets/24d112bd-a6e0-4999-935d-0815a69439b0)

### Relevant log output

INFO | ??:??:?? 6 Scan for i2c devices

DEBUG | ??:??:?? 6 Scan for I2C devices on port 2

INFO | ??:??:?? 6 SHT31 found at address 0x44

DEBUG | ??:??:?? 6 Scan for I2C devices on port 1

INFO | ??:??:?? 6 1 I2C devices found

Stephan-2025:
The SHT4x is wrong recognized as SHT31

So that could be the issue that no correct data is readable / display´able

andreas-xa86:
The firmware (file ScanI2CTwoWire.cpp) up to now only recognizes sensors with registerValue  0x11a2, 0x11da, 0xe9c and 0xc8d as a SHT4x sensor. Other values will be regarded as SHT31 sensor.


I added a line to print the register value to the debug log.
My sensor shows 0x1320. (Use the serial console on flasher.meshtastic.org)
I added this value to the IF statement.
The next time the sensor was recognized correctly and temperature an humidity are displayed.


How to build your own firmware: https://meshtastic.org/docs/development/firmware/build/
Sensor used: https://de.aliexpress.com/item/1005008021051062.html

ScanI2CTwoWire.cpp after line 375

`                LOG_DEBUG("Register Value: 0x%x", registerValue);
                if (registerValue == 0x11a2 || registerValue == 0x11da || registerValue == 0xe9c || registerValue == 0x1320 || registerValue == 0xc8d) {
[...]

PR #8723, opened 2025-11-23
XIAO NRF52: Support for DS18B20 Temp Sensor + Improved battery readings

FFAMax:
Hello, Team.

New: XIAO NRF52: Added support for DS18B20 temperature sensor (1 sensor for now).
New: XIAO NRF52: Added battery status reporting (connected/not connected).
New: XIAO NRF52: Initial support for battery charging status (based on voltage levels to detect if the battery is charging or discharging).
Affect: All platforms: Devices with 101% battery will no longer show as "Line Powered." This will be migrated to a separate metric once the feature is fully implemented. https://github.com/meshtastic/firmware/issues/8711

Tested on XIAO NRF52.

### DIFF
--- a/src/modules/Telemetry/Sensor/DS18B20Sensor.cpp
+++ b/src/modules/Telemetry/Sensor/DS18B20Sensor.cpp
@@ -1,0 +1,46 @@
+#include "configuration.h"
+
+#if HAS_TELEMETRY && !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR 
+
+#include "DS18B20Sensor.h"
+#include "TelemetrySensor.h"
+#include <typeinfo>
+#include <DallasTemperature.h>
+#include "../mesh/generated/meshtastic/telemetry.pb.h"
+
+DS18B20Sensor::DS18B20Sensor() : TelemetrySensor(meshtastic_TelemetrySensorType_DS18B20, "DS18B20") {
+    pin = PIN_WIRE_DS18B20; // Define this pin in your board's variant.h
+    oneWire = new OneWire(pin);
+    sensors = new DallasTemperature(oneWire);
+    sensors->begin();
+    status = sensors->getDeviceCount() > 0;
+}
+
+int32_t DS18B20Sensor::runOnce()
+{
+    if (!hasSensor()) {
+        return DEFAULT_SENSOR_MINIMUM_WAIT_TIME_BETWEEN_READS;
+    }
+    sensors->requestTemperatures();
+    status = sensors->getDeviceCount() > 0;
+    LOG_INFO("DS18B20: pin=%d, device_count=%d", pin, sensors->getDeviceCount());
+    return initI2CSensor();
+}
+
+void DS18B20Sensor::setup() {}
+
+bool DS18B20Sensor::getMetrics(meshtastic_Telemetry *measurement)
+{
+    if (!hasSensor()) return false;
+
+    sensors->requestTemperatures();
+    float tempC = sensors->getTempCByIndex(0);
+
+    if(tempC != DEVICE_DISCONNECTED_C) {
+        measurement->variant.environment_metrics.has_temperature = true;
+        measurement->variant.environment_metrics.temperature = tempC;
+        return true;
+    }
+    return false;
+}
+#endif
--- a/src/modules/Telemetry/Sensor/DS18B20Sensor.h
+++ b/src/modules/Telemetry/Sensor/DS18B20Sensor.h
@@ -1,0 +1,22 @@
+#include "configuration.h"
+
+#if !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR
+
+#include "TelemetrySensor.h"
+#include <DallasTemperature.h>
+#include <OneWire.h>
+
+class DS18B20Sensor : public TelemetrySensor {
+  private:
+    OneWire *oneWire;
+    DallasTemperature *sensors;
+    uint8_t pin;
+
+  public:
+    DS18B20Sensor();
+    virtual int32_t runOnce() override;
+    virtual void setup() override;
+    virtual bool getMetrics(meshtastic_Telemetry *measurement) override;
+};
+
+#endif


NomDeTom:
> New: XIAO NRF52: Added support for DS18B20 temperature sensor (1 sensor for now).
New: XIAO NRF52: Added battery status reporting (connected/not connected).
New: XIAO NRF52: Initial support for battery charging status (based on voltage levels to detect if the battery is charging or discharging).
Affect: All platforms: Devices with 101% battery will no longer show as "Line Powered." This will be migrated to a separate metric once the feature is fully implemented. https://github.com/meshtastic/firmware/issues/8711

I would suggest splitting this into separate pull requests.

With regard to DS18B20 sensors (which I know relatively well, and use a number outside of the Meshtastic project) I believe 1-wire sensors were removed at one point to save on pins and to reduce flash consumption - the functionality is replicated by I2C sensors and they're a similar order of cost.

Do you have any idea of how this will affect flash consumption?

garthvh:
Dallas sensors were removed because they can't be auto connected and configured, not really sure what moving away from 101 as the powered value does, would make graphs worse and require rework in both clients

FFAMax:
>I would suggest splitting this into separate pull requests.

Agree. Will do. 

>With regard to DS18B20 sensors (which I know relatively well, and use a number outside of the Meshtastic project) I believe 1-wire sensors were removed at one point to save on pins and to reduce flash consumption - the functionality is replicated by I2C sensors and they're a similar order of cost.

1. This request came from the community so at least somebody happy by having it again :) 
2. On the Internet I found people also asking for it, even building firmware
https://github.com/jeremyd/meshtastic-firmware/releases/tag/v2.5.16-ds18b20 (https://www.reddit.com/r/meshtastic/comments/1ixhhuv/ds18b20_renegade/)
https://github.com/jle1n0/meshtastic-firmware/commit/1afbe14b458e567eb84be80fdacb5b5a7f0bfa54
So anyway better to keep it onboarded.

 >Do you have any idea of how this will affect flash consumption?

Can leave it only on specific platforms like XIAO NRF52 if any concerns. So compiler will skip if not enabled.

>Dallas sensors were removed because they can't be auto connected and configured

@garthvh could you please elaborate, what do you mean "they can't be auto connected and configured". Once sensor connected to proper pin and telemetry enabled, it will work. This code particularly can support even many on the same pin, like environment, components, etc.

>not really sure what moving away from 101 as the powered value does, would make graphs worse and require rework in both clients

The 101% reading was initially a workaround, but now I've encountered this historically developed issue. The battery is one entity, and the power line is another. It can't be called 'warm' because it's more like 'milk'.
Let's work on https://github.com/meshtastic/firmware/issues/8711 to separate these entities properly so we can report accurate information.
The last time I checked, the device was connected to the power line, and it showed 100% charge. It was difficult to determine whether the battery was fully charged or not, because the graph still showed 100%, even though the device was plugged into the power line."

garthvh:
Not sure where you got the workaround info, but it is not correct.

thebentern:
Support for this i2c conversion for the Dallas 1-wire sensors is on my list. As @garthvh said, we only support i2c sensors unless something is baked into the hardware. It gives us the most options without complicating the configuration with pin definitions
https://www.adafruit.com/product/5976

FFAMax:
I just double-checked the XIAO NRF52840: the SDA/SCL pins are used for NSS and RF_SW.
An I²C conversion might work if there’s a built-in option; otherwise, it would require additional components.
For now, the only zero-cost option I see on the XIAO NRF52840 is to use a spare pin that’s 1-Wire compatible.
If I overlooked any details, please feel free to point them out.

>It gives us the most options without complicating the configuration with pin definitions

@thebentern I agree. We need to make sure it doesn’t interfere with the main project while still being useful on specific platforms, and it should be easy to enable without requiring new code. This isn’t only about the code itself - it also involves testing and everything around it.
It’s not easy to onboard a well-tested feature, but very easy to off-board one.
This approach satisfies the needs of a portion of the community who want specific features, and at the same time shows that their effort has value and their contributions aren’t ignored or dumped. That helps keep people motivated to do good work, test properly, maintain compatibility, and avoid ending up with “it works only in my private branch” solutions.

pinout2
Screen Shot 2025-11-23 at 6 45 52 AM

NomDeTom:
@FFAMax 
> If I overlooked any details, please feel free to point them out.

The NRF is capable of moving most functions around. If you look at the list of build environments, you'll see there's an i2c variant of the seeeeeeeeeeed NRF52 kit, using the back 2 pins for i2c. It's my preferred arrangement, actually.

I've been discussing this with TheBen on the discord (where the conversation is quicker!). While I'm a fan of the ds18b20 (not sure if I've mentioned it, but I'm a big fan!), given the current limitations and the comparable cost between the ds18b20 and e.g. AHT20 or SHT40 which provide the same function, I can see that incorporating the extra library, even behind a build flag, is adding a future support issue.

As for building special variants, I'm working on a solution to that, as well, because I like to move pins around to suit my own needs, and I don't want to have to mess around any more than I have to.

You're more than welcome to pop over to the discord to discuss what is on the to-do list for the project right now.

sandos:
One huge advantage of onewire is the amount of sensors you can connect. No address clashes.

NomDeTom:
> One huge advantage of onewire is the amount of sensors you can connect. No address clashes.

I think it's outweighed by the fact that there's not much more than temperature sensors that are suitable for Meshtastic use, and current telemetry protobufs don't support more than one sensor of that type.

I agree *in general* that they're a good sensor type to support, but here it's a case of "not now".

inraphil:
Hi, Just one comment about the fact that "there is supported sensors which provide the same function" ... Not really, I'm using waterproof DS18B20 to mesure water temperature up to 1m depth. I can not find any sensor for this purpose in the list of supported sensors.

NomDeTom:
> Hi, Just one comment about the fact that "there is supported sensors which provide the same function" ... Not really, I'm using waterproof DS18B20 to mesure water temperature up to 1m depth. I can not find any sensor for this purpose in the list of supported sensors.

A brief google reveals that the FS304-SHT2X utilises a SHT20 i2c sensor, and claims to be completely sealed and enclosed against water ingress. I'm fairly certain this is on the supported list.

FFAMax:
Be careful with SHT when buying, example https://github.com/meshtastic/firmware/issues/7522

FFAMax:
But they are still not comparable with DS18B20 (on pic), 5pcs for $8 - I didn't saw that low price for SHTxx
photo_2026-01-12 18 56 20

inraphil:
The FS304-SHT2X seems to be an interesting alternative but very too expensive (minimum ten times) in my case (I want to install dozens of sensors) and not sure it supports immersion (short circuit on hygrometry measure). After browsing the net, gemini and chatgpt, i guess there is no equivalent to DS18B20 with I2C, low cost and really waterproof to imersion. I guess meshtastic will not be an option. Using serial interface with additional module will cost more energy and complexity.

FFAMax:
@inraphil, since the topic started, it is still usable with your own code, but the only thing is that it becomes more and more outdated.
More stuff also needs to be coded, like smarter power management to let the device run if there is no sun for a week or so.

I started with DS18B20 because in that area from where I got the request, the community just has these sensors lying around, even without any need to purchase them. So why do we need to spend money if they are already in the hands of some friend — literally sitting in a box and waiting? So this not only saves money, but it is also smart to use what you already have instead of purchasing more and more.

I am focusing on building cheap versions of the node, so I 100% understand your case. It is reasonable, especially for your case of dozens. Feel free to build firmware based on my commit or take the code and integrate it with a recent version. If you don’t need new features, it will work.

inraphil:
@FFAMax I completely agree with your point of view. It seems to me that Meshtastic was started to offer a robust and resilient communication solution. In this pursuit of resilience, being able to use popular and readily available components without having to purchase them should be a priority. Furthermore, I guess 1-Wire support was available in older versions but has been removed. Using your modified firmware is an option, and I think your contribution should be integrated into Meshtastic's features. However, since I'm considering using it for sensors in scientific studies, I also need to be able to ensure I understand the code and adapt it to future firmware updates. I need to find the time to determine if this is within my capabilities...

Issue #8565, opened 2025-11-06
[Bug]: t-lora pager core panic


mesb1:
### Category

Other

### Hardware

Other

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [x] OLED slide UI on any display

### Firmware Version

2.7.13

### Description

All my 2.7 for t-lora pager experienced it.
Mostly while some work with trace for example, it just reboots.
Also i see no watch had got time from lora while t-deck doing it easily.

[13:07:52]DEBUG | ??:??Guru Meditation Error: Core  1 panic'ed (Interrupt wdt timeout on CPU1).

### Relevant log output

```Shell
[13:07:22]INFO  | ??:??:?? 671 [Router] Route traced:
[13:07:22]0x9eecb2c4 --> 0xb2a72678 (0.00dB) --> 0x9ea13f34 (-19.25dB) --> 0x6982b394 (0.00dB) --> 0xdb5177e8 (-17.25dB) --> 0x433bf034 (3.50dB) --> 0xb3b
[13:07:22]DEBUG | ??:??:?? 671 [Router] Module 'traceroute' considered
[13:07:22]DEBUG | ??:??:?? 671 [Router] Module 'routing' wantsPacket=1
[13:07:22]INFO  | ??:??:?? 671 [Router] Received routing from=0xb3b2420b, id=0x231f5ad, portnum=70, payloadlen=127
[13:07:22]DEBUG | ??:??:?? 671 [Router] Routing sniffing (id=0x0231f5ad fr=0xb3b2420b to=0x9eecb2c4, transport = 1, WantAck=1, HopLim=0 Ch=0x0 Portnum=70 requestId=1caafd05 rxSNR=5.5 rxRSSI=-87 hopS
[13:07:22]DEBUG | ??:??:?? 671 [Router] cancelSending id=0x1caafd05, removed=0
[13:07:22]DEBUG | ??:??:?? 671 [Router] Module 'routing' considered
[13:07:27]DEBUG | ??:??:?? 676 Rotary event CW
[13:07:27]DEBUG | ??:??:?? 676 State: ON
[13:07:27]INFO  | ??:??:?? 676 Input event 28! kb 0
[13:07:27]INFO  | ??:??:?? 676 [Screen] Turn on screen
[13:07:28]DEBUG | ??:??:?? 676 Rotary event CW
[...]

PR #8746, opened 2025-11-25
Prevent double-registering of Rotary Encoder on TLora Pager

thebentern:
Prior to this change, we experienced multiple logging events per operation
```
INFO  | ??:??:?? 591 Input event 10! kb 0
INFO  | ??:??:?? 591 Input event 10! kb 0
DEBUG | ??:??:?? 592 Rotary event Press
DEBUG | ??:??:?? 592 Rotary event Press
INFO  | ??:??:?? 592 Input event 10! kb 0
INFO  | ??:??:?? 592 Input event 10! kb 0
```

Also removed some chatty logs from ext. notification


mverch67:
Maybe this fixes https://github.com/meshtastic/firmware/issues/8565

PR #8754, opened 2025-11-25
Preliminary Thinknode M4 Support

caveman99:


lyusupov:
@caveman99 

Both M3 and M4 products use LR1110 radio and likely the same RF switch hardware design.

image

<br>


**Which one is incorrect ?**

caveman99:
The M3 uses https://www.elecrow.com/download/product/CIL13510X/nRFLR1110_Datasheet_V1.0.pdf while the M4 has a bespoke arrangement of nrf52 and LR1110 on the main PCB. However you are right, the M3 definition of HIGH-HIGH for the LP output is also true for the M4, i think i made a copy/paste error.

mverch67:
The blue LED stops blinking after ~30s, so there's no indication if the device is on or off.

PR #8766, opened 2025-11-26
Detection sensor sleep & interrupt

Columbo818:
Presented by @rbomze and myself, this PR updates the DetectionSensor module to use an interrupt-driven approach when the following conditions are met:

• The device is nRF52840-based
• The node role is Sensor
• Power saving mode is enabled

When these criteria are satisfied, the module uses GPIO interrupts instead of polling, allowing the device to reliably enter deep sleep for ultra-low power operation.

These changes are contained within an `#ifdef ARCH_NRF52` to prevent build failures on other platforms, ensuring that they only affect nRF52 devices. Devices that do not meet all of the above conditions, including nRF52-based devices, continue using the existing polling behaviour to maintain compatibility.

All existing configuration values (`sds_secs` and `min_wake_secs`) continue to be respected, preserving expected module behaviour.

## 🤝 Attestations

- [✅] I have tested that my proposed changes behave as described.
- [✅] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [✅] Heltec WiFi LoRa 32 (V2)
  - [✅] Heltec T114
  - [✅] Seeed Wio Tracker L1

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -4,31 +4,31 @@ cli:
 plugins:
   sources:
     - id: trunk
-      ref: v1.7.3
+      ref: v1.7.4
       uri: https://github.com/trunk-io/plugins
 lint:
   enabled:
-    - checkov@3.2.492
-    - renovate@42.5.4
+    - checkov@3.2.495
+    - renovate@42.24.1
     - prettier@3.6.2
-    - trufflehog@3.90.13
+    - trufflehog@3.91.1
     - yamllint@1.37.1
-    - bandit@1.8.6
+    - bandit@1.9.2
     - trivy@0.67.2
     - taplo@0.10.0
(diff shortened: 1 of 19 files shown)


Columbo818:
Some additional testing has revealed an issue where the nRF52 does not wake up.
Given this, we've had to revert to sleeping forever.

My feeling is to close the PR as sleeping forever ignores the SDS_Secs and will confuse inexperienced users, making them think their nodes are dead.
@rbomze - I think this works for your use case, but isnt ready to be made default until we can sort the wake issue.

rbomze:
Yeah, trash the pull request, Colombo. I am about to finish what works in both cases.
- sensing when sleeping indefinitely during a shutdown
- low power polling during a sleep with timeout

rbomze:
replaced by #8778, please drop PR.

PR #8768, opened 2025-11-27
Update Heltec_T190 and E290

hereway:
Hello everyone, I have updated the GPS and rotary encoder configuration for the Heltec-T190, and also added GPS adaptation support for the Heltec-E290.

Update Summary:
- Optimized GPS module configuration parameters for Heltec-T190
- Improved the rotary encoder driver for Heltec-T190
- Added complete GPS functionality support for Heltec-E290
- Ensured GPS data parsing stability for both devices

Testing Status:
- Successfully tested GPS positioning functionality on actual Heltec-T190 hardware
- Verified operation response of the rotary encoder on T190
- Completed basic GPS module functionality testing for E290

- [√] I have tested that my proposed changes behave as described.
- [√] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [√] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


src/mesh/NodeDB.cpp
```diff
@@ -849,6 +849,12 @@ void NodeDB::installDefaultModuleConfig()
     moduleConfig.canned_message.inputbroker_event_cw = meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar(28);
```
mverch67, line 874: I'm not sure if it's a good idea to hard-code these pin values into the firmware. These input pins are configurable for a reason. If they are hard-coded they cannot be used for sth different (e.g. detection sensor).
Instead these values should be configured via app or CLI.

variants/esp32s3/heltec_vision_master_e290/variant.h
```diff
@@ -3,6 +3,24 @@
 #define PIN_BUTTON2 21             // Second built-in button
 #define ALT_BUTTON_PIN PIN_BUTTON2 // Send the up event
 
+/*
+ * GPS pins
+ */
+
+#define GPS_L76K
```
mverch67, line 10: Heltec vision master doesn't have a built-in GPS so the default should always be not present.
`#define GPS_DEFAULT_NOT_PRESENT 1`

variants/esp32s3/heltec_vision_master_t190/variant.h
```diff
@@ -1,7 +1,31 @@
...
+#define INPUTDRIVER_ENCODER_TYPE 2
+
+#define inputbroker_CW_pin          16 //turn on right
+#define inputbroker_CCW_pin         15 //turn on left
+#define inputbroker_press_pin       21
+
+/*
+ * GPS pins
+ */
+#define GPS_L76K
```
mverch67, line 16: Heltec vision master doesn't have a built-in GPS so the default should always be not present.
`#define GPS_DEFAULT_NOT_PRESENT 1`

src/mesh/NodeDB.cpp
hereway: Thank you for the reminder. Regarding the hardware of the T190, we have already upgraded it to V2.0, adding a ROTARY and a GPS interface. Therefore, hard-coding the DIP switch settings into the firmware is likely the most convenient modification.

variants/esp32s3/heltec_vision_master_e290/variant.h
hereway: Thank you once again for the reminder. We have upgraded the hardware of the E290 to version V1.0. However, it has not been released yet as we are still in the testing phase. My sincere apologies for not mentioning the hardware changes in the previous description. In the E290 V1.0, we have added a hardware GPS interface.

PR #8771, opened 2025-11-27
Add initial support for Hackaday Communicator

jp-bennett:
To support the screen, I've pulled in Arduino_GFX, a library similar to LovyanGFX, and added support for it in the same way.

The only other notable change is that in src/platform/esp32/architecture.h, we now fall through to private HW ID by default, if none is selected.

### DIFF
--- a/src/input/HackadayCommunicatorKeyboard.cpp
+++ b/src/input/HackadayCommunicatorKeyboard.cpp
@@ -194,8 +194,8 @@ void HackadayCommunicatorKeyboard::released()
 
     uint32_t now = millis();
     last_tap = now;
-
-    queueEvent(HackadayCommunicatorTapMap[last_key][modifierFlag % HackadayCommunicatorTapMod[last_key]]);
+    if (HackadayCommunicatorTapMod[last_key])
+        queueEvent(HackadayCommunicatorTapMap[last_key][modifierFlag % HackadayCommunicatorTapMod[last_key]]);
     if (isModifierKey(last_key) == false)
         modifierFlag = 0;
 }
--- a/variants/esp32s3/hackaday-communicator/variant.h
+++ b/variants/esp32s3/hackaday-communicator/variant.h
@@ -1,6 +1,3 @@
-
-#define TFT_CS 41
-
 #define TFT_BL 2
 #define SPI_FREQUENCY 2000000
 #define SPI_READ_FREQUENCY 16000000


src/graphics/draw/DebugRenderer.cpp
```diff
@@ -97,8 +97,7 @@ void drawFrame(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t x, int16
                                           (storeForwardModule->heartbeatInterval * 1200))) { // no heartbeat, overlap a bit
 #if (defined(USE_EINK) || defined(ILI9341_DRIVER) || defined(ILI9342_DRIVER) || defined(ST7701_CS) || defined(ST7735_CS) ||      \
      defined(ST7789_CS) || defined(USE_ST7789) || defined(ILI9488_CS) || defined(HX8357_CS) || defined(ST7796_CS) ||             \
-     defined(USE_ST7796) ||            \
-     ARCH_PORTDUINO) &&                                                                                                                                                                                                    \
+     defined(HACKADAY_COMMUNICATOR) || defined(USE_ST7796) || ARCH_PORTDUINO) &&                                                 \
```
mverch67, line 100: is here a "defined" missing -> defined(ARCH_PORTDUINO)

src/input/HackadayCommunicatorKeyboard.cpp
```diff
@@ -0,0 +1,271 @@
...
+    ledcWrite(LEDC_BACKLIGHT_CHANNEL, _brightness);
+#endif
+}
+
+void HackadayCommunicatorKeyboard::pressed(uint8_t key)
+{
+    if (state == Init || state == Busy) {
+        return;
+    }
+    if (config.device.buzzer_mode == meshtastic_Config_DeviceConfig_BuzzerMode_ALL_ENABLED ||
```
mverch67: if Hackaday communicator has no vibration motor then these lines can be removed?

src/graphics/draw/DebugRenderer.cpp
jp-bennett: No, ARCH_PORTDUINO only gets defined to "1". I'd more prefer to drop "defined" from more of these macros, and just always set them to 1 for enabled.

src/input/HackadayCommunicatorKeyboard.cpp
jp-bennett: Good catch. Pushing that update.

src/graphics/TFTDisplay.cpp
```diff
@@ -1271,12 +1276,15 @@ void TFTDisplay::display(bool fromBlank)
                     x_LastPixelUpdate = x;
                 }
             }
-
+#if defined(HACKADAY_COMMUNICATOR)
+            tft->draw16bitBeRGBBitmap(x_FirstPixelUpdate, y, &linePixelBuffer[x_FirstPixelUpdate],
+                                      (x_LastPixelUpdate - x_FirstPixelUpdate + 1), 1);
```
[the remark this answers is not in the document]
jp-bennett: It was backwards before, using the Big Endian version fixes the colors.

src/input/HackadayCommunicatorKeyboard.cpp
```diff
@@ -0,0 +1,217 @@
...
+    if (last_key < 0 || last_key >= _TCA8418_NUM_KEYS) {
+        last_key = -1;
+        state = Idle;
+        return;
+    }
+
+    uint32_t now = millis();
+    last_tap = now;
+
+    queueEvent(HackadayCommunicatorTapMap[last_key][modifierFlag % HackadayCommunicatorTapMod[last_key]]);
```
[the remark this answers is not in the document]
jp-bennett: Copilot earned its keep on this one. Very good spot.

PR #8779, opened 2025-11-27
Remove screen activation in powerExit function

jp-bennett:
This seems to be a potential source of unintended screen wakes.
Only tested with the Hackaday Communicator so far.

### DIFF
--- a/src/PowerFSM.cpp
+++ b/src/PowerFSM.cpp
@@ -219,8 +219,6 @@ static void powerIdle()
 
 static void powerExit()
 {
-    if (screen)
-        screen->setOn(true);
     setBluetoothEnable(true);
 }
 


thebentern:
This one could have some side effects. Definitely want to test with a number of other hardware variants

jp-bennett:
I'm actually wondering if this should be moved out of PowerFSM completely, and handled in Screen::handleInputEvent()

jp-bennett:
> This one could have some side effects. Definitely want to test with a number of other hardware variants

This is why I targeted Develop. ;)

thebentern:
> I'm actually wondering if this should be moved out of PowerFSM completely, and handled in Screen::handleInputEvent()

That makes a lot of sense, similar to how we stop ext. notification nagging there.

jp-bennett:
Tested with one of the few devices that has both a screen and EXT_PWR_DETECT pin, and it worked as expected. Also tried a RAK4631 with screen, and it also seems to be working as expected. I feel pretty good about this change, but let me know if you find where it breaks something.

Issue #8763, opened 2025-11-26
[Feature Request]: Power-fault Flooding Protection (network congestion)


dem1tris:
### Platform

Any

### Description

After a power reboot, nodes immediately send telemetry, position, and nodeinfo packets. In case of power-faulty equipment, nodes can severely overload the network, and such issues are difficult to detect. The firmware must include flood protection in case of frequent reboots.

## Solution (from must-have to nice-to-have):

1. **Persistent uptime calculation** using two monotonic relative timestamps. The node maintains a persistent monotonic timestamp while it's alive. On reboot, the node backs up the latest persisted value and continues incrementing the current timestamp. On the next reboot, the node calculates the latest uptime using the "backed up" timestamp and the latest persisted "current" timestamp. If the latest uptime is less than the "radio cooldown" threshold, the node keeps silence on core portnums.
2. **Power-fault reports.** Every too-short uptime is counted. For several such events within some interval, the node reports itself as power-faulty to the client app, or even to the mesh.
3. **Power-faulty neighbours protection.** Mesh nodes receiving a power-fault report reduce NODEINFO packet retransmission and mark the faulty node in the app. Or just limit NODEINFO retransmission by the minimal interval (3 hours) for every node heard

dem1tris:
Here's the top part of NODEINFO histogram for the packet dump I collected. The first node reboots 10 times/hour. The two other nodes translated theirs NODEINFO also more often than the minimum defaults (once per 3 hours).

<img width="1496" height="226" alt="Image" src="https://github.com/user-attachments/assets/9ce84b29-e366-4785-aa76-ce813a17c85d" />

dem1tris:
Here's also the chart with more data, if anyone interested. 10000 packets collected, ~4000 filtered out as self-owned and never sent to mesh. 5500 of the remaining 6000 are the "service" packets.

<img width="1996" height="664" alt="Image" src="https://github.com/user-attachments/assets/97ed24d2-4727-4dd0-b7fb-4ff9af79916b" />

NomDeTom:
Would this not be better accomplished simply by making the routine broadcasts wait the full interval period following reboot before sending out a packet? Bootlooping nodes will then send fewer packets (because each reboot will reset the timer), but will still be able to participate in the mesh.

If you have an idea for how to implement your suggestions in the code, I would suggest making and testing the changes, then submitting a pull request.

dem1tris:
> wait the full interval following reboot

[...]

PR #8793, opened 2025-11-28
NRF52 - low voltage management, fixing data corruption

phaseloop:
Draft - PR during testing and development.

Strayed a bit from original PR but in a good direction.

This PR solves various issues with NRF52 boards when voltage is below 3V - fixing some data corruption scenarios and making them less prone to boot loops and problems like power fail radio flood: https://github.com/meshtastic/firmware/issues/8763

While ESP32 boards die at around 3.0V, most NRF52 + LDO combinations easily allow to go down to 2.5V. Usually firmware will try to put device to sleep at around 3.1V but this often fails and it's a different story.

Introduced API will also make it possible to implement on other platforms in the future.

Changes:

- implement `isPowerLevelSafe()` method to detect if current power state is safe for node to operate (driven by NRF52 POF power failure detection circuit). This will be especially important to prevent any NodeDB/config saving on power failure conditions to prevent flash corruption. Note that this does not use battery pin ADC but internal chip protection circuits.

- prevent device boot when voltage is too low (boot will resume when voltage rises to safe level - for example on solar nodes). This will prevent bootloops. During waiting state - user led will flash in sequence to indicate device state.

- move NRF52 power failure threshold from 2.4V to 2.7V which is above automatic cutoff value for li-ion battery protectors so we do not get unexpected power cutoff

TODO:

- prevent any NodeDB saves during power failure condition. NRF52 automatically blocks flash writes - but Meshtastic interprets that as flash corruption and reboots and formats filesystem as soon as POF condition goes away. So we shoot ourselves in the foot with bazooka.

- put device to deep sleep when power failure is detected without waiting for classic battery readings and ADC voltage detection because it may be already too late for that. One example is bad battery going from 3.2V to 2.5V in few seconds or powering boards from various external regulators (and countless unexpected failure modes).

### DIFF
--- a/src/platform/nrf52/main-nrf52.cpp
+++ b/src/platform/nrf52/main-nrf52.cpp
@@ -35,6 +35,10 @@
 #define SAFE_VDD_VOLTAGE_THRESHOLD 2.7
 #endif
 
+#ifndef SAFE_VDD_VOLTAGE_THRESHOLD_HIST
+#define SAFE_VDD_VOLTAGE_THRESHOLD_HOST 0.2
+#endif
+
 // Weak empty variant initialization function.
 // May be redefined by variant files.
 void variant_shutdown() __attribute__((weak));
@@ -63,17 +67,10 @@ bool powerHAL_isVBUSConnected()
 
 bool powerHAL_isPowerLevelSafe()
 {
-    // some variants use AREF_VOLTAGE as 3.0
-    // but this is fine as we hunt for VDD values less than 3V
-    // otherwise either set here the reference for each measure and make sure
-    // same is done for battery reading (which long term should be implemented here as HAL function)
 
-    // we use the same values as regular battery read so there is no conflict on SAADC
-    analogReadResolution(BATTERY_SENSE_RESOLUTION_BITS);
+    uint16_t threshhold = SAFE_VDD_VOLTAGE_THRESHOLD * 1000; // convert V to mV
 
-    uint16_t vddADCRead = analogReadVDD();
-    float voltage = ((1000 * AREF_VOLTAGE) / pow(2, BATTERY_SENSE_RESOLUTION_BITS)) * vddADCRead;
-    LOG_INFO("VDD VOLTAGE: %f", voltage);
+
 
     return true;
 }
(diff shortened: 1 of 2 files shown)


src/platform/nrf52/main-nrf52.cpp
```diff
@@ -38,12 +38,56 @@ void variant_shutdown() {}
...
+
+    // Many NRF52 boards have decent LDO which goes down to 2V
+    // In the future - boards with crappy LDO can be set to prevent memory corruption at higher voltage - like 3V
+    // using custom variant definition. Remember that above 2.8V you need to monitor VDDH voltage threshold using different
+    // registers
+
+    // SoftDevice is only enabled by Adafruit Bluetooth library (Bluefruit) and there is no good way to change it or integrate with it.
+    // This is started at boot before bluetooth so we use raw registers instead of sd_power*
+
+    NRF_POWER->POFCON = ((POWER_POFCON_THRESHOLD_V27 << POWER_POFCON_THRESHOLD_Pos) | (POWER_POFCON_POF_Enabled << POWER_POFCON_POF_Pos));
```
rbomze: @phaseloop You can set `POFCON` with `sd_power_pof_threshold_set(...)`. Isn't it possible to run all your code after bluetooth is start up? and in case it's completely disabled (by EXCLUDE) just run other code, guarded by #ifdef

src/platform/nrf52/main-nrf52.cpp
```diff
@@ -192,9 +217,18 @@ extern "C" void lfs_assert(const char *reason)
...
+    // TODO: this will/can crash CPU if bluetooth stack is not compiled in or bluetooth is not initialized
+    // (regardless if enabled or disabled) - as there is no live SoftDevice stack
+    // implement "safe" functions detecting softdevice stack state and using proper method to set registers
     if (!(sd_power_gpregret_clr(0, 0xFF) == NRF_SUCCESS && sd_power_gpregret_set(0, NRF52_MAGIC_LFS_IS_CORRUPT) == NRF_SUCCESS)) {
         NRF_POWER->GPREGRET = NRF52_MAGIC_LFS_IS_CORRUPT;
     }
+
+    // TODO: this should not be done when SoftDevice is enabled as device will not boot back on soft reset
+    // as some data is retained in RAM which will prevent re-enabling bluetooth stack
+    // Google what Nordic has to say about NVIC_* + SoftDevice
```
rbomze: Isn't it possible to run `sd_softdevice_disable()` after Bluetooth is properly shutdown, so this is safe again?

phaseloop:
closing, will reopen later as new PR

PR #8797, opened 2025-11-29
Meshtastic Plugins

benallfree:
This PR introduces the concept of **Meshtastic Plugins**: a clean, decentralized way to extend Meshtastic without touching the core firmware.

Currently, the plugin registry is running as part of the https://meshforge.org/ project at https://registry.meshforge.org but I am open to moving it to an official registry if you guys like this concept. The pip package is also currently named `mesh-plugin-manager` but can be renamed to `meshtastic-plugin-manager` if you take it on officially. 

### Key features

- Plugins live completely outside the core. They are installed as needed via a simple `git clone` into `./plugins/<plugin-slug>/`.
- A new build helper script `./bin/mpm_pio.py` (Meshtastic Plugin Manager) is invoked automatically via a tiny change in `platformio.ini`:
  - Scans all plugin directories
  - Automatically injects them into the build
  - Discovers any `*.proto` files in plugins and generates the corresponding protobuf code
  - Allows plugin authors to use protobuf messages for plugin-specific data without ever modifying the core `meshtastic/protobufs` files
- `mpm_pio.py` can also be used as a standalone CLI tool (`python bin/mpm_pio.py list`, etc.) to list discovered plugins or manually generate protobuf artifacts

- New lightweight dynamic module registration system in `src/modules/ModuleRegistry.cpp`
  - Plugin authors simply add `MESHTASTIC_REGISTER_MODULE(MyModule)` in their code
  - No more need to edit `Modules.cpp` in the core

### Working examples

- `./src/plugins/sample-plugin` – a fully functional demo plugin included in this repo
- Real-world plugins already migrated:
  - https://github.com/MeshEnvy/lobbs
  - https://github.com/MeshEnvy/lodb
- Community plugin index maintained as part of the MeshForge project:  
  https://github.com/MeshEnvy/mesh-forge/blob/main/registry

### Next step (suggestion)

If we like this direction, we could spin up an official `registry.meshtastic.org` that serves the latest `registry.json`. Plugin authors would contribute via PRs to a canonical registry repository, making discovery seamless for users.

Looking forward to your feedback! This should make third-party extensions much easier and keep the core lean. There may even be existing modules that could be migrated out of the core.

> Note: CI will fail until the pip package is published

### DIFF
--- a/plugins/sample-plugin/src/SampleModule.h
+++ b/plugins/sample-plugin/src/SampleModule.h
@@ -1,8 +1,6 @@
 #ifndef SAMPLE_MODULE_H
 #define SAMPLE_MODULE_H
 
-#pragma MPM_MODULE(MySampleModule, sampleModule)
-
 #include "SinglePortModule.h"
 
 class MySampleModule : public SinglePortModule
--- a/plugins/sample-plugin/src/plugin.h
+++ b/plugins/sample-plugin/src/plugin.h
@@ -1,0 +1,7 @@
+#pragma once
+
+#include "SampleModule.h"
+
+#pragma MPM_MODULE(MySampleModule, sampleModule)
+
+#define SAMPLE_PLUGIN_VERSION "1.0.0"


fifieldt:
Thanks for being willing to work upstream :)

benallfree:
You're welcome. I'm closing this one though, as the approach is out of date.

PR #8800, opened 2025-11-29
Fix for high power drain after shutdown on Heltec T114. 

rbomze:
This PR ~limits the effects of PR #7882 to RAK4630 modules~ disables ADC monitoring of PR #7882 explicitly on the `Heltec T114` variant.

During power consumption measurements on my Heltec T114 i realized that the module consumes 2.9mA after initializing `cpuDeepSleep(portMAX_DELAY)`, which also happens after every shutdown (eg. by software or buttons).

The reason is that #7882 enables the ADC to monitor the battery voltage after power down. This drives ADC_CTRL to high and results in an consumption of 2.9mA on Heltec T114, because the resistor used at the base of the transistor is 1kOhm. ~~(3.3V / 0.0029A = 1138 Ohm)~~.

The drain calculates:
Base current through R9:  I_B = (3.3−0.7) V / 1  kOhm ≈ 2.6 mA
Resistive bleed through R25:  I_R25 = 3.3 V / 10 kOhm ≈ 0.33 mA
Total drawn from ADC_Ctrl ≈ 2.6 mA +0.33 mA ≈  2.93 mA
Which perfectly matches the measurement.

## 🤝 Attestations

  - [x] Heltec T114

### DIFF
--- a/variants/nrf52840/ELECROW-ThinkNode-M1/variant.h
+++ b/variants/nrf52840/ELECROW-ThinkNode-M1/variant.h
@@ -157,15 +157,15 @@ External serial flash WP25R1635FZUIL0
 #define PIN_GPS_STANDBY (32 + 2) // An output to wake GPS, low means allow sleep, high means force wake
 // Seems to be missing on this new board
 // #define PIN_GPS_PPS (32 + 4)  // Pulse per second input from the GPS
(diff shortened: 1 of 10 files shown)


phaseloop:
This is such a weird board design - driving Q8 ADC_CTRL transistor even every minute will consume much more energy than leaving this divider alone.

PR #8803, opened 2025-11-29
Reduce key duplication by enabling hardware RNG

Komzpa:
## Summary

- add portable HardwareRNG helper that taps NRF52 crypto RNG, ESP32/TRNG, RP2040 hwrand, Portduino host RNG, and modem-provided entropy as available
- stir hardware bytes into CryptRNG during key generation to avoid repeated keys on NRF devices
- route NRF52, Portduino, and RP2040 platform setup through the new helper so runtime random seeds also benefit
- expose RadioLibInterface::randomBytes so modem entropy can be mixed in when radios are initialized

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


jp-bennett:
I particularly like that this is just stirring in another potential randomness source. Will go through it in more detail as I am able.

src/platform/nrf52/main-nrf52.cpp
```diff
@@ -288,15 +289,13 @@ void nrf52Setup()
...
-    } seed;
-    nRFCrypto.begin();
-    nRFCrypto.Random.generate(seed.seed8, sizeof(seed.seed8));
-    LOG_DEBUG("Set random seed %u", seed.seed32);
-    randomSeed(seed.seed32);
-    nRFCrypto.end();
+    uint32_t seed = 0;
+    if (!HardwareRNG::seed(seed)) {
+        LOG_WARN("Hardware RNG seed unavailable, using PRNG fallback");
+        seed = random();
```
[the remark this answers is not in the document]
jp-bennett: This actually seems like good feedback. We might consider an XOR of the deviceID, MAC address, and micros() as the fallback seed.

src/mesh/HardwareRNG.cpp
```diff
@@ -0,0 +1,158 @@
...
+    if (generated == static_cast<ssize_t>(length)) {
+        filled = true;
+    }
+
+    if (!filled) {
+        fillWithRandomDevice(buffer, length);
+        filled = true;
+    }
+#else
+    fillWithRandomDevice(buffer, length);
```
jp-bennett: How good/bad of a source of entropy is std::random_device on various platforms? STM32 will fall over into this code, for instance. I think we want to just return false here instead.

src/mesh/CryptoEngine.cpp
```diff
@@ -25,6 +26,15 @@ void CryptoEngine::generateKeyPair(uint8_t *pubKey, uint8_t *privKey)
...
     // Mix in any randomness we can, to make key generation stronger.
     CryptRNG.begin(optstr(APP_VERSION));
+
+    uint8_t hardwareEntropy[64] = {0};
+    if (HardwareRNG::fill(hardwareEntropy, sizeof(hardwareEntropy))) {
+        CryptRNG.stir(hardwareEntropy, sizeof(hardwareEntropy));
+    } else {
+        LOG_WARN("Hardware entropy unavailable, falling back to software RNG");
+    }
+    memset(hardwareEntropy, 0, sizeof(hardwareEntropy));
```
phaseloop, line 37: Hardware entropy is filled then zeroed again always? Or should it be above in  "else" condition?

jp-bennett: The stir() function is where it's actually used, so setting back to nulls after use seems to be the correct behavior.

phaseloop: ah, makes sense

Komzpa:
I'll be happy if someone can take over this and get it live - we've got three new same-pubkey NRFs in the town nowadays

jp-bennett:
> I'll be happy if someone can take over this and get it live - we've got three new same-pubkey NRFs in the town nowadays

I like basically about this approach, and it's clearly needed. I'll take over and get it pushed across the line. Thanks!

Issue #8440, opened 2025-10-24
[Bug]: Station G2 - illegal TX power in every country configured by default


serialrf433:
### Category

Other

### Hardware

Station G2

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

Any since Station G2 support have been added

### Description

This is a followup from https://github.com/meshtastic/firmware/issues/7239

I first was really happy to see finally stopping those more and more illegal stations out there. Since mid of the year they got much more. And they are screaming with measured illegal TX power out in the air. This is bringing the whole Meshtastic project in danger!

The arguments against https://github.com/meshtastic/firmware/pull/8273 are fake. The people want to keep their illegal TX power to ignore the law. Now they want a second feature to still be able to do their illegal TX power thing. It should be now named 'loss gain compensation': https://github.com/meshtastic/firmware/pull/8332

People use really strange arguments like here:
https://github.com/meshtastic/firmware/pull/8273#issuecomment-3394840272

If you spend 3 digits on your meshtastic gear and it does not perform like on the main website from the company producing it, https://wiki.uniteng.com/en/meshtastic/station-g2#summary-for-lora-power-amplifier-conduction-test then open a warranty claim.
The Meshtastic firmware is not there to fix hardware production issues where the 'fix' is allowing the masses to use illegal high tx power.

This revert should have never happened https://github.com/meshtastic/firmware/pull/8332

The dev's of the project should have told that all should wait for that fake functionality implemented to make again illegal TX power possible. I hope such functionality wont ever be implemented. It would be abused all the time. My WLAN router also do not have such 'loss gain compensation' and it have a RP-SMA antenna. I have connected 10meter of the cheapest crap antenna cable to this RP-SMA connector. Should i now request at every OEM to build a function to make illegal transmission power possible?

If you want something illegal, then compile it on your own. There should not be prebuild images with illegal TX power capabilities build in.

### Relevant log output

```Shell

```

b8b8:
Not arguing for or against just merely pointing out that due to the inverse square law that radio waves follow... Quadrupling the power from 1W to 4W can only double the range and is still subject to line of sight (not taking into account further line losses).

PR #8805, opened 2025-11-29
Revert the Revert "Fix Station G2 Lora Power Settings"

serialrf433:
There is no need for any 'an additional lora param to handle line loss gain compensation'! https://github.com/meshtastic/firmware/issues/8445

This revert fixing the issue should never have been merged: https://github.com/meshtastic/firmware/pull/8332

This PR here is fixing https://github.com/meshtastic/firmware/issues/8440


serialrf433:
Do i have to create now a fake github account with a fake email to not have to publish my email address? This is getting insane. Just fix the issue i have spend endless hours into writing about it.

PR #8811, opened 2025-11-30
Plain RAK4631 should not compile EInk and TFT display code

thebentern:
I believe this is a safe assumption, but please correct me if I'm missing something.

### DIFF
--- a/variants/nrf52840/rak4631/platformio.ini
+++ b/variants/nrf52840/rak4631/platformio.ini
@@ -31,8 +31,6 @@ build_src_filter = ${nrf52_base.build_src_filter} \
   +<mqtt/> \
   -<graphics/EInkDisplay2.cpp> \
   -<graphics/EInkDynamicDisplay.cpp> \
-  -<graphics/TFTDisplay.cpp> \
-  -<graphics/tftSetup.cpp> \
   -<graphics/fonts/EinkDisplayFonts.cpp>
 lib_deps = 
   ${nrf52840_base.lib_deps}


Xaositek (approved):
Seems sane enough - good item to test in develop

jp-bennett:
The weird bit is that HAS_TFT really means MUI. Ideally we rename that, and then add another define, to indicate that we want to use the TFTDisplay adapter code. Seems like that would help this issue as well.

jp-bennett:
Let's split this into two PRs, since the TFTDisplay issue is a bigger one. I'll work on adding USE_TFTDISPLAY to all the variants that need it, and when that is ready, it should help the RAK4631 as well as everything else.

variants/esp32s3/heltec_v4/variant.h
```diff
@@ -34,6 +34,39 @@
 #define LORA_PA_EN 2
 #define LORA_PA_TX_EN 46 // enable tx
 
+#if HAS_TFT
```
mverch67, line 37: I think this won't work with MUI as MUI is linked in as a library. The device-ui library does not have any dependencies to the meshtastic repo (as any other library), i.e. it will not include variant.h which means all MUI related define macros have to be in platformio.ini

The heltec V4 related macros are used in this MUI file (I counted 9 macros):
https://github.com/meshtastic/device-ui/blob/3bf332240416c5cb8c919fac2a0ec7260eb3be75/include/graphics/LGFX/LGFX_HELTEC_V4_TFT.h#L115

jp-bennett: Understood. I'll flip this change back.

PR #8829, opened 2025-12-01
MESHTASTIC_REGISTER_MODULE macro

benallfree:
This PR introduces the `MESHTASTIC_REGISTER_MODULE()` macro which allows modules to register without needing to modify `Modules.cpp`


benallfree:
Copilot is correct, this seems to vary by platform. I'm going to keep experimenting with options.  The fundamental problem is that code can typically only be executed inside a function and attempts to use static initializers are optimized away because they aren't actually referenced anywhere.

benallfree:
Closing as not feasible.

PR #8845, opened 2025-12-03
Fix command syntax for erase and write flash

bmidi:
I've tested this when flashing a lilygo device. Changing the parameters allowed me to use device-install shell script to interact with esptool again.

esptool.py v4.8.1
    read_flash          Read SPI flash content
    write_flash         Write a binary blob to flash

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)

PR #8868, opened 2025-12-05
device-install: Consistently use write-flash

tyll:
Esptool accepts the write-flash command only on newer versions while older versions only accepted write_flash. Use write-flash consistently to avoid partial flashing with older esptool version.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/bin/device-install.sh
+++ b/bin/device-install.sh
@@ -32,6 +32,19 @@ if ! command -v jq >/dev/null 2>&1; then
     exit 1
 fi
 
+# esptool v5 supports commands with dashes and deprecates commands with
+# underscores. Prior versions only support commands with underscores
+if ${ESPTOOL_CMD} | grep --quiet write-flash
+then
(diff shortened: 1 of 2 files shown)


fifieldt:
Thanks for your contribution :)

This is fun! At the same time we also have a patch in to suggest the opposite: https://github.com/meshtastic/firmware/pull/8845

fifieldt:
Idea: can run esp tool, grep for the version, and then use the detected version to choose the best command syntax?

tyll:
Here is an updated version. I didn't test it, yet, as it seems to require new firmware names for installation (it is now checking for factory in the name).

fifieldt:
Thank you very much - let's test!

PR #8872, opened 2025-12-05
fix(MQTT): Send first MapReport as soon as possible

ndoo:
1. fix(MQTT): First MapReport does not get sent 
   
   Throttle::isWithinTimespanMs(last_report_to_map, map_publish_interval_msecs) is used to maintain the map reporting interval, but because last_report_to_map has an initial value of 0, the map report routine does not start until the system
millis() time has passed map_publish_interval_msecs.
   
   Fix this by adding a check that last_report_to_map is not 0.
2. feat(MQTT): Send MapReporting immediately upon location fix 
   
   Do not update last_report_to_map when Map Report is attempted without a valid location, as this results in waiting up to an hour (or configured Map Report interval).
   
   That usually happens because most nodes do not keep GPS warm, so GPS usually locks after the first attempt at Map Report.
   
   This change also results in the log WARNing message getting spammed until a location is obtained, so remove the message for now.

## 🤝 Attestations

  - [x] Other: Heltec T114v2
  - [x] Other: Lilygo T3-S3

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -53,6 +53,9 @@ static uint8_t bytes[meshtastic_MqttClientProxyMessage_size + 30]; // 12 for cha
 static bool isMqttServerAddressPrivate = false;
 static bool isConnected = false;
 
+static uint32_t lastPositionUnavailableWarning = 0;
+static const uint32_t POSITION_UNAVAILABLE_WARNING_INTERVAL_MS = 15000; // 15 seconds
+
 inline void onReceiveProto(char *topic, byte *payload, size_t length)
 {
     const DecodedServiceEnvelope e(payload, length);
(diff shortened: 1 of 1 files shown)


ndoo:
After discussion with Ixitxachitl on Discord, it looks like Map Reporting does get sent, but only after the millis() timer has passed the first interval.

I can't think of any good reason not to send the first possible map report instead of letting the clock tick up to the first interval.

So, keeping the code the same, but re-wording the PR and commit to match the actual fix (at least, how I perceive it).

Please help to request re-review.

fifieldt (approved):
This makes logical sense to me

src/mqtt/MQTT.cpp
```diff
@@ -841,14 +841,11 @@ void MQTT::perhapsReportToMap()
         map_position_precision = default_map_position_precision;
     }
 
-    if (Throttle::isWithinTimespanMs(last_report_to_map, map_publish_interval_msecs))
+    if (Throttle::isWithinTimespanMs(last_report_to_map, map_publish_interval_msecs) && last_report_to_map != 0)
         return;
 
-    if (localPosition.latitude_i == 0 && localPosition.longitude_i == 0) {
-        last_report_to_map = millis();
```
GUVWAF, line 852: The reason this was there was to avoid spamming the debug line below when there is no position available. 
However, I still think it's useful to have this line. Maybe we can do the throttle for the printing of the debug line?

fifieldt: good idea. there is throttle logic in gps/rtc.cpp if you want to copy that

ndoo:
Thank you for the feedback i’ll give it a go later today.

src/mqtt/MQTT.cpp
ndoo: Added as suggested, using the same throttle interval as GPS time.

GUVWAF (approved):
Looks good to me now, thanks!

ndoo:
Hi there, I think this PR got missed for a few release cycles, would be a nice enhancement at least for myself.

h3lix1:
@ndoo I found my tests not working and traced it to this change. Created diff https://github.com/meshtastic/firmware/pull/9784 to help fix things, but you can make your own.

ndoo:
> @ndoo I found my tests not working and traced it to this change. Created diff https://github.com/meshtastic/firmware/pull/9784 to help fix things, but you can make your own.

Let me know if you need any action on my part, thanks for the fix.

h3lix1:
Looks like the diff was merged. No changes needed unless someone wants to improve it.

Issue #8732, opened 2025-11-24
[Bug]: LiliGo T-Beam consumes 89+% of the heap when wifi is enabled on 2.7.15


samm-git:
### Category

Other

### Hardware

T-Beam

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.15.567b8ea

### Description

LiliGo T-Beam consumes 89+% of the heap just after start if WIFI/MQTT is enabled. I tried to reflash with full erase (as some report suggests) and reset node db but it does not have any visible impact.

Some tests (measure done after start with emty nodedb):

1. Device with WIFI disabled: **61%** of the HEAP in use
2. Device with WIFI enabled but MQTT disabled: **85%** 
3. Device with both WIFI and MQTT enabled: **86%**

korbinianbauer:
> I'm not sure if it is expected behavior, so I decided to report

Can you collect those numbers for the last stable release for comparison?

samm-git:
@korbinianbauer  sure, how to get it from 2.6? I do not see anything like that in the menu

ndoo:
Seeing 92% heap usage on Lilygo T3 V1.6.1 as well, when Wi-Fi and MQTT are enabled.

Only around 60+% with Wi-Fi off, MQTT module enabled (but not functional).

MartinEmrich:
Facing the same issue on a Lilygo T3 V2 1.6.1 (which has no PSRAM).

After a few hours with `git bisect`, I narrowed it down to around commits 70ac3601b04ef3f9243777fde545befbe254c08e (last one working fine) and 99770354995c9b463fbf570993b7e67289f65dc9 (bad); the few in between do not link.

The working ones report ~189000 bytes of heap, the bad ones later only 169000 bytes. the heap usage never changes, but the available heap is shrunk by ~20k.

I will play around with this a bit...

samm-git:
@MartinEmrich my case seems to be different, as i do have PSRAM, but seems that its not really in use for the heap, etc

samm-git:
Ok, I was able to reduce HEAP usage back to 76% on my device.

I did two changes: 
1. Enabled dynamic allocator for the devices with PSRAM (getting 82-84% consumption)
[...]

Issue #8729, opened 2025-11-23
[Bug]: Configuring wifi on LILYGO T-LoRa V2.1-1.6 makes node unreachable for clients


VanceVagell:
### Category

WiFi

### Hardware

T-Lora v2 1.6

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.15.567b8ea Alpha

### Description

I installed firmware 2.7.15.567b8ea Alpha because I was facing this bug: https://github.com/meshtastic/Meshtastic-Android/issues/3667 which was preventing me from configuring wifi at all (Android app just hung when attempting to). The alpha firmware resolved that issue, and allowed me to configure wifi (while connected via Bluetooth), and I can indeed see that my LILYGO device is assigned an IP on my network.

However, the LILYGO node is now unreachable via either bluetooth OR network connection. The Android app shows that the LILYGO device is on my local network and attempts to connect to it, but it hangs with the circular spinner spinning indefinitely while trying to connect via the network. And attempting to tap the Bluetooth connection I previously used with the device does nothing (no visual feedback), so the node is unreachable.

I can reflash it and use Bluetooth only, but then I can't use MQTT with this node.

### Relevant log output

```Shell

```

VanceVagell:
More information:

The Android app says "Module config (10 / 13)" when trying to connect to the device via the wifi network, it just hangs there with the circular spinner. After maybe a minute, the LILYGO device resets (it shows the Meshtastic logo on the screen and reboots).

The debug log from the LILYGO via USB when I'm trying to connect to it in the Android app via wifi looks like this:

DEBUG | 19:40:05 36 [RadioIf] Corrected frequency offset: 3079.667725
DEBUG | 19:40:05 36 [RadioIf] Lora RX (id=0xd9d02bf7 fr=0x088aaa7c to=0xffffffff                                                                                                 , transport = 0, WantAck=0, HopLim=2 Ch=0x8 encrypted len=99 rxSNR=10 rxRSSI=-93                                                                                                  hopStart=3 relay=0xb4)
DEBUG | 19:40:05 36 [RadioIf] Packet RX: 969ms
[...]

PR #8880, opened 2025-12-06
Revert to old dynamically allocated memory for low-memory ESP32 devices

MartinEmrich:
Potentially fixes: #8732 #8729

Effectively reverts https://github.com/meshtastic/firmware/commit/9211b1bb4b0cf9fcfa6a889cbab49938743f7414 
for ESP32 devices without PSRAM, like my Lilygo V2 1.6.1.

All versions later have ~20kB less heap, making them slow and unresponsive when using with Wifi (App cannot connect, using API is very slow, regular network dropouts).

This reverts that change for plain ESP32 devices that have no PSRAM, which I assume would be used if present.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Lilygo T3 V2 1.6.1

### DIFF
--- a/boards/heltec_v4.json
+++ b/boards/heltec_v4.json
@@ -9,7 +9,7 @@
     "extra_flags": [
       "-DBOARD_HAS_PSRAM",
       "-DARDUINO_USB_CDC_ON_BOOT=1",
-      "-DARDUINO_USB_MODE=0",
+      "-DARDUINO_USB_MODE=1",
       "-DARDUINO_RUNNING_CORE=1",
       "-DARDUINO_EVENT_RUNNING_CORE=1"
     ],


MartinEmrich:
After leaving the device running for a while, the heap usage still creeps up, and at the 88-90% mark the device starts to drop IP connectivity (Mesh node activity seems to be less affected).

jp-bennett:
It would be nice to identify what is leaking memory.

MartinEmrich:
@jp-bennett TBH I don't think it's a memory leak in the classical sense (memory gets malloc()d and forgotten), but the dynamic allocation over time allocates the same amount of memory as was pre-allocated without that patch. It was worth a try, but we (me and some participants in the linked issues) now think it its also caused by some bloat creeping in since 2.6.x, and a feature-reduced build for non-PSRAM devices (or an expert optimizing the code for memory usage) is probably necessary.

If a maintainer clearly agrees, I am happy to close the PR without a merge, clearly even in my less-populared area the memory allocated quickly reaches the same amount as pre-allocated.

MartinEmrich:
The change definitely has not the intended effect for more than a few minutes, even stripping out other optional features does not. So I cancel this, sorry to bother...

PR #8884, opened 2025-12-07
Fix #8883 (lora-Pager fter playing the notification, voltage does not disappear from the speaker)

### LINKED ISSUE #8883: [Bug]: Tlora-Pager fter playing the notification, voltage does not disappear from the speaker
### Category

Other

### Hardware

Other

- [ ] InkHUD ePaper

### Firmware Version

2.7.15.567b8ea

### Description

Are there any reasons to check the condition:
`audioThread->isPlaying()`
Before stopping audioThread?

The thing is, if no melody is playing, then audioThread never stops.
Because of this, the speaker pin is never set to LOW state:

```
class AudioThread : public concurrency::OSThread
method: stop()

#ifdef T_LORA_PAGER
        LOG_DEBUG("AudioThread: EXPANDS_AMP_EN, LOW");
        io.digitalWrite(EXPANDS_AMP_EN, LOW);
#endif
```

### Relevant log output

```Shell
ExternalNotificationModule::stopNow()
#ifdef HAS_I2S
    LOG_INFO("Stop audioThread playback");
    if (audioThread->isPlaying())
        audioThread->stop();
#endif
```

polarikus:
Fix #8883

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [x] LilyGo T-LoraPager

### DIFF
--- a/src/AudioThread.h
+++ b/src/AudioThread.h
@@ -50,8 +50,11 @@ class AudioThread : public concurrency::OSThread
             delete i2sRtttl;
             i2sRtttl = nullptr;
         }
-        delete rtttlFile;
-        rtttlFile = nullptr;
+
+        if (rtttlFile != nullptr) {
+            delete rtttlFile;
+            rtttlFile = nullptr;
+        }
 
         setCPUFast(false);
 #ifdef T_LORA_PAGER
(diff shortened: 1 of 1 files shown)


polarikus:
If this solution doesn't work for you, let's think about alternatives. My suggestions:

1. Add a new method to AudioThread, something like "enableSpeaker", and call it where needed in NotificationModule.cpp.
2. Move the power management of EXPANDS_AMP_EN completely to NotificationModule, but that might not look right - it doesn't seem like its responsibility.
3. Move the low-level control of EXPANDS_AMP_EN somewhere else, for example, add a wrapper around AudioOutput.

mverch67 (approved):
Looks OK to me. If audioThread->stop() would never be called it would also leak memory (rtttlFile) which was allocated in beginRttl().

thebentern (approved):

PR #8261, opened 2025-10-08
Increase bluetooth 5.0 PHY speed and MTU on esp32_s3

h3lix1:
- Added ESP32-S3-only startup tuning so we explicitly request MTU 517 (was stack default 23), a 2 M PHY (was 1 M), and 251-byte PDU suggested data length (was the legacy 27-byte default) when bringing up NimBLE 

- On every new connection we now immediately re-request 2 M PHY (instead of leaving the link on 1 M) and 251-byte data length, while logging the initial/target MTU to confirm the negotiation

- Will negotiate down if needed for devices that do not support Bluetooth 5.0

Tested on ESP32-S3 hardware and observed the BLE logs to confirm the negotiated MTU/PHY/data length

## 🤝 Attestations

- [X] I have tested that my proposed changes behave as described.
- [X] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [x] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Seeed Studio T-1000E tracker card
  - [x] Other (please specify below)
Heltec (Lora32) V4

Tested on the following phones:

- [x] Motorola Edge 2024
- [x] iPhone 15 Pro Max

Issue #8876, opened 2025-12-06
Showcase: reducing ESP32 power usage by 40% without sleeping


phaseloop:
### Platform

ESP32

### Description

With few bluetooth timing tweaks and migrating to ESP-IDF V5.5 we can cut ESP32 node power usage down to 60 mA and later to 50 mA with some FreeRTOS tweaks.

This requires migrating to pioarduino because PlatformIO refuses to support newer arduino-esp32 3.x versions and we are forced to use ancient and buggy ESP-IDF versions.

This is already done for C6 variants. Nothing crazy with adapting codebase to Arudino 3.x framework - mostly using different ADC mode for battery status.

I did an experiment on my branch with some quick migration hacks and BLE timing changes.

Result below - Heltec V3 before and after, BLE enabled and phone connected.

![Image](https://github.com/user-attachments/assets/3b73cf85-c105-440a-a1be-3604fbbd9307)

![Image](https://github.com/user-attachments/assets/1765c510-8340-4500-a894-5589b2a63c89)

korbinianbauer:
Cool, what what exactly does this do differently to save so much power?

miky2k:
> Cool, what what exactly does this do differently to save so much power?

You need to ask to expressif , IDF is closed source

korbinianbauer:
> > Cool, what what exactly does this do differently to save so much power?
> 
> You need to ask to expressif , IDF is closed source

Ok, let me rephrase.

@phaseloop What exactly did you do to save so much power? Just use a newer IDF version and that's it? Or does the newer IDF version provide new/different functionality that you explicitly used?

phaseloop:
The firmware I uploaded had both IDF updated and BLE advertisement interval changed (from 50 ms I guess to around 300 ms) - frankly I'm not sure if this power saving is due to IDF or both changes. I did not investigate it further for now. I don't expect such drastic change to be only caused by interval change.

Changing advertisement interval did nothing on IDF v4 when it comes to power usage which was weird so I assumed there were some bugs with radio not sleeping during interval pauses - it seems I was right. Also IDFv5 and arduino 3.x framework allow to compile and enable "FreeRTOS tickless idle" which will allow to save power even more.

MartinEmrich:
Looks like I took a similar route, with similar incompatibilities discovered...

I just got my Heltec V4 working with current Espressif Toolstack via Pioarduino, dropped idle consumption from 110mA to 48mA

[...]

PR #8890, opened 2025-12-07
Guard 2M PHY mode for NimBLE

thebentern:
Original change introduced by #8261 
We have seen a number of reports of issues with particularly desktop PC / Mac bluetooth regressions due to the PHY increase. This change macro-guards that mode. 
If someone wishes to use the 2M PHY mode for NimBLE, they can add `-DNIMBLE_ENABLE_2M_PHY=1` to their PIO ini environment build flags.

### DIFF
--- a/src/nimble/NimbleBluetooth.cpp
+++ b/src/nimble/NimbleBluetooth.cpp
@@ -671,16 +671,17 @@ class NimbleBluetoothServerCallback : public NimBLEServerCallbacks
 
         LOG_INFO("BLE conn %u initial MTU %u (target %u)", connHandle, connInfo.getMTU(), kPreferredBleMtu);
         pServer->updateConnParams(connHandle, 6, 12, 0, 200);
-#endif
     }
+#endif
 
+#ifdef NIMBLE_TWO
     virtual void onDisconnect(NimBLEServer *pServer, NimBLEConnInfo &connInfo, int reason)
     {
         LOG_INFO("BLE disconnect reason: %d", reason);
 #else
-virtual void onDisconnect(NimBLEServer *pServer, ble_gap_conn_desc *desc)
-{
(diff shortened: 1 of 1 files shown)


phaseloop:
Just FYI - this is 99% caused by buggy ESP-IDF version as we are using Arduino 2.x and ESP-IDF V4.X

I advocate for some plan to move to pioarduino based on ESP-IDF V5.5 which has a ton of bugs fixed, both for BT and power management. 

Example: https://github.com/meshtastic/firmware/issues/8876


There is also suspicion (by me) that some NRF52 boards with external LF crystal (for example T114) have faulty crystals throwing off bluetooth connections on some faster timings. This was reported already by several Nordic customers. So there is mess to manage on both platforms :/

thebentern:
> Just FYI - this is 99% caused by buggy ESP-IDF version as we are using Arduino 2.x and ESP-IDF V4.X
> 
> I advocate for some plan to move to pioarduino based on ESP-IDF V5.5 which has a ton of bugs fixed, both for BT and power management.
> 
> Example: #8876
> 
> There is also suspicion (by me) that some NRF52 boards with external LF crystal (for example T114) have faulty crystals throwing off bluetooth connections on some faster timings. This was reported already by several Nordic customers. So there is mess to manage on both platforms :/

Would you mind propoosing that PR separately @phaseloop. I'm sure @caveman99 has thoughts as well :-)

phaseloop:
> Would you mind propoosing that PR separately @phaseloop. I'm sure @caveman99 has thoughts as well :-)

I created a GH issue to discuss it along with migration plan. AFAIK this aligns well with idea to move ADC and power management functions to each `/platform` codebase and not litter main files with hundreds of #ifdefs

https://github.com/meshtastic/firmware/issues/8896

PR #8893, opened 2025-12-07
Add keyboard layout [Ru, Pl, Cz, Ua] for T-Deck and CardKB devices

timurey:
Hey!
I've implemented the ability to change the keyboard layout for locales supported by Meshtastic fonts: RU, UA, PL, CZ.
It works on devices with CardKB and T-Desk, the keyboard shortcut for changing the layout is fn-L.

In addition to adding a layout change handler and a mapping table for various layouts, I've added code to support UTF-8 strings in the message input field.

I can't add support for other input devices, but I think this will be a good start for adding other layouts and expanding the supported input device types in different languages.

I only tested the firmware on devices with high resolutions (e-ink, T-Deck), but it should work well on OLED displays as well.
out

  - [x] T-Deck
  - [x] Heltec E213
  - [x] Heltec E290
  - [x] NRF52 Pro-micro DIY


fifieldt:
Thanks, this looks like a nice implementation.

jp-bennett:
I don't love that the character replacement logic is in cannedMessages. That module is already too much of a catch-all.

timurey:
Great point, I completely agree. cannedMessages is indeed turning into a module that does too much. The character replacement logic for keyboard layouts is a separate responsibility.

My proposal: I'll move it out into a dedicated kbLayout.c/.h module, as you suggested.
It will expose a function, ```const char *applyCurrentLayout(char key)``` which cannedMessages can call.
Also I’ll move keyboard layouts mapping tables, so all layout-specific data will be encapsulated in the new module. 
The current module will have a layout indicator and utf8 character processing functions

timurey:
Sorry for messy commit history

timurey:
Hey, @jp-bennett @fifieldt, please review the changes if possible.

timurey:
There was a strange changes in last commit, constant changed to a magic number: https://github.com/meshtastic/firmware/commit/9da4396c6f600a00e9519dbcbc0698205fa50fdb#r173521514
```uint16_t maxChars = meshtastic_Constants_DATA_PAYLOAD_LEN - (moduleConfig.canned_message.send_bell ? 1 : 0);```
changed to
```const uint16_t maxChars = 200 - (moduleConfig.canned_message.send_bell ? 1 : 0);```
@HarukiToreda is it a bug?

PR #8906, opened 2025-12-09
Use truncated position for smart position

RCGV1:
If we have imprecise location and smart position on lets not send smart position even if we move if our new position will get truncated to the previous position. We now take into account truncating when deciding the distance travelled for smart pos.
Tested by walking around and sending client notification if the moved distance is greater than the threshold but the truncated position isnt.

This is useful for reducing the amount of positions that are sent.
IMG_5186

## 🤝 Attestations

- [ x ] I have tested that my proposed changes behave as described.
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ x ] RAK WisBlock 4631

### DIFF
--- a/src/modules/PositionModule.cpp
+++ b/src/modules/PositionModule.cpp
@@ -45,8 +45,12 @@ bool PositionModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mes
 {
     auto p = *pptr;
 
-    // If inbound message is a replay (or spoof!) of our own messages, we shouldn't process
-    // (why use second-hand sources for our own data?)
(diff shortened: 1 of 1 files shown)


phaseloop:
Just FYI as this may be another edge case but I'm not into GPS code in Meshtastic - GPS devices and phones also report DOP (dilution of precision) and we should not trigger smart position send if previous and new position are within uncertainty radius of GPS position.

I think someone a week ago reported here a situation when node with poor GPS reception "travels" constantly on a map and sends unnecesary position updates.

thebentern (approved):

PR #8858, opened 2025-12-04
Cut NRF52 bluetooth power usage by 300% - testers needed!

phaseloop:
According to Nordic online power profiler and general industry recommendation for low power BLE devices (see below), this change should reduce raw bluetooth subsystem power usage by at least 300%. 

In advertisement mode (no phone connection) simulated current drop should be from 146 uA to 56 uA for the RF layer only. 

When phone is connected to node (theoretical) yield should be around 400% as node is allowed to skip up to 5 extra interrogation/keep-alive frames longer if there is no data to send (so called slave latency).

Physics behind it:

https://docs.silabs.com/bluetooth/2.13/bluetooth-general-system-and-performance/optimizing-current-consumption-in-bluetooth-low-energy-devices

Changes introduced:

- increase advertisement frame interval from 152.5 ms to 417.5 ms for slow discovery mode - which is still in the lower range of available BLE intervals and way within values recommended by Apple (link in the source code). For many low power BLE devices intervals starts above 500 ms.

This change to be in effect was confirmed by nRF Connect logs. I don't see any visible impact on Meshtastic app connection/discovery times to node or general Android visibility of the device.

- increase Slave Latency factor to 5 (from default value of 0), which means node can freely ignore up to 5 keep-alive frames from a phone when it has no data to transmit - and it will not affect anything with connection stability. Which means we don't need to waste power sending BLE frames that do nothing.

Screenshot by SiLabs from link above:

slave-latency

While this is a generic industry knowledge and we know power savings are there, it would be good if someone with a decent power profiler can check how theoretical yield looks in practice. It would be great if other people can test this change with their phones and nodes.

Those are still pretty conservative settings so we can have further fun once this is verified to be stable across general population.

I suppose similar change to ESP32 may bring much drastic power savings.

## 🤝 Attestations

- [X] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
 
  - [X] Promicro/NiceNanoV2 - Faketec custom board

PR #8913, opened 2025-12-09
Revert "Cut NRF52 bluetooth power usage by 300% - testers needed!"

thebentern:
Reverts meshtastic/firmware#8858

Having issues on iOS


neopiccolorat:
Does it work on Android? I wish I could test.
Is it going to be implemented in a new pull request?

PR #8940, opened 2025-12-11
More blinkenlights work for Thinknode-m3

jp-bennett:
Uses the green LED as a message indicator, and adds a fast blink critical battery warning.


src/mesh/NodeDB.cpp
```diff
@@ -805,11 +805,11 @@ void NodeDB::installDefaultModuleConfig()
     moduleConfig.external_notification.output_ms = 500;
     moduleConfig.external_notification.nag_timeout = 2;
 #endif
-#if defined(RAK4630) || defined(RAK11310) || defined(RAK3312) || defined(MUZI_BASE)
-    // Default to RAK led pin 2 (blue)
+#if defined(RAK4630) || defined(RAK11310) || defined(RAK3312) || defined(MUZI_BASE) || defined(ELECROW_ThinkNode_M3)
```
fifieldt, line 808: I look forward to a future where maybe we can check for PIN_LED2 being defined here or something.

src/modules/StatusLEDModule.cpp
```diff
@@ -26,7 +26,11 @@ int StatusLEDModule::handleStatusUpdate(const meshtastic::Status *arg)
                 power_state = charged;
             }
         } else {
-            power_state = discharging;
+            if (powerStatus->getBatteryChargePercent() > 5) {
```
fifieldt, line 29: FWIW, I looked up LI ion battery curves and 5% seems to be the start of the precipitous drop, so feels like the right number here.

PR #8772, opened 2025-11-27
Make GPS_TX_PIN the serial TX and GPS_RX_PIN the serial RX for all NRF variants

Stary2001:
https://github.com/meshtastic/firmware/pull/8486 inadvertantly changed the behaviour of the GPS_TX_PIN and GPS_RX_PIN defines on non-ESP32 architectures - GPS_TX_PIN is now the serial TX pin and GPS_RX_PIN is the serial RX pin, rather than being some combination defined per variant. This was already the behaviour on ESP32 platforms.
Try to straighten it out.

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)

PR #8951, opened 2025-12-12
Fix for GPS not using pins defined in variant.h if set as "Pin 0" through app. ONLY WORKING ON PROMICRO

macvenez:
**Don't accept** as it will break GPS on all boards but nrf52_promicro_tcxo

As discussed on discord [here](https://discord.com/channels/867578229534359593/1202456506587418664/1449057901619712222) this should fix the GPS issue.

Tested and working on `nrf52_promicro_tcxo` **won't work** on other boards as pins are swapped now

Had to swap them again in promicro. How do we want to go with the other boards?


Xaositek:
Also let's confirm I think we have this done? Or what does it materially change?
https://github.com/meshtastic/firmware/pull/8772
https://github.com/meshtastic/firmware/pull/8760

NomDeTom:
I need to check this properly. The premise is correct - the `#define` is incorrect and not getting picked up properly. However, swapping the pins needs investigating.

PR #7180, opened 2025-06-30
High performance arduino libs for RGB displays

mverch67:
Experimental code for Seeed Indicator to avoid display [drift](https://github.com/esp-arduino-libs/arduino-esp32-sdk?tab=readme-ov-file#for-sdks-suffixed-with--h):
- arduino-esp32 version 3.3 (ESP-IDF v5.5.0) 
- esp-arduino-libs (see https://github.com/esp-arduino-libs/arduino-esp32-sdk)
- pioarduino platform (see https://github.com/pioarduino/platform-espressif32)

## Status
- [x]  arduino 3.3.5 (esp-idf v5.5) packages and config
- [x] 3.3 API library migration
- [x] io expander I2C migration
- [x] hw rotation
- [ ] CI build
- [x] bring-up
  - [x] task watchdog
  - [x] radiolib interrupts
  - [x] bluetooth
  - [x] wifi
  - [ ] SPI gpio
  - [ ] 120 MHz bootloader
  - [x] MUI
  - [ ] BaseUI 

## 🤝 Attestations

- [ ] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)

PR #8952, opened 2025-12-12
Migrate esp32 families to pioarduino platform

vidplace7:
Switch all ESP32 builds to [pioarduino/platform-espressif32](https://github.com/pioarduino/platform-espressif32).

https://docs.espressif.com/projects/arduino-esp32/en/latest/migration_guides/2.x_to_3.0.html

Switching to pioarduino will be necessary in order to support newer chipsets (esp32h2, esp32c5, esp32c6, esp32p4, etc)

**DRAFT** for now; this mostly serves as a CI probe to see what's failing in Arduino 3.x

TODO:
- [x] ADC API Changes
- [ ] CI Work
  - https://github.com/pioarduino/platform-espressif32/blob/main/.github/workflows/examples.yml

Depends on:
- Remove `syslog` lib  (replace?) 
  - #8964 
- NimBLE
  - #8983 
  - #7880
  - #9118
- libpax
  - Fixed in @mverch67 [fork](https://github.com/mverch67/libpax). PR upstream?


mverch67:
Check also my work for https://github.com/meshtastic/firmware/pull/7180

default_8MB.csv
```diff
@@ -0,0 +1,7 @@
+# Name,   Type, SubType, Offset,  Size, Flags
+nvs,      data, nvs,     0x9000,  0x5000,
+otadata,  data, ota,     0xe000,  0x2000,
+app0,     app,  ota_0,   0x10000, 0x330000,
+app1,     app,  ota_1,   0x340000,0x330000,
+spiffs,   data, spiffs,  0x670000,0x180000,
+coredump, data, coredump,0x7F0000,0x10000,
```
mverch67, line 7: Note that enabling the core dump requires some further logic to **prevent** writing core dumps again and again when the device got into a boot loop..

phaseloop:
just fyi - using pioarduino forces use of pioarduino-core and in vs code automatically installs Pioarduino IDE. So stupid pissing contest betweeen platformio and espressif will force us to rework pipelines and different core while this is only one of many platforms we use.

Maybe better solution is to use tasmota platform-espressif32 which is maintained by Espressif employees - it's a drop-in replacement for pioarduino and also supports hybrid compile for ESP-IDF. Myself and few other people here did experiments with that.

vidplace7:
"Merging" this into a new branch `pioarduino` for better collaboration. (Will re-open PR from this new branch)

PR #8953, opened 2025-12-12
Add SNR, RSSI and Hops to the Messages screen

### LINKED ISSUE #8949: [Feature Request]: Show RSSI & SNR or hops in Messages view
### Platform

ESP32

### Description

On the BaseUI, it would be nice to have the RSSI and SNR (or number of hops) displayed in the Message view like on the Android app.
This would be useful for the Range Test on the receiver side to check the metrics directly on the screen.

romainreignier:
Following my Feature Request in #8949, I went ahead in tried to implement it.

I  have only Heltec V3 nodes so I have tested the code only on them.

I understand that the display is small and wasting 1 line for these metrics might not be important for everyone.
If you think so, you can just ignore this PR and close it.

I have not modified the code path specific to the M5Stack C6L Unit because I could not test it.

I got the inspiration from the Android App and I have used the same logic to compute the `hopsAway` and display it if not a direct message. But I have kept the SNR and RSSI data displayed in case of hops because it is still useful for my use case using the Range Test Module.

Preview:

IMG_20251212_185329

Fixes #8949

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/graphics/draw/MessageRenderer.cpp
+++ b/src/graphics/draw/MessageRenderer.cpp
@@ -218,6 +218,15 @@ void drawTextMessageFrame(OLEDDisplay *display, OLEDDisplayUiState *state, int16
     }
 
     // === Header Construction ===
+    char metricsStr[64];
+    // Logic from Android app
+    int hopsAway = (mp.hop_start == 0 || mp.hop_limit > mp.hop_start) ? -1 : mp.hop_start - mp.hop_limit;
+    if(hopsAway == 0) {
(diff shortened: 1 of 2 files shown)


fifieldt:
Thanks for being a great collaborator & following the process :) Will let someone with more experience in the baseUI comment about the screen real estate.

Xaositek:
I think this is a pretty good implementation. I would say let's take a look at this after we land Multi Message as its going to really modify how this whole thing lays out.

Issue #5767, opened 2025-01-06
[Bug]: GPS Buffer full


cracky22:
### Category

Hardware Compatibility, Other

### Hardware

Seeed Card Tracker T1000-E

### Firmware Version

2.5.18

### Description

WARN | 10:26:13 342 [GPS] GPS Buffer full with 1023 bytes waiting. Flush to avoid corruption

### Relevant log output

```Shell
WARN | 10:26:13 342 [GPS] GPS Buffer full with 1023 bytes waiting. Flush to avoid corruption
```

cracky22:
```
DEBUG | 10:29:59 568 [Router] Module 'routing' considered

INFO | 10:29:59 568 [GPS] GPS power state move from IDLE to ACTIVE

WARN | 10:29:59 568 [Position] Ch. util >25%. Skip send

DEBUG | 10:29:59 568 [RadioIf] Lora RX (id=0x49311782 fr=0xda63587c to=0xffffffff, WantAck=0, HopLim=3 Ch=0x1e encrypted len=49 rxSNR=10 rxRSSI=-34 via MQTT hopStart=7)

DEBUG | 10:29:59 568 [RadioIf] Packet RX: 600ms

DEBUG | 10:29:59 568 [RadioIf] rx_snr found. hop_limit:5 rx_snr:9.750000

DEBUG | 10:29:59 568 [RadioIf] rx_snr found in packet. Setting tx delay:5467

DEBUG | 10:29:59 568 [Router] Add packet record (id=0x49311782 fr=0xda63587c to=0xffffffff, WantAck=0, HopLim=3 Ch=0x1e encrypted len=49 rxSNR=10 rxRSSI=-34 via MQTT hopStart=7)

DEBUG | 10:29:59 568 [Router] Use channel 1 (hash 0x1e)

DEBUG | 10:29:59 568 [Router] Use AES128 key!

DEBUG | 10:29:59 568 [Router] decoded message (id=0x49311782 fr=0xda63587c to=0xffffffff, WantAck=0, HopLim=3 Ch=0x1 Portnum=3 rxtime=1736155799 rxSNR=10 rxRSSI=-34 via MQTT hopStart=7)

[...]

PR #8956, opened 2025-12-12
Fix GPS Buffer full issue on NRF52480 (Seeed T1000E)

fifieldt:
We set the buffer size to 1024 on NRF52480, less than other platforms:

esp32.ini:  -DSERIAL_BUFFER_SIZE=4096
esp32c6.ini:  -DSERIAL_BUFFER_SIZE=4096
nrf52.ini:  -DSERIAL_BUFFER_SIZE=1024

However, 115200 baud, like the T1000e uses is about 12 times that - almost 15 bytes per millisecond. 15 bytes * 200 millisecond (our GPS poll rate)  = 3000 bytes, which is longer than our buffer on the nrf52 platform. This causes "GPS Buffer full" errors on the T1000e and other devices based on NRF52480 with newer GPS chips.

This patch increases SERIAL_BUFFER_SIZE for nrf52480 to 4096 to align with other platforms. It keeps the original 1024 for the nrf52832, which has fewer resources.

Fixes https://github.com/meshtastic/firmware/issues/5767

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/variants/nrf52840/nrf52.ini
+++ b/variants/nrf52840/nrf52.ini
@@ -19,7 +19,6 @@ build_type = release
 build_flags =
   -include variants/nrf52840/cpp_overrides/lfs_util.h
   ${arduino_base.build_flags}
-  -DSERIAL_BUFFER_SIZE=1024
   -Wno-unused-variable
   -Isrc/platform/nrf52
   -DLFS_NO_ASSERT                      ; Disable LFS assertions , see https://github.com/meshtastic/firmware/pull/3818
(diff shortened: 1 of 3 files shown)


fifieldt:
This setting has been tested long-term (1 month continuous) on the T1000E by @cmastro25 who confirms it fixes the issue and introduces no other problems.

drupol:
Should we do the same for the wismesh tag or it is not necessary?

fifieldt:
> Should we do the same for the wismesh tag or it is not necessary?

Hi @drupol , the wismesh tag will also have its buffer extended by this change. However, it looks like the gps module in the tag, wishmesh tag, only uses 9600 baud, so it wouldn't have been affected by the bug.

drupol:
Thank you.

PR #8958, opened 2025-12-13
Wait for location to set time

jp-bennett:
Some GPS will give a bogus time right from boot. Consider waiting for an initial position before setting time.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -1133,13 +1133,10 @@ int32_t GPS::runOnce()
         // if gps_update_interval is <=10s, GPS never goes off, so we treat that differently
         uint32_t updateInterval = Default::getConfiguredOrDefaultMs(config.position.gps_update_interval);
 
-        // 1. Got a time for the first time
+        // Got a time for the first time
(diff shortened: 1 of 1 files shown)


fifieldt:
Actually, we won't set the time if it's a bogus time. See the epoch logic

Xaositek:
I like having redundancy - sometimes the epoch bits only fire from the hosted builds, last I heard they didn't work if you built from scratch so this is a nice option to have.

jp-bennett:
Not 100% sold on this approach, but seemed worth a PR to discuss.

fifieldt:
> I like having redundancy - sometimes the epoch bits only fire from the hosted builds, last I heard they didn't work if you built from scratch so this is a nice option to have.

This was fixed too, actually. They'll always be set.

fifieldt:
> Not 100% sold on this approach, but seemed worth a PR to discuss.

In bad signal scenarios you can sometimes get a time, say, 15mins before you can get a position, so this is not without user impact :)

What bogus times are we protecting from here?

jp-bennett:
Going to close this, as this scenario should be handled. Will re-open if I run into this problem again.

PR #8967, opened 2025-12-14
Add LilyGO T-Beam 1W support

santosvivos:
## Summary
Adds support for the LilyGO T-Beam 1W, an ESP32-S3 based LoRa device with a 1-watt external power amplifier.

## Hardware
- ESP32-S3 (16MB Flash, 8MB OPI PSRAM)
- SX1262 + 1W PA (XY16P35 module)  
- SH1106 OLED 128x64
- Quectel L76K GPS
- Cooling fan for PA

## Changes
- Add board definition and variant files for ESP32-S3 based T-Beam 1W
- Add RF95_FAN_EN support to SX126xInterface for PA cooling fan
- Add SX126X_PA_RAMP_US for configurable PA ramp time (800us for 1W PA)
- Configure RF switch: DIO2 for PA, GPIO 21 for LNA control

### New Files
- `boards/t-beam-1w.json` - Board definition
- `variants/esp32s3/t-beam-1w/` - Variant files

### Modified Files  
- `src/mesh/SX126xInterface.cpp` - Added RF95_FAN_EN and SX126X_PA_RAMP_US support (wrapped in #ifdef, no impact on other boards)

## Key Configuration
- RF switch: DIO2 controls PA, GPIO 21 (SX126X_RXEN) controls LNA
- PA ramp time: 800µs (required for 1W PA stabilization)
- TCXO voltage: 1.8V

## 🤝 Attestations
- [x] I have tested on: LilyGO T-Beam 1W (the new hardware this PR adds)
------- Note: Changes to `SX126xInterface.cpp` are guarded by `#ifdef` and only activate when the relevant macros are defined in a variant, so there is no regression risk for existing boards. -----

### DIFF
--- a/src/mesh/SX126xInterface.cpp
+++ b/src/mesh/SX126xInterface.cpp
@@ -63,8 +63,8 @@ template <typename T> bool SX126xInterface<T>::init()
 #endif
 
 #ifdef RF95_FAN_EN
-    pinMode(RF95_FAN_EN, OUTPUT);
     digitalWrite(RF95_FAN_EN, HIGH);
+    pinMode(RF95_FAN_EN, OUTPUT);
 #endif
 
 #if ARCH_PORTDUINO


fifieldt:
Very exciting!

To stay within regulatory power limits in all jurisdictions, we need a bit of information about the performance of the power amp. Do you happen to have a table of the expected output based on certain output power levels from the SX1262?

We use those values to calculate how much gain the PA provides for the various SX1262 power levels, see eg:
https://github.com/meshtastic/firmware/blob/bf32f17f287195603bb6926faf3b219fd726c3ed/src/configuration.h#L146

santosvivos:
> Very exciting!
> 
> To stay within regulatory power limits in all jurisdictions, we need a bit of information about the performance of the power amp. Do you happen to have a table of the expected output based on certain output power levels from the SX1262?
> 
> We use those values to calculate how much gain the PA provides for the various SX1262 power levels, see eg:
> 
> https://github.com/meshtastic/firmware/blob/bf32f17f287195603bb6926faf3b219fd726c3ed/src/configuration.h#L146

Greetings! I believe I found some tables specifically for this device posted by the manufacturer at this link: https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/docs/en/t_beam_1w/t_beam_1w_hw.md


currnetA
currnetB

fifieldt:
Thanks, that looks like the graph for max power (SX1262 at 22dBm + ~10dBm gain at the power amp = 1W or so)

Nothing for the lower power ranges?
Is there a part number for the PA on diagrams somewhere so we can look up at component manufacturer instead perhaps?

santosvivos:
> Thanks, that looks like the graph for max power (SX1262 at 22dBm + ~10dBm gain at the power amp = 1W or so)
> 
> Nothing for the lower power ranges?
> Is there a part number for the PA on diagrams somewhere so we can look up at component manufacturer instead perhaps?


I was not able to find any information for low power ranges in the manufacturer's documentation. I disassembled one unit and was able to get some numbers from the RF module, but I was not able to find any references to it online. I have emailed the manufacturer of the device to ask them if they can provide the module manufacturer or additional information regarding the power amplifier / low power output. I will provide an update when they respond : )

fifieldt:
@santosvivos , while we wait, maybe just add TX_GAIN_LORA 10 ? it won't be perfect, but I don't want to block this important work :)

santosvivos:
No problem I will have that done tonight : )

jstockdale:
Thanks so much for getting this working!! Already using it on my device. :tada: :raised_hands:

santosvivos:
@fifieldt Is there anything else I could do to help? Not really sure how this all works :sweat_smile:

jstockdale:
For what it’s worth, everything seems to be working here with my unit (current with your branch as of last night) … except the battery indicator is not properly calibrated for the ~8.2V F550 battery.

It shows 100% at 8.2V and at 7.6V and at 6.95V so it doesn’t seem to be tracking the capacity.

Maybe it is doing battery math for a ~4.1V battery still?

fifieldt:
@santosvivos , it looks great ... have you tried asking the #alphanauts channel to see if anyone in there has one they can test on?

santosvivos:
@jstockdale Oh okay cool I can take a look at that and maybe throw it on a fixed voltage DC supply then adjust the calculations if that's what's happening. Thanks for the bug!

@fifieldt I have not found anyone there yet, but I believe there is at least one in the Meshtastic Discord with it running on this fork. I have 2 of these devices personally running on this latest update with performance similar or better than my other devices(T-beam, T1000-e, etc). I plan on confirming the voltage calculations this week per @jstockdale bug find.

ethan1980:
I have my T-beam 1W running with this version and it seems to be working correctly. I haven't used a battery, so I can't confirm the capacity calculation issue. I'm using 22dB in the LoRa configuration and I'm getting similar results to other nodes with similar capacities, like an e22.
camphoto_1804928587

shalberd:
Is the relationship between all those variables documented somewhere? Mostly, I am curious about whether the 27 dBm / 0.5W maximum on EU_868 holds here. Asking out of interest in all sincerity.

Not583:
Fan keeps running? Is there a way to switch it off?

santosvivos:
@shalberd I have only as much info as provided by the manufacturer. You will need to attach a spectrum analyzer and run through each setting in your configuration to know for certain.

 @Not583 I set the fan to on all of the time for safety. I could not find information on any integrated temperature sensor for the PA. It will require deeper investigation. Alternatively you can unplug the PH2.0 connector, which powers the fan, to shut it off, but I personally wouldn't do that if you are doing any TX. There is no way to know thermal characteristics without access to a sensor or with external testing equipment. 

I have been very busy, but I received two additional units for testing, and intend to take another bite out of this soon!

jstockdale:
@santosvivos I took a look at the schematic and the general recommendations by LilyGo at https://github.com/Xinyuan-LilyGO/LilyGo-LoRa-Series/blob/master/docs/en/t_beam_1w/t_beam_1w_hw.md and from what I'm gathering the PA gets switched on when transmitting, and is off (with the LNA being on) during receive. I'd be happy to test, but I would imagine we can use airtime as a proxy for how much of the time the PA is on. I don't think the LNA should generate *nearly* as much heat as the PA. So the passive heatsink is likely more than sufficient for the LNA in receive mode.

In general we might be able to speed control the fan by setting up pin 41 as a pwm. But that would require testing to confirm. I am pretty sure we could try something like turn on the fan low whenever transmitting and then scale the speed based on airtime. But even if pwm / speed control doesn't work with this fan, just turning the fan on during tx and then off a short time (~250ms?) later would probably work as a simple solution. That won't account for environmental variation but I think any attempt at improving the fan code will significantly improve battery life of the device. And like. It's easy to be pretty conservative and run the fan more than necessary but still not 100% all the time.

I'm also been thinking setting up some way to quickly toggle transmit between like 100mw, 500mw, and 1000mw would be pretty cool. Or dynamic transmission power based on retries and acks. But now I'm just thinking out loud.

Thanks for your excellent work on this device! Let me know if I can be helpful with support for anything in particular (or if you want me to take a crack at a fan implementation / curves / testing). Cheers!!

Not583:
Blackout Comms got the fan issue sorted....... In their review they say this: 

"The LNA (low noise amplifier) needs to be disabled any time you’re about to TX, or the radio can be permanently damaged. It needs to be re-enabled for receive mode. Blackout Comms switches between TX/RX mode often multiple times per second. 

There is some other low-level timing controls that need to be put in place. I probably need to tune the timings a little better, but they are working pretty well at this point for the T-Beam 1W.

Additionally, the fan (other models don’t have) needs to be controlled, so I need to read the T-Beam’s temp every so often and turn the fan on if it gets too hot."

And it works!

teffcz:
In datasheet (page 3) i can see there is a NTC sensor (NCP18XH103F03RB) on pin ESP IO14. Maybe we could use this for fan control

PR #8970, opened 2025-12-14
Add Rebooting to DFU mode notification as a simple pop-up

Xaositek:
- Add Rebooting to DFU mode notification as a simple pop-up

IMG_1539

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -417,7 +417,9 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
     }
     case meshtastic_AdminMessage_enter_dfu_mode_request_tag: {
         LOG_INFO("Client requesting to enter DFU mode");
+#if HAS_SCREEN
         IF_SCREEN(screen->showSimpleBanner("Device is rebooting\ninto DFU mode.", 0));
+#endif
 #if defined(ARCH_NRF52) || defined(ARCH_RP2040)
         enterDfuMode();
 #endif


src/modules/AdminModule.cpp
```diff
@@ -417,6 +417,7 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
     }
     case meshtastic_AdminMessage_enter_dfu_mode_request_tag: {
         LOG_INFO("Client requesting to enter DFU mode");
+        screen->showSimpleBanner("Device is rebooting\ninto DFU mode.", 0);
```
thebentern: Let's add an if (screen) statement and #if HAS_SCREEN? Perhaps this could crash on screenless devices?

Xaositek: Added now

thebentern (approved):

src/modules/AdminModule.cpp
```diff
@@ -417,6 +417,9 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
     }
     case meshtastic_AdminMessage_enter_dfu_mode_request_tag: {
         LOG_INFO("Client requesting to enter DFU mode");
+#if HAS_SCREEN
+        IF_SCREEN(screen->showSimpleBanner("Device is rebooting\ninto DFU mode.", 0));
+#endif
```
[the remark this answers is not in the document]
Xaositek: @thebentern thoughts? It seems double-safe?

PR #8975, opened 2025-12-15
InkHUD Menu improvements

HarukiToreda:
### Summary

This PR introduces a major update to the InkHUD menu that allows most node configuration to be done directly on the device without needing a phone.

A new section called **Node Config** was added to group all node related settings in one place.

### What is included

**LoRa Config**
- Shows current region, modem preset, and frequency
- Allows changing region and preset from the device
- Uses the same reload and reboot behavior as existing config paths

**Channel Config**
- Lists active channels by name
- Per channel controls for uplink, downlink, and position sharing
- Position precision can be selected per channel
- Precision labels follow the selected metric or imperial display units

**Device Config**
- Shows current device role
- Allows switching between supported roles
- Timezone selection added

**Power Config**
- Toggle powersave mode (ESP32 only)

**Display Config**
- Toggle 12 hour clock mode
- Toggle display units between metric and imperial

**Bluetooth Config**
- Toggle Bluetooth on or off
- Switch pairing mode between fixed and random PIN
- If toggled while Wi-Fi is enabled, it will disable Wi-Fi and enable Bluetooth for esp32

**Network Config**
- Toggle Wi-Fi on or off
- Visible only by esp32
- Shows connection status
- When toggled, it disables Bluetooth

**Reset NodeDB**
- Options to delete All or Keep only Favorites

### UI improvements

- This also adds the option to select a region on setup tips when region is UNSET
- Added a Back option consistently across all menus
- Reduced line spacing to fit more options without scrolling
- No changes outside InkHUD
- Next tile function now unselects so screen can close

### Video Demo

https://github.com/user-attachments/assets/bc69f3c2-2ba9-4021-b207-fafe94ec4ad5

### DIFF
--- a/bin/device-install.sh
+++ b/bin/device-install.sh
@@ -32,6 +32,19 @@ if ! command -v jq >/dev/null 2>&1; then
     exit 1
 fi
 
+# esptool v5 supports commands with dashes and deprecates commands with
+# underscores. Prior versions only support commands with underscores
+if ${ESPTOOL_CMD} | grep --quiet write-flash
+then
+    ESPTOOL_WRITE_FLASH=write-flash
+    ESPTOOL_ERASE_FLASH=erase-flash
+    ESPTOOL_READ_FLASH_STATUS=read-flash-status
+else
+    ESPTOOL_WRITE_FLASH=write_flash
+    ESPTOOL_ERASE_FLASH=erase_flash
+    ESPTOOL_READ_FLASH_STATUS=read_flash_status
+fi
+
 set -e
 
 # Usage info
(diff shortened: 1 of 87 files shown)


HarukiToreda:
Now that I think about it, there's still some other config options I can still include, putting it on draft

Xaositek:
Some initial feedback:

- Super snappy on ThinkNode M1 - fastest EInk I've ever used
- Backlight toggle on back button isn't consistent
- Shutdown tone plays twice about 7-10 seconds apart

HarukiToreda:
Tested for a few weeks, ready for review

PR #8985, opened 2025-12-16
Implement Long Turbo preset

### LINKED ISSUE #8214: [Feature Request]: LongModTurbo Preset
### Platform

Cross-Platform

### Description

New LongModTurbo Radio Preset:
-500kHz Bandwidth
- 11 Spread Factor
- 8 Coding Rate
- 902.6875 (or close to 902 as possible)

As various meshes are moving to different presets do distance themselves from saturated default LongFast traffic the 2 logical directions are MediumSlow (faster but lower SF) and LongModerate (more reliable but slower and increased airtime). 

I’m proposing that those regions with the option have a new preset to choose from to leverage 500kHz bandwidth. I’m nicknaming it LongModTurbo because it’s essentially Long Range / Moderate settings with the bandwidth cranked up but call it whatever 🤣

Since 500kHz is only limited to certain regions, make the preset only available depending on set region. 

While people can override presets, this makes it a lot simpler for less savvy folk to change mesh radio settings and join a new mesh. The setting proposed offer the following benefits:

- 33% increase in speed over LongFast reducing airtime.
- Same spread factor of LongFast/LongModerate for larger geographically spread-out meshes. 
- Higher Coding Rate for longer/noisier links 
- this speed has the option to be a 100% (double) increase if variable CR is introduced letting people start with higher reliability and increasing speed if needed. 
- A frequency that is far from the noisy 915mHz centre, but close enough to 906.875 that antennas tuned for LongFast won’t suffer a big hit. 
- From limited testing, a side benefit seems to be reduced hops, as the more resilient paths means smart flooding isn’t forced to take less efficient routes 
- A way to move from LongFast without making any sacrifices (with the exception of a (slightly reduce Link Budget)

thebentern:
Closes #8214

### DIFF
--- a/src/graphics/draw/MenuHandler.cpp
+++ b/src/graphics/draw/MenuHandler.cpp
@@ -33,26 +33,23 @@ namespace graphics
 namespace
 {
 
+// Caller must ensure the provided options array outlives the banner callback.
 template <typename T, size_t N, typename Callback>
-BannerOverlayOptions createBannerOptions(const char *message, const MenuOption<T> (&options)[N],
-                                         std::array<const char *, N> &labels, Callback &&onSelection)
+BannerOverlayOptions createStaticBannerOptions(const char *message, const MenuOption<T> (&options)[N],
+                                               std::array<const char *, N> &labels, Callback &&onSelection)
 {
     for (size_t i = 0; i < N; ++i) {
         labels[i] = options[i].label;
     }
 
-    std::array<MenuOption<T>, N> optionsCopy{};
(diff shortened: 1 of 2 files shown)


Xaositek (approved):
Only thing I notice is that the ordering of Presets in the BaseUI menu isn't consistent. Should probably be:
"LongModerate", "LongFast", "LongTurbo", 
"MediumSlow", "MediumFast", 
"ShortSlow", "ShortFast", "ShortTurbo",

fifieldt (approved):

PR #8992, opened 2025-12-16
Cut NRF52 bluetooth power usage

phaseloop:
Resubmitting https://github.com/meshtastic/firmware/pull/8858 after issues with iPhones were fixed.

300% Bluetooth subsystem power saving after measurements seem to save around 3% overall power node usage (not counting power drawn on Lora tx).

Tested by multiple users on their NRF52 nodes and with their phones (iPhone, Android, etc). Personally testes on Promicro/Faketec board.

### DIFF
--- a/src/platform/nrf52/NRF52Bluetooth.cpp
+++ b/src/platform/nrf52/NRF52Bluetooth.cpp
@@ -243,10 +243,8 @@ int NRF52Bluetooth::getRssi()
 
 // Valid BLE TX power levels as per nRF52840 Product Specification are: "-20 to +8 dBm TX power, configurable in 4 dB steps".
 // See https://docs.nordicsemi.com/bundle/ps_nrf52840/page/keyfeatures_html5.html
-#define VALID_BLE_TX_POWER(x) \
-    ((x) == -20 || (x) == -16 || (x) == -12 || \
-     (x) == -8  || (x) == -4  || (x) == 0  || \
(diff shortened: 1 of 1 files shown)


fifieldt (approved):

phaseloop:
fyi: I ran trunk formatter on the code but CI/CD still shows an error without any information why does it seem formatting is wrong :/

I can run clang-format but it reformats whole file and there are thousands of changes.

phaseloop:
@thebentern - is something else here blocking merge? thanks :)

src/platform/nrf52/main-nrf52.cpp
```diff
@@ -313,6 +313,23 @@ void nrf52Setup()
 
     r = nrfx_wdt_channel_alloc(&nrfx_wdt, &nrfx_wdt_channel_id_nrf52_main);
     assert(r == NRFX_SUCCESS);
+
+
+    // print LFCLK debug info
```
thebentern: Can we uncomment this but guard it in a #if DEBUG_LFCLK?

phaseloop: This was lefover from debugging iPhone issues - I removed it as there is little of value to print that information.

phaseloop:
everything should be fine now

thebentern:
@phaseloop I plan on merging this one after the 2.7.18 alpha

PR #8996, opened 2025-12-16
Adding support for Adafruit Soil Moisture Sensor (ADA4026)

rpnahm:
I added support for the Adafruit Soil Moisture: https://www.adafruit.com/product/4026.
image

The default I2C address (0x36) overlaps with a couple other supported sensors. This implementation utilizes the address 0x39. You can change the address by soldering closed the two jumpers on the back of the sensor. This implementation also doesn't support the temperature readings that the soil moisture sensor can provide. 

There are some extra commits to facilitate the custom hardware config we were developing in tandem. The final changes are up to spec though.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631
  -   - Custom node using nrf52840-dk and SX1262MB2CAS  -


src/mesh/generated/meshtastic/telemetry.pb.h
```diff
@@ -102,8 +102,10 @@ typedef enum _meshtastic_TelemetrySensorType {
     meshtastic_TelemetrySensorType_SEN5X = 43,
     /* TSL2561 light sensor */
     meshtastic_TelemetrySensorType_TSL2561 = 44,
+    /* Adafruit STEMMA Soil Sensor */
```
fifieldt, line 105: Thanks for your work! Just so you know, these files are automatically generated and so the edits should go to the meshtastic/protobufs repo

PR #8999, opened 2025-12-17
Prep work for better Store and Forward

jp-bennett:
Make getHash() a public accessor, and make the p_encrypted packet copy accessible for modules to use.


shalberd:
quick question: Does device role ROUTER_LATE also support Store & Forward? Great to see the feature is being worked on.

PR #9009, opened 2025-12-17
update docker configuration

lpmi-13:
# Description

This commit updates the docker configuration for the project in the following ways:

- moves the ARG lower down in the 2 Dockerfiles to take advantage of layer caching if the image is rebuilt with a different ARG value (the layers above it don't need to be rebuilt).
- removes unnecessary `mkdir` statements, since `WORKDIR` runs `mkdir` under the hood (https://docs.docker.com/reference/dockerfile/#workdir)
- binds the compose stack to the loopback interface instead of all interfaces, which means if running on an unsecured subnet (eg, coffee shop wifi) no other hosts will be able to send traffic to the container. 

## 🤝 Attestations

- [x] My proposed changes don't update any code and are only updates to the container configuration

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
 
     # Forward the container’s port 4403 to the host
     ports:
-      - 127.0.0.1:4403:4403
+      - "${NETWORK_INTERFACE}:4403:4403"
 
     restart: unless-stopped
 
--- a/.env.example
+++ b/.env.example
@@ -2,3 +2,5 @@
 CONFIG_PATH=/path/to/meshtastic/config.yaml
 # USB device to passthrough (`lsusb -t`: look for `ch341`)
 USB_DEVICE=/dev/bus/usb/001/037
+# Default loopback binding for the docker-compose stack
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
 
     # Forward the container’s port 4403 to the host
     ports:
-      - 4403:4403
+      - 127.0.0.1:4403:4403
```
vidplace7: Can you actually break this out as a variable (Add in `.env.example`) and **default** to 127.0.0.1?

Beyond that nitpick this PR looks good ❤️ thanks for the submission!

lpmi-13: Yep, can do 🫡 

I _was_ actually wondering, given the nature of this project, if there are some times you'd like it to run on one host and be reachable from another (or maybe there's some other reason I'm missing about why you'd want it bound to `0.0.0.0/0`...CI maybe?).

lpmi-13:
I also added the `.env` file to the `.gitignore`, but if it's intentional that it's _not_ in there, I can easily take it out and add a comment about why we don't want it.

PR #9014, opened 2025-12-18
Be more judicious about responding to want_response in existing meshes

thebentern:
tl;dr two major changes in behavior:
~~- Don't send a want_response in the initial send unless we have less than 10 nodes in the nodedb~~
- Don't reply to want_response NodeInfos if we've just heard a nodeinfo from the node in the past 12 hours

### DIFF
--- a/src/modules/NodeInfoModule.cpp
+++ b/src/modules/NodeInfoModule.cpp
@@ -9,9 +9,13 @@
 #include <Throttle.h>
 #include <algorithm>
 
+#ifndef USERPREFS_NODEINFO_REPLY_SUPPRESS_SECS
+#define USERPREFS_NODEINFO_REPLY_SUPPRESS_SECS (12 * 60 * 60)
+#endif
+
 NodeInfoModule *nodeInfoModule;
 
-static constexpr uint32_t NodeInfoReplySuppressSeconds = 12 * 60 * 60; // 12 hours
+static constexpr uint32_t NodeInfoReplySuppressSeconds = USERPREFS_NODEINFO_REPLY_SUPPRESS_SECS;
 
 bool NodeInfoModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshtastic_User *pptr)
 {
@@ -93,11 +97,6 @@ void NodeInfoModule::sendOurNodeInfo(NodeNum dest, bool wantReplies, uint8_t cha
         bool requestWantResponse = (config.device.role != meshtastic_Config_DeviceConfig_Role_TRACKER &&
                                     config.device.role != meshtastic_Config_DeviceConfig_Role_SENSOR) &&
                                    wantReplies;
-
-        if (requestWantResponse && lastSentToMesh == 0 && nodeDB && nodeDB->numMeshNodes >= 10) {
-            LOG_DEBUG("Stripping want_response because we have 10 or nodes in our NodeDB");
-            requestWantResponse = false;
-        }
 
         p->decoded.want_response = requestWantResponse;
         if (_shorterTimeout)
--- a/userPrefs.jsonc
+++ b/userPrefs.jsonc
@@ -55,6 +55,7 @@
   // "USERPREFS_MQTT_TLS_ENABLED": "false",
   // "USERPREFS_MQTT_ROOT_TOPIC": "event/REPLACEME",
   // "USERPREFS_RINGTONE_NAG_SECS": "60",
+  // "USERPREFS_NODEINFO_REPLY_SUPPRESS_SECS": "43200",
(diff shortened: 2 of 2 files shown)


RCGV1:
@thebentern I approve but also guard the want response so routers can send want response true

src/modules/NodeInfoModule.cpp
NomDeTom, line 1: Can we have some userprefs overrides for these?

src/modules/NodeInfoModule.cpp
```diff
@@ -68,9 +90,16 @@ void NodeInfoModule::sendOurNodeInfo(NodeNum dest, bool wantReplies, uint8_t cha
 
     if (p) { // Check whether we didn't ignore it
         p->to = dest;
-        p->decoded.want_response = (config.device.role != meshtastic_Config_DeviceConfig_Role_TRACKER &&
+        bool requestWantResponse = (config.device.role != meshtastic_Config_DeviceConfig_Role_TRACKER &&
                                     config.device.role != meshtastic_Config_DeviceConfig_Role_SENSOR) &&
                                    wantReplies;
+
+        if (requestWantResponse && lastSentToMesh == 0 && nodeDB && nodeDB->numMeshNodes >= 10) {
```
NomDeTom: Can we have a userprefs override for this?

src/modules/NodeInfoModule.cpp
```diff
@@ -7,17 +7,37 @@
...
 
     auto p = *pptr;
+
+    if (mp.decoded.want_response) {
+        const NodeNum sender = getFrom(&mp);
+        const uint32_t now = mp.rx_time ? mp.rx_time : getTime();
+        auto it = lastNodeInfoSeen.find(sender);
+        if (it != lastNodeInfoSeen.end()) {
+            uint32_t sinceLast = now >= it->second ? now - it->second : 0;
+            if (sinceLast < NodeInfoReplySuppressSeconds) {
```
NomDeTom, line 37: Can we have a userprefs override for this?

fifieldt:
Suggested modification - may be hard to implement since it involves time.

rather than just the raw number of nodes >=10, make it the  number of nodes we've seen within X days?

If a device has been off for some time, the mesh has likely changed significantly.

fifieldt:
Also, should this be targeted at master or develop?

RCGV1:
I don't think it should be seen online since what if we are boot looping and not able to receive packets from nodes. Also I don't like solutions that involve being able to receive well since there can be bad connections to the mesh.

thebentern:
After reviewing the actual behavior, the original send premise was incorrect. We only send want_response on the initial announce after a factory reset, so this change did nothing from that end. Any witnessed behavior on that end currently is most likely very outdated firmware. The response side however still holds more value, so we can ship that.

rob-fi:
Sorry to jump in on this, but if anyone reads this that is able to provide some clarification then I'd appreciate it. Is it the case that want_response to nodeinfos are being stripped even with a manual request? If so then it might explain the problems I've been having with others during testing when nodes have had a Node DB reset. It's been causing quite a headache - nodes with good consistent connectivity through a hop or two just never seemed to exchange node info - things just seemed broken. That then resulted in more resets and so on.

If this is documented somewhere other than this PR then could someone link me to it so I can let others know? Seems a pretty major thing to change. I'd rather a node temporarily put another in a naughty list for sending too many nodeinfos with want_response in a short space of time to the same node, rather than blanket 12 hours. Just gives the impression that the mesh is broken.

If I'm misunderstanding this change then I'd be delighted to be corrected also!

There really should be an easy way to batch export and import nodes via the Android app to share with others rather than screenshots of the QR codes of various nodes so that others can add them and get the public keys and names again. Of course favouriting before resetting is an option if you remember.

NomDeTom:
> Sorry to jump in on this, but if anyone reads this that is able to provide some clarification then I'd appreciate it. Is it the case that want_response to nodeinfos are being stripped even with a manual request? If so then it might explain the problems I've been having with others during testing when nodes have had a Node DB reset. It's been causing quite a headache - nodes with good consistent connectivity through a hop or two just never seemed to exchange node info - things just seemed broken. That then resulted in more resets and so on.

This only affects packets that are sent broadcast, not with a specific destination.
 
> If this is documented somewhere other than this PR then could someone link me to it so I can let others know? Seems a pretty major thing to change. I'd rather a node temporarily put another in a naughty list for sending too many nodeinfos with want_response in a short space of time to the same node, rather than blanket 12 hours. Just gives the impression that the mesh is broken.

The 12 hour naughty list is specifically to address older node firmware that sends broadcast nodeinfo with a want_ack. Newer firmware only does this on a factory reset. If you think it needs documenting somewhere specific then suggest it. The cure to the specific problem is to keep a factory reset node off a busy mesh for 12 hours before a reset.

> If I'm misunderstanding this change then I'd be delighted to be corrected also!
> 
> There really should be an easy way to batch export and import nodes via the Android app to share with others rather than screenshots of the QR codes of various nodes so that others can add them and get the public keys and names again. Of course favouriting before resetting is an option if you remember.

Raise a feature request in the relevant repo, please.

PR #9023, opened 2025-12-20
First position send validation

thebentern:
For our first positionmodule send after boot, validate if we have received a fresh position since boot from GPS or phone, if not do not send yet. We have one exception to this, if the node is set to fixed_position. This should prevent unnecessary and stale positions from being sent to the mesh on first boot.


thebentern:
Tested and confirmed working

PR #9031, opened 2025-12-21
action: skip trying to comment binary size change results if it is not a PR

Jorropo:


.github/workflows/main_matrix.yml
vidplace7, line 237:
```suggestion
    if: github.repository == 'meshtastic/firmware' && github.event_name == 'pull_request_target'
```

vidplace7: You should just be able to add this here

vidplace7: Also we use `pull_request_target` not `pull_request` :+1:

Jorropo: We need `Upload combined manifests for later commit and global stats crunching.` to run on master & develop.
I am thinking maybe there should be two steps.

But given both needs to download all the manifests to merge them I thought it made sense to have it together as one.

Jorropo: I changed `pull_request_target`, opened #9033 since I forgot to make a branch sorry

PR #9032, opened 2025-12-21
Shame do not complain about missing targets

Jorropo:
PR CI only runs a small subset of all tests.

It is very likely a file we didn't found in the PR is just not tested in PR.


Jorropo:
I hope this fixes

Issue #7481, opened 2025-07-27
[Feature Request]: 3.x have narrow bandwidth and low spreading factor presets


mikecarper:
### Platform

Cross-Platform

### Description

Using a narrow bandwidth like 62.5 and lower allows for a lower noise floor to be found. Add in some presets that support 62.5khz SF7-9; and then for sx 1262 chips support 8khz SF5 up to 41.7khz SF5-8. Basically limit the lowest Kbps to be bigger than 0.8kbps with the new presets.

So presets would be narrow fast, narrow medium and narrow slow for 62.5 sf7,8,9

This could be a 2.8 change in theory as well but I figured that a new classification of presets might need to wait until 3.x

mikecarper:
<img width="1844" height="1571" alt="Image" src="https://github.com/user-attachments/assets/ad7fe3c3-3e6a-4459-b8e8-d8e8fe17ea45" />

NomDeTom:
@fifieldt can you reopen this, please?

NomDeTom:
The discussions on the discord indicated that there may be some fine-tuning required to try and make sure that the slots land in a good place between the peaks. I'm not sure that is extensible across the whole of the US - electricity meters may be standardised, but whether they all pick the same frequency sequence is unclear.

I do think some narrow presets are needed, and the EU868 band is so small that there is no wiggle room anyway. The waterfall plots I've seen where it would be most beneficial, there are high noise sources in the very centre of the 869.525 band, so sitting to one or other side would be best.

I support a new narrow preset, and I'd suggest that we tweak it such that Meshtastic lands in slot 1, i.e. the lower edge of the band, simply to avoid conflict with other lora systems. _Edit: this is in reference to EU868. US915 still needs work._

fifieldt:
Lower edge of the band often has overflow interference from high power telco freqs as I understand it ...

mikecarper:
<img width="1653" height="753" alt="Image" src="https://github.com/user-attachments/assets/3a6ebb2b-324f-454b-9a83-44f53b35a700" />

Top and bottom of the spectrum don't look too bad. If you want to only support sx1262 chips getting more narrower than 62.5 would have even more benefits for finding a quiet place. Looking at an SDR in Boston and its similar to what we have here in Seattle

NomDeTom:
[...]

PR #9038, opened 2025-12-21
New_EU_presets

NomDeTom:
Some new presets for a wider range of operations in the EU. Ham bands are now removed, pending further discussion.
Relies on https://github.com/meshtastic/protobufs/pull/824 for protobufs.

## What has changed?
- 2 new region presets
  - EU_866, 
  - NARROW_868,
- 4 new modem presets
  - LITE_FAST/SLOW, 
  - NARROW_FAST/SLOW
- new region parameters
  - licensedOnly, to limit a region to ham operators, i.e. those operating with the licensed flag active - placeholder for now.
  - overrideSlots, to allow a region to always default to a particular slot
  - defaultPreset, to allow a default other than LongFast
  - availablePreset, to allow limiting preset availability to a particular region
- refactoring of the modem setting checks to happen before restarting the node
- adding additional client notifications for invalid setting combinations

## What is needed to integrate these changes more widely
- BaseUI / InkHud
  - The new regions and presets have been added to the menu lists, but the list logic has not been updated to exclude invalid combinations
- MUI
  - The new regions and presets need to be added to the menu lists, and the list logic included.
- Client Apps
  - These will need updating to take account of the valid region/preset combinations, ideally hiding invalid combinations within the menus
  - Their display logic will need updating to include the new overrideSlots information, which will affect both the lora config menu and the behaviour when new primary channels are generated

## Acknowledgements
This change has been supported extensively by @Stary2001, @phaseloop, and @caveman99. Their input cannot be understated.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Pro-micro DIY


src/mesh/RadioInterface.cpp
```diff
@@ -187,6 +187,24 @@ const RegionInfo regions[] = {
     */
     RDEF(BR_902, 902.0f, 907.5f, 100, 0, 30, true, false, false),
 
+    /*
+        EU 866MHz RFID band (ETSI EN 302 208): 4 channels at 865.7/866.3/866.9/867.5 MHz
+        475 kHz gap between channels, 27 dBm, duty cycle 2.5% (mobile) or 10% (fixed)
+        https://www.etsi.org/deliver/etsi_en/302200_302299/302208/03.04.01_60/en_302208v030401p.pdf
+    */
+    RDEF(EU_866, 865.6375f, 867.5625f, 2.5, 0.475, 27, true, false, false),
```
phaseloop: I still have doubts about this as ETSI states that this band is open only for RFID tags and readers



> The present document applies to RFID interrogators and tags operating together as a system. 

>For each specified band,
>multiple high power channels are made available for use by interrogators. The tags respond with a modulated >signal
> preferably in the adjacent low power channels. Interrogators may be used with either integral or external antennas.
The types of equipment covered by the present document are as follows:
• fixed interrogators;
• portable interrogators;
• batteryless tags;
• battery assisted tags;
• battery powered tags.
These types of radio equipment are capable of operating in the frequency ranges given in table 1 and table 2.

src/mesh/RadioInterface.cpp
```diff
@@ -518,6 +553,31 @@ void RadioInterface::applyModemConfig()
...
+                bw = 62.5;
+                cr = 6;
+                sf = 7;
+                break;
+            case meshtastic_Config_LoRaConfig_ModemPreset_NARROW_SLOW:
+                bw = 62.5;
+                cr = 6;
+                sf = 8;
+                break;
+            case meshtastic_Config_LoRaConfig_ModemPreset_HAM_FAST:
```
phaseloop: If we use "narrow" here then I suppose instead of "HAM" it should be called "WIDE". AFAIK such wide setting could also be used by non-hams in 915 MHz band (US and EU coming soon) as this band allows it.

NomDeTom: The intention here is to give the HAM_US433 a specific set of presets for use in that band, which consists of the two narrow settings (already in use) plus an equivalent of shortfast for use at meetups. I should emphasise that this particular setup has been handed to me, so I'm not familiar with the details.

src/mesh/RadioInterface.cpp
NomDeTom: I've redone the definition. It now refers to 47b: SRD.

src/mesh/RadioInterface.cpp
phaseloop: I understand original intention - in that case we will probably end up later with same setting duplicated later under different name as "wide/fast" will be usable on 915/2.4 GHz too.

NomDeTom: @wehooper4 what was the intended setting for the "wide" ham preset?

src/mesh/MeshRadio.h
```diff
@@ -5,6 +5,32 @@
...
+    bool allowPreset_SHORT_TURBO; // 500kHz BW
+    bool allowPreset_LONG_TURBO;  // 500kHz BW
+    bool allowPreset_LITE_FAST;   // For EU_866
+    bool allowPreset_LITE_SLOW;   // For EU_866
+    bool allowPreset_NARROW_FAST; // Narrow BW
+    bool allowPreset_NARROW_SLOW; // Narrow BW
+    bool allowPreset_HAM_FAST;    // 500kHz BW
+    bool reserved;
+};
+
```
phaseloop: I think we don't need this struct at all. In `config.pb.h` available modes are already listed as enum/int.

So preset can actually be a simple array of modes allowed in a given band:

```
PRESETS_EU_868 = [meshtastic_Config_LoRaConfig_ModemPreset_NARROW_FAST,  meshtastic_Config_LoRaConfig_ModemPreset_LITE_FAST]
```
(I know presets don't match the band but this is an example).

And this array is then passed to RDEF. And we now have a ready list of available modes to show to user when they select a band, no need even to do any filtering.

NomDeTom: Thanks - this is the kind of insight I needed.

I'll make the changes later.

Do you have any view on how to iterate the presets list for the menu display?

src/mesh/RadioInterface.cpp
```diff
@@ -279,10 +267,15 @@ void initRegion()
...
 }
 
-const RegionInfo *getRegion(meshtastic_Config_LoRaConfig_RegionCode code) {
+const RegionInfo *getRegion(meshtastic_Config_LoRaConfig_RegionCode code)
+{
     const RegionInfo *r = regions;
     for (; r->code != meshtastic_Config_LoRaConfig_RegionCode_UNSET && r->code != code; r++)
         ;
+    if (r->licensedOnly == true && !devicestate.owner.is_licensed) {
+        LOG_WARN("Region code %d not permitted without license, defaulting to unset", config.lora.region);
```
phaseloop: This should be checked in validateModemConfig and return to default band preset as we do for any other problem with presets. Otherwise someone will end with bricked device with no lora if they try to set it remotely

NomDeTom: OK - understood. I'll move it to ln 627, but I don't know how to exit that check cleanly.

NomDeTom:
@jp-bennett This PR now includes your custom CR check. Can you please re-check it in the context of this change? I don't know if we need additional logic to handle if moving to a preset with a higher default preset than the custom CR set, or indeed if we need to take account of people wanting to run CR4.

In any case, can you help me avoid a reversion of your change?

Jckf:
As requested on Discord:
Untitled
A visualization of the presets and their slots, plus the band used by a different well-known system (Alt 1).

rob-fi:
As per comments in https://github.com/meshtastic/firmware/pull/8855 an that have been happening elsewhere, I'd like to further discussions on this somehow. Happy for someone to let me know what the best channel for that is.

As I understand it, the general concerns are:

- Compatibility with existing widely used settings e.g. in Finland that use 62.5 kHz bandwith starting from the band edge. Even if a different CR was used in an official new preset, it's understood that it at least would allow interoperability.
- Purpose of the channel spacing considering that other presets (or indeed other LoRa or other devices using the band) do not use any, and there's no apparent technical or legal requirement (that has been presented) to do so. Furthermore, as per the Meshtastic [code](https://github.com/meshtastic/firmware/blob/8bc35ac6846f4b3f1e965b08dce2f2b9302b2718/src/mesh/RadioInterface.cpp#L816), channel slots and spacings are calculated from the very start of a band, and spaces are only added *after*. 
- The 3 slots will often awkwardly overlap with other services operating in the band whereas with no spacing, but with 4 slots, it woud be more likely to find at least one that doesn't.

This addition is such an important one that it really needs as much community consultation as possible, either to determine the best path forward that works for the users, or simply to educate (myself included) them, as there's likely some misunderstanding surrounding all of this.

Hoping that there can be some useful discussions surrounding this going forward :).

NomDeTom:
I won't respond to all the comments and questions here and in #8855 and #7481, but I will explain as much as I can. I will preface this by saying **I am not a subject matter expert, nor a legal professional**. This PR is submitted as an independent contributor who lives in the former EU, and while it may be adopted by Meshtastic, I am not writing for or on behalf of the project.

Firstly, I should say that there are the following concerns:
1) Out of band emissions must be controlled.
2) *If* the current EU settings are non-compliant (I am not saying either way), that has no bearing here. I am seeking to make this new setting compliant, regardless of what went before, or is accepted by others.
3) The radios that are usable with Meshtastic have no specific technical requirements for clocks or oscillators. Non-TCXO radios are permitted, as are the standard 22dbm SX1262, and also amplified radios of various types. I do not know the degree of frequency drift they incur, but given that the non-TCXO radios are able to drift so far that they are unusable past 400mS TX time, I would venture that it is there.

With specific regard to the EU868 band from 869.400-869.650MHz, I've been reading ETSI EN 300 220-2 V3.3.1 which, according to Annex A, is a harmonised standard to meet the requirements of 2014/53/EU (the Radio Equipment Directive). This is publicly available for everyone to read, and I thoroughly recommend everyone does so. Figure 6 in that shows how emissions in the out of band region are to be controlled, and shows that immediately outside of the operating channel there is a limit of 0dbm/1khz (noting that 0dbm = 1mW). I've interpreted this as a Spectral Density figure, but this is an assumption on my part.

For a radio outputting 500mW (27dbm) across 250kHz, that equates to 2mW/khz. If the frequency drifts at all, then the radio will be non-compliant. By reducing the radio output by 3dbm (27 -> 24dbm), to 250mW, the output will be 1mW/khz, or 0dbm/khz. The out of band limit reduces by (250khz x 2)/36 ~= 1dbm every 14khz, so the radio needs to be even lower if the frequency drifts outside the domain. At 5% drift (12.5khz for 250khz BW), one would need to be at 23dbm output, and so on. From this it's possible to see that a 22dbm radio operating at BW = 250khz across the whole band is **Probably OK**.

However, when we move to BW = 62.5khz, the spectral density is much higher - 8mW/khz for a 500mW radio. Luckily for us, the out of band domain applies outside of the operating channel, and the operating channel can be defined as a smaller section of the total permitted band, and you don't have to use it all. So by dividing the channel into 3x operating channels that are 83khz wide, and then using BW = 62.5khz, we give ourselves space to drift up and down before we go outside of the operating channel and have to start decreasing the output power.

If we used a band that was directly adjacent to the lower or upper edge of the band, we would have to reduce the output power to 1mW/khz or 62.5mW output (i.e. 18dbm) even with an almost perfect TCXO, and would have to reduce it further to accommodate any kind of oscillator drift.

tl;dr: We need padded operating channels to allow us to use the full power of the radios. If we operated closer to the edge of the permitted frequency band, we would need to lower the output power accordingly.

NomDeTom:
@psimolin @j0uni @tekomula @oherrala @petrisi in case you're not subscribed here, too.

rob-fi:
> I won't respond to all the comments and questions here and in #8855 and #7481, but I will explain as much as I can.

Thanks so much for writing that @NomDeTom. Although tedious to write out, it's really helpful to understand the logic and reasoning for implmenenting things in certain ways. There's some great information there to digest, and I/perhaps others will do a bit of digging (in particular with the spectral density charactaristics of LoRa modulation) and see if we come to the same conclusions, but your reasoning seems sound enough based on what's written.

Although there are many SRD devices on-band that, let's just say, aren't too strict with compliance, there's always the possibility to lead by example, especially with a project of this nature that now has become so popular. We don't want to make excuses for it being outlawed. At the same time, if there's a way to legally cater for a not insignificant user base already using particular settings, then it's worth exploring a little.

derpyspike:
@NomDeTom First of all thank you for you analysis and work.🙂

I must also say I'm not a lawyer or field expert either, but looking into the document IMHO Figure 6 (and Table 7) doesn't apply to the frequency band but rather the transmission band, in other words if you use 62.5kHz channel, your OCW, f and fnom would be the ones of that 62.5kHz channel indeed. In other words it applies to the side channel emissions resulting from the modulation.
The one that does apply to the out of frequency band emissions is the one in 4.4.8, so Figure 7 and Table 9, and that by the way does specify 0dBm/1kHz at the frequency band edges. (And you kinda reach the same conclusion you did with the Figure 6 for the current EU868 presets😅)

Although LoRa modulation doesn't have a rectangular spectral density and does have a drop at the edges (although I don't know how much), and if my understanding is correct part of certifying the European Conformity (CE mark) for EU868 is certifying it does comply with these regulations. If they didn't comply with the 4.4.6 section you brought up they would be illegal regardless of firmware settings unless operated at a low enough power.

On the TCXO part, I'm pretty sure that the only 27dBm+ EU868 device that exist is the RA01SHP that some people use for DIY devices (which If I'm not mistaken, it doesn't have a CE mark) ,other than those all other both commercial , like Heltec V4, Station G2,TBeam 1W, etc, and the ones used in most 27dBm DIY devices like the Ebyte E22/E22P all have TCXO.

That said it would be interesting to measure the devices with a Spectrum Analyzer to see what the real actual transmitted spectral density looks like😀

Other than that, I think it's important to take into consideration collisions with other LoRa devices, because using the orthogonality between Spreading Factors it can be completely avoided, also the 4 channel division, allows to use 62.5kHz and 125kHz channels simultaneously (although they would collide with the LoRaWAN RX2) while the 3 channel division you either use 62.5kHz or it collides with everything and if you transmit in the Channel 3, any other LoRa device in that part of the band with same BW but different center frequency would also collide😉, and that would be a lose-lose situation for both, and if you were to use same frequency but different SF they would both be orthogonal and be able to transmit at the same time (Like it would be if you use the current 62.5kHz 4 channel division and slot 4).
Or you could use channel 1 and channel 4 as 62.5kHz channels and use the 869.525/125kHz as a 125kHz channel and be orthogonal with LoRaWAN

Another benefit of using spacing between channels that also must be said, is that if you use the 4 channel divisions and you have Meshtastic nodes next to each other in side channels (For example one node in channel 2 and the other in channel 3 but otherwise same settings), they will hear each other and correctly decode each others packets, which even though is a bit of an edge case, we did try it as we are currently doing tests with 62.5kHz in Spain, and it works (although obviously with terrible SNR) 😂
I haven't tried it but I assume it wouldn't happen with the 3 channel division you are working on.

IMHO if the EU868 RDEF stays as it is and a new 3 channel one is created (which if I understand correctly is the idea) this shouldn't be a problem and everyone could choose based on what's best for them, like @rob-fi in Finland for example.🙂

rob-fi:
Thanks for the insight @derpyspike!

But yes the collision with other LoRa devices is a concern because so many are generally operating in the middle of the band, so keeping well away from them is an important consideration. I believe that was the whole reason the EdgeFastLow "setting" came about. There could be a 3rd channel in the middle of course if spacing was really wished, but having 4 makes more sense.

I just hope that this topic can remain open and with good discussion going forward to determine what is absoutely possible or not. It's my worry that this just gets barged through and merged without any further consideration, so I'll try to flag this for discussion as much as I can both here and in Discord, and I'll try to consult with some experts on the subject too. There's discussions brewing about a Meshtastic fork that would include such edge of frequency settings and that's the last thing I want to see happening.

With regards to the meshtastic collision of adjacent channels, well, having devices next to each other is going to do that :D. That's unfortunately a general RF problem for more than just meshtastic!

By the way here's an illustration of a 62.5kHz bandwidth meshtastic message with the marker set to 62.5kHz bandwith also. This a completely uncalibrated (other than the ppm offset of the receiver adjusted) measurement of a node a few meters away outside a window (in -15°C, clearly has TCXO), so the signal is very strong. But it serves to illustrate how the energy is concentrated within the bandwidth.

efl

NomDeTom:
@derpyspike 
> IMHO Figure 6 (and Table 7) doesn't apply to the frequency band but rather the transmission band, in other words if you use 62.5kHz channel, your OCW, f and fnom would be the ones of that 62.5kHz channel indeed. In other words it applies to the side channel emissions resulting from the modulation. The one that does apply to the out of frequency band emissions is the one in 4.4.8, so Figure 7 and Table 9, and that by the way does specify 0dBm/1kHz at the frequency band edges. (And you kinda reach the same conclusion you did with the Figure 6 for the current EU868 presets😅)

Yup. I do a lot of this late at night, and it's hard to keep one greyscale picture separate from another. The details are subtly different - e.g. the rate of slope at the edges. 4.4.8.2 says that specific limits apply to the edges of the "Permitted Frequency Band" - which drop at 0.18dBm/khz in Figure 7, rather than those in Figure 6, which uses 2x OCW for the slope (giving 0.288dBm/khz for a 62.5khz OCW). Noting that this is a stricter requirement, I think it doesn't undermine the point - the stricter of the requirements will apply, unless stated otherwise.

> Although LoRa modulation doesn't have a rectangular spectral density and does have a drop at the edges

Lora is specifically crafted so that each chirp sweeps the whole of the operating band equally.

> part of certifying the European Conformity (CE mark) for EU868 is certifying it does comply with these regulations. If they didn't comply with the 4.4.6 section you brought up they would be illegal regardless of firmware settings unless operated at a low enough power.

This is a whole other patch of thorns. My brief understanding is that you're welcome to purchase components and put them together legally. Selling them as a complete device would require a certificate.

> Other than that, I think it's important to take into consideration collisions with other LoRa devices, because using the orthogonality between Spreading Factors it can be completely avoided, also the 4 channel division, allows to use 62.5kHz and 125kHz channels simultaneously (although they would collide with the LoRaWAN RX2) while the 3 channel division you either use 62.5kHz or it collides with everything and if you transmit in the Channel 3, any other LoRa device in that part of the band with same BW but different center frequency would also collide😉, and that would be a lose-lose situation for both, and if you were to use same frequency but different SF they would both be orthogonal and be able to transmit at the same time (Like it would be if you use the current 62.5kHz 4 channel division and slot 4). Or you could use channel 1 and channel 4 as 62.5kHz channels and use the 869.525/125kHz as a 125kHz channel and be orthogonal with LoRaWAN

Lora-on-lora interference is minimal, unless the devices are on top of each other - it changes the noise floor, nothing more. Partial overlap with a source of interference is probably fine, too. I've been informed that Lora is better at recovering from random noise rather than a few scrambled bits in one place.

> Another benefit of using spacing between channels that also must be said, is that if you use the 4 channel divisions and you have Meshtastic nodes next to each other in side channels (For example one node in channel 2 and the other in channel 3 but otherwise same settings), they will hear each other and correctly decode each others packets, which even though is a bit of an edge case, we did try it as we are currently doing tests with 62.5kHz in Spain, and it works (although obviously with terrible SNR) 😂 I haven't tried it but I assume it wouldn't happen with the 3 channel division you are working on.

What you're describing is what we're trying to avoid - spurious emissions in the adjacent bands. It is literally what the document is instructing you to do.

> IMHO if the EU868 RDEF stays as it is and a new 3 channel one is created (which if I understand correctly is the idea) this shouldn't be a problem and everyone could choose based on what's best for them, like @rob-fi in Finland for example.🙂

Yup - you're all invdividuals, with your own thoughts and feelings and opinions. Noone should be telling you what you should or should not do, within the boundaries of the possible, the legal, the sensible and the safe. If one setting doesn't work, and another does, then use it. Meshtastic is all about communication, after all.

@rob-fi 
> But yes the collision with other LoRa devices is a concern because so many are generally operating in the middle of the band, so keeping well away from them is an important consideration. I believe that was the whole reason the EdgeFastLow "setting" came about. There could be a 3rd channel in the middle of course if spacing was really wished, but having 4 makes more sense.

Collision with other Lora devices in the middle of the 869.525 is **not** the primary concern - as above, I am not concerned with lora-on-lora interference in the slightest, except from general good-actor perspectives. The principle concern is with other types of uses that dominate the central band - strong FSK signals and so on - from things like temporary traffic lights and other unidentified utility type transmissions that simply block us out.

Having 4 bands would certainly be nice, but from my comments above, these are the regulations that we are trying to work inside. Out of band is measured at the edges of the operating channel. The nominal frequency is defined as the centre of the OCW. By adding some padding to the bands, we can ensure that the chances of interference are minimised and everyone will be happy.

> I just hope that this topic can remain open and with good discussion going forward to determine what is absoutely possible or not. It's my worry that this just gets barged through and merged without any further consideration, so I'll try to flag this for discussion as much as I can both here and in Discord, and I'll try to consult with some experts on the subject too. There's discussions brewing about a Meshtastic fork that would include such edge of frequency settings and that's the last thing I want to see happening.

As I've said before: if you can read the same text as me and tell me it doesn't apply, or my interpretation is incorrect then please do so. I am always happy to be corrected, have my understanding improved or simply shown the error in my process. To some degree, I wish we *had* just gone ahead with this in the first form. We've been discussing this for several months now, and I want to see progress as much as anyone.

I think the idea of a fork simply to skip setting a few custom settings is interesting. If there is motivation to make and maintain such a fork, perhaps those individuals would instead consider looking at the other areas of improvement for the project? As you can see, the code for the change is progressing, and in a way that doesn't cause confusion and disruption to the wider userbase. There is an open issues list, and plenty of improvements to be made without duplicating the efforts of other teams.

> With regards to the meshtastic collision of adjacent channels, well, having devices next to each other is going to do that :D. That's unfortunately a general RF problem for more than just meshtastic!

Yes - this is something we specifically need to avoid. The absolute last thing I want is for an enforcement action to be served on well-meaning Meshtastic users, and this can be avoided by applying ourselves to the planning stage.

> By the way here's an illustration of a 62.5kHz bandwidth meshtastic message

If you read the later sections of the ETSI EN document, it will tell you how the official tests are performed. Your picture fully illustrates the out-of-band emissions and the rectangular distribution of the radiated energy within the band. The limits are given in the standard - we must attempt to comply!

derpyspike:
@NomDeTom Thanks again for your answer, just a few quick notes🙂

> Lora is specifically crafted so that each chirp sweeps the whole of the operating band equally.

What I meant with LoRa not having a rectangular spectral density is indeed what you can see in @rob-fi image, there's a very clear drop at the edges and not constant spectral density across the channel (there is even distribution everywhere except the edges), in fact just by the image I would say you see around a 6-7dB drop at the edge frequencies compared to the center frequency, where is it indeed evenly distributed like you say.

>What you're describing is what we're trying to avoid - spurious emissions in the adjacent bands. It is literally what the document is instructing you to do.

This is just a property of RF circuits, even if you had a theoretical perfect filter and no out of channel or out of band emissions, the receiver would still have the same problem, in fact if you look at the SX1262 you would see it has "88dB blocking immunity at 1MHz offset", so with good enough signal even at 1MHz offset the same problem would appear, this is a problem with electronics generally

>Yes - this is something we specifically need to avoid. The absolute last thing I want is for an enforcement action to be served on well-meaning Meshtastic users, and this can be avoided by applying ourselves to the planning stage.

But again, I think what you are doing is extremely important, as with compliance one must make sure to check every single box and this is something where Meshtastic can and should lead by example😇

rob-fi:
Thanks again for that @NomDeTom  and @derpyspike 

> @NomDeTom Thanks again for your answer, just a few quick notes🙂
> 
> > Lora is specifically crafted so that each chirp sweeps the whole of the operating band equally.
> 
> What I meant with LoRa not having a rectangular spectral density is indeed what you can see in @rob-fi image, there's a very clear drop at the edges and not constant spectral density across the channel (there is even distribution everywhere except the edges), in fact just by the image I would say you see around a 6-7dB drop at the edge frequencies compared to the center frequency, where is it indeed evenly distributed like you say.
> 

What I'm most interested in is precisely what that drop is for LoRa. I don't know at this point does LoRa and the typical Semtech radios already have a specified fall-off at the edges in order to comply with such regulations? I'd really like to find that out but I haven't found the right documentation yet that goes into that detail.

This is important not simply for steering the course of this issue, but rather knowing that do the existing Meshtastic presets comply, and does the continued use of both Meshtastic users of EFL and also does Meshcore comply, that does the same at the other end of the band. I have a suspicion that they _might_ do BUT I do not know this, and I'd really like to find out.

> > What you're describing is what we're trying to avoid - spurious emissions in the adjacent bands. It is literally what the document is instructing you to do.
> 
> This is just a property of RF circuits, even if you had a theoretical perfect filter and no out of channel or out of band emissions, the receiver would still have the same problem, in fact if you look at the SX1262 you would see it has "88dB blocking immunity at 1MHz offset", so with good enough signal even at 1MHz offset the same problem would appear, this is a problem with electronics generally
> 
> > Yes - this is something we specifically need to avoid. The absolute last thing I want is for an enforcement action to be served on well-meaning Meshtastic users, and this can be avoided by applying ourselves to the planning stage.
> 
> But again, I think what you are doing is extremely important, as with compliance one must make sure to check every single box and this is something where Meshtastic can and should lead by example😇

I fully agree with this. However, what I will say is that it questionable about enforcing the limits for these specific presets if the existing presets also turn out to be in violation of the regulations, and there's no desire to immediately correct that.

One could argue the current Meshtastic presets need reworked to comply with regulations if legal compliance is top priority and it turns out that they violate regulations.

Jckf:
> However, what I will say is that it questionable about enforcing the limits for these specific presets if the existing presets also turn out to be in violation of the regulations, and there's no desire to immediately correct that.

I have little to no technical knowledge to contribute here, but want to point out that what you're saying is that; because the old presets might not be compliant, no effort should be made to make new presents comply either. This is in general a pretty dangerous stance.

psimolin:
> > However, what I will say is that it questionable about enforcing the limits for these specific presets if the existing presets also turn out to be in violation of the regulations, and there's no desire to immediately correct that.
> 
> I have little to no technical knowledge to contribute here, but want to point out that what you're saying is that; because the old presets might not be compliant, no effort should be made to make new presents comply either. This is in general a pretty dangerous stance.

I interpreted it as "if existing presets are in violation, there should be desire to correct them as well."

rob-fi:
> > However, what I will say is that it questionable about enforcing the limits for these specific presets if the existing presets also turn out to be in violation of the regulations, and there's no desire to immediately correct that.
> 
> I have little to no technical knowledge to contribute here, but want to point out that what you're saying is that; because the old presets might not be compliant, no effort should be made to make new presents comply either. This is in general a pretty dangerous stance.

The complete opposite :)

I was saying that there should be an immediate desire to correct them also, as well as informing those using "EFL" and also Meshcore that they're breaching regulations, if it turns out to be the case. I'm only trying to establish with absolute certainty what's in compliance and what isn't.

Jckf:
> I was saying that there should be an immediate desire to correct them also, as well as informing those using "EFL" and also Meshcore that they're breaching regulations, if it turns out to be the case. I'm only trying to establish with absolute certainty what's in compliance and what isn't.

That is probably outside the scope of this PR.

NomDeTom:
> > > However, what I will say is that it questionable about enforcing the limits for these specific presets if the existing presets also turn out to be in violation of the regulations, and there's no desire to immediately correct that.
> > 
> > 
> > I have little to no technical knowledge to contribute here, but want to point out that what you're saying is that; because the old presets might not be compliant, no effort should be made to make new presents comply either. This is in general a pretty dangerous stance.
> 
> The complete opposite :)
> 
> I was saying that there should be an immediate desire to correct them also, as well as informing those using "EFL" and also Meshcore that they're breaching regulations, if it turns out to be the case. I'm only trying to establish with absolute certainty what's in compliance and what isn't.

My view is what I stated in the first long comment - that the existing LongFast EU868 preset is **probably ok**, especially where a 22dbm radio is used. The "shoulders" of the spectrum mask imply that the fall-off is quite gentle for that bandwidth, and the lower spectral density of the wider BW is also in our favour.

If there is a correction to be made, I would seek to do it separately than this change, and I would seek to make sure that this "new" activity is compliant first and foremost. I hope that existing EdgeFastLow Meshtastic setups would move to this setup ASAP, once released, simply because it will have different channel names, frequency ranges, etc., and new users would expect to be able to use a preloaded preset rather than a custom one.

Re: other lora meshes: I have no immediate interest in copying exactly what anyone else is doing. Meshtastic has previously used The Things Network and LoraWAN as templates, but I think that can only take you so far. I also have no desire to tell anyone else how they should set themselves up to comply with the regulations. To reiterate: I am not a subject matter expert, and apparently I have no credibility ;-)

@Jckf Precisely - One PR cannot carry all this responsibility!

NomDeTom:
@rob-fi 
> I'm only trying to establish with absolute certainty what's in compliance and what isn't.

I'm afraid to obtain that level of certainty would require consulting a suitable professional. I am a professional - not for this - and I can tell you that it will not be simple, cheap or unambiguous that all radios and use-cases will be given a stamp of approval. Without crowd-funding a project to answer the question (which, to be 100% clear, I am not in favour of), the best we can do is be humble, be realistic, attempt to interpret the law as best we can, and be conscious of the fact that we may be asked to justify ourselves or change our operating methods.

rob-fi:
> > I was saying that there should be an immediate desire to correct them also, as well as informing those using "EFL" and also Meshcore that they're breaching regulations, if it turns out to be the case. I'm only trying to establish with absolute certainty what's in compliance and what isn't.
> 
> That is probably outside the scope of this PR.

I don't entirely agree with that due to the fact that the whole reason this discussion occurred was because this PR was modified to change the location and spacing of the slots in the band, which put them at odds with existing user's settings as well as general band usage.

It was modified on the premise that it would have been breaching regulations if implemented the previous way. However, if it turns out that it's *not* breaching regulations then there's an argument to discussing reverting back to as it was originally planned, which would be beneficial for existing users.

I do understand everything you've been stating, and the resources you've linked to @NomDeTom, please don't think I'm discrediting you because it's not the case. You've given lengthy and helpful explanations and I have no reason to doubt what you've been saying. My only niggle right now is that's not including the context of LoRa modulation specifically, and I really would like to apply the characteristics of LoRa into the equation, because I'm suspicious that these considerations may have been included in the specification. Unfortunately I've not yet found a paper or information on that topic in the limited time I've had to dig into it, or had time to contact someone at a regulatory body about it. I may have an opportunity to discuss with someone next week. Yes these things can be expensive but sometimes with the right connections, things happen.

I know you're keen to move this forward but I feel this needs to be paused for now to give a little more time to look into this. I can't force you to nor expect you to do that of course :).

And for sure this PR isn't the place for these wider discussions so let's leave this here, and I'll continue to look into it at least for my own satisfaction.

I didn't wade in to this discussion to just cause a fuss, I have only as good intentions as you do.

Many thanks for your efforts both for the project and for addressing this topic. It's much appreciated.

src/mesh/MeshRadio.h
```diff
@@ -10,12 +10,21 @@ struct RegionInfo {
     meshtastic_Config_LoRaConfig_RegionCode code;
     float freqStart;
     float freqEnd;
-    float dutyCycle;
-    float spacing;
+    float dutyCycle;    // modified by getEffectiveDutyCycle
+    float spacing;      // gaps between radio channels
+    float padding;      // padding at each side of the "operating channel"
```
NomDeTom, line 15: @caveman99 this is the method we're using to centralise the frequency slots within the allowed bands.

src/mesh/MeshRadio.h
```diff
@@ -10,12 +10,21 @@ struct RegionInfo {
...
     uint8_t powerLimit; // Or zero for not set
     bool audioPermitted;
     bool freqSwitching;
     bool wideLora;
+    bool licensedOnly;        // Only allow in HAM mode
```
NomDeTom, line 20: this is unused for now, but lays the ground work for the hambands

src/mesh/MeshRadio.h
```diff
@@ -10,12 +10,21 @@ struct RegionInfo {
...
+    int8_t textThrottle;      // text broadcast throttle - signed to allow future changes
+    int8_t positionThrottle;  // position broadcast throttle - signed to allow future changes
```
NomDeTom, line 22: this is used for the PositionModule.cpp changes - current check is just for non-zero.

src/mesh/MeshRadio.h
```diff
@@ -10,12 +10,21 @@ struct RegionInfo {
...
+    int8_t telemetryThrottle; // telemetry broadcast throttle - signed to allow future changes
```
NomDeTom, line 23: This is used for all the other telemetry changes - current check is just for non-zero.

src/mesh/NodeDB.cpp
```diff
@@ -1292,7 +1288,7 @@ void NodeDB::loadFromDisk()
     // Coerce LoRa config fields derived from presets while bootstrapping.
     // Some clients/UI components display bandwidth/spread_factor directly from config even in preset mode.
     if (config.has_lora && config.lora.use_preset) {
-        RadioInterface::bootstrapLoRaConfigFromPreset(config.lora);
+        RadioInterface::validateModemConfig(config.lora);
```
NomDeTom, line 1295: this is a rewrite of @thebentern checks, but using the tools we developed for this PR.

NomDeTom (commented):
Some notes

PR #9044, opened 2025-12-22
Add Rak 6421 autoconf

jp-bennett:
This adds a method to derive the MAC Address and Device ID from the UUID of a Pi HAT+.
It also adds the properly named .yaml file for the new RAK 6421 hat


jp-bennett:
Did not mean to double PR this branch. Closing this request in favor of https://github.com/meshtastic/firmware/pull/9010

PR #9045, opened 2025-12-22
Throttle GPS initial position warning

jp-bennett:


src/modules/PositionModule.cpp
```diff
@@ -428,8 +428,9 @@ int32_t PositionModule::runOnce()
     bool waitingForFreshPosition = (lastGpsSend == 0) && !config.position.fixed_position && !nodeDB->hasLocalPositionSinceBoot();
 
     if (lastGpsSend == 0 || msSinceLastSend >= intervalMs) {
-        if (waitingForFreshPosition) {
+        if (waitingForFreshPosition && !Throttle::isWithinTimespanMs(lastWarn, ONE_MINUTE_MS)) {
```
fifieldt, line 431: would you prefer to do this throttling, or just not have the log? I'd lean toward the latter since this is a log for a noop and we have adequate logging for position update ops.

fifieldt: Marking as approved anyway, support either decision :)

thebentern:
I've got this one flagged as GPS verbose guarded one upstream

PR #9063, opened 2025-12-24
Set USB_MODE=1 for CDC boards

vidplace7:
For boards where `ARDUINO_USB_CDC_ON_BOOT=1`; `-DARDUINO_USB_MODE=` should be `1`.

This is *enforced* in Arduino 3.0 and I'm not sure why this works now?

DRAFT: this merits further testing on affected devices.


fifieldt:
We've set this to 1 and 0 in various patches in the past ... there was interesting behaviour and reasoning :) Will have to dig it up

PR #9064, opened 2025-12-24
Cleanup: Remove icarus custom framework-arduinoespressif32

vidplace7:
`icarus` uses a custom (outdated) `framework-arduinoespressif32`.

This is pure tech debt... and not going to work with the coming Arduino 3.x update. It's gotta go!


fifieldt (approved):
I checked the upstream diff and it seems like it just applies options that can probably be handled by firmware patches if needed

https://github.com/mskvortsov/meshtastic-esp32-arduino-libs/compare/main...PowerFeather:powerfeather-meshtastic-arduino-lib:main

PR #9039, opened 2025-12-22
Fix Rotary enc long press

brad112358:
Rotary encoder INPUT_BROKER_SELECT_LONG events were not being generated on long press of the encoder select switch.  Thus, upper case characters could not be entered in "Free Text" messages via the encoder.

The root cause was RotaryEncoderInterruptBase::runOnce() only being called once after RotaryEncoderInterruptBase::intPressHandler() because it always returned INT32_MAX.

The easy fix is to return 20 if the press is still detected and LONG_PRESS_DURATION has not passed.

## 🤝 Attestations

- [ X] I have tested that my proposed changes behave as described.
- [ X] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [X] Other (please specify below)
Tested on DIY NRF52 Promicro E22 board

PR #7613, opened 2025-08-12
T-Lora Pager 

mverch67:
includes PR #4684 +fix for greater performance 😛

image

## Status

- [x] SX1262 LoRa
- [x] display ST7796
- [x] IO expander XL9555 (0x20)
- [x] audio codec ES8311 (0x18)
- [x] battery gauge BQ27220 (0x55)
- [x] battery charger BQ25896 (0x6B)
- [x] GPS M10Q
- [x] RTC PCF85063A (0x51)
- [x] rotary encoder
- [x] keyboard TCA8418 (0x34) + backlight
- [x] SD card
- [x] haptic DRV2605 (0x5A)
- [ ] IMU BHI260AP (0x28)
- [ ] NFC

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)
    - [x] Seeed SenseCAP Indicator

PR #9070, opened 2025-12-24
Make rotary encoder long press move to next screen

brad112358:
With this trivial improvement, a single rotary encoder is sufficient to fully use the BaseUI.  Without this change, I could find no way to access the many available information screens.

## 🤝 Attestations

- [x ] I have tested that my proposed changes behave as described.
- [x ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x ] Other (please specify below)
          Tested on DIY NRF52 Promicro E22 board

### DIFF
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -1687,7 +1687,8 @@ int Screen::handleInputEvent(const InputEvent *event)
         if (!inputIntercepted) {
             if (event->inputEvent == INPUT_BROKER_LEFT || event->inputEvent == INPUT_BROKER_ALT_PRESS) {
                 showPrevFrame();
-            } else if (event->inputEvent == INPUT_BROKER_RIGHT || event->inputEvent == INPUT_BROKER_USER_PRESS) {
+            } else if (event->inputEvent == INPUT_BROKER_RIGHT || event->inputEvent == INPUT_BROKER_USER_PRESS ||
+                       event->inputEvent == INPUT_BROKER_SELECT_LONG) {
                 showNextFrame();
             } else if (event->inputEvent == INPUT_BROKER_UP_LONG) {
                 // Long press up button for fast frame switching


Xaositek:
Shouldn't rotary encoder up/down change frames and then Press open the action menu? Help me understand why long-press is needed to change frames.

brad112358:
I don't know what the design intention was, but currently, up/down immediately activates and moves through the canned message menu, no matter which frame is current.

brad112358:
I couldn't find much documentation on BaseUI using a rotary encoder, but it's clear this hasn't been tested much, given that long press doesn't work and thus, capital letters can't be selected in free text.  See #9039.

If someone will let me know how BaseUI is supposed to work with a rotary encoder, I can probably make it work that way.

Xaositek:
If you want to match other devices like TLora Pager, it would be CW turn is next frame, CCW is previous frame. Depressing the encoder will activate the Action Menus. That would be the limit at the moment; again similar to TLora Pager.

jp-bennett:
There are special, combined left/up and down/right events that the encoder should fire.
Edit: Apparently not any longer. We just use left/right and interpret those as up/down as needed.

mverch67:
> If someone will let me know how BaseUI is supposed to work with a rotary encoder, I can probably make it work that way.

There are different rotary encoder implementations and each can be configured what action should be applied to up/down GPIOs. I assume those defined actions are the root cause of your issue. 

I suggest you try setting the encoder GPIO input events of the canned message module (using phone app or CLI) to 28, 29:
```
    moduleConfig.canned_message.inputbroker_event_cw = meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar(28);
    moduleConfig.canned_message.inputbroker_event_ccw = meshtastic_ModuleConfig_CannedMessageConfig_InputEventChar(29);
```

See also https://github.com/meshtastic/firmware/pull/7613/changes/c0f8437c0ae2db12e3d2ce4f1f966a2500d7e635.

brad112358:
> I suggest you try setting the encoder GPIO input events of the canned message module (using phone app or CLI) to 28, 29:

Thanks!

Yes, this makes it mostly usable except when entering free text the longer beep generated by USER_PRESS and ALT_PRESS greatly slow down input vs the shorter beep of UP/DOWN.  Note, that even when configured for UP/DOWN, rotary events are slowed down by the beep.  Usability would be much improved by a shorter "click" or, buffered input events.  Or, perhaps the buzzer feedback can be disabled for these events when entering free text.  The advantage of the rotary encoder is supposed to be rapid intuitive movement in long lists like free text via spinning the dial, but that advantage is not realized yet.  I may as well have used buttons.

Note, I had to use the CLI to set this as the android app does not offer "user press" or "alt press" as options for input events in the canned message module.

My bug fix PR #9039 still is being ignored, so upper case letters still can't be entered via my rotary encoder.

brad112358:
I have opened [a bug for the Android app](https://github.com/meshtastic/Meshtastic-Android/issues/4112) to request that it support the required events for rotary encoders.

brad112358:
I have opened a new PR, #9146 which improves the responsiveness of the rotary encoder by reducing the delay caused by buzzer feedback for USER_PRESS and ALT_PRESS events.

I'll go ahead and close this PR.

PR #9071, opened 2025-12-25
Implement HAS_PHYSICAL_KEYBOARD for devices with physical keyboards

Xaositek:
- Implement HAS_PHYSICAL_KEYBOARD for devices with physical keyboards
- Add HAS_PHYSICAL_KEYBOARD to variant.h for:
  - TDeck
  - TLora Pager
  - TDeck Pro


thebentern:
@Xaositek what if we add a HAS_PHYSICAL_KEYBOARD macro, so that we can guard against current and future devices at the variant level?

Xaositek:
@thebentern Let me know if you have other devices I've missed

thebentern (approved):

PR #9075, opened 2025-12-25
KZ_863 is not wide lora

fifieldt:
KZ_863 was set to wide_lora = true. This was a mistake, induced because the regulations would allow SHORT_TURBO. However, that interpretation of the code was incorrect and wide_lora has a different meaning.

Fixes https://github.com/meshtastic/firmware/issues/9054


fifieldt:
Simple/limited change that fixes a clear bug needed by a community, no review/objections for a week, merging :)

NRG316:
> KZ_863 was set to wide_lora = true. This was a mistake, induced because the regulations would allow SHORT_TURBO. However, that interpretation of the code was incorrect and wide_lora has a different meaning.
> 
> Fixes #9054

Thanks!

PR #9080, opened 2025-12-26
Add menus for Smart Position, Broadcast Interval and Position Interval

Xaositek:
Add menus for Smart Position, Broadcast Interval and Position Interval to the Position screen

### DIFF
--- a/src/graphics/draw/MenuHandler.cpp
+++ b/src/graphics/draw/MenuHandler.cpp
@@ -1406,7 +1406,7 @@ void menuHandler::GPSUpdateIntervalMenu()
     BannerOverlayOptions bannerOptions;
     bannerOptions.message = "Update Interval";
     bannerOptions.optionsArrayPtr = optionsArray;
-    bannerOptions.optionsCount = 15;
+    bannerOptions.optionsCount = 16;
     bannerOptions.bannerCallback = [](int selected) -> void {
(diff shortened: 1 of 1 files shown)


src/graphics/draw/MenuHandler.cpp
```diff
@@ -1346,6 +1367,188 @@ void menuHandler::GPSFormatMenu()
...
+        bannerOptions.InitialSelected = 12;
+    } else {
+        bannerOptions.InitialSelected = 0;
+    }
+    screen->showOverlayBanner(bannerOptions);
+}
+
+void menuHandler::GPSPositionBroadcastMenu()
+{
+    static const char *optionsArray[] = {"Back",     "1 Hour",   "2 Hours",  "3 Hours",  "4 Hours",  "5 Hours", "6 Hours",
```
fifieldt: These need work. 1 hour is too high for the lowest, and I've not seen a use case to date for more than 24 hours. Android has a list - use that?

Xaositek: The options in this list are from the iOS app which I copied.

Xaositek: Aligned options to Android app to provide necessary options

fifieldt (approved):

PR #8061, opened 2025-09-21
RTC: PCF85063 support, port to SensorLib 0.3.1

WillyJL:
- Support for PCF85063, enables RTC on T-Lora Pager
- Supersedes #7886
- Fixes #7885
- Port all `SensorLib` usages to `0.3.1`, port `PCF8563_Library` usages to `SensorLib`
  - Notably, this fails at linking step on nRF52 platform, I have opened a bug report:
  - Could also use some testing that BMA423 on t-watch-s2 still works, API changed quiet a bit and I hope I ported it correctly
- Removes PCF8563 and related defines for devices that AFAICT do not have a PCF8563
  - Could use some help double checking I did not add any regressions as for some devices I cuold not find very detailed information on this
- Adds log warnings in case RTC address is defined but RTC was not found at runtime

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [x] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)
    - [x] T-Lora Pager

PR #9084, opened 2025-12-26
Fix PR#8061 SensorLib nRF ThinkNode M-series

vidplace7:
Addendum to https://github.com/meshtastic/firmware/pull/8061
Try-fix Elecrow thinknode NRF M-series.

`lewisxhe/PCF8563_Library` has been replaced with `lewisxhe/SensorLib` and it seems just these two were missed.

### DIFF
--- a/variants/nrf52840/ELECROW-ThinkNode-M3/platformio.ini
+++ b/variants/nrf52840/ELECROW-ThinkNode-M3/platformio.ini
@@ -15,5 +15,5 @@ lib_deps =
   ${nrf52840_base.lib_deps}
   # renovate: datasource=custom.pio depName=nRF52_PWM packageName=khoih-prog/library/nRF52_PWM
   khoih-prog/nRF52_PWM@1.0.1
-  # renovate: datasource=custom.pio depName=PCF8563 packageName=lewisxhe/library/PCF8563_Library
-  lewisxhe/PCF8563_Library@1.0.1
+  ; # renovate: datasource=custom.pio depName=SensorLib packageName=lewisxhe/library/SensorLib
+  ; lewisxhe/SensorLib@0.3.3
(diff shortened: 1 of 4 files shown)


vidplace7:
@WillyJL please review :pray:

WillyJL:
i dont believe this is the optimal fix: these 2 devices include the library and define PCF8563_RTC, but *do not* define HAS_RTC. this means they do not actually query the RTC nor use it in any way from what i can understand of the code. this is similar to other things in #8061 where i removed PCF8563 and sensorlib from a lot of devices that included it but didnt use it, because they in fact do not have an RTC and it was just included as copy paste from other variant files.

from what i can tell the correct fix here is to remove sonsorlib for these 2 variants aswell and remove the PCF8563_RTC from their variant.h files.

WillyJL:
of course its possible im misunderstanding the code, in that case updating like you did would work yes. but if these devices dont come with an rtc then no need to include the library

vidplace7:
> i dont believe this is the optimal fix: these 2 devices include the library and define PCF8563_RTC, but _do not_ define HAS_RTC. this means they do not actually query the RTC nor use it in any way from what i can understand of the code. this is similar to other things in #8061 where i removed PCF8563 and sensorlib from a lot of devices that included it but didnt use it, because they in fact do not have an RTC and it was just included as copy paste from other variant files.
> 
> from what i can tell the correct fix here is to remove sonsorlib for these 2 variants aswell and remove the PCF8563_RTC from their variant.h files.

I've modified this PR to remove rather than replace :+1:

WillyJL:
from a quick google it seems like thinknodes come with rtc, but i cant find any info on which rtc they use, and regardless for how the code was setup previously it was not in use anyway, i would assume more work would be needed to make it work.

happy holidays!

vidplace7:
> Maybe the lines should be commented with a link to this or other than removed ...

:+1: done

PR #6625, opened 2025-04-18
T5-4.7-S3 Epaper Pro support

mverch67:
resolves #6607 

## Status

V1: hw version v1.0-240810
V2: hw version v1.0-241224

- [x] V1: SX1262 LoRa
- [x] V2: SX1262 LoRa
- [x] E-Ink display ED047TC1
- [x] V2: EInk power driver TPS65185 (0x68)
- [x] GT911 touch controller (0x5D)
- [x] PMU BQ25896 (0x6B) 
- [x] Battery Mgmt BQ27220 (0x55)
- [x] RTC PCF85063A (0x51) see #8061 
- [x] V2: GNNS MIA-M10Q
- [x] SD card
- [x] V2: IO expander XL9555 (0x20)

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [x] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)
    - [x] T5-4.7-S3 Epaper Pro V1
    - [x] T5-4.7-S3 Epaper Pro V2

PR #9096, opened 2025-12-28
Fix T5 E-Paper Pro V2 display initialization

Ringmast4r:
## Summary
- Fixed T5 E-Paper Pro V2 (H752-01/H752-02) display initialization
- Initialize all PCA9535 Port 0 pins (0-7) as outputs with HIGH values to match stock firmware

## Problem
The display would not power on correctly on the LilyGO T5S3-4.7-e-paper-PRO V2 because only IO expander pin 0 was being initialized. The stock firmware initializes all 8 pins on Port 0.

## Solution
Changed the initialization loop to set all 8 pins (0-7) as outputs with HIGH values, matching the behavior in the stock firmware's `epd_board_v7.c`.

## Testing
- Tested on LilyGO T5S3-4.7-e-paper-PRO V2 (H752-01, 915MHz variant)
- Display now powers on and renders correctly
- Touch input working
- Bluetooth connection working
- LoRa radio functional

### DIFF
--- a/src/graphics/EInkParallelDisplay.cpp
+++ b/src/graphics/EInkParallelDisplay.cpp
@@ -83,8 +83,13 @@ bool EInkParallelDisplay::connect()
         epaper->initPanel(BB_PANEL_LILYGO_T5PRO, 28000000);
 #elif defined(T5_S3_EPAPER_PRO_V2)
         epaper->initPanel(BB_PANEL_LILYGO_T5PRO_V2, 28000000);
-        epaper->ioPinMode(0, OUTPUT);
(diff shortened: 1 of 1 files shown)


mverch67:
Wouldn't it be better to commit the changes into https://github.com/meshtastic/firmware/pull/6625 instead of duplicating the entire PR?

Ringmast4r:
Closing this PR per @mverch67's suggestion. I'll contribute my V2 display initialization fix to #6625 instead to keep everything unified. Thanks for the guidance\!

PR #9041, opened 2025-12-22
Store and Forward Plus Plus

jp-bennett:
Been working on this one for quite a while:
https://discord.com/channels/867578229534359593/931765413350613052/1144001594178936853

This is strictly limited to native for the moment, but seems to work in the limited test case my local mesh presents.

The idea is that a central Linux node tracks messages on the primary channel as a git-style hash chain, and periodically broadcast the tip-of-chain information. Other Linux nodes will check their local copy of the chain, and request missing messages, as well as cache and forward messages not yet on the chain.

This needs more work before mass deployment, but is *probably* ready for some targeted testing.

## Summary by CodeRabbit

- **New Features**
  - Added Store-and-Forward++ support for persisting, synchronizing, validating, and forwarding messages across mesh networks.
  - Added configurable storage, synchronization, chain, hop, and announcement settings.
  - Added optional routing port whitelists and no-hop port controls.
- **Bug Fixes**
  - Prevented Store-and-Forward++ from operating when forced simulated radio mode is enabled.
  - Ensured configured no-hop ports are not rebroadcast through additional mesh hops.
- **Configuration**
  - Added support for loading and exporting the new Store-and-Forward++ and routing settings.

PR #9102, opened 2025-12-28
[feat] user config flag to skip hop decrement

matkam:
In a large mesh such as http://bayme.sh, we want to make sure all edges of the mesh feel included, and not just the middle. Hop decrement skipping has taken us far using Router type nodes, but there should an option to allow packets to travel far with all other device roles as well. This introduces the option to skip hops for all device types.

## 🤝 Attestations

  - [x] RAK WisBlock 4631
  - [x] B&Q Station G2

### DIFF
--- a/src/mesh/NextHopRouter.cpp
+++ b/src/mesh/NextHopRouter.cpp
@@ -138,7 +138,7 @@ bool NextHopRouter::perhapsRebroadcast(const meshtastic_MeshPacket *p)
                     if (shouldDecrementHopLimit(p)) {
                         tosend->hop_limit--; // bump down the hop count
                     } else {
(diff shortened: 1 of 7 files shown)


jp-bennett:
Please take a look at https://github.com/meshtastic/firmware/pull/9041 as a possible alternative to hop count solutions. In fact, I would love to get some bay mesh volunteers to help test this soon.

matkam:
> Please take a look at https://github.com/meshtastic/firmware/pull/9041 as a possible alternative to hop count solutions. In fact, I would love to get some bay mesh volunteers to help test this soon.

Wild (and welcomed) solution! Can we do both? Happy to test it and find more volunteers once you are ready.

PR #9112, opened 2025-12-29
feat: Add optional Perfect Forward Secrecy for direct messages

krisclarkdev:
## Summary

This PR adds optional Perfect Forward Secrecy (PFS) for direct messages. When two nodes both support ephemeral keys, the message encryption gains forward secrecy - meaning even if long-term keys are later compromised, past messages remain secure.

**This is fully backward compatible.** Nodes running this code can communicate with stock firmware using the existing PKI encryption.

### Architecture

Each node generates an ephemeral Curve25519 keypair at boot. This ephemeral public key gets broadcast via NodeInfo so other nodes learn about it. When sending a DM to a node that supports PFS, we derive a session key using Triple-DH:

```
DH1 = DH(local_identity_private, remote_ephemeral_public)
DH2 = DH(local_ephemeral_private, remote_identity_public) 
DH3 = DH(local_ephemeral_private, remote_ephemeral_public)
session_key = SHA256(DH1 || DH2 || DH3)
```

This provides mutual authentication (DH1, DH2) and forward secrecy (DH3).

### Key Files

- `EphemeralKeyManager.h/.cpp` - Manages ephemeral keypairs, rotation, and remote key caching
- `CryptoEngine.cpp` - `deriveTripleDHSessionKey()`, `encryptWithPFS()`, `decryptWithPFS()`
- `Router.cpp` - Integration in `perhapsEncode()` and `perhapsDecode()`
- `NodeInfoModule.cpp` - Broadcasting ephemeral keys
- `main.cpp` - Initialization
- `test/test_pfs/` - Unit tests

### Key Rotation

Keys rotate after 100 messages or 24 hours (whichever comes first). On rotation, NodeInfo is re-broadcast so peers update their cache before the next outgoing message.

### Fallback Behavior

The decryption path tries legacy PKI first, then PFS. This means:
- PFS node → stock node: works (falls back to legacy encryption)
- Stock node → PFS node: works (legacy decryption succeeds)
- PFS node → PFS node: uses forward secrecy

## Testing Status

**Compiles clean on:**
- ESP32 (tbeam) - 95.1% flash
- STM32 (rak3172) - 73.6% flash

**What I've verified:**
- Builds without warnings
- Key generation logic reviewed
- Rotation logic reviewed
- Fallback paths reviewed

**What needs testing:**
I only have one device right now - my second T-Beam is being replaced. I'd really appreciate if anyone in the community with spare devices could flash this and test:
1. PFS node ↔ PFS node communication
2. PFS node ↔ stock firmware node (should fall back gracefully)
3. Power cycle and verify keys reload correctly (`PFS: Loaded from flash` in logs)
4. Send 100+ messages and verify rotation happens

## Things to Watch

1. **Flash usage** - ESP32 is at 95.1% which is tight but not blocker-level
2. **Timing** - Key rotation broadcasts NodeInfo which adds mesh traffic. In normal use (100 msgs/24h) this should be fine
3. **Edge case** - If Alice rotates mid-conversation and the NodeInfo doesn't reach Bob before the next message, that message could fail to decrypt. I've added immediate NodeInfo broadcast after rotation to minimize this window.

## Debug Mode

For testing key rotation quickly, you can add `-D PFS_TEST_MODE` to build_flags. This reduces the rotation threshold to 5 messages instead of 100.

Boot logs show the current state:
```
PFS: Loaded from flash - keyId=3, age=2 hours, msgs=45/100, remoteKeys=2
PFS: Ready - keyId=3, pubkey[0:3]=a3b2c1d0
```

## Related Issues

Forward secrecy has been discussed in the community for improving message security. This implementation aims to provide it transparently without requiring user configuration.


jp-bennett:
Unfortunately, it looks like a bunch of other commits got mixed in, here. The easiest way forward may be to create a new branch based on master, cherry-pick these commits in, and create a new PR. Hard to review with extra stuff mixed in.

PR #9114, opened 2025-12-29
Noop "download" portion of #shame

vidplace7:
FYI @Jorropo 

This disables the "Download old Manifest" portion of #shame until it can be re-engineered to not do a nasty for loop behind the scenes.


Jorropo:
Yeah thanks, later this week I'll try to see if we can find the run ID for a given commit.
If not easily within CI we are gonna remove this from CI and I'll make a bot off CI that handle sending comments.

PR #9118, opened 2025-12-29
Fix NimBLE 2 migration

### LINKED ISSUE #9049: [Bug]: Heltec V3 has broken BLE on 2.7.17
### Category

BLE

### Hardware

Heltec V3

- [ ] InkHUD ePaper

### Firmware Version

2.7.17.83c6161

### Description

After updating to version 2.7.17, BLE connectivity on the Heltec V3 stopped working. The device boots normally, the serial debug output looks fine, but when attempting to connect from the Android app nothing happens—there is no error on the phone or in the serial log. After flashing back 2.7.16 then BLE works fine. Tested twice, same result.

### Relevant log output

```Shell

```

thebentern:
Closes #9049

### DIFF
--- a/src/nimble/NimbleBluetooth.cpp
+++ b/src/nimble/NimbleBluetooth.cpp
@@ -853,6 +853,7 @@ void NimbleBluetooth::setupService()
 
 void NimbleBluetooth::startAdvertising()
 {
+#if defined(CONFIG_IDF_TARGET_ESP32S3) || defined(CONFIG_IDF_TARGET_ESP32C3) || defined(CONFIG_IDF_TARGET_ESP32C6)
     NimBLEExtAdvertising *pAdvertising = NimBLEDevice::getAdvertising();
     NimBLEExtAdvertisement legacyAdvertising;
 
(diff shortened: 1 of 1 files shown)


variants/esp32/esp32.ini
```diff
@@ -37,8 +37,8 @@ build_flags =
   -DMYNEWT_VAL_BLE_HS_LOG_LVL=LOG_LEVEL_CRITICAL
   -DAXP_DEBUG_PORT=Serial
   -DCONFIG_BT_NIMBLE_ENABLED
+  -DCONFIG_BT_NIMBLE_NVS_PERSIST=1
   -DCONFIG_BT_NIMBLE_MAX_BONDS=6 # default is 3
-  -DCONFIG_BT_NIMBLE_ROLE_CENTRAL_DISABLED
```
vidplace7, line 41: Do we use the `CENTRAL` role? It's got a significant flash/heap footprint.
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/performance/size.html#bluetooth-nimble

```
-DCONFIG_BT_NIMBLE_ROLE_CENTRAL_DISABLED
```
should probably be kept. Maybe even also add:
```
-DCONFIG_BT_NIMBLE_ROLE_OBSERVER_DISABLED
```

PR #9123, opened 2025-12-30
Arduino 3.x: Test Tasmota Platform (3.1.x)

vidplace7:
TEST CI probe (DO NOT MERGE)

---

Checking out the tasmota platform (vs pioarduino)

@phaseloop


vidplace7:
I guess CI only runs against `master` and `develop`...
https://github.com/vidplace7/meshtastic-firmware/actions/runs/20600505486
Here we go

vidplace7:
Tasmota Platform strips out NetworkClientSecure. Which we use in the webserver and mqtt modules.

We'll need to instead use pioarduino and strip it down ourselves.

PR #9116, opened 2025-12-29
feat: Add optional Perfect Forward Secrecy for direct messages

krisclarkdev:
## Summary

Adds optional Perfect Forward Secrecy (PFS) for direct messages using ephemeral Curve25519 key exchange.

**Fully backward compatible** - nodes running this code can still communicate with stock firmware.

## How It Works

Each node generates an ephemeral keypair at boot. When two nodes both support PFS, messages between them use Triple-DH key derivation:

```
DH1 = DH(local_identity_priv, remote_ephemeral_pub)
DH2 = DH(local_ephemeral_priv, remote_identity_pub)
DH3 = DH(local_ephemeral_priv, remote_ephemeral_pub)
session_key = SHA256(DH1 || DH2 || DH3)
```

This provides mutual authentication and forward secrecy.

## Key Features

- **EphemeralKeyManager** - Generates/rotates Curve25519 ephemeral keys
- **Triple-DH derivation** - Mutual auth + forward secrecy
- **Key rotation** - After 100 messages or 24 hours
- **LRU cache** - 32 remote key entries
- **Persistence** - Keys saved to flash with checksum validation
- **Graceful fallback** - Falls back to legacy PKI if PFS unavailable

## Files Changed

- `src/mesh/EphemeralKeyManager.h/.cpp` - New ephemeral key manager
- `src/mesh/CryptoEngine.h/.cpp` - Triple-DH and PFS encrypt/decrypt
- `src/mesh/Router.cpp` - PFS integration in encode/decode paths
- `src/main.cpp` - EphemeralKeyManager initialization

## Build Status

Tested on tbeam (ESP32) - 95.8% flash usage. Compiles clean.

## Testing Needed

I only have one device currently. Would appreciate community testing:
1. PFS node ↔ PFS node communication
2. PFS node ↔ stock firmware (fallback test)
3. Key persistence across power cycles
4. Key rotation behavior (build with `-D PFS_TEST_MODE` for 5-message rotation)

PR #9154, opened 2026-01-02
add a .clang-format file

Jorropo:
The previous setup would be extremely lenient.
AFAIK would basically only enforce tabulation more or less.

Add a .clang-format file, the point is you can't accidently have whitespace changes. See #9116 for what sitatuation we are in before this patch.

The point is there is a single valid way to format the file, and running trunk fmt -a will bring you to it.


Jorropo:
In VC @thebentern pointed out this is a big merge conflicts risk.

I agree, however it is easy to fix.
If a conflict happen you can copy the `.clang-format` file from my PR, run `trunk fmt -a`, commit then merge this PR into your branch and the conflict should go away.

PR #9172, opened 2026-01-04
Revert "add a .clang-format file (#9154)"

Jorropo:
I thought git would be smart enough to understand all the whitespace changes but even with all the flags I know to make it ignore theses it still blows up if there are identical changes on both sides.
I have a solution but it require creating a new commit at the merge base for each conflicting PR and merging it into develop.
I don't think blowing up all PRs is worth for now, maybe if we can coordinate this for V3 let's say.

This reverts commit 0d11331d185a60adf84e64ec98642baf6a15a95a.


Jorropo:
Actually I have let an other idea.

Jorropo:
Well it's much better but still more work than I thought.

Jorropo:
Basically `git merge 0d11331d185a60adf84e64ec98642baf6a15a95a~1 && git show 0d11331d185a60adf84e64ec98642baf6a15a95a:.clang-format > .clang-format && trunk fmt -a && git add . && git commit -m "add .clang-format on local branch" && git merge 0d11331d185a60adf84e64ec98642baf6a15a95a` but the second merge doesn't automatically succeeds.

Issue #8790, opened 2025-11-28
[Bug]: own message sent out received again about 20 minutes later


shalberd:
### Category

Other

### Hardware

Rak4631

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.15.567b8ea Alpha

### Description

Hello

I have my RAK4631 node set as client_mute and using firmware 2.7.15.

I send out a message, replying to another message that comes in in channel 0.

![Image](https://github.com/user-attachments/assets/b49c71b0-9eb6-443e-ab10-c5a078f3cc35)

It shows up in the app, so far, so good and unsuspect. Debug log of my sent-out-message:
```

to: 4294967295 (!ffffffff)
id: 4080155345
want_ack: true
decoded {
  portnum: TEXT_MESSAGE_APP
  payload: "xxxxxxxxx 3 hops. Gr\303\274ess  xxxx"
  reply_id: 3756052286
}
Decoded Payload:
{
xxxxxxxxx 3 hops. Grüess xxxx
}
```

34 minutes later, I get my own message again in the app, with the same id, this time indicating though that it has been around in the mesh and now returning to me. Coming from me, yes, same message id. Same replay-id.
Now, maybe this is important: the node mentioned as relay_node runs on firmware 2.6.11:

from: 1306336680 (!xxxxxxda8)
to: 4294967295 (!ffffffff)
id: 4080155345
rx_time: 1764358238
rx_snr: 2.75
rx_rssi: -105
hop_start: 5
relay_node: xxxxxxxxxxxch (!xxxxxx697)
transport_mechanism: TRANSPORT_LORA
decoded {
  portnum: TEXT_MESSAGE_APP
  payload: "xxxxxxxxx 3 hops. Gr\303\274ess  xxxx"
  reply_id: 3756052286
  bitfield: 0
}
Decoded Payload:
{
xxxxxxxxx 3 hops. Grüess xxxx
}

@GUVWAF I remember in FW 2.13, you merged stuff into master that made duplicates disappear. 
https://github.com/meshtastic/firmware/pull/8450

Here now in 2.7.15, the behavior happens again. I am unsure if it is related.

@h3lix1 I think you also worked on some duplicate removal stuff in 2.7.12: https://github.com/meshtastic/firmware/pull/8148/files

[...]

PR #9180, opened 2026-01-04
Add list of text message packet IDs, and check for dupes

jp-bennett:
Adds a trivial list of seen packet IDs in the text message module, and then checks for a matching packet ID in FloodingRouter, to avoid retransmitting old text packets.

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -246,7 +246,8 @@ uint32_t RadioInterface::getPacketTime(const meshtastic_MeshPacket *p, bool rece
 /** The delay to use for retransmitting dropped packets */
 uint32_t RadioInterface::getRetransmissionMsec(const meshtastic_MeshPacket *p)
 {
-    size_t numbytes = pb_encode_to_bytes(bytes, sizeof(bytes), &meshtastic_Data_msg, &p->decoded);
+    size_t numbytes =p->which_payload_variant == meshtastic_MeshPacket_decoded_tag ? 
+      pb_encode_to_bytes(bytes, sizeof(bytes), &meshtastic_Data_msg, &p->decoded) : p->encrypted.size+MESHTASTIC_HEADER_LENGTH;
     uint32_t packetAirtime = getPacketTime(numbytes + sizeof(PacketHeader));
     // Make sure enough time has elapsed for this packet to be sent and an ACK is received.
     // LOG_DEBUG("Waiting for flooding message with airtime %d and slotTime is %d", packetAirtime, slotTimeMsec);


GUVWAF:
So we are currently seeing duplicates of text messages that are not in the PacketHistory anymore?

jp-bennett:
@GUVWAF Yes. Both through higher hop counts, and in the new store and forward++ tests. It seems to only be text messages that are causing the problem, so I went for the simple solution.

GUVWAF:
I'm wondering whether it's really only text messages that cause this, or whether it's just only noticeable for text messages.

Not so long time ago we moved to a static memory allocation for the PacketHistory and it might be we just need to increase its size. PacketHistory is now also used for routing decisions (using the next_hop and relay_node bytes).

jp-bennett:
> I'm wondering whether it's really only text messages that cause this, or whether it's just only noticeable for text messages.

Store and Forward ++ is limited to text messages, so half of the cause is only texts.

> 
> Not so long time ago we moved to a static memory allocation for the PacketHistory and it might be we just need to increase its size. PacketHistory is now also used for routing decisions (using the next_hop and relay_node bytes).

Something I noticed about PacketHistory is that we seem to have 4 PacketHistory pools: one for each class that inherits from Router, unless I'm misunderstanding how the class inheritance works. And at that, it looks like only flooding router and next hop router actually write to their pools.

My original idea was to make this a lighter message store (only 4 bytes per message) that would grab text messages when they fell off the end of the PacketHistory. But with the split between the two router types, that became a pretty complicated task.

thebentern:
> Something I noticed about PacketHistory is that we seem to have 4 PacketHistory pools: one for each class that inherits from Router, unless I'm misunderstanding how the class inheritance works. And at that, it looks like only flooding router and next hop router actually write to their pools.

The better approach perhaps would be to ensure that we have a global singleton packethistory to consolidate these rather than add the complexity of another one?

GUVWAF:
> Store and Forward ++ is limited to text messages, so half of the cause is only texts.

Hmm, I see. However, considering the existing issue #8790, which seems to be caused by `ROUTER_LATE` we might be better off increasing the PacketHistory size in general.

> Something I noticed about PacketHistory is that we seem to have 4 PacketHistory pools: one for each class that inherits from Router, unless I'm misunderstanding how the class inheritance works. 

No, we only have one router object, which is the most-derived class, namely ReliableRouter: https://github.com/meshtastic/firmware/blob/9f5170a0bc034d52e42205f3d80b1ead63fba9b8/src/main.cpp#L805

jp-bennett:
> No, we only have one router object, which is the most-derived class, namely ReliableRouter:
> 
> https://github.com/meshtastic/firmware/blob/9f5170a0bc034d52e42205f3d80b1ead63fba9b8/src/main.cpp#L805

Ah-hah! That was the detail I was missing.

Regardless, SF++ will deliver missed text messages up to an hour later, so some sort of dedicated text message pool seems reasonable.

GUVWAF:
> so some sort of dedicated text message pool seems reasonable.

Yes, fair enough. Then in general this looks good to me.

GUVWAF (approved):

jp-bennett:
@GUVWAF Had a thought about this last night: Should we check the new message list first, so as not to prevent these rebroadcasts from poisoning the PacketHistory?

GUVWAF:
I think it's better to have them in the PacketHistory first, as we now store things like `next_hop` and `relayed_by` there also, which is used for next-hop routing decisions.

jp-bennett:
> I think it's better to have them in the PacketHistory first, as we now store things like `next_hop` and `relayed_by` there also, which is used for next-hop routing decisions.

Yes, that's what I mean. If these are strangely routed packets coming from the wrong direction, we really don't want them influencing next-hop routing decisions.

shalberd:
> I think it's better to have them in the PacketHistory first

> If these are strangely routed packets coming from the wrong direction

I think the Copilot review was only requested after this PR was merged. From what I understand with my limited view, it seems the copilot comments above, one of you should have a look again. 

What do you think of the comments, especially the one on order-of-operations?

PR #9181, opened 2026-01-05
ELECORW-ThinkNode-M8-V0.1-firmware

Koi-elecrow:
Hello, Team.

New: ThinkNode-M8: Main control logic for ThinkNode M8.

Tested on ThinkNode-M8.

### DIFF
--- a/boards/ThinkNode-M8.json
+++ b/boards/ThinkNode-M8.json
@@ -1,0 +1,53 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "nrf52840_s140_v6.ld"
+    },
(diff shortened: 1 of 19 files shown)


Koi-Sunyh:
> Quick review of this the "PresetMessageModule.cpp/.h" appear to be a repeat of "CannedMessageModule.cpp/.h" but gated only for the ThinkNode M8. Is there a reason for this?
These two files are preset pages created for the ThinkNode M8 device. They contain the predefined alert levels and preset content we have configured. The actual display can be viewed directly on the ThinkNode M8 device.

PR #9184, opened 2026-01-05
RadioInterface::getRetransmissionMsec now handles encrypted packets c…

rbreesems:
This is a one line PR (no issue created) that changes `RadioInterface::getRetransmissionMsec` to handle the case of the packet being encrypted. This method is called by NextHopRouter when adding a packet for retranmssion. When the node is at least one hop away from the sending node, the packet handed to this method is encrypted.

When the packet is encrypted, the current code throws an error `Panic: can't encode protobuf reason='bytes size exceeded'` because of the call to `pb_encode_to_bytes`. When this happens whatever value that gets placed in `numbytes` by this method results in the packet never being retransmitted about 90% of time (at least in my tests).

The fix sets the `numbytes` value to the correct size of the encrypted packet if it is encrypted.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -246,7 +246,8 @@ uint32_t RadioInterface::getPacketTime(const meshtastic_MeshPacket *p, bool rece
 /** The delay to use for retransmitting dropped packets */
 uint32_t RadioInterface::getRetransmissionMsec(const meshtastic_MeshPacket *p)
 {
-    size_t numbytes = pb_encode_to_bytes(bytes, sizeof(bytes), &meshtastic_Data_msg, &p->decoded);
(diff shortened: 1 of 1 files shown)


rbreesems:
For some reason, I cannot add a tag to this Merge Request, should be 'bugfix'.

Also, cannot run the formatter for some reason.

rbreesems:
I now realize that the `pb_encode_to_bytes` PANIC is intentional and returns 0 resulting in a packet size is just `sizeof(PacketHeader)`.  However, it is disconcerting to see PANIC error messages in the debug log and it should just return the encrypted packet size.

PR #9188, opened 2026-01-05
feat: Add ACK batching to reduce airtime in busy meshes

### LINKED ISSUE #9187: [Feature Request]: Batch ACK responses to reduce LoRa transmission overhead
### Platform

Cross-Platform

### Summary

Aggregate multiple ACK responses into a single LoRa transmission to reduce airtime overhead in busy mesh networks.

### Problem Statement

Currently, every packet sent with `want_ack=true` generates its own individual ACK packet in response. In a busy mesh with many nodes exchanging messages, position updates, and telemetry, this creates significant airtime overhead:

**Example scenario (current behavior):**
```
Node A sends 5 messages to Node B in quick succession (want_ack=true) 
Node B sends 5 separate ACK packets back to Node A = 5 individual LoRa transmissions, each with its own preamble and headers
```

Each ACK transmission consumes:
- LoRa preamble time
- Packet header overhead
- Transmission slot (which could be used for actual data)
- Radio duty cycle budget

In a mesh with 20+ active nodes during a busy period, ACK traffic can consume a substantial portion of available channel capacity.

### Proposed Solution

Implement an optional ACK batching mechanism that queues pending ACKs for a short window (100-500ms) and sends them as a single aggregated packet:

**Proposed behavior:**
```
Node A sends 5 messages to Node B in quick succession (want_ack=true) 
Node B waits briefly (e.g., 200ms), collecting pending ACKs 
Node B sends 1 batched ACK packet containing all 5 acknowledgments = 1 LoRa transmission instead of 5
```

### Benefits

| Benefit | Description |
|---------|-------------|
| **Reduced Airtime** | Significant reduction in ACK-related transmissions during burst traffic (e.g., 5 packets batched into 1 ACK = 80% fewer transmissions for that burst) |
| **Improved Channel Availability** | More capacity for actual messages instead of acknowledgments |
| **Better Scalability** | Meshes with many nodes benefit the most—ACK overhead doesn't scale linearly with traffic |
| **Lower Power Consumption** | Fewer radio transmissions = longer battery life on all nodes |

### Technical Approach

1. **New Protobuf Message**: Add `BatchedRouting` message type to aggregate multiple ACK IDs
2. **ACK Queue**: Buffer non-urgent ACKs in `RoutingModule` with a configurable time window
3. **Flush Triggers**: Send batched ACK when either:
   - Batch window expires (e.g., 200ms since first pending ACK)
   - Batch size limit reached (e.g., 10 ACKs)
4. **Preserve Urgent ACKs**: ACKs with `ackWantsAck=true` (reliable ACKs) still send immediately
5. **Receiver Handling**: Update `ReliableRouter` to handle batched ACK packets and stop retransmissions for all included IDs

### Configuration Options

Consider adding a new config option:
```protobuf
message DeviceConfig {
    // ...existing fields...
    bool ack_batching_enabled = X;       // Enable/disable ACK aggregation
    uint32 ack_batch_window_ms = Y;      // Time window to collect ACKs (default: 200ms)
}
```

### Trade-offs to Consider

| Consideration | Mitigation |
|---------------|------------|
| **Increased ACK latency** (100-500ms) | Acceptable for most use cases; urgent ACKs bypass batching |
| **If batch packet is lost, all ACKs are lost** | Senders will retransmit, triggering new ACKs |
| **Additional memory for ACK queue** | Minimal—stores only packet IDs, not full packets |
| **Complexity** | Moderate—isolated to `RoutingModule` and `ReliableRouter` |

### Use Cases That Benefit Most

- **Store & Forward servers** processing many queued messages
- **Dense deployments** with 20+ nodes in range
- **Telemetry-heavy meshes** with frequent position/sensor broadcasts
- **Events or gatherings** where many users are actively messaging

### Related Work

This is similar to TCP's delayed ACK mechanism (RFC 1122) which batches acknowledgments to reduce overhead, and WiFi block acknowledgments (802.11e).

krisclarkdev:
## Summary

Fixes #9187

This PR adds ACK batching to reduce airtime consumption in busy meshes. Instead of sending individual ACK packets for each received message, ACKs are queued and sent as a single combined packet.

## Problem

In busy meshes, ACK traffic can consume significant airtime. Each ACK requires a separate transmission with full packet overhead, even though ACKs are small.

## Solution

Queue ACKs per-destination and flush them as a single batched packet after:
- 200ms timeout (configurable via `DEFAULT_BATCH_WINDOW_MS`)
- 8 ACKs accumulated for the same destination

Each batched ACK uses only 5 bytes (4-byte packet ID + 1-byte error code), significantly reducing overhead compared to individual ACK packets.

## Features

- **Disabled by default** for backwards compatibility
- Two ways to enable:

### Build-time Enable

Add to `platformio.ini`:
```ini
build_flags = 
  -DACK_BATCHING_ENABLED=1
```

### Runtime Enable

Send a direct message to yourself with one of these commands:
- `!ackbatch on` - Enable ACK batching
- `!ackbatch off` - Disable ACK batching  
- `!ackbatch` - Show current status

The response appears in the serial log.

## Testing

- ✅ Builds clean for `heltec-v3` and `heltec-v4`
- ✅ Tested on Linux (native build)
- ✅ Tested on Heltec V4 hardware
- ✅ Unit tests included (`test/test_ackbatcher/`)

## Backwards Compatibility

- Disabled by default - no behavior change unless explicitly enabled
- Nodes without this feature will ignore batched ACK packets (unrecognized payload format)
- Batched ACKs use the existing `ROUTING_APP` portnum with a magic byte prefix (`0xBA`) for identification

## Files Changed

- `src/mesh/AckBatcher.h` / `.cpp` - Core batching logic
- `src/mesh/NextHopRouter.h` - Periodic flush timer
- `src/mesh/ReliableRouter.cpp` - Batched ACK detection and processing
- `src/modules/RoutingModule.cpp` - Queue ACKs instead of immediate send
- `src/modules/Modules.cpp` - AckBatcher initialization
- `src/modules/TextMessageModule.cpp` - Runtime toggle command
- `test/test_ackbatcher/test_main.cpp` - Unit tests

### DIFF
--- a/test/test_ackbatcher/test_main.cpp
+++ b/test/test_ackbatcher/test_main.cpp
@@ -90,7 +90,7 @@ void test_parseBatchedAck_MultipleEntries(void) {
       0x02, 0x00, 0x00, 0x00, // PacketId = 2
       0x00,                   // Error = NONE
       0x03, 0x00, 0x00, 0x00, // PacketId = 3
-      0x01,                   // Error = TOO_LARGE (1)
+      0x07,                   // Error = TOO_LARGE (7)
   };
   memcpy(p.decoded.payload.bytes, payload, sizeof(payload));
   p.decoded.payload.size = sizeof(payload);


GUVWAF:
Thanks for the PR. 

Have you done analysis on how often this will actually be used? 200ms for the default flush timeout sounds *very* short. A single LoRa packet on LongFast preset is already quite some longer than that.

Besides, it would be better to use protbuf encoding for the block ACK.

krisclarkdev:
@GUVWAF the time was just an arbitrary number, I figured with some testing and user feedback that we could find a good default.  I also think that this feature if implemented should probably be left option as an admin command since it would only be beneficial in large meshes so it'd be good to have the option.

GUVWAF:
With which LoRa preset have you been testing this?

For this to be effective, for LongFast we will be talking about seconds of delay. I'm not sure that's worth it, as multiple packets to the same node quickly after each other is not common, and you'll have to wait longer until you get an acknowledgment.

caveman99:
Extensive behavioural change, not backward compatibble, not tested in real life, academic use case, likely AI generated.

PR #9196, opened 2026-01-06
Morse Code Input for Single-Button Devices

sjoerdvanderhoorn:
This PR adds morse code input with live hints to devices that only have one button available to the user, such as the  LILYGO T3S3 and Heltec V3. This enables these devices to not only receive, but also send messages over the Meshtastic network without the need for pairing them to a device like a phone or computer.

Video of the feature in use:

https://github.com/user-attachments/assets/38c57717-4f19-425f-8eaf-f4f56eea2d98

Alternate input method (grid keyboard):

https://github.com/user-attachments/assets/6dc1c174-1cf5-4e21-8628-d9438b253b80

### User Experience (UX)
The input screen is designed to guide the user through morse code interactively:
- **Visual Feedback:** Displays a breakdown of available characters (`A-Z`, `0-9`, punctuation).
- **Live Filtering:** As the user inputs dots/dashes, the character grid hides invalid options.
- **Live Hints:** Remaining characters display a dot or dash above them indicating the next required stroke:
- **Timing Logic:**
  - **Dot:** Short Press (<300ms)
  - **Dash:** Long Press (>300ms)
  - **Commit Current Character:** Idle > 1000ms
  - **Auto-Space:** Idle > 3000ms
- **Corrections:** Sending a keystroke of 8 dots will remove the last entered character, making it easy to correct mistakes.
  
```
┌────────────────────────────────────────┐
│ Hello_                                 │ ← Input Text with blinking cursor
│ ────────────────────────────────────── │ ← Divider Line
│ .--- .... ..-. ---. --.-               │ ← Row 1 Hints (Dots/Dashes above chars)
│ ABCD EFGH IJKL MNOP QRST               │ ← Row 1 Characters
│                                        │
│ ...- --- --.. ----- .... -...          │ ← Row 2 Hints
│ UVW XYZ ,.? 0123 456 789               │ ← Row 2 Characters
└────────────────────────────────────────┘
```

### Menu System
Accessible via a **Long Press (> 2s)** from the morse code input screen:

*   **Morse Menu**
    *   **Back**: Return to typing.
    *   **Backspace**: Delete last character.
    *   **Shift**: Toggle capitalization (Auto-enables after `.`, `!`, `?`).
    *   **Char Picker**: Access special characters (`' ! / - ( ) & : ; = + " - _ $ @`).
    *   **Send**: Transmit message.
    *   **Exit**: Cancel and return to main screen.

### Integration
This feature reuses the existing **New Freetext Msg** menu option.
- On devices without a physical keyboard (detected via `kb_found`), selecting "New Freetext Msg" will now launches the Morse Input interface instead of the standard text entry screen.
- This ensures the option is discoverable on single-button devices where `BUTTON_PIN` is defined but no keyboard is attached.
- Devices with a physical keyboard will not launch the Morse Input interface but keep using the physical keyboard instead.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] LILYGO T3S3


BurdMahn:
awesome

caveman99 (changes requested):
What's the function of all the different input method files under modules apart from single button input? Architecture wise this belongs in the input broker, not in moduls.

sjoerdvanderhoorn:
Thanks @caveman99. The other files are additional input methode that I committed to the same branch this PR is in, not realizing they would appear here. I will try and provide a video of them in action soon. In short, the Grid Keyboard provides a full on screen display where the user can toggle through using the single button. Even though the Morse code implementation allows for faster typing, the Grid Keyboard may feel more familiar to some users.

Can you please at a high level lay out how this would need to be added into the input broker, and if you think it has a chance of being merged into the main branch at all before I spend more time at this?

mcastillof:
This is great!!

sjoerdvanderhoorn:
> What's the function of all the different input method files under modules apart from single button input? Architecture wise this belongs in the input broker, not in moduls.

@caveman99 I refactored the code as per requested. I also attached below video displaying of the other Grid Keyboard input method. The Special Character input method is similar to it, but just shows a different charset.


https://github.com/user-attachments/assets/6dc1c174-1cf5-4e21-8628-d9438b253b80

korbinianbauer:
Cool!

I wonder if the morse input could be combined with sound output on devices that have speaker. Both for feedback when typing as well as audible morse code as an output option for incoming messages.

Ultimately I'd like to be able to communicate without a screen at all, just one button and a speaker/earbud, on a preselected channel.

sjoerdvanderhoorn:
> I wonder if the morse input could be combined with sound output on devices that have speaker. Both for feedback when typing as well as audible morse code as an output option for incoming messages.

For devices that have speakers, you might as well include (rudimentary) TTS capabilities like [SAM](https://discordier.github.io/sam/) so the message is played to you out loud. I bet for input, you could also make a basic voice recognition system that uses the NATO alphabet (where you would press the button, say the code word, and it would determine which of 26 options it most likely is) or re-implement something like [Harpy](https://iiif.library.cmu.edu/file/Newell_box00092_fld06359_doc0001/Newell_box00092_fld06359_doc0001.pdf).

3geETR:
https://github.com/user-attachments/assets/d0359ab8-0c39-4bbe-9632-e01707ef0d31

My device is Faketec v5, and it works great—I just have a few problems:
- For those who don't know Morse code, putting a space every 3 seconds is a bit of a hassle; it constantly puts spaces. If there is a space in Morse code, it can be used instead. Or we should be able to adjust the duration.
- When I switch to Grid mode and restart the device, it switches back to Morse mode

emojisrDEV:
Me and a dude are making a custom Key for this with custom software and circuit boards so this would be a massive help if we can fork it 🤞here are the two discords we use: https://discord.gg/wbZWDzu5x https://discord.gg/F23mnytRUv

sjoerdvanderhoorn:
> For those who don't know Morse code, putting a space every 3 seconds is a bit of a hassle; it constantly puts spaces. If there is a space in Morse code, it can be used instead. Or we should be able to adjust the _duration._

A space in Morse code is defined as seven empty units. The current implementation is aimed at beginners and thus a standard wait time applies. A next version could be made adaptive, where we keep track of the timing between dots and dashes and calculate how long the space should take (plus the pause between two symbols).

seafloor507:
is there a way to make this work with the wio tracker l1 pro? i would love morse code on that

3geETR:
> is there a way to make this work with the wio tracker l1 pro? i would love morse code on that

https://github.com/3geETR/firmware/actions/runs/25256342848
I built it, try it maybe it will work

PR #9204, opened 2026-01-07
Add support for VDDHDIV5 voltage measurement

brad112358:
Allow a NRF52 variant, like the DIY Promicro NRF52 development boards which power the NRF52 directly via the VDDH pin to obtain battery voltage directly from this pin using VDDHDIV5.  This can simplify DIY builds by avoiding the need to connect voltage divider resistors to an analog pin.  This method also avoids the small current drain of the voltage divider.  Note however, that a tradeoff in some cases is that this method may not allow monitoring battery voltage while charging via USB since some designs, including the Promicro boards, supply USB voltage to VDDH when USB is connected.

Note that there are a couple of NRF52 variants which are already defining BATTERY_PIN to -1; presumably to indicate the absence of a battery sense pin.  It is not immediately obvious to me if these variants might happen to have battery voltage connected to VDDH and would thus benefit from this feature, but I don't expect these variants to suffer any issues do to this change.

If, however, it is preferred to have a new #DEFINE rather than overload BATTERY_PIN, please suggest a new symbol name, and I'll update this PR.

I would like to take advantage of this in the future to simplify my "Easy E22" build instructions: https://github.com/brad112358/easy_E22 

See also #7579 

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  Tested on DIY NRF52 Promicro E22 board

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -331,16 +331,24 @@ class AnalogBatteryLevel : public HasBatteryLevel
             raw = espAdcRead();
             scaled = esp_adc_cal_raw_to_voltage(raw, adc_characs);
             scaled *= operativeAdcMultiplier;
-#else // block for all other platforms
+#else                 // block for all other platforms
+#if BATTERY_PIN != -1 // Skip if pin not used
             for (uint32_t i = 0; i < BATTERY_SENSE_SAMPLES; i++) {
-#if (BATTERY_PIN == -1)
-                raw += analogReadVDDHDIV5();
-#else
                 raw += analogRead(BATTERY_PIN);
-#endif
             }
             raw = raw / BATTERY_SENSE_SAMPLES;
             scaled = operativeAdcMultiplier * ((1000 * AREF_VOLTAGE) / pow(2, BATTERY_SENSE_RESOLUTION_BITS)) * raw;
+#endif
+#if defined(USE_VDDHDIV5) and defined(ARCH_NRF52)
+            if (scaled <
+                1000) { // Battery pin not used or not connected to voltage divider, fallback to reading voltage of VDDH pin
+                for (uint32_t i = 0; i < BATTERY_SENSE_SAMPLES; i++) {
+                    raw += analogReadVDDHDIV5();
+                }
+                raw = raw / BATTERY_SENSE_SAMPLES;
+                scaled = 5 * ((1000 * AREF_VOLTAGE) / pow(2, BATTERY_SENSE_RESOLUTION_BITS)) * raw;
+            }
+#endif
 #endif
             battery_adcDisable();
 
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -333,7 +333,11 @@ class AnalogBatteryLevel : public HasBatteryLevel
             scaled *= operativeAdcMultiplier;
 #else // block for all other platforms
             for (uint32_t i = 0; i < BATTERY_SENSE_SAMPLES; i++) {
+#if (BATTERY_PIN == -1)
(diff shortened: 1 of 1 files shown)


phaseloop:
The reading is divided by 5 so I think you need to use different OPERATIVE_ADC_MULTIPLIER (5x) instead of the one configured for voltage divider. I assume with current setting your readings are like 30-50% lower than the actual battery state, unless you manually set ADC to 5.0 in device settings.

brad112358:
> The reading is divided by 5 so I think you need to use different OPERATIVE_ADC_MULTIPLIER (5x) instead of the one configured for voltage divider. I assume with current setting your readings are like 30-50% lower than the actual battery state, unless you manually set ADC to 5.0 in device settings.

Yes, of course, instead of #define ADC_MULTIPLIER  to a value based on the specific voltage divider resistor values, when BATTERY_PIN is set to -1, ADC_MULTIPLIER should be set to 5 in variant.h or in device settings.

brad112358:
I suspect it is fairly common to supply battery voltage to the VDDH pin when a single LiPo or LiIon cell is used in a design.
For DIY builds, especially, it might be handy to offer VDDHDIV5 as a fall-back option for cases where the designer or builder chooses not to include voltage divider resistors.

I think I will I'll re-work this PR

brad112358:
I have re-worked this PR to allow, but not require, automatic run-time choice of voltage measurement via a BATTERY_PIN or via VDDHDIV5.

It works like this:
- If USE_VDDHDIV5 is not defined and BATTERY_PIN is defined and not -1, everything works as before; i.e., the voltage is determined via analogRead() based on the specified multiplier, etc.
- If BATTERY_PIN is defined to be -1, the analogRead() loop is skipped, since analogRead(-1) would have returned 0 anyway.  This just saves a little time and code for variants that currently define BATTERY_PIN to -1 for whatever reason.
- For NRF52 boards, if USE_VDDHDIV5 is defined and the analogRead() loop was skipped or measured less than 1 volt, analogReadVDDHDIV5() is used to obtain a voltage reading from the VDDH pin.

This allows a variant to choose to use either an analog pin or the VDDH pin to sense the battery voltage and also allows a variant to use a single firmware with fallback to use of the VDDH pin if the BATTERY_PIN is connected to ground.  In practice, I have found that, a floating BATTERY_PIN is also likely to result in a measurement of less than 1 volt, but obviously, the recommendation is to connect the BATTERY_PIN to ground on DIY builds when voltage divider resistors are not installed.

phaseloop:
Makes sense - I can test that later on promicro. Some variants use external 3.3V LDO regulator connected to VDD+VDDH in "low voltage mode" as Nordic coins that. I use similar VDD direct reading to protect device from undervoltage: https://github.com/meshtastic/firmware/pull/9211

BTW - I think it will be better to actually force using ADC_MULTIPLIER=5 when using VDDH reading instead of relying on any other compiled or configured value to be 5 to avoid problems or misconfigurations. Especially if we are in the territory of automatic fallbacks :)

brad112358:
> BTW - I think it will be better to actually force using ADC_MULTIPLIER=5 when using VDDH reading instead of relying on any other compiled or configured value to be 5 to avoid problems or misconfigurations. Especially if we are in the territory of automatic fallbacks :)

Yes, and the current PR already does exactly that by just hard coding 5.

brad112358:
@phaseloop were you able to test this?

brad112358:
What can I do to help get this merged?

PR #9213, opened 2026-01-07
Feature T-Deck Pro and T-Deck: add Chinese IME and built-in GB2312 pinyin table

vicliu624:
## Summary
Enable Chinese display and IME on T-Deck (LCD) and refine IME behavior/UX for T-Deck Pro.

## Changes
- Enable U8g2 text rendering on T-Deck (LCD) and use GB2312-capable font for Chinese display.
- Add/enable Chinese IME for T-Deck (alongside T-Deck Pro) and keep IME disabled by default until CN mode is selected.
- Gate IME paging to active candidate sessions; reset/disable IME on exit from freetext.
- Adjust IME candidate bar position upward to avoid footer overlap.
- Insert space directly when CN mode is active but IME buffer is empty.
- Harden UTF‑8 apostrophe normalization with bounds checks.
- Pin U8g2_for_Adafruit_GFX to an immutable commit SHA (both t-deck / t-deck-pro).

## Testing
- [x] Built on T-Deck (ESP32-S3)
- [x] Built on T-Deck Pro (ESP32-S3)

## Notes
- IME is only active in CN mode; other modes remain EN/123.
- IME buffer is cleared and disabled when leaving freetext.

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

**Other:**
- LilyGo T-Deck Pro (ESP32-S3)

## IME usage (T-Deck vs T-Deck Pro)
- T-Deck Pro: Press sym in any mode to enter the IME; use space to cycle candidates and Enter to confirm. The IME buffer resets when leaving the freetext screen.
- T-Deck: Short-press the joystick to enter/switch the IME; in CN mode, move the joystick to select a candidate and press to confirm. The IME buffer also resets on exit from freetext.

35D0C1A1C40FB36A8FBB7EB6EE3FF546
62BFB47C9909DEA661562411297161CE


src/graphics/draw/MessageRenderer.cpp
```diff
@@ -36,6 +44,86 @@ static std::vector<std::string> cachedLines;
...
+{
+    return getTextWidth(display, std::string(text));
+}
+
+static inline void drawText(OLEDDisplay *display, int x, int y, const std::string &text)
+{
+#if defined(USE_U8G2_EINK_TEXT)
+    if (auto *u8g2 = getU8g2Fonts(display)) {
+        const int baseline = y + FONT_HEIGHT_SMALL - 1;
+        u8g2->drawUTF8(x, baseline, text.c_str());
```
Xaositek, line 120: In testing this it appears the font or font size has changed? Message fonts are much smaller now.

Image

vicliu624: Thanks for pointing this out — I’ve made an adjustment to address the issue you observed.

I’ve aligned the message text with the time / node info by adjusting how the U8g2 baseline is calculated in MessageRenderer / CannedMessageModule. Instead of applying a manual offset, the baseline now uses the font’s actual ascent. This ensures that the rendered text height matches the original OLED font’s line height, so the message text no longer appears smaller than other UI elements.

Regarding the earlier choice of u8g2_font_wqy12_t_gb2312, that was originally done to minimize Flash usage. However, after adjusting the baseline calculation, I’ve switched to u8g2_font_wqy16_t_gb2312, which now renders at a size that is visually consistent with the previous behavior and matches the time / node info as intended.

PR #8678, opened 2025-11-19
Adding support for InkHUD joystick navigation for the Seeed Wio Tracker L1 E-ink

zeropt:
Made additions to InkHUD for navigating with the joystick on the Seeed Wio Tracker L1 E-ink

Key changes:

- Modified `seeed_wio_tracker_L1_eink/nicheGraphics.h` to enable joystick navigation and connect joystick events to InkHUD handlers
- `TwoButtonExtended` mirrors the functionality `TwoButton` but with additional joystick specific functions
- Up, down, left, right, and center handlers added to `InkHUD` with matching events in `Events`
- Add `prevApplet` and `prevTile` to `WindowManager`
- Added persistent settings for enabling the joystick and rotating it relative to the screen
- Declare additional input handlers in `Applet` and handle joystick inputs in `TipsApplet`, `NotificationApplet`, and `Menu Applet`
- Add `AlignStickApplet` for aligning the joystick to the screen (accessible through the menu)
- `MenuApplet` stores previous page data for navigating backwards through the menu
- Update the Tips Applet to describe basic joystick navigation and show the Align Stick Applet after if the joystick is enabled and not aligned

Align Joystick Applet:

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [x] Seeed Wio Tracker L1 E-ink (with joystick disabled)
  - [x] Seeed Wio Tracker L1 E-ink (with joystick enabled)
  - [ ] Other InkHUD capable devices

We have only tested our changes on the Seeed Wio Tracker L1 E-ink by setting the `settings->joystick.enable` to true and false to confirm that navigation with and without a joystick still works properly.

PR #9227, opened 2026-01-09
Change canned message recipient's previous page to send page

scobert969:
Fixes a minor bug in #8678. Pressing the back button on the recipients page would lead to the options page. Now leads to send page.

Update 1/29/2026
This PR has been updated to change the way `navLeft()` works to function with the new menu pages in #8975. When the Back `MenuItem` is executed, it opens its `nextPage`, which is set to `previousPage` or when pressing the left joystick button it calls `showPage()` with `previousPage` directly.


scobert969:
Updated the back menu actions for the new menu pages introduced in #8975

HarukiToreda:
Looks good. Could you update the first comment to clearly describe what this PR does now? Since the original purpose has changed from just the back button(Left toggle of Joystick) to a clean up of how the menu remembers where you came from across all lists, it would be helpful for people with less coding experience to understand what was modified and why. 

image

PR #9228, opened 2026-01-09
New extra variant: Xiao ESP32c3

vidplace7:
Adds new variant(s) for the [Seeed Xiao ESP32-C3](https://wiki.seeedstudio.com/XIAO_ESP32C3_Getting_Started/).

This MCU is not sold in a bundle with a LoRa radio.
Definitions have been provided to build a firmware that is compatible with either the "ESP32-S3" Xiao LoRa radio or the "NRF52840" Xiao LoRa radio (they use different pinouts).


NomDeTom:
Oops. Should have a look at this at some point!

PR #9231, opened 2026-01-09
Meshtastic unified OTA

thebentern:
@jake-b's work. I'm just PRing it. :)


ChrisHi0:
Did you test if it also work with a Heltec V4? Since the new firmeware update  to version 2.7.18 it is not possible with the command meshtastic --reboot-ota. Does anyone know how else to do it? I don't know where else to ask to help with this problem.

PR #9234, opened 2026-01-09
Fix TFT_MESH settings across setting and recalling

Xaositek:
- Fix TFT_MESH settings across setting and recalling
- Log at boot if we used user defined, TFT_MESH_OVERRIDE, or default
- Actually set the correct colors


HarukiToreda:
tested on the T114, colors are properly saved.

PR #8866, opened 2025-12-05
Xiao NRF - define suitable i2c pins for the sub-variants

NomDeTom:
This PR is a tidying up and extending of the defaults for the Xiao NRF, to allow each of the variations in use to coexist harmoniously whilst also sharing a single variant.h. Pins are assigned as follows:

| Pin   | Default  | I2C  | BTB  | BLE-L |     | Pin   | Default | I2C  | BTB  | BLE-L |
| ----- | -------- | ---- | ---- | ----- | --- | ----- | ------- | ---- | ---- | ----- |
|       |          |      |      |       |     |       |         |      |      |       |
| D0    |          | UBTN | DIO1 | CS    |     | 5v    |         |      |      |       |
| D1    | DIO1     | DIO1 | Busy | DIO1  |     | GND   |         |      |      |       |
| D2    | NRST     | NRST | NRST | Busy  |     | 3v3   |         |      |      |       |
| D3    | Busy     | Busy | CS   | NRST  |     | D10   | MOSI    | MOSI | MOSI | MOSI  |
| D4    | CS       | CS   | RXEN | SDA   |     | D9    | MISO    | MISO | MISO | MISO  |
| D5    | RXEN     | RXEN | UBTN | SCL   |     | D8    | SCK     | SCK  | SCK  | SCK   |
| D6    | G_TX     | SDA  | G_TX |       |     | D7    | G_RX    | SCL  | G_RX | RXEN  |
|       |          |      |      |       |     |       |         |      |      |       |
|       | End      |      |      |       |     |       |         |      |      |       |
| NFC1/ | SDA      | G_TX | SDA  | G_TX  |     | NFC2/ | SCL     | G_RX | SCL  | G_RX  |
| D30   |          |      |      |       |     | D31   |         |      |      |       |
|       |          |      |      |       |     |       |         |      |      |       |
|       | Internal |      |      |       |     |       |         |      |      |       |
| D16   | SCL1     | SCL1 | SCL1 | SCL1  |     |       |         |      |      |       |
| D17   | SDA1     | SDA1 | SDA1 | SDA1  |     |       |         |      |      |       |

Basically, Default and I2C variants just swap whether GPS or I2C gets to sit on "real" pins, vs the NFC pads at the back.

The Board-to-board connector version of the WIO radio board is assumed to follow the default pinout apart from that, and BLE legacy just shoves them on the back, because it can.

[...]

PR #9238, opened 2026-01-09
Change I2C pins to NFC pads for XIAO nRF52840

wehooper4:
Change the default I2C pins to match the Seeed Solar Node, allowing use of I2C sensors in DIY nodes based on the Seeed NRF52840 kit

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
 Seeed NRF52840 Kit

### DIFF
--- a/variants/nrf52840/seeed_xiao_nrf52840_kit/variant.h
+++ b/variants/nrf52840/seeed_xiao_nrf52840_kit/variant.h
@@ -189,8 +189,8 @@ static const uint8_t SCK = PIN_SPI_SCK;
 #define PIN_WIRE_SCL D7
 #else
 // Use NFC pads for I2C as done on Seeed Solar Node 
-#define PIN_WIRE_SDA D14 // P0.09
-#define PIN_WIRE_SCL D15 // P0.10
+#define PIN_WIRE_SDA (9) // P0.09
+#define PIN_WIRE_SCL (10) // P0.10
 #endif
 
 static const uint8_t SDA = PIN_WIRE_SDA;


lbibass:
I'm not sure why, but changing the i2c pins as done in this patch caused my solarxiao to not run correctly. reverting the patch allowed it to boot again. that was the only change made. it would hang on checking the i2c bus.

variants/nrf52840/seeed_xiao_nrf52840_kit/variant.h
```diff
@@ -188,9 +188,9 @@ static const uint8_t SCK = PIN_SPI_SCK;
 #define PIN_WIRE_SDA D6
 #define PIN_WIRE_SCL D7
 #else
-// Internal LSM6DS3TR on XIAO nRF52840 Series
-#define PIN_WIRE_SDA (17)
-#define PIN_WIRE_SCL (16)
+// Use NFC pads for I2C as done on Seeed Solar Node 
+#define PIN_WIRE_SDA (9) // P0.09
+#define PIN_WIRE_SCL (10) // P0.10
```
cwalther, line 193: I have not tried it yet, but these numbers look wrong to me. Does this work for you? According to https://github.com/meshtastic/firmware/blob/a092f6bb22f6fae223acf50e3306a9063f790fbd/variants/nrf52840/seeed_xiao_nrf52840_kit/variant.cpp#L54-L55 it should be 30 and 31. See also [PR #8866](https://github.com/meshtastic/firmware/pull/8866/changes#diff-29d6644e1e9057ffcb2ac6700e32926aee16e2c5a7ce71039a65626f141d9d9aR221-R224).

cwalther: I have now tested it with my Xiao nRF52840 Sense and an SSD1306 display: It indeed works with 30 (NFC1, closer to center of board) and 31 (NFC2, closer to pin header), but not with 9 and 10 (with the same symptom as @lbibass).

(However it turns out I still can‘t use this variant because it leaves me without a `BUTTON_PIN`.)

cwalther:
I believe this is now obsoleted by #8866.

PR #9258, opened 2026-01-11
CI: Unified ESP32 OTA firmware

vidplace7:
Add new [esp32-unified-ota](https://github.com/meshtastic/esp32-unified-ota) by @jake-b to the firmware packaging process. Fall-back to BT-Only firmware on ESP32-C-series.


vidplace7:
This still isn't being properly included in the manifest (just like bleota.bin before it)

I'll submit a followup PR to adjust platformio extrascripts to include these in the manifest properly.

PR #9326, opened 2026-01-15
Fix uMesh RF POWER configuration error

linser233:
Fix uMesh RF POWER configuration error
(see https://github.com/linser233/uMesh/blob/main/RF_Power.md)


fifieldt:
Thanks. Since this uses a power amp, we should also set a TX_GAIN_LORA array in configuration.h , based on the charts at https://github.com/linser233/uMesh/blob/main/RF_Power.png

linser233:
> Thanks. Since this uses a power amp, we should also set a TX_GAIN_LORA array in configuration.h , based on the charts at https://github.com/linser233/uMesh/blob/main/RF_Power.png

I have checked the relevant `TX_GAIN_LORA` config and it does not seem to fit the formula. 

For the 30 dBm SKU:  
`Pout ≈ 11.75 + 0.85 * Pset`  
For the 33 dBm SKU:  
`Pout ≈ 21 + 0.55 * Pset`  

The output power cannot simply be treated as a linear offset. In this case, is there a better way to handle it rather than just using `TX_GAIN_LORA 10`?

fifieldt:
@linser233 , see eg : https://github.com/meshtastic/firmware/blob/d493f5f171b50b1f75062ed052e8cd3416b0c226/src/configuration.h#L149

for how to make an array of gains

(first value is gain added when 1dBm does into the SX1262, second value is gain added when 2dBm into the SX1262, ... 22nd value is gain added when 22dBm into the SX1262)

linser233:
> @linser233 , see eg :
> 
> https://github.com/meshtastic/firmware/blob/d493f5f171b50b1f75062ed052e8cd3416b0c226/src/configuration.h#L149
> 
> for how to make an array of gains
> 
> (first value is gain added when 1dBm does into the SX1262, second value is gain added when 2dBm into the SX1262, ... 22nd value is gain added when 22dBm into the SX1262)

In that case, I would need four configuration files — 1262 and 1268, each with 30S and 33S. Would it be better to place them in a separate directory?

fifieldt:
> > @linser233 , see eg :
> > https://github.com/meshtastic/firmware/blob/d493f5f171b50b1f75062ed052e8cd3416b0c226/src/configuration.h#L149
> > 
> > for how to make an array of gains
> > (first value is gain added when 1dBm does into the SX1262, second value is gain added when 2dBm into the SX1262, ... 22nd value is gain added when 22dBm into the SX1262)
> 
> In that case, I would need four configuration files — 1262 and 1268, each with 30S and 33S. Would it be better to place them in a separate directory?

Since the values are relatively close (at least for the different frequencies with the same device), maybe just take an average or a maximum to keep it simple ...

linser233:
> > > @linser233 , see eg :
> > > https://github.com/meshtastic/firmware/blob/d493f5f171b50b1f75062ed052e8cd3416b0c226/src/configuration.h#L149
> > > 
> > > for how to make an array of gains
> > > (first value is gain added when 1dBm does into the SX1262, second value is gain added when 2dBm into the SX1262, ... 22nd value is gain added when 22dBm into the SX1262)
> > 
> > 
> > In that case, I would need four configuration files — 1262 and 1268, each with 30S and 33S. Would it be better to place them in a separate directory?
> 
> Since the values are relatively close (at least for the different frequencies with the same device), maybe just take an average or a maximum to keep it simple ...

What I mean is that my module is available in four SKUs: the SX1268-based versions rated at 30 dBm and 33 dBm, and the SX1262-based version.😂

linser233:
@fifieldt I’ve considered that the 33 dBm SKU is unlikely to be produced in significant numbers, so it may be more practical to simply rename the current YAML files to reflect the 30 dBm SKU.

linser233:
uMesh’s prototype validation has been tested and is functioning correctly with the new configuration file.
f16e6b7c587808938f56a218de34ee49 (2)

fifieldt (approved):

PR #9327, opened 2026-01-15
Meshtastic OTA (moar)

thebentern:


ChrisHi0:
Can someone tell me if the ota update works on a heltec v4? Because when I do the command meshtastic --reboot-ota, nothing happens. Sorry for that may stupid question

slomkowski:
Hello,

@thebentern since you made changes in the OTA subsystem, I think I can ask you.

I'm writing OTA firmware that uses Ethernet instead of Wi-Fi (for the WT32-ETH01 module). Is it a good idea to define `METHOD_OTA_ETHERNET` and push that upstream, or should I stick with `METHOD_OTA_WIFI` for Ethernet OTA updates?

I don't want to interfere with your work on combined BLE/Wi-Fi OTA, but I'm planning to take over https://github.com/meshtastic/firmware/pull/9586, and perhaps adding Ethernet to the combined OTA might be a good idea.

PR #9332, opened 2026-01-16
Add interrupt for external charge detection

jp-bennett:
Tested on Thinknode m4, m6, and T1000-e

We already have an interrupt for EXT_PWR_DETECT, this adds an interrupt for the charge detect pin as well.


fifieldt (approved):

DGxInfinitY:
Hey wondering how this is used. I'd like to have a way to see if the solar panel is meeting charging voltage on my t114 and I noticed that when it was I could physically see the charge led light up. I'd like for a simple metric that can be viewed over the mesh like a charging: yes or no when requesting telemetry. I can't visibility see the node to know if it's charging just guessing based on voltage. The problem is that when you have a large battery it's challenging to see the voltage increase substantially. It would be nice to know if the panel is achieving charging because then you might detect failing batteries and other power problems sooner before having a node go down.

PR #9346, opened 2026-01-18
Add IKOKA STICK 0.3.0 variant (SX1268/E22-400M33S)

PhilipLykov:
## Summary
Add support for IKOKA STICK 0.3.0 board with EBYTE E22-400M33S (SX1268) module.

## Hardware
- **Board:** IKOKA STICK 0.3.0 (Seeed XIAO nRF52840)
- **LoRa:** EBYTE E22-400M33S (SX1268 chip)
- **Frequency:** 433/470 MHz
- **Max Power:** 33 dBm with internal PA
- **Display:** SSD1306 OLED

## Why Needed
Existing `seeed_xiao_nrf52840_kit` is for SX1262 (868/915 MHz). 
This variant supports SX1268 (433/470 MHz) with:
- Different chip configuration (USE_SX1268)
- TCXO voltage (2.2V vs 1.8V)
- RF switch (DIO2 + RXEN)
- PA configuration (12 dB gain, virtual gain scaling)

## Changes
- New variant: `variants/nrf52840/ikoka_stick_0_3_0/`
- Modified: `src/configuration.h` (added E22-400M33S power config)

## Testing
✅ Radio init successful, TX/RX working, PA functional (0-33 dBm)
✅ Display and button working
✅ Comprehensive documentation included

## Safety Notes
⚠️ 33 dBm may require amateur radio license
⚠️ Regional compliance necessary


PhilipLykov:
@CLAassistant check

NomDeTom:
Does this need to be a separate variant?

PhilipLykov:
> Does this need to be a separate variant?

It does at least to have workable I2C (display connector) and user button. But main changes are in the support of EBYTE E22-400M33S which requires special TCXO voltage of 2.2 and power calculation and limitation (internal PA will blow out if you supply more than 21dBm to it)

src/configuration.h
```diff
@@ -124,6 +124,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define SX126X_MAX_POWER 8
 #endif
 
+#ifdef EBYTE_E22_400M33S
+// E22-400M33S: 433/470MHz variant - per RF output power curve from E22-M Series manual
+// Actual PA gain: 12 dB (21 dBm input → 33 dBm output from curve)
+// Virtual gain: 9 dB - scales app's 0-30 dBm range to module's 0-33 dBm range
+// This makes: App 30dBm = SX1268 21dBm = Module 33dBm (full power)
+#define TX_GAIN_LORA 9
```
fifieldt, line 137: Is this device's gain linear across the entire input range?

Have a look down at USE_GC1109_PA for an example of how to handle devices with non-linear power amps.

PhilipLykov: Conform my testing it's quite linear but I don't have equipment better then LibraVNA.

NomDeTom:
> > Does this need to be a separate variant?
> 
> It does at least to have workable I2C (display connector) and user button. But main changes are in the support of EBYTE E22-400M33S which requires special TCXO voltage of 2.2 and power calculation and limitation (internal PA will blow out if you supply more than 21dBm to it)

Can you link the docs where that TCXO voltage is stated? The datasheet doesn't mention it.

PhilipLykov:
> > > Does this need to be a separate variant?
> > 
> > 
> > It does at least to have workable I2C (display connector) and user button. But main changes are in the support of EBYTE E22-400M33S which requires special TCXO voltage of 2.2 and power calculation and limitation (internal PA will blow out if you supply more than 21dBm to it)
> 
> Can you link the docs where that TCXO voltage is stated? The datasheet doesn't mention it.
[E22-M+Series+Module_UserManual_EN.pdf](https://github.com/user-attachments/files/24839664/E22-M%2BSeries%2BModule_UserManual_EN.pdf)
Here is a PDF received from EBYTE tech support. Please pay attention that E22-900M33S has different PA parameters and allows lower power from SX126x chip. TCXO voltage is the same for both - 2.2v

NomDeTom:
> [E22-M+Series+Module_UserManual_EN.pdf](https://github.com/user-attachments/files/24839664/E22-M%2BSeries%2BModule_UserManual_EN.pdf)
> Here is a PDF received from EBYTE tech support. Please pay attention that E22-900M33S has different PA parameters and allows lower power from SX126x chip. TCXO voltage is the same for both - 2.2v

Hmmmm. We're using 1.8v for all of the 900mhz ones, and not generally seeing any consequences from that. Have you confirmed that the 400mhz variant doesn't init without the higher voltage?

PhilipLykov:
> > [E22-M+Series+Module_UserManual_EN.pdf](https://github.com/user-attachments/files/24839664/E22-M%2BSeries%2BModule_UserManual_EN.pdf)
> > Here is a PDF received from EBYTE tech support. Please pay attention that E22-900M33S has different PA parameters and allows lower power from SX126x chip. TCXO voltage is the same for both - 2.2v
> 
> Hmmmm. We're using 1.8v for all of the 900mhz ones, and not generally seeing any consequences from that. Have you confirmed that the 400mhz variant doesn't init without the higher voltage?

The role of TCXO, as I understand it, is to provide stable frequency within full temperature range. There is possibility that vendor find out that 1.8v is just not enough and increased it to 2.2v in the latest manual. But in any case it will work the question how stable. I don't have necessary lab equipment to measure the output frequency with necessary precision within all temperature range but with 2.2v my LibreVNA shows quite good results (the center frequency is on set frequency).

NomDeTom:
> with 2.2v my LibreVNA shows quite good results (the center frequency is on set frequency).

Can you recheck at 1.8v?

The choices here are to add extra functions, or simply make a sweeping change.

PhilipLykov:
> > with 2.2v my LibreVNA shows quite good results (the center frequency is on set frequency).
> 
> Can you recheck at 1.8v?
> 
> The choices here are to add extra functions, or simply make a sweeping change.

It's not a simple task: we need to check it with full temperature range (-15C - 40C) and in full frequency range (at least: 410-510MHz). My level of equipment (and lack of specific knowledge) doesn't allow me to make such statements as what is better 1.8 or 2.2v in this case. But the vendor gives this info in their document (for all E22-xxxMxxS series), so I suppose this should be a reason to change 1.8v to 2.2v.

PhilipLykov:
I've made a non-lab power test of E22-400M33S on Ikoka Stick with proposed firmware:Set Value     Measured Value
1-	8,10
5-	13,10
10-	18,10
15-	22,60
20-	27,10
21-	28,10
22-	29,10
23-	29,60
24-	30,10
25-	31,10
26-	31,60
27-	32,10
28-	33,10
29-	33,10
30-	33,60
Measurements were made with TinySA (non-calibrated) by single max power peak, averaged by 10 measurements each.

PR #9347, opened 2026-01-18
Noise floor

RCGV1:
I implemented a noise floor calculation by getting the radios current RSSI if its not transmitting nor receiving, this value is then added to a sliding window average which can hold up to 20 measurements. Noise Floor is sent via local stats.
This is extremely useful to determine if a cavity filter is needed for a particular location.

Depends on:
https://github.com/meshtastic/protobufs/pull/846

iOS Interface:
Screenshot 2026-01-17 at 7 26 34 PM

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631


compumike:
Very interesting!

We're not measuring Noise Floor (NF) in the sense of just random (stationary random process) background: thermal, amplifier, etc. 

We're really interested in interference (potentially non-LoRa as well as LoRa, though excluding samples where a receive preamble is already detected by the radio).

**It might be interesting to extract other statistics of these NF samples, not just the average.**

For example: if we also extract the 10th-percentile and 90th-percentile values, this could tell us whether the interference is present all the time (10th and 90th close together), versus occasional (10th and 90th far apart). If they're far apart, it might actually inform us to delay a packet transmit if we measure a `getRSSI(false)` above a certain percentile, so we can attempt to avoid even non-LoRa interference! (Or maybe delay if it's more than X dB above the 10th percentile NF.) That would be pretty cool. 😄

compumike:
Another thought: about how long does the `getRSSI(false)` call take to return? 

Assuming it's quick, have you looked at taking a burst of consecutive samples (say, 10) and averaging them before storing in the array? (Maybe not just averaging the 10 samples as dBm values, but instead converting them to linear power mW values, averaging in mW, and then converting back to dBm.) This might give a cleaner short-term NF measurement which then goes into the longer-term array.

kamcio2603:
any update?

jock78:
Hi can provide Seeed Xiao Esp32, LILYGO T3v1.6.1, Heltec v4 and Heltec Pocket for testing.

src/mesh/RadioLibInterface.cpp
```diff
@@ -255,6 +347,9 @@ currently active.
 */
 void RadioLibInterface::onNotify(uint32_t notification)
 {
+
+    updateNoiseFloor();
```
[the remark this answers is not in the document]
GUVWAF: Copilot has a valid comment here. We shouldn't call this here, because every delay added here can cause the LoRa radio queue to get full (we have to read it out immediately because it can hold only 256 bytes). Doing it after entering receive mode is fine, and a period call e.g. at the end of `Router::runOnce()` or something.

src/mesh/LR11x0Interface.cpp
```diff
@@ -317,4 +317,10 @@ template <typename T> bool LR11x0Interface<T>::sleep()
 
     return true;
 }
+
+// LR11x0 chips do not support getRSSI(bool) for instantaneous RSSI - noise floor measurement is disabled for these chips
```
GUVWAF: LR11x0 chips have the `getRssiInst()` method for this.

michalpan:
Hi,

this is a super cool feature. It would help a lot with installations in locations with high noise. But it would be possible to know this ahead of time, instead of hitting it during debugging why a node is deaf…

it would be super cool to see in the app.

cheers,
Michal

src/mesh/LR11x0Interface.cpp
GUVWAF: This is not yet resolved.

src/mesh/RadioLibInterface.cpp
```diff
@@ -419,6 +509,9 @@ void RadioLibInterface::completeSending()
 
         // We are done sending that packet, release it
         packetPool.release(p);
+
+        // Update noise floor after transmitting (radio is now in a good state to sample)
+        updateNoiseFloor();
```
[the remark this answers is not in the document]
GUVWAF: Well, if `sendingPacket == NULL` then it doesn't return early.

But indeed I don't think we should do it here, we should first return to receive mode as fast as possible after transmitting.

GUVWAF (changes requested):

RCGV1:
@GUVWAF I had to enable "GodMode" to access that fucntion for lr11

GUVWAF:
@RCGV1 Ah, I see. We shouldn't do that, I'll see if I can make that a public method in RadioLib.

src/mesh/RadioLibInterface.h
```diff
@@ -97,11 +97,45 @@ class RadioLibInterface : public RadioInterface, protected concurrency::Notified
     /// are _trying_ to receive a packet currently (note - we might just be waiting for one)
     bool isReceiving = false;
 
+  protected:
+    // Noise floor tracking - rolling window of samples
+    static const uint8_t NOISE_FLOOR_SAMPLES = 20;
+    static const int32_t NOISE_FLOOR_MIN = -120; // Minimum noise floor clamp in dBm
```
GUVWAF: Why clamping to -120dBm? If there is only thermal noise, at 0 degrees Celsius and with a small bandwidth (62.5kHz), you can go down to -126dBm.

GUVWAF: When testing with the LR1110, I noticed that a value of -128 is returned when it is invalid (i.e. not in receive mode while calling the function). Hence I would say let's check for this specific value, and not clamp it to -120.

GUVWAF:
@RCGV1 For LR11x0, when we update RadioLib to include https://github.com/jgromes/RadioLib/pull/1714, we can use `lora.getRSSI(false, true)`.

For the rest, please see my comment on clamping.

Furthermore, we should still remove updating the noise floor from [here](https://github.com/meshtastic/firmware/pull/9347/changes#diff-4d7a147d99939f7143419faf6383d8d0bd2ed3cc8d25e76f527b082030c7c13dR516), as we need the radio to go the receive mode as fast as possible in order not to lose packets. Moreover, at least for the LR11x0, it is necessary to have it in receive mode before doing the RSSI measurement.

Lastly, in `startReceive()` it is also done only before enabling the Rx interrupt, which is also not ideal.
I would say maybe we could call `updateNoiseFloor()` from within DeviceTelemetry, just before we send our telemetry to the phone, which happens every minute. Then every 15 min. when the LocalStats are sent, we have a fairly good average.

src/modules/Telemetry/DeviceTelemetry.cpp
```diff
@@ -141,10 +143,11 @@ meshtastic_Telemetry DeviceTelemetryModule::getLocalStatsTelemetry()
         telemetry.variant.local_stats.num_tx_relay_canceled = router->txRelayCanceled;
     }
 
-    LOG_INFO("Sending local stats: uptime=%i, channel_utilization=%f, air_util_tx=%f, num_online_nodes=%i, num_total_nodes=%i",
+    LOG_INFO("Sending local stats: uptime=%i, channel_utilization=%f, air_util_tx=%f, num_online_nodes=%i, num_total_nodes=%i, "
+             "noise_floor=%f",
```
robekl: `telemetry.variant.local_stats.noise_floor` is int32_t, so `%d` should be used

src/mesh/RadioLibInterface.cpp
```diff
@@ -246,6 +254,94 @@ bool RadioLibInterface::findInTxQueue(NodeNum from, PacketId id)
...
+
+    LOG_DEBUG("Noise floor: %d dBm (samples: %d, latest: %d dBm)", currentNoiseFloor, getNoiseFloorSampleCount(), rssi);
+}
+
+int32_t RadioLibInterface::getAverageNoiseFloor()
+{
+    uint8_t sampleCount = getNoiseFloorSampleCount();
+
+    if (sampleCount == 0) {
+        return 0; // Return 0 if no samples
```
robekl: API/docs state uncalibrated should be -120 dBm (src/mesh/RadioLibInterface.h:129). This leaks impossible telemetry (0 dBm) via src/modules/Telemetry/DeviceTelemetry.cpp:122 and hurts data accuracy for fresh boot / low-traffic nodes.

src/mesh/RF95Interface.cpp
robekl, line 290: RF95Interface::startReceive() does not call RadioLibInterface::startReceive(), so the new noise-floor sampling hook in base (src/mesh/RadioLibInterface.cpp:616) is bypassed on RF95. Result: RF95 devices may never refresh noise floor except after TX-complete path, which biases or stalls measurements on RX-heavy nodes.

src/mesh/RadioLibInterface.cpp
```diff
@@ -246,6 +254,94 @@ bool RadioLibInterface::findInTxQueue(NodeNum from, PacketId id)
...
+        isNoiseFloorBufferFull = true;
+    }
+
+    // Calculate the new average using the rolling window
+    currentNoiseFloor = getAverageNoiseFloor();
+
+    LOG_DEBUG("Noise floor: %d dBm (samples: %d, latest: %d dBm)", currentNoiseFloor, getNoiseFloorSampleCount(), rssi);
+}
+
+int32_t RadioLibInterface::getAverageNoiseFloor()
```
robekl: Noise-floor buffer/index state is read and written without synchronization (getAverageNoiseFloor() vs updateNoiseFloor()), and telemetry calls into this from a different module thread (src/modules/Telemetry/DeviceTelemetry.cpp:122).

src/mesh/LR11x0Interface.cpp
```diff
@@ -317,4 +317,11 @@ template <typename T> bool LR11x0Interface<T>::sleep()
 
     return true;
 }
+
+template <typename T> int16_t LR11x0Interface<T>::getCurrentRSSI()
+{
+    float rssi = -110;
+    lora.getRssiInst(&rssi);
```
robekl: return value should be checked instead of always assuming success

Leoguy77:
any updates? This feature would be great🙏

michalpan:
Having the ability check the background signals. This is super useful to
have especially when placing a node on a roof, in connected environment.
Some of my nodes are deaf and it is sometimes  hard to say if it is caused
by the DIY radio or a nearby BTS.

It is a really good feature to have. Especially that other LoRa
applications have it.

Regards,
Michal

RCGV1:
I am waiting for the next RadioLib update because one of the chops doesn't support instant rssi and we need the next RadioLib update to support it (already got merged there)

RCGV1:
LR11x0 now uses lora.getRSSI(false, true).
Removed noise-floor updates from radio hot paths; sampling now happens from DeviceTelemetry.
Invalid LR11x0 RSSI -128 is rejected without clamping to -120.
Added locking around noise-floor sample state and fixed telemetry logging/default value handling.

platformio.ini
```diff
@@ -50,6 +50,7 @@ build_flags = -Wno-missing-field-initializers
 	-DRADIOLIB_EXCLUDE_FSK4=1
 	-DRADIOLIB_EXCLUDE_APRS=1
 	-DRADIOLIB_EXCLUDE_LORAWAN=1
+	-DRADIOLIB_GODMODE=1
```
GUVWAF: We don't need GODMODE anymore.

src/mesh/RadioLibInterface.cpp
```diff
@@ -256,6 +265,102 @@ bool RadioLibInterface::findInTxQueue(NodeNum from, PacketId id)
...
+    int16_t rssi = getCurrentRSSI();
+    if (rssi == NOISE_FLOOR_INVALID || rssi >= 0 || rssi < NOISE_FLOOR_VALID_MIN) {
+        LOG_DEBUG("Skipping invalid RSSI reading: %d", rssi);
+        return;
+    }
+
+    uint8_t sampleCount = 0;
+    int32_t average = NOISE_FLOOR_DEFAULT;
+    {
+        concurrency::LockGuard guard(&noiseFloorLock);
```
GUVWAF: Not sure we need the lock TBH, we don't do real multi-threading.

GUVWAF (commented):
Looks good to me now, left two comments. At least RadioLib's GODMODE should now be disabled.

GUVWAF (approved):

thebentern (approved):

PR #9350, opened 2026-01-18
Delete unused code

EricSesterhennX41:
CryptoEngine::clearKeys() is not used in the code base, therefore this cleanup removes the code. It might give casual reviewers the impression, that keys are wiped.

Since the code uses memset() which might be optimized away by the compiler, using the code might not even cause the memory to be wiped.

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -9,14 +9,14 @@ plugins:
 lint:
   enabled:
     - checkov@3.2.497
-    - renovate@42.81.8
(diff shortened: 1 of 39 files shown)


EricSesterhennX41:
I have no idea on how to resolve the trunk check runner issue. Any pointers would be appreciated, I do not get any warnings from trunk locally and the runner just reports:

src/mesh/CryptoEngine.cpp
 -:-  fmt  Incorrect formatting, autoformat by running 'trunk fmt'  clang-format

PR #9351, opened 2026-01-18
Add StatusMessage module and config overrides

jp-bennett:
Depends on https://github.com/meshtastic/protobufs/pull/847

Adds the twice-a-day Status message, to give users a place other than the long name to put extra status text.


rob-fi:
Before opening an issue I just wanted to check if I'm missing something. I can set the status message (using latest Android build and latest firmware) no problem and it shows up for my node that I'm connected to, however it doesn't show up for other nodes (also running latest) even after many days of it being set. Not sure if an Android app or firmware issue,

PR #9352, opened 2026-01-18
Slow down NodeInfo responses when on a "big" mesh

jp-bennett:
Detect a "bigMesh", and modify our NodeInfo throttling appropriately.

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -9,14 +9,14 @@ plugins:
 lint:
   enabled:
     - checkov@3.2.497
-    - renovate@42.81.8
+    - renovate@42.84.2
     - prettier@3.8.0
-    - trufflehog@3.92.4
+    - trufflehog@3.92.5
     - yamllint@1.38.0
-    - bandit@1.9.2
+    - bandit@1.9.3
     - trivy@0.68.2
     - taplo@0.10.0
-    - ruff@0.14.11
+    - ruff@0.14.13
     - isort@7.0.0
     - markdownlint@0.47.0
     - oxipng@10.0.0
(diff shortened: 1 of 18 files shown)


fifieldt (approved):

thebentern:
I'm not in favor of this change as-is, since it's a jarring cliff-edge behavior instead of a gradual scale up of intervals and preset-aware coefficients like we already have in the codebase. My proposal is to bump the existing base timeout to 10 minutes (maybe even more would be a good idea) and let the existing scaling logic take it from there:

```
meshtastic_MeshPacket *NodeInfoModule::allocReply()
{
    if (suppressReplyForCurrentRequest) {
        LOG_DEBUG("Skip send NodeInfo since we heard the requester <12h ago");
        ignoreRequest = true;
        suppressReplyForCurrentRequest = false;
        return NULL;
    }

    if (!airTime->isTxAllowedChannelUtil(false)) {
        LOG_DEBUG("Skip send NodeInfo > 40%% ch. util");
        return NULL;
    }

    // Use graduated scaling based on active mesh size
    uint32_t timeoutMs = Default::getConfiguredOrDefaultMsScaled(
        0, 10 * 60, nodeStatus->getNumOnline()) * 1000;

    if (lastSentToMesh && Throttle::isWithinTimespanMs(lastSentToMesh, timeoutMs)) {
        LOG_DEBUG("Skip send NodeInfo since we sent it <%ums ago", timeoutMs / 1000);
        ignoreRequest = true;
        return NULL;
    }

    ignoreRequest = false;
    // ... rest of function
}
```

In practice, this looks like:
image

Xaositek:
What does this look like for LongTurbo since it had promise in the future.

PR #9358, opened 2026-01-19
Traffic Management Module for packet forwarding logic

h3lix1:
## Summary

Large Mesh networks can get noisy fast: repeated position packets, bursty senders, and unknown/undecryptable traffic all consume limited airtime and power. When reaching 1000+ nodes, there needs to be more controls around handling a lot of broadcast traffic.

This PR introduces an optional `TrafficManagementModule` to reduce that noise without hard-coding one policy for all devices. It gives routers and clients controllable guardrails so they can prioritize useful traffic and keep the mesh healthier under load.

### Why?

- **Less unnecessary traffic:** position deduplication, rate limiting, and unknown-packet suppression reduce avoidable rebroadcasts.
- **Better routing behavior:** hop controls let deployments preserve range for useful packets while still containing telemetry/position spread when desired.
- **Predictable embedded resource usage:** a fixed-size 10-byte cache entry and O(1) cuckoo lookups keep memory and per-packet cost bounded.
- **Operational visibility and control:** new protobuf config/stats are exposed through Admin and Phone APIs for tuning and monitoring.
- **Safer rollout:** feature enablement is gated at both build time (`HAS_TRAFFIC_MANAGEMENT`, `TRAFFIC_MANAGEMENT_CACHE_SIZE`) and runtime (`moduleConfig.traffic_management.enabled`). It is about as self contained as it can be for a new feature.

### Architecture

```mermaid
classDiagram
    class MeshModule {
        +handleReceived(MeshPacket) ProcessMessage
        +alterReceived(MeshPacket)
        +wantPacket(MeshPacket) bool
        #isPromiscuous bool
        #encryptedOk bool
    }

    class OSThread {
        +runOnce() int32_t
        #setIntervalFromNow(ms)
    }

    class TrafficManagementModule {
        -cacheLock Lock
        -cache UnifiedCacheEntry*
        -stats TrafficManagementStats
        -cacheEpochMs uint32_t
        +getStats() TrafficManagementStats
        +resetStats()
        +recordRouterHopPreserved()
        -shouldDropPosition() bool
        -shouldDropUnknown() bool
        -isRateLimited() bool
        -shouldRespondToNodeInfo() bool
        -computePositionFingerprint()$ uint8_t
    }

    MeshModule <|-- TrafficManagementModule
    OSThread <|-- TrafficManagementModule

    TrafficManagementModule --> Router : allocForSending
    TrafficManagementModule --> MeshService : sendToMesh
```

- **Dual inheritance**: `MeshModule` (packet handling) + `concurrency::OSThread` (periodic maintenance)
- **Location**: `src/modules/TrafficManagementModule.cpp` and `src/modules/TrafficManagementModule.h`
- **Thread model**: Runs as a cooperative OSThread with 60-second maintenance intervals (`kMaintenanceIntervalMs`)
- **Promiscuous mode**: `isPromiscuous = true` allows inspection of all packets, not just those addressed to this node
- **Encrypted packet support**: `encryptedOk = true` permits handling undecryptable packets for unknown-packet tracking

### Compile-Time Gating

Module instantiation in `setupModules()` requires all three conditions:

```mermaid
flowchart TD
    A[setupModules called] --> B{HAS_TRAFFIC_MANAGEMENT<br/>== 1?}
    B -->|No| Z[Module not compiled]
    B -->|Yes| C{MESHTASTIC_EXCLUDE_<br/>TRAFFIC_MANAGEMENT?}
    C -->|Yes| Z
    C -->|No| D{moduleConfig.<br/>traffic_management.<br/>enabled?}
    D -->|No| Y[Module compiled but<br/>not instantiated]
    D -->|Yes| E[new TrafficManagementModule]
    E --> F[trafficManagementModule = instance]
```

1. `HAS_TRAFFIC_MANAGEMENT` defined as 1 (variant-level)
2. `!MESHTASTIC_EXCLUDE_TRAFFIC_MANAGEMENT` (build exclusion flag)
3. `moduleConfig.traffic_management.enabled` (runtime config)

---

### Unified Cache with Cuckoo Hashing

The module uses a **unified cache** combining all per-node tracking into a single 10-byte structure with **cuckoo hashing** for O(1) lookups. This provides:

- **O(1) lookup** via cuckoo hashing - critical at packet processing rates
- **10 bytes per entry** - unified across all platforms
- **Position fingerprinting** - deterministic 8-bit fingerprint from truncated coordinates
- **Adaptive timestamp resolution** - 8-bit timestamps with ~24 hour range
- **Saturating counters** - uint8_t (max 255) for rate/unknown counts

### UnifiedCacheEntry Structure (10 bytes) - All Platforms

A single compact structure used across ESP32, NRF52, and all other platforms:

```cpp
struct __attribute__((packed)) UnifiedCacheEntry {
    NodeNum node;            // 4 bytes - Node identifier (0 = empty)
    uint8_t pos_fingerprint; // 1 byte  - Lower 4 bits of lat + 4 bits of lon
    uint8_t rate_count;      // 1 byte  - Packet count (saturates at 255)
    uint8_t unknown_count;   // 1 byte  - Unknown packet count
    uint8_t pos_time;        // 1 byte  - Position timestamp (adaptive resolution)
    uint8_t rate_time;       // 1 byte  - Rate window start (adaptive resolution)
    uint8_t unknown_time;    // 1 byte  - Unknown tracking start (adaptive resolution)
};
static_assert(sizeof(UnifiedCacheEntry) == 10);
```

### Position Fingerprinting

Instead of storing full coordinates (8 bytes) or using a computed hash, the module uses a deterministic 8-bit fingerprint derived from the truncated lat/lon:

```cpp
fingerprint = (lat_low4 << 4) | lon_low4
```

Where `lat_low4` and `lon_low4` are the lower 4 significant bits of each truncated coordinate.

**Benefits over hash-based approach:**
- **No hash collisions for adjacent positions**: Sequential grid cells have sequential fingerprints
- **Deterministic**: Same input always produces same output
- **Collision only at distance**: Two positions collide only if they differ by 16+ grid cells in BOTH dimensions

**Guard for low precision**: If `position_precision_bits < 4`, uses `min(precision, 4)` bits from each coordinate.

### Adaptive Timestamp Resolution

All timestamps use 8-bit values with **adaptive resolution** calculated from config at startup:

```cpp
resolution = max(60, min(339, interval / 2))
```

- **60 second minimum** ensures reasonable precision for all intervals
- **339 second maximum** allows ~24 hour range (255 × 339 ≈ 86400 seconds)
- **interval/2** ensures at least 2 ticks per configured interval for accuracy

Since config changes require reboot, resolution is calculated once in the constructor.

**Memory Efficiency vs NodeDB:**

The traffic management cache is dramatically more memory-efficient than full NodeDB entries:

| Structure | Per-Entry Size | 2048 Entries | Contents |
| --------- | -------------- | ------------ | -------- |
| `meshtastic_NodeInfoLite` | **196 bytes** | ~400 KB | Full node: user info, position, device metrics, public key, etc. |
| `UnifiedCacheEntry` | **10 bytes** | ~20 KB | Traffic state only: fingerprint, counters, timestamps |

The unified cache achieves **~20x memory reduction** compared to duplicating NodeDB data because it stores only the minimal state needed for traffic decisions:
- No encryption keys (32 bytes saved)
- No user strings (long_name, short_name - ~50 bytes saved)
- No device metrics (~24 bytes saved)
- No full coordinates (8 bytes → 1 byte fingerprint)
- Adaptive 8-bit timestamps with ~24 hour range
- Saturating counters (8-bit vs 32-bit)

### Cuckoo Hashing Algorithm

Cuckoo hashing provides O(1) worst-case lookup using two hash functions:

```mermaid
flowchart TD
    A[lookup node] --> B{cache h1 node == node?}
    B -->|Yes| C[Return entry at h1]
    B -->|No| D{cache h2 node == node?}
    D -->|Yes| E[Return entry at h2]
    D -->|No| F[Not found]

    G[insert node] --> H{slot h1 empty?}
    H -->|Yes| I[Insert at h1]
    H -->|No| J{slot h2 empty?}
    J -->|Yes| K[Insert at h2]
    J -->|No| L[Kick entry from h1]
    L --> M[Place kicked entry at its alternate]
    M --> N{Max kicks reached?}
    N -->|No| L
    N -->|Yes| O[Drop displaced entry]
```

**Hash functions:**

```cpp
// h1: Simple mask for sequential NodeNums
uint16_t cuckooHash1(NodeNum n) { return n & cacheMask(); }

// h2: Golden ratio multiplicative hash (decorrelated from h1)
uint16_t cuckooHash2(NodeNum n) {
    return ((n * 2654435769u) >> (32 - bits)) & cacheMask();
}
```

**Cuckoo displacement algorithm:**

1. Insert new entry at h1, displacing existing entry
2. Move displaced entry to its alternate location (h2 if was at h1, or vice versa)
3. If alternate location occupied, repeat displacement
4. After 16 kicks (cycle detected), drop the displaced entry to preserve cache integrity

The displaced entry is dropped (not placed at an arbitrary slot) because placing it outside its two valid cuckoo locations would make it unreachable by `findEntry()`, corrupting the cache. This is the standard resolution for cuckoo cycles with fixed-size tables.

This O(1) lookup is critical for packet processing performance:

- **Predictable latency**: Every lookup checks exactly 2 slots (vs O(n) linear scan)
- **Scales with cache size**: 1000 entries has same lookup cost as 100 entries
- **No dynamic allocation**: All operations work on pre-allocated array
- **Cache-friendly**: Sequential memory access for displacement chain

**Performance comparison:**

| Operation | Old (Linear Scan) | New (Cuckoo Hash) |
| --------- | ----------------- | ----------------- |
| Lookup    | O(n) = ~1000 ops  | O(1) = 2 ops      |
| Insert    | O(n) = ~1000 ops  | O(1) amortized    |

### Cache Sizing

Cache size rounds up to power-of-2 for efficient cuckoo hash indexing:

```cpp
// TRAFFIC_MANAGEMENT_CACHE_SIZE=1000 → cacheSize()=1024
constexpr uint16_t cacheSize() {
    uint16_t n = TRAFFIC_MANAGEMENT_CACHE_SIZE;
    n--; n |= n >> 1; n |= n >> 2; n |= n >> 4; n |= n >> 8;
    return n + 1;
}
```

### Memory Budget Calculations

**Cache memory at 2048 entries (unified 10-byte structure):**

| Platform          | Allocation | Default Entries | Total Memory |
| ----------------- | ---------- | --------------- | ------------ |
| ESP32 + PSRAM     | PSRAM      | 2048            | **20 KB** |
| ESP32 (no PSRAM)  | Heap       | 2048            | **20 KB** |
| nRF52840          | Heap       | 1024            | **10 KB** |
| Native/Portduino  | Heap       | 2048            | **20 KB** |

**Note**: `TRAFFIC_MANAGEMENT_CACHE_SIZE` can be overridden per-variant. The value rounds up to the next power of 2 for efficient cuckoo hash indexing (e.g., 1000 → 1024, 2000 → 2048).

**Typical available RAM by platform:**

| Platform       | Total RAM                | Free After Boot | NodeDB Overhead     |
| -------------- | ------------------------ | --------------- | ------------------- |
| ESP32 + PSRAM  | 320KB DRAM + 4-8MB PSRAM | 80-150KB DRAM   | ~16.6KB (100 nodes) |
| ESP32 no PSRAM | ~320KB                   | 80-150KB        | ~16.6KB (100 nodes) |
| nRF52840       | 256KB                    | 50-80KB         | ~13.3KB (80 nodes)  |

**Risk assessment:**

| Platform       | Free Heap    | Cache Size   | % of Free RAM |
| -------------- | ------------ | ------------ | ------------- |
| ESP32 + PSRAM  | PSRAM: 4-8MB | 20KB → PSRAM | <1%           |
| ESP32 no PSRAM | 80-150KB     | 20KB → heap  | 13-25%        |
| nRF52840       | ~160KB       | 10KB → heap  | ~6%           |

### Current Enablement Status

Traffic management is enabled on:

- **Heltec V4** (ESP32-S3 with PSRAM) - 2048 entries, cache in PSRAM
- **Station G2** (ESP32-S3) - 2048 entries
- **Tracker T1000-E** (nRF52840) - 1024 entries, cache in heap
- **Native/Portduino** - 2048 entries

Other variants can enable via `HAS_TRAFFIC_MANAGEMENT=1` in their `variant.h`.

**Example variant enablement:**

```cpp
// In variant.h
#ifndef HAS_TRAFFIC_MANAGEMENT
#define HAS_TRAFFIC_MANAGEMENT 1
#endif
#ifndef TRAFFIC_MANAGEMENT_CACHE_SIZE
#define TRAFFIC_MANAGEMENT_CACHE_SIZE 2048  // ~20KB with 10-byte entries
#endif
```

---

### Lock Architecture

The module uses a single `mutable concurrency::Lock cacheLock` member to protect all cache and statistics operations.

**Codebase Lock Pattern Comparison:**

| Pattern            | Declaration               | Usage                      | Used By                     |
| ------------------ | ------------------------- | -------------------------- | --------------------------- |
| Global pointer     | `extern Lock *spiLock;`   | `LockGuard g(spiLock);`    | SPI bus, crypto             |
| Member by value    | `Lock mutex;`             | `LockGuard g(&mutex);`     | PhoneAPI, Screen            |
| **Mutable member** | `mutable Lock cacheLock;` | `LockGuard g(&cacheLock);` | **TrafficManagementModule** |

**Why `mutable`?** The `getStats()` method is `const` but must acquire a lock. Declaring the lock as `mutable` allows locking in const methods without requiring a dangerous `const_cast`. This is semantically correct: locking doesn't affect the logical const-ness of the object.

### Critical Sections Analysis

| Function               | Lock Scope            | Duration           | Notes                              |
| ---------------------- | --------------------- | ------------------ | ---------------------------------- |
| `getStats()`           | Entire function       | ~32 byte copy      | Returns copy to prevent torn reads |
| `resetStats()`         | Entire function       | Single memset      | Fast zero-initialization           |
| `incrementStat()`      | Entire function       | Single increment   | Called frequently                   |
| `runOnce()`            | Full cache sweep      | O(CACHE_SIZE)      | Runs every 60s on OSThread         |
| `shouldDropPosition()` | Cache lookup + update | O(1) cuckoo lookup | Per-packet cost                    |
| `isRateLimited()`      | Cache lookup + update | O(1) cuckoo lookup | Per-packet cost                    |
| `shouldDropUnknown()`  | Cache lookup + update | O(1) cuckoo lookup | Per-packet cost                    |

### Statistics Counters

- Protected increment via `incrementStat()` with lock guard
- Statistics struct (`meshtastic_TrafficManagementStats`) is 32 bytes (8 x uint32_t)
- `getStats()` returns copy under lock to prevent torn reads across multiple counters

---

### handleReceived() Processing Order

```mermaid
flowchart TD
    A[Packet Received] --> B{Module Enabled?}
    B -->|No| Z[Return CONTINUE]
    B -->|Yes| C[Increment packets_inspected]

    C --> D{Undecoded Packet?}
    D -->|Yes| E{drop_unknown enabled?}
    E -->|Yes| F{Threshold exceeded?}
    F -->|Yes| G[/"Drop: unknown_packet_drops++<br/>Return STOP"/]
    F -->|No| H[Track in cache]
    E -->|No| H
    H --> Z
    D -->|No| I{NodeInfo Request?}

    I -->|Yes| J{nodeinfo_direct_response<br/>enabled?}
    J -->|Yes| K{want_response &&<br/>!broadcast && !toUs?}
    K -->|Yes| L{hopsAway <= maxHops?}
    L -->|Yes| M[/"Send cached User<br/>nodeinfo_cache_hits++<br/>Return STOP"/]
    L -->|No - too far| N[Continue to next check]
    K -->|No| N
    J -->|No| N
    I -->|No| N

    N --> O{Position Packet?<br/>Not from us?}
    O -->|Yes| P{position_dedup_enabled?}
    P -->|Yes| Q[Decode & truncate lat/lon]
    Q --> R{Duplicate within<br/>min_interval?}
    R -->|Yes| S[/"Drop: position_dedup_drops++<br/>Return STOP"/]
    R -->|No| T[Update cache]
    T --> U
    P -->|No| U
    O -->|No| U

    U{Rate limit check?<br/>Not from us?<br/>Not ROUTING/ADMIN?}
    U -->|Yes| V{rate_limit_enabled?}
    V -->|Yes| W{Limit exceeded?}
    W -->|Yes| X[/"Drop: rate_limit_drops++<br/>Return STOP"/]
    W -->|No| Y[Increment counter]
    Y --> Z
    V -->|No| Z
    U -->|No| Z

    Z[Return CONTINUE]
```

### NodeInfo Direct Response - Role-Based Limits

Both routers and clients use maxHops logic: respond when `hopsAway <= threshold`.
**Role determines the enforced maximum** (not just a default).

| Device Role | Max Limit | Responds to |
|-------------|-----------|-------------|
| ROUTER, ROUTER_LATE, CLIENT_BASE | 3 | Nodes within 0-3 hops |
| All other roles (clients) | 0 | Direct neighbors only (0 hops) |

**Examples:**
- Router with config=0: uses limit (3) → responds to 0, 1, 2, 3 hops
- Router with config=2: uses 2 → responds to 0, 1, 2 hops
- Router with config=5: clamped to 3 → responds to 0, 1, 2, 3 hops
- Client with config=0: uses limit (0) → responds to 0 hops only
- Client with config=2: clamped to 0 → responds to 0 hops only

**Note**: `nodeinfo_direct_response` must be enabled separately for this feature to work.

**Adjustable constants** in `TrafficManagementModule.cpp`:
- `kRouterDefaultMaxHops = 3` - Router limit: max 3 hops
- `kClientDefaultMaxHops = 0` - Client limit: direct only (cannot increase)

### alterReceived() - Relayed Broadcast Hop Exhaustion

For relayed broadcasts from other nodes (`!isFromUs(&mp) && isBroadcast(mp.to)`):

- If `exhaust_hop_telemetry` and port is `TELEMETRY_APP`: exhaust hops
- If `exhaust_hop_position` and port is `POSITION_APP`: exhaust hops
- Sets `hop_start = hop_start - hop_limit + 1` to preserve correct `hopsAway` for the one remaining relay hop
- Sets `hop_limit = 0` and signals `exhaustRequested` to `perhapsRebroadcast()`
- Prevents relayed telemetry/position from propagating further through the mesh
- Our own outgoing packets are unaffected

---

## Position Truncation Algorithm

Coordinates are truncated to configurable precision for deduplication:

```cpp
int32_t truncateLatLon(int32_t value, uint8_t precision)
{
    if (precision == 0 || precision >= 32)
        return value;

    uint32_t mask = UINT32_MAX << (32 - precision);
    uint32_t truncated = static_cast<uint32_t>(value) & mask;
    truncated += (1u << (31 - precision));  // Round to center of cell
    return static_cast<int32_t>(truncated);
}
```

**Precision examples** (approximate):
| Bits | Grid Size | Use Case |
|------|-----------|----------|
| 32 | ~1cm | No truncation |
| 24 | ~10m | Default - pedestrian/general |
| 20 | ~40m | Vehicle |
| 16 | ~600m | City block |
| 12 | ~10km | Regional |

### Default Values

Defaults are defined in `src/mesh/Default.h` and applied at runtime via `Default::getConfiguredOrDefault()`:

| Setting | Default | Meaning |
|---------|---------|---------|
| `position_precision_bits` | **24** | ~10m grid cells |
| `position_min_interval_secs` | **86400** (ONE_DAY) | 1 day between identical positions |

These defaults provide aggressive deduplication suitable for reducing network traffic while still allowing meaningful position updates. Users can adjust via app settings.

---

### Router.cpp Changes

When `traffic_management.router_preserve_hops` is enabled and the local device is a router role (ROUTER, ROUTER_LATE, or CLIENT_BASE), hops are preserved for decoded packets **except** position and telemetry (which have their own `exhaust_hop_*` controls).

This is a "best effort" approach that does not attempt to identify whether the previous relay was also a router, since the `relay_node` field only contains the last byte of the node ID (8 bits), making reliable identification impossible due to collision risk.

The feature preserves hop budget when router-role devices relay non-sensor packets (text messages, nodeinfo, routing, etc.), extending effective mesh range through infrastructure routers while still allowing position/telemetry to be managed independently via the exhaust hop settings.

---

### TrafficManagementConfig (module_config.proto)

14 configuration fields using nanopb STATIC encoding:

```protobuf
message TrafficManagementConfig {
    bool enabled = 1;
    bool position_dedup_enabled = 2;
    uint32 position_precision_bits = 3;
    uint32 position_min_interval_secs = 4;
    bool nodeinfo_direct_response = 5;
    uint32 nodeinfo_direct_response_max_hops = 6;
    bool rate_limit_enabled = 7;
    uint32 rate_limit_window_secs = 8;
    uint32 rate_limit_max_packets = 9;
    bool drop_unknown_enabled = 10;
    uint32 unknown_packet_threshold = 11;
    bool exhaust_hop_telemetry = 12;
    bool exhaust_hop_position = 13;
    bool router_preserve_hops = 14;
}
```

Generated C struct: `meshtastic_ModuleConfig_TrafficManagementConfig` (~36 bytes)

### TrafficManagementStats (telemetry.proto)

7 counter fields for telemetry reporting:

```protobuf
message TrafficManagementStats {
    uint32 packets_inspected = 1;
    uint32 position_dedup_drops = 2;
    uint32 nodeinfo_cache_hits = 3;
    uint32 rate_limit_drops = 4;
    uint32 unknown_packet_drops = 5;
    uint32 hop_exhausted_packets = 6;
    uint32 router_hops_preserved = 7;
}
```

Generated C struct: `meshtastic_TrafficManagementStats` (28 bytes)

---

### AdminModule.cpp Changes

**Get handler** (`handleGetModuleConfig`):

```cpp
case meshtastic_AdminMessage_ModuleConfigType_TRAFFICMANAGEMENT_CONFIG:
    res.get_module_config_response.which_payload_variant =
        meshtastic_ModuleConfig_traffic_management_tag;
    res.get_module_config_response.payload_variant.traffic_management =
        moduleConfig.traffic_management;
    break;
```

**Set handler** (`handleSetModuleConfig`):

```cpp
case meshtastic_ModuleConfig_traffic_management_tag:
    moduleConfig.has_traffic_management = true;
    moduleConfig.traffic_management = c.payload_variant.traffic_management;
    break;
```

### PhoneAPI.cpp Changes

Traffic management config included in module config enumeration sent to connected clients. The enumeration loop uses the auto-updating `_meshtastic_AdminMessage_ModuleConfigType_MAX + 1` pattern to detect when all module configs have been sent.

---

## Periodic Maintenance Thread

The `runOnce()` method executes every 60 seconds to garbage-collect stale cache entries.

For each cache entry, individual tracking data (position, rate, unknown) is checked against its TTL:
- **Position data**: 4x the configured `position_min_interval_secs`
- **Rate limit data**: 2x the configured `rate_limit_window_secs` (or 10 minutes if unconfigured)
- **Unknown tracking data**: 5x the unknown reset window (5 minutes)

If all tracking data for an entry has expired, the entire slot is freed. The epoch is also checked and reset when relative timestamps approach 8-bit overflow (~19 hours at maximum resolution).

---

### mesh-pb-constants.h Additions

```cpp
// Default disabled; variants opt-in
#ifndef HAS_TRAFFIC_MANAGEMENT
#define HAS_TRAFFIC_MANAGEMENT 0
#endif

// Default cache size when enabled
#ifndef TRAFFIC_MANAGEMENT_CACHE_SIZE
#if HAS_TRAFFIC_MANAGEMENT
#define TRAFFIC_MANAGEMENT_CACHE_SIZE 1000
#else
#define TRAFFIC_MANAGEMENT_CACHE_SIZE 0
#endif
#endif
```

### Variant Example: Heltec V4

`variants/esp32s3/heltec_v4/variant.h`:

```cpp
#ifndef HAS_TRAFFIC_MANAGEMENT
#define HAS_TRAFFIC_MANAGEMENT 1
#endif
#ifndef TRAFFIC_MANAGEMENT_CACHE_SIZE
#define TRAFFIC_MANAGEMENT_CACHE_SIZE 2048
#endif
```

### DIFF
--- a/src/graphics/EInkDisplay2.cpp
+++ b/src/graphics/EInkDisplay2.cpp
@@ -143,6 +143,10 @@ bool EInkDisplay::connect()
 #ifdef ELECROW_ThinkNode_M1
     // ThinkNode M1 has a hardware dimmable backlight. Start enabled
     digitalWrite(PIN_EINK_EN, HIGH);
+#elif defined(MINI_EPAPER_S3)
+    // T-Mini Epaper S3 requires panel power rail enabled before SPI transfer.
+    digitalWrite(PIN_EINK_EN, HIGH);
+    delay(10);
 #else
     digitalWrite(PIN_EINK_EN, LOW);
 #endif
@@ -202,7 +206,8 @@ bool EInkDisplay::connect()
     }
 
 #elif defined(HELTEC_WIRELESS_PAPER_V1_0) || defined(HELTEC_VISION_MASTER_E290) || defined(TLORA_T3S3_EPAPER) ||                 \
-    defined(CROWPANEL_ESP32S3_5_EPAPER) || defined(CROWPANEL_ESP32S3_4_EPAPER) || defined(CROWPANEL_ESP32S3_2_EPAPER)
+    defined(CROWPANEL_ESP32S3_5_EPAPER) || defined(CROWPANEL_ESP32S3_4_EPAPER) || defined(CROWPANEL_ESP32S3_2_EPAPER) ||         \
+    defined(MINI_EPAPER_S3)
     {
         // Start HSPI
         hspi = new SPIClass(HSPI);
@@ -216,9 +221,13 @@ bool EInkDisplay::connect()
 
         // Init GxEPD2
         adafruitDisplay->init();
+#if defined(MINI_EPAPER_S3)
+        adafruitDisplay->setRotation(3);
+#else
         adafruitDisplay->setRotation(3);
 #if defined(CROWPANEL_ESP32S3_5_EPAPER) || defined(CROWPANEL_ESP32S3_4_EPAPER)
         adafruitDisplay->setRotation(0);
+#endif
 #endif
     }
 #elif defined(PCA10059) || defined(ME25LS01)
@@ -259,17 +268,6 @@ bool EInkDisplay::connect()
         adafruitDisplay->setRotation(3);
         adafruitDisplay->setPartialWindow(0, 0, EINK_WIDTH, EINK_HEIGHT);
     }
-#elif defined(MINI_EPAPER_S3)
-    spi1 = new SPIClass(HSPI);
-    spi1->begin(PIN_SPI1_SCK, PIN_SPI1_MISO, PIN_SPI1_MOSI, PIN_EINK_CS);
-
-    // Create GxEPD2 objects
-    auto lowLevel = new EINK_DISPLAY_MODEL(PIN_EINK_CS, PIN_EINK_DC, PIN_EINK_RES, PIN_EINK_BUSY, *spi1);
-    adafruitDisplay = new GxEPD2_BW<EINK_DISPLAY_MODEL, EINK_DISPLAY_MODEL::HEIGHT>(*lowLevel);
-
-    // Init GxEPD2
-    adafruitDisplay->init();
-    adafruitDisplay->setRotation(1);
 #elif defined(HELTEC_WIRELESS_PAPER) || defined(HELTEC_VISION_MASTER_E213)
 
     // Detect display model, before starting SPI
--- a/src/graphics/EInkDisplay2.h
+++ b/src/graphics/EInkDisplay2.h
@@ -89,12 +89,12 @@ class EInkDisplay : public OLEDDisplay
     // If display uses HSPI
 #if defined(HELTEC_WIRELESS_PAPER) || defined(HELTEC_WIRELESS_PAPER_V1_0) || defined(HELTEC_VISION_MASTER_E213) ||               \
     defined(HELTEC_VISION_MASTER_E290) || defined(TLORA_T3S3_EPAPER) || defined(CROWPANEL_ESP32S3_5_EPAPER) ||                   \
-    defined(CROWPANEL_ESP32S3_4_EPAPER) || defined(CROWPANEL_ESP32S3_2_EPAPER) || defined(ELECROW_ThinkNode_M5)
+    defined(CROWPANEL_ESP32S3_4_EPAPER) || defined(CROWPANEL_ESP32S3_2_EPAPER) || defined(ELECROW_ThinkNode_M5) ||               \
+    defined(MINI_EPAPER_S3)
     SPIClass *hspi = NULL;
 #endif
 
-#if defined(HELTEC_MESH_POCKET) || defined(SEEED_WIO_TRACKER_L1_EINK) || defined(HELTEC_MESH_SOLAR_EINK) ||                      \
-    defined(MINI_EPAPER_S3)
+#if defined(HELTEC_MESH_POCKET) || defined(SEEED_WIO_TRACKER_L1_EINK) || defined(HELTEC_MESH_SOLAR_EINK)
     SPIClass *spi1 = NULL;
 #endif
 
--- a/src/graphics/niche/Drivers/EInk/GDEW0102T4.cpp
+++ b/src/graphics/niche/Drivers/EInk/GDEW0102T4.cpp
@@ -1,0 +1,178 @@
+#ifdef MESHTASTIC_INCLUDE_NICHE_GRAPHICS
+
+#include "./GDEW0102T4.h"
+
+#include <cstring>
+
+using namespace NicheGraphics::Drivers;
+
+// LUTs from GxEPD2_102.cpp (GDEW0102T4 / UC8175).
+static const uint8_t LUT_W_FULL[] = {
+    0x60, 0x5A, 0x5A, 0x00, 0x00, 0x01, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+};
+
+static const uint8_t LUT_B_FULL[] = {
+    0x90, 0x5A, 0x5A, 0x00, 0x00, 0x01, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+};
+
+static const uint8_t LUT_W_FAST[] = {
+    0x60, 0x01, 0x01, 0x00, 0x00, 0x01, //
+    0x80, 0x12, 0x00, 0x00, 0x00, 0x01, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+};
+
+static const uint8_t LUT_B_FAST[] = {
+    0x90, 0x01, 0x01, 0x00, 0x00, 0x01, //
+    0x40, 0x14, 0x00, 0x00, 0x00, 0x01, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, //
+};
+
+GDEW0102T4::GDEW0102T4() : UC8175(width, height, supported) {}
+
+void GDEW0102T4::setFastConfig(FastConfig cfg)
+{
+    // Clamp out only clearly invalid PLL settings.
+    if (cfg.reg30 < 0x05)
+        cfg.reg30 = 0x05;
+    fastConfig = cfg;
+}
+
+GDEW0102T4::FastConfig GDEW0102T4::getFastConfig() const
+{
+    return fastConfig;
+}
+
+void GDEW0102T4::configCommon()
+{
+    // Init path aligned with GxEPD2_GDEW0102T4 (UC8175 family).
+    sendCommand(0xD2);
+    sendData(0x3F);
+
+    sendCommand(0x00);
+    sendData(0x6F);
+
+    sendCommand(0x01);
+    sendData(0x03);
+    sendData(0x00);
+    sendData(0x2B);
+    sendData(0x2B);
+
+    sendCommand(0x06);
+    sendData(0x3F);
+
+    sendCommand(0x2A);
+    sendData(0x00);
+    sendData(0x00);
+
+    sendCommand(0x30); // PLL / drive clock
+    sendData(0x13);
+
+    sendCommand(0x50); // Last border/data interval; subtle but can affect artifacts
+    sendData(0x57);
+
+    sendCommand(0x60);
+    sendData(0x22);
+
+    sendCommand(0x61);
+    sendData(width);
+    sendData(height);
+
+    sendCommand(0x82); // VCOM DC setting
+    sendData(0x12);
+
+    sendCommand(0xE3);
+    sendData(0x33);
+}
+
+void GDEW0102T4::configFull()
+{
+    sendCommand(0x23);
+    sendData(LUT_W_FULL, sizeof(LUT_W_FULL));
+    sendCommand(0x24);
+    sendData(LUT_B_FULL, sizeof(LUT_B_FULL));
+
+    powerOn();
+}
+
+void GDEW0102T4::configFast()
+{
+    uint8_t lutW[sizeof(LUT_W_FAST)];
+    uint8_t lutB[sizeof(LUT_B_FAST)];
+    memcpy(lutW, LUT_W_FAST, sizeof(LUT_W_FAST));
+    memcpy(lutB, LUT_B_FAST, sizeof(LUT_B_FAST));
+
+    // Second stage duration bytes are the main "darkness vs ghosting" control for this panel.
+    lutW[7] = fastConfig.lutW2;
+    lutB[7] = fastConfig.lutB2;
+
+    sendCommand(0x30);
+    sendData(fastConfig.reg30);
+
+    sendCommand(0x50);
+    sendData(fastConfig.reg50);
+
+    sendCommand(0x82);
+    sendData(fastConfig.reg82);
+
+    sendCommand(0x23);
+    sendData(lutW, sizeof(lutW));
+    sendCommand(0x24);
+    sendData(lutB, sizeof(lutB));
+
+    powerOn();
+}
+
+void GDEW0102T4::writeOldImage()
+{
+    // On this panel, FULL refresh is most reliable when "old image" is all white.
+    if (updateType == FULL) {
+        sendCommand(0x10);
+        // Use buffered writes of 0xFF to avoid per-byte SPI transactions.
+        const uint16_t chunkSize = 64;
+        uint8_t ffBuf[chunkSize];
+        memset(ffBuf, 0xFF, sizeof(ffBuf));
+
+        uint32_t remaining = bufferSize;
+        while (remaining > 0) {
+            uint16_t toSend = remaining > chunkSize ? chunkSize : static_cast<uint16_t>(remaining);
+            sendData(ffBuf, toSend);
+            remaining -= toSend;
+        }
+        return;
+    }
+
+    // FAST refresh uses differential data (previous frame as old image).
+    if (previousBuffer) {
+        writeImage(0x10, previousBuffer);
+    } else {
+        writeImage(0x10, buffer);
+    }
+}
+
+void GDEW0102T4::finalizeUpdate()
+{
+    // Keep panel out of deep-sleep between updates for better reliability of repeated FAST refresh.
+    powerOff();
+}
+
+#endif // MESHTASTIC_INCLUDE_NICHE_GRAPHICS
--- a/src/graphics/niche/Drivers/EInk/GDEW0102T4.h
+++ b/src/graphics/niche/Drivers/EInk/GDEW0102T4.h
@@ -1,0 +1,55 @@
+/*
+
+E-Ink display driver
+    - GDEW0102T4
+    - Controller: UC8175
+    - Size: 1.02 inch
+    - Resolution: 80px x 128px
+
+*/
+
+#pragma once
+
+#ifdef MESHTASTIC_INCLUDE_NICHE_GRAPHICS
+
+#include "configuration.h"
+
+#include "./UC8175.h"
+
+namespace NicheGraphics::Drivers
+{
+
+class GDEW0102T4 : public UC8175
+{
+  private:
+    static constexpr uint16_t width = 80;
+    static constexpr uint16_t height = 128;
+    static constexpr UpdateTypes supported = (UpdateTypes)(FULL | FAST);
+
+  public:
+    struct FastConfig {
+        uint8_t reg30;
+        uint8_t reg50;
+        uint8_t reg82;
+        uint8_t lutW2;
+        uint8_t lutB2;
+    };
+
+    GDEW0102T4();
+    void setFastConfig(FastConfig cfg);
+    FastConfig getFastConfig() const;
+
+  protected:
+    void configCommon() override;
+    void configFull() override;
+    void configFast() override;
+    void writeOldImage() override;
+    void finalizeUpdate() override;
+
+  private:
+    FastConfig fastConfig = {0x13, 0xF2, 0x12, 0x0E, 0x14};
+};
+
+} // namespace NicheGraphics::Drivers
+
+#endif // MESHTASTIC_INCLUDE_NICHE_GRAPHICS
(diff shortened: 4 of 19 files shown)


src/modules/TrafficManagementModule.cpp
```diff
@@ -0,0 +1,1392 @@
...
+        logAction("exhaust", &mp, reason);
+        // Adjust hop_start so downstream nodes compute correct hopsAway (hop_start - hop_limit).
+        // Without this, hop_limit=0 with original hop_start would show inflated hopsAway.
+        mp.hop_start = mp.hop_start - mp.hop_limit + 1;
+        mp.hop_limit = 0;
+        // Signal perhapsRebroadcast() to allow one final relay with hop_limit=0.
+        // Without this flag, perhapsRebroadcast() would skip the packet since hop_limit==0.
+        // The flag is checked in NextHopRouter::perhapsRebroadcast() which forces
+        // tosend->hop_limit=0, ensuring no further propagation beyond the next node.
+        exhaustRequested = true;
```
robekl, line 1021: `exhaustRequested` is set here and only reset in `handleReceived()`. But duplicate packets can be handled in `shouldFilterReceived()` before `handleReceived()` runs, and that path calls `perhapsRebroadcast()`, which reads `shouldExhaustHops()`. This makes hop-exhaust state non-packet-scoped and can cause incorrect extra rebroadcast behavior.

src/modules/TrafficManagementModule.cpp
```diff
@@ -0,0 +1,1392 @@
...
+ * @param lat_truncated  Precision-truncated latitude
+ * @param lon_truncated  Precision-truncated longitude
+ * @param precision      Number of significant bits (1-32)
+ * @return               8-bit fingerprint (4 bits lat + 4 bits lon)
+ */
+uint8_t TrafficManagementModule::computePositionFingerprint(int32_t lat_truncated, int32_t lon_truncated, uint8_t precision)
+{
+    // Guard: no truncation means no fingerprint
+    if (precision == 0 || precision >= 32)
+        return 0;
```
robekl: all positions will share the same fingerprint at precision = 32 or precision > 32. is this okay?

src/modules/TrafficManagementModule.cpp
```diff
@@ -0,0 +1,1392 @@
...
+    (void)nowMs;
+    return false;
+#else
+    if (!pos->has_latitude_i || !pos->has_longitude_i)
+        return false;
+
+    uint8_t precision = Default::getConfiguredOrDefault(moduleConfig.traffic_management.position_precision_bits,
+                                                        default_traffic_mgmt_position_precision_bits);
+    if (precision > 32)
+        precision = 32;
```
robekl: may interact with the "all same fingerprint" at precision >= 32 concern

src/modules/TrafficManagementModule.cpp
```diff
@@ -0,0 +1,1392 @@
...
+ * Invalidates all cached timestamps while preserving node associations.
+ */
+void TrafficManagementModule::resetEpoch(uint32_t nowMs)
+{
+#if TRAFFIC_MANAGEMENT_CACHE_SIZE > 0
+    TM_LOG_DEBUG("Resetting cache epoch");
+    cacheEpochMs = nowMs;
+
+    // Invalidate all relative timestamps
+    for (uint16_t i = 0; i < cacheSize(); i++) {
```
robekl: timestamps are cleared but not pos_fingerprint, leaving old position identity active while time base changes.

src/modules/TrafficManagementModule.cpp
h3lix1: Updated in ef1af77 - Added tests as well for potential future regressions here.

src/modules/TrafficManagementModule.cpp
h3lix1: The default should be default_traffic_mgmt_position_precision_bits (which is 24). I'll use that as the default if precision is invalid. Pondering the merrit of putting a lower limit as well. (i.e. precision = 4) but I don't think I'm too concerned here. Fixed in 34b89a1

src/modules/TrafficManagementModule.cpp
h3lix1: Also fixed with the issue below in 34b89a1

h3lix1:
@robekl Feel free to send me your ko-fi (or the like) for apprecaition of finding these.

src/modules/TrafficManagementModule.cpp
h3lix1: Did a better way of flishing everything for stale resources in 2c98d2f .. And added tests to validate a flush for future me.

robekl:
> @robekl Feel free to send me your ko-fi (or the like) for apprecaition of finding these.

Thanks, not necessary. I have also been asked to stop reviewing code. 🤷‍♂️

src/modules/TrafficManagementModule.cpp
```diff
@@ -0,0 +1,1410 @@
...
+    if (!shouldExhaust || !isBroadcast(mp.to))
+        return;
+
+    if (mp.hop_limit > 0) {
+        const char *reason = isTelemetry ? "exhaust-hop-telemetry" : "exhaust-hop-position";
+        logAction("exhaust", &mp, reason);
+        // Adjust hop_start so downstream nodes compute correct hopsAway (hop_start - hop_limit).
+        // Without this, hop_limit=0 with original hop_start would show inflated hopsAway.
+        mp.hop_start = mp.hop_start - mp.hop_limit + 1;
+        mp.hop_limit = 0;
```
NomDeTom, line 1015: I'm slightly concerned that this is a universal block, and I'm not happy with it.

I suggest (in no particular order):
* Rather than block, exclude from free hops on routers
* Rather than block, enforce a maximum hops (make them the equivalent of hopstart =3 or something)
* Gracefully throttle based on chutil
* Gracefully throttle based on frequency (enforce a minimum interval)
* Rather than block, charge additional hops, e.g. hops remaining-2

NomDeTom: I would also ask that the role of the originator be considered. What is the point of sensor or tracker roles if not to send telemetry and position packets?

h3lix1: I'll answer, but in the particular order provided.

It's not a universal block, but it does limit the number of hops left. I should probably gate this behind ROUTER-like roles though to enforce that it's only for nodes that can send packets with a large radius. My expectation is that it will take a couple hops to get to a router, and that it will likely be heard by multiple routers before the hops are limited.

* Position and telemetry packets are already excluded from free hops on routers (at least with this feature)
* This is possible.. for our mesh that's still a very large radius, and this will run amuck with "overheard hoptimization". Thus the reason why hop_limit of 0 was chosen..
* chutil is a metric that I'd like to stay away from to base these decisions on.. Meshes considering using this module are likely already in the weeds when it comes to chutil.
* Tracking that will likely exceed memory constraints. I can see what can be done here.
* I think this goes with point 2, where overheard hoptimization will start reverting hops back on un-updated nodes.

I agree about the originator to be considered, but to do that we'll need to increase the size of nodedb. The meshes that would want this are likely much larger than than MAX_NUM_NODES, making knowing if a node is a SENSOR or TRACKER role difficult.

I think we can look into this as a solution for now for large meshes until someone can think of something better. It's a bit heavy handed, but I think that's necessary for now until more clients are updated with better congestion management.

h3lix1:
@thebentern @NomDeTom What are the blockers left to getting this into .21 or .22? This with #9921 puts a good guard rail in place for misconfiguration on small networks that don't need it.

PR #9363, opened 2026-01-19
Allow using Spreading Factors 5 and 6

derpyspike:
Modified AdminModule.cpp to allow setting Spreading Factor 5 and 6 in custom presets.
In modern LoRa chips like the SX1262 or LR1110 these seem to work without issues and I have tested both SF5 and SF6 extensively with Meshtastic (with this AdminModule change) even with 30 km links and everything seems to work fine.

## 🤝 Attestations

  - [X] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [X] Heltec V4
  - [X] Heltec  E290
  - [X] Station G2


thebentern (approved):

caveman99:
@derpyspike can you please create a PR against documentation, detailing the 2 new spread factors and that they don't work in 1st gen chips.

src/modules/AdminModule.cpp
```diff
@@ -767,7 +767,7 @@ void AdminModule::handleSetConfig(const meshtastic_Config &c)
             validatedLora.coding_rate = 5;
         }
 
-        if (validatedLora.spread_factor < 7 || validatedLora.spread_factor > 12) {
+        if (validatedLora.spread_factor < 5 || validatedLora.spread_factor > 12) {
```
caveman99: can you check for a 1st gen radio (SX127x, RF95) and refuse 5 and 6 in that case?

caveman99 (changes requested):

src/modules/AdminModule.cpp
derpyspike: @caveman99 I've updated the check to check for these, I've also included the detect/LoRaRadioType.h in the includes to use the radioType for this check.

derpyspike:
@caveman99 OK, change made and docs PR opened.

meshtastic/meshtastic#2294

derpyspike:
@thebentern Hi, I see you added conflicting logic to check for the Spreading Factor in your #9808 PR, should I integrate my check into that one? Or modify the LORA_SF_MIN on yours and leave mine as a separate check?

derpyspike:
Closed, and reopened in #10160

PR #9374, opened 2026-01-20
Remove GPS Baudrate locking for Seeed Xiao S3 Kit

### LINKED ISSUE #9373: [Bug]: Seeed Xiao S3 fixed GPS baud rate
### Category

Hardware Compatibility

### Hardware

DIY

- [ ] InkHUD ePaper

### Firmware Version

2.7.15.567b8ea

### Description

The Seeed Xiao ESP32S3 has a fixed baud rate for GPS modules, which makes so that any other GPS module cannot connect. See https://github.com/meshtastic/firmware/blob/eefc08087d66b3b4298b76d9f8d8803ae372fe30/variants/esp32s3/seeed_xiao_s3/variant.h#L53

This is the same issue as in https://github.com/meshtastic/firmware/issues/7012

### Relevant log output

```Shell

```

fifieldt:
The Seeed Xiao S3 Kit's default GPS is an L76K which operates at 9600 baud, so when this variant was defined that baud rate was specified.

However, this is a development board and it is expected that users can attach their own devices. This includes GPS, which may operate at a different baud rate. The current fixed baud rate prevents this, so this patch removes that setting.

This will revert to the regular automatic probe method. This will successfully detect the L76K as before (the same speed as before since 9600 baud is the first baud rate checked), but also allow other GPSes at other baud rates to be detected.

Thanks to @ScarpMarc for the report

Fixes https://github.com/meshtastic/firmware/issues/9373

### DIFF
--- a/variants/esp32s3/seeed_xiao_s3/variant.h
+++ b/variants/esp32s3/seeed_xiao_s3/variant.h
@@ -50,7 +50,6 @@ L76K GPS Module Information : https://www.seeedstudio.com/L76K-GNSS-Module-for-S
 #define GPS_RX_PIN 44
 #define GPS_TX_PIN 43
 #define HAS_GPS 1
-#define GPS_BAUDRATE 9600
 #define GPS_THREAD_INTERVAL 50
 #define PIN_SERIAL1_RX PIN_GPS_TX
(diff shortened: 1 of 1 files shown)


fifieldt:
2 months with no review, going to smash merge.

PR #9428, opened 2026-01-25
feat(power): automatic WiFi disable when running on battery

### LINKED ISSUE #9427: [Feature Request]: Automatic WiFi disable when running on battery
## Use Case

**Docked mode (at home, charging):**
- Node connected to USB power
- WiFi enabled for web UI access from PC, MQTT, remote management

**Portable mode (on the go):**
- Node running on battery
- WiFi not needed — phone connects via BLE
- WiFi wastes ~100-150mA for no benefit

Currently, users must manually toggle WiFi settings when switching between these modes.

## Proposed Behavior

Add new configuration options:

- `network.wifi_on_external_power_only` (bool) — enable automatic WiFi management based on power source
- `network.wifi_power_loss_timeout_secs` (uint32, default: 30) — delay before disabling WiFi after power loss

When enabled:

1. Monitor power source using existing USB/external power detection
2. Automatically enable WiFi when external power is connected
3. Wait for `wifi_power_loss_timeout_secs` after power loss before disabling WiFi (prevents accidental disconnects from briefly unplugged cable)
4. If power is restored within timeout — cancel WiFi disable

## Expected Benefits

- Significant battery life improvement when portable
- Seamless transition between docked and portable modes
- No manual configuration changes needed

## Platform

Cross-Platform (ESP32, RP2040, Linux Native — platforms with WiFi support)

lexfrei:
## Summary

Implement automatic WiFi management based on power source. When enabled, WiFi is automatically disabled when running on battery and re-enabled when external power is connected.

- Add `wifi_on_external_power_only` config option to enable feature
- Add `wifi_power_loss_timeout_secs` config option for delay before disabling WiFi (default: 30s)
- Implement proper WiFi hot toggle without device reboot

## Use Case

**Docked mode (at home, charging):**
- Node connected to USB power
- WiFi enabled for web UI access from PC, MQTT, remote management

**Portable mode (on the go):**
- Node running on battery
- WiFi not needed — phone connects via BLE
- WiFi wastes ~100-150mA for no benefit

### WiFi Service Cleanup (hot toggle support)
- Add `deinitWebServer()` to stop HTTPS/HTTP servers
- Add `stop()` method to `UdpMulticastHandler`
- Add `deinitWifiServices()` to properly shutdown all WiFi-dependent services (mDNS, NTP, Syslog, WebServer, API Server)
- Reset `APStartupComplete` flag to allow service reinitialization on reconnect

### Power-Based WiFi Control
- `handleWifiPowerManagement()` in `Power.cpp` monitors USB power state
- Timer-based approach prevents WiFi toggle on brief power interruptions
- WiFi automatically re-enables when external power is restored

## Dependencies

- Protobufs PR: https://github.com/meshtastic/protobufs/pull/851

## Test Plan

- [ ] Build for ESP32-S3 target
- [ ] Verify WiFi connects normally with external power
- [ ] Disconnect USB, wait 30s — verify WiFi disabled
- [ ] Reconnect USB — verify WiFi re-enables
- [ ] Test "accidental unplug" — disconnect and reconnect within 30s — WiFi should stay on
- [ ] Monitor heap usage for memory leaks

## Platforms

- [x] ESP32 (tested build: seeed-xiao-s3)
- [ ] RP2040 (untested, should work)
- [ ] Linux Native (untested, should work)

Closes: https://github.com/meshtastic/firmware/issues/9427

### DIFF
--- a/src/mesh/wifi/WiFiAPClient.cpp
+++ b/src/mesh/wifi/WiFiAPClient.cpp
@@ -248,12 +248,47 @@ bool isWifiAvailable()
     }
 }
 
+// Deinit all WiFi-dependent services (called before WiFi off)
+static void deinitWifiServices()
+{
+    LOG_DEBUG("Deinit WiFi services");
+
+    // Disable syslog
+    syslog.disable();
+
+    // Stop API server
+    deInitApiServer();
+
+#ifdef ARCH_ESP32
+#if !MESHTASTIC_EXCLUDE_WEBSERVER
+    // Stop web server
+    deinitWebServer();
+#endif
+
+    // Stop mDNS
+    MDNS.end();
+#endif
+
+#ifndef DISABLE_NTP
+    // Stop NTP client
+    timeClient.end();
+#endif
+
+    // Reset flag so services reinitialize on reconnect
+    APStartupComplete = false;
+
+    LOG_INFO("WiFi services stopped");
+}
+
 // Disable WiFi
 void deinitWifi()
 {
     LOG_INFO("WiFi deinit");
 
     if (isWifiAvailable()) {
+        // First stop all services that depend on WiFi
+        deinitWifiServices();
+
 #ifdef ARCH_ESP32
         WiFi.disconnect(true, false);
 #elif defined(ARCH_RP2040)
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -30,7 +30,7 @@
 #include "input/LinuxInputImpl.h"
 #endif
 
-#if HAS_WIFI
+#if HAS_WIFI && !defined(ARCH_PORTDUINO)
 #include "mesh/wifi/WiFiAPClient.h"
 #endif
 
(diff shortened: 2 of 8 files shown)


src/mesh/wifi/WiFiAPClient.cpp
```diff
@@ -248,12 +248,47 @@ bool isWifiAvailable()
...
+// Deinit all WiFi-dependent services (called before WiFi off)
+static void deinitWifiServices()
+{
+    LOG_DEBUG("Deinit WiFi services");
+
+    // Disable syslog
+    syslog.disable();
+
+    // Stop API server
+    deInitApiServer();
```
aido, line 260: Hi,

Before stopping these services, should the code check if ethernet is currently enabled? Since `initEthernet()` also initializes several of these services, we might want to ensure they aren't killed if a valid ethernet connection is still active, even if WiFi is being deinitialized.

Should a check for `config.network.eth_enabled` be added here and also for `timeClient.end();`?

A separate pull request for "automatic disable **ethernet** when running on battery" can take care of stopping these services if both `config.network.eth_enabled` and `config.network.wifi_on_external_power_only` are set.

lexfrei:
Replaced by #11911, since GitHub would not let me reopen this one after the fork it came from spent time archived.

Two changes against what is here: the service teardown now leaves everything running while an Ethernet link is up, per @aido's review above, and the radio is suspended rather than deinited, because the reconnect timer reads config.network.wifi_enabled and undoes a bare deinitWifi() within a second.

PR #9438, opened 2026-01-26
Move device code from main.cpp to earlyInitVariant

jp-bennett:
Tested on the HaD Communicator.
Tested on the tlora pager.
Tested on the t-deck. Found and fixed an unrelated trackball disaster.
Tested Thinknode-M5

### DIFF
--- a/src/input/TrackballInterruptBase.cpp
+++ b/src/input/TrackballInterruptBase.cpp
@@ -61,11 +61,7 @@ int32_t TrackballInterruptBase::runOnce()
         uint32_t pressDuration = millis() - pressStartTime;
         bool buttonStillPressed = false;
 
-#if defined(T_DECK)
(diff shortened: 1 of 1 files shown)


jp-bennett:
Could probably use a quick test with the actual devices. I'll update as I test with my devices (starting tomorrow).

src/graphics/Screen.h
```diff
@@ -103,6 +103,11 @@ class Screen
 #include <string>
 #include <vector>
 
+#ifdef ELECROW_ThinkNode_M5
```
NomDeTom: Is this an opportunity to make the PCA9557 a generic define?

jp-bennett: In retrospect, it probably belongs in main.h

PR #9440, opened 2026-01-26
Trackball revamp

jp-bennett:
The trackball on the T-Deck has always been a bit terrible on BaseUI. This is because we've been treating it too much as a button. This PR does a bit of general cleanup, but on the Trackball it counts GPIO pulses and triggers an event when the pulse count exceeds a threshold. Seems to greatly improve the experience.


src/input/TrackballInterruptBase.cpp
```diff
@@ -55,6 +55,20 @@ int32_t TrackballInterruptBase::runOnce()
 {
     InputEvent e = {};
     e.inputEvent = INPUT_BROKER_NONE;
+#if TB_THRESHOLD
+    if (lastInterruptTime && lastInterruptTime < millis() - 1000) {
```
thebentern: I believe there is a helper in Throttle.cpp to clean this up

Issue #3910, opened 2024-05-16
[Feature Request]: Battery Chemistry Type Selection


CheeseIsGood27:
### Platform

NRF52, ESP32, RP2040, Linux Native, other

### Description

Feature request to add a drop-down menu or way to select battery chemistry for INA219, INA260, or INA3221 sensor under the power config menu.

Lines 14-28 of power.h seem to define several battery chemistry types under the "OCV_ARRAY" variable:

`#ifndef OCV_ARRAY
#ifdef CELL_TYPE_LIFEPO4
#define OCV_ARRAY 3400, 3350, 3320, 3290, 3270, 3260, 3250, 3230, 3200, 3120, 3000
#elif defined(CELL_TYPE_LEADACID)
#define OCV_ARRAY 2120, 2090, 2070, 2050, 2030, 2010, 1990, 1980, 1970, 1960, 1950
#elif defined(CELL_TYPE_ALKALINE)
#define OCV_ARRAY 1580, 1400, 1350, 1300, 1280, 1250, 1230, 1190, 1150, 1100, 1000
#elif defined(CELL_TYPE_NIMH)
#define OCV_ARRAY 1400, 1300, 1280, 1270, 1260, 1250, 1240, 1230, 1210, 1150, 1000
#elif defined(CELL_TYPE_LTO)
#define OCV_ARRAY 2700, 2560, 2540, 2520, 2500, 2460, 2420, 2400, 2380, 2320, 1500
#else // LiIon
#define OCV_ARRAY 4190, 4050, 3990, 3890, 3800, 3720, 3630, 3530, 3420, 3300, 3100
#endif
#endif`

REQUEST: A drop-down menu, or even a few lines to define a custom "OCV_ARRAY" value in the power config would be very useful for alternative battery chemistries.

CONCERN: Most boards used for Meshtastic have an on-board LiPo charge circuit wired direct to an ADC pin for voltage reading. I don't think the modified "OCV_ARRAY" would cause an issue related to the ADC circuit and voltage divider, but I'm uncertain.

[...]

PR #9442, opened 2026-01-26
Remove the unused OCV_ARRAYs and move one to a variant.h

jp-bennett:
Most of the OCV_ARRAY entries were completely unused. Remove these, and move the one non-default to its proper variant.h
Custom hardware should define this string in variant.h

### DIFF
--- a/src/power.h
+++ b/src/power.h
@@ -15,19 +15,7 @@
 
 // Device specific curves go in variant.h
 #ifndef OCV_ARRAY
-#ifdef CELL_TYPE_LIFEPO4
-#define OCV_ARRAY 3400, 3350, 3320, 3290, 3270, 3260, 3250, 3230, 3200, 3120, 3000
-#elif defined(CELL_TYPE_LEADACID)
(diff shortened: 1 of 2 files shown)


thebentern (approved):

fifieldt:
I think these were used by DIY hardware people ...

jp-bennett:
> I think these were used by DIY hardware people ... 

I'll add them as examples in the variant readme. It's super easy to just copy the array into a local variant.

shalberd:
@phaseloop @jp-bennett 

this is really bad for people using e.g. variants/nrf52840/rak4631/variant.h with e.g. the CELL_TYPE_LTO battery arguments.
Nothing else is different, it is a custom RAK4631 board.

https://pcb.uart.cz/datasheets/solar-node-revE-datasheet.pdf

The way it was before, with easy to use battery chemistry variables, was much easier.

https://github.com/meshtastic/firmware/issues/3910

Now, one has to manually set the OCV_ARRAY values. Really bad ...

shalberd:
> I'll add them as examples in the variant readme. It's super easy to just copy the array into a local variant.

I think they were not kept somewhere, they are just gone now.

PR #9447, opened 2026-01-27
Add error handling for SPI command failures in LR11x0, RF95, and SX128x interfaces

thebentern:
I dug into some of the reports that the Seeed T1000-E would sometimes mysteriously transmit on 915MHz, despite reporting successfully initializing and being configured for the default LongFast slot on US region. As it turns out, I found an unhandled RadioLib error code (RADIOLIB_ERR_SPI_CMD_FAILED) that we handle in the other radio interface implementations. 

This also suggests that there _could_ be some issues with radio / txco needing additional time to init on boot up. There could also hardware issues at play, but this PR seeks to give the benefit of the doubt by add a short delay and creating an additional retry before giving up. I'm hoping to get some testing from at least one person with affected node(s)


Xaositek:
Sample firmware file has been turned over to user experiencing issues - pending feedback

PR #9456, opened 2026-01-28
Add reply bot module with DM-only responses and rate limiting

mattatat25:
Adds a reply bot module that listens for /ping, /hello, and /test commands received via direct messages or broadcasts on the primary channel. The module always replies via direct message to the sender only, reporting hop count, RSSI, and SNR. Per-sender cooldowns are enforced to reduce network spam. Updates include the new module source files and required build configuration changes.

## 🤝 Attestations

  - [X] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [X] RAK WisBlock 4631
  - [X] Seeed Studio Wio Tracker L1


Xaositek (approved):

mattatat25:
image

max-arnold:
Could the bot have an option (runtime or compile time) to reply in the channel? Other people on the channel actually might prefer to see the replies. I think the following behavior makes some sense:

1. Pings sent as a DM message get replies as a DM (without auto favoriting the sender)
2. Pings sent in any channel (not just the primary one) get replies in the same channel

platformio.ini
```diff
@@ -50,6 +50,7 @@ build_flags = -Wno-missing-field-initializers
 	-DRADIOLIB_EXCLUDE_APRS=1
 	-DRADIOLIB_EXCLUDE_LORAWAN=1
 	-DMESHTASTIC_EXCLUDE_DROPZONE=1
+	-DMESHTASTIC_EXCLUDE_REPLYBOT=1
 	-DMESHTASTIC_EXCLUDE_REMOTEHARDWARE=1
```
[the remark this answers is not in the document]
mattatat25: Will update the docs

src/modules/ReplyBotModule.cpp
```diff
@@ -0,0 +1,183 @@
...
+    replybotCooldownIdx = (replybotCooldownIdx + 1) % REPLYBOT_COOLDOWN_SLOTS;
+    return false;
+}
+
+// Constructor – registers a single text port and marks the module promiscuous
+// so that broadcast messages on the primary channel are visible.
+ReplyBotModule::ReplyBotModule() : SinglePortModule("replybot", meshtastic_PortNum_TEXT_MESSAGE_APP)
+{
+    isPromiscuous = true;
+}
```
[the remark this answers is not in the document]
mattatat25: This is intentional. The module is disabled by default, unless it’s
explicitly enabled, so I didn’t add it to the global setup.

mattatat25:
> Could the bot have an option (runtime or compile time) to reply in the channel? Other people on the channel actually might prefer to see the replies. I think the following behavior makes some sense:
> 
> 1. Pings sent as a DM message get replies as a DM (without auto favoriting the sender)
> 2. Pings sent in any channel (not just the primary one) get replies in the same channel

This is my first PR, so I kept the idea pretty small and went with DM only
replies, to avoid extra channel noise. I agree an option to reply in-channel
would be fun to explore.

mattatat25:
As of 1/28/26 it's verified working on the Heltec V3, working to verify other devices in my free time.

mattatat25:
1/29/26 Verified on Seeed Studio Wio Tracker L1.

thebentern (approved):

mattatat25:
Docs will be coming soon! Stay tuned!

shalberd:
A bot on channel 0, at least on the one with the well-known key, would be a very bad idea, in my opinion. For replies as well as reacting to bot commands. In fact, it would, especially for shared infrastructure mountain nodes, be preferable to have a way to disable bot-related traffic altogether. Rate-limiting is good, but is the lookup table large enough for e.g. a network of 600 plus nodes?
I like the idea of a bot for certain use cases, I just think we need to balance the needs of all interest groups.
Here, we have very limited frequency slots, on EU_868 ... always important to keep that in mind.

src/modules/ReplyBotModule.cpp
```diff
@@ -0,0 +1,183 @@
...
+/*
+ * ReplyBotModule.cpp
+ *
+ * This module implements a simple reply bot for the Meshtastic firmware.  It listens for
+ * specific text commands ("/ping", "/hello" and "/test") delivered either via a direct
+ * message (DM) or a broadcast on the primary channel.  When a supported command is
+ * received the bot responds with a short status message that includes the hop count
+ * (minimum number of relays), RSSI and SNR of the received packet.  To avoid spamming
+ * the network it enforces a per‑sender cooldown between responses.  By default the
+ * module is enabled; define MESHTASTIC_EXCLUDE_REPLYBOT at build time to exclude it
```
shalberd, line 11: is it really enabled be default? I see it set to one, that is, exclude=true. 
line 53 in platformio.ini

Does this not mean it is disabled by default, for now?

Xaositek: There is a separate PR that I wrote that fixes the documentation of this functionality. No it is not enabled by default

shalberd: good, thank you, and thank you for fixing the documentation. Have a good day

PR #9469, opened 2026-01-29
Nodes Rendering - Cache/page based rendering

ItsLimitlezz:
I’ve redesigned the way nodes are displayed on the nodes page in MUI to enhance the overall experience.

Instead of continuously updating a list of every node, the system now creates a cache that gets refreshed every 10 seconds. This cache is then used to render pages with a limited number of note buffers. The idea is to render a single page at a time, rather than rendering multiple pages simultaneously. This approach allows for a smoother flow between pages and improves performance.

I’ve tested this system with a few others, and they’ve confirmed that it indeed improves performance. While it may not provide a high frame rate experience, it makes browsing through a node list of up to 250 nodes much more efficient.

It’s worth noting that the nodes page may be slow for the first minute after booting up. However, it becomes much quicker after that initial period.

To be clear most of this code was done vibe coding. 
## 🤝 Attestations

- [Y] I have tested that my proposed changes behave as described.
- [Y] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [X] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/graphics/draw/NodeListRenderer.cpp
+++ b/src/graphics/draw/NodeListRenderer.cpp
@@ -564,12 +564,6 @@ void drawNodeListScreen(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t
                 continue;
             if (locationScreen && !n->has_position)
                 continue;
-            if (!n)
-                continue;
-            if (n->num == nodeDB->getNodeNum())
-                continue;
-            if (locationScreen && !n->has_position)
-                continue;
 
             cachedFilteredNodesList.push_back(n->num);
         }
@@ -634,8 +628,6 @@ void drawNodeListScreen(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t
         // Scroll Popup Overlay
         if (millis() - popupTime < POPUP_DURATION_MS) {
             popupTotal = totalEntries;
-
-            int perPage = visibleNodeRows * totalColumns;
 
             popupStart = startIndex + 1;
             popupEnd = std::min(startIndex + perPage, totalEntries);


src/graphics/draw/NodeListRenderer.h
```diff
@@ -63,6 +63,10 @@ void drawColumns(OLEDDisplay *display, int16_t x, int16_t y, const char **fields
 void scrollUp();
 void scrollDown();
 
+// Infinite scroll management (internal use, called by drawNodeListScreen)
+// These functions manage dynamic loading/unloading of nodes based on scroll position
+// to optimize memory usage for large node lists
+
```
Xaositek: This appears to be tied to no additional code?

platformio.ini
```diff
@@ -120,7 +120,7 @@ lib_deps =
 [device-ui_base]
 lib_deps =
 	# renovate: datasource=git-refs depName=meshtastic/device-ui packageName=https://github.com/meshtastic/device-ui gitBranch=master
-	https://github.com/meshtastic/device-ui/archive/69739b84f87a91568d3c421498bc89977937a141.zip
+	/Users/juanpena/Documents/Git/device-ui-Limitlezz
```
Xaositek: Probably don't need to update this.

bin/org.meshtastic.meshtasticd.metainfo.xml
```diff
@@ -87,7 +87,7 @@
   </screenshots>
 
   <releases>
-    <release version="2.7.19" date="2026-01-22">
+    <release version="2.7.19" date="2026-01-29">
```
Xaositek: Should not be part of the PR

debian/changelog
```diff
@@ -2,7 +2,7 @@ meshtasticd (2.7.19.0) unstable; urgency=medium
 
   * Version 2.7.19
 
- -- GitHub Actions <github-actions[bot]@users.noreply.github.com>  Thu, 22 Jan 2026 22:17:40 +0000
+ -- GitHub Actions <github-actions[bot]@users.noreply.github.com>  Thu, 29 Jan 2026 01:17:09 +0000
```
Xaositek: Should not be part of the PR

src/graphics/draw/NodeListRenderer.h
ItsLimitlezz: Just notes to remember things

src/graphics/draw/NodeListRenderer.cpp
```diff
@@ -531,60 +540,139 @@ void drawNodeListScreen(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t
...
+                continue;
+            if (n->num == nodeDB->getNodeNum())
+                continue;
+            if (locationScreen && !n->has_position)
+                continue;
+
+            drawList.push_back(n->num);
+        }
+        totalEntries = drawList.size();
+>>>>>>> 01f8f2063 (Optimize node list pagination and rendering PT.2)
```
Xaositek: Looks like there is still some merge conflicts in the submitted code.

src/graphics/draw/NodeListRenderer.cpp
```diff
@@ -531,60 +540,139 @@ void drawNodeListScreen(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t
...
+        int lastNodeY = y;
+        int shownCount = 0;
+        int rowCount = 0;
+
+        for (int idx = startIndex; idx < endIndex; idx++) {
+<<<<<<< HEAD
+            uint32_t nodeNum = cachedFilteredNodesList[idx];
+=======
+            uint32_t nodeNum = drawList[idx];
+>>>>>>> 01f8f2063 (Optimize node list pagination and rendering PT.2)
```
Xaositek: Looks like there is still some merge conflicts in the submitted code.

src/graphics/draw/NodeListRenderer.cpp
```diff
@@ -605,70 +702,121 @@ void drawNodeListScreen(OLEDDisplay *display, OLEDDisplayUiState *state, int16_t
...
-
 // =============================
 // Screen Frame Functions
 // =============================
+=======
+
+    // =============================
+    // Screen Frame Functions
+    // =============================
+>>>>>>> 01f8f2063 (Optimize node list pagination and rendering PT.2)
```
Xaositek: Looks like there is still some merge conflicts in the submitted code.

src/graphics/draw/NodeListRenderer.cpp
ItsLimitlezz: Corrected merging issues

fifieldt:
Thanks a lot, great idea :)

PR #9470, opened 2026-01-29
Remove stale variant.h defines

jp-bennett:


jp-bennett:
I feel like this is enough for one PR, and any more would make it hard to review.

PR #9471, opened 2026-01-29
Log incoming client IP addresses for API connections

miketweaver:
## Problem

Currently, only a single client can connect at once to the ApiServer. When multiple clients (e.g., mobile app, web interface, CLI tool) attempt to connect simultaneously, they will compete for the connection, often resulting in confusing disconnection issues. The existing logs only show "Incoming API connection" without identifying which client is connecting, making it difficult to troubleshoot these scenarios.

## Solution

This PR modifies the connection logging in three API server implementations to include the remote client's IP address:
- `ServerAPI.cpp`: Generic API connection handler
- `WiFiServerAPI.cpp`: WiFi-specific connection handler  
- `ethServerAPI.cpp`: Ethernet-specific connection handler

By adding this logging it makes for **easier troubleshooting** to quickly identify which client is connecting or causing connection conflicts.

### Before
```
INFO  | Incoming API connection
INFO  | Incoming wifi connection
INFO  | Incoming ethernet connection
```

### After
```
INFO  | Incoming API connection from 192.168.1.100
INFO  | Incoming wifi connection from 192.168.1.100
INFO  | Incoming ethernet connection from 192.168.1.100
```

## Testing

I have tested these changes on the following hardware:
- ✅ M5Stack Cardputer ADV (ESP32-S3)
- ✅ Portduino

Minimal testing, but it worked and I didn't see anything wrong.
- ✅ Heltec V3 (minimal testing)
- ✅ Xiao nRF kit (minimal testing)

All devices correctly log the connecting client's IP address when API connections are established via WiFi and Ethernet interfaces.

**Note:**
I do not have an ipv6 network to test with and I am unsure what this would do in that situation. This is UNTESTED with ipv6

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [x] RAK WisBlock 4631 (Xiao nRF kit)
  - [x] Other: M5Stack Cardputer ADV (ESP32-S3), Portduino

### DIFF
--- a/src/modules/StoreForwardModule.cpp
+++ b/src/modules/StoreForwardModule.cpp
@@ -513,7 +513,7 @@ bool StoreForwardModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp,
             LOG_DEBUG("StoreAndForward_RequestResponse_ROUTER_BUSY");
             // retry in messages_saved * packetTimeMax ms
             retry_delay = millis() + getNumAvailablePackets(this->busyTo, this->last_time) * packetTimeMax *
-                                         (meshtastic_StoreAndForward_RequestResponse_ROUTER_ERROR ? 2 : 1);
+                                         (p->rr == meshtastic_StoreAndForward_RequestResponse_ROUTER_ERROR ? 2 : 1);
         }
         break;
 


miketweaver:
Closed to research this. I'll open a new one when I'm ready. I'm not sure this handles ipv6 either
https://github.com/meshtastic/firmware/blob/b2f2f6b305e9cdc66fdb129fd6795b642fd96c2f/src/mesh/udp/UdpMulticastHandler.h#L44-L50

PR #9481, opened 2026-01-30
Add custom ringtone definition for RAK4631 and enable buzzer pin

thebentern:
This *shouldn't* interfere with anything as far as I can tell. Credit to @tropho23 for the ringtone RTTL which works on the resonant frequency for the RAK wisblock buzzer and doesn't get the buzzer HOT!


fifieldt (approved):
If it works on that device shouldn't change anything else, should be safe

PR #9490, opened 2026-01-30
re-enable RTC support on THINKNODE M3 and M6

jp-bennett:
M3 and M6 RTC support was pulled due to the ambiguity around how HAS_RTC works. These devices do have a working RTC, so re-adding the defines.


jp-bennett:
Closing as this should go in master

Issue #9406, opened 2026-01-23
[Bug]: The text is too large on the Heltec E213


ccchdro:
### Category

Other

### Hardware

Heltec Vision Master E213

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [x] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.18 and older version

### Description

The text is too large on the Heltec E213

![Image](https://github.com/user-attachments/assets/484b1e5f-7c29-46d6-9c47-6b91da9aa46e)

![Image](https://github.com/user-attachments/assets/cded69a8-00a3-4a61-9dc5-6f36c101f16c)

![Image](https://github.com/user-attachments/assets/f70ca582-c070-4245-9507-dccda78b8833)

### Relevant log output

```Shell

```

herbenderbler:
Hey all!  I came across this bug while perusing the `first contribution` label and thought I might be able to help.  I've taken a first pass at this and have a PR put together.  I'll be sure to push it today as soon as I make sure I adhere to the contribution requirements documented.

PR #9495, opened 2026-01-31
BUGNUM-9406: Decreased text size for the Heltec E213

herbenderbler:
Problem Statement:

Bug Number [#9406](https://github.com/meshtastic/firmware/issues/9406) mentioned an issue with text size being too large on the Heltec E213.  This PR is my attempt at fixing that issue.  I unfortunately don't have the affected hardware to test my code changes, so please keep that in mind.  I chose to take the font size one step down, in hopes of alleviating this issue without being too heavy-handed about my choice.

## 🤝 Attestations

- [] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/bin/check-all.sh
+++ b/bin/check-all.sh
@@ -13,7 +13,7 @@ if [[ $# -gt 0 ]]; then
 	# can override which environment by passing arg
 	BOARDS="$@"
 else
-	BOARDS="tlora-v2 tlora-v1 tlora_v1_3 tlora-v2-1-1.6 tbeam heltec-v2.0 heltec-v2.1 tbeam0.7 meshtastic-diy-v1 rak4631 rak4631_eink rak11200 t-echo canaryone pca10059_diy_eink"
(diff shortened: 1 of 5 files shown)


Xaositek:
The one concern I have with this change is that the reported issue is in BaseUI, nor InkHUD and your change impacts only InkHUD. I think @HarukiToreda has this hardware and is looking into fixing this up in time.

herbenderbler:
> The one concern I have with this change is that the reported issue is in BaseUI, nor InkHUD and your change impacts only InkHUD. I think @HarukiToreda has this hardware and is looking into fixing this up in time.

I'm happy to fix my PR and move these changes to BaseUI, but it's really your call.

Xaositek:
Haruki has one of these devices so let's leave it to him as he can test it live before committing

PR #9504, opened 2026-02-01
BaseUI: Implementation of Status Message for Favorite and NodeList views

Xaositek:
- Add status message to Favorite Frame
- Add status message to Node List if on LongName view
- Keep 5 Status Messages in the active buffer
- Concatenate long status messages to not overflow available screen space


src/modules/StatusMessageModule.cpp
```diff
@@ -29,10 +29,23 @@ int32_t StatusMessageModule::runOnce()
...
-        meshtastic_StatusMessage incomingMessage;
+        meshtastic_StatusMessage incomingMessage = meshtastic_StatusMessage_init_zero;
+
         if (pb_decode_from_bytes(mp.decoded.payload.bytes, mp.decoded.payload.size, meshtastic_StatusMessage_fields,
                                  &incomingMessage)) {
+
             LOG_INFO("Received a NodeStatus message %s", incomingMessage.status);
+
+            RecentStatus entry;
+            entry.fromNodeId = mp.from;
```
jp-bennett, line 40: If we're just displaying for favorite nodes, suggest only storing the string for favorite nodes as well.

Xaositek: Since we are showing on both Favorite Screen and Node List, we should probably leave for now?

Xaositek:
IMG_1993
IMG_1994

Xaositek:
IMG_1997
IMG_1998

src/modules/StatusMessageModule.h
```diff
@@ -19,16 +20,28 @@ class StatusMessageModule : public SinglePortModule, private concurrency::OSThre
...
+
+    const std::vector<RecentStatus> &getRecentReceived() const { return recentReceived; }
+
   protected:
     /** Called to handle a particular incoming message
      */
     virtual ProcessMessage handleReceived(const meshtastic_MeshPacket &mp) override;
 
   private:
+    static constexpr size_t MAX_RECENT_STATUSMESSAGES = 20;
```
jp-bennett: This ends up being quite a bit of memory on constrained devices. I'm still of the opinion this should be reduced, and possibly limited to devices that are favorited.

Xaositek: I originally set it at 5, I can move it back but I think display wise, there's more value of the implementation if we use it on more than just the Favorite frame.

caveman99 (approved):

fifieldt (approved):

PR #9509, opened 2026-02-02
fix(power): reset LPF when adc_multiplier_override changes at runtime

### LINKED ISSUE #9508: power.adc_multiplier_override runtime changes have no immediate effect
## Device Information

- **Firmware version**: 2.7.18.4744010
- **Device**: Seeed XIAO ESP32S3 (`seeed-xiao-s3`)
- **Platform**: ESP32-S3

## Bug Description

When changing `power.adc_multiplier_override` at runtime via CLI or Admin API, the setting is saved correctly but has no visible effect on voltage readings for 1-2 minutes. Compile-time `ADC_MULTIPLIER` in variant.h works correctly and immediately.

## Steps to Reproduce

1. Device has compile-time `ADC_MULTIPLIER 2.0` (default)
2. Battery shows ~4.08V at full charge
3. Set runtime override: `meshtastic --set power.adc_multiplier_override 2.06`
4. Verify setting saved: `meshtastic --get power.adc_multiplier_override` returns `2.059999...`
5. Reboot device
6. Wait 2+ minutes
7. Check voltage: still shows ~4.08V instead of expected ~4.20V

**Expected**: Voltage should show ~4.20V after multiplier change (or converge quickly)
**Actual**: Voltage remains at ~4.08V for extended period, slowly converges over 1-2 minutes

## Root Cause

Found in `src/Power.cpp`, method `AnalogBatteryLevel::getBattVoltage()`.

The Low Pass Filter (LPF) uses cached `last_read_value` that was calculated with the **old** multiplier. The `initial_read_done` flag is set to `true` on first reading and never reset when multiplier changes.

```cpp
// Line 315-316: Multiplier is correctly read from config
float operativeAdcMultiplier =
    config.power.adc_multiplier_override > 0 ? config.power.adc_multiplier_override : ADC_MULTIPLIER;

// Lines 339-348: But LPF uses old cached value
if (!initial_read_done) {
    if (scaled > last_read_value)
        last_read_value = scaled;
    initial_read_done = true;  // Set to true FOREVER
} else {
    // LPF slowly adapts with 0.5 coefficient - takes 5+ readings to converge
    last_read_value += (scaled - last_read_value) * 0.5;
}
```

With coefficient 0.5 and minimum read interval of 5 seconds, full convergence takes ~1-2 minutes.

## Why compile-time works immediately

When `ADC_MULTIPLIER` is changed in variant.h and firmware is reflashed, device reboots with `initial_read_done = false`, so first reading sets `last_read_value` directly without LPF filtering.

## Proposed Fix

Track the current multiplier and reset the filter when it changes:

```cpp
// Add member variable:
float last_adc_multiplier = 0;

// In getBattVoltage(), after computing operativeAdcMultiplier:
if (last_adc_multiplier != operativeAdcMultiplier) {
    initial_read_done = false;
    last_adc_multiplier = operativeAdcMultiplier;
}
```

## Additional Context

The comment in `AdminModule.cpp` (line 702) says:
> "Really just the adc override is the only thing that can change without a reboot"

This implies runtime changes should work, but the LPF filter prevents immediate effect.

lexfrei:
## Summary

- Reset the Low Pass Filter when `power.adc_multiplier_override` changes at runtime
- Track last used ADC multiplier value and reset `initial_read_done` flag when it changes
- Allows runtime calibration to take effect immediately instead of waiting for slow LPF convergence

## Problem

When changing `power.adc_multiplier_override` at runtime, the new value is saved but has no visible effect on battery voltage readings for 1-2 minutes due to the LPF slowly converging to the new value.

The LPF uses a cached `last_read_value` calculated with the old multiplier, and the `initial_read_done` flag (set on first reading) prevents immediate application of the new multiplier.

## Solution

Add a member variable to track the last used multiplier. When the multiplier changes, reset `initial_read_done` to `false`, causing the next ADC reading to initialize the filter directly with the new scaled value.

## Test plan

- [ ] Set `power.adc_multiplier_override` via CLI
- [ ] Verify voltage reading updates immediately (within one ADC read cycle, ~5 seconds)
- [ ] Change multiplier again, verify immediate effect
- [ ] Reboot device, verify multiplier is still applied correctly

Fixes #9508

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -314,6 +314,13 @@ class AnalogBatteryLevel : public HasBatteryLevel
         // Override variant or default ADC_MULTIPLIER if we have the override pref
         float operativeAdcMultiplier =
             config.power.adc_multiplier_override > 0 ? config.power.adc_multiplier_override : ADC_MULTIPLIER;
+
+        // Reset filter if multiplier changed (allows runtime calibration to take effect immediately)
+        if (last_adc_multiplier != operativeAdcMultiplier) {
+            initial_read_done = false;
+            last_adc_multiplier = operativeAdcMultiplier;
+        }
+
         // Do not call analogRead() often.
         const uint32_t min_read_interval = 5000;
         if (!initial_read_done || !Throttle::isWithinTimespanMs(last_read_time_ms, min_read_interval)) {
@@ -520,6 +527,8 @@ class AnalogBatteryLevel : public HasBatteryLevel
     bool initial_read_done = false;
     float last_read_value = (OCV[NUM_OCV_POINTS - 1] * NUM_CELLS);
     uint32_t last_read_time_ms = 0;
+    // Track last used ADC multiplier to reset filter when config changes at runtime
+    float last_adc_multiplier = 0;
 
 #if HAS_TELEMETRY && !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR && defined(HAS_RAKPROT)
 


lexfrei:
Reopened as #11910 — GitHub would not let me reopen this one. The fix there rescales the cached value by the ratio of old and new multiplier instead of resetting the filter, so lowering the multiplier takes effect too.

PR #9510, opened 2026-02-02
Power off control pin on Thinknode m5 during deepsleep and add RTC

jp-bennett:
Adds the variant_shutdown function for esp32, and uses it to power down a pin on the m5 during shutdown.


jp-bennett:
Surprise benefit, dropping the RTC autodetection class saves a bit of ram and flash space, though not a whole lot.

PR #9513, opened 2026-02-03
Inkhud battery icon improvements.

Vortetty:
- Fixes the battery icon draining from the flat side towards the bump, which is backwards from general design language seen on most devices.
- ~~By request of kr0n05_ on discord, adds the ability to mirror the battery icon which fixes that issue in another way, and is also a common design seen on other devices, and fixes the above.~~ removed by developer request

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other: Seeed Wio Tracker L1 E-ink
  
edit: it appears i forgot to change the label from hardware support.. would someone mind fixing that

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -459,6 +459,8 @@ class AnalogBatteryLevel : public HasBatteryLevel
         }
         // if it's not HIGH - check the battery
 #endif
+        // If we have an EXT_PWR_DETECT pin and it indicates no external power, believe it.
+        return false;
 
 // technically speaking this should work for all(?) NRF52 boards
 // but needs testing across multiple devices. NRF52 USB would not even work if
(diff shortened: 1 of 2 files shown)


HarukiToreda:
I suggest we go with the battery graphics with the new change as the default and discard the old one entirely. It does not need to be a settings choice for such a minor change. Let's leave the settings choice for other major future features 

The new one drains from the side the positive side as expected for most other devices
IMG_20260203_090620 

Old one drains from the negative side which looks odd.

ForestedDev:
New commit adds a 1px border around the battery icon to prevent text from rendering under the icon, as well as a small dither similar to that used in the message applet near the header.

Previous behaviour:
ResizedImage_2026-02-03_08-03-12_9596

New behaviour:
ResizedImage_2026-02-03_10-18-52_5979

The border is always rendered to prevent any user applets from rendering under the battery, and the dither is handled in the header rendering code so that applets which don't use the standard header are unaffected.

HarukiToreda (approved):

PR #9519, opened 2026-02-03
Database for telemetry readings

oscgonfer:
This is a proof of concept PR to be elaborated further, but featuring a small database for telemetry. The idea is to create something similar to: https://docs.smartcitizen.me/hardware/firmware/?h=flash#flash-storage

The database contains a set of records in memory, ideally (although not working yet) encoded to protobuf and writable to the FS. This would allow to:

- [ ] Retrieve average/min/max readings for telemetry with admin commands
- [ ] Store delivery status, enforcing retries for certain roles (`SENSOR`)  in which telemetry is interesting
- [ ] Serve as a backup ifor storing data to an sd-card later on (with additional flags)

Current implementation relies on a protobuf definition implemented in https://github.com/meshtastic/protobufs/pull/854
This is simply a PoC, not sure it's a good idea. If moving forward, it'd require work on:

- [ ] Make sure writing / reading works to FS.
- [ ] Writing can be done per `GROUPS`, keeping a separate index for the currently used group, to avoid writing / reading big groups every time.
- [ ] Improvements to recover data from storage to MQTT for certain roles (`SENSOR`), to avoid gaps in received packets.

This is currently disabled by default and needs: `#define MESHTASTIC_EXCLUDE_AIR_QUALITY_DB 0`

## 🤝 Attestations

  - [ ] Seeed Studio ESP32-S3 (partially)


oscgonfer:
I believe this can be closed in favour of #11520, which can be them dumped to memory periodically.

PR #9520, opened 2026-02-03
Add SHT21 sensor support and related configurations

Tomas-Kuchta-FPV:
Hi yall,
I've gotten my hands on the SHT21 sensor for a school project. And have written code for it with the help of Copilot.  
I have no idea if my PR is welcomed here as you already have SHT3x in te codebase which is way more common.  
The code is tested and it works on the target device.  
The sensors shares the same addres with `INA_ADDR` (0x40). I have implemented a flag `-D HAS_SHT21=1`, but maybe for user confort a probe would be better. What do you think?
Thanks for any sudgestions.  

## 🤝 Attestations

- [ x I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [x] DIY_V1 (ESP32)

### DIFF
--- a/src/graphics/draw/MessageRenderer.cpp
+++ b/src/graphics/draw/MessageRenderer.cpp
@@ -877,15 +877,15 @@ void drawTextMessageFrame(OLEDDisplay *display, OLEDDisplayUiState *state, int16
                 // Send Message (Right side)
                 display->drawRect(x1 + 2 - bubbleW, y1 - bubbleH, bubbleW, bubbleH);
                 // Top Right Corner
-                display->drawRect(x1, topY, 2, 1);
+                display->drawRect(x1 - 1, topY, 2, 1);
                 display->drawRect(x1, topY, 1, 2);
                 // Bottom Right Corner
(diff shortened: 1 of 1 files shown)


src/mesh/generated/meshtastic/telemetry.pb.h
```diff
@@ -103,7 +103,9 @@ typedef enum _meshtastic_TelemetrySensorType {
     /* TSL2561 light sensor */
     meshtastic_TelemetrySensorType_TSL2561 = 44,
     /* BH1750 light sensor */
-    meshtastic_TelemetrySensorType_BH1750 = 45
+    meshtastic_TelemetrySensorType_BH1750 = 45,
+    /* STH21 temperature, R. humidity sensor */
+    meshtastic_TelemetrySensorType_SHT21 = 46
```
fifieldt: This code is automatically generated. So another PR to the protobufs repo is needed (with basically the same content)

Tomas-Kuchta-FPV: Alright regenerated and made a PR in protobufs https://github.com/meshtastic/protobufs/pull/855, thanks

PR #9521, opened 2026-02-03
Fix config.display.use_long_node_name not saving

Xaositek:
After forgetting about this for some time, I realized we missed a line in the save which kept config.display.use_long_node_name from saving properly.


Xaositek:
Will redo against `master`

PR #9529, opened 2026-02-04
Remove unused hmx variable

EricSesterhennX41:
The variable is not used at all in the function, remove it to silence the compiler warning.

## 🤝 Attestations

- [x] I have compile tested that my proposed changes behave as described.


fifieldt:
Thanks a lot for the cleanups recently :)

EricSesterhennX41:
> Thanks a lot for the cleanups recently :)

thanks, still getting familiar with the code base and the workflow.

PR #8010, opened 2025-09-16
M5Stack Cardputer Adv

WillyJL:
- Implemented mostly by @Szetya in #7985
- This is NOT the "Meshtastic For Cardputer-Adv" on M5Burner, that firmware is closed source currently in violation of Meshtastic's GPLv3, and M5Stack said "There is no open source code yet, we are still optimizing", see #7989 and the discord for more info
- Whether or not they will opensource and upstream their "optimizations" will remain to be seen, in the meantime I thought I'd open a PR with @Szetya's code since it atleast makes the device work so we can do our own community work on it, I for one want to improve keyboard support for it like in #7989 (feel free to merge either one of these PRs first, I will update the other accordingly with the keyboard improvements for Cardputer Adv shortly after)
- I do not have a device myself to test, am acting as a proxy between someone who does (my friend and youtuber sn0ren) to get feedback
- What I heard is "it works" from sn0ren and @Szetya mentioned sound is not working yet and also that they need to agree with @mverch67 on which display driver to use
- Apologies for opening the PR myself when I had little to do with the development, I just wanted to get things moving to then improve the keyboard for this device too, but feel free to close and open another PR yourself if that works better
- Needs https://github.com/meshtastic/protobufs/pull/776 merged first and then an update to protobufs submodule to match the merged commit

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)
    - [x] LilyGo T-Lora Pager
    - [x] M5Stack Cardputer Adv

PR #9540, opened 2026-02-05
Cardputer Kit

caveman99:
BMI270 WIP


caveman99:
This combines the original work from https://github.com/meshtastic/firmware/pull/8010 and the observations during the review process, as well as findings from https://www.reddit.com/r/CardPuter/comments/1pxjryj/working_meshtastic_firmware_for_the_cardputer_adv/ It adds Support for the accelerometer and sound module.

caveman99:
thanks for all the detective work and constructive discussion on this device to @RaymiiOrg, @mverch67, @WillyJL and @Szetya - i hope to pull in the inprovements from https://github.com/meshtastic/firmware/pull/7989 next.

WillyJL:
> thanks for all the detective work and constructive discussion on this device to @RaymiiOrg, @mverch67, @WillyJL and @Szetya - i hope to pull in the inprovements from #7989 next.

about #7989, i had ported my keyboard changes to the WIP cardputer adv keyboard here: https://github.com/WillyJL/meshtastic-firmware/commit/a7928b1d3363aa9b4523ef7d56361c21b87efdf3

im assuming the keyboard implementation has changed in the meantime so this commit will probably need additional changes, but i hope this commit can help update your pr once 7989 is merged (or update 7989 if this pr is merged first).

thanks for taking over and the further attention, i had only opened my cardputer adv pr to start off but dont even have that device and couldnt help further.

RaymiiOrg:
> thanks for all the detective work and constructive discussion on this device to @RaymiiOrg, @mverch67, @WillyJL and @Szetya - i hope to pull in the inprovements from #7989 next.

Thanks to you we might get first class support for all hardware in the Cardputer! 😄

RaymiiOrg:
It seems that ( with or without ) the led enabled, the screen doesn't turn off its backlight anymore. The top cardputer runs this branch, the bottom runs my old branch. The screen is set to turn off after 30 sec on both. (The bottom cardputer has the ambient led off, but is running)




os

psifertex:
Thanks for the work @caveman99 -- did you happen to check my branch as well which was linked in that thread? A few of QoL features in there as well:

 -  Audio in https://github.com/psifertex/meshtastic-firmware/tree/feat/m5stack-cardputer-adv-audio (but it sounds like you have that covered already)
 - Brightness in settings: https://github.com/WillyJL/meshtastic-firmware/commit/3503e8fabb36c8c47f6f247414fb64bbe1503465
 - Better navigation via context-sensitive hotkeys: https://github.com/WillyJL/meshtastic-firmware/commit/b6384f345ad70afe7228cc9793046ac40fff9ae2

I'll try to check it out when I have more time and re-apply them to this branch if not.

RaymiiOrg:
After a few days of testing it seems that the screen freezes after a while. The app connection (wifi) still works, sending and such, but no response to input (the screen doesn't change and no messages can be sent from the device).

radonslide:
If you set the display to a timeout of, for example, 5 minutes, it will not come back on after pressing a button, etc. The display remains off.

xMasterX:
Hello!

I made patch that combines all work of this branch with @psifertex sound and kb modifications (however I was unable to make it work using m5unified, so I used m5cardputer lib based on unified (strange))

And @WillyJL 's brightness + keyboard navigation fixes ported to latest source 
Also applied fix to brightness since it was making screen go fully off when using lower values

This patch can be applied at develop branch at commit 545826d319553f3bb570cc37a33688a1e172ec6b
(latest for 15 feb)

There are some issues with current implementation in this branch, like ES8311 pins are not correct
This is not relevant for the patch since it uses other lib to work with ES chip
just noticed that while was working on it

[cardADV_full_support.patch](https://github.com/user-attachments/files/25328288/cardADV_full_support.patch)

(Can't make PR to the existing PR, so if possible - apply that on top of this PR, or make new (anybody can do, I'm too busy with other things 😢 ))

Also display timeout and backlight issues described in comments above, are not present with those changes applied, or I fixed them along the way and haven't noticed :))))

Keymap hint for those who wanna test:

fn+t - test beep sound
fn+y - test melody sound rttl
fn+g - disable/enable gps hardware
fn+q - reboot the meshtastic
fn+b - enable/disable bluetooth
fn+m - mute
fn+p - send ping

xMasterX:
Update:
There was an issue with sound init in previous patch and original @psifertex code
`es8311` needs to be "asked" to start working via i2c
Especially - we need to feed whole table of parameters in it
Sound was working only if other firmware was used and device had no power off between reflashing to meshtastic build
ES chip stayed at its config and we was getting sound, one power cycle - no more sound

I fixed that issue, here is final updated patch:
[cardputeradv_with_fixed_sound.patch](https://github.com/user-attachments/files/25376350/cardputeradv_with_fixed_sound.patch)

xMasterX:
@jp-bennett 
Hi, please check my last patch in the comment, it has no issue with screen wakeup, it seems like double work is being done

RaymiiOrg:
Commit [d1c406b](https://github.com/meshtastic/firmware/pull/9540/commits/d1c406bc5cbec3d499346b4fd2b20fc518c1bd6b) seems to fix the unresponsive screen after a while. The screen however still does not turn off (in my case after 1 minute). It stays on the last screen it was (before it would have turned off). Pressing a key does flash the screen and then it responds like normally.

RaymiiOrg:
After the last 3 commits (c0f289e, de04d90, 32521d4), the display is now off when it should be, but the backlight is still on (while the display is off). (with or without ambient led disabled.)

src/modules/CannedMessageModule.cpp
```diff
@@ -858,8 +867,7 @@ bool CannedMessageModule::handleFreeTextInput(const InputEvent *event)
     // All hardware keys fall through to here (CardKB, physical, etc.)
 
     if (event->kbchar == INPUT_BROKER_MSG_EMOTE_LIST) {
-        runState = CANNED_MESSAGE_RUN_STATE_EMOTE_PICKER;
-        requestFocus();
+        updateState(CANNED_MESSAGE_RUN_STATE_EMOTE_PICKER);
```
thebentern, line 873: Does this one need to preserve the requestFocus()?

src/modules/CannedMessageModule.cpp
```diff
@@ -457,6 +453,21 @@ int CannedMessageModule::handleInputEvent(const InputEvent *event)
...
+void CannedMessageModule::updateState(cannedMessageModuleRunState newState)
+{
+    runState = newState;
+    if (runState == CANNED_MESSAGE_RUN_STATE_FREETEXT) {
+        inputBroker->menuMode =
+            false; // Allow any key input to be sent to the message composer instead of being interpreted as menu navigation
+    } else {
+        inputBroker->menuMode = true; // Re-enable menu navigation for destination selection
+    }
+    if (runState == CANNED_MESSAGE_RUN_STATE_ACTIVE || runState == CANNED_MESSAGE_RUN_STATE_FREETEXT ||
```
thebentern: This isn't maintainable. We need an overload request focus by callers instead of maintaining a list of states inside the method. I'll push an update

RaymiiOrg:
wanna-know-ur-feedback-about-meshtastic-performance-on-v0-k43il5vrq9ug1

https://x.com/M5Stack/status/2035004082373742650

More info being tested about the actual kit.

PR #9543, opened 2026-02-05
Ebyte E22P module support for existing PCBs. New Hydra E22P variant.

And137:
Hydra E22P - enables support for the Ebyte E22P-868M30S and E22P-915M30S modules.

This variant allows the use of standard board designs for Hydra without having to modify the RX control connection.

New definition added: SX126X_ME
If defined, captures the SX126X_RXEN pin and sets the state to always ON.

Added condition in SX126xInterface.cpp that checks the X126X_ME definition in variant.h. It allows for creating other board variants for the Ebyte E22P module based on the E22 template.

WARNING: Do not flash Hydra-E22P variant on a standard E22 Hydra, as it may result in hardware damage by setting the RX and TX pin states to "1" at the same time!

### DIFF
--- a/variants/esp32s3/heltec_v4/platformio.ini
+++ b/variants/esp32s3/heltec_v4/platformio.ini
@@ -124,11 +124,8 @@ build_flags =
   -D SCREEN_TOUCH_RST=TOUCH_RST_PIN
 
 lib_deps = ${heltec_v4_base.lib_deps}
-  ; ${device-ui_base.lib_deps}
+  ${device-ui_base.lib_deps}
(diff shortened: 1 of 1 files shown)


NomDeTom:
This just needs an alternative line in the main hydra FW to add an ifdef E22P - and then just use pinmode high for the equivalent of the rxen pin. No need to mess with the main FW. Variant specific things should stay in the variant.

And137:
All changes are now contained within the Hydra variant. E22P support is now avaliable as a build flag “-D USE_EBYTE_E22P”. Should I open a new pull request?

PR #9029, opened 2025-12-20
🔧 Fix LNA/PA power control for Heltec v4, wireless tracker v2

weebl2000:
Root Cause

UPDATE 23/12/2025:
Actually, we should just the SX1262 chip handle the RF path. Operating pin 46 as a TX switch is not the right way, it should be treated as CPS pin.

Key fixes:
1. Removed double-control: Removed SX126X_TXEN=LORA_PA_TX_EN which caused both RadioLib and setTransmitEnable() to fight over GPIO46
2. Corrected documentation: GPIO46 is CPS (PA mode select), NOT the TX/RX switch. DIO2 handles TX/RX path via CTX
3. Added GC1109 datasheet control logic to variant headers

~~The GC1109 FEM (Front-End Module) chip requires specific control logic:~~
~~- PA_EN (pin 2): Must be HIGH for both RX and TX modes~~
~~- PA_TX_EN (pin 46): RF switch control - LOW=RX path (LNA), HIGH=TX path (PA)~~

~~The bugs prevented proper RX operation, causing 10-20 dB sensitivity loss.~~

## 🤝 Attestations

I haven't tested yet, but derived from [HERE](https://github.com/meshcore-dev/MeshCore/pull/1249). But someone please do.

- [ ] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V4

PR #9572, opened 2026-02-08
Hold GC1109 FEM power during deep sleep for LNA RX wake

weebl2000:
## Summary

- Use `rtc_gpio_hold_en` to latch `PA_POWER` (LDO) and `PA_EN` (CSD) HIGH during deep sleep, keeping the GC1109 LNA powered for wake-on-packet RX
- Skip these pins in the blanket RTC hold release on wake (`initDeepSleep`)
- Release them in `SX126xInterface::init()` **after** setting GPIO registers HIGH first, avoiding a power glitch
- Add 1ms LDO startup delay on cold boot before driving CSD/CPS, per GC1109 power-on sequence requirement (TLV75733P tSTR ~550us)
- Applies to Heltec V4 and Heltec Wireless Tracker V2

## Background

Previously, the GC1109 FEM power pins used weak internal pull-ups (`gpio_pullup_en`) during deep sleep, which is less reliable than explicitly latching the pad state. Additionally, `SX126xInterface::sleep()` sets `PA_EN` LOW before `enableLoraInterrupt()` re-enables it, creating a window where the GC1109 could lose power.

The blanket `rtc_gpio_hold_dis` loop in `initDeepSleep()` also released these pins before `SX126xInterface::init()` had a chance to set them HIGH, causing a brief power glitch on every deep sleep wake.

Without this fix, the GC1109 LNA could be unpowered during sleep, degrading RX wake sensitivity by ~17dB.

The TLV75733P LDO has a startup time (tSTR) of ~550us. On cold boot, `PA_EN` (CSD) was being driven HIGH only ~12.5us after `PA_POWER`, before the LDO output had stabilised. The GC1109 datasheet requires "VBAT must be prior to CSD/CPS/CTX for the power on sequence". On deep sleep wake the LDO is held on via RTC latch, so no delay is needed.

**Trade-off:** ~1.2mA additional deep sleep current for significantly improved wake-on-packet range.

Reference: https://github.com/meshcore-dev/MeshCore/pull/1600
Reference: https://github.com/meshtastic/firmware/pull/9029 (comments by @compumike, @ascendr, @lindorffs)

## Test plan

- [ ] Flash Heltec V4, enable ROUTER role (triggers deep sleep with LoRa wake)
- [ ] Verify wake-on-packet works at expected range
- [ ] Verify TX still works correctly after wake
- [ ] Measure deep sleep current (~1.2mA extra expected with LNA active)
- [ ] Verify no regression on cold boot (1ms delay is negligible)
- [ ] Confirm normal RX/TX operation after cold boot

### DIFF
--- a/src/mesh/SX126xInterface.cpp
+++ b/src/mesh/SX126xInterface.cpp
@@ -8,6 +8,7 @@
 #endif
 #if defined(USE_GC1109_PA) && defined(ARCH_ESP32)
 #include <driver/rtc_io.h>
+#include <esp_sleep.h>
 #endif
 
 #include "Throttle.h"
@@ -68,6 +69,18 @@ template <typename T> bool SX126xInterface<T>::init()
     pinMode(LORA_PA_POWER, OUTPUT);
     digitalWrite(LORA_PA_POWER, HIGH);
     rtc_gpio_hold_dis((gpio_num_t)LORA_PA_POWER);
+
+    // TLV75733P LDO has ~550us startup time (datasheet tSTR). On cold boot, wait
+    // for VBAT to stabilise before driving CSD/CPS, per GC1109 requirement:
+    // "VBAT must be prior to CSD/CPS/CTX for the power on sequence"
+    // On deep sleep wake the LDO was held on via RTC latch, so no delay needed.
+#if defined(ARCH_ESP32)
+    if (esp_sleep_get_wakeup_cause() == ESP_SLEEP_WAKEUP_UNDEFINED) {
+        delayMicroseconds(1000);
+    }
+#else
+    delayMicroseconds(1000);
+#endif
 
     // CSD (LORA_PA_EN): Chip enable - must be HIGH to enable GC1109 for both RX and TX
     pinMode(LORA_PA_EN, OUTPUT);


compumike:
In support of this PR, I did some more research and found that the **official** `HelTecAutomation/Heltec_ESP32` does include a 1 ms delay after `LORA_PA_EN`, as well as the use of `rtc_gpio_hold_dis((gpio_num_t)LORA_PA_POWER);`:

```c
#if defined(USE_GC1109_PA)
	pinMode(LORA_PA_POWER,OUTPUT);
    digitalWrite(LORA_PA_POWER,HIGH);


    rtc_gpio_hold_dis((gpio_num_t)LORA_PA_EN); // <==============
	pinMode(LORA_PA_EN,OUTPUT);
	digitalWrite(LORA_PA_EN,HIGH);
    delay(1);  // <==============


    pinMode(LORA_PA_TX_EN,OUTPUT);
	digitalWrite(LORA_PA_TX_EN,HIGH);
    delay(2);
#endif
```
https://github.com/HelTecAutomation/Heltec_ESP32/blame/7663bf84c4c54afddb181d64706c7ced77aea0a5/src/radio/radio.c#L889-L901C4

PR #9582, opened 2026-02-09
Add SHT21 sensor support

Tomas-Kuchta-FPV:
## Description

This PR adds support for the SHT21 humidity and temperature sensor to Meshtastic. The SHT21 is a low-cost, low-power environmental sensor that complements the existing SHT3x support.

## Changes

- Added SHT21 sensor driver implementation
- Added telemetry configuration for SHT21 sensor readings
- Integrated sensor with the telemetry module
- Generated corresponding protobuf definitions

## Sensor Details

- **I2C Address**: 0x40 (same as `INA_ADDR`)
- **Control**: Enabled via `-D HAS_SHT21=1` compile flag
- **Use Case**: Environmental monitoring (temperature/humidity)

## Testing

- ✅ Tested on DIY_V1 (ESP32)
- ✅ Sensor reads temperature and humidity correctly
- ✅ No regressions on test device

## Notes

The SHT21 shares the I2C address (0x40) with the INA power monitoring chip. Current implementation uses a compile-time flag to select which sensor is used. Future improvements could include runtime detection/probing if both sensors need to be supported simultaneously.

Related: meshtastic/protobufs#855


src/detect/ScanI2CTwoWire.cpp
```diff
@@ -401,6 +403,14 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
 
                 break;
 
+#if HAS_SHT21
```
fifieldt, line 406: is there a register you can look up on the SHT21 that is unique to that model, or at least able to distinguish it from the other chips that share the address? That way we could detect it and use it without a firmware rebuild.

NomDeTom:
Looking at the datasheets, maybe the INA219 calibration register (05h) being all zero at reset would be enough?

oscgonfer:
Hi @Tomas-Kuchta-FPV - This is likely superseeded by #9859. Did you get a chance to test that one?

oscgonfer:
Closing as superseded by https://github.com/meshtastic/firmware/pull/9859

PR #9590, opened 2026-02-09
Fix seeed_xiao_nrf52840_kit_i2c build

cwalther:
Building the _seeed_xiao_nrf52840_kit_i2c_ environment fails with the following linker error:

```
Linking .pio/build/seeed_xiao_nrf52840_kit_i2c/firmware-seeed_xiao_nrf52840_kit_i2c-2.7.18.baed99c.elf
/Users/cwalther/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld: .pio/build/seeed_xiao_nrf52840_kit_i2c/firmware-seeed_xiao_nrf52840_kit_i2c-2.7.18.baed99c.elf section `.text' will not fit in region `FLASH'
/Users/cwalther/.platformio/packages/toolchain-gccarmnoneeabi/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld: region `FLASH' overflowed by 123144 bytes
collect2: error: ld returned 1 exit status
```

This is fixed by the attached commit. The reason, discovered by comparing the verbose build output between `seeed_xiao_nrf52840_kit` (which builds successfully) and `seeed_xiao_nrf52840_kit_i2c`, was that `[env:seeed_xiao_nrf52840_kit_i2c]` mistakenly overwrites instead of extends `build_unflags` from the base environment and thereby loses `build_unflags: -Ofast, -Og, -ggdb3, -ggdb2, -g3, -g2, -g, -g1, -g0` which affect the code size.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Seeed Xiao NRF52840 Kit

### DIFF
--- a/.clusterfuzzlite/Dockerfile
+++ b/.clusterfuzzlite/Dockerfile
@@ -20,7 +20,7 @@ ENV PIP_ROOT_USER_ACTION=ignore
 RUN apt-get update && apt-get install --no-install-recommends -y \
         cmake git zip libgpiod-dev libbluetooth-dev libi2c-dev \
         libunistring-dev libmicrohttpd-dev libgnutls28-dev libgcrypt20-dev \
-        libusb-1.0-0-dev libssl-dev pkg-config libsqlite3-dev && \
+        libusb-1.0-0-dev libssl-dev pkg-config libsqlite3-dev libsdl2-dev && \
     apt-get clean && rm -rf /var/lib/apt/lists/* && \
     pip install --no-cache-dir -U \
         platformio==6.1.16 \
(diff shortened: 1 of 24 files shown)


cwalther:
Doh, I mistakenly rebased on an old version of `develop`. Here it is again on the current one.

cwalther:
I have discovered in the meantime that this is redundant with #8866 (which solves the problem in a different way), so if that one is planned to land soon, don’t bother reviewing this.

caveman99:
closing in favor of #8866 as suggested - thanks for your work nonetheless.

PR #9594, opened 2026-02-10
Round telemetry precision

fresh3nough:
## Summary
- round telemetry float values to fixed precision before send/serialize
- add shared telemetry precision helper
- update telemetry serializer tests for rounded values

## Testing
- pio test -e native -f test_meshpacket_serializer (fails: pio not installed)

Co-Authored-By: Warp <agent@warp.dev>


caveman99:
Sorry, this is pointless. Over the air, floating point values are always the same size, as noted in the discussion.

PR #9601, opened 2026-02-11
security: replace sprintf/strcpy with bounds-checked variants

weebl2000:
## Summary
- Replaces unsafe `sprintf` calls with `snprintf` across multiple modules
- Replaces unsafe `strcpy` calls with `strncpy` + null termination
- Adds `assert()` for statically-bounded writes and `LOG_ERROR`/`LOG_WARN` + `assert` for runtime-dependent writes, so overflows are caught in debug and reported in release
- Refactors `DetectionSensorModule` to write directly into `payload.bytes`, eliminating a heap allocation, fixing a packet leak on the public-channel error path, and removing a bogus null-terminator write after the bell character

### Changes by file
| File | What changed |
|---|---|
| `DetectionSensorModule.cpp` | Rewrote `sendDetectionMessage` and `sendCurrentStateMessage` — `snprintf` directly into `payload.bytes`, use return value instead of `strlen`, `packetPool.release()` on error paths |
| `DropzoneModule.cpp` | `snprintf` + assert-only for bounded `matchCompare` (short/long name), `LOG_ERROR` + assert for `replyStr` formatting |
| `KeyVerificationModule.cpp` | `snprintf` + assert for all 5 notification messages, cleaned up `strncpy` of `remote_longname` with cached `strlen` and explicit null-termination, fixed `IF_SCREEN` block to use `sizeof(message)` instead of hardcoded `25` |
| `NodeDB.cpp` | `snprintf` with `LOG_ERROR` + assert + hardcoded fallback message for the duplicate-key notification |
| `RedirectablePrint.cpp` | `strncpy` + `LOG_WARN` for BLE log message truncation (non-critical), assert-only for thread name (bounded) |
| `FSCommon.cpp` | `strncpy` + `LOG_WARN` + assert for file path copies, deduplicated the ESP32/other `#ifdef` branches |
| `NodeInfoModule.cpp` | assert + `strncpy` for node ID copy (`"!%08x"` = 10 chars into 16-byte buffer) |

### Background
MeshCore's security audit flagged widespread `sprintf`/`strcpy` usage without bounds checking. The same patterns exist in Meshtastic. While most buffers are sized adequately for typical data, user-configurable fields (sensor names, node names) and mesh-received data (owner names) could exceed buffer limits. Rather than silently truncating, overflows now trigger asserts in debug builds and log errors in release builds.

## Test plan
- [ ] Verify `tbeam` build succeeds
- [ ] Verify detection sensor messages still send correctly
- [ ] Verify dropzone conditions responses format correctly
- [ ] Verify key verification UI notifications display correctly
- [ ] Verify BLE log records don't truncate normal messages

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -25,6 +25,7 @@
 #include "modules/NeighborInfoModule.h"
 #include <ErriezCRC32.h>
 #include <algorithm>
+#include <assert.h>
 #include <pb_decode.h>
 #include <pb_encode.h>
 #include <power/PowerHAL.h>
@@ -1837,7 +1838,13 @@ bool NodeDB::updateUser(uint32_t nodeId, meshtastic_User &p, uint8_t channelInde
                 meshtastic_ClientNotification *cn = clientNotificationPool.allocZeroed();
                 cn->level = meshtastic_LogRecord_Level_WARNING;
                 cn->time = getValidTime(RTCQualityFromNet);
-                snprintf(cn->message, sizeof(cn->message), warning, p.long_name);
+                int written = snprintf(cn->message, sizeof(cn->message), warning, p.long_name);
+                if (written < 0 || (size_t)written >= sizeof(cn->message)) {
+                    LOG_ERROR("Duplicate key warning truncated (wrote %d, buf %u)", written, sizeof(cn->message));
+                    assert(false);
+                    snprintf(cn->message, sizeof(cn->message),
+                             "Remote device is advertising your public key. Possible key compromise.");
+                }
                 service->sendClientNotification(cn);
             }
             return false;
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -1837,7 +1837,7 @@ bool NodeDB::updateUser(uint32_t nodeId, meshtastic_User &p, uint8_t channelInde
                 meshtastic_ClientNotification *cn = clientNotificationPool.allocZeroed();
                 cn->level = meshtastic_LogRecord_Level_WARNING;
                 cn->time = getValidTime(RTCQualityFromNet);
-                sprintf(cn->message, warning, p.long_name);
+                snprintf(cn->message, sizeof(cn->message), warning, p.long_name);
                 service->sendClientNotification(cn);
             }
             return false;
--- a/src/modules/DetectionSensorModule.cpp
+++ b/src/modules/DetectionSensorModule.cpp
@@ -6,6 +6,7 @@
 #include "configuration.h"
 #include "main.h"
 #include <Throttle.h>
+#include <assert.h>
 DetectionSensorModule *detectionSensorModule;
 
 #define GPIO_POLLING_INTERVAL 100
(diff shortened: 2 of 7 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -1830,7 +1830,7 @@ bool NodeDB::updateUser(uint32_t nodeId, meshtastic_User &p, uint8_t channelInde
                 meshtastic_ClientNotification *cn = clientNotificationPool.allocZeroed();
                 cn->level = meshtastic_LogRecord_Level_WARNING;
                 cn->time = getValidTime(RTCQualityFromNet);
-                sprintf(cn->message, warning, p.long_name);
+                snprintf(cn->message, sizeof(cn->message), warning, p.long_name);
```
Jorropo: This should probably error rather than truncating.

src/modules/DetectionSensorModule.cpp
```diff
@@ -120,7 +120,7 @@ void DetectionSensorModule::sendDetectionMessage()
 {
     LOG_DEBUG("Detected event observed. Send message");
     char *message = new char[40];
-    sprintf(message, "%s detected", moduleConfig.detection_sensor.name);
+    snprintf(message, 40, "%s detected", moduleConfig.detection_sensor.name);
```
Jorropo: The code you are changing is bad in other ways:
1. `message` shouldn't even exists in the first place (we should `snprintf`-in-place in `p->decoded.payload.bytes`)
1. `strlen(message)` shouldn't exists, `snprintf` returns the length of the string
1. this line of code `p->decoded.payload.bytes[p->decoded.payload.size + 1] = '\0'; // Bell character` shouldn't exists, it may do an oob write and this is not even a null terminated strings to begin with, it's a length terminated string.

src/modules/DetectionSensorModule.cpp
```diff
@@ -142,7 +142,7 @@ void DetectionSensorModule::sendDetectionMessage()
 void DetectionSensorModule::sendCurrentStateMessage(bool state)
 {
     char *message = new char[40];
-    sprintf(message, "%s state: %i", moduleConfig.detection_sensor.name, state);
+    snprintf(message, 40, "%s state: %i", moduleConfig.detection_sensor.name, state);
```
Jorropo: Exact same issues as https://github.com/meshtastic/firmware/pull/9601/changes#r2799283338

src/modules/DropzoneModule.cpp
```diff
@@ -32,13 +32,13 @@ ProcessMessage DropzoneModule::handleReceived(const meshtastic_MeshPacket &mp)
     auto &p = mp.decoded;
     char matchCompare[54];
     auto incomingMessage = reinterpret_cast<const char *>(p.payload.bytes);
-    sprintf(matchCompare, "%s conditions", owner.short_name);
+    snprintf(matchCompare, sizeof(matchCompare), "%s conditions", owner.short_name);
```
Jorropo: `short_name` is bounded, this can never trigger.
Using a bounded form is good in case the size of `short_name` but it should assert or error on failure.

src/modules/DropzoneModule.cpp
```diff
@@ -32,13 +32,13 @@ ProcessMessage DropzoneModule::handleReceived(const meshtastic_MeshPacket &mp)
...
     if (strncasecmp(incomingMessage, matchCompare, strlen(matchCompare)) == 0) {
         LOG_DEBUG("Received dropzone conditions request");
         startSendConditions = millis();
     }
 
-    sprintf(matchCompare, "%s conditions", owner.long_name);
+    snprintf(matchCompare, sizeof(matchCompare), "%s conditions", owner.long_name);
```
Jorropo: Same as https://github.com/meshtastic/firmware/pull/9601/changes#r2799299181

src/modules/DropzoneModule.cpp
```diff
@@ -77,11 +77,11 @@ meshtastic_MeshPacket *DropzoneModule::sendConditions()
         auto windDirection = telemetry.variant.environment_metrics.wind_direction;
         auto temp = telemetry.variant.environment_metrics.temperature;
         auto baro = UnitConversions::HectoPascalToInchesOfMercury(telemetry.variant.environment_metrics.barometric_pressure);
-        sprintf(replyStr, "%s @ %02d:%02d:%02dz\nWind %.2f kts @ %d°\nBaro %.2f inHg %.2f°C", dropzoneStatus, hour, min, sec,
-                windSpeed, windDirection, baro, temp);
+        snprintf(replyStr, sizeof(replyStr), "%s @ %02d:%02d:%02dz\nWind %.2f kts @ %d°\nBaro %.2f inHg %.2f°C", dropzoneStatus,
```
Jorropo: This should never overflow since it is easy to make `replyStr` longer than the longest possible string,
an assert on overflow would be good.

src/modules/KeyVerificationModule.cpp
```diff
@@ -69,7 +69,7 @@ bool KeyVerificationModule::handleReceivedProtobuf(const meshtastic_MeshPacket &
 
         meshtastic_ClientNotification *cn = clientNotificationPool.allocZeroed();
         cn->level = meshtastic_LogRecord_Level_WARNING;
-        sprintf(cn->message, "Enter Security Number for Key Verification");
+        snprintf(cn->message, sizeof(cn->message), "Enter Security Number for Key Verification");
```
Jorropo: Should fail on overflow instead.

src/modules/KeyVerificationModule.cpp
```diff
@@ -83,7 +83,7 @@ bool KeyVerificationModule::handleReceivedProtobuf(const meshtastic_MeshPacket &
     } else if (currentState == KEY_VERIFICATION_RECEIVER_AWAITING_HASH1 && r->hash1.size == 32 && r->nonce == currentNonce) {
         if (memcmp(hash1, r->hash1.bytes, 32) == 0) {
             memset(message, 0, sizeof(message));
-            sprintf(message, "Verification: \n");
+            snprintf(message, sizeof(message), "Verification: \n");
```
Jorropo: Should fail on overflow instead.

src/modules/KeyVerificationModule.cpp
```diff
@@ -101,7 +101,7 @@ bool KeyVerificationModule::handleReceivedProtobuf(const meshtastic_MeshPacket &
                       screen->showOverlayBanner(options);)
             meshtastic_ClientNotification *cn = clientNotificationPool.allocZeroed();
             cn->level = meshtastic_LogRecord_Level_WARNING;
-            sprintf(cn->message, "Final confirmation for incoming manual key verification %s", message);
+            snprintf(cn->message, sizeof(cn->message), "Final confirmation for incoming manual key verification %s", message);
```
Jorropo: Should error instead.

Jorropo (commented):
I am not gonna finish that review since all of them are identical.

This PR assumes truncating messages when we have bugs is acceptable.
I don't think so, all of theses unbounded operations exists because we think they are ok (the arguments have limited sizes and can't overflow).
If we are wrong in that assumption we should be told about it instead.

For something that has no runtime dependency (like `sprintf(x, "example")`) this should fail an assert.
For something that has a runtime dependency (easier to get wrong) crashing isn't ideal but you could turn them into error messages.
Like for example a push notification who can't fit a string should send you a push notification with an error message telling you we couldn't deliver a push notification to you, ...

weebl2000:
> If we are wrong in that assumption we should be told about it instead.

In this case, being told about it in way of user facing error or error logging for developers? (or both)

Jorropo:
> In this case, being told about it in way of user facing error or error logging for developers? (or both)

I would expect all of them to be developers issues.
It might make sense to show them to users, just so that they report bugs to us (like sending an error push notification if something manage to send a push notification that does not fit).

PR #9610, opened 2026-02-11
Some fixes for xeddsa pr

weebl2000:
- fix null dereference crash in NodeInfoModule when receiving NodeInfo from unknown nodes                                                                                                       
- signature verification failure didn't drop the packet — now it does
- HAS_XEDDSA_SIGNED bitfield was never set — now set on successful verification, so unsigned packets from previously-signing nodes are rejected                                             
- Signature only covered payload bytes — now covers metadata too: fromNode, packetId, and portnum to prevent replay, reattribution, and portnum redirection
- xeddsa_sign() had no key initialization check — now returns false if keys are all zeros                                                                                                   
- AdminModule didn't load keys when user provided both private and public key                                                                                                               
- curve_to_ed_pub() field inversion ran on every signed packet — now cached per sender
- createNewIdentity() ran unnecessarily on every boot when the key hadn't changed
- Hardcoded 120 byte signing limit replaced with derivation from XEDDSA_SIGNATURE_SIZE and DATA_PAYLOAD_LEN
- Added const qualifiers to input-only parameters in xeddsa_verify and curve_to_ed_pub

### DIFF
--- a/arch/portduino/portduino.ini
+++ b/arch/portduino/portduino.ini
@@ -24,7 +24,6 @@ lib_deps =
   ${radiolib_base.lib_deps}
   ${environmental_base.lib_deps}
   # renovate: datasource=custom.pio depName=rweather/Crypto packageName=rweather/library/Crypto
-  #rweather/Crypto@0.4.0
   https://github.com/meshtastic/Crypto/archive/1aa30eb536bd52a576fde6dfa393bf7349cf102d.zip
   # renovate: datasource=custom.pio depName=LovyanGFX packageName=lovyan03/library/LovyanGFX
   lovyan03/LovyanGFX@^1.2.0
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -2079,21 +2079,25 @@ bool NodeDB::generateCryptoKeyPair(const uint8_t *privateKey)
 
 bool NodeDB::createNewIdentity()
 {
-    // Remove the old node from the NodeDB
     uint32_t oldNodeNum = getNodeNum();
+    uint32_t newNodeNum = crc32Buffer(config.security.public_key.bytes, config.security.public_key.size);
+
+    // If the key hasn't changed, nothing to do
+    if (newNodeNum == oldNodeNum)
+        return false;
+
+    // Retire the old node entry
     meshtastic_NodeInfoLite *node = getMeshNode(oldNodeNum);
-
-    // Set my node num uint32 value to bytes from the new public key
-    myNodeInfo.my_node_num = crc32Buffer(config.security.public_key.bytes, config.security.public_key.size);
-
-    if (node != NULL && myNodeInfo.my_node_num != oldNodeNum) {
-        LOG_DEBUG("Old node num %u is now %u", oldNodeNum, myNodeInfo.my_node_num);
+    if (node != NULL) {
+        LOG_DEBUG("Old node num %u is now %u", oldNodeNum, newNodeNum);
         node->is_ignored = true;
         node->has_device_metrics = false;
         node->has_position = false;
         node->user.public_key.size = 0;
-        node->user.public_key.bytes[0] = 0;
+        memset(node->user.public_key.bytes, 0, sizeof(node->user.public_key.bytes));
     }
+
+    myNodeInfo.my_node_num = newNodeNum;
 
     meshtastic_NodeInfoLite *info = getOrCreateMeshNode(getNodeNum());
     info->user = TypeConversions::ConvertToUserLite(owner);
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -501,22 +501,28 @@ DecodeState perhapsDecode(meshtastic_MeshPacket *p)
             p->decoded.want_response |= p->decoded.bitfield & BITFIELD_WANT_RESPONSE_MASK;
 
         if (p->decoded.has_xeddsa_signature) {
-            LOG_WARN("packet shows XEdDSA");
             meshtastic_NodeInfoLite *node = nodeDB->getMeshNode(p->from);
             if (node && node->user.public_key.size == 32) {
-                LOG_WARN("attempting to verify");
                 p->xeddsa_signed = crypto->xeddsa_verify(node->user.public_key.bytes, p->decoded.payload.bytes,
                                                          p->decoded.payload.size, p->decoded.xeddsa_signature.bytes);
+                if (p->xeddsa_signed) {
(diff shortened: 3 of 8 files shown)


arch/portduino/portduino.ini
```diff
@@ -24,7 +24,6 @@ lib_deps =
   ${radiolib_base.lib_deps}
   ${environmental_base.lib_deps}
   # renovate: datasource=custom.pio depName=rweather/Crypto packageName=rweather/library/Crypto
-  #rweather/Crypto@0.4.0
   https://github.com/meshtastic/Crypto/archive/1aa30eb536bd52a576fde6dfa393bf7349cf102d.zip
```
caveman99, line 27: The renovate definition does not fit the link to the commit ref.

Please do not check in changes unrelated to this PR

src/mesh/NodeDB.cpp
```diff
@@ -2079,22 +2079,26 @@ bool NodeDB::generateCryptoKeyPair(const uint8_t *privateKey)
 
 bool NodeDB::createNewIdentity()
 {
-    // Remove the old node from the NodeDB
     uint32_t oldNodeNum = getNodeNum();
```
caveman99, line 2082: This change is not in scope of the PR. Please open a separate PR for this.

src/mesh/Router.cpp
```diff
@@ -568,11 +575,16 @@ meshtastic_Routing_Error perhapsEncode(meshtastic_MeshPacket *p)
...
             p->decoded.bitfield |= (config.lora.config_ok_to_mqtt << BITFIELD_OK_TO_MQTT_SHIFT);
             p->decoded.bitfield |= (p->decoded.want_response << BITFIELD_WANT_RESPONSE_SHIFT);
-            if (p->pki_encrypted == false && isBroadcast(p->to) && p->decoded.payload.size < 120) {
-                crypto->xeddsa_sign(p->decoded.payload.bytes, p->decoded.payload.size, p->decoded.xeddsa_signature.bytes);
-                p->decoded.xeddsa_signature.size = 64;
-                p->decoded.has_xeddsa_signature = true;
-                LOG_WARN("XEDDSA Signed!");
+            // Sign broadcast packets if payload + signature fits within the max Data payload.
+            // The actual encoded size is checked after pb_encode (TOO_LARGE).
+            if (!p->pki_encrypted && isBroadcast(p->to) &&
```
caveman99, line 580: What happens if it is too large? Reject? Send unsigned?

src/modules/NodeInfoModule.cpp
```diff
@@ -24,8 +24,10 @@ bool NodeInfoModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mes
     }
     NodeNum sourceNum = getFrom(&mp);
     auto node = nodeDB->getMeshNode(sourceNum);
-    if ((node->bitfield & NODEINFO_BITFIELD_HAS_XEDDSA_SIGNED_MASK) && !mp.xeddsa_signed)
+    if (node && (node->bitfield & NODEINFO_BITFIELD_HAS_XEDDSA_SIGNED_MASK) && !mp.xeddsa_signed) {
+        LOG_WARN("Dropping unsigned NodeInfo from node 0x%08x that previously signed", sourceNum);
         return true;
```
caveman99, line 29: Don't add excessive logging

caveman99 (changes requested):

arch/portduino/portduino.ini
weebl2000: It's related. The dependency was removed in favor of the fork.

src/mesh/Router.cpp
weebl2000: If the message becomes too large it is sent unsigned (it was already like this, just added named constant). Perhaps we can do a debug log in an else to indicate unsigned messages are being sent.

src/modules/NodeInfoModule.cpp
weebl2000: LOG_DEBUG maybe? But I think it's a security concern if this happens so at least some logging is warranted.

src/mesh/NodeDB.cpp
weebl2000: But the entire function is new.

https://github.com/meshtastic/firmware/compare/develop...XEdDSA#diff-d3ca504aa401bdfaa42fcd99349d867b504756a86721afbbb46482909e8b1001R2080

src/modules/NodeInfoModule.cpp
caveman99: The string is the problem. We have too much logging already.

arch/portduino/portduino.ini
caveman99: It's a cosmetic change and again not in the scope of this PR, now if you'd fix the renovate string to point to the correct repository that would be a different story.

src/mesh/NodeDB.cpp
caveman99: You are right, sorry, i made that remark in the haste of a quick review.

src/mesh/Router.cpp
caveman99: I don't think this is a good idea, since the average user that sends messages from the app doesn't know if it is signed or not. Either refuse with message too large or notify the API back that it is sent unsigned

weebl2000: I agree, it's a bit dangerous like this.

arch/portduino/portduino.ini
weebl2000: > It's a cosmetic change and again not in the scope of this PR, now if you'd fix the renovate string to point to the correct repository that would be a different story.

I removed it because it was done in all other places except this place.

image

src/modules/NodeInfoModule.cpp
jp-bennett: For the initial implementation, extra logging is warranted. We can cut the strings out once we know things are working

src/modules/AdminModule.cpp
```diff
@@ -817,6 +817,12 @@ void AdminModule::handleSetConfig(const meshtastic_Config &c)
         else if (config.security.private_key.size == 32 && config.security.public_key.size == 0) {
             nodeDB->generateCryptoKeyPair(config.security.private_key.bytes);
         }
+        // User provided both keys — load them into the crypto engine
```
jp-bennett: We intentionally don't do this. Too much of a footgun for users to use mismatching keys.

src/mesh/Router.cpp
jp-bennett: We'll address this on the main PR.

jp-bennett (approved):

PR #9623, opened 2026-02-12
fix a lot of low level cppcheck warnings

caveman99:
let's see where this barfs ;-)

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -366,8 +366,7 @@ std::unique_ptr<RadioInterface> initLoRa()
 
     if ((!rIf) && (config.lora.region != meshtastic_Config_LoRaConfig_RegionCode_LORA_24)) {
         // If specified TCXO voltage fails, attempt to use DIO3 as a reference instead
-        rIf =
-            std::unique_ptr<SX1262Interface>(new SX1262Interface(loraHal, SX126X_CS, SX126X_DIO1, SX126X_RESET, SX126X_BUSY));
+        rIf = std::unique_ptr<SX1262Interface>(new SX1262Interface(loraHal, SX126X_CS, SX126X_DIO1, SX126X_RESET, SX126X_BUSY));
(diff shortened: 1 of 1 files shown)


src/Power.cpp
```diff
@@ -705,9 +705,9 @@ bool Power::setup()
     } else if (analogInit()) {
         found = true;
     }
-
 #ifdef NRF_APM
-    found = true;
+    if (!found)
```
thebentern: Was this throwing a warning? Seems odd

src/input/TCA8418Keyboard.cpp
```diff
@@ -72,7 +71,7 @@ void TCA8418Keyboard::pressed(uint8_t key)
     }
 
     // Compute key index based on dynamic row/column
-    next_key = row * _TCA8418_COLS + col;
+    next_key = (int8_t)(row * _TCA8418_COLS + col);
```
thebentern, line 74: Shouldn't this be a uint8_t?

src/Power.cpp
caveman99: that's the last one standing out - it's cppcheck being weird.

src/input/TCA8418Keyboard.cpp
caveman99: that was also one of my question marks. i'll review once more

src/Power.cpp
caveman99: The original warning was 'found is changes several times without computation in between. now it complains that the if part is superfluous.

src/input/TCA8418Keyboard.cpp
caveman99: Ah. next_key and last_key are class variables defined as int_8 - the original issue was variable shadowing inside the pressed() function by redefining as uint8 and then signed/unsigned comaprison to int8 last_key. 

Apparently last_key can be <0 (checked in release()) - are they prone to go over 128? If so we shouold increase to INt16

PR #9625, opened 2026-02-12
guard against overflow or null-dereference

robekl:
Changes based on tool analysis of missing or inadequate bounds checks or null checks. Unchecked; I don't have hardware to verify.

### DIFF
--- a/src/FSCommon.cpp
+++ b/src/FSCommon.cpp
@@ -136,9 +136,9 @@ std::vector<meshtastic_FileInfo> getFiles(const char *dirname, uint8_t levels)
         } else {
             meshtastic_FileInfo fileInfo = {"", static_cast<uint32_t>(file.size())};
 #ifdef ARCH_ESP32
-            strcpy(fileInfo.file_name, file.path());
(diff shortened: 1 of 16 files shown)


caveman99:
Similar to https://github.com/meshtastic/firmware/pull/9601 - please read and understand the comments by @Jorropo in this PR, then come back and explain to us why we would waste our time on this one

robekl:
> Similar to #9601 - please read and understand the comments by @Jorropo in this PR, then come back and explain to us why we would waste our time on this one

If multiple tools are independently identifying code that could potentially cause a crash or memory corruption, maybe it's worth reviewing or refactoring. That's the tax of c-style code and being responsible for your own memory management and direct access. If you have so little time to "waste" to review/accept fixes that other people are offering, you must be pretty busy.

caveman99:
Did you read the comments in the other PR? Almost every occurance there identified by a tool was either invalid or should be handled differently. I acknowledge this PR changes more than the one i referenced, but if there's something to be concerned about, pinpoint the actual issues one by one and don't dump a machinated big changeset here for us to review.

This is by no means meant in a hostile way, it's just our way to deal with the limited human ressources we have here.

PR #9626, opened 2026-02-12
fix: respect DontMqttMeBro flag regardless of channel PSK

weebl2000:
~~The memcmp calls for defaultpsk/eventpsk were used in boolean context without comparing to 0, causing the condition to be true when keys did NOT match rather than when they did. This meant the DontMqttMeBro flag was enforced for custom PSKs but ignored for default/event PSKs — the opposite of the intended behavior described in PR #4643.~~

Removes the PSK-matching check from the MQTT uplink filter entirely. The `dontUplink` flag is now respected for all non-local packets on public MQTT servers, regardless of channel key.

The PSK check was broken from its introduction in #4643 — `memcmp` was used in boolean context without comparing to 0, inverting the condition. Since no one noticed for over a year, the PSK-based filtering provided no practical value. Simplifying to always respect the sender's preference is both more correct and easier to reason about.

Discovered in #9600

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -754,10 +754,8 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp_encrypted, const meshtastic_Me
     if (mp_decoded.which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
         // For uplinking other's packets, check if it's not OK to MQTT or if it's an older packet without the bitfield
         bool dontUplink = !mp_decoded.decoded.has_bitfield || !(mp_decoded.decoded.bitfield & BITFIELD_OK_TO_MQTT_MASK);
-        // check for the lowest bit of the data bitfield set false, and the use of one of the default keys.
-        if (!isFromUs(&mp_decoded) && !isMqttServerAddressPrivate && dontUplink &&
-            (ch.settings.psk.size < 2 || (ch.settings.psk.size == 16 && memcmp(ch.settings.psk.bytes, defaultpsk, 16) == 0) ||
-             (ch.settings.psk.size == 32 && memcmp(ch.settings.psk.bytes, eventpsk, 32) == 0))) {
+        // Respect the DontMqttMeBro flag for other nodes' packets on public MQTT servers
+        if (!isFromUs(&mp_decoded) && !isMqttServerAddressPrivate && dontUplink) {
             LOG_INFO("MQTT onSend - Not forwarding packet due to DontMqttMeBro flag");
             return;
         }
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -756,8 +756,8 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp_encrypted, const meshtastic_Me
         bool dontUplink = !mp_decoded.decoded.has_bitfield || !(mp_decoded.decoded.bitfield & BITFIELD_OK_TO_MQTT_MASK);
         // check for the lowest bit of the data bitfield set false, and the use of one of the default keys.
         if (!isFromUs(&mp_decoded) && !isMqttServerAddressPrivate && dontUplink &&
-            (ch.settings.psk.size < 2 || (ch.settings.psk.size == 16 && memcmp(ch.settings.psk.bytes, defaultpsk, 16)) ||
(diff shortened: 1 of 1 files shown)


Jorropo (approved):
Thanks

Jorropo (commented):
Ok actually this code is too hard to read I have no idea what the intent is.
Don't merge until we make a decision.

src/mqtt/MQTT.cpp
```diff
@@ -756,8 +756,8 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp_encrypted, const meshtastic_Me
         bool dontUplink = !mp_decoded.decoded.has_bitfield || !(mp_decoded.decoded.bitfield & BITFIELD_OK_TO_MQTT_MASK);
         // check for the lowest bit of the data bitfield set false, and the use of one of the default keys.
         if (!isFromUs(&mp_decoded) && !isMqttServerAddressPrivate && dontUplink &&
-            (ch.settings.psk.size < 2 || (ch.settings.psk.size == 16 && memcmp(ch.settings.psk.bytes, defaultpsk, 16)) ||
-             (ch.settings.psk.size == 32 && memcmp(ch.settings.psk.bytes, eventpsk, 32)))) {
+            (ch.settings.psk.size < 2 || (ch.settings.psk.size == 16 && memcmp(ch.settings.psk.bytes, defaultpsk, 16) == 0) ||
+             (ch.settings.psk.size == 32 && memcmp(ch.settings.psk.bytes, eventpsk, 32) == 0))) {
```
Jorropo:
```suggestion
       if (!isFromUs(&mp_decoded) && !isMqttServerAddressPrivate && dontUplink) {
```
This is my guess of what should happen:
if a user asks to not be uplinked to the public MQTT server; respect that.

For some reason the previous code would respect that only on:
1. one byte AES keys
2. 16 & 32 bytes AES keys who aren't `defaultpsk` & `eventpsk`

Which is wrong since `defaultpsk` is equal to the `{1}` one byte AES key.
And it is nonsensical to take a different decision for `defaultpsk` based on how you configured it.

I don't see a reason to not respect `dontUplink` irrespective of the channel, it's practically  the existing behavior anyhow since no one would configure `defaultpsk` & `eventpsk` without using the 1 byte key solution.

Jorropo (changes requested):

src/mqtt/MQTT.cpp
weebl2000: Good call — removed the PSK check entirely. The dontUplink flag is now respected for all non-local packets on public MQTT servers, regardless of channel key. The PSK filtering was broken since #4643 and nobody noticed, so I guess it's not needed.

src/mqtt/MQTT.cpp
```diff
@@ -754,10 +754,8 @@ void MQTT::onSend(const meshtastic_MeshPacket &mp_encrypted, const meshtastic_Me
     if (mp_decoded.which_payload_variant == meshtastic_MeshPacket_decoded_tag) {
         // For uplinking other's packets, check if it's not OK to MQTT or if it's an older packet without the bitfield
         bool dontUplink = !mp_decoded.decoded.has_bitfield || !(mp_decoded.decoded.bitfield & BITFIELD_OK_TO_MQTT_MASK);
-        // check for the lowest bit of the data bitfield set false, and the use of one of the default keys.
-        if (!isFromUs(&mp_decoded) && !isMqttServerAddressPrivate && dontUplink &&
-            (ch.settings.psk.size < 2 || (ch.settings.psk.size == 16 && memcmp(ch.settings.psk.bytes, defaultpsk, 16)) ||
-             (ch.settings.psk.size == 32 && memcmp(ch.settings.psk.bytes, eventpsk, 32)))) {
```
fifieldt, line 760: I think probably not, but are there any  meshtastic-supported events that don't use the public server?

weebl2000: I don't think it matters in practice. The eventpsk exemption was broken since it was introduced (inverted memcmp), so DontMqttMeBro has effectively always been respected on event channels. This change just makes the code match the actual behavior.

caveman99: There was a private server at DEFCON25 but it was only used for data collection. That would have been an Event where uplink was desired since it was not passed on just processed for statistics.

dshot92:
Can this fix be the reason of this signal propagation increase almost overnight?

https://github.com/meshtastic/firmware/discussions/10965

PR #9627, opened 2026-02-12
Add phone API fanout for multiple clients

kotovasia5120:
Replace single to-phone queues with per-client fanout queues and pools so multiple PhoneAPI clients (BLE/WiFi/Serial/HTTP/Ethernet/etc.) can be served concurrently. Introduces PhoneClientSlot, per-client StaticPointerQueues, fanout entry pools, and a phoneClientsLock to protect registration and queue operations. Adds register/unregister APIs, enqueue/get/release helpers for packets, queue statuses, MQTT proxy messages and notifications, and implements refcounted release semantics and drop-oldest behavior for full client queues. Tracks active APIs with api_state_mask and per-state counts, updates api_state selection logic, and updates UI/debug rendering to show combined active APIs. Also: add HTTP activity tracking to HttpAPI/ContentHandler and unit tests (test_api_fanout) to validate fanout behavior and API state mask. Misc: small headers/includes and utility functions added.

## 🤝 Attestations

  - [x] DIY esp32s3 e22
  - [x] seeed xiao s3

### DIFF
--- a/src/graphics/SharedUIDisplay.cpp
+++ b/src/graphics/SharedUIDisplay.cpp
@@ -427,7 +427,7 @@ const int *getTextPositions(OLEDDisplay *display)
 // *************************
 void drawCommonFooter(OLEDDisplay *display, int16_t x, int16_t y)
 {
-    if (!isAPIConnected(service->api_state))
+    if (!isAnyAPIConnected(service->api_state_mask))
         return;
 
     const int scale = (currentResolution == ScreenResolution::High) ? 2 : 1;
--- a/src/graphics/SharedUIDisplay.h
+++ b/src/graphics/SharedUIDisplay.h
@@ -77,4 +77,9 @@ static inline bool isAPIConnected(uint8_t state)
     return state < sizeof(connectedStates) ? connectedStates[state] : false;
 }
 
+static inline bool isAnyAPIConnected(uint32_t mask)
+{
+    return mask != 0;
+}
+
 } // namespace graphics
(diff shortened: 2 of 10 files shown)


thebentern:
What's the goal here?

kotovasia5120:
> What's the goal here?

Basically, I lost access to the settings via the phone if there was a serial transmission.

GUVWAF:
This is potentially very interesting, as e.g. it would also enable running the MUI in combination with a Bluetooth connection, but I think it needs extensive testing with all clients. Furthermore, I would say supporting 8 simultaneous connections is way over the top. Things like: `MemoryPool<PacketFanoutEntry, MAX_PHONE_API_CLIENTS * MAX_RX_TOPHONE>`, which are statically allocated, will increase RAM usage very significantly. I would say starting with max. 2 API clients would already be a huge improvement (we can't run BLE and Wi-Fi reliably at the same time anyway).

caveman99:
I am trying to phrase this as politely as possible. A first time contribution of this size that shuffles part of the inner workings of the firmware and has a few telltale signs of generative AI needs heavy scrutiny. "Basically, I lost access to the settings via the phone if there was a serial transmission." is not a compelling answer either. The problem of singular API connection has been there for years, and frankly I lack the confidence in a drive-by PR to fix that magically. If you are really sincere about adding this, AND commit to maintaining it, please state a few details about WHY you implemented a few things like they are in the PR. Please your case, explain the ideas and the architecture, the advantages and the pitfalls. And i suggest doing it yourself, not having an LLM express that for you.

PR #9628, opened 2026-02-12
convert `GPS*` global and some new in gps.cpp to `unique_ptr<GPS>`

Jorropo:
Trying this to see if anything bad happen if I were to replace most raw pointers with unique_ptr.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -52,7 +52,7 @@ SerialUART *GPS::_serial_gps = &GPS_SERIAL_PORT;
 HardwareSerial *GPS::_serial_gps = nullptr;
 #endif
 
-GPS *gps = nullptr;
+std::unique_ptr<GPS> gps = nullptr;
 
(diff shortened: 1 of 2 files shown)


fifieldt:
This may be a useless idea , but: one thing you could try is reducing the serial buffer size, getting rid of the NMEA disable lines and see what happens. We used to run up against limits for that when it was 1024 for Nrf52. It may still happen on occasion for some chips on some architectures, so would be interesting to see of after these changes the platform crashes or something...

Jorropo:
@fifieldt I am not sure what you are talking about, CI is failing since even tho I've used `build_flags = -std=c++14` the xtensa esp32 builds use ¿ `c++11` ?

@vidplace7 is telling me it is an esp-idf issue apparently.

PR #9633, opened 2026-02-13
Add INKHUD build support for T-Echo Plus

kuroanji:
Tested, working like charm.

- [✅ I have tested that my proposed changes behave as described.
- [✅ I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  -  ✅ LilyGo T-Echo Plus


caveman99:
-1 AI agent

PR #9638, opened 2026-02-14
envrc: use flake

HannesGitH:
nix flake is the successor of legacy nix

similarly `use flake` can be seen as a successor to `use nix`

the main benefit is that everyone (using direnv) get the exact same versions of platformIO and python etc

with the old approach the system nix is used, which in my case was broken and did not work

when using flakes, if it works on one machine it basically works everywhere, reliably

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,1 @@
-use nix
+use flake


HannesGitH:
@agustinmista

agustinmista:
Hey @HannesGitH! :hand: 

Please bear in mind that the current implementation _does_ use flakes under the hood (via [flake-compat](https://github.com/NixOS/flake-compat)), while retaining backward compatibility with non-flake setups.

This change would break the entry point of anyone who (for any reason) prefers _not_ to use flakes :cry:

HannesGitH:
edit: okay seemingly flake-compat should also use the locked versions, I wonder why it didnt work on my end then..

---
but the packages are not pinned to the lockfile with the old approach, right?
at least I could not build using the `use nix` hook as my current system nix had some incompatible python packages
doing `use flake` solved the issue as then I used the exact same versions you did when locking the flake

so imo the benefit of actual reproducibility outweighs the legacy compatibility, especially as legacy user can still `nix-develop` and only loose the direnv hook

but I might be mistaken

PR #9639, opened 2026-02-14
nix flake: fix shell hook to enable building esp32

HannesGitH:
Using the flake did not successfully build a binary as python dependencies could not correctly be installed, this fixes that

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  
custom esp32-s3 with sx1276

### DIFF
--- a/src/graphics/niche/InkHUD/Applet.cpp
+++ b/src/graphics/niche/InkHUD/Applet.cpp
@@ -144,6 +144,21 @@ void InkHUD::Applet::resetDrawingSpace()
     setFont(fontSmall);
 }
 
+// Sets one or more inputs to enabled/disabled for this applet and if they should be sent to it
+void InkHUD::Applet::setInputsSubscribed(uint8_t input, bool captured)
+{
+    if (captured)
+        subscribedInputs |= input;
+    else
+        subscribedInputs &= ~input;
+}
+
+// Checks if a specific input is enabled for this applet and should be sent to it
+bool InkHUD::Applet::isInputSubscribed(InputMask input)
+{
+    return (subscribedInputs & input) == input;
+}
+
 // Tell InkHUD::Renderer that we want to render now
 // Applets should internally listen for events they are interested in, via MeshModule, CallbackObserver etc
 // When an applet decides it has heard something important, and wants to redraw, it calls this method
--- a/src/graphics/niche/InkHUD/Applet.h
+++ b/src/graphics/niche/InkHUD/Applet.h
@@ -89,6 +89,9 @@ class Applet : public GFX
     virtual void onForeground() {}
     virtual void onBackground() {}
     virtual void onShutdown() {}
+
+    // Input Events
+
     virtual void onButtonShortPress() {}
     virtual void onButtonLongPress() {}
     virtual void onExitShort() {}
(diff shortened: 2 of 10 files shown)


agustinmista:
FYI: these changes break the shell in NixOS (something that the initial PR introducing `flake.nix` was careful to avoid):

```console
$ pio run -e heltec-v3
...
[nanopb] Installing Protocol Buffers dependencies
error: externally-managed-environment

× This environment is externally managed
╰─> This command has been disabled as it tries to modify the immutable
    `/nix/store` filesystem.
    
    To use Python with Nix and nixpkgs, have a look at the online documentation:
    <https://nixos.org/manual/nixpkgs/stable/#python>.
```

HannesGitH:
edit: consider this message resolved
<details><summary>original message</summary>

> FYI: these changes break the shell in NixOS (something that the initial PR introducing `flake.nix` was careful to avoid):
> 
> ```
> $ pio run -e heltec-v3
> ...
> [nanopb] Installing Protocol Buffers dependencies
> error: externally-managed-environment
> 
> × This environment is externally managed
> ╰─> This command has been disabled as it tries to modify the immutable
>     `/nix/store` filesystem.
>     
>     To use Python with Nix and nixpkgs, have a look at the online documentation:
>     <https://nixos.org/manual/nixpkgs/stable/#python>.
> ```

It does output a binary in the end tho, which the original approach did not do for me :D (see comment in flake)

what exactly is this nanopb doing?

```
➜  meshtastic git:(fix-nix-flake) ✗ nix develop                                                                                                           
bash-5.3$ which pio
/nix/store/n8hwkxg5zhyz9dpk2w0ayfid6ydz13wl-platformio-6.1.18/bin/pio
bash-5.3$ pio run -e heltec-v3
...
[nanopb] Installing Protocol Buffers dependencies
error: externally-managed-environment

× This environment is externally managed
╰─> This command has been disabled as it tries to modify the immutable
    `/nix/store` filesystem.
    
    To use Python with Nix and nixpkgs, have a look at the online documentation:
    <https://nixos.org/manual/nixpkgs/stable/#python>.
...
heltec-v3      SUCCESS   00:03:27.863
```

</details>

HannesGitH:
I think having a deterministic build environment (including all dependencies) brings us much closer to reproducible binaries, think `.#binary.heltecV3`, via nix, which would allow us to use all of nix's features not only the dev-shell

for this reason, its per design of this pr, that the python directory is read-only and managed by nix

wdyt @agustinmista ? 
<sub>(if you disagree, I suppose we can keep both, the dynamic and the reproducible as dev-shells, and just fight over which one becomes the default ;))</sub>

---

ps: last commit improved on this and also completely removed the error, which was caused by missmatching python versions

HannesGitH:
```nix
packages = forAllSystems ({ pkgs, ... }:
  let
    pioPyDeps = ps: with ps; [ protobuf grpcio-tools intelhex ];
    platformio = pkgs.platformio;
    python3 = platformio.python.withPackages pioPyDeps;
  in pkgs.lib.genAttrs [ "heltec-v3"  /* ... (*) */ ] (target:
    pkgs.stdenv.mkDerivation {
      name = "meshtastic-firmware-${target}";
      src = ./.;
      buildInputs = [ platformio python3 ];
      buildPhase = ''
        pio run -e ${target}
      '';
      installPhase = ''
        cp .pio/build/${target}/firmware.bin $out/firmware.bin
      '';
    }));
          
          
# (*): `pio project config --json-output 2>/dev/null | jq -r '.[] | .[0] | select(startswith("env:")) | sub("^env:"; "")'`
```

something in the lines of this is what I had in mind as a next step, but currently dont have the time to find my way around platformIOs heavy reliance on non-reproducible `$HOME/.platformio`

this might also be something rather upstreamed into nix platformio

agustinmista:
> I think having a deterministic build environment (including all dependencies) brings us much closer to reproducible binaries, think `.#binary.heltecV3`, via nix, which would allow us to use all of nix's features not only the dev-shell
> 
> for this reason, its per design of this pr, that the python directory is read-only and managed by nix
> 
> wdyt @agustinmista ? (if you disagree, I suppose we can keep both, the dynamic and the reproducible as dev-shells, and just fight over which one becomes the default ;))
> 
> ps: last commit improved on this and also completely removed the error, which was caused by missmatching python versions

I'd personally love to see a fully Nixified setup, but I don't think PlatformIO would make it very easy for us without a non-trivial amount of work --- that's why my original commit only included a Nix shell with the bare minimum to bootstrap the rest on my system.

I also don't have write permissions to this repo, so I'm not far from being just a bystander :sweat_smile:

HannesGitH:
> I also don't have write permissions to this repo, so I'm not far from being just a bystander 😅

A knowledge one, that started nixification of this project, so I'd like to have your opinion on stuff :) 

> I don't think PlatformIO would make it very easy for us without a non-trivial amount of work

thats very likely true, but I think this PR would be a step in the right direction

HannesGitH:
I think I forced pushed develop once, and now github is in a broken state and does not accept any more pushes to this pr ? 🤨

I am sorry, opening a replacement

PR #9641, opened 2026-02-14
[Security FIX]: Hide Predefined Bluetooth PIN from Pairing Screen

kuroanji:
# Hide Predefined Bluetooth PIN from Pairing Screen

## Summary

When Bluetooth pairing mode is set to **Fixed PIN**, the device no longer displays the PIN code on screen during pairing. Instead, it shows a message indicating that a predefined PIN is in use.

## Motivation

When a user sets a fixed Bluetooth PIN, displaying it on screen during every pairing attempt is a security concern — anyone with physical access to the device can read the PIN. The fixed PIN is meant to be known only to the device owner, not broadcast on the display.

With a **random PIN**, showing the code on screen is necessary and expected — it's the only way the user can learn the generated code.

### Random PIN mode (`RANDOM_PIN`)

No change. The screen displays the generated PIN code as before:

```
     Bluetooth
   Enter this code
     123 456
   Name: device_name
```

### Fixed PIN mode (`FIXED_PIN`)

The PIN code is hidden. The screen shows:

```
     Bluetooth
     Pairing with
    predefined PIN
   Name: device_name
```

### No PIN mode (`NO_PIN`)

Not affected — pairing screen is not shown.

## Affected Platforms

| Platform | File | Display Type |
|---|---|---|
| NRF52 | `src/platform/nrf52/NRF52Bluetooth.cpp` | OLED (Screen class) |
| ESP32 (NimBLE) | `src/nimble/NimbleBluetooth.cpp` | OLED (Screen class) |
| InkHUD (e-ink) | `src/graphics/niche/InkHUD/Applets/System/Pairing/PairingApplet.cpp` | E-Ink |

## Implementation

All three display handlers check `config.bluetooth.mode` at render time. If the mode is `FIXED_PIN`, the PIN code is replaced with a text message. The condition is evaluated inside the render callback, so it reacts to runtime config changes without restart.

The Bluetooth pairing logic itself is unchanged — the fixed PIN is still used for authentication. Only the on-screen display is affected.

- [✅ ] I have tested that my proposed changes behave as described.
- [✅ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ✅
Predefined_PIN
] Lilygo T-Echo PLUS
[BT_Pairing_PIN_Display.md](https://github.com/user-attachments/files/25316787/BT_Pairing_PIN_Display.md)
[BT_Pairing_PIN_Changelog.md](https://github.com/user-attachments/files/25316788/BT_Pairing_PIN_Changelog.md)

### DIFF
--- a/src/graphics/niche/InkHUD/Applets/System/Pairing/PairingApplet.cpp
+++ b/src/graphics/niche/InkHUD/Applets/System/Pairing/PairingApplet.cpp
@@ -14,12 +14,20 @@ void InkHUD::PairingApplet::onRender(bool full)
     // Header
     setFont(fontMedium);
     printAt(X(0.5), Y(0.25), "Bluetooth", CENTER, BOTTOM);
-    setFont(fontSmall);
-    printAt(X(0.5), Y(0.25), "Enter this code", CENTER, TOP);
 
-    // Passkey
-    setFont(fontMedium);
-    printThick(X(0.5), Y(0.5), passkey.substr(0, 3) + " " + passkey.substr(3), 3, 2);
+    if (config.bluetooth.mode == meshtastic_Config_BluetoothConfig_PairingMode_FIXED_PIN) {
+        // Predefined PIN: don't reveal the code on screen
+        setFont(fontSmall);
+        printAt(X(0.5), Y(0.45), "Pairing with", CENTER, BOTTOM);
+        printAt(X(0.5), Y(0.55), "predefined PIN", CENTER, TOP);
+    } else {
+        setFont(fontSmall);
+        printAt(X(0.5), Y(0.25), "Enter this code", CENTER, TOP);
+
+        // Passkey
+        setFont(fontMedium);
+        printThick(X(0.5), Y(0.5), passkey.substr(0, 3) + " " + passkey.substr(3), 3, 2);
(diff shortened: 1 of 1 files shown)


tropho23:
If one pairs devices rarely and uses a fixed PIN that they might forget, showing it onscreen is critical.  If one forgets the PIN, either serial connection and viewing logs, or a factory reset is required to resolve this and many users will not want to do either.  This degrades usability for the sake of security, when we could maintain convenience and improve physical security.

I recommend adding a toggle to this feature so users can decide whether they want to hide PINs, and have it disabled by default so anyone who wants to enable it can do so.

Jorropo:
If you can see the screen of the device, you are arms away from plugging it in a USB port, accessing the uart console (which give you the same power bluetooth give you) or even worst bootloader mode and accessing any secrets it might have.
Thus calling this PR a security fix is very optimistic. In my threat model hackers have arms and USB cables.

I agree with @tropho23 it can be inconvenient.
So given the limited gains and rare but existent downside I would rather close this PR.

Xaositek:
I feel like there is minimal benefit with this change. I will more quickly forget the predefined PIN and end up having to some recovery effort.

PR #9647, opened 2026-02-14
Undefine LED_BUILTIN for Heltec v2 variant

ericbarch:
As a result of #8830, the Heltec v2 variant no longer compiles and throws the following error:

`<command-line>: error: expected unqualified-id before '-' token /Users/ebarch/.platformio/packages/framework-arduinoespressif32/variants/heltec_wifi_lora_32_V2/pins_arduino.h:10:22: note: in expansion of macro 'LED_BUILTIN'`

This undefines LED_BUILTIN for this variant which resolves the build error.

Probably not a super popular board, but still nice to be able to build for it.

### DIFF
--- a/variants/esp32/heltec_v2.1/platformio.ini
+++ b/variants/esp32/heltec_v2.1/platformio.ini
@@ -14,3 +14,4 @@ build_flags =
   ${esp32_base.build_flags}
   -D HELTEC_V2_1
   -I variants/esp32/heltec_v2.1
+  -ULED_BUILTIN


Jorropo:
I think that a bug in #8830 where it does:
```
-DLED_BUILTIN=-1
```
It should do:
```
-DLED_BUILTIN=(-1)
```
Since the `-` might leak into a binary operator rather than the unary one.
I don't really understand why adding or removing `-ULED_BUILTIN` does anything if not undoing the `-DLED_BUILTIN=-1` line.

ericbarch:
> I think that a bug in #8830 where it does:
> 
> ```
> -DLED_BUILTIN=-1
> ```
> 
> It should do:
> 
> ```
> -DLED_BUILTIN=(-1)
> ```
> 
> Since the `-` might leak into a binary operator rather than the unary one. I don't really understand why adding or removing `-ULED_BUILTIN` does anything if not undoing the `-DLED_BUILTIN=-1` line.

Hmm, it looks like doing `-DLED_BUILTIN=(-1)` breaks the build.

I was simply following the existing pattern if you look in other variant files where `-ULED_BUILTIN` is added to the build_flags. That's undefining the -1 value which is breaking the build.

Jorropo:
Sure but then I don't understand why we are defining `LED_BUILTIN` to begin with, if you want to force push your version again (sorry my client already gc-ed your commit as I can't `git checkout` it) would be nice thanks.

Jorropo:
I copy pasted the wrong number and based my changes on #8830 rather than your PR so overwrote your commit.

ericbarch:
> Sure but then I don't understand why we are defining `LED_BUILTIN` to begin with, if you want to force push your version again (sorry my client already gc-ed your commit as I can't `git checkout` it) would be nice thanks.

Yeah, I'm not sure why that previous change was necessary.

No worries, just force pushed.

Jorropo:
> No worries, just force pushed.

I've managed to find the missing commit by manually fetching from your repo, rebased it (that why I've rere-force-pushed).

Jorropo:
Also added Heltec v2.1 variant.

Jorropo:
Thanks, confirmed theses two targets now build by building them locally.

ericbarch:
> Thanks, confirmed theses two targets now build by building them locally.

Appreciate it!

3geETR:
I got the same error when building meshtastic-diy-v1_1 (esp32/diy/v1_1). A permanent solution is needed for this.

```
n of the previous definition
<command-line>: error: expected unqualified-id before '-' token
/home/user/.platformio/packages/framework-arduinoespressif32/variants/doitESP32devkitV1/pins_arduino.h:6:22: note: in expansion of macro 'LED_BUILTIN'
static const uint8_t LED_BUILTIN = 2;
^~~~~~~~~~~
<command-line>: error: expected unqualified-id before '-' token
/home/user/.platformio/packages/framework-arduinoespressif32/variants/doitESP32devkitV1/pins_arduino.h:6:22: note: in expansion of macro 'LED_BUILTIN'
static const uint8_t LED_BUILTIN = 2;
^~~~~~~~~~~
<command-line>: error: expected unqualified-id before '-' token
/home/user/.platformio/packages/framework-arduinoespressif32/variants/doitESP32devkitV1/pins_arduino.h:6:22: note: in expansion of macro 'LED_BUILTIN'
static const uint8_t LED_BUILTIN = 2;
^~~~~~~~~~~
<command-line>: error: expected unqualified-id before '-' token
/home/user/.platformio/packages/framework-arduinoespressif32/variants/doitESP32devkitV1/pins_arduino.h:6:22: note: in expansion of macro 'LED_BUILTIN'
static const uint8_t LED_BUILTIN = 2;
^~~~~~~~~~~
*** [.pio/build/meshtastic-diy-v1_1/src/FSCommon.cpp.o] Error 1
*** [.pio/build/meshtastic-diy-v1_1/src/DebugConfiguration.cpp.o] Error 1
*** [.pio/build/meshtastic-diy-v1_1/src/BluetoothCommon.cpp.o] Error 1
*** [.pio/build/meshtastic-diy-v1_1/src/DisplayFormatters.cpp.o] Error 1
```

PR #9650, opened 2026-02-14
replace delete in DebugConfiguration.cpp with std::unique_ptr

Jorropo:
Is part of the unique_ptr modernization effort.

### DIFF
--- a/src/DebugConfiguration.cpp
+++ b/src/DebugConfiguration.cpp
@@ -22,9 +22,12 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.*/
 
+#include "SerialConsole.h"
 #include "configuration.h"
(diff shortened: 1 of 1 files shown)


src/DebugConfiguration.cpp
```diff
@@ -120,26 +123,23 @@ bool Syslog::vlogf(uint16_t pri, const char *fmt, va_list args)
 
 bool Syslog::vlogf(uint16_t pri, const char *appName, const char *fmt, va_list args)
 {
-    char *message;
     size_t initialLen;
     size_t len;
     bool result;
 
     initialLen = strlen(fmt);
```
harry-iii-lord, line 130: This is probably out of scope of this PR but, what is reason to set `initialLen` to length of format? Assuming typically you want to "printf" something in log passing arguments this value will be almost always too short right?

Jorropo: I have no idea, I was considering removing the first allocation and passing a `nullptr` to `vsprintf` to compute the real length, then running the `vnsprintf` a second time.

I would expect all calls to be over `initialLen` anways since how often is a formated result smaller than it's verb ?
I guess numbers `<100` but outside of that ¿

PR #9654, opened 2026-02-15
InkHUD: Favorite Map Applet

HarukiToreda:
This PR introduces a new map applet that only shows nodes that have been favorited. this way the map won't zoom out with all other random nodes.


caveman99:
Is the 2 month commit history intentional? That usually happens when you create a PR from a long running existing branch instead of creating a new one off develop.

HarukiToreda:
> ? That usually happens when you create a PR from a long running existing branch instead of creating a new one off develop.

not intentional, I forgot this branch was old. I just re-used it

PR #8163, opened 2025-09-29
Add agc reset attempt

jp-bennett:
Still WIP.
An attempt to trigger an AGC and AFC reset while the radio is unused.

PR #9571, opened 2026-02-08
Apply SX1262 register 0x8B5 patch for improved GC1109 RX sensitivity

weebl2000:
## Summary

- Apply undocumented SX1262 register 0x8B5 patch (set LSB=1) on Heltec V4 and Wireless Tracker V2 boards with the GC1109 FEM
- Patch recommended by Heltec engineer (@Quency-D) based on Semtech guidance
- Guarded by `USE_GC1109_PA` so only affects GC1109 boards

## Background

Testing in MeshCore PR [#1398](https://github.com/meshcore-dev/MeshCore/pull/1398) showed this register patch significantly improved RX sensitivity on the Heltec V4, reducing packet loss from ~55% to ~25% — bringing it in line with a RAK4631 reference node at the same location.

The patch was originally identified via discussion in [meshtastic/firmware#9029](https://github.com/meshtastic/firmware/pull/9029) and MeshCore PR [#1249](https://github.com/meshcore-dev/MeshCore/pull/1249).

## Test plan

- [ ] Flash Heltec V4, verify RX sensitivity improvement via RSSI/SNR comparison with a reference node
- [ ] Verify TX still works correctly
- [ ] Verify no regression on other SX1262 boards (patch is behind `USE_GC1109_PA` guard)

PR #9658, opened 2026-02-15
RadioLib edge-triggered interrupts robustness

compumike:
While investigating https://github.com/meshtastic/firmware/pull/8163 (AGC reset attempt) I found a few possible race conditions in interrupt handling that might cause `RX_DONE` to get lost.

### Edge-Triggered Interrupts

`RadioLibInterface::enableInterrupt` calls `lora->setDio0Action` which configures a rising-edge-triggered interrupt: https://github.com/jgromes/RadioLib/blob/1d68c7635eeb44b61179df5c62b46e7ca5c13fb6/src/modules/SX126x/SX126x_config.cpp#L12-L14

Edge-triggered interrupts are particularly vulnerable to being missed because if the line rises between the time when `startReceive` tells the radio chip its IRQ configuration and when `enableInterrupt` runs, the interrupt line will be high, but the interrupt handler will never fire. (Instead, if it were configured as a level-triggered interrupt, then calling `enableInterrupt` would immediately run the interrupt handler when the line was high.)

It's a small window of vulnerable time, but it happens every packet.

The consequences of a missed interrupt are serious. (Waits for the watchdog agc reset.)

### Workarounds

This PR adds three fixes here:

1. Make `NotifiedWorkerThread` use `atomic` to prevent a tiny read-ISRwrite-clear race condition.
2. Adds `RadioLibInterface::checkRxDoneIrqFlag()` to poll the radio over SPI and notify if `RX_DONE` is set, as a backup to the interrupt handler. This gets called immediately after `enableInterrupt(ISR_RX)` to catch interrupts in the setup window.
3. Adds a 1 second `lastRadioMissedIrqPoll` to the main loop, near the `agc reset`. This is a backup to the interrupt handler, and it costs one SPI roundtrip every second at most.

### Does it catch anything?

See the `caught missed RX_DONE` lines below:

```
DEBUG | 12:39:30 1566 [DeviceTelemetry] Node status update: 19 online, 250 total
INFO  | 12:39:30 1566 [DeviceTelemetry] Send packet to phone
INFO  | 12:39:30 1566 Tell client we have new packets 70
DEBUG | 12:39:30 1566 [ServerAPI] FromRadio=STATE_SEND_PACKETS
DEBUG | 12:39:31 1566 [ServerAPI] phone downloaded packet (id=0x96e3855c fr=0xdeadbeef to=0xffffffff, transport = 0, WantAck=0, HopLim=7 Ch=0x0 Portnum=67 rxtime=1771177170 priority=10)
WARN  | 12:39:32 1568 caught missed RX_DONE
ERROR | 12:39:32 1568 [RadioIf] Ignore received packet due to error=-7 (maybe id=0x52898e0a fr=0x69852c30 to=0xfffdffff flags=0xa4 rxSNR=-10.5 rxRSSI=-111 nextHop=0x0 relay=0x4a)
DEBUG | 12:39:32 1568 [RadioIf] Packet RX (noise?) : 968ms
DEBUG | 12:39:47 1582 [ServerAPI] Got client heartbeat
DEBUG | 12:39:47 1582 [ServerAPI] FromRadio=STATE_SEND_QUEUE_STATUS, numbytes=6
ERROR | 12:39:47 1583 [RadioIf] Ignore received packet due to error=-7 (maybe id=0x134b7eee fr=0x43b73cc8 to=0xffffffff flags=0xc6 rxSNR=-20.5 rxRSSI=-121 nextHop=0xcc relay=0x92)
```

```
DEBUG | 12:57:55 2671 [Router] Ignore dupe incoming msg (id=0x85ddd227 fr=0x69852c30 to=0xffffffff, transport = 1, WantAck=0, HopLim=2 Ch=0x8 encrypted len=39 rxSNR=5.5 rxRSSI=-16 hopStart=
DEBUG | 12:57:55 2671 [Router] cancelSending id=0x85ddd227, removed=0
DEBUG | 12:57:55 2671 [Router] Incoming msg was filtered from 0x69852c30
DEBUG | 12:58:08 2684 [ServerAPI] Got client heartbeat
DEBUG | 12:58:08 2684 [ServerAPI] FromRadio=STATE_SEND_QUEUE_STATUS, numbytes=6
ERROR | 12:58:21 2697 [RadioIf] Ignore received packet due to error=-7 (maybe id=0x956a3e68 fr=0x71a9adf4 to=0xff55ffff flags=0xa2 rxSNR=-14 rxRSSI=-115 nextHop=0x0 relay=0x4a)
DEBUG | 12:58:22 2697 [RadioIf] Packet RX (noise?) : 559ms
WARN  | 12:58:23 2699 caught missed RX_DONE
DEBUG | 12:58:23 2699 [RadioIf] Corrected frequency offset: 940.656189
DEBUG | 12:58:23 2699 [RadioIf] Lora RX (id=0x956a3e68 fr=0x7ba9adfe to=0xffffffff, transport = 0, WantAck=0, HopLim=1 Ch=0x8 encrypted len=44 rxSNR=-18.75 rxRSSI=-119 hopStart=5 relay=0xb2)
DEBUG | 12:58:23 2699 [RadioIf] Packet RX: 559ms
INFO  | 12:58:23 2699 [Router] Packet History - insert: Using new slot @uptime 2699.646s TRACE NEW
DEBUG | 12:58:23 2699 [Router] Use channel 0 (hash 0x8)
```

```
ERROR | 13:03:19 2995 [RadioIf] Ignore received packet due to error=-7 (maybe id=0x123c1e39 fr=0xe9852c30 to=0xffffffff flags=0xab rxSNR=-19.5 rxRSSI=-120 nextHop=0x0 relay=0xb2)
DEBUG | 13:03:19 2995 [RadioIf] Packet RX (noise?) : 559ms
DEBUG | 13:03:21 2997 [ServerAPI] Got client heartbeat
DEBUG | 13:03:21 2997 [ServerAPI] FromRadio=STATE_SEND_QUEUE_STATUS, numbytes=6
WARN  | 13:03:26 3001 caught missed RX_DONE
ERROR | 13:03:26 3001 [RadioIf] Ignore received packet due to error=-7 (maybe id=0xbb340269 fr=0x7ba9adfe to=0xffffffff flags=0xa1 rxSNR=-19.25 rxRSSI=-119 nextHop=0x0 relay=0xf4)
DEBUG | 13:03:26 3001 [RadioIf] Packet RX (noise?) : 559ms
INFO  | 13:03:31 3007 [DeviceTelemetry] Send: air_util_tx=0.737639, channel_utilization=5.863334, battery_level=101, voltage=0.000000, uptime=3007
DEBUG | 13:03:31 3007 [DeviceTelemetry] Partially randomized packet id 869859702
DEBUG | 13:03:31 3007 [DeviceTelemetry] updateTelemetry LOCAL
```

```
DEBUG | 13:10:37 3433 [RadioIf] Ignore false preamble detection
DEBUG | 13:10:37 3433 [RadioIf] Started Tx (id=0x4b9b42a4 fr=0xda621dc4 to=0xffffffff, transport = 1, WantAck=0, HopLim=1 Ch=0x8 encrypted len=44 rxtime=1771179036 rxSNR=-10.75 rxRSSI=-112 h
DEBUG | 13:10:37 3433 [RadioIf] 0 packets remain in the TX queue
DEBUG | 13:10:38 3434 [RadioIf] Packet TX: 559ms
DEBUG | 13:10:38 3434 [RadioIf] Completed sending (id=0x4b9b42a4 fr=0xda621dc4 to=0xffffffff, transport = 1, WantAck=0, HopLim=1 Ch=0x8 encrypted len=44 rxtime=1771179036 rxSNR=-10.75 rxRSSI
WARN  | 13:10:40 3436 caught missed RX_DONE
DEBUG | 13:10:40 3436 [RadioIf] Corrected frequency offset: -126.906242
DEBUG | 13:10:40 3436 [RadioIf] Lora RX (id=0x4b9b42a4 fr=0xda621dc4 to=0xffffffff, transport = 0, WantAck=0, HopLim=0 Ch=0x8 encrypted len=44 rxSNR=-21 rxRSSI=-122 hopStart=7 relay=0x28)
DEBUG | 13:10:40 3436 [RadioIf] Packet RX: 559ms
INFO  | 13:10:40 3436 [Router] Packet History - insert: Reusing slot aged 3.616s TRACE MATCHED PACKET
DEBUG | 13:10:40 3436 [Router] Ignore dupe incoming msg (id=0x4b9b42a4 fr=0xda621dc4 to=0xffffffff, transport = 1, WantAck=0, HopLim=0 Ch=0x8 encrypted len=44 rxSNR=-21 rxRSSI=-122 hopStart
DEBUG | 13:10:40 3436 [Router] cancelSending id=0x4b9b42a4, removed=0
DEBUG | 13:10:40 3436 [Router] Incoming msg was filtered from 0xda621dc4
```

(It is also possible that these aren't truly missed, but had `notification` set by the ISR, and not yet handled by the main loop. That might be worth calling out in the log line.)

## 🤝 Attestations

  - [X] Station G2

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -192,8 +192,6 @@ bool kb_found = false;
 // global bool to record that on-screen keyboard (OSK) is present
 bool osk_found = false;
 
-unsigned long last_listen = 0;
-
 // The I2C address of the RTC Module (if found)
 ScanI2C::DeviceAddress rtc_found = ScanI2C::ADDRESS_NONE;
 // The I2C address of the Accelerometer (if found)
@@ -1127,12 +1125,6 @@ void loop()
             lastRadioMissedIrqPoll = millis();
             RadioLibInterface::instance->pollMissedIrqs();
         }
-    }
-
-    if (RadioLibInterface::instance != nullptr && !Throttle::isWithinTimespanMs(last_listen, 1000 * 60) &&
-        !(RadioLibInterface::instance->isSending() || RadioLibInterface::instance->isActivelyReceiving())) {
-        RadioLibInterface::instance->startReceive();
-        LOG_DEBUG("attempting AGC reset");
     }
 
 #ifdef DEBUG_STACK
--- a/src/main.h
+++ b/src/main.h
@@ -33,7 +33,6 @@ extern ScanI2C::DeviceAddress cardkb_found;
 extern uint8_t kb_model;
 extern bool kb_found;
 extern bool osk_found;
-extern unsigned long last_listen;
 extern ScanI2C::DeviceAddress rtc_found;
 extern ScanI2C::DeviceAddress accelerometer_found;
 extern ScanI2C::FoundDevice rgb_found;
--- a/src/mesh/RadioLibInterface.cpp
+++ b/src/mesh/RadioLibInterface.cpp
@@ -517,8 +517,6 @@ void RadioLibInterface::handleReceiveInterrupt()
 
 void RadioLibInterface::startReceive()
 {
-    // Note the updated timestamp, to avoid unneeded AGC resets
-    last_listen = millis();
     isReceiving = true;
     powerMon->setState(meshtastic_PowerMon_State_Lora_RXOn);
 }
--- a/src/mesh/SX126xInterface.cpp
+++ b/src/mesh/SX126xInterface.cpp
@@ -6,6 +6,10 @@
 #ifdef ARCH_PORTDUINO
 #include "PortduinoGlue.h"
 #endif
+#if defined(USE_GC1109_PA) && defined(ARCH_ESP32)
+#include <driver/rtc_io.h>
+#include <esp_sleep.h>
+#endif
 
 #include "Throttle.h"
 
@@ -55,14 +59,33 @@ template <typename T> bool SX126xInterface<T>::init()
 #if defined(USE_GC1109_PA)
     // GC1109 FEM chip initialization
     // See variant.h for full pin mapping and control logic documentation
+    //
+    // On deep sleep wake, PA_POWER and PA_EN are held HIGH by RTC latch (set in
+    // enableLoraInterrupt). We configure GPIO registers before releasing the hold
+    // so the pad transitions atomically from held-HIGH to register-HIGH with no
+    // power glitch. On cold boot the hold_dis is a harmless no-op.
 
     // VFEM_Ctrl (LORA_PA_POWER): Power enable for GC1109 LDO (always on)
     pinMode(LORA_PA_POWER, OUTPUT);
     digitalWrite(LORA_PA_POWER, HIGH);
+    rtc_gpio_hold_dis((gpio_num_t)LORA_PA_POWER);
+
+    // TLV75733P LDO has ~550us startup time (datasheet tSTR). On cold boot, wait
+    // for VBAT to stabilise before driving CSD/CPS, per GC1109 requirement:
+    // "VBAT must be prior to CSD/CPS/CTX for the power on sequence"
+    // On deep sleep wake the LDO was held on via RTC latch, so no delay needed.
+#if defined(ARCH_ESP32)
+    if (esp_sleep_get_wakeup_cause() == ESP_SLEEP_WAKEUP_UNDEFINED) {
+        delayMicroseconds(1000);
+    }
+#else
+    delayMicroseconds(1000);
+#endif
 
     // CSD (LORA_PA_EN): Chip enable - must be HIGH to enable GC1109 for both RX and TX
     pinMode(LORA_PA_EN, OUTPUT);
     digitalWrite(LORA_PA_EN, HIGH);
+    rtc_gpio_hold_dis((gpio_num_t)LORA_PA_EN);
 
     // CPS (LORA_PA_TX_EN): PA mode select - HIGH enables full PA during TX, LOW for RX (don't care)
     // Note: TX/RX path switching (CTX) is handled by DIO2 via SX126X_DIO2_AS_RF_SWITCH
--- a/src/sleep.cpp
+++ b/src/sleep.cpp
@@ -162,6 +162,13 @@ void initDeepSleep()
     if (wakeCause != ESP_SLEEP_WAKEUP_UNDEFINED) {
         LOG_DEBUG("Disable any holds on RTC IO pads");
         for (uint8_t i = 0; i <= GPIO_NUM_MAX; i++) {
+#if defined(USE_GC1109_PA)
+            // Skip GC1109 FEM power pins - they are held HIGH during deep sleep to keep
+            // the LNA active for RX wake. Released later in SX126xInterface::init() after
+            // GPIO registers are set HIGH first, avoiding a power glitch.
+            if (i == LORA_PA_POWER || i == LORA_PA_EN)
+                continue;
+#endif
             if (rtc_gpio_is_valid_gpio((gpio_num_t)i))
                 rtc_gpio_hold_dis((gpio_num_t)i);
 
@@ -556,8 +563,13 @@ void enableLoraInterrupt()
 #endif
 
 #if defined(USE_GC1109_PA)
-    gpio_pullup_en((gpio_num_t)LORA_PA_POWER);
-    gpio_pullup_en((gpio_num_t)LORA_PA_EN);
+    // Keep GC1109 FEM powered during deep sleep so LNA remains active for RX wake.
+    // Set PA_POWER and PA_EN HIGH (overrides SX126xInterface::sleep() shutdown),
+    // then latch with RTC hold so the state survives deep sleep.
+    digitalWrite(LORA_PA_POWER, HIGH);
+    rtc_gpio_hold_en((gpio_num_t)LORA_PA_POWER);
+    digitalWrite(LORA_PA_EN, HIGH);
+    rtc_gpio_hold_en((gpio_num_t)LORA_PA_EN);
     gpio_pulldown_en((gpio_num_t)LORA_PA_TX_EN);
 #endif
 


src/concurrency/NotifiedWorkerThread.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "OSThread.h"
+#include <atomic>
```
caveman99, line 4: There are no true atomics on RP2040 (Cortex M0+). These are simulated by briefly disabling interrupts. Wouldn't that have exactly the opposite effect of what we actually want?

caveman99 (changes requested):

src/concurrency/NotifiedWorkerThread.h
compumike: Interesting, did not know that about the RP2040!

Assuming that the "simulated" behavior is implemented properly, I still think "briefly disabling interrupts" is the desired behavior:

-----

Without this PR:

```
n = notification;
// INTERRUPT HANDLER FIRES HERE: sets notification=1
notification = 0; // BUG: this clobbers the new notification!
```

The `notification=1` from the interrupt handler is lost forever!

-----

With this PR:

```
save_and_disable_interrupts(); // pseudocode
n = notification;
// INTERRUPT PIN RISES HERE
notification = 0;
restore_interrupts_from_disabled(); // pseudocode

// INTERRUPT HANDLER FIRES HERE: while the rising edge is detected and latched, the interrupt handler only runs immediately *after* re-enabling interrupts, and it sets notification=1.
// While this won't be seen on this iteration of checkNotification, it will be seen on the next call of checkNotification.
// (Is this another potential issue if we go to sleep for a while?)
```

This does depend on some assumptions what it means for interrupts to be disabled. I think it means a CPU-wide disabling of interrupt handlers running, but doesn't prevent the interrupt from being latched for later dispatch when interrupts are re-enabled, and doesn't remove the interrupt handler. (This is very different from `disableInterrupt()` which actually detaches the interrupt handler.)

To support that: https://pip-assets.raspberrypi.com/categories/609-microcontroller-boards/documents/RP-009085-KB-1-raspberry-pi-pico-c-sdk.pdf?disposition=inline says:
- page 167: "edge interrupts are stored in the INTR register and can be cleared by writing to the INTR register."

`save_and_disable_interrupts()` and `restore_interrupts_from_disabled()` are defined here: https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2_common/hardware_sync/include/hardware/sync.h#L227-L291

pico_atomic is here: https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2_common/pico_atomic/atomic.c

stdatomic.h is here: https://github.com/raspberrypi/pico-sdk/blob/a1438dff1d38bd9c65dbd693f0e5db4b9ae91779/src/rp2_common/pico_atomic/include/stdatomic.h

GCC at least says "If there is no pattern or mechanism to provide a lock-free instruction sequence, a call is made to an external routine with the same parameters to be resolved at run time." https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html and I believe this references the definitions in atomic.c.

compumike:
To get a sense of order of magnitude for these issues causing missed `RX_DONE` interrupts:

### `NotifiedWorkerThread::checkNotification()`

Before this PR, `NotifiedWorkerThread::checkNotification()` is vulnerable to a race condition (losing the notification interrupt forever!) for only probably ~1 cycle per call to `checkNotification`.

Back of the envelope: 1 cycle @ 100 MHz (10 ns), with 1 call to `checkNotification` every 100 ms: any given interrupt has a **1 in 10 million** chance of getting lost. Not very common, but not nothing.

### `checkRxDoneIrqFlag()` immediately after `enableInterrupts`:

`enableInterrupts` happens once per packet. However, the vulnerable duration is much much much longer:

`SX126x::startReceiveCommon` has SPI `setPacketParams` after `clearIrqStatus`, https://github.com/jgromes/RadioLib/blob/87bf5785a3ca58dae184cff7b9860bef84a09b5c/src/modules/SX126x/SX126x.cpp#L553C17-L584 which writes about 6 bytes over SPI. 48 bits at 4 Mbps = 12 us. About 1000x larger vulnerable window compared to `checkNotification`.

Hopefully it is rare to get a RX_DONE interrupt within 12 us after `clearIrqStatus`. This would depend partially on the SX1262's internal implementation, for example whether it could still be in the middle of hearing a new preamble during standby mode just after receiving a packet (I think it likely would!). But order-of-magnitude wise, if randomly distributed and we assume a packet is in the 100s of milliseconds duration, then it could be something like **1 in 10000 packets**.

## Consequences

The consequence of either of these race conditions is a missed `RX_DONE` interrupt, which shows up as a receiver that

> radios going deaf from time to time

until some other watchdog like https://github.com/meshtastic/firmware/pull/8163 kicks in.

## Does `TX_DONE` need similar?

Probably, but I wanted to keep this PR relatively focused for now.

src/mesh/RF95Interface.cpp
```diff
@@ -301,6 +301,7 @@ void RF95Interface::startReceive()
 
     // Must be done AFTER, starting receive, because startReceive clears (possibly stale) interrupt pending register bits
     enableInterrupt(isrRxLevel0);
+    checkRxDoneIrqFlag();
```
GUVWAF, line 304: Shall we move `checkRxDoneIrqFlag()` into RadioLibInterface's `enableInterrupt()`?

thebentern:
@compumike amazing detective work! Tag me when you have things buttoned up and we can take another look.

src/mesh/RF95Interface.cpp
compumike: @GUVWAF I looked at that, but in `RadioLibInterface.h` it's just `virtual void enableInterrupt(void (*)()) = 0;` which then is overridden, for example in `SX126xInterface.h` as `virtual void enableInterrupt(void (*callback)()) { lora.setDio1Action(callback); }`.

I'd be happy to move the `checkRxDoneIrqFlag` call into each of those (four files, `SX126xInterface.h` etc), such as `virtual void enableInterrupt(void (*callback)()) { lora.setDio1Action(callback); checkRxDoneIrqFlag(); }` if you think that's cleaner than the current implementation!

compumike:
@thebentern I've been running a build with this PR + https://github.com/meshtastic/firmware/pull/9571 + https://github.com/meshtastic/firmware/pull/9572 _(or, at least, whatever those two PRs looked like 5 days ago!)_ for 5 days and everything looks good. Anecdotally it seems like I'm getting more packets through successfully (more successful roundtrip traceroutes from a parking lot where it's usually intermittent), but that's not a controlled experiment and I don't have hard data to back that up. And if there is any noticeable improvement, it's probably due to the other two PRs, not this one.

src/mesh/RF95Interface.cpp
GUVWAF: Ah, I see, sorry for the confusion. Hmm, no then maybe doing it explicitly like you have now is better I think.

PR #9659, opened 2026-02-15
Ebyte E22P-868M30S and E22P-915M30S modules support for existing Hydra PCBs

And137:
All changes are contained within the Hydra variant. E22P support is now available as a build flag “-D EBYTE_E22P-868M30S”. Additional commented flags have been added for Hydra customization.

## Summary by CodeRabbit

* **New Features**
  * Added optional build configuration for EBYTE E22P-868M30S and E22P-915M30S radio modules.
  * Added optional PSRAM support and configuration for increasing the maximum node count to 150.
  * Updated radio pin configuration for compatible EBYTE modules while preserving standard Hydra board behavior.


fifieldt:
@And137 thanks for the path. this should be targeted against the develop branch, prior to inclusion in master :)

And137:
> @And137 thanks for the path. this should be targeted against the develop branch, prior to inclusion in master :)

Sorry, I had simply used the master branch before (quite some time ago), and there were no problems.

And137:
I apologize for the confusion. I haven’t added much recently, and the project is moving forward.

KokoSoft (approved):

variants/esp32/diy/hydra/variant.h
```diff
@@ -33,7 +33,13 @@
 #define SX126X_DIO1 33  // EBYTE module's DIO1 pin
 
 #define SX126X_TXEN 13 // Schematic connects EBYTE module's TXEN pin to MCU
-#define SX126X_RXEN 14 // Schematic connects EBYTE module's RXEN pin to MCU
+
+#if defined(USE_EBYTE_E22P) // For Ebyte E22P-868M30S and E22P-915M30S modules - pin 14 always HIGH
+    #define SX126X_RXEN -1
```
KokoSoft: Why not use `RADIOLIB_NC` instead `-1`?

variants/esp32/diy/hydra/variant.h
```diff
@@ -33,7 +33,13 @@
 #define SX126X_DIO1 33  // EBYTE module's DIO1 pin
 
 #define SX126X_TXEN 13 // Schematic connects EBYTE module's TXEN pin to MCU
```
KokoSoft, line 35: Do we need `SX126X_TXEN` for `USE_EBYTE_E22P`? Is this pin not controlled by RF_SW (DIO2)?

variants/esp32/diy/hydra/variant.h
```diff
@@ -33,7 +33,13 @@
...
-#define SX126X_RXEN 14 // Schematic connects EBYTE module's RXEN pin to MCU
+
+#if defined(USE_EBYTE_E22P) // For Ebyte E22P-868M30S and E22P-915M30S modules - pin 14 always HIGH
+    #define SX126X_RXEN -1
+    #define E22P_ME 14
```
KokoSoft: Wouldn't it be better to use `SX126X_POWER_EN` instead of adding more definitions to the project?

KokoSoft (commented):
Maybe adding global support for E22P modules would make more sense, since not only Hydra boards will use them.

And137:
> Maybe adding global support for E22P modules would make more sense, since not only Hydra boards will use them.

I’ve tried but I got reply that the changes should stay within viariant…

variants/esp32/diy/hydra/variant.h
And137: I will check this solution when I have time, thank you.

variants/esp32/diy/hydra/variant.h
And137: The standard template for E22 uses TX and RX pins and the switching mechanism from radiolib. People are wiring the EN pin to 3,3V for E22P and they leave the TX control as it were for E22. That’s the reasoning behind coding this way.

KokoSoft: Unless `TXEN` is connected to `DIO2` on the Hydra board, we should probably just leave it as is

KokoSoft:
There appear to be differences between the E22P modules.
> For the E22P-433M30S, the RF_switch is controlled by TX_EN and RX_EN.

variants/esp32/diy/hydra/variant.h
And137: The changes are now minimal. I've deleted variant.cpp because the variant now uses SX126X_EN, as suggested.

variants/esp32/diy/hydra/variant.h
And137: Done, thanks.

And137:
> There appear to be differences between the E22P modules.
> 
> > For the E22P-433M30S, the RF_switch is controlled by TX_EN and RX_EN.

Versions for higher frequencies have modified controls, which make them less prone to damage. The 433 MHz version can be operated using the standard Hydra variant. I've added a warning in the comment.

NomDeTom:
@And137 Do you have the power curves for these modules? Do you want to add them to the defines in here:

https://github.com/meshtastic/firmware/blob/f88bc732ccc7f852986179b91939322b89ef8e09/src/configuration.h#L107-L190

And137:
> firmware/src/configuration.h

For now, I only have a rather imprecise power graph provided by the manufacturer. However, I have standardized the naming of the compilation flag so that in the near future it will be easy to introduce scaling in TrafficManagementModule.cpp after performing power measurements.

PR #9660, opened 2026-02-16
Add USB_MODE=1 for Station G2 - Solving all my serial issues.

h3lix1:
## Summary
Enable native USB CDC mode (ARDUINO_USB_MODE=1) for the Station G2 board so it uses the ESP32-S3’s built-in USB for serial communication instead of UART.

This little change has fixed all my problems connecting to the Station G2 from my linux node over serial.. including:
* Meshtastic CLI not saving anything correctly the first time
* The inability to use --configure properly
* The G2 falling off the USB bus, requiring a reboot to recover
* Many firmware upgrade failures trying to get the G2 into GW upgrade mode
* Random strange disconnects, complaining about "port in use" etc.
* meshtastic command taking much longer to complete than it should

## 🤝 Attestations

* [x] I have tested that my proposed changes behave as described.
* [x] I have tested that my proposed changes do not cause any obvious regressions on the following devices:

**Devices tested**

* [x] Heltec (Lora32) V4
* [] LilyGo T-Deck
* [] LilyGo T-Beam
* RAK WisBlock 4631
* Seeed Studio T-1000E tracker card *(Does not add/subtract functionality)*
* [x] Station G2
* [] LILYGO Pager

### DIFF
--- a/src/AudioThread.h
+++ b/src/AudioThread.h
@@ -4,6 +4,7 @@
 #include "configuration.h"
 #include "main.h"
 #include "sleep.h"
+#include <memory>
 
 #ifdef HAS_I2S
 #include <AudioFileSourcePROGMEM.h>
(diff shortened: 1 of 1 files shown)


h3lix1:
I'd rather keep things consistent, but I'll take the advice of the maintainers here.

PR #9666, opened 2026-02-16
Heltec-v3 on battery: button and rotary encoder responsiveness fix

### LINKED ISSUE #8783: [Bug]: Heltec LoRa32 V3 Button issue
### Category

Hardware Compatibility

### Hardware

Heltec V3

- [ ] InkHUD ePaper
- [x] OLED slide UI on any display

### Firmware Version

2.7.16.c3a7ad286

### Description

User button press registration is very inconsistent. Sometimes i need to click 3 or 4 times to switch to another screen slide.

### Relevant log output

```Shell

```

tomagaw:
## Heltec-V3 button (and rotary encoder) issue resolved ##
I made changes so the OS is forced to handle user button and rotary encoder input constantly instead of relying on interrupts. The problem was that those input methods were very unreliable on Heltec-V3 when working on battery.

### This is just a quick hack. ###
The "real" issue that caused corresponding functions not to be called reliably still remains somewhere else.
Additionally - although there is one more instance of **runOnce()** returning  **INT32_MAX** in *src/input/RotaryEncoderInterruptBase.cpp*, it doesn't seem to cause any trouble here.

fixes #8783

## 🤝 Attestations

  - [x] Heltec (Lora32) V3

### DIFF
--- a/src/input/HackadayCommunicatorKeyboard.cpp
+++ b/src/input/HackadayCommunicatorKeyboard.cpp
@@ -106,8 +106,8 @@ static unsigned char HackadayCommunicatorTapMap[_TCA8418_NUM_KEYS][2] = {{},
                                                                          {}};
 
 HackadayCommunicatorKeyboard::HackadayCommunicatorKeyboard()
-    : TCA8418KeyboardBase(_TCA8418_ROWS, _TCA8418_COLS), modifierFlag(0), last_modifier_time(0), last_key(-1), next_key(-1),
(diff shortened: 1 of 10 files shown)


caveman99:
So you essentially switch from IRQ based handling to polling ? That essentially strains the rest of the cooperative multitasking with yet unknown consequences. Please file a bug report instead so we can look into this.

tomagaw:
I see. Exactly what I expected with this *"solution"*. The bug report is already there #8783

Issue #7093, opened 2025-06-21
[Feature Request]: Allow temperature sensor offset calibration on devices like SenseCAP T1000-E


argoufra:
### Platform

NRF52

### Description

Hi,

I’m using a SenseCAP T1000-E with Meshtastic. The built-in temperature sensor seems to measure a few degrees too high, probably because of heat from the electronics inside the case.

It would be great to have a simple setting to apply a temperature offset (for example: -3°C), so the reported temperature is more accurate.

Thanks for your work on this great project!

Issue #5476, opened 2024-11-29
[Feature Request]: Possibility to manually apply offset for temperature, humidity and barometric pressure measurements (BME280 and other sensors)


Synthacon:
### Platform

ESP32

### Description

It would be useful to have built-in support for applying user-defined offsets to the temperature, humidity, and barometric pressure readings for the BME280 results. This sensor can't be recalibrated and humidity value drifts with the age. This option would allow users to easily adjust for systematic inaccuracies or environmental influences (e.g., heat from nearby components).

Correct for localized heating affecting temperature and humidity readings.
Adjust for known inaccuracies in factory calibration.
Enable quick fine-tuning for specific environmental conditions.

Just as example my sensor shows +0.64c and -10% for humidity and currently there is no way to adjust this.

HarukiToreda:
this is indeed a problem plaguing most BME680 sensors as well as they are not calibrated for accuracy at factory and there's no way to do so on the board itself. It must be offset by the firmware to be able to get accurate readings.

dm5tt:
They are measuring the PCB temperature and not environmental temperature. So maybe they work as designed?

Issue #5469, opened 2024-11-28
 for [Feature Request]: Possibility to manually apply offset for temperature, humidity and barometric pressure measurements (BME280))


Synthacon:
### Platform

ESP32, Cross-Platform, other

### Description

It would be useful to have built-in support (Android app) for applying user-defined offsets to the temperature, humidity, and barometric pressure readings for the BME280 results. This would allow users to easily adjust for systematic inaccuracies or environmental influences (e.g., heat from nearby components). 

- Correct for localized heating affecting temperature and humidity readings.
- Adjust for known inaccuracies in factory calibration.
- Enable quick fine-tuning for specific environmental conditions.

Just as example my sensor  shows +0.64c and -10% for humidity and currently there is no way to adjust this.

Issue #4267, opened 2024-07-11
[Feature Request]: temperature offset


jranma:
### Platform

other

### Description

TELEMETRY: the reported temperature may be quite inaccurate. This may be caused by the sensor close to another component that emits heat. 
An offset or calibration system for the sensor would be appreciated.

garthvh:
Sensor calibration needs to be handled by the device

PR #9682, opened 2026-02-18
BaseUI: Temperature offset and number picker upgrade

HarukiToreda:
This PR adds a temperature offset to local temperature before being broadcasted so users can manually correct sensor readings that are a slightly off.

What this PR does:
- Adds a new **Set Temp Offset** option in the Environment screen action menu.
- Lets users enter positive or negative values with decimals.
- Shows and edits the offset in the user’s preferred units (C or F).
- Saves the offset so it stays after reboot.
- Applies the offset to this device’s own environment telemetry readings.
- Local sensor data now shows if there's no data received from the mesh or is invalid, while showing local sensor measurements, data displayed is live to make calibration more accurate.

It also improves consistency:
- Key Verification now uses the same number picker style, so the UI feels consistent.
- Shared picker logic reduces duplicate code and makes future reuse easier.

Notes:
- This does **not** change telemetry values received from other nodes.

https://github.com/user-attachments/assets/29c57244-4313-4736-b286-99d2ee389ac9


HarukiToreda:
Addresses issues


- https://github.com/meshtastic/firmware/issues/4267
- https://github.com/meshtastic/firmware/issues/5469

- https://github.com/meshtastic/firmware/issues/5476

Discussion:
- https://github.com/meshtastic/firmware/discussions/5482

DOES NOT Address Issue
- https://github.com/meshtastic/firmware/issues/7093
This feature uses protos that the apps and other clients can use to offset the value, but since this is currently only available on BaseUI which requires a screen, it is not possible to do on the T1000E

weerdenburg:
Thank you very much for adding this new **Set Temp Offset** option.
I only wonder if we don't also need to be able to set an offset for "humidity and barometric pressure" as @Synthacon mentioned the following in https://github.com/meshtastic/firmware/discussions/5482
> Just as example my sensor shows **+0.64c** and **-10%** for humidity and currently there is no way to adjust this.

HarukiToreda:
> Thank you very much for adding this new **Set Temp Offset** option. I only wonder if we don't also need to be able to set an offset for "humidity and barometric pressure" as @Synthacon mentioned the following in #5482
> 
> > Just as example my sensor shows **+0.64c** and **-10%** for humidity and currently there is no way to adjust this.

The sensors do not use temperature to calculate those other values, so applying a temperature offset will not affect them.

Temperature is something I am familiar with and have the proper tools to calibrate accurately. For now, this PR will focus on temperature calibration.

I do not currently have a reliable way to test or validate the other two values. Making changes to them without fully understanding their behavior or having the ability to properly test is not something I'm comfortable including in this PR.

jem0n:
Any chance that this functionality could also be included in InkHUD?

HarukiToreda:
> Any chance that this functionality could also be included in InkHUD?

certainly

weerdenburg:
> I do not currently have a reliable way to test or validate the other two values. Making changes to them without fully understanding their behavior or having the ability to properly test is not something I'm comfortable including in this PR.

OK, that makes sense. I agree we should get the temperature offset merged first. After that’s in a release, we can look into whether any of the other sensor values also need an offset (and if there’s a reliable way to test/validate them).

Xaositek (approved):
Tested which allowed me to offset my temperature by -3.5F - great effort here!

caveman99:
We discussed this before and decided against simple offset compensation for suboptimal sensor placement. For one you need a compensation curve, not a simple offset to do it right. Secondly most of these problems can be solved by correct sensor placement, especially on a sensor extension cable. Thirdly there are some combined sensors out there that have a very bad temperature probe, whichis only used to compensate internal sensor readings of another value. While these readings are exposed, it's not advisable to use this as a valid ambient temperature, most of the times this is the sensor operating temperature.

This is not a verdict, just some items to add to the discussion. I fear if we merge this it will open up a can of worms of people needing compensation values on ALL the other reading we support.

weerdenburg:
> Secondly, most of these problems can be solved by correct sensor placement, especially on a sensor extension cable.
> 
> This is not a verdict, just some items to add to the discussion. I fear if we merge this, it will open up a can of worms of people needing compensation values on ALL the other reading we support.

Even if I place my BME280 outside the box, it still has an offset against the temperature that I measure on a dedicated temp meter next to it. And that can simply be adjusted with a simple offset, like used in this pull request.

I do understand a bit of your fear, but at this moment, the temperature readings are not usable at all, so I would rather have a simple offset to get the reading back into the correct "range".

And, yes, as I wrote before, I do think that other sensor values should have an offset option.
> After that’s in a release, we can look into whether any of the other sensor values also need an offset (and if there’s a reliable way to test/validate them).

I do like your opinion, but I gave your message a thumbs down, because I don't want you guys to downvote this long-awaited pull request ;-)

HarukiToreda:
> We discussed this before and decided against simple offset compensation for suboptimal sensor placement. For one you need a compensation curve, not a simple offset to do it right. Secondly most of these problems can be solved by correct sensor placement, especially on a sensor extension cable. Thirdly there are some combined sensors out there that have a very bad temperature probe, whichis only used to compensate internal sensor readings of another value. While these readings are exposed, it's not advisable to use this as a valid ambient temperature, most of the times this is the sensor operating temperature.
> 
> This is not a verdict, just some items to add to the discussion. I fear if we merge this it will open up a can of worms of people needing compensation values on ALL the other reading we support.

Unfortunately, this is not something that can be corrected through sensor placement. I have tested every sensor currently on the supported list, placing each one outside of the enclosure and away from any other components to eliminate additional variables.

Across all of them, I consistently observed a ±1 to 2°F deviation from ambient temperature. These tests were conducted over full day periods to allow the internal sensor elements to reach stable operating temperatures and eliminate warm-up bias.

Based on those results, I found that applying an offset is very effective. It remains stable even as temperatures fluctuate, including across a range from 34°F to 85°F during my testing.

If the concern is that this creates an illusion of accuracy, I would argue the opposite. Displaying a known incorrect value without correction is more misleading than allowing a user to apply a measured and calibrated adjustment, which is exactly what this PR enables.

src/modules/Telemetry/EnvironmentTelemetry.cpp
```diff
@@ -272,6 +299,29 @@ int32_t EnvironmentTelemetryModule::runOnce()
         return disable();
     }
 
+    auto refreshLocalMeasurementPacket = [this]() {
```
oscgonfer, line 290: Could this part be left from other changes not needed on this PR?

src/modules/Telemetry/EnvironmentTelemetry.cpp
```diff
@@ -631,11 +686,15 @@ bool EnvironmentTelemetryModule::sendTelemetry(NodeNum dest, bool phoneOnly)
             p->priority = meshtastic_MeshPacket_Priority_RELIABLE;
         else
             p->priority = meshtastic_MeshPacket_Priority_BACKGROUND;
-        // release previous packet before occupying a new spot
-        if (lastMeasurementPacket != nullptr)
-            packetPool.release(lastMeasurementPacket);
-
-        lastMeasurementPacket = packetPool.allocCopy(*p);
+        const bool shouldReplaceDisplayPacket =
```
oscgonfer, line 687: Similar to this part

src/mesh/MeshModule.h
```diff
@@ -81,6 +81,7 @@ class MeshModule
     static AdminMessageHandleResult handleAdminMessageForAllModules(const meshtastic_MeshPacket &mp,
                                                                     meshtastic_AdminMessage *request,
                                                                     meshtastic_AdminMessage *response);
+    const char *getName() const { return name; }
```
oscgonfer, line 84: And this

oscgonfer:
If I can add my two cents to the conversation... In the SC project we did something [similar](https://forum.smartcitizen.me/t/new-firmware-feature-temperature-and-relative-humidity-corrections/2018) to this correction. This helps users trust the data and not be thrown off by the readings. Temperature is something that people can relate to very easily, and it can generate mistrust on the data right away. In our case, however, there is a key difference - we added the possibility of this correction only for a certain sensor, and for hardware that we _really_ know well. We tested it consistently and concluded that an offset was enough in our case. In our case (and in general), there are two possible sources of error in temp/rh readings: device self-heat, and inter-device deviation, and they _should_ be treated differently, and it's true that they may not be linear.

That being said, considering all the tests you did for the supported sensors, it'd be maybe worth documenting them _somewhere_ for future traceability and to justify that the offset is a good-enough correction.

From a _metrology_ point of view, however, it is literally the same to have an offset or not if the offset is not reported _somewhere_ (because the offset is applied, but what was it?). Because of that, readings are basically _always going to be subject to correction_. The good thing is that, since this correction doesn't _fundamentally_ change the values, it can be used to enhance user trust, while more complex _postprocessing_ can still be applied on a separate platform (this is what we do in SC) in a more formal way.

One thing that could be a follow up to this effort (and I'd strongly suggest to consider it) is to add, as others say, RH compensation at the very least, even if it can't be tested at the moment with accurate sensors, _in principle_ it should work very similarly. If a T+RH sensor is off in temperature, it will be always be off in both metrics, as those sensors don't measure RH per se, but they estimate it based on the same signal as the one used for temperature.

src/modules/Telemetry/EnvironmentTelemetry.cpp
HarukiToreda: I could remove it from changes, but I’d like to clarify why it’s there.

With the way the current environmental screen is implemented, locally measured sensor data does not display right away. Even though the readings come from our own sensors, they only appear on screen after being sent over the mesh and received back while connected to the app.

Having to connect to the app and wait for that packet round trip just to see a local reading makes calibration very innefficient.

What I added allows locally measured data to be displayed if no environmental packets have yet been received from the mesh. Once the first environmental packet is heard, the screen reverts to the normal behavior and displays only mesh-received data, just as it did before. Now that I think about it I probably show label the source on the screen to say "Local" as to not be confused with a Mesh packet. 

This keeps the existing mesh logic intact while making calibration and initial setup significantly more practical. If you think this is unnecessary I can remove it.

src/modules/Telemetry/EnvironmentTelemetry.cpp
HarukiToreda: Purpose: don’t replace a remote mesh packet with local data once remote data is already being displayed.

HarukiToreda:
> ggest to consider it) is to add, as others say, RH compensation at the very least, even if it can't be tested at the moment with accurate sensors, _in principle_ it should work very similarly. If a T+RH sensor is off in temperature, it will be always be off in both metri

I completely agree, I am getting ready to redo all my test for this and document them on my[ site](https://harukitoreda.github.io/Meshtastic-Experiments/Battery-Runtime-Tests) where I dump all the random tests I make, once fully complete I'll see how to add them onto the Mesh repo. My first tests were done with only one sensor of each but ideally I want to redo them with at least 3 of each to account for manufacturing inconsistencies and a heat vision module to display the heat generated by each sensor if any. I do want to follow up with calibrating RH next, I haven't figure out a way to do this yet. 

Thank you for the SC reference, I didn't know it existed till now, this will save me time on how to best approach the future improvements

src/modules/Telemetry/EnvironmentTelemetry.cpp
weerdenburg: So this code always "fixes" the issue of not seeing the environmental data right after booting up the system?

I really hate it that a lot of times when I pick up the device, press a button... I first need to wait for the complete flow... 30sec. a few minutes,  before I can see the local environmental data. It's LOCAL data, so please ALWAYS show it to me.

> after being sent over the mesh and received back while connected to the app.
In my opinion, it's great if this PR also fixes this big irritation.

And maybe even a step further... why wouldn't we always want to show the most relevant local data?

HarukiToreda: yeah, the next step after this is to allow the user to choose the source like for example Local Only, Latest Mesh Packets, or Favorite Only Packets. This way you have control over what shows, But that's much later when this can get through.

CodeSkunk521:
Some information about how the offset may or may not be applied to RH calculation: https://github.com/esphome/issues/issues/4752

Synthacon:
> > We discussed this before and decided against simple offset compensation for suboptimal sensor placement. For one you need a compensation curve, not a simple offset to do it right. Secondly most of these problems can be solved by correct sensor placement, especially on a sensor extension cable. Thirdly there are some combined sensors out there that have a very bad temperature probe, whichis only used to compensate internal sensor readings of another value. While these readings are exposed, it's not advisable to use this as a valid ambient temperature, most of the times this is the sensor operating temperature.
> > This is not a verdict, just some items to add to the discussion. I fear if we merge this it will open up a can of worms of people needing compensation values on ALL the other reading we support.
> 
> Unfortunately, this is not something that can be corrected through sensor placement. I have tested every sensor currently on the supported list, placing each one outside of the enclosure and away from any other components to eliminate additional variables.
> 
> Across all of them, I consistently observed a ±1 to 2°F deviation from ambient temperature. These tests were conducted over full day periods to allow the internal sensor elements to reach stable operating temperatures and eliminate warm-up bias.
> 
> Based on those results, I found that applying an offset is very effective. It remains stable even as temperatures fluctuate, including across a range from 34°F to 85°F during my testing.
> 
> If the concern is that this creates an illusion of accuracy, I would argue the opposite. Displaying a known incorrect value without correction is more misleading than allowing a user to apply a measured and calibrated adjustment, which is exactly what this PR enables.

There might also be a slight heating effect if the values are read very frequently. This cheap sensor boards usually have multiple components, and reading values at a high frequency can cause heating and lead to errors. So limiting this could potentially reduce error. 

The whole idea of scientific measurement is to reduce error, it is utterly pointless to even have option for sensors if they are knowingly showing ;incorrect values. There must be away to calibrate or add offset.

Probably best source for actual fix is the datasheet of each supported swnsor...

HarukiToreda:
As I mentioned in previous comments, this is not meant to be a perfect solution. It is simply an option that allows users to consciously make small adjustments to compensate for known inconsistencies in sensors. It is not forced on anyone, if a user does not want to use it or does not trust it, they can simply leave it untouched.

This is similar to how the ADC calibration value works. Not everyone takes the time to measure their battery with a multimeter to verify that the voltage shown on the screen is accurate. However, for those who do want that level of accuracy. like I do, it provides the tool needed to offset it slightly, I know the BME680 Sensors I have, are always about 1.3° to 2° F off so knowing this I can close the degree of error manually. Maybe in the future someone can take the time to make a proper calibration per sensor, but till then leaving known incorrect values with no way to adjust it in a way is worse than a slight offset.

weerdenburg:
Hello, I would like to know the status of this great fix.
Maybe not a 100% perfect solution, but better than no adjustments.

When will this fix be released? I don't understand why this issue is automatically closed.

PR #9703, opened 2026-02-21
Add T-Beam BPF (144-148 Mhz LoRa)

vidplace7:
Add [LilyGo T-Beam BPF](https://lilygo.cc/en-us/products/t-beam-bpf), the first / only 144-148Mhz LoRa device :meat_on_bone:

Requires:
- Protobuf regen to include https://github.com/meshtastic/protobufs/pull/864
- Protobuf Ham 2M (144-148mhz) region additions


caveman99:
EU (ITU Region 1) 2m Band is 144-146MHz - Can we either make 2 presets or (assuming a HAM knows what they are doing) make a tighter default preset and tell people to deliberately set a custom frequency if they wanna go outside? This way an EU operator may not unknowingly exceed their boundaries.

vidplace7:
Blocked by #10120 (enabling frequency overrides, narrow presets)

vidplace7:
Depends on protobuf regen that includes:
- https://github.com/meshtastic/protobufs/commit/2e284d91a75d0cbf0fc08ab1d0fc33aaeb659bd8

vidplace7:
Moving this to a branch in-org `t-beam-bpf`

PR #9706, opened 2026-02-21
Add VL53L0 distance sensor.

smavridis:
Added new Environment Telemetry class VL53L0XSensor. Tested with a Heltec V3.

I have deployed it with a Heltec V3, currently my only meshtastic/lora device.
Expecting RAK and Xiao boards to ensure no issues, but delayed due to Chinese New Year.

This is my first contribution, i read the contribution guide and signed the CLA, let me know if i have missed anything.

Thank you for the Meshtastic project.

## 🤝 Attestations

  - [X] Heltec (Lora32) V3

### DIFF
--- a/src/detect/ScanI2CTwoWire.cpp
+++ b/src/detect/ScanI2CTwoWire.cpp
@@ -626,13 +626,17 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
                 }
                 break;
             case TSL25911_ADDR:
-                registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0xA0 | 0x12), 1);
-                if (registerValue == 0x50) {
-                    type = TSL2591;
-                    logFoundDevice("TSL25911", (uint8_t)addr.address);
+                if(getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0x12), 1) == 0x50) {
+                        type = TSL2591;
+                        logFoundDevice("TSL25911", (uint8_t)addr.address);
+                } else if ((getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0x0A), 1)&0xF0) == 0x50) {
+                        type = TSL2561;
+                        logFoundDevice("TSL2561", (uint8_t)addr.address);
+                } else if (getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0xC0), 1) == 0xEE) {
+                        type = VL53L0X;
+                        logFoundDevice("VL53L0X", (uint8_t)addr.address);
                 } else {
-                    type = TSL2561;
-                    logFoundDevice("TSL2561", (uint8_t)addr.address);
+                        LOG_INFO("Unknown device found at address 0x%x", (uint8_t)addr.address);
                 }
                 break;
 
--- a/src/modules/Telemetry/Sensor/VL53L0XSensor.cpp
+++ b/src/modules/Telemetry/Sensor/VL53L0XSensor.cpp
@@ -1,21 +1,91 @@
 #include "configuration.h"
-#include <memory>
 
 #if !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR && __has_include(<Adafruit_VL53L0X.h>)
 
 #include "../mesh/generated/meshtastic/telemetry.pb.h"
-#include "TelemetrySensor.h"
+#include <memory>
+
+
+
+#include "../mesh/generated/meshtastic/telemetry.pb.h"
+#include "FSCommon.h"
 #include "VL53L0XSensor.h"
+#include "SPILock.h"
+#include "SafeFile.h"
+#include "TelemetrySensor.h"
 
 #include <Adafruit_VL53L0X.h>
 #include <typeinfo>
+#include <pb_decode.h>
+#include <pb_encode.h>
+
+
+bool VL53L0XSensor::loadState()
+{
+#ifdef FSCom
+    spiLock->lock();
+    auto file = FSCom.open(VL53L0XStateFileName, FILE_O_READ);
+    bool okay = false;
+    if (file) {
+        LOG_INFO("%s state read from %s", sensorName, VL53L0XStateFileName);
+        pb_istream_t stream = {&readcb, &file, meshtastic_VL53L0XState_size};
+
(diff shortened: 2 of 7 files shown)


fifieldt:
Hey! Thanks for the patch :) For adding the sensor type, a patch needs to be made to the meshtastic/protobufs repo. If you can't find it let us know!

smavridis:
> Hey! Thanks for the patch :) For adding the sensor type, a patch needs to be made to the meshtastic/protobufs repo. If you can't find it let us know!

Thanks for the heads-up, i added the VL53L0 enum in TelemetrySensorType.
Is there anything else i am missing?

fifieldt:
That should be all that's required, thanks!

smavridis:
Thank you @fifieldt and @caveman99 for your help and guidance.
Glad to be part of this project.

src/detect/ScanI2CTwoWire.cpp
```diff
@@ -566,12 +566,21 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
...
             case TSL25911_ADDR:
                 registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0xA0 | 0x12), 1);
-                if (registerValue == 0x50) {
-                    type = TSL2591;
-                    logFoundDevice("TSL25911", (uint8_t)addr.address);
-                } else {
-                    type = TSL2561;
-                    logFoundDevice("TSL2561", (uint8_t)addr.address);
+                switch (registerValue) {
+                    case 0x00 ... 0x04:
```
robekl: `getRegisterValue()` defaults to `0x00` when no bytes are read, so bus/read failures can be misdetected as VL53L0X.

src/modules/Telemetry/Sensor/VL53L0XSensor.cpp
```diff
@@ -0,0 +1,39 @@
...
+    initI2CSensor();
+    return status;
+}
+
+bool VL53L0XSensor::getMetrics(meshtastic_Telemetry *measurement)
+{
+    VL53L0X_RangingMeasurementData_t measure;
+    vl53l0x.rangingTest(&measure, false);
+
+    measurement->variant.environment_metrics.has_distance = true;
```
robekl: `measure.RangeStatus` should be checked for out-of-range/error readings before deciding that the result is valid

src/detect/ScanI2CTwoWire.cpp
smavridis: Did not thinkof that and was testing with the sensor so wasnt checking that scenario, will fix.

src/modules/Telemetry/Sensor/VL53L0XSensor.cpp
smavridis: Good idea, i didn't know it had such Status.

smavridis:
I addressed the issues by @robekl (thanks), detection of sensors and ranging should be more robust.
Ideally i should have checked a setup with the other sensors, but i don't have any at hand.

src/detect/ScanI2CTwoWire.cpp
```diff
@@ -565,13 +565,17 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
                 }
                 break;
             case TSL25911_ADDR:
-                registerValue = getRegisterValue(ScanI2CTwoWire::RegisterLocation(addr, 0xA0 | 0x12), 1);
```
smavridis, line 629: Out of curiosity the bitwise OR of the register addresses always seemed wierd to me.
Doesn't that mean that the old code was reading register 0xB2 rather than 0xA0 and/or 0x12.
Am i missing some weird overload or c++ magic?

smavridis:
I converted this to a draft, since i decided to extend it a bit more.
Looking around i found i can add a State Message in to implement configuration of the sensor by the user.
I will be adding an option for setting the sensitivity/range of the sensor.
I think this will make it more usable for any other user that might be interested in the sensor.
Ill be following the SEN5X sensor as it seems to also have State.

smavridis:
I have added the State message and serialization/de-serialization logic.
I can see from the boot log that the load function gets called, but since the state was never written, no state gets loaded.

I think i am still missing some sort of registration so that the application knows that a configuration/option exists for the sensor but i cant really understand where/how that is done.

I did see a handleAdminMessage() function which i guess is relevant but i still think i am missing something.

smavridis:
So i tried to figure out how to add/modify the android application to add a configuration menu for the sensor ... unfortunately it is way above my skill level of Android/kotlin, etc ... i might try to touch it again at a latter time or hopefully some good soul does it instead.

Failing that i switched to providing a menu option in the firmware.
I have made the first steps and added a top level menu 'Sensors' in the System view.
I also considered initially to put it on the Environment Telemetry view, but since it is about received telemetry rather that local sensors i think it is not suitable.

At this point i wanted to get some feedback from you @fifieldt @robekl on the approach and any suggestions or ideas you might have.

I will try tomorrow to see how to call access the VL53L0X object to get/set state and configuration, so ill get back to you.

Thank you for your time.

src/mesh/generated/meshtastic/telemetry.pb.h
```diff
@@ -113,9 +113,18 @@ typedef enum _meshtastic_TelemetrySensorType {
     /* SCD30 CO2, humidity, temperature sensor */
     meshtastic_TelemetrySensorType_SCD30 = 49,
     /* SHT family of sensors for temperature and humidity */
-    meshtastic_TelemetrySensorType_SHTXX = 50
+    meshtastic_TelemetrySensorType_SHTXX = 50,
```
caveman99, line 116: Please revert ALL changes in generated files. Those will clash with the autogenerated headers.

caveman99 (changes requested):

caveman99:
The settings changes for your sensor need to be wired in the AdminModule and with additional functions in your sensor code, to process the changes. For testing this, it's best to use the python CLI client, since the protobufs can be updated there without too much hassle. Android App would be the next step.

src/mesh/generated/meshtastic/telemetry.pb.h
smavridis: Reverted all in updated commit.

smavridis:
Thanks for the feedback @caveman99.
I removed the generated files as requested.
I also modified the admin.proto and implemented a handleAdminMessage.
However i cant get/set the new variable so i am still missing something?
Perhaps that's what you mean by AdminModule, should i modify AdminModule.cpp?

PR #9712, opened 2026-02-22
Add possibility to disable sensors in multisensor devices

oscgonfer:
This PR is linked to https://github.com/meshtastic/protobufs/pull/865

The objective is to reduce airtime for those sensors that have multiple metrics (for instance a PM sensor). Many of them have groups of metrics that aren't always necessary or wanted, such as PM, PN, TRH, ...  This PR introduces a way to disable them via admin commands and to store that in `moduleConfig.telemetry`. Currently on AQ telemetry, but can be expanded to other multisensor devices easily.


src/modules/Telemetry/Sensor/SFA30Sensor.cpp
```diff
@@ -183,16 +183,67 @@ bool SFA30Sensor::getMetrics(meshtastic_Telemetry *measurement)
...
+    if (setDisables.has_disable_hcho) {
+        moduleConfig.telemetry.sensordisables.sfa30.disable_hcho = setDisables.disable_hcho;
+        LOG_INFO("%s %s HCHO metrics", sensorName, setDisables.disable_hcho ? "disabling" : "enabling");
+    }
+    if (setDisables.has_disable_trh) {
+        moduleConfig.telemetry.sensordisables.sfa30.disable_trh = setDisables.disable_trh;
+        LOG_INFO("%s %s T/RH metrics", sensorName, setDisables.disable_trh ? "disabling" : "enabling");
+    }
+
+    if (!nodeDB->saveToDisk(SEGMENT_MODULECONFIG))
```
oscgonfer, line 227: @HarukiToreda I saw how you handled this on your temp offset PR, but I don't seem to be able to make these settings persist. Do you see any flag of what I might be doing wrong here?

PR #9728, opened 2026-02-23
Update platformio.ini and variant.h for MeshLink V1.3 board configuration and pin assignments

ponzano:
Fix for GPS Serial pins, they had been messed with in some recent commits.
Updated with new GPS enable pin (Now working!)
Updated platformio.ini to get it in line with other boards
Some optimizations for E-INK screen variant reactivity and speed
Defined a GPIO (10) as default for user button


ponzano:
Hi, any chance this could get merged eventually? @caveman99 @thebentern @jp-bennett

PR #9734, opened 2026-02-24
Fix WisMesh Tap V2 env mess

thebentern:


mverch67:
Is it meant to be landscape or portrait orientation? Asking because of
```
  -D LGFX_SCREEN_WIDTH=240
  -D LGFX_SCREEN_HEIGHT=320
```

MUI currently only supports landscape mode for that resolution.

caveman99 (approved):
Maybe fold the TFT definition into the base build_flags

caveman99:
> Is it meant to be landscape or portrait orientation? Asking because of
> > ```
>   -D LGFX_SCREEN_WIDTH=240
>   -D LGFX_SCREEN_HEIGHT=320
> ```
> > MUI currently only supports landscape mode for that resolution.

The device i got is in landscape mode, antenna out the side. Sensible would be portrait for that one.

mverch67:
> The device i got is in landscape mode, antenna out the side. Sensible would be portrait for that one.

Use an antenna with a 90 degree angle 😉

PR #9737, opened 2026-02-24
InkHUD: Nodelist cleanup

HarukiToreda:
Added lines to separate nodes in nodelist applets like heard and recent, reduced the spacing so we can fit more on screen, removed the faded look when the name is too long and replaced it with ... so it doesn't look like the panel is failing.  Cleaned up the code to be more efficient with no regressions.
Old Look

New Look


src/graphics/niche/InkHUD/Applets/Bases/NodeList/NodeListApplet.cpp
```diff
@@ -120,9 +120,23 @@ void InkHUD::NodeListApplet::onRender(bool full)
...
+        constexpr const char *ellipsis = "...";
+        const uint16_t ellipsisW = getTextWidth(ellipsis);
+        if (maxWidth == 0)
+            return std::string();
+        if (getTextWidth(text) <= maxWidth)
+            return text;
+        if (ellipsisW > maxWidth)
+            return std::string();
+        while (!text.empty() && (getTextWidth(text) + ellipsisW > maxWidth))
+            text.pop_back();
```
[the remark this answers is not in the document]
HarukiToreda: names are short, row count is low, and redraw frequency is low, so impact is negligible. No need to do this

src/graphics/niche/InkHUD/Applets/Bases/NodeList/NodeListApplet.cpp
```diff
@@ -185,41 +194,46 @@ void InkHUD::NodeListApplet::onRender(bool full)
...
+        // Right-side labels and long name are rendered in small font.
         setFont(fontSmall);
-        printAt(width() - 1, lineBY, distance, RIGHT, MIDDLE);
+        uint16_t rightContentW = 0;
 
-        // If we have a direct connection to the node, draw the signal indicator
-        if (hopsAway == 0 && signal != SIGNAL_UNKNOWN) {
-            uint16_t signalW = getTextWidth("Xkm"); // Indicator should be similar width to distance label
+        // Top row right: signal indicator whenever signal data is known.
+        if (signal != SIGNAL_UNKNOWN) {
```
robekl: does removing the `hopsAway == 0` condition in the "if" mean that we will show the last-hop signal instead of the directly observed signal? that may be misleading

src/graphics/niche/InkHUD/Applets/Bases/NodeList/NodeListApplet.cpp
```diff
@@ -120,9 +120,23 @@ void InkHUD::NodeListApplet::onRender(bool full)
...
+    auto ellipsizeToWidth = [this](std::string text, uint16_t maxWidth) {
+        constexpr const char *ellipsis = "...";
+        const uint16_t ellipsisW = getTextWidth(ellipsis);
+        if (maxWidth == 0)
+            return std::string();
+        if (getTextWidth(text) <= maxWidth)
+            return text;
+        if (ellipsisW > maxWidth)
+            return std::string();
+        while (!text.empty() && (getTextWidth(text) + ellipsisW > maxWidth))
```
robekl: `getTextWidth(text)` is called multiple times here. call it once and save the result to avoid recalculation

src/graphics/niche/InkHUD/Applets/Bases/NodeList/NodeListApplet.cpp
HarukiToreda: you have a point, I will put it back, also this will allow distance to show again.

PR #9742, opened 2026-02-25
Fix RAK13302 1W FEM control: drive SKY66122-11 CSD/CPS/CTX directly

weebl2000:
The RAK13302's SKY66122-11 FEM has three control pins (CSD, CPS, CTX) that must be actively driven by the MCU. The existing code only managed CPS via SX126X_POWER_EN and relied on DIO2 for TX/RX switching, but R25 (DIO2→CTX) is NC by default, leaving CSD floating and CTX undriven. This put the FEM in an undefined state — the PA wouldn't reliably engage during TX and the 16 dB RX LNA was never active.

Add USE_SKY66122_FEM support (following the existing USE_GC1109_PA pattern) to explicitly initialize, sleep, and toggle all three FEM pins for correct TX/RX/shutdown modes.

See also https://github.com/meshcore-dev/MeshCore/pull/1836.

I don't have a RAK13302 device so I cannot test this myself.

References:

- [RAK13302 docs](https://docs.rakwireless.com/product-categories/wisblock/RAK13302/overview/)
- [SKY66122-11 FEM datasheet](https://www.skyworksinc.com/-/media/SkyWorks/Documents/Products/2901-3000/SKY66122-11_204782G.pdf)
- [SX1262 datasheet](https://semtech.my.salesforce.com/sfc/p/E0000000JelG/a/RQ000008nKCH/hp2iKwMDKWl34g1D3LBf_zC7TGBRIo2ff5LMnS8r19s)
rak13302-schematic
rak13302-power-explanation


weebl2000:
Fixed formatting

caveman99 (changes requested):
That's more or less identical patterns for both PA's Why don't aggregate them user USE_LORA_PA or something similar and only differentiate in the sleep call to not switch off the power for one of these?

I really don't like duplicate code blocks that do almost the same.

weebl2000:
I'm not sure if it would be clearer if we unified the FEM logic: https://github.com/weebl2000/meshtastic-firmware/pull/1/changes

Thoughts?

thebentern:
> I'm not sure if it would be clearer if we unified the FEM logic: https://github.com/weebl2000/meshtastic-firmware/pull/1/changes
> 
> Thoughts?

Looks cleaner to me

weebl2000:
> Looks cleaner to me

updated PR to include the unifying

src/sleep.cpp
```diff
@@ -567,15 +571,15 @@ void enableLoraInterrupt()
...
-    digitalWrite(LORA_PA_POWER, HIGH);
-    rtc_gpio_hold_en((gpio_num_t)LORA_PA_POWER);
-    digitalWrite(LORA_PA_EN, HIGH);
-    rtc_gpio_hold_en((gpio_num_t)LORA_PA_EN);
-    gpio_pulldown_en((gpio_num_t)LORA_PA_TX_EN);
+    digitalWrite(LORA_FEM_POWER, HIGH);
+    rtc_gpio_hold_en((gpio_num_t)LORA_FEM_POWER);
+    digitalWrite(LORA_FEM_CSD, HIGH);
+    rtc_gpio_hold_en((gpio_num_t)LORA_FEM_CSD);
+    gpio_pulldown_en((gpio_num_t)LORA_FEM_CPS);
```
robekl, line 582: return value is ignored from `gpio_pulldown_en((gpio_num_t)LORA_FEM_CPS)`, unlike nearby GPIO setup calls (LORA_DIO1, LORA_RESET) that log failures. Please check res and log on non-ESP_OK to avoid silent FEM sleep-state misconfiguration (potential CPS float / unstable wake behavior).

weebl2000: Thanks for the review! This was the same in the pre-existing code (`LORA_PA_TX_EN`) — I kept the pattern as-is to keep the diff focused.

I'd say it's cosmetic at best: `gpio_pulldown_en` can really only fail if the pin isn't RTC-capable, which is a static hardware/variant config issue that would surface immediately during development. And even then, CPS is already driven LOW by `SX126xInterface::sleep()` before we get here, and the FEM is shut down (CSD LOW) so CPS state is a don't care at that point.

If `gpio_pulldown_en` is returning non-OK, something weird is happening at the hardware level and I suspect users would see far more severe failures than just this pin not being held correctly.

weebl2000:
> That's more or less identical patterns for both PA's Why don't aggregate them user USE_LORA_PA or something similar and only differentiate in the sleep call to not switch off the power for one of these?
> 
> I really don't like duplicate code blocks that do almost the same.

Addressed this, please review again! @caveman99

weebl2000:
Actually, I think it's already correct, closing this!

PR #9748, opened 2026-02-25
Add Transmit history persistence for respecting traffic intervals between reboots

thebentern:
It's a fair bit of code. Additionally, I did create a way to run the native test suite (in docker) locally for those of us on macs.

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -4,15 +4,23 @@ FROM python:3.14-slim-trixie
 ENV DEBIAN_FRONTEND=noninteractive
 ENV PIP_ROOT_USER_ACTION=ignore
 
+# hadolint ignore=DL3008
 RUN apt-get update && apt-get install --no-install-recommends -y \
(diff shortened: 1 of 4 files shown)


caveman99:
Logic seems fine, i just would not persist on every change because of flash wear. How many datapoints are we expecting here?

Xaositek (approved):
Tested on RAK3401 1-watt which was having brownout issues - confirmed this fixed the DDOS of NodeInfos

thebentern:
> Logic seems fine, i just would not persist on every change because of flash wear. How many datapoints are we expecting here?

I have a 5 minute throttle on the save to FS currently.  The first save should still prevent the nodeinfo blasted out and then the node crapped itself and rebooted.

https://github.com/meshtastic/firmware/pull/9748/changes#diff-cf385cef8d6553d11ab3c97584c677bf9a423f692f62c1fb98a626b52fdc1806R63

PR #9761, opened 2026-02-26
feat: add new configuration files for LR11xx variants

NomDeTom:
@jp-bennett can you review these and tell me if the format is correct?

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


NomDeTom:
@jp-bennett can you double check my logic here?

jp-bennett (approved):

jp-bennett:
Looks correct. Proof will be testing with the real hardware, but I believe the formatting is right.

fifieldt (approved):

PR #9772, opened 2026-02-27
feat: add configurable syslog severity level filter

PhilipLykov:
## Summary

- Add configurable syslog severity level filter to reduce syslog noise
- Add `LogLevel` enum and `rsyslog_level` field (tag 12) to `NetworkConfig` protobuf
- Wire the new config field to the existing `Syslog::logMask()` filtering mechanism via a new `syslogLevelToMask()` helper
- Default behavior: `UNSET` (0) is treated as `INFO`, filtering out `DEBUG` messages

## Motivation

Currently the firmware forwards all log messages to the configured rsyslog server with no way to filter by severity. This generates significant traffic from DEBUG messages that are rarely needed in production. The `Syslog` class already has a `_priMask` filtering mechanism, but it was never wired to a user-configurable setting.

## Changes

| File | Change |
|---|---|
| `src/mesh/generated/meshtastic/config.pb.h` | Add `LogLevel` enum, `rsyslog_level` field, tags, FIELDLIST, init macros, size updates |
| `src/DebugConfiguration.h` | Add `syslogLevelToMask()` inline helper function |
| `src/mesh/eth/ethClient.cpp` | Apply `logMask()` during Ethernet syslog initialization |
| `src/mesh/wifi/WiFiAPClient.cpp` | Apply `logMask()` during WiFi syslog initialization |

## Companion PR

Protobuf definition: meshtastic/protobufs#870

## Backwards Compatibility

- Default (`UNSET` = 0) maps to `INFO` level - existing devices will filter out DEBUG after upgrade
- No breaking changes to wire format or existing config fields
- Apps that don't set `rsyslog_level` get sensible default behavior

## Test plan

- [x] Firmware builds successfully for `rak4631_eth_gw`
- [x] Verify DEBUG messages are filtered by default on rsyslog server
- [x] Set `rsyslog_level` to `LOG_DEBUG` and verify all messages appear
- [x] Set `rsyslog_level` to `LOG_WARNING` and verify only WARNING/ERROR/CRITICAL appear
- [ ] Verify WiFi-based syslog filtering works on ESP32


src/mesh/generated/meshtastic/config.pb.h
```diff
@@ -166,6 +166,26 @@ typedef enum _meshtastic_Config_NetworkConfig_AddressMode {
     meshtastic_Config_NetworkConfig_AddressMode_STATIC = 1
 } meshtastic_Config_NetworkConfig_AddressMode;
 
+/* Syslog log level to control which messages are sent to the rsyslog server.
```
caveman99: Please revert the change to the 'generated' file. This will casue ugly merge conflicts when the protos re re-generated

src/DebugConfiguration.h
```diff
@@ -199,4 +199,23 @@ class Syslog
 
 }; // namespace meshtastic
 
+inline uint8_t syslogLevelToMask(int configLevel)
+{
+    switch (configLevel) {
+    case 5:  // meshtastic_Config_NetworkConfig_LogLevel_LOG_TRACE
+    case 10: // meshtastic_Config_NetworkConfig_LogLevel_LOG_DEBUG
+        return LOG_UPTO(SYSLOG_DEBUG);
```
caveman99, line 206: Meshtastic has a trace level too.

caveman99 (changes requested):

PhilipLykov:
@caveman99 Thanks for the review! Both points addressed in ca5a0e2:

**1. Generated file reverted** — `config.pb.h` is now back to the upstream version. The proto definition lives in the companion PR meshtastic/protobufs#870. The `rsyslog_level` usage in firmware is now guarded with `#ifdef meshtastic_Config_NetworkConfig_rsyslog_level_tag`, so the code compiles cleanly with or without the proto field — defaulting to INFO when absent.

**2. TRACE level** — Made it an explicit case with a comment explaining that syslog has no native TRACE level, so it maps to `SYSLOG_DEBUG` (the most verbose syslog level available):
```cpp
case 5: // LOG_TRACE — syslog has no TRACE; map to DEBUG (most verbose syslog level)
    return LOG_UPTO(SYSLOG_DEBUG);
case 10: // LOG_DEBUG
    return LOG_UPTO(SYSLOG_DEBUG);
```

Also merged latest upstream/develop — no conflicts.

llamafilm:
The name of the protocol is 'syslog', not 'rsyslog'.  It's defined by RFC 3164 and RFC 5424.  Rsyslog is just one software implementation of said protocol.

PhilipLykov:
@llamafilm Good point — "syslog" is indeed the protocol name (RFC 3164/5424), and "rsyslog" is just one implementation. However, the existing config field is already named `rsyslog_server` in the Meshtastic protobuf, so we named `rsyslog_level` to be consistent with that established convention. Renaming would be a separate discussion affecting the existing field as well.

PR #9777, opened 2026-02-28
Unlock 0x8B5 register macro guard for SX162

thebentern:


ndoo:
Seems like the log text should have been updated to not mention the specific GC1109 FEM too @thebentern

thebentern:
> Seems like the log text should have been updated to not mention the specific GC1109 FEM too @thebentern

Good call. I'll update the log too

PR #9778, opened 2026-02-28
fix(airtime): set secSinceBoot from millis() to address inaccurate airtime calculations

m1nl:
The current airtime calculation relies on the OSThread being triggered every second. However, when the device enters sleep mode or other tasks take longer to execute, the airtime OSThread may be delayed, leading to inaccurate calculations.

I propose replacing the `secSinceBoot` timer, which is currently incremented inside the airtime OSThread, with the `millis()` function from the Arduino framework. The `millis()` function returns the number of milliseconds since the board began running the current program. According to the documentation, `millis()` provides a reliable uptime value regardless of whether the device is sleeping or temporarily busy with other tasks, making it a more robust basis for airtime calculations.

I've also decreased interval for airtime OSThread to 100ms - I consider this to be a safe value as most of maintenance tasks run with this interval. We don't need to make it equal to 1000ms, when `millis()` invocation provides accurate uptime.

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


src/airtime.cpp
```diff
@@ -97,7 +97,7 @@ uint32_t AirTime::getSecondsPerPeriod()
 
 uint32_t AirTime::getSecondsSinceBoot()
 {
-    return this->secSinceBoot;
+    return millis() / 1000;
```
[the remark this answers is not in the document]
m1nl: it's not needed as value from millis is always used to perform modulo calculations:

```
uint8_t AirTime::currentPeriodIndex()
{
    return ((getSecondsSinceBoot() / SECONDS_PER_PERIOD) % PERIODS_TO_LOG);
}

uint8_t AirTime::getPeriodUtilMinute()
{
    return (getSecondsSinceBoot() / 10) % CHANNEL_UTILIZATION_PERIODS;
}

uint8_t AirTime::getPeriodUtilHour()
{
    return (getSecondsSinceBoot() / 60) % MINUTES_IN_HOUR;
}
```

src/airtime.cpp
```diff
@@ -206,5 +204,5 @@ int32_t AirTime::runOnce()
             this->utilizationTX[utilPeriodTX] = 0;
         }
     }
-    return (1000 * 1);
+    return 100;
```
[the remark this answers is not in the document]
m1nl: most of maintenance tasks use 100ms, but I can change it if maintainers recommend value of 1000ms.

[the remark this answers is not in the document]
thebentern: Nah, this is fine. I'll mark as resolved

src/airtime.cpp
```diff
@@ -206,5 +204,5 @@ int32_t AirTime::runOnce()
             this->utilizationTX[utilPeriodTX] = 0;
         }
```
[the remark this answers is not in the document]
m1nl: I think this makes sense but it's quite unlikely that we'll sleep for more than a minute. Such condition will occur in environments with very little traffic so this won't matter much. Still this can be fixed by adding a loop to zero all periods between `lastUtilPeriodTX+1` and `utilPeriodTX`.

GUVWAF:
In general this sounds good to me. I don't really like that it's running every 100ms, though. There are not a lot of things that are by default running at a certain interval, and in principle we could get rid of it here. We only need to update airtime when receiving or transmitting a packet, and at that time we could check in which "bin" we are with `millis()`. Not sure you think this is a reasonable change, but it would be really nice to do it like this.

m1nl:
> In general this sounds good to me. I don't really like that it's running every 100ms, though. There are not a lot of things that are by default running at a certain interval, and in principle we could get rid of it here. We only need to update airtime when receiving or transmitting a packet, and at that time we could check in which "bin" we are with `millis()`. Not sure you think this is a reasonable change, but it would be really nice to do it like this.

I like this idea too! However that's way much complicated change than I proposed. If needed we can restore 1000ms interval, that should not impact airtime calculation accuracy.

I can try to redesign airtime calculation in line with idea you brought up in a separate PR.

src/airtime.cpp
robekl, line 200: Only the current bucket is cleared when lastUtilPeriod != utilPeriod, so if multiple 10s windows were skipped, intermediate buckets remain stale. This can overstate channel utilization and incorrectly block TX.

robekl, line 222: Only one utilizationTX minute bucket is reset even if multiple minute boundaries were crossed while the thread did not run.

robekl, line 52: If the thread is delayed across multiple hour periods, historical windows are not fully advanced and old values persist longer than intended.

robekl, line 20: unlikely, but calling `this->getPeriodUtilHour()` twice to calculate the index in the same array could result in different values being returned. it should be saved to a local variable or done some other way to avoid that opportunity for undesired behavior

robekl, line 31: unlikely, but calling `this->getPeriodUtilMinute()` twice to calculate the index in the same array could result in different values being returned. it should be saved to a local variable or done some other way to avoid that opportunity for undesired behavior

m1nl: @robekl this has been addressed in my last commit

m1nl: @robekl this has been addressed in my last commit

m1nl: @robekl this has been addressed in my last commit

m1nl: @robekl this has been addressed in my last commit

m1nl:
@robekl I applied changes from your comments, please review

src/airtime.cpp
fifieldt, line 100: This is what the AI is referring to. Does it make sense to give the millis treatment here too?

src/airtime.cpp
```diff
@@ -161,20 +161,22 @@ AirTime::AirTime() : concurrency::OSThread("AirTime"), airtimes({}) {}
 
 int32_t AirTime::runOnce()
 {
-    secSinceBoot++;
+    // set internal counter once per thread call
+    // to ensure period won't change during its execution
+    this->secSinceBoot = millis() / 1000;
```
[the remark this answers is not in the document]
m1nl: I think we should promote stability of secSinceBoot over its exact accuracy. The idea @robekl raised was to keep the timer value consistent throughout entire runOnce execution, which makes sense. Time drifts lesser than 1 minute won't have much impact.

src/airtime.cpp
```diff
@@ -193,18 +195,49 @@ int32_t AirTime::runOnce()
...
+            // not including previous one as it got increased by 1
+            for (int i = lastUtilPeriodTX; i <= utilPeriodTX; i++)
+                this->utilizationTX[i] = 0;
+
+            lastUtilPeriodTX = utilPeriodTX;
         }
     }
-    return (1000 * 1);
-}
+    return 1000;
```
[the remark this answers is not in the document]
m1nl: It was restored to original value as recommended by @GUVWAF

src/airtime.cpp
```diff
@@ -193,18 +195,49 @@ int32_t AirTime::runOnce()
...
+                lastUtilPeriod = 0;
+            }
 
-            this->channelUtilization[utilPeriod] = 0;
+            // clear all periods between (but not including) previous one and (including) new one
+            // not including previous one as it got increased by 1
+            for (uint8_t i = lastUtilPeriod; i <= utilPeriod; i++)
+                this->channelUtilization[i] = 0;
+
+            lastUtilPeriod = utilPeriod;
```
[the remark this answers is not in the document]
m1nl: The worst what can happen is that we'll be left with same channel utilization after the when the entire counter wraps-around. This is unlikely and fixing this will make the code even more complicated.

src/airtime.cpp
```diff
@@ -193,18 +195,49 @@ int32_t AirTime::runOnce()
...
+                // then continue from 0
+                lastUtilPeriodTX = 0;
+            }
+
+            // clear all periods between (but not including) previous one and (including) a new one
+            // not including previous one as it got increased by 1
+            for (int i = lastUtilPeriodTX; i <= utilPeriodTX; i++)
+                this->utilizationTX[i] = 0;
+
+            lastUtilPeriodTX = utilPeriodTX;
```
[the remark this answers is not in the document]
m1nl: The worst what can happen is that we'll be left with same channel utilization after the when the entire counter wraps-around. This is unlikely and fixing this will make the code even more complicated.

src/airtime.cpp
m1nl: Hm, from what I can see, Copilot seems to question almost every change just to raise concerns. Earlier, @robekl brought up a valid point that we should keep a stable value of `secSinceBoot`. Small individual drifts of less than a minute don’t matter much; at least we wouldn’t be accumulating the error, which is what’s happening right now.

The change I proposed initially seemed very small to me, but it looks like we might actually benefit from a full rewrite of the airtime calculation logic. I might try rewriting it from scratch, although I’m not sure how welcome those changes would be, since it touches some internals that haven’t been modified in a long time...

NomDeTom:
closed as https://github.com/meshtastic/firmware/pull/11291 has landed, and a different approach was taken. Many thanks and much of the credit is due to @m1nl for this contribution.

PR #9808, opened 2026-03-04
Add some lora parameter clamping logic to coalesce to defaults and enforce some bounds

thebentern:


NomDeTom:
I will look at this later, but on the face of it I am opposed to setting the defaults for lora like this. They must be defined based on regions, as some regions or presets prevent the use of defaults like these.

PR #9809, opened 2026-03-04
Add back FEM LNA mode configuration for LoRa

thebentern:
Was removed from #9753 but we just wanted the menu UI removed


caveman99 (approved):

src/mesh/NodeDB.cpp
```diff
@@ -568,6 +568,11 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
         true; // FIXME: maybe false in the future, and setting region to enable it. (unset region forces it off)
     config.lora.override_duty_cycle = false;
     config.lora.config_ok_to_mqtt = false;
+#if HAS_LORA_FEM
```
caveman99, line 571: That doesn't make sense now ;-)

caveman99 (changes requested):

PR #9813, opened 2026-03-05
More RAK6421 work

jp-bennett:
Adds the ability to specify a list of pins that need pulled high to power LoRa. Also adds the ability to read custom entries from Hat+ devices.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -672,7 +672,8 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
 #endif
     config.position.broadcast_smart_minimum_distance = 100;
     config.position.broadcast_smart_minimum_interval_secs = default_broadcast_smart_minimum_interval_secs;
-    if (config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER)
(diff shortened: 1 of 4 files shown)


bin/config.d/lora-hat-rak-6421-pi-hat.yaml
```diff
@@ -5,9 +5,11 @@ Lora:
...
+    - 12
+    - 13
   DIO3_TCXO_VOLTAGE: true
   DIO2_AS_RF_SWITCH: true
   spidev: spidev0.0
-  GPIO_DETECT_PA: 13
-  TX_GAIN_LORA: [8, 8, 11, 11, 11, 11, 11, 11, 11, 11, 11, 11, 11, 11, 11, 11, 10, 10, 9, 8, 8, 7]
\ No newline at end of file
+  # GPIO_DETECT_PA: 13
+  TX_GAIN_LORA: [8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8]
```
fifieldt: Since it's the same value you could set a single number rather than the array and it should work I think...

PR #9818, opened 2026-03-05
fix: add ROUTER_LATE and TAK_TRACKER to congestion scaling exemption

h3lix1:
## Summary

Adds `ROUTER_LATE` and `TAK_TRACKER` to the congestion scaling exemption in `getConfiguredOrDefaultMsScaled()`, so these roles skip unnecessary interval dilation just like `ROUTER` and `TRACKER`/`SENSOR`.

## Changes

| File                   | Line(s) | Before                             | After                                           | Effect                                                      |
| ---------------------- | ------- | ---------------------------------- | ----------------------------------------------- | ----------------------------------------------------------- |
| `src/mesh/Default.cpp` | 41      | `role == ROUTER` skips scaling     | `role == ROUTER \|\| role == ROUTER_LATE`       | `ROUTER_LATE` no longer has its intervals congestion-scaled |
| `src/mesh/Default.cpp` | 45      | `IS_ONE_OF(role, SENSOR, TRACKER)` | `IS_ONE_OF(role, SENSOR, TRACKER, TAK_TRACKER)` | `TAK_TRACKER` no longer has its intervals congestion-scaled |

## Rationale

Congestion scaling multiplies broadcast intervals based on online node count. Infrastructure roles (`ROUTER`, `ROUTER_LATE`) already have significantly higher base intervals and should not be further scaled — they need to maintain predictable relay timing. Similarly, `TAK_TRACKER` is a tracker variant with priority position sends; its intervals should not be dilated by congestion scaling any more than `TRACKER`'s are. The original code predates `ROUTER_LATE` and `TAK_TRACKER` and was not updated when they were added.

## 🤝 Attestations

- [X] I have tested that my proposed changes behave as described.*
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Heltec V4

* Tested on ROUTER_LATE as well as I can.. at least it doesn't break. Unable to test with TAK_TRACKER.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -672,7 +672,8 @@ void NodeDB::installDefaultConfig(bool preserveKey = false)
 #endif
     config.position.broadcast_smart_minimum_distance = 100;
     config.position.broadcast_smart_minimum_interval_secs = default_broadcast_smart_minimum_interval_secs;
-    if (config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER)
+    if (config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER &&
+        config.device.role != meshtastic_Config_DeviceConfig_Role_ROUTER_LATE)
         config.device.node_info_broadcast_secs = default_node_info_broadcast_secs;
     config.security.serial_enabled = true;
     config.security.admin_channel_enabled = false;
--- a/src/modules/CannedMessageModule.cpp
+++ b/src/modules/CannedMessageModule.cpp
@@ -1072,12 +1072,14 @@ void CannedMessageModule::sendText(NodeNum dest, ChannelIndex channel, const cha
     } else {
         sm.dest = dest;
         sm.type = MessageType::DM_TO_US;
-        // Only add as favorite if our role is NOT CLIENT_BASE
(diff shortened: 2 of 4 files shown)


src/mesh/Default.cpp
```diff
@@ -38,11 +38,13 @@ uint32_t Default::getConfiguredOrDefault(uint32_t configured, uint32_t defaultVa
 uint32_t Default::getConfiguredOrDefaultMsScaled(uint32_t configured, uint32_t defaultValue, uint32_t numOnlineNodes)
 {
     // If we are a router, we don't scale the value. It's already significantly higher.
-    if (config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER)
+    if (config.device.role == meshtastic_Config_DeviceConfig_Role_ROUTER ||
```
thebentern, line 41: @h3lix1 can we add router_late to the IF_ROUTER macro in Default? I think that makes this change in behavior make more sense, as the intent is to make Routers have course grained intervals which are fixed (predictable) values out of the box

thebentern: Never mind! I just saw that you did this in #9815

PR #9830, opened 2026-03-05
feat: add integration with berry app framework

punkle:
## Summary

Adds a runtime app framework to Meshtastic firmware, enabling community-developed apps to be written in Berry scripting language, loaded from SD card or flash at runtime. No firmware recompilation required.

This is the firmware side implementation working toward meshtastic/device-ui#282.

## Features

- app developer signing and user approval flows
- app permissions explicitly requested and user approval flows
- http client
- app state
- mesh receive on firmware, agnostic to ui framework (didnt want to implement mesh send just yet)
- UI works on bui and mui
- json parser
- users can install apps to sd card or flash depdending on what is available
- probably more I am missing

## Architecture

The implementation adds ~2,000 lines across a new `AppModule` in firmware:

| File | Purpose |
|------|---------|
| `AppModule.cpp/h` | App lifecycle, Berry VM management, permission enforcement |
| `AppDisplay.cpp/h` | Display drawing primitives bridged to Berry |
| `AppHttp.cpp/h` | HTTP client bindings |
| `AppMesh.cpp/h` | Mesh packet receive subscriptions |
| `AppState.cpp/h` | Persistent app state storage |
| `FlashFileBackend.h` | Flash filesystem backend for app storage |
| `SDFileBackend.h` | SD card filesystem backend for app storage |

The framework depends on the [mapps](https://github.com/punkle/mapps) library (pulled in via `platformio.ini`) which provides `AppLibrary`, `MappTrustStore`, and the Berry runtime integration. The device-ui side has corresponding changes in [my device-ui fork](https://github.com/punkle/device-ui) for the app browsing/approval UI screens.

## Testing

Tested on **T-Deck Plus** in both TFT and non-TFT (BUI) modes. All four sample apps run successfully on both UI frameworks.

## How to try it

1. Install the [mapps CLI](https://github.com/punkle/mapps) — used for building, signing, and uploading app bundles to devices
2. Check out the [sample apps](https://github.com/punkle/mapp-apps) for working examples:
   - Mesh messages dashboard (demonstrating mesh receive)
   - Jokes API client (demoing http request)
   - Hello world (basic demo)
   - Couter (demonstrating app state)
3. All four examples work with both BUI and MUI

## Dependencies & open questions

This PR introduces dependencies on two external repos:

- **firmware → mapps** (`punkle/mapps`) — the Berry runtime, app library, and trust store
- **device-ui → mapps** (`punkle/mapps`) — shared app types and UI integration

I am unsure about this dependency structure in general. I am used to more higher level languages where dependency management is managed on a library level so I am unsure how this type of interdependency plays out.

I'm am relatively new to C++, so I'd welcome an architectural review and any guidance on patterns, memory management, or conventions I may have gotten wrong. A high-level overview/walkthrough with a maintainer would be very helpful if this direction has legs. However i have tried to be thrifty about memory and resource management in general as I have developed this: I can answer to specifics if neccessary.

## Related

- Feature request: meshtastic/device-ui#282
- mapps CLI: https://github.com/punkle/mapps
- Sample apps: https://github.com/punkle/mapp-apps
- device-ui fork: https://github.com/punkle/device-ui

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [x] Heltec V4
  - [x] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


variants/esp32s3/t-deck/platformio.ini
```diff
@@ -21,9 +21,10 @@ build_src_filter =
   ${esp32s3_base.build_src_filter}
   +<../variants/esp32s3/t-deck>
 
-build_flags = ${esp32s3_base.build_flags} 
-  -D T_DECK 
+build_flags = ${esp32s3_base.build_flags}
+  -D T_DECK
   -D BOARD_HAS_PSRAM
+  -D HAS_SDCARD
```
punkle: Im happy to back this out if it has unintended consequences. It was useful for my testing and the t-deck does have an sd card reader.

platformio.ini
```diff
@@ -122,7 +123,7 @@ lib_deps =
 [device-ui_base]
 lib_deps =
 	# renovate: datasource=git-refs depName=meshtastic/device-ui packageName=https://github.com/meshtastic/device-ui gitBranch=master
-	https://github.com/meshtastic/device-ui/archive/6c75195e9987b7a49563232234f2f868dd343cae.zip
+	https://github.com/punkle/device-ui/archive/a4a1e3b4fbf452fca8e298baf54636ec184735d7.zip
```
punkle: This will need to be adjusted to the real device ui.

src/graphics/tftSetup.cpp
```diff
@@ -39,7 +42,7 @@ void tftSetup(void)
 #ifndef ARCH_PORTDUINO
     deviceScreen = &DeviceScreen::create();
     PacketAPI::create(PacketServer::init());
-    deviceScreen->init(new PacketClient);
+    deviceScreen->init(new PacketClient, appLibrary);
```
punkle: We can choose to back this part out to simplify the roll out of this for the moment. This will allows us to roll this out here and then later fold in the device-ui

platformio.ini
punkle: Ive decided to leave device-ui out of this to simplify.

src/graphics/tftSetup.cpp
punkle: Ive decided to leave device-ui out to simplify.

variants/esp32s3/t-deck/platformio.ini
punkle: Ive decided to back this out to simplify this review

mverch67:
Did you measure how much larger the firmware gets for devices without SD card? E.g. the T-Echo or STM32 targets?

Regarding all the compilation errors I suggest to use exclude compiler macros, e.g. EXCLUDE_BERRY_APP for targets that do not include this feature. This macro may default to 1 and is then disabled for individual targets that have large enough screens for displaying the apps.

mverch67:
I suggest to GLOBALLY exclude the app framework by default and turn it only on for tested targets otherwise you'll end up with dozens of broken devices, e.g. after line 460 in configuration.h add:

```
#ifndef MESHTASTIC_EXCLUDE_APPS
#define MESHTASTIC_EXCLUDE_APPS 1
#endif
```

and in T-DEck platformio.ini add `-DMESHTASTIC_EXCLUDE_APPS=0`
Revert the last commit regarding nrf52.ini

fifieldt:
Thanks for being a great contributor.

I'd just quickly like to echo mverch 67s comments , particularly about firmware size. It's something we are quite sensitive about :)

punkle:
Thanks for the feedback. I have made the inclusion a build variant opt-in and Ill roll out as I test on additional devices.

It occurs to be that it should be called something like "INCLUDE_MAPP_FRAMEWORK" i.e. not a negative.

punkle:
Here are the firmware size differences based on the files that are produced by running `pio run -e rak3172` and `pio run -e t-echo`. Let me know if this is what you are looking for.

| File | develop | app-framework branch | Diff |
|---|---|---|---|
| firmware-rak3172.bin | 173.4 KiB | 173.5 KiB | +128 |
| firmware-rak3172.elf | 6.9 MiB | 6.9 MiB | -352 |
| firmware-rak3172.hex | 487.8 KiB | 488.1 KiB | +360 |
| firmware-rak3172.map | 2.8 MiB | 2.8 MiB | +4557 |
| firmware-rak3172.mt.json | 737 B | 737 B | 0 |
| firmware-t-echo.elf | 1.3 MiB | 1.3 MiB | 0 |
| firmware-t-echo.hex | 2.0 MiB | 2.0 MiB | +495 |
| firmware-t-echo.mt.json | 1005 B | 1005 B | 0 |
| firmware-t-echo.uf2 | 1.4 MiB | 1.4 MiB | +512 |
| firmware-t-echo.zip | 711.8 KiB | 712.0 KiB | +177 |

PR #9870, opened 2026-03-09
fix(tlora-pager): Remove SDCARD_USE_SPI1 so SX1262 and SD can share bus

ndoo:
Problem:
- Inserting a µSD card causes RadioLib to hit a critical error and reboot
- Device enters a boot loop as the SD card remains inserted

Reproduction:
- Insert a µSD card and power on
- RadioLib reports a critical error on boot
- Device reboots, repeating indefinitely

Root cause:
- On T-Lora Pager, SX1262 and the µSD slot share the same physical SPI bus (same SCK/MOSI/MISO pins, differentiated only by CS)
- SDCARD_USE_SPI1 is intended for boards where SD is on a separate SPI bus; it initializes a second ESP32 SPI peripheral (SPI3) for SD
- SPI2 is already driving those same pins for LoRa, so both controllers simultaneously drive the same GPIO lines, causing bus contention

Fix:
- Remove SDCARD_USE_SPI1 so both devices share a single SPI peripheral (SPI2), with CS pins providing device selection as intended
- Tested on a custom fork of device-ui; LoRa and SD card map tiles both work correctly with an SD card inserted

Testing:

tlora-pager-tft is currently broken due to no mainline device-ui support. If you replace the device-ui in lib_deps with `https://github.com/Meshtastic-Malaysia/device-ui/archive/refs/heads/tlora-pager-tft-pluginbased.zip` you should be able to build a working Device UI and test with a µSD with maps on it.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - [x] TLora Pager TFT


ndoo:
This tripped me up:

SDCARD_USE_SPI1 -> HSPI → SPI3
SX1262 is on SPI → SPI2 (FSPI)

Removing SDCARD_USE_SPI1 makes SD share SPI2 (FSPI).

IMG_9449

thebentern (approved):

mverch67 (commented):
I just tried it with my T-Lora Pager and SD card. It didn't find the SD card whether SDCARD_USE_SP1 was defined or not. It seems there are further adaptations necessary.

ndoo:
Weird. Works for me.

ndoo:
Samsung 256GB, the green one from Amazon

mverch67 (approved):
Looking at the SPI GPIO definitions SD card and radio are using the same GPIO, so they supposed to share indeed the same SPI.

ndoo:
Sorry, accidentally pushed some testing commits, resetting and force push…

PR #9874, opened 2026-03-10
Experiment: C++17 support

thebentern:
I'm not in love with all of the unflags, but it seems to be the main way of making some of the platforms work.

### DIFF
--- a/variants/native/portduino.ini
+++ b/variants/native/portduino.ini
@@ -55,7 +55,7 @@ build_flags =
   -li2c
   -luv
   -std=gnu17
-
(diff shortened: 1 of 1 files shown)


variants/native/portduino.ini
```diff
@@ -55,7 +55,6 @@ build_flags =
   -li2c
   -luv
   -std=gnu17
```
Jorropo, line 57: why doesn't it match `esp32-common.ini` ?

NomDeTom:
Once this is released, do we need an update to the LLM instructions?

PR #9895, opened 2026-03-12
fix(native): implement BinarySemaphorePosix with proper pthread synchronization

iannucci:
## Summary

The `BinarySemaphorePosix` class, used on all Linux/portduino/native builds, has stub implementations that completely break the cooperative thread scheduler:

- `give()` is a **no-op** (empty function body)
- `take(uint32_t msec)` just calls `delay(msec)` and **always returns false**
- `giveFromISR()` is a **no-op**

The `// FIXME` comment in `take()` confirms this was known to be incomplete.

This means the `mainDelay` semaphore (and any other `BinarySemaphorePosix` usage) cannot function: threads cannot wake the main loop early via `give()`, the `NotifiedWorkerThread` pattern is broken (`notify()` → `mainDelay.interrupt()` → `give()` does nothing), and the scheduler degrades to a fixed-interval poller that always sleeps the full timeout.

### What this PR does

Replaces the stubs with a proper binary semaphore implementation using POSIX threads:

**Header (`BinarySemaphorePosix.h`):**
- Adds `#include <pthread.h>` (guarded by `#ifndef HAS_FREE_RTOS`)
- Replaces commented-out `SemaphoreHandle_t` with `pthread_mutex_t mutex`, `pthread_cond_t cond`, `bool signaled`

**Implementation (`BinarySemaphorePosix.cpp`):**
- **Constructor:** `pthread_mutex_init` + `pthread_cond_init`, `signaled = false`
- **Destructor:** `pthread_cond_destroy` + `pthread_mutex_destroy`
- **`take(msec)`:** `pthread_cond_timedwait` with `CLOCK_REALTIME` timeout. Uses a `while (!signaled)` loop to handle spurious wakeups. Atomically consumes the signal (binary semaphore semantics). Returns `true` if signaled, `false` on timeout.
- **`give()`:** Sets `signaled = true`, calls `pthread_cond_signal` under mutex
- **`giveFromISR()`:** Delegates to `give()`, sets `*pxHigherPriorityTaskWoken = true`

### Symptoms observed without this fix

Tested on Raspberry Pi 3 Model B (ARM64, Debian 12 Bookworm) with Adafruit LoRa Radio Bonnet (RFM9x/SX1276):

- Radio TX works only on fixed timer intervals (no early wake from `give()`)
- Radio RX interrupts are not propagated to the main loop
- Telemetry packets are sent to the phone API but never transmitted over the mesh
- MQTT gateway receives LoRa packets but relay timing is degraded
- OLED display (SSD1306 via I2C) was non-functional (likely related to I2C thread scheduling)

After the fix: bidirectional LoRa, MQTT gateway relay, telemetry over mesh, and OLED display all work correctly.

### Scope

This change only affects `#ifndef HAS_FREE_RTOS` builds (Linux/portduino/native). FreeRTOS builds use `BinarySemaphoreFreeRTOS` and are unaffected.

## Test plan

- [x] Built with `bash ./bin/build-native.sh native` (Docker cross-compilation, `python:3.11-slim-bookworm`, `--platform linux/arm64`)
- [x] Deployed to Raspberry Pi 3 Model B running Debian 12 (Bookworm) ARM64
- [x] Verified RF95/SX1276 radio init success
- [x] Verified bidirectional LoRa communication (TX and RX confirmed with multiple nodes)
- [x] Verified MQTT gateway operation (uplink and downlink)
- [x] Verified telemetry transmission over mesh
- [x] Verified SSD1306 OLED display working
- [x] Verified stable operation across power cycles
- [ ] Needs testing on other native platforms (x86_64 Linux, macOS simulator)


GUVWAF:
Very nice work. I believe this might also solve the issue mentioned in #9939, or at least make it *way* less common, and it is a better solution IMO.

jp-bennett:
It needs to be specific for portduino rather than just any target without HAS_FREE_RTOS, but yes, this looks good.

jp-bennett:
Testing this out with my DMShell branch, and it appears to fix the RX detection issue I was seeing. Which in retrospect is exactly what the PR said it did.

jp-bennett (approved):

PR #9897, opened 2026-03-13
Fix rak_wismeshtag low‑voltage reboot hang after App configuration

Ethan-chen1234-zy:
Improve low‑voltage handling for wisMeshtag TAG to avoid boot hang after App config

Summary
This PR fixes the wisMeshtag (rak_wismeshtag) “hang after App config reboot” issue by making the device proactively avoid running in the POF/VDDH‑sensitive low‑voltage region. Concretely, it:

Raises the software safe VDD threshold and hysteresis.
Adds automatic System OFF when VDD stays unsafe for a period (no USB case).
Uses LPCOMP wake so the device recovers when voltage returns to a safe range.
The low‑VDD System OFF logic is implemented via a variant hook (variant_nrf52LoopHook) so the platform code remains generic.

In addition, this PR updates the OCV curve used for battery SOC/telemetry display, improving the accuracy of the App’s battery percentage.
OCV changes only affect battery SOC estimation/telemetry display; they do not change the boot safety gate (waitUntilPowerLevelSafe) or POF behavior.

Background / Why this matters (what was actually failing)
Root symptom (reported/observed)
On v2.7.3, when the battery is around ~2.8 V or below, performing App configuration followed by reboot can leave the device completely stuck. Recovery typically requires a true cold start (full power removal / draining).

Why v2.7.3 hangs (clearer root-cause story)
On VDDH‑mode hardware, Nordic POF has two independent concepts:

THRESHOLD (VDD comparator)
THRESHOLDVDDH (VDDH‑only high‑voltage mode)
From main-nrf52.cpp comments and the dedicated POF/VDDH analysis doc:

The firmware history/configuration focused on VDD threshold, while the hardware POF is in VDDH mode.
As a result, POF behavior effectively shows up near ~2.8 V, even when the configured value does not appear to match the expected VDDH threshold.
When POF triggers in this low‑voltage “grey zone” while the App is saving config:

App config changes → Flash/LittleFS writes during the reboot sequence.
Low voltage + POF causes write failure / corruption risk.
LittleFS may hit lfs_assert() → sets a “corrupt/format-needed” path or forces reboot.
After reboot, the device is still in the same low‑voltage band → it can re-enter the reboot / retry / blocked chain until the battery is fully drained (true cold start).
This explains why the issue was much worse on v2.7.3 (directly tied to the POF/Flash/LittleFS failure chain), and why soft reset alone was not sufficient.

What this PR changes (the actual fix strategy)
This PR does not attempt to “debug or reconfigure POF VDDH comparator” in the current code path. Instead, it prevents the problematic operating window by policy.

1) Treat low VDD as unsafe (software guardrail)
In variants/nrf52840/rak_wismeshtag/variant.h:

SAFE_VDD_VOLTAGE_THRESHOLD = 2.9
SAFE_VDD_VOLTAGE_THRESHOLD_HYST = 0.1
Result: we avoid performing “heavy operations” (BLE + Flash writes + reboot path) in the range where the v2.7.3 POF/VDDH + LittleFS failure chain is most likely.

2) Auto System OFF when VDD stays unsafe (runtime mitigation)
Using the new variant hook:

When !powerHAL_isVBUSConnected() and !powerHAL_isPowerLevelSafe() persists longer than LOW_VDD_SYSTEMOFF_DELAY_MS (5000 ms),
call cpuDeepSleep(portMAX_DELAY) → enter System OFF.
This ensures the device stops “staying alive” in the POF-sensitive grey band and avoids repeated failure attempts.

3) LPCOMP wake on voltage recovery
In variant.h:

BATTERY_LPCOMP_INPUT = NRF_LPCOMP_INPUT_3
BATTERY_LPCOMP_THRESHOLD = NRF_LPCOMP_REF_SUPPLY_9_16
So from System OFF, when voltage rises to a safe region (≈3.1 V at the divider), LPCOMP wakes the device and recovery can happen without fully draining the battery.

4) Variant hook integration (code organization)
In src/platform/nrf52/main-nrf52.cpp:

Add weak variant_nrf52LoopHook()
Call it at the end of nrf52Loop()
In rak_wismeshtag/variant.cpp:

Implement the low‑VDD System OFF logic inside that hook.
Scope / Known limitations (explicit)
POF VDDH comparator configuration is not explicitly added/verified in this PR.
This PR focuses on actively avoiding the unsafe region rather than validating VDDH comparator programming.
Separate hang can still occur at 3.0–3.4 V after App-triggered soft reset (v2.7.20+ class issue) due to SAADC under‑reporting after soft reset, causing waitUntilPowerLevelSafe() to block. This PR does not address that SAADC/soft-reset issue (documented as out-of-scope).
Behavior after this change
VDD ≤ 2.9 V
Device does not proceed normally; it enters System OFF after the configured delay (no USB).
Low-voltage indication remains as designed (“3 green blinks every 2 s” behavior during the unsafe period).
3.0 V ≤ VDD ≤ 3.4 V
Cold boot behavior should be OK, but App-triggered soft reset may still stop at the power-safe gate due to the known SAADC under-reporting issue (manual reset may be needed).
VDD ≥ 3.5 V
App config + reboot behaves normally (no hang observed).
Testing (TAG hardware)
Below ~2.9 V: enters System OFF as intended; recovery only after charging + reset.
~3.1 V: auto wake from System OFF when voltage recovers.
3.0–3.4 V: cold boot passes; App config reboot may require manual reset (SAADC known issue).
≥3.5 V: App config reboot works normally.

### DIFF
--- a/src/platform/nrf52/Nrf52SaadcLock.cpp
+++ b/src/platform/nrf52/Nrf52SaadcLock.cpp
@@ -1,6 +1,6 @@
-#include "configuration.h"
 #include "Nrf52SaadcLock.h"
 #include "concurrency/Lock.h"
+#include "configuration.h"
 
 #ifdef ARCH_NRF52
 
--- a/src/platform/nrf52/main-nrf52.cpp
+++ b/src/platform/nrf52/main-nrf52.cpp
@@ -341,7 +341,7 @@ void nrf52Loop()
     checkSDEvents();
     reportLittleFSCorruptionOnce();
 
-    variant_nrf52LoopHook();  // Optional variant hook called each nrf52Loop();
+    variant_nrf52LoopHook(); // Optional variant hook called each nrf52Loop();
 }
 
 #ifdef USE_SEMIHOSTING
--- a/variants/nrf52840/rak_wismeshtag/variant.cpp
+++ b/variants/nrf52840/rak_wismeshtag/variant.cpp
@@ -19,13 +19,13 @@
 */
 
 #include "variant.h"
+#include "Arduino.h"
+#include "FreeRTOS.h"
 #include "nrf.h"
-#include "wiring_constants.h"
-#include "wiring_digital.h"
 #include "power/PowerHAL.h"
 #include "sleep.h"
-#include "FreeRTOS.h"
-#include "Arduino.h"
+#include "wiring_constants.h"
+#include "wiring_digital.h"
 
 const uint32_t g_ADigitalPinMap[] = {
     // P0
--- a/variants/nrf52840/rak_wismeshtag/variant.h
+++ b/variants/nrf52840/rak_wismeshtag/variant.h
@@ -225,7 +225,7 @@ SO GPIO 39/TXEN MAY NOT BE DEFINED FOR SUCCESSFUL OPERATION OF THE SX1262 - TG
 #define AREF_VOLTAGE 3.0
 #define VBAT_AR_INTERNAL AR_INTERNAL_3_0
 #define ADC_MULTIPLIER 1.73
-#define OCV_ARRAY 4160, 4020, 3940, 3870, 3810, 3760, 3740, 3720, 3680, 3620, 2990  // updated OCV array for rak_wismeshtag
+#define OCV_ARRAY 4160, 4020, 3940, 3870, 3810, 3760, 3740, 3720, 3680, 3620, 2990 // updated OCV array for rak_wismeshtag
 
 // Wake from System OFF when battery rises again (LPCOMP).
 // BAT_ADC divider: R22=1M (top), R24=1.5M (bottom) => V_BAT_ADC = VBAT * (1.5 / (1.0 + 1.5)) = 0.6 * VBAT


Ethan-chen1234-zy:
Addressed review: fixed low-VDD timer when millis()==0, rate-limited powerHAL_isPowerLevelSafe to 100ms in the variant hook, clarified LPCOMP 9/16 vs divider math, added integer mV threshold macros + platform support, removed trailing whitespace.

Ethan-chen1234-zy:
We added a shared mutex (concurrency::nrf52SaadcLock from src/platform/nrf52/Nrf52SaadcLock.cpp) and take it in getVDDVoltage() (covers analogReadResolution / analogReference / analogReadVDD) and in the nRF52 analog battery read path in Power.cpp, so VDD and battery reads cannot interleave SAADC configuration.

Ethan-chen1234-zy:
This change only adjusts the WisMeshTag LPCOMP wake threshold:
BATTERY_LPCOMP_THRESHOLD is updated from REF_SUPPLY_9_16 to REF_SUPPLY_5_8.
Reason: cpuDeepSleep() configures LPCOMP for DETECT_UP (wake on rising crossing).
With the 0.6× battery divider, REF_SUPPLY_9_16 can result in the input already
being above the threshold when VDD is drooping/tracking VBAT, so no rising-edge
crossing occurs and the device may stay in System OFF. REF_SUPPLY_5_8 improves
wake reliability on voltage recovery.

PR #9903, opened 2026-03-14
feat: Support INA219/INA226 as primary battery sensor without ADC pin

tobiascrackz:
## Problem
On DIY boards with an INA219 directly on the battery, there was no 
initialization path for battery level sensing without a BATTERY_PIN.
batteryLevel remained NULL, causing PWD (power unknown) in the node list.

## Solution
- Added `inaInit()` to Power::setup() as a new init path
- Added `BATTERY_INA_ADDRESS` fallback in `hasINA()` so the sensor 
  works without config being set via app/CLI
- Added `BATTERY_INA_ADDRESS` define support in `variant.h` for 
  boards that have an INA sensor hardwired to the battery

## Testing
Tested on custom ESP32 board with INA219 at 0x40 directly on battery+.
Battery voltage and SOC now displayed correctly in node list.

## 🤝 Attestations

-  I have tested that my proposed changes behave as described.
- I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - Other (please specify below):
     DIY ESP32 & WROVER32 Modules wired to RFM95


korbinianbauer:
What if the INA is not wired up to measure the nodes bat voltage though?

NomDeTom:
> What if the INA is not wired up to measure the nodes bat voltage though?

Having had a brief look, it appears that the default ina address is added at compile time in the variant.h, and this is only engaged if either the default address is used, or a new address is supplied.

tobiascrackz:
The change only affects boards that explicitly define BATTERY_INA_ADDRESS in their variant.h, meaning the board designer has intentionally hardwired an INA sensor for battery monitoring. In that case it is assumed the sensor is physically present.

madeofstown:
I tested this out for my own variant and it worked great. Thank you for this submission, I hope it gets merged in soon.

NomDeTom:
@tobiascrackz are you still interested in merging this?

tobiascrackz:
> @tobiascrackz are you still interested in merging this?

yes! i think this would be a legit improvement

Simplycissmus:
A second board that needs this init path.

`seeed-xiao-s3` carries `BATTERY_PIN -1` next to `ADC_CHANNEL ADC_CHANNEL_0`, added in
#6070 so an INA could report without claiming a GPIO. But `ADC1_CHANNEL_0_GPIO_NUM` is 1,
and the same variant gives GPIO1 to `PIN_GPS_STANDBY`, so a board without an INA samples
the GPS standby line instead: measured on a bare unit, 3084 mV at 0 percent for 81 s after
every boot; on a unit with the L76K expansion board, 5338 to 5838 mV unbroken over 11 days.
Neither has a battery. Written up in #11908.

The obvious fix, dropping the three defines, takes the INA with it — `analogInit()` returns
false without `BATTERY_PIN`, so `batteryLevel` is never set. This PR is what would let that
variant have both.

tobiascrackz:
> A second board that needs this init path.
> 
> `seeed-xiao-s3` carries `BATTERY_PIN -1` next to `ADC_CHANNEL ADC_CHANNEL_0`, added in #6070 so an INA could report without claiming a GPIO. But `ADC1_CHANNEL_0_GPIO_NUM` is 1, and the same variant gives GPIO1 to `PIN_GPS_STANDBY`, so a board without an INA samples the GPS standby line instead: measured on a bare unit, 3084 mV at 0 percent for 81 s after every boot; on a unit with the L76K expansion board, 5338 to 5838 mV unbroken over 11 days. Neither has a battery. Written up in #11908.
> 
> The obvious fix, dropping the three defines, takes the INA with it — `analogInit()` returns false without `BATTERY_PIN`, so `batteryLevel` is never set. This PR is what would let that variant have both.

great to know this PR would also help the seeed-xiao-s3 case.

PR #9906, opened 2026-03-14
Heltec v4.3: enable LNA by default

weebl2000:
It should only be disabled by users that have problems with it. Anyone using Heltec v4.3 and not setting it explicitly will have bad RX.

Get firmware here:

- [UPDATE BIN](https://x230.weebl.me/public/firmware-heltec-v4-2.7.21.ce37a6e.bin)
- [FACTORY ERASE BIN](https://x230.weebl.me/public/firmware-heltec-v4-2.7.21.ce37a6e.factory.bin)

### DIFF
--- a/src/mesh/LoRaFEMInterface.cpp
+++ b/src/mesh/LoRaFEMInterface.cpp
@@ -25,7 +25,7 @@ void LoRaFEMInterface::init(void)
         pinMode(LORA_KCT8103L_PA_CSD, OUTPUT);
         digitalWrite(LORA_KCT8103L_PA_CSD, HIGH);
         pinMode(LORA_KCT8103L_PA_CTX, OUTPUT);
-        digitalWrite(LORA_KCT8103L_PA_CTX, HIGH);
+        digitalWrite(LORA_KCT8103L_PA_CTX, LOW); // LNA enabled by default
         setLnaCanControl(true);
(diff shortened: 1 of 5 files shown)


fifieldt:
Thanks! Can we get a few v4 users to share their results? Code seems fine, just feels like we're flipping back and forth a little on this device.

weebl2000:
> Thanks! Can we get a few v4 users to share their results? Code seems fine, just feels like we're flipping back and forth a little on this device.

Fairly certain ON is the way to go by default. FYI, Heltec v4.2 and earlier didn't have a way to turn off the low noise amplifier at all. It was always physically in the RX path.

Heltec v4.3 came out with the main purpose being for users to able to bypass the LNA without doing a hardware bypass mod. Realistically you'd only want to disable it if you have problematic RF interference.

iuvi7:
I'm Alien here, but maybe it can be an option to choose in web flasher for v4 like "Choose version you have 4.1\4.2 or 4.3" ?

weebl2000:
> I'm Alien here, but maybe it can be an option to choose in web flasher for v4 like "Choose version you have 4.1\4.2 or 4.3" ?

Not needed. All versions are supported by the firmware. 4.2 and earlier didn't have an LNA you could toggle on/off with software so the configuration does not apply there. It's only relevant to 4.3+

PR #9916, opened 2026-03-16
Fix for preserving pki_encrypted and public_key when relaying UDP multicast packets to radio.

niklaswall:
Fix for preserving pki_encrypted and public_key when relaying UDP multicast packets to radio.

PKI DMs sent over UDP multicast had their pki_encrypted flag and public_key fields explicitly cleared before being forwarded to the LoRa radio. This caused the receiving node to treat the packet as a channel-encrypted message it couldn't decrypt, silently dropping it.

The MQTT ingress path correctly preserves these fields. The UDP multicast ingress path should behave the same way.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [x] Heltec Wireless Tracker V2


thebentern:
@jp-bennett tagging you since this was added in a97df4bb524d0f53276a0662e286b8b385a625b1

niklaswall:
Yes, it would be interesting to know why it was added! The commit message "Sanity check incoming UDP" does not really make sense since we were basically just "dropping" parts of the MeshPacket related to PKI encryption (pki_encrypted and public_key).

My best guess is that you might sometimes end up with garbage in the MeshPacket since memory was not initialized before decoding the protobuf.

Initializing MeshPacket to meshtastic_MeshPacket_init_zero before decoding the received data should fix that potential issue. 
`meshtastic_MeshPacket mp = meshtastic_MeshPacket_init_zero;`

Issue #9704, opened 2026-02-21
[Bug]: STM32/E5 settings not saved


einolto:
### Category

Hardware Compatibility

### Hardware

Other

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.19

### Description

Variant: STM32/E5
The last working firmware I have tried is 2.7.5 (edit: 2.7.17)
2.7.19 (edit: and 2.7.18) runs, but settings do not stick. After a few seconds default values are restored. Full erase done on each try.
For testing I am fine with 2.7.5 (edit: now 2.7.17) but since I have a testing setup I am able to test whatever needed.

### Relevant log output

```Shell

```

einolto:
Edit: According to my testing, the last working firmware in terms of the issue is v2.7.17.9058cce.
Repeatable procedure:
* change region
* device reboots (I guess this is intentional due to region change)
* settings reverted back to default

Edit after investigation. Looks like restoring this:

---
    { // Scope for lock
        concurrency::LockGuard g(spiLock);
        // brief window of risk here ;-)
        if (fullAtomic && FSCom.exists(filename.c_str()) && !FSCom.remove(filename.c_str())) {
            LOG_ERROR("Can't remove old pref file");
            return false;
        }
    }
---
to a file: "SafeFile.cpp" restores the saving functionality 2.7.18 onwards.

According to change log, the modification was done: #9296

PR #5987, opened 2025-02-04
try implementing littlefs for stm32

caveman99:
i can't find my wio-e5 to test this. Compiles without error though :-)

PR #9927, opened 2026-03-17
fix(STM32WL) fix LittleFS FILE_O_WRITE truncation for persistent prefs

dunaevai135:
# Fix persisted config corruption on CDEBYTE_E77-MBL. 

**Likely fixes** #9704 see Testing

STM32WL LittleFS reopened `FILE_O_WRITE` without truncation, which could leave stale bytes in rewritten protobuf files. After reboot this broke prefs loading and the node fell back to defaults.

## Fix
use `LFS_O_TRUNC` LittleFS flags for STM32WL `FILE_O_WRITE`

Only the STM32WL LittleFS backend is changed.

According to @caveman99 's littlefs implementation #5987 and @vidplace7 's report, it worked for them without this fix. I'd love to hear your feedback.

With the fix, I've got a fully functional variant for serial from E77-900m22s.

## Reproduction:
```
  meshtastic --port /dev/ttyUSB0 --set device.led_heartbeat_disabled true --timeout 20
  meshtastic --port /dev/ttyUSB0 --set display.flip_screen true --timeout 20
  meshtastic --port /dev/ttyUSB0 --set lora.hop_limit 7 --timeout 20
  meshtastic --port /dev/ttyUSB0 --reboot --wait-to-disconnect 3 --timeout 20
  sleep 12
  meshtastic --port /dev/ttyUSB0 --get device.led_heartbeat_disabled --timeout 20
  meshtastic --port /dev/ttyUSB0 --get display.flip_screen --timeout 20
  meshtastic --port /dev/ttyUSB0 --get lora.hop_limit --timeout 20
```

## Testing
I have tested only on `CDEBYTE_E77-MBL` and `heltec-v3` (for regressions). Since this only affected the STM32WL LittleFS backend, it would be a good idea to test it on other STM32WL variants, such as the wio-e5, but I don't have any of those. I'd appreciate some help with this.

  - [x] Heltec (Lora32) V3
  - [x] CDEBYTE_E77-MBL

### DIFF
--- a/src/platform/stm32wl/STM32_LittleFS_File.cpp
+++ b/src/platform/stm32wl/STM32_LittleFS_File.cpp
@@ -54,7 +54,7 @@ File::File(char const *filename, uint8_t mode, STM32_LittleFS &fs) : File(fs)
 
 bool File::_open_file(char const *filepath, uint8_t mode)
 {
-    int flags = (mode == FILE_O_READ) ? LFS_O_RDONLY : (mode == FILE_O_WRITE) ? (LFS_O_RDWR | LFS_O_CREAT) : 0;
+    int flags = (mode == FILE_O_READ) ? LFS_O_RDONLY : (mode == FILE_O_WRITE) ? (LFS_O_RDWR | LFS_O_CREAT | LFS_O_TRUNC) : 0;
 
     if (flags) {
         _file = (lfs_file_t *)rtos_malloc(sizeof(lfs_file_t));
@@ -71,10 +71,6 @@ bool File::_open_file(char const *filepath, uint8_t mode)
             _file = NULL;
             return false;
         }
-
-        // move to end of file
-        if (mode == FILE_O_WRITE)
-            lfs_file_seek(_fs->_getFS(), _file, 0, LFS_SEEK_END);
 
         _is_dir = false;
     }


Stary2001:
Thanks for the reproducer! I can see it on wio-e5:
```
INFO  | ??:??:?? 0 Load /prefs/config.proto
ERROR | ??:??:?? 0 Error: can't decode protobuf array overflow
INFO  | ??:??:?? 0 Install default LocalConfig
```

It's unclear why this doesn't trigger on nrf52 as the arduino/littlefs glue is afaik taken from there...
(edit: It's because SafeFile doesn't do renaming on NRF52. Argh.)

Stary2001:
There's another way to fix this - rather than modifying that to truncate all files on open (which changes behaviour from what's expected), just don't use the broken implementation of `renameFile` and use the one built into LittleFS instead, which definitely works...
I need to sit down and analyze how many flash pages each approach uses, or take the flash cache out of arduino for nrf52 which would allow lowering the littlefs block size to 512 so we don't have to quite worry so much about it.
We reserve 24k, which is 12 2k pages, and we have 5 config files that are all ~100 bytes.

```
diff --git a/src/FSCommon.cpp b/src/FSCommon.cpp
index f215be80f..906c10ba5 100644
--- a/src/FSCommon.cpp
+++ b/src/FSCommon.cpp
@@ -80,7 +80,7 @@ bool renameFile(const char *pathFrom, const char *pathTo)
 {
 #ifdef FSCom
 
-#ifdef ARCH_ESP32
+#if defined(ARCH_ESP32) || defined(ARCH_STM32WL)
     // take SPI Lock
     spiLock->lock();
     // rename was fixed for ESP32 IDF LittleFS in April
     ```

ndoo:
Question: could the STM32WL Milesight bathroom sensor I assumed had dead flash actually just be a manifestation of this bug? What's the symptom of this issue?

einolto:
Does not work on STM32WL (wio-e5 module). Procedure:
1. Flash full erase
2. Flashing "firmware-wio-e5-le-2.7.21.efd2613.hex"
3. "meshtastic --info" works
4. "meshtastic --set lora.region EU_868" -> hangs on writing to device
5. device reset
6. "meshtastic --info does not work
7. RAW debug console prints Meshtastic+version information on device reset (<- device is not completely bricked)

Full output of the step 4:
...
meshtastic --port com3 --set lora.region EU_868
Connected to radio
Set lora.region to EU_868
Writing lora configuration to device
...

Stary2001:
> wio-e5: region `FLASH' overflowed by 296 bytes

really 😭

Stary2001:
Should have been fixed by PR #10171

thebentern:
Inadvertently reopened

Issue #9876, opened 2026-03-10
[Bug]: MQTT config does not persist reliably on RAK4631 RAK13300 Wifi and Ethernet gateway devices running 2.7.19; proxy/UI/API behavior is also unstable


benseidler01:
### Category

WiFi

### Hardware

Rak4631

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.19.bb3d6d5  Rak4631_eth_gw

### Description

````markdown id="hqoict"
# MQTT configuration does not persist on RAK gateway devices running Meshtastic 2.7.19, with repeated API/UI instability and inconsistent proxy behavior

> I fed ChatGPT a mountain of serial logs and terminal history because, after hours of wrestling with this, my original notes had evolved from “technical observations” into an enormous pile of notes full of explicit adjectives and the thought of sanitizing the documentation by hand was more than I could handle today. The result below is the cleaned-up version of the same evidence, minus the sleep deprivation and creative adjectives.

## Summary

MQTT behavior on Meshtastic `2.7.19.bb3d6d5` appears broken on multiple new RAK ethernet and wifi gateway devices.  My documented testing over the last 12 hours was on the RAK4631 with the RAK13800 ethernet module, but I experienced the same exact behavior over the last two weeks with the RAK WisMesh WiFi Gateways.

Across four new devices total:
- 2 × RAK WisMesh WiFi Gateway devices
- 2 × RAK4631 devices with RAK13800 Ethernet modules

...a basic MQTT gateway configuration resulted in unstable or non-persistent behavior. The detailed evidence below is from a RAK4631 Ethernet gateway device running:

- `firmware-rak4631_eth_gw-2.7.19.bb3d6d5.uf2`
- `pioEnv: rak4631_eth_gw`

In addition to the Ethernet gateway testing, I also tested a Wi-Fi gateway-class device running the same Meshtastic release branch on:

- `pioEnv: rak11200`
- firmware version `2.7.19.bb3d6d5`

The most critical defect is that `mqtt.enabled` does not persist. The CLI accepts the write, reports success, and says it wrote the MQTT configuration to the device, but read-back immediately returns `False`, including after reboot.

At the same time, the device also exhibited:
- inconsistent `mqtt.proxy_to_client_enabled` persistence
[...]

PR #9934, opened 2026-03-18
fix: MQTT settings silently fail to persist when broker is unreachable

### LINKED ISSUE #9107: [Bug]: MQTT cannot be configured
### Category

Other

### Hardware

Heltec V3

- [ ] InkHUD ePaper

### Firmware Version

2.7.15.567b8ea

### Description

The node is accepting MQTT configurations without any issues/errors etc. but the settings remain unchanged.

### Relevant log output

```Shell
meshtastic --set mqtt.address "<own mqtt>" --set mqtt.username "<user>" --set 
mqtt.password "<password>"

Connected to radio
Set mqtt.address to <own mqtt>
Set mqtt.username to <user>
Set mqtt.password to <password>
Writing mqtt configuration to device

meshtastic --get mqtt.address
Connected to radio
mqtt.address: mqtt.meshtastic.org
Completed getting preferences
```

rcatal01:
## Summary

- `isValidConfig()` was testing broker connectivity via `connectPubSub()` during config validation
- When the broker was unreachable (network not ready, DNS failure, server down), the function returned `false`, causing `AdminModule::handleSetModuleConfig()` to skip saving settings entirely — silently
- This removes the connectivity test from `isValidConfig()`, which now only validates configuration correctness (TLS support check, default server port check)
- Connectivity is handled by the MQTT module's existing reconnect loop after settings are saved

Fixes #9107

## Root Cause

In `MQTT::isValidConfig()` (MQTT.cpp line 668-670), `connectPubSub()` was called to test broker connectivity as part of validation. When this failed, `isValidConfig()` returned `false`, which caused `AdminModule` (line 924-926) to `return false` before reaching line 930 where `moduleConfig.mqtt = c.payload_variant.mqtt` actually persists the settings.

The result: settings appeared to save but silently reverted. No error was shown to the user.

## Changes

- **`src/mqtt/MQTT.cpp`**: Removed `connectPubSub()` call and associated `MQTTClient`/`PubSubClient` object creation from `isValidConfig()`. The function now only validates config correctness, not connectivity.
- **`test/test_mqtt/MQTT.cpp`**: Updated 4 unit tests to reflect that `isValidConfig()` no longer tests connectivity. `test_configWithConnectionFailure` renamed to `test_configWithUnreachableServerIsStillValid` and now asserts `true`.

## Test plan

- [x] Built firmware for Heltec Wireless Tracker (ESP32-S3)
- [x] Changed MQTT settings via Meshtastic iOS app with no WiFi configured (broker unreachable)
- [x] Device rebooted after save
- [x] Reconnected via BLE — settings persisted correctly
- [x] Verified this was the exact scenario that failed on stock 2.7.15 firmware
- [ ] Unit tests updated and should pass (removed connectivity assertions from validation tests)

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -657,9 +657,29 @@ bool MQTT::isValidConfig(const meshtastic_ModuleConfig_MQTTConfig &config, MQTTC
             return false;
 #endif
         }
-        // Connectivity is intentionally NOT validated here. Settings must be saved even
-        // when the network is temporarily unavailable. The MQTT module's existing reconnect
-        // loop will establish the connection after settings are persisted and the device reboots.
+        // Perform a lightweight TCP connectivity check without using connectPubSub(),
+        // which mutates the module's isConnected state. This only checks if the server
+        // is reachable — it does not establish an MQTT session.
+        // Settings are always saved regardless of the result.
+        if (isConnectedToNetwork()) {
+            MQTTClient testClient;
+            if (!testClient.connect(parsed.serverAddr.c_str(), parsed.serverPort)) {
+                const char *warning = "Could not reach the MQTT server. Settings will be saved, but please verify the server "
+                                      "address and credentials.";
+                LOG_WARN(warning);
+#if !IS_RUNNING_TESTS
+                meshtastic_ClientNotification *cn = clientNotificationPool.allocZeroed();
+                if (cn) {
+                    cn->level = meshtastic_LogRecord_Level_WARNING;
+                    cn->time = getValidTime(RTCQualityFromNet);
+                    strncpy(cn->message, warning, sizeof(cn->message) - 1);
+                    cn->message[sizeof(cn->message) - 1] = '\0';
+                    service->sendClientNotification(cn);
+                }
+#endif
+            }
+            testClient.stop();
+        }
 #else
         const char *warning = "Invalid MQTT config: proxy_to_client_enabled must be enabled on nodes that do not have a network";
         LOG_ERROR(warning);
--- a/test/test_mqtt/MQTT.cpp
+++ b/test/test_mqtt/MQTT.cpp
@@ -818,7 +818,8 @@ void test_configEmptyIsValid(void)
     TEST_ASSERT_TRUE(MQTT::isValidConfig(config));
 }
 
-// Empty 'enabled' configuration is valid. No connectivity check is performed.
+// Empty 'enabled' configuration is valid. A lightweight TCP check may be performed
+// but does not affect the result.
 void test_configEnabledEmptyIsValid(void)
 {
     meshtastic_ModuleConfig_MQTTConfig config = {.enabled = true};
@@ -842,7 +843,7 @@ void test_configWithDefaultServerAndInvalidPort(void)
     TEST_ASSERT_FALSE(MQTT::isValidConfig(config));
 }
 
-// Custom host and port is valid. No connectivity check is performed.
+// Custom host and port is valid. TCP reachability is checked but does not block saving.
 void test_configCustomHostAndPort(void)
 {
     meshtastic_ModuleConfig_MQTTConfig config = {.enabled = true, .address = "server:1234"};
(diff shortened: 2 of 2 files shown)


Xaositek:
This may have a positive impact on #9876

thebentern:
@rcatal01 I think I'm in favor of this change because its currently confusing users, but I'm not sure silently failing a different way is better. I think perhaps we should throw a client notification up to the client letting them know that we could not reach the configured MQTT broker and to double check the settings.

rcatal01:
@thebentern Good point — silently succeeding isn't much better than silently failing. I've pushed an update that:

- **Still always saves settings** (the core fix)
- **Attempts a connectivity check** when the network is available
- If the broker is unreachable, sends a **WARNING client notification**: "MQTT settings saved, but could not reach the MQTT server. Please verify the server address and credentials."
- If the network isn't available at all (WiFi not connected yet), skips the check with a log message

This way users get clear feedback without losing their settings.

rcatal01:
Pushed a follow-up commit addressing the Copilot review feedback:

- Fixed warning message wording to reflect correct timing ("Settings will be saved" instead of "Settings saved")
- Added null check on `clientNotificationPool.allocZeroed()` to match the `AdminModule::sendWarning()` pattern
- Cleaned up test comments to accurately describe conditional behavior

rcatal01:
After reviewing the Copilot feedback on the advisory connectivity check, I've removed it entirely and returned to the simpler approach from the first commit.

The issue: `connectPubSub()` mutates the file-scope `isConnected` flag, which can temporarily put the running MQTT module into a disconnected state during what should be a config validation. It also performs synchronous network operations that can block the config-save path on DNS/TCP timeouts.

The cleaner solution: `isValidConfig()` now only validates configuration correctness (TLS support check, default server port check). The MQTT reconnect loop handles connectivity after settings are persisted and the device reboots. If the broker is unreachable, the user will see it in the MQTT connection status — no special validation-time notification needed.

This is the version that was tested on hardware (Heltec Wireless Tracker) and confirmed working: MQTT settings now persist regardless of broker reachability.

thebentern:
@rcatal01 I'm not in favor of this as-is. Removing the connectivity check diagnostic with client notification strips that whole process of any value it had before and people will end up being confused the other direction when their mqtt connection doesn't work, with no feedback.

rcatal01:
@thebentern @Xaositek Understood — I agree users need connectivity feedback. The issue Copilot raised is that `connectPubSub()` mutates the module's `isConnected` state as a side effect.

Would a **lightweight TCP-only connection test** be acceptable? The idea: use a standalone `MQTTClient` scoped locally within `isValidConfig()` to check if the server IP/port is reachable, send a WARNING notification if not, then destroy the test client. No MQTT session, no state mutation, settings still always save.

Want to confirm this direction before I push more code.

ackstorm23:
> @rcatal01 I think I'm in favor of this change because its currently confusing users, but I'm not sure silently failing a different way is better. I think perhaps we should throw a client notification up to the client letting them know that we could not reach the configured MQTT broker and to double check the settings.

I still believe there are bugs in the code because it will claim to not be able to reach a server that it was connecting to without issue just fine for days until I disable MQTT, and when re-enabling it *with the same settings* it will log that the server is unreachable and refuse to enable it.   Network captures at the time showed no attempt to connect so I think it's logging does not provide enough information about the failure to determine what is *actually* failing.  

The original pubsub codebase has not been updated since 2020, and was absolutely full of unresolved git issues.  They are using [a fork maintained by one man](https://github.com/thingsboard/pubsubclient )  which received some updates but has not been updated in a year.  It may simply be that the pubsub codebase is buggy and we are victim to using that implementation. :shrug: 

Looks like [hmueller01](https://github.com/hmueller01/pubsubclient3) is the only person actively maintaining a fork at this point, with many updates.

ackstorm23:
> @rcatal01 I'm not in favor of this as-is. Removing the connectivity check diagnostic with client notification strips that whole process of any value it had before and people will end up being confused the other direction when their mqtt connection doesn't work, with no feedback.

to be fair, the 'feedback' it provides now is not helpful.   It provides no details about the failure, simply stating that it could not connect.  

Was it a failure because of temporarily DNS resolution?  Was it actually unable to reach the target? did it take too long to get a response from the target? did it actually try to connect at all?   (I've seen packet captures where it did not even try to connect over the fire)  or maybe it received an unexpected reply from the MQTT server.  No clues are given. 

We don't even know if it would self-recover after the first failure because it never gives the opportunity to try.   

On another note, anyone who might want to pre-configure MQTT ahead of time for a device so that it's already ready to transmit when deployed is actively blocked from doing so because the MQTT server simply has to be reachable at the time it is turned on. :shrug: 

IMHO it should turn MQTT on or off, and not care if it can connect or not.  It will already log each time  it cannot reach the MQTT server so it's not as if the user has no way to know if there is a problem.

Issue #9580, opened 2026-02-09
[Bug]: RPI Zero 2W Linux Native meshtasticd freezes after constant `Can not send yet, busyRx`


Luigi-Pizzolito:
### Category

Hardware Compatibility

### Hardware

Linux Native

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.15

### Description

I am trying to setup Meshtastic on a RPI Zero 2W using the Linux native firmware and `meshtasticd`. I have a MeshAdv Pi Hat equivalent using the same pinout and the E22-900M30S.

I've been able to reproduce this issue on both the `alpha` and `beta` releases of `meshtasticd`, currently using `alpha`.

Below is my `config.yaml` output from `meshtasticd`:
```yaml
Lora:
  Module: sx1262
  CS:
    pin: 21
    line: 21
    gpiochip: 0
  IRQ:
    pin: 16
    line: 16
    gpiochip: 0
  Busy:
    pin: 20
    line: 20
    gpiochip: 0
  Reset:
    pin: 18
    line: 18
    gpiochip: 0
  TXen:
    pin: 13
    line: 13
    gpiochip: 0
  RXen:
    pin: 12
    line: 12
    gpiochip: 0
  DIO2_AS_RF_SWITCH: false
  DIO3_TCXO_VOLTAGE: 1.8
  spidev: /dev/spidev0.0
  spiSpeed: 2000000
Input:
  {}
Logging:
  LogLevel: info
Webserver:
  RootPath: /usr/share/meshtasticd/web
  SSLKey: /etc/meshtasticd/ssl/private_key.pem
  SSLCert: /etc/meshtasticd/ssl/certificate.pem
  Port: 443
General:
  ConfigDirectory: /etc/meshtasticd/config.d/
  AvailableDirectory: /etc/meshtasticd/available.d/
  MaxMessageQueue: 100
  MaxNodes: 200
```

This problem happens right after trying to transmit a packet, e.g. when running the following CLI command: `meshtastic --host 192.168.8.199 --sendtext "This is a longer test message for radio verification"`. After running and getting the `[RadioIf] Can not send yet, busyRx` warning, the API interface stops working until a `meshtasticd` restart. I have tried resetting the transceiver through the GPIO reset pin, as well as power cycling the entire system.

[...]

Issue #4298, opened 2024-07-17
[Bug]: Issues with GPIO Pins and SX1262 Radio Initialization


orhunavcu:
### Category

Serial

### Hardware

Other

### Firmware Version

2.3.13.83f5ba0

### Description

Hi, I am encountering issues while setting up Meshtastic on my Raspberry Pi 5 with a Waveshare LoRa 433MHz GNSS module. Below is the log output and specific errors I am facing:

`Portduino is starting, HWID=1, VFS root at /root/.portduino/default
Setting up Meshtastic on Portduino...
Using /etc/meshtasticd/config.yaml as config file
Warning, cannot claim pin GPIO21St16invalid_argument
Warning, cannot claim pin GPIO16St16invalid_argument
Warning, cannot claim pin GPIO20St16invalid_argument
Warning, cannot claim pin GPIO18St16invalid_argument
INFO  | ??:??:?? 0 

//\ E S H T /\ S T / C

DEBUG | ??:??:?? 0 Filesystem files:
DEBUG | ??:??:?? 0  /.. (0 Bytes)
DEBUG | ??:??:?? 0  /prefs/.. (0 Bytes)
DEBUG | ??:??:?? 0  /prefs/db.proto (512 Bytes)
DEBUG | ??:??:?? 0  /prefs/channels.proto (57 Bytes)
DEBUG | ??:??:?? 0  /prefs/. (0 Bytes)
DEBUG | ??:??:?? 0  /. (0 Bytes)
INFO  | ??:??:?? 0 No I2C device configured, skipping.
INFO  | ??:??:?? 0 No I2C devices found
INFO  | ??:??:?? 0 Meshtastic hwvendor=37, swver=2.3.13.83f5ba0
INFO  | ??:??:?? 0 Initializing NodeDB
INFO  | ??:??:?? 0 Loading /prefs/db.proto
INFO  | ??:??:?? 0 Loaded /prefs/db.proto successfully
INFO  | ??:??:?? 0 Loaded saved devicestate version 22, with nodecount: 1
INFO  | ??:??:?? 0 File /prefs/config.proto not found
INFO  | ??:??:?? 0 Installing default LocalConfig
INFO  | ??:??:?? 0 Setting default channel and radio preferences!
DEBUG | ??:??:?? 0 Expanding short PSK #1
INFO  | ??:??:?? 0 Wanted region 0, using UNSET
INFO  | ??:??:?? 0 File /prefs/module.proto not found
[...]

PR #9939, opened 2026-03-18
Fix intermittent busyRx on Portduino SX1262 (stale preamble IRQ)

### LINKED ISSUE #9933: SX1262 intermittent busyRx on Raspberry Pi 5 with TXen/RXen (PiMesh 1W / MeshAdv)
## Description

Intermittent `[RadioIf] Can not send yet, busyRx` errors on Raspberry Pi 5 with an SX1262-based LoRa HAT using TXen/RXen RF switching (PiMesh 1W V1 / MeshAdv Pi Hat with Ebyte E22-900M30S). The radio initializes successfully but intermittently gets stuck in RX state due to false preamble detection, preventing transmission.

## Hardware

- **Board:** Raspberry Pi 5 Model B Rev 1.1
- **OS:** DietPi (Debian Trixie 13, kernel 6.12.75+rpt-rpi-2712)
- **LoRa HAT:** PiMesh 1W V1 (Ebyte E22-900M30S / SX1262, 1W with external PA)
- **GPS:** ATGM336H (UART on /dev/ttyAMA0)
- **meshtasticd:** 2.7.20 (016e68e, native-tft)
- **Connection:** SPI (spidev0.0), GPIO via RP1 (gpiochip4)

## LoRa Config (`config.d/lora-pimesh-1w-v1.yaml`)

```yaml
Lora:
  Module: sx1262
  CS: 21
  IRQ: 16
  Busy: 20
  Reset: 18
  TXen: 13
  RXen: 12
  DIO3_TCXO_VOLTAGE: true
  SX126X_MAX_POWER: 22
  spidev: spidev0.0
  gpiochip: 4
```

## config.txt (relevant lines)

```
dtparam=spi=on
enable_uart=1
dtoverlay=disable-bt
```

Note: `dtoverlay=spi0-0cs` must NOT be used on Pi 5 — it conflicts with the RP1 SPI controller and causes `SX126x init result -2` (chip not found) followed by a crash.

## Symptoms

1. Radio initializes successfully (`SX126x init result 0`)
2. Some packets transmit fine (e.g., NodeInfo at 30s after startup)
3. Other packets intermittently fail with `busyRx` (e.g., HostMetrics at 60s, GPS position at ~28s)
4. Debug log shows `Ignore false preamble detection` immediately after the busyRx warnings
5. Once stuck, the radio never recovers to transmit that packet — the queue fills and discards

## Debug Log (verbose mode)

```
INFO  | Set radio: region=US, name=LongFast, config=0, ch=19, power=30
INFO  | Final Tx power: 22 dBm
INFO  | SX126x init result 0
INFO  | Frequency set to 906.875000
INFO  | Bandwidth set to 250.000000
INFO  | Power output set to 22
INFO  | Set RX gain to boosted mode; result: 0
INFO  | Applied SX1262 register 0x8B5 patch for RX improvement
INFO  | sx1262 init success
...
INFO  | [GPS] Send pos@...:6 to mesh (wantReplies=0)
DEBUG | [GPS] enqueue for send (id=... encrypted len=68 ...)
WARN  | [RadioIf] Can not send yet, busyRx
WARN  | [RadioIf] Can not send yet, busyRx
WARN  | [RadioIf] Can not send yet, busyRx
DEBUG | [RadioIf] Ignore false preamble detection
```

After longer runs without recovery:
```
INFO  | [RadioIf] tophone queue status queue is full, discard oldest
```

### Pi 5 GPIO chip matters
- `gpiochip: 0` (default) → radio initializes but BUSY pin reads incorrectly, busyRx on every TX attempt
- `gpiochip: 4` → radio works, busyRx is intermittent rather than constant
- On Pi 5 DietPi/Trixie, `/dev/gpiochip4` symlinks to `/dev/gpiochip0`, but the RP1 GPIO driver behaves differently depending on which chip number is requested

### SX126X_MAX_POWER: 22 reduces frequency
- Without `SX126X_MAX_POWER: 22`: busyRx on nearly every TX attempt
- With `SX126X_MAX_POWER: 22`: busyRx is intermittent (~1 in 3 TX attempts)

### spi0-0cs overlay breaks SX1262 init on Pi 5
- With `dtoverlay=spi0-0cs`: `SX126x init result -2` → crash (`free(): invalid pointer`)
- Without it: init succeeds. Default Pi 5 SPI config works fine since meshtasticd manages CS via GPIO21

## Likely Root Cause

The RP1 GPIO controller on Pi 5 has different interrupt/edge detection behavior than the BCM GPIO on Pi 4 and earlier. The SX1262's IRQ pin (GPIO16) edge detection or the BUSY pin (GPIO20) level reading appears to be unreliable through the RP1, causing false preamble detections that prevent the radio from transitioning out of RX mode.

This may be related to:
- How `libgpiod` handles edge events on the RP1
- Timing of BUSY pin polling after an IRQ event
- The RP1's GPIO interrupt coalescing or debouncing behavior

## Related Issues

- #9580 — Same busyRx freeze on RPi Zero 2W with identical hardware config (TXen/RXen, SX1262)
- #4298 — GPIO pin issues with SX1262 on RPi5 (resolved by correct gpiochip setting)
- jgromes/RadioLib#1218 — SX1262 stuck in FS mode on RPi5

## Steps to Reproduce

1. Raspberry Pi 5 with any SX1262 HAT using TXen/RXen (PiMesh 1W V1, MeshAdv, etc.)
2. Install meshtasticd 2.7.20
3. Configure with gpiochip: 4, SX126X_MAX_POWER: 22
4. Set region, wait for GPS fix or HostMetrics timer (60s)
5. Observe intermittent busyRx in logs: `journalctl -u meshtasticd -f`

NearlCrews:
## Summary

On Linux-native (Portduino) builds, the SX1262 radio gets stuck in RX with `Can not send yet, busyRx` / `Ignore false preamble detection` errors, blocking TX until the existing stale-flag fallback in `RadioLibInterface::receiveDetected` times out (`2 * preambleTimeMsec`). Under high radio activity I measured this on 60-80% of TX attempts; at low activity it was sporadic.

Fixes #9933. Related: #9580 (same symptoms on RPi Zero 2W), #4298 (GPIO pin issues with SX1262 on RPi 5).

## Root cause

`startReceiveDutyCycleAuto` puts the SX1262 into a sleep/CAD/sleep cycle. Every CAD pass can latch a `PREAMBLE_DETECTED` IRQ flag even when no real packet is arriving. On bare-metal MCUs those stale flags get cleared/overwritten within a few symbol times before anything sees them, because IRQ polling happens at nanosecond scale. On Linux, `gpiod` reads are microseconds each, so `getIrqFlags()` regularly catches a stale CAD-induced flag. `isActivelyReceiving()` then returns `true`, TX is gated, and we wait out the `2 * preambleTimeMsec` stale-flag fallback before unblocking — by which time `setTransmitDelay()` has already pushed the TX attempt into a new random window where the same race can recur.

## Fix

On Portduino only, replace `startReceiveDutyCycleAuto(preambleLength, 8, ...)` with `startReceive(RADIOLIB_SX126X_RX_TIMEOUT_INF, ...)`. Continuous receive has no sleep/CAD cycles, so the stale-flag source is eliminated at the root. Power saving from duty cycling is irrelevant on mains-powered Linux-class devices (Raspberry Pi, x86 gateways, femtofox).

`isActivelyReceiving()` is unchanged — the existing `receiveDetected` logic with its `2 * preambleTimeMsec` fallback remains the correct place to handle any residual stale flags on any platform.

## What this PR no longer does

An earlier revision also tried to accelerate stale-flag recovery inside `isActivelyReceiving()` on Portduino by clearing `PREAMBLE_DETECTED` and re-reading after 5 ms. That was wrong on two counts (thanks to @GUVWAF for catching it):

1. The SX1262 does not re-assert `PREAMBLE_DETECTED` later in the same reception — preamble detection is a one-shot IRQ, after which the chip moves to sync word → header → payload. If the flag was cleared mid-reception (e.g. between end-of-preamble and `HEADER_VALID`), the recheck would see "no preamble," declare the reception stale, and TX could fire into an in-progress RX.
2. The 5 ms recheck window was a constant, but preamble time scales with LoRa settings (`preambleLength * (2^sf) / bw`). LongFast is ~131 ms; LongSlow is ~524 ms. 5 ms was 25–100× too short to distinguish "real preamble, still decoding" from "stale flag."

That code is removed in the current commit. The continuous-receive change alone is sufficient on my hardware — I did not observe any residual `busyRx` after applying it.

## Behavior change

| Platform | Before | After |
| --- | --- | --- |
| Portduino SX126x | `startReceiveDutyCycleAuto(16-symbol preamble, 8-symbol RX window)` | `startReceive(RX_TIMEOUT_INF, ...)` |
| ESP32 / nRF52 / RP2040 / STM32WL SX126x | `startReceiveDutyCycleAuto(...)` | unchanged |

No change to `isActivelyReceiving` on any platform. No change on battery-powered builds.

## Test plan

Tested on Raspberry Pi 5 with SX1262 (E22-900M30S) HAT, full config (4 channels, MQTT, boosted RX gain, GPS, neighbor info, telemetry).

- [x] Before fix: `busyRx` on 60-80% of TX attempts under heavy traffic (firmware 2.7.15 and 2.7.20)
- [x] After fix: 0 `busyRx` across 10 consecutive 45-second test runs
- [x] After fix: 0 `busyRx` in a ~24 h sustained soak with mixed traffic
- [x] `Ignore false preamble detection` log no longer appears in steady state
- [x] Packet reception unchanged (NodeInfo / HostMetrics / GPS position / DeviceTelemetry all transmit and receive)
- [x] Radio init still succeeds (`SX126x init result 0`)

CLA will be signed before merge.

### DIFF
--- a/src/mesh/SX126xInterface.cpp
+++ b/src/mesh/SX126xInterface.cpp
@@ -315,14 +315,21 @@ template <typename T> void SX126xInterface<T>::startReceive()
     setTransmitEnable(false);
     setStandby();
 
+    // On Linux-native (Portduino), use continuous receive instead of duty cycle auto-receive.
+    // The duty cycle mode causes intermittent false preamble detections due to GPIO read latency
+    // in userspace (gpiod), which leads to "Can not send yet, busyRx" errors.
+    // Power saving from duty cycle is irrelevant on mains-powered Linux devices.
+#if ARCH_PORTDUINO
+    int err = lora.startReceive(RADIOLIB_SX126X_RX_TIMEOUT_INF, MESHTASTIC_RADIOLIB_IRQ_RX_FLAGS);
+    if (err != RADIOLIB_ERR_NONE) {
+        LOG_ERROR("SX126X startReceive %s%d", radioLibErr, err);
+        portduino_status.LoRa_in_error = true;
+    }
+#else
     // We use a 16 bit preamble so this should save some power by letting radio sit in standby mostly.
     int err = lora.startReceiveDutyCycleAuto(preambleLength, 8, MESHTASTIC_RADIOLIB_IRQ_RX_FLAGS);
     if (err != RADIOLIB_ERR_NONE)
         LOG_ERROR("SX126X startReceiveDutyCycleAuto %s%d", radioLibErr, err);
-#ifdef ARCH_PORTDUINO
-    if (err != RADIOLIB_ERR_NONE)
-        portduino_status.LoRa_in_error = true;
-#else
     assert(err == RADIOLIB_ERR_NONE);
 #endif
 


src/mesh/SX126xInterface.cpp
```diff
@@ -368,7 +375,25 @@ template <typename T> bool SX126xInterface<T>::isActivelyReceiving()
...
+#if ARCH_PORTDUINO
+    // On Linux-native platforms, the SX1262's PREAMBLE_DETECTED IRQ flag can become stale —
+    // it is latched and persists from internal CAD checks during receive mode. Due to GPIO read
+    // latency through gpiod, these stale flags cause false "busyRx" blocks that prevent TX.
+    // Fix: if we see PREAMBLE_DETECTED without HEADER_VALID, clear it and re-read. If the
+    // preamble was real, the radio will re-assert it within a few ms; if not, we avoid blocking TX.
+    constexpr uint32_t STALE_PREAMBLE_RECHECK_MS = 5;
+    uint16_t irq = lora.getIrqFlags();
+    if ((irq & RADIOLIB_SX126X_IRQ_PREAMBLE_DETECTED) && !(irq & RADIOLIB_SX126X_IRQ_HEADER_VALID)) {
+        lora.clearIrqFlags(RADIOLIB_SX126X_IRQ_PREAMBLE_DETECTED);
```
GUVWAF: You mention "Real receptions re-assert the flag within a few ms.". This may often be true, but I don't think it's guaranteed. What happens if you clear the IRQ *just* before the preamble ends and it continues decoding the LoRa header?
Even then, `STALE_PREAMBLE_RECHECK_MS` can not be a constant, it will depend on `preambleTimeMsec`, which depends on the LoRa settings. For slow presets, 5ms will not be enough to detect (part of) a preamble.

NearlCrews:
@GUVWAF you are right on both counts — thanks for the careful read. I've pushed an update (`797601f`, rebased onto current `master` at `68383c8`) that **drops the `isActivelyReceiving()` recheck entirely** and keeps only the `startReceive(RX_TIMEOUT_INF, ...)` change on Portduino.

**On your in-flight-reception concern:** the SX1262 fires `PREAMBLE_DETECTED` once per reception (when the preamble detector confirms the pattern) and then moves on to sync word / header / payload — it does not re-assert that flag while the same reception continues. So if my recheck cleared the flag during the tail end of a real preamble, `HEADER_VALID` might never arrive by the next poll and the code would correctly-but-wrongly mark the reception stale. That's a silent TX-into-RX collision risk, and it's indefensible.

**On the preset-dependent timing:** right too — `STALE_PREAMBLE_RECHECK_MS` as a constant was dimensionally wrong. `RadioInterface.cpp` already tracks the real preamble duration as `preambleTimeMsec = preambleLength * (pow_of_2(sf)) / bw`, and it varies from ~131 ms (LongFast) to ~524 ms (LongSlow) with 16-symbol preambles. 5 ms was 25–100× too short to distinguish "preamble decoding in progress" from "stale CAD flag."

**Why the remaining change is enough.** The root cause is `startReceiveDutyCycleAuto`'s internal CAD producing latched `PREAMBLE_DETECTED` flags that bare-metal MCUs overwrite within microseconds but Linux `gpiod` reads catch. Switching Portduino to continuous receive eliminates that CAD source, so the stale flags stop being produced. Any residual stale flags from other sources (boot state, noise) are already handled correctly by the existing `2 * preambleTimeMsec` fallback in `RadioLibInterface::receiveDetected` — that path will log `Ignore false preamble detection` and release TX. On the hardware I have (RPi 5 + E22-900M30S), I did not observe any `busyRx` after the continuous-receive change alone, including a ~24 h soak.

@Copilot — the 5 ms vs 10 ms mismatch is moot now, that code is gone.

CLA is still unsigned; I'll sign before merge.

GUVWAF:
I don't see how not using the duty cycle receiving would solve the issue, since this only affects the radio and has nothing to do with handling the IRQ flag. Moreover, with our current preamble length, the `startReceiveDutyCycleAuto()` function falls back to just `startReceive()` as the period the radio can sleep is too low, so it's already equivalent to what you propose. 

I think the reason it works for you now is because #9895 is merged in the meantime.

NearlCrews:
You're right on both counts. We pass `(preambleLength=16, minSymbols=8)` at `SX126xInterface.cpp:330`, so RadioLib 7.6.0 computes `sleepSymbols = 16 - 2*8 = 0` (`SX126x.cpp:523`), `sleepPeriod = 0`, and trips the early return at `SX126x.cpp:546-547`. We land in plain `startReceive(RX_TIMEOUT_INF, ...)` either way, so the Portduino branch was a no-op for our config.

Plausible that #9895 explains why my soak stopped reproducing the busyRx. I haven't isolated that. Closing this.

PR #9950, opened 2026-03-19
t2000 WIP

caveman99:


Szetya:
Hi! I have a T2000C device too. I see a lot of potential in it, but I haven't had time to work on it yet. Is the first flash done via the USB port? Could you send me a photo showing the order of the VCC, GND, D+, and D- pins?

Szetya:
> Hi! I have a T2000C device too. I see a lot of potential in it, but I haven't had time to work on it yet. Is the first flash done via the USB port? Could you send me a photo showing the order of the VCC, GND, D+, and D- pins?

It looks like it worked. :)
I used ST-Link to flash the firmware. I also flashed the bootloader using ST-Link (OOCD).
In the end, I decided not to use USB; I didn't want to solder it onto the PADs.

PR #9956, opened 2026-03-19
Fix TFTDisplay::display to align pixels at 32-bit boundary

notmasteryet:
The device, such as esp32c6, require 32-bit alignment for the data.

The patch aligns start and end of the update pixels buffer. In worst case, it will add two extra pixels per row.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [X] Private hardware based on esp32c6 with 1.47" LCD

### DIFF
--- a/src/graphics/TFTDisplay.cpp
+++ b/src/graphics/TFTDisplay.cpp
@@ -1278,6 +1278,10 @@ void TFTDisplay::display(bool fromBlank)
             // Most displays will have even number of pixels in a row -- this will be in bounds
             // of the displayWidth. (Hopefully odd displays will just ignore that extra pixel.)
             x_LastPixelUpdate |= 1;
+            // Ensure the last pixel index does not exceed the display width.
+            if (x_LastPixelUpdate >= displayWidth) {
+                x_LastPixelUpdate = displayWidth - 1;
+            }
 #if defined(HACKADAY_COMMUNICATOR)
             tft->draw16bitBeRGBBitmap(x_FirstPixelUpdate, y, &linePixelBuffer[x_FirstPixelUpdate],
                                       (x_LastPixelUpdate - x_FirstPixelUpdate + 1), 1);


notmasteryet:
The alternative is just start filling `linePixelBuffer` from start, but we need to maintain separate index for that.

notmasteryet:
> I'm not sure if the original author of this AI generated PR deeply understood the problem that was suggested to be solved by the AI (typically via trial and error).

Problem was observed on real hardware. AI was not even trying to solve the issue. New code is all human.

> Neither is the case if we look at the [allocation](https://github.com/notmasteryet/firmware/blob/79259cf88d210ee196a3797577f43981103b00c9/src/graphics/TFTDisplay.cpp#L1513) of linePixelBuffer:
> 
> ` this->linePixelBuffer = (uint16_t *)malloc(sizeof(uint16_t) * displayWidth);`

malloc allocations are at least 4 bytes aligned, that is what my fix was counting on.

> I have tested the code on a T-Deck and a T-Lora-Pager and it still works, it seems the change doesn't break anything.

Thanks for testing on other TFT displays. I tried not to break it there.

notmasteryet:
> So I guess the issue that is solved by having an even x coordinate is rather related to the display hardware of the mentioned 1.47" LCD which may require an even pixel coordinate for the SPI transfer. I know some amoled LEDs require this (e.g. the CO5300).

That's very plausible, but I did not have time to verify this theory I pursue initial solution since all my googling ended up with esp32c6 DMA thing. Maybe some ST7789 are picky about even pixels.

P.S. I didn't look at underlined display libraries to see what they do with linePixelBuffer.

whywilson:
I've confirmed that this PR fix the display Pixel issues on Arduino Nesso N1.

fifieldt (approved):
This makes sense logically to me, mverch found no badness, and a user reported it helped

PR #9988, opened 2026-03-23
Add regular reboot option, disabled by default

korbinianbauer:
This adds the option to enable regular reboots every `DEFAULT_REGULAR_REBOOT_DAYS` days.
Allowed range is 1 - 49 days.

Per default, this is disabled (set to -1).

For me this has proven vital to keep nodes that I connect and disconnect to via Bluetooth many times per day working for more than ~1 month without a manual reboot.

I deliberately chose whole days as the unit for this PR, to make sure and clear this is not suitable to create nodes that keep restarting so much that they hurt the mesh.

## 🤝 Attestations

  - [X] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [X] RAK WisBlock 4631

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -366,6 +366,26 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define DEFAULT_REBOOT_SECONDS 7
 #endif
 
+// Allows scheduling a periodic reboot at boot-time.
+// Default is -1 (disabled).
+// Unit is whole days.
+// Effective range: 1 to 49 days.
+#ifndef DEFAULT_REGULAR_REBOOT_DAYS
+#define DEFAULT_REGULAR_REBOOT_DAYS -1
+#endif
+
+#if DEFAULT_REGULAR_REBOOT_DAYS < -1
+#error DEFAULT_REGULAR_REBOOT_DAYS must be -1 (disabled) or a positive whole number of days
+#endif
+
+#if DEFAULT_REGULAR_REBOOT_DAYS == 0
+#error DEFAULT_REGULAR_REBOOT_DAYS must be -1 (disabled) or a positive whole number of days
+#endif
+
+#if DEFAULT_REGULAR_REBOOT_DAYS > 49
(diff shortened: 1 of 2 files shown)


shalberd:
> For me this has proven vital to keep nodes that I connect and disconnect to via Bluetooth many times per day working

the question is: why does Bluetooth / the Bluetooth Stack have the effect of not making nodes work anymore

korbinianbauer:
@shalberd Agreed, though that's just an example for why you might want to use this.

For me it's a patch until I (or someone else) figures out why this happens and why the watchdog does not trigger.

fifieldt:
I believe our current policy on a reboot timer feature is to treat the need for it as a bug that should instead be fixed. How can we help debug those lockups, do you think?

korbinianbauer:
@fifieldt 

Here's what I know:

- Hardware is a RAK4631 on a 19003 baseboard, with 3x 18650 NMC and a 6V 3W solar panel.

- I observed the issue twice. The first time with 2.6.10 beta, the second time with 2.6.11 beta. The node is now running my own 2.7.21 build with the reboot timer enabled.

- The node crashed after around 32 and 36 days respectively.
Here's an Uptime graph. You can see the uptime reach ~3M seconds twice, then the node crashes and the graph stays flat (no new data. Then the uptime drops back to 0 when I noticed the issue and reset the node. On the right end you can see the uptime resetting regularly, since the node reboots once per day now.

IMG_20260405_123021_593~2


- We have nodes with identical firm- and hardware running without issues for 6 months+, running as Router_Late.

- Problematic node is Client_Mute, Good nodes are Router_Late

- The good nodes are on mountaintops. They have Bluetooth enabled and are likely seen by dozens of smartphones per day, but the visitors don't have the PIN to connect, so I assume no one does.

- The bad node is on my balcony, where my smartphone running the companion app gets in and out of reach multiple times per day.

- They're all in the same small mesh (Medium Fast) with ~10 nodes and very little traffic. Ch_Util has always been <1% for all of them in the observed timeframe.

- Power is not an issue, they all sit at 4.0-4.15V at all times.

PR #9995, opened 2026-03-24
feat(telemetry): add DS18B20 1-Wire multi-thermometer support

Elwimen:
## Summary

Adds support for the Dallas/Maxim DS18B20 1-Wire digital thermometer, enabling multiple temperature sensors on a single GPIO pin with full runtime configuration via the Meshtastic CLI — no compile flags or firmware recompile needed.

### Features

- **Runtime pin configuration**: `meshtastic --set telemetry.ds18b20.pin 42` — live reload, no reboot required
- **Up to 16 sensors per bus** per packet; additional sensors sent as overflow chunk packets
- **FNV-1a 32-bit `device_id`** per sensor — stable across reboots, compact on the wire (saves 4 bytes vs fixed64 ROM code)
- **Temperature aggregation modes** for the legacy `temperature` scalar: `FIRST` (default), `AVERAGE`, `MIN`, `MAX`
- **Mesh broadcast control**: `mesh_enabled` flag controls whether overflow chunk packets go over LoRa or stay phone-only
- **Hot-plug support**: bus rescanned every cycle to discover added/removed sensors
- **Async non-blocking conversion**: 2-phase cycle avoids blocking during the 750ms DS18B20 conversion time
- **No compile flag needed**: `__has_include(<DallasTemperature.h>)` guard activates the driver on any board that has the library in `lib_deps`

### CLI usage

```bash
meshtastic --set telemetry.environment_measurement_enabled true
meshtastic --set telemetry.ds18b20.pin 42
meshtastic --set telemetry.ds18b20.mode AVERAGE      # optional, default FIRST
meshtastic --set telemetry.ds18b20.mesh_enabled true # optional, default false
```

### Protobuf changes

- New `DS18B20Reading` message (`fixed32 device_id`, `float temperature`) added to `EnvironmentMetrics` as `repeated DS18B20Reading ds18b20_readings = 23`
- New `DS18B20Config` sub-message in `TelemetryConfig` (field 16): pin, TemperatureMode enum, mesh_enabled
- `DS18B20 = 51` added to `TelemetrySensorType` enum
- Protobufs submodule bumped to v2.7.20

### Hardware note

> ⚠️ The upstream `paulstoffregen/OneWire` library does not fully support GPIO > 33 on ESP32/ESP32-S3. The firmware emits a `LOG_WARN` at runtime if such a pin is configured. GPIO ≤ 33 is recommended until the upstream library is patched. Testing with a patched OneWire on GPIO 44 (T3S3) confirmed correct operation on higher-numbered pins.

### Wire it up

```
DS18B20 VDD → 3.3V
DS18B20 GND → GND
DS18B20 DQ  → GPIO pin (e.g. GPIO 42)
4.7kΩ pull-up between DQ and 3.3V
```

## 🤝 Attestations

  - [x] Heltec (Lora32) V4
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631
  - [x] Other: LilyGo T3S3 v1 (ESP32-S3, GPIO 42 and GPIO 44 with patched OneWire), T-Echo


oscgonfer:
Hi @Elwimen ,

Thank you for the PR. Normally, protobufs are not included in these PRs, but rather referenced as a separate one in the protobuf repo in case this makes it in. Can you remove all pb.h and pb.cpp changes from the PR?

In addition, this PR would force using the sensor via one-wire interface, when there are i2c to 1-wire converters with multiple channels like [this one](https://github.com/adafruit/Adafruit_DS248x/tree/main), supporting multiple channels. I'd say it'd be better to add support to the driver, instead of using 1-wire directly, as it will take up the bus and we wouldn't be able to use it otherwise. I understand you're defining a pin for it, but imho this breaks somehow with the advantages of having a plug and play system relying on I2C. There has been a previous attempt to work on this here: https://github.com/meshtastic/firmware/pull/8723. See discussion.

I'm happy to support on this, but the overall idea to integrate one-wire as a one-of solution for just this sensor is probably outweighed by the other option via the i2c driver.

In addition, can you acknowledge the use (or not use) of AI to generate the code?

Elwimen:
Hi @oscgonfer, thanks for the review.

**On the protobuf files:** Fully agreed — I'll strip the generated pb.h/pb.cpp changes from this PR and submit a separate one against the protobufs repo.

**On 1-Wire vs DS2482:**

A few things worth clarifying here:

1. **1-Wire does not occupy I2C.** It runs on a dedicated GPIO pin (user-configured, defaults to unset). I2C sensors continue working exactly as before — the two buses are completely independent.

2. **The DS2482 bridge is not implemented in the firmware.** Recommending an I2C-to-1-Wire bridge as the preferred path when there's no existing driver, no PR for it, and the previous attempt (#8723) was never merged leaves users with no DS18B20 support at all. This PR gives them working support today.

3. **1-Wire has a fundamental advantage over I2C for this sensor class: no address collisions.** Every DS18B20 ships with a factory-burned unique 64-bit ROM code. You can wire 20+ sensors to one GPIO line and the driver enumerates them automatically — no DIP switches, no jumpers, no address remapping. With I2C sensors you eventually run out of available addresses. For temperature arrays (greenhouse rows, tank monitoring, HVAC zones) this matters.

4. **The DS2482 adds cost and complexity for no gain.** It's an extra IC ($1-2), extra footprint on the PCB, a separate driver to maintain, additional flash overhead, and an extra failure point — all to reach sensors that the GPIO pin already talks to natively. The DS18B20 is arguably the most widely deployed digital temperature sensor in embedded/IoT precisely because a single wire and a pullup resistor is all it takes.

5. **Community demand is real.** Looking at the #8723 discussion, people have been asking for DS18B20 support for years. The direct 1-Wire approach is what hobbyists and professionals already have wired up. Making them buy and solder a bridge chip to use hardware they already own is a poor user experience.

I'm happy to discuss the I2C bridge path as a *future* addition if someone wants to write that driver. But I don't think blocking a working, well-tested direct 1-Wire implementation on a hypothetical alternative is the right call for the project.

The code is based on 20+ years of embedded experience — this approach is battle-tested in production deployments.

Elwimen:
Is there any reason not to add the support for OneWire devices?

oscgonfer:
> Is there any reason not to add the support for OneWire devices?

Hi @Elwimen, there will be support for one-wire devices, but via I2C bridges. The reason is that the agreement for sensor implementation in meshtastic has been that, for any other protocol other than I2C (one-wire, UART, etc.), if there is an alternative possibility through I2C or an external MCU that handles the protocol and custom implementation, that should be the preferred option. However, if there is no alternative, and the sensor is strategic, then an implementation can be evaluated. The idea is to first agree among involved contributors on what is the best approach, and then implement it. 

For the Dallas sensors, since there is an alternative for implementing them via fairly simple and inexpensive bridges, we think it's a good trade-off and support will be added, as it's a feature requested many times.

If 8-channel bridges is not if enough, a separate approach with a small MCU handling the communication could be developed if needed.

Elwimen:
Hi @oscgonfer,

sorry for the defensive tone earlier, that was on me. Re-reading your
reply, the I2C-first policy makes sense at Meshtastic's scale. Consistent
architecture across hundreds of board variants beats case-by-case exceptions.
Understood.

One thing worth front-loading in docs: none of the current Meshtastic vendors ship
a 1-Wire bridge on-board, so users are always adding external hardware regardless.
The bare IC path is cheap enough that it shouldn't be a barrier, but users need to
find it before they land on an Adafruit breakout:

| Hardware | Price | Notes |
|---|---|---|
| Wio-SX1262 + XIAO ESP32S3 | $9.90 | Complete node |
| Heltec WiFi LoRa 32 V3 | $17.90-19.90 | Complete node |
| LilyGO T3S3 | $23.65 | Complete node |
| DS2484 bare IC + 4.7k pullup | ~$2.61 | Single channel, 4 I2C addresses |
| DS2482-100 bare IC + 4.7k pullup | ~$2.87 | Single channel, 4 I2C addresses |
| DS2482-800 bare IC | ~$14 | 8 independent channels |

Analog Devices is the only manufacturer of I2C-to-1-Wire bridge silicon, but the
family covers the relevant use cases well:

| Bridge IC | Channels | I2C Addresses | Price | Notes |
|---|---|---|---|---|
| DS2484 | 1 | 4 (ADDR pin) | ~$2.61 | Integrated level translator, SOT-23 |
| DS2482-100 | 1 | 4 (AD0, AD1) | ~$2.87 | 8-SOIC, PCTLZ for external MOSFET |
| DS2482-800 | 8 | 8 (AD0-AD2) | ~$14 | 8 independent buses, up to 64 with 8 chips |

DS2484 and DS2482-100 share a compatible command set, so one driver covers both.
More importantly, the bridge handles ROM Search and bus mechanics once - every
additional 1-Wire device type is just a thin handler on top that knows its
commands and how to parse the scratchpad. The device family is broader than most
people realize:

| Device | Family | Type | Notes |
|---|---|---|---|
| DS18B20 | 28h | Temperature | Primary target, 9-12-bit |
| DS18S20 | 10h | Temperature | Same read sequence |
| DS1822 | 22h | Temperature | +-2C economy variant |
| DS1825 | 3Bh | Temperature | Hardware address pins, no ROM search |
| DS28EA00 | 42h | Temperature + I/O | + 2 GPIO |
| DS2438 | 26h | Voltage + current + temp | Foundation for humidity/barometric/solar boards |
| DS2408 | 29h | I/O expander | 8-channel open-drain |
| DS2413 | 3Ah | I/O expander | Dual bidirectional I/O |
| DS1923 | 21h | Data logger | Temp + humidity iButton |

One implementation note: the dominant flash cost in the current code is STL
container overhead, not the protocol logic. Fixed arrays keep the whole thing lean
and new device support stays incremental.

Happy to rework this as a bridge-based implementation (as soon as I get my hands on one). Do you have a preferred
chip target or driver structure in mind? Better to align before writing any code.

oscgonfer:
Hi @Elwimen 

Thank you for your comments. I have ordered some drivers and we can work on this together if you like. I'd suggest we move this to a discussion and we move forward there.

oscgonfer:
https://github.com/meshtastic/firmware/discussions/10010

PR #9999, opened 2026-03-24
Use UDP as roof node <---> indoor nodes backchannel

slp:
For deployments with a roof node and one or more indoor nodes, having the latter to reliably receive broadcast messages from the Mesh is a challenge. CLIENT_BASE now behaves as ROUTER_LATE, but only for packets having one of the indoor nodes as source or final destination (assuming they have been configured as favourites). This has prompted some users to configure their roof nodes as ROUTER_LATE, which can degrade the Mesh.

This PR promotes the use of UDP broadcast as the backchannel between the roof and the indoor nodes, increasing the reliability by which broadcast packets arrive to the latter while also decreasing the overall LoRa traffic. This is achieved by slightly changing the behaviour of CLIENT_MUTE and CLIENT_BASE when UDP broadcast in enabled:

- CLIENT_MUTE: for locally generated packets, if there's an active `udpHandler` and sending the packet via UDP has succeeded, LoRa broadcasting is skipped.
- CLIENT_BASE: if there's an active `udpHandler`, incoming packets are also rebroadcasted via UDP.

## Open questions

- Should this behaviour be gated behind a configuration option? Since AFAIK the UDP broadcast feature is not yet stable, I've assumed it's safe to change the behaviour of CLIENT_MUTE and CLIENT_BASE when it's enabled without the burden of maintaining an additional configuration option, but can be done if deemed necessary. 
- Should going through the roof node via UDP count as a hop? This PR doesn't implement a special treatment for packets coming from the UDP transport, meaning it does count as a hop. If people think it's better to not count as a hop, we can implement the exception either here or in a future PR (TBH, I'd prefer the latter). 
- Is UDP broadcast reliable on arbitrary consumer-grade WiFi equipment? I have my doubts about it, so if people like this approach and this PR goes in, I'll try to implement a UDP unicast option with a configurable list of IPs in another PR.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] RAK 3112

### DIFF
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -394,7 +394,12 @@ ErrorCode Router::send(meshtastic_MeshPacket *p)
 
 #if HAS_UDP_MULTICAST
     if (udpHandler && config.network.enabled_protocols & meshtastic_Config_NetworkConfig_ProtocolFlags_UDP_BROADCAST) {
-        udpHandler->onSend(const_cast<meshtastic_MeshPacket *>(p));
+        bool res = udpHandler->onSend(const_cast<meshtastic_MeshPacket *>(p));
+        if (res && config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT_MUTE) {
+            LOG_DEBUG("Packet successfully sent via UDP, skipping LoRa");
+            packetPool.release(p);
+            return ERRNO_OK;
+        }
     }
 #endif
 
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -830,6 +830,23 @@ void Router::handleReceived(meshtastic_MeshPacket *p, RxSource src)
             }
         }
 #endif
+#if HAS_UDP_MULTICAST
+        if (config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT_BASE && udpHandler &&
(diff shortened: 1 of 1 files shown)


GUVWAF:
It's a neat idea, but it indeed changes the behaviour of existing roles and makes it even more complicated. I'm not against it per se, as indeed UDP broadcast is not (yet) in widespread use, but maybe we can do this in combination with a specific [Rebroadcast Mode](https://meshtastic.org/docs/configuration/radio/device/#rebroadcast-mode). Maybe @thebentern has some opinions on this one too?

slp:
> It's a neat idea, but it indeed changes the behaviour of existing roles and makes it even more complicated. I'm not against it per se, as indeed UDP broadcast is not (yet) in widespread use, but maybe we can do this in combination with a specific [Rebroadcast Mode](https://meshtastic.org/docs/configuration/radio/device/#rebroadcast-mode). Maybe @thebentern has some opinions on this one too?

I was also thinking it may be nicer to decouple this behaviour from the configured role. For instance, it might be useful having a CLIENT_MUTE that still rebroadcasts via UDP. How about adding two toggles, to go alongside the UDP config, like this:

- Rebroadcast messages from the mesh via UDP.
- For locally generated packets, if sending via UDP succeeded, skip LoRa.

Issue #4166, opened 2024-06-23
[Bug]: store and forward chronology


jranma:
### Category

Other

### Hardware

T-Beam S3

### Firmware Version

2.3.12

### Description

when one retrieves store and forward messages using "SF", all messages will arrive at the same time. Above all, they will all be displayed with the time they were retrieved, not the time they were received on the relay. It is therefore difficult, if not impossible, to understand the thread of the conversation because the chronologies are wrong.

garthvh:
Where are you seeing the messages with the same time?

jranma:
each time I request messages with a "SF" command, I get all the messages with the same time, which is the time at wich I requested them. 
Using Android APP.

GUVWAF:
Yes, that sounds correct. Currently the timestamp when the packet was actually received is not stored and send with the packet upon request.

jranma:
OK, not sure the messages come in the right order. 
Last time I received many messages and as the timestamp are all the same, I had a hard time understanding the logic of the conversation. 
Having the message sorted by their "real" timestamp would be a nice emprovement.

GUVWAF:
I think we can make this work. Currently we do store a relative time in the StoreForward history, but we can change it to an absolute time if we have one, and then not updating the `rx_time` for StoreForward packets when requesting them.

NeimadTL:
@GUVWAF @caveman99 @fifieldt , is this still up for grab ? Not familiar at all with the project but looks this [method](https://github.com/meshtastic/firmware/blob/develop/src/mesh/MeshService.cpp#L456) needs to be updated and this [line](https://github.com/meshtastic/firmware/blob/develop/src/mesh/MeshService.cpp#L195) needs to be removed based on what has been discussed already, correct ?

GUVWAF:
@NeimadTL Yes, it's still on the TODO, but to be honest it's more involved than you might think. You would have to add the original time somewhere here: https://github.com/meshtastic/firmware/blob/develop/src/modules/StoreForwardModule.cpp#L195, and then also send that with the packet. So it needs to be added to the protobuf definition of a StoreForward message: https://github.com/meshtastic/protobufs/blob/cb1f89372a70b0d4b4f8caf05aec28de8d4a13e0/meshtastic/storeforward.proto

And then you'd likely need to add an `alterReceived()` method to the StoreForwardModule to overwrite the `rx_time` with the original time from the StoreForward message.

NeimadTL:
[...]

PR #10045, opened 2026-03-31
Add `sent_at` member to packet history

NeimadTL:
This new member is assigned when the packet is sent and will be used to display messages chronologically in conversation.

Fix for #4166, I don't have the hardware to test it.

### DIFF
--- a/src/modules/StoreForwardModule.cpp
+++ b/src/modules/StoreForwardModule.cpp
@@ -193,6 +193,7 @@ void StoreForwardModule::historyAdd(const meshtastic_MeshPacket &mp)
     }
 
     this->packetHistory[this->packetHistoryTotalCount].time = getTime();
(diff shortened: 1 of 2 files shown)


GUVWAF:
This will not yet be enough. It will need to be sent over the air, because the recipient of the S&F packet needs to know it, and then update the `rx_time` of the packet for the apps to properly parse the time of reception.

NeimadTL:
Does this mean to set the `sent_at` in that method [here](https://github.com/meshtastic/firmware/blob/develop/src/mesh/Router.cpp#L311) as well ? 

Regarding the `rx_time`, should I set it with the `sent_at` [here](https://github.com/meshtastic/firmware/blob/develop/src/mesh/MeshService.cpp#L195)?

PR #10052, opened 2026-04-02
Add new development board DFRobot_LoRaWAN_ESP32-S3

Martin-hub1:
I added a new development board DFRobot_LoRaWAN_ESP32-S3, which works properly and has no impact on the operation of other development boards.

### DIFF
--- a/src/graphics/TFTDisplay.cpp
+++ b/src/graphics/TFTDisplay.cpp
@@ -76,7 +76,7 @@ class LGFX : public lgfx::LGFX_Device
             cfg.offset_y = TFT_OFFSET_Y;   // Panel offset amount in Y direction
 #if defined(TFT_OFFSET_ROTATION)
             cfg.offset_rotation = TFT_OFFSET_ROTATION;
-#elif
+#else
             cfg.offset_rotation = 0;       // Rotation direction value offset 0~7 (4~7 is upside down)
 #endif
             cfg.dummy_read_pixel = 8;      // Number of bits for dummy read before pixel readout
--- a/src/graphics/TFTDisplay.cpp
+++ b/src/graphics/TFTDisplay.cpp
@@ -74,8 +74,8 @@ class LGFX : public lgfx::LGFX_Device
             cfg.panel_height = TFT_HEIGHT; // actual displayable height
             cfg.offset_x = TFT_OFFSET_X;   // Panel offset amount in X direction
             cfg.offset_y = TFT_OFFSET_Y;   // Panel offset amount in Y direction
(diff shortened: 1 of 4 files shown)


fifieldt:
Thanks! Is that this one: https://www.aliexpress.com/item/1005010372478586.html ?

Martin-hub1:
@fifieldt Yes, this is the official store link for this product: https://www.dfrobot.com/product-2933.html

AteBraak:
I can close my [PR](https://github.com/meshtastic/firmware/pull/10014) in favor of EOM.

fifieldt:
Ah, sorry @AteBraak , you did indeed commit a little earlier. I do like that you've done cfg.offset_rotation = TFT_OFFSET_ROTATION instead of setting directly to 2. 

@Martin-hub1 , is it OK if we move our efforts to https://github.com/meshtastic/firmware/pull/10014 ? Perhaps you can review and collaborate with @AteBraak to get the best bits of both?

Martin-hub1:
Thank you @AteBraak @fifieldt ,I read your PR and think using TFT_OFFSET_ROTATION is better. I have updated my PR; please check it out.

src/graphics/TFTDisplay.cpp
```diff
@@ -74,8 +74,8 @@ class LGFX : public lgfx::LGFX_Device
             cfg.panel_height = TFT_HEIGHT; // actual displayable height
             cfg.offset_x = TFT_OFFSET_X;   // Panel offset amount in X direction
             cfg.offset_y = TFT_OFFSET_Y;   // Panel offset amount in Y direction
-#if defined(DFRobot_LoRaWAN_ESP32_S3)
-            cfg.offset_rotation = 2;
+#if defined(TFT_OFFSET_ROTATION)
+            cfg.offset_rotation = TFT_OFFSET_ROTATION;
 #elif
```
AteBraak: can you use #else instead of #elif?
example elif usage (else where in file):
#elif defined(ST7789_CS)
if no define is used as a condition please use #else

Martin-hub1: oh sorry,I have revised it.

AteBraak: thanks

AteBraak (approved):
I tested this on two dfr1195 devices. Tx/Rx work from both, screen functions as expected, BLE with meshtastic app works.

Code change looks good. I would ask Meshtastic familiar dev to check as well.

AteBraak:
Thank you @Martin-hub1 !

Martin-hub1:
@fifieldt  Hello, what else do I need to do for the submission and merging of this PR?

PR #10057, opened 2026-04-02
fix: replace weak PRNG with CSPRNG in crypto paths

Patrickschell609:
## Summary

- Replace all 4 uses of Arduino `random()` in cryptographic code paths with hardware-backed CSPRNG (`RNG.rand()` from the Crypto library)
- Affected paths: PKI key generation entropy, PKI nonce generation (extraNonce), and packet ID generation (used as nonce input for ALL channel encryption)
- Add `cryptoSecureRandom32()` helper function accessible from all platform builds, with graceful fallback to `random()` on minimal builds that exclude PKI
- Fix type mismatch: `extraNonceTmp` was `long` (sign-extended on 32-bit platforms) but written as `uint32_t` via memcpy

## Why this matters

Arduino `random()` uses a linear congruential generator (LCG) seeded from `analogRead()`. LCG output is trivially predictable after observing a few values. Since packet IDs feed directly into AES-CTR/CCM nonces, predictable packet IDs = predictable nonces = catastrophic for encryption security (nonce reuse enables plaintext recovery).

## Test plan

- [x] All `test_crypto` tests pass (SHA256, ECB, DH25519, AES-CTR, PKC round-trip)
- [x] All other native test suites pass (one pre-existing failure in `test_transmit_history` unrelated to this change)
- [ ] Verify on nRF52840 (RAK4631) that packet IDs are generated and encryption works normally
- [ ] Verify on ESP32 that `cryptoSecureRandom32()` links correctly

### DIFF
--- a/src/mesh/CryptoEngine.cpp
+++ b/src/mesh/CryptoEngine.cpp
@@ -29,7 +29,8 @@ void CryptoEngine::generateKeyPair(uint8_t *pubKey, uint8_t *privKey)
     if (myNodeInfo.device_id.size == 16) {
         CryptRNG.stir(myNodeInfo.device_id.bytes, myNodeInfo.device_id.size);
     }
-    auto noise = random();
+    uint32_t noise;
+    CryptRNG.rand((uint8_t *)&noise, sizeof(noise));
     CryptRNG.stir((uint8_t *)&noise, sizeof(noise));
 
     LOG_DEBUG("Generate Curve25519 keypair");
@@ -106,11 +107,12 @@ bool CryptoEngine::encryptCurve25519(uint32_t toNode, uint32_t fromNode, meshtas
                                      uint64_t packetNum, size_t numBytes, const uint8_t *bytes, uint8_t *bytesOut)
 {
     uint8_t *auth;
-    long extraNonceTmp = random();
+    uint32_t extraNonceTmp;
+    RNG.rand((uint8_t *)&extraNonceTmp, sizeof(extraNonceTmp));
     auth = bytesOut + numBytes;
     memcpy((uint8_t *)(auth + 8), &extraNonceTmp,
            sizeof(uint32_t)); // do not use dereference on potential non aligned pointers : *extraNonce = extraNonceTmp;
-    LOG_DEBUG("Random nonce value: %d", extraNonceTmp);
+    LOG_DEBUG("Random nonce value: %u", extraNonceTmp);
     if (remotePublic.size == 0) {
         LOG_DEBUG("Node %d or their public_key not found", toNode);
         return false;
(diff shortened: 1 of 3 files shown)


src/mesh/CryptoEngine.cpp
```diff
@@ -29,7 +29,8 @@ void CryptoEngine::generateKeyPair(uint8_t *pubKey, uint8_t *privKey)
     if (myNodeInfo.device_id.size == 16) {
         CryptRNG.stir(myNodeInfo.device_id.bytes, myNodeInfo.device_id.size);
     }
-    auto noise = random();
+    uint32_t noise;
+    CryptRNG.rand((uint8_t *)&noise, sizeof(noise));
```
jp-bennett, line 33: No, you clearly don't understand what this code is doing. CryptRNG is *not* hardware backed on all platforms, and this code pulls more randomness data from random() which *is* backed with hardware randomness on some platforms.

src/mesh/CryptoEngine.cpp
```diff
@@ -145,7 +147,11 @@ bool CryptoEngine::encryptCurve25519(uint32_t toNode, uint32_t fromNode, meshtas
 bool CryptoEngine::decryptCurve25519(uint32_t fromNode, meshtastic_UserLite_public_key_t remotePublic, uint64_t packetNum,
                                      size_t numBytes, const uint8_t *bytes, uint8_t *bytesOut)
 {
-    const uint8_t *auth = bytes + numBytes - 12; // set to last 8 bytes of text?
+    if (numBytes < 12) {
+        LOG_WARN("PKI decrypt: packet too small (%u bytes), need at least 12 for tag+nonce", numBytes);
+        return false;
+    }
+    const uint8_t *auth = bytes + numBytes - 12; // 8-byte CCM tag + 4-byte extraNonce
```
jp-bennett, line 154: Should be a different PR, and is probably dead code branch.

src/mesh/CryptoEngine.cpp
```diff
@@ -106,11 +107,12 @@ bool CryptoEngine::encryptCurve25519(uint32_t toNode, uint32_t fromNode, meshtas
                                      uint64_t packetNum, size_t numBytes, const uint8_t *bytes, uint8_t *bytesOut)
 {
     uint8_t *auth;
-    long extraNonceTmp = random();
+    uint32_t extraNonceTmp;
+    RNG.rand((uint8_t *)&extraNonceTmp, sizeof(extraNonceTmp));
```
jp-bennett, line 111: Last I checked, this collides with another object name on one of the platforms. STM32, I believe.

jp-bennett:
I think #8803 better represents how we want to address this. I'll take some time to fix the issues pointed out on that PR, and then reviews welcome.

PR #10060, opened 2026-04-02
fix: use constant-time comparison for admin key checks

Patrickschell609:
## Summary

- Add `constantTimeCompare()` utility function to `meshUtils.h/cpp` for timing-safe byte array comparison
- Replace 4 `memcmp()` calls in `AdminModule.cpp` that compare security-sensitive key material:
  - 3 admin public key comparisons (authorization gate for remote admin access)
  - 1 session passkey comparison (session authentication)

## Why this matters

`memcmp()` returns early on the first differing byte, leaking which bytes of the secret match via response timing. For admin key comparison, this means an attacker can brute-force the 32-byte admin public key one byte at a time (256 * 32 = 8,192 attempts) instead of brute-forcing the full key space.

While LoRa timing attacks are harder than TCP, an attacker with an SDR at close range can measure response latency with sub-millisecond precision. Admin key compromise grants full remote control: config changes, channel key extraction, firmware updates, and factory reset.

The new `constantTimeCompare()` uses XOR accumulation with `volatile` to prevent compiler optimization, matching the existing pattern in `aes-ccm.cpp` (which was `static` and not available to other modules).

## Test plan

- [x] All native test suites pass (151/151, one pre-existing flaky failure in transmit_history)
- [x] `test_admin_radio` passes (exercises admin module code paths)
- [ ] Verify on hardware that admin key authentication still works correctly

### DIFF
--- a/src/meshUtils.cpp
+++ b/src/meshUtils.cpp
@@ -79,6 +79,19 @@ bool memfll(const uint8_t *mem, uint8_t find, size_t numbytes)
     return true;
 }
 
+int constantTimeCompare(const uint8_t *a, const uint8_t *b, size_t len)
+{
+    if (len == 0)
+        return 0;
+    if (!a || !b)
+        return -1;
+    volatile uint8_t d = 0;
+    for (size_t i = 0; i < len; i++) {
+        d |= (a[i] ^ b[i]);
+    }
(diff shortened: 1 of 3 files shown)


Xaositek:
Complete duplicate of https://github.com/meshtastic/firmware/pull/9888

PR #10065, opened 2026-04-03
fix: stop logging cryptographic material and credentials

Patrickschell609:
## Summary
- **Private key** logged in cleartext on config restore (`NodeDB.cpp:648`)
- **Shared encryption key** (first 8 bytes) and nonce logged on every PKI encrypt/decrypt (`CryptoEngine.cpp:125-126, 166-167`)
- **Session admin passkey** logged on set and verify (`AdminModule.cpp:1470, 1477-1478`)
- **Public keys** (full 32 bytes) logged on node discovery (`NodeDB.cpp:1866, 1906`)

All `printBytes()` calls dumping key material replaced with safe `LOG_DEBUG` messages that confirm the operation occurred without exposing sensitive bytes. Also fixes an `sprintf` -> `snprintf` in the duplicate key warning path.

## Test plan
- [ ] Verify PKI encrypt/decrypt still works (no functional change, only log output)
- [ ] Verify admin session passkey flow still works
- [ ] Verify node discovery still processes public keys
- [ ] Check serial output no longer contains raw key bytes
- [ ] All 152 native tests pass


jp-bennett (approved):
This stuff was very useful early in the lifetime of adding these features, but it's good hygiene to no longer print the key material.

fifieldt (approved):

fifieldt:

PR #1594, opened 2022-08-04
Littlefs rename fix

caveman99:
fix for long garbage collection runs in littlfs below v2.5

PR #10072, opened 2026-04-03
Remove copyFile/renameFile reimplementations, adjust STM32 SafeFile to match NRF52

Stary2001:
Use the built-in rename, or none instead.
Temporarily added for NRF52 in #1594, but we're not using it on NRF52 any more, and it caused problems for STM32.
Needs testing on RP2040.
Will supersede #9927 .

## 🤝 Attestations

  - [x] Wio-E5
  - [x] RP2040


einolto:
Between "develop" and this commit there is a hardfault.
//\ E S H T /\ S T / C
Version 2.7.21.cd06055 for wio-e5 from meshtastic/firmware
Debug mute is enabled, there will be no serial output.
HardFault!
r0: 00000000
r1: 000003e3
r2: 00000000
r3: 20010000
r12: 20001524
lr: 0802915f
pc[return address]: 0800fc68
xpsr: 21000000

Edit: copying the three files limited to this commit over the develop (2f19a1d7a421c67c2674e73732a7019f13156953) config save seems like working on STM32WL

thebentern (approved):

Stary2001:
Should have been fixed by PR https://github.com/meshtastic/firmware/pull/10171

PR #10080, opened 2026-04-04
Update RadioLib to 7.6.0

fifieldt:
Released February 20th.

Since we're soon going to have to upgrade RadioLib again to add in LR2021 support, let's first upgrade to the latest officially released version. This way we can test our existing modules and have a little bit less of a delta for the next jump.


fifieldt:
I didn't synch branch, this is already done last week.

PR #10120, opened 2026-04-10
Enable Narrow and Lite regions for EU

NomDeTom:
Some new presets for a wider range of operations in the EU, based on https://github.com/meshtastic/firmware/pull/9038 but with improvements to the position and telemetry scaling implementation.

Relies on https://github.com/meshtastic/protobufs/pull/824 for protobufs (as before)

## What has changed?
- 2 new region presets
  - EU_866, 
  - NARROW_868,
- 4 new modem presets
  - LITE_FAST/SLOW, 
  - NARROW_FAST/SLOW
- new region parameters that make use of default slots, etc.

## What is needed to integrate these changes more widely
- BaseUI / InkHud
  - The new regions and presets have been added to the menu lists, but the list logic has not been updated to exclude invalid combinations
- MUI
  - The new regions and presets need to be added to the menu lists, and the list logic included.
- Client Apps
  - These will need updating to take account of the valid region/preset combinations, ideally hiding invalid combinations within the menus
  - Their display logic will need updating to include the new overrideSlots information, which will affect both the lora config menu and the behaviour when new primary channels are generated

## Acknowledgements
This change has been supported extensively by @Stary2001, @phaseloop, and @caveman99. Their input cannot be understated.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


NomDeTom:
oops - forgot to put the modem settings in!

PR #10123, opened 2026-04-10
RemoteShell SSH-like experience over PKI

jp-bennett:
Still WIP, but actually working, and ready for some general sanity checks from others. does not work super well with a ch341 device, but you can use any Meshtastic device as the local node, and connect using the client script over USB, to connect to a remote Linux node.

LLM was heavily used to generate this. I feel like I kept it reigned on reasonable well in the c++ code (though I'm sure I'll make changes there before I'll completely happy). I'm much less certain about the Python code. Ideally this function can land in the Meshtastic python CLI, but for now, the standalone client is super useful for testing.


SpudGunMan:
Can there be a 2fa in some snazzy way

jp-bennett:
> Can there be a 2fa in some snazzy way 

We might be able to go to a system login instead of a shell. Then the Meshtastic node is the 2fa, and you're prompted for password.

jp-bennett:
@thebentern decent cleanup punchlist from copilot. I have a few other todo items to knock out before it's ready to merge.

PR #10142, opened 2026-04-12
Switch PlatformIO deps from PIO Registry to tagged GitHub zips

vidplace7:
Switch all dependencies defined in the base `platformio.ini` from PlatformIO Registry sources to GitHub tag sources.
This does not include any library upgrades. Simply a source change.

Should help further with PlatformIO Registry ratelimiting issues.

This should be backmerged to `develop` ASAFP after merge (the build containers are cache based on the state of `master`)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -131,107 +131,115 @@ lib_deps =
 ; Common libs for environmental measurements in telemetry module
 [environmental_base]
 lib_deps =
-	# renovate: datasource=custom.pio depName=Adafruit BusIO packageName=adafruit/library/Adafruit BusIO
-	adafruit/Adafruit BusIO@1.17.4
-	# renovate: datasource=custom.pio depName=Adafruit Unified Sensor packageName=adafruit/library/Adafruit Unified Sensor
-	adafruit/Adafruit Unified Sensor@1.1.15
-	# renovate: datasource=custom.pio depName=Adafruit BMP280 packageName=adafruit/library/Adafruit BMP280 Library
-	adafruit/Adafruit BMP280 Library@3.0.0
-	# renovate: datasource=custom.pio depName=Adafruit BMP085 packageName=adafruit/library/Adafruit BMP085 Library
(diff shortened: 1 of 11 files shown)


platformio.ini
```diff
@@ -131,107 +131,115 @@ lib_deps =
...
+	# renovate: datasource=github-tags depName=Adafruit BusIO packageName=adafruit/Adafruit_BusIO
+	https://github.com/adafruit/Adafruit_BusIO/archive/refs/tags/1.17.4.zip
+	# renovate: datasource=github-tags depName=Adafruit Unified Sensor packageName=adafruit/Adafruit_Sensor
+	https://github.com/adafruit/Adafruit_Sensor/archive/refs/tags/1.1.15.zip
+	# renovate: datasource=github-tags depName=Adafruit GFX packageName=adafruit/Adafruit-GFX-Library
+	https://github.com/adafruit/Adafruit-GFX-Library/archive/refs/tags/1.12.6.zip
+	# renovate: datasource=github-tags depName=NeoPixel packageName=adafruit/Adafruit_NeoPixel
+	https://github.com/adafruit/Adafruit_NeoPixel/archive/refs/tags/1.15.4.zip
+	# renovate: datasource=github-tags depName=Adafruit SSD1306 packageName=adafruit/Adafruit_SSD1306
+	https://github.com/adafruit/Adafruit_SSD1306/archive/refs/tags/2.5.16.zip
```
vidplace7, line 143: These 3 "new" dependencies were being unconditionally added by other Adafruit libs.
- Adafruit-GFX-Library
- Adafruit_NeoPixel
- Adafruit_SSD1306

Adding them here explicitly forces them to use github as the source, and makes our builds more reproducible.

Example:
https://github.com/adafruit/Adafruit_INA219/blob/2.1.0/library.properties#L10

Explicit dependencies on NeoPixel have been removed in other variants since they're redundant.

jp-bennett (approved):

PR #10143, opened 2026-04-12
fix(xmodem): truncate file on open instead of appending

benallfree:
This PR fixes what I perceive is a bug in the XModem module: subsequent file writes to the same path append data rather than truncating.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [x] LilyGo T-Beam
  - [x] RAK WisBlock 4631


thebentern (approved):

benallfree:
@thebentern Copilot brought up a good point about preserving atomicity by writing to a tmp file and then doing a rename. I updated the PR to reflect that.

benallfree:
@thebentern Sorry for the delay, for some reason I am not getting GH notifications on this PR.

PR #10168, opened 2026-04-15
pioarduino esp32c6: fix build due to LED_BUILTIN compile error.

cpatulea:
<command-line>: error: expected unqualified-id before '-' token /home/<snip>/.platformio/packages/framework-arduinoespressif32/variants/esp32c6/pins_arduino.h:9:22: note: in expansion of macro 'LED_BUILTIN'
    9 | static const uint8_t LED_BUILTIN = SOC_GPIO_PIN_COUNT + PIN_RGB_LED;
      |                      ^~~~~~~~~~~
<command-line>: error: expected unqualified-id before '-' token

Similar to commit dcfead5.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/variants/esp32c6/esp32c6.ini
+++ b/variants/esp32c6/esp32c6.ini
@@ -22,6 +22,7 @@ build_flags =
   -DHAS_BLUETOOTH=0
   -DMESHTASTIC_EXCLUDE_PAXCOUNTER
   -DMESHTASTIC_EXCLUDE_BLUETOOTH
+  -ULED_BUILTIN
 
 lib_deps =
   ${arduino_base.lib_deps}


mverch67:
This is likely to happen for _all_ devices that don't have their own boards/board.json definition because it will pickup the standard pins_arduino.h provided by platformio which _always_ contain a LED_BUILTIN definition.

The only solution for this is to either revert PR https://github.com/meshtastic/firmware/pull/8830 or to undefine -ULED_BUILTIN for these devices that are using platformio provided board definitions.

cpatulea:
@jp-bennett what was the intent behind https://github.com/meshtastic/firmware/pull/8830 explicitly setting `-DLED_BUILTIN=-1`? It causes build errors in multiple specific cases described above.

jp-bennett:
@cpatulea because *not* setting it at all also causes build failures. As far as I can tell, it's a placeholder that doesn't actually belong in libraries or board support files.

cpatulea:
pioarduino (base commit d9ee3d6e6):

```
$ git diff
diff --git a/platformio.ini b/platformio.ini
index 43cb1fcd1..473977c59 100644
--- a/platformio.ini
+++ b/platformio.ini
@@ -58,7 +58,7 @@ build_flags = -Wno-missing-field-initializers
        -DMESHTASTIC_EXCLUDE_POWERMON=1
        -DMESHTASTIC_EXCLUDE_STATUS=1
        -D MAX_THREADS=40 ; As we've split modules, we have more threads to manage
-       -DLED_BUILTIN=-1
+       #-DLED_BUILTIN=-1
        #-DBUILD_EPOCH=$UNIX_TIME ; set in platformio-custom.py now
        #-D OLED_PL=1
        #-D DEBUG_HEAP=1 ; uncomment to add free heap space / memory leak debugging logs
```

```
Environment    Status    Duration
-------------  --------  ------------
heltec-v3      SUCCESS   00:02:31.899
```

(heltec-v3 does *not* have a boards file)

develop (commit 0cab43fb)

```
$ git diff
diff --git a/platformio.ini b/platformio.ini
index 46170fe09..f0af061cb 100644
--- a/platformio.ini
+++ b/platformio.ini
@@ -58,7 +58,7 @@ build_flags = -Wno-missing-field-initializers
        -DMESHTASTIC_EXCLUDE_POWERMON=1
        -DMESHTASTIC_EXCLUDE_STATUS=1
        -D MAX_THREADS=40 ; As we've split modules, we have more threads to manage
-       -DLED_BUILTIN=-1
+       #-DLED_BUILTIN=-1
        #-DBUILD_EPOCH=$UNIX_TIME ; set in platformio-custom.py now
        #-D OLED_PL=1
        #-D DEBUG_HEAP=1 ; uncomment to add free heap space / memory leak debugging logs
```

```
Environment    Status    Duration
-------------  --------  ------------
heltec-v3      SUCCESS   00:03:24.639
```

[edit] Also tried develop branch `heltec-wireless-paper`, build successfully.

@jp-bennett is there a specific device that I should test?

cpatulea:
Tried to build at commit just before #8830 (commit 1d219a9), seems good?

```
Environment            Status    Duration
---------------------  --------  ------------
heltec-v3              SUCCESS   00:04:11.973
heltec-wireless-paper  SUCCESS   00:02:28.386
```

> The Arduino framework really wants this set, but then does nothing with it

I do see this text in the PR, but I can't find any specific place where Arduino framework needs this.

mverch67:
> I do see this text in the PR, but I can't find any specific place where Arduino framework needs this.

LED_BUILTIN is used in every arduino blink example. This is used to build and flash a simple example and see that the tools platform works together with the hardware without having to know the detailed pinout of that hardware. It just works to let the led blinking.

With #8830 I think we didn't win anything. There are exactly two groups of devices: those with platformio provided boards definition and those without (or those that required adaptations). Those without usually copy the original pins_arduino.h (if available) into the variant folder. Usually these can be used right away as they are without changes.

Now, #8830 introduced additional effort: 
- For boards with own board definition the pins_arduino.h cannot be used as it is, because the LED_BUILTIN definition has to be removed (as has been done in #8830 for many boards)
- For boards without own definition the LED_BUILTIN has duplicated defines. It requires to add -ULED_BUILTIN into every target of platformio.ini for that variant.

So overall I don't see any real benefit of #8830, it just adds a bit of confusion because it requires non-standard pins_arduino.h files.

Just my two cents.

jp-bennett:
Let's watch the CI here and on https://github.com/meshtastic/firmware/pull/10179 to see how this idea shakes out.

The reality is that LED_BUILTIN is an upstream mess, and we're going to have to deal with it one way or another. My only ask is that we don't add the defines back into variant.h files.

mverch67:
> My only ask is that we don't add the defines back into variant.h files.

The LED_BUILTIN define belongs to pins_arduino.h, not variant.h. Otherwise the define won't be used when external arduino files are built (except when pins_arduino.h includes variant.h, but that would be non-standard as well). 

If the board does not have a LED then LED_BUILTIN can be set to -1 in file pins_arduino.h. If there are existing external(!) pins_arduino.h files somewhere in a toolchain that don't define LED_BUILTIN then that toolchain probably does not make use of LED_BUILTIN (anymore).


Edit: #https://github.com/meshtastic/firmware/pull/10179 reveals that LED_BUILTIN is only used in nrf52 toolchain, so I think #8830 can be safely reverted and the -DLED_BUILTIN=-1 can be inserted into nrf52.ini which does not clash with the existing pins_arduino.h files that are defined for esp32 devices.

jp-bennett:
I went ahead and added -DLED_BUILTIN=-1 to the nrf .ini in 10179. If that builds cleanly, I'm 100% on board with merging it. Thanks @cpatulea for wrestling us into taking another look at this.

jp-bennett:
Merged that one, I think this one can be closed.

cpatulea:
Great! Thanks for the collaboration.

PR #10169, opened 2026-04-15
refactor(nodedb): stable-slot NodeDB with streamed persistence and backend-aware storage with increased node counts.

h3lix1:
# NodeDB Refactor

This takes NodeDB off the old "one vector does everything" setup and turns it into a layered model:

- dense storage slots for full `meshtastic_NodeInfoLite` records
- a compact `NodeMeta` cache for hot reads
- a private `NodeNum -> meta index` hash
- a separate `displayOrder` view for UI and exports
- one backend seam so the same NodeDB logic can run against RAM/PSRAM or flash

The point was not to get fancy. The point was to stop mixing storage order, display order, lookup, persistence, and pointer lifetime into one container. Also, to increase nodedb size without breaking the memory budget.

## AI Asistance disclosure

A lot of this was assisted with AI. This would have taken way too long otherwise, with entirely too many bugs. It is absolutely no skin off by back if this is not accepted due to this, but I think it deserves a fair shot.

## Why this version

I tried pretty hard to keep the final shape boring on purpose.

There are a bunch of more elaborate ways to build a large NodeDB: packed hot indexes, cuckoo hashing, generalized write-behind buffers, deeper scratch-object APIs, and so on. Some of that is interesting, but I wanted to keep this code flow simple. (unlike that guy who wrote the TMM.. oh wait)

Instead this is the smallest version that still fixes nodedb size (and memory) problems:

- sorting no longer rewrites full records
- save/load no longer stages the full DB in RAM
- callers that keep long-lived pointers got pushed toward `NodeNum` first
- flash-backed boards get a real storage path instead of pretending they can carry the whole record set the same way PSRAM boards can

That makes the system easier to extend later if we actually need something fancier.

## The model

```mermaid
graph TB
    subgraph NodeDB
        META[NodeMeta array]
        HASH[NodeNum to meta index hash]
        ORDER[displayOrder array]
    end

    subgraph Full Records
        STORE[NodeStore]
        ARRAY[ArraySlotStore]
        FLASH[FlashSlotStore]
    end

    HASH --> META
    ORDER --> META
    META --> STORE
    STORE --> ARRAY
    STORE --> FLASH
```

The practical split is:

- `NodeMeta` is the cheap hot path for lookup, sorting, flags, counts, and routing-adjacent reads
- the active `NodeStore` holds the full protobuf-backed record bodies
- `displayOrder` is the presentation view
- dense storage order stays internal to NodeDB

### 1. storage and presentation are no longer the same thing

`sortMeshDB()` now sorts `displayOrder[]` instead of swapping `NodeInfoLite` records around.

That was one of the bigger behavioral fixes. The old model made it too easy for a cached pointer to keep pointing at "some slot in the vector" while the logical identity of that slot had changed under it.

### 2. hot metadata is separated from cold record bodies

Each live node gets a compact `NodeMeta` entry with:

- node number
- last heard
- storage index
- quantized SNR
- channel and hops data
- frequently checked flags
- role and hardware model

That is enough for most NodeDB decisions without reopening the full record.

### 3. lookup moved to a private open-addressed hash

The hash is simple linear probing.

- it stays private to NodeDB
- it is rebuilt in one pass
- it is easy to reason about in code review
- it is plenty for the current load factor

If this ever shows as slow, it is easy to replace without touching the rest of the design.

### 4. persistence got split by backend

`ArraySlotStore` keeps the in-memory path and uses streamed `nodes.proto` save/load.

`FlashSlotStore` keeps fixed-width page files and fixed-width records with protobuf payloads inside them.

The key thing there is that flash records store protobuf bytes, not raw struct bytes. That keeps the durable format stable even if compiler layout details change.

### 5. flash saves stopped rewriting the world on every change

The flash-backed path now has:

- dirty tracking by dense storage index
- a full-rewrite flag for layout changes
- page-level batching so one save rewrites each touched page once

That still keeps the logic close to before with nodedb flushing:

- full rewrite when layout changed or migration/normalization is needed
- incremental page-batched flush when only existing live slots changed

No persistent cache, no write-behind thread, no invisible deferred state.

## Boot and save flow

```mermaid
flowchart TD
    START[start]
    LEGACY{valid nodes.proto?}
    FLASHPREF{flash-preferred target?}
    FLASH{valid flash manifest?}
    ARRAYBOOT[load streamed protobuf]
    MIGRATE[migrate to flash and verify]
    FLASHBOOT[scan flash slots into NodeMeta]
    DEFAULTS[install defaults]
    READY[continue boot]

    START --> LEGACY
    LEGACY -->|yes| ARRAYBOOT
    ARRAYBOOT --> FLASHPREF
    FLASHPREF -->|yes| MIGRATE
    FLASHPREF -->|no| READY
    MIGRATE --> READY
    LEGACY -->|no| FLASHPREF
    FLASHPREF -->|yes| FLASH
    FLASH -->|yes| FLASHBOOT
    FLASHBOOT --> READY
    FLASH -->|no| DEFAULTS --> READY
```

On flash-backed saves, the flow is:

```mermaid
flowchart TD
    SAVE[start save]
    FULL{layout changed or migration?}
    REWRITE[rewrite dense live slots by page]
    DIRTY[collect dirty storage slots]
    GROUP[group by flash page]
    PATCH[patch one page image in memory]
    VERIFY[verify page or full store]
    CLEAR[clear dirty bits]

    SAVE --> FULL
    FULL -->|yes| REWRITE --> VERIFY --> CLEAR
    FULL -->|no| DIRTY --> GROUP --> PATCH --> VERIFY --> CLEAR
```

The page batching matters to keep write churn low.

## Backend choice

Right now the split is straightforward:

- `ESP32-S3` with PSRAM prefers `ArraySlotStore`
- `nRF52` and no-PSRAM `ESP32` family targets prefer `FlashSlotStore`
- `native` and `Portduino` stay on the array-backed path unless a build overrides the cap

The landed target caps are:

| target class                       | backend | target cap |
| ---------------------------------- | ------- | ---------: |
| `ESP32-S3` with PSRAM              | array   |       3000 |
| `ESP32` / `ESP32-S3` without PSRAM | flash   |        500 |
| `nRF52`                            | flash   |        100 |

On PSRAM builds, the runtime store still clamps that target based on actual boot-time headroom. So `NODEDB_TARGET_CAP` is the build upper bound, but it can be less.

### Simple hash over clever hash

The cuckoo-hash used in TMM, while works, was more complex than it needed to be for this purpose. The current linear-probe table is easier to inspect and easier to trust, and I don't expect it to be the bottleneck.

### Protobuf payloads over raw record dumps

Flash records keep encoded protobuf bytes. That costs a little encode/decode work, but we're already using protobufs for everything else.

### Dirty-slot batching over generalized caching

The flash optimization stops at "only rewrite what changed, and only touch each page once per save." That gets most of the value without adding another layer of state to debug.

### Compatibility bridge

`meshNodes` still exists as a compatibility alias, and `slotAt()` / `slotPtr()` still feed pointer-based callers.

The surrounding codebase used meshNodes a lot, so the storage rewrite keeps the bridge local to NodeDB instead of forcing a giant caller rewrite in the same diff. Maybe in future.

## Side cleanup that made this safer

Before the backend work, there were a few pieces of cleanup that mattered more than they might look at first glance:

- screen frame allocation got decoupled from `MAX_NUM_NODES`
- packet history stopped scaling from NodeDB size
- favorite / ignore / mute / key verification writes moved behind NodeDB mutators
- several UI/module pointer caches moved from `NodeInfoLite *` to `NodeNum`

Those changes reduced the number of hidden assumptions the storage rewrite had to carry.

## User-visible behavior

Not much. The visible outcomes are mostly about the system not painting itself into a corner anymore:

- higher practical node caps on hardware that can handle them
- sane flash-backed persistence on boards that cannot
- node lists still sort the same way people expect
- device displays now show active and total nodedb node counts
- longer connection times, esp. for bluetooth.

Also, the screen now reads `<online>/<total>` for nodes.

## Review notes

If I were reviewing this cold, I would look at it in this order:

1. `NodeDB` storage invariants
2. `NodeMeta` rebuild/refresh rules
3. `displayOrder` behavior after sort, remove, reset, and eviction
4. flash boot scan and lazy hydration
5. flash full rewrite vs incremental dirty flush
6. migration rule that verifies the destination before deleting legacy state

That is where the actual correctness lives.

## What this does not try to do

- It does not add a persistent page cache.
- It does not hide write-behind behavior in the background.
- It does not move unrelated subsystems into PSRAM just because NodeDB changed.
- It does not pretend every caller is already cleanly handle-based.

This is a storage-model cleanup first. If later profiling shows it needs to be more efficient, it will hopefully be easier.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Heltec V4
Heltec T114


h3lix1:
The anticipation is that with something like this, TMM will no longer need its own cache and DB. Everything hot will stay in memory, everything elese is put on storage.

PR #10179, opened 2026-04-16
Test PR to remove LED_BUILTIN

jp-bennett:
Do not merge, simply to demonstrate how many targets break without this define.


mverch67:
Interesting experiment. It looks like the LED_BUILTIN is only required in some nrf52 toolchain file (packages/framework-arduinoadafruitnrf52/libraries/InternalFileSytem/src/flash/flash_cache.c) and not for any esp32 devices.

jp-bennett:
@mverch67 yeah, looks like it. I'll wait for CI to fully complete, but this suggests an even cleaner solution, of adding the define in the nRF .in file.

mverch67 (approved):

PR #10205, opened 2026-04-19
Fix LED state handling in StatusLEDModule to clarify power state condition

### LINKED ISSUE #9912: [Bug]: T-Beam Supreme LED stuck on in 2.7.20.6658ec2 Alpha
### Category

Other

### Hardware

T-Beam S3

- [ ] InkHUD ePaper

### Firmware Version

2.7.20.6658ec2 Alpha

### Description

On 2.7.20.6658ec2 Alpha there's a blue LED next to the T-Beam Supreme's display panel that is stuck on since boot. Problem is new in this alpha build, wasn't present in prior builds.

### Relevant log output

```Shell

```

### LINKED ISSUE #10170: [Bug]: TBeam Blue LED stays ON forever
### Category

Other

### Hardware

T-Beam

- [ ] InkHUD ePaper

### Firmware Version

2.7.22

### Description

After installing 2.7.22 using the latest device-install.sh, the blue LED behind the display of the TBeam stays ON and can't be disabled. Before, i was running 2.7.10 and did not observe this behaviour.
(For the installation of 2.7.10 i used an older version of device-install.sh which did not work anymore for installing 2.7.22 Could this have to do with the issue?)

### Relevant log output

```Shell

```

thebentern:
Users have been confused with the new solid LED behavior for fully charged devices as it looks like things are "stuck" where prior to this, they would use the periodic blinking behavior. 
Closes #10170
Closes #9912

Original regression in #9512


jp-bennett:
I actually really like the behavior to indicate fully charged via the LED. Would like to find a way to keep it in some form. Probably also need to first figure out of there's a bug.

Edit: maybe use the solid LED for charged only if we have a plugged-in detect mechanism. It's probably pretty non-ideal on a device that can't detect when power is unplugged, staying lit until the battery drops below 100%

jp-bennett:
The t-beam is broken because `#define LED_STATE_ON 0 // State when LED is lit`
Fix incoming.

Jckf:
Adding my vote for this. Many nodes are permanently USB powered, so I consider the new behavior broken.

george-black:
I have a Heltec v4 that is constantly plugged into USB power.  The white LED stays on solid in recent versions (past 2.7.19, i think) -- and I miss the blink.  This node will never have a battery and never be charging, so the steady-on state of the LED loses the heartbeat feature for me.

So I support this PR!  👍

jp-bennett:
Another possible solution, only do the solid LED when we know we have a battery.

george-black:
FYI this issue persists for me in 2.7.24 despite https://github.com/meshtastic/firmware/pull/10423 being closed

PR #10216, opened 2026-04-20
Meshtasticd notifications

jp-bennett:
This pull request introduces support for desktop notifications using `libnotify` on native (Portduino) builds, and updates packaging and dependencies to ensure `libnotify` is available across all supported environments. It also adds a desktop icon for Meshtasticd and updates several development and CI dependencies.

**Desktop notification support:**

* Added `libnotify` integration for native/Portduino builds, enabling Meshtasticd to show desktop notifications when messages are received. This includes new code in `ExternalNotificationModule` to generate notifications, and updates to `platformio.ini` to enable and link against `libnotify`. [[1]](diffhunk://#diff-8f0f833122f0577cc7d091d7954100cba2e48c1208d118f5830f68417eace402R27-R30) [[2]](diffhunk://#diff-8f0f833122f0577cc7d091d7954100cba2e48c1208d118f5830f68417eace402R415-R417) [[3]](diffhunk://#diff-8f0f833122f0577cc7d091d7954100cba2e48c1208d118f5830f68417eace402R531-R567) [[4]](diffhunk://#diff-9ff83e2a3de8cfc6dbf0dfb19a3242ee3058939b69c33d00ec005e5c4ba33c96R80-R83) [[5]](diffhunk://#diff-0ec8c666cca735920304d5f9d3e6ccd013c1b679b120339e3d85edca11d12d25R5-R6) [[6]](diffhunk://#diff-0ec8c666cca735920304d5f9d3e6ccd013c1b679b120339e3d85edca11d12d25L30-R33) [[7]](diffhunk://#diff-0ec8c666cca735920304d5f9d3e6ccd013c1b679b120339e3d85edca11d12d25L58-R62) [[8]](diffhunk://#diff-0ec8c666cca735920304d5f9d3e6ccd013c1b679b120339e3d85edca11d12d25L87-R92) [[9]](diffhunk://#diff-0ec8c666cca735920304d5f9d3e6ccd013c1b679b120339e3d85edca11d12d25L115-R122) [[10]](diffhunk://#diff-d3ca504aa401bdfaa42fcd99349d867b504756a86721afbbb46482909e8b1001R866-R869)

**Dependency and packaging updates:**

* Added `libnotify-dev` (and runtime equivalents) to all Dockerfiles, build scripts, and package specs to ensure the notification feature is available in all environments (Debian, Alpine, Fedora, CI). [[1]](diffhunk://#diff-13bd9d7a30bf46656bc81f1ad5b908a627f9247be3f7d76df862b0578b534fc6R37) [[2]](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L17-R17) [[3]](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L56-R56) [[4]](diffhunk://#diff-5faf10fde2b24acdf4f05e70be1be9d50921d685e64f781d242e056b11d79dccL12-R12) [[5]](diffhunk://#diff-0d11c9650bb11fa0a8bec94254d8b1860290b241e626c237c7d5f4a6c5a35719L14-R14) [[6]](diffhunk://#diff-0d11c9650bb11fa0a8bec94254d8b1860290b241e626c237c7d5f4a6c5a35719L45-R45) [[7]](diffhunk://#diff-2a13f1344504379e877324d3a0375adcbcb4cb27d7e575ad8f4618a52d6a00e0L30-R31) [[8]](diffhunk://#diff-b2e8a475cd83d0d5af11cbc1ef5baa64c173c8f0ffb15f6bea6e71555b790bc3L14-R14) [[9]](diffhunk://#diff-a4a56c120d0b650654383df6618c355ac387bc78c01da639cd5e2512dcd38dc0R53)

* Updated Fedora/RPM and Debian packaging to include `libnotify` as a build dependency. [[1]](diffhunk://#diff-a4a56c120d0b650654383df6618c355ac387bc78c01da639cd5e2512dcd38dc0R53) [[2]](diffhunk://#diff-2a13f1344504379e877324d3a0375adcbcb4cb27d7e575ad8f4618a52d6a00e0L30-R31)

**Desktop integration:**

* Added installation of a scalable SVG desktop icon for Meshtasticd in both Debian and RPM packaging. [[1]](diffhunk://#diff-12cd30243b932906e261fc337cd7992e59f71f7a6e3e6e974f61bb34c1b9e513R10-R11) [[2]](diffhunk://#diff-a4a56c120d0b650654383df6618c355ac387bc78c01da639cd5e2512dcd38dc0R116-R119) [[3]](diffhunk://#diff-a4a56c120d0b650654383df6618c355ac387bc78c01da639cd5e2512dcd38dc0R179)

## Summary by CodeRabbit

* **New Features**
  * Native builds can display desktop notifications for alerts, including sender and message details.
  * Linux packages now include the application icon.
  * Portduino builds enable external notifications and alert messages by default when supported.

* **Bug Fixes**
  * Notification support now works across additional Linux build and runtime environments.
  * Notification delivery remains resilient when desktop notification services are unavailable, with automatic recovery when service availability returns.


caveman99 (approved):

fifieldt (approved):

src/modules/ExternalNotificationModule.cpp
```diff
@@ -631,4 +638,108 @@ int ExternalNotificationModule::handleInputEvent(const InputEvent *event)
     }
     return 0;
 }
+
+#if HAS_LIBNOTIFY
```
jp-bennett, line 649: **Blocking — the missing `#endif` goes here.**

This `#if HAS_LIBNOTIFY` was inserted *before* the `#endif` that used to close `#if !MESHTASTIC_EXCLUDE_INPUTBROKER` (line 630). The result is that the file's trailing `#endif` (line 745) closes `HAS_LIBNOTIFY`, and the input-broker conditional is never closed — so the translation unit fails to preprocess on every target.

```suggestion
#endif

#if HAS_LIBNOTIFY
```

Verified: with that `#endif` added, `gcc -E -P` on the file exits 0; without it, it reports `unterminated #if` at line 630, which is exactly what CI reports.

Note that adding the `#endif` at the bottom of the file instead would balance the count but nest the entire libnotify implementation inside `!MESHTASTIC_EXCLUDE_INPUTBROKER`, which isn't the intent.

---

src/modules/ExternalNotificationModule.cpp
```diff
@@ -631,4 +638,108 @@ int ExternalNotificationModule::handleInputEvent(const InputEvent *event)
...
+            }
+            g_object_unref(G_OBJECT(notification));
+        } else {
+            LOG_WARN("Failed to create notification");
+        }
+        lock.lock();
+
+        consecutiveFailures = failed ? consecutiveFailures + 1 : 0;
+        if (consecutiveFailures >= maxNotifyFailures) {
+            LOG_WARN("Disabling desktop notifications after %d consecutive failures", consecutiveFailures);
```
jp-bennett: **`LOG_WARN` from the worker thread races the main loop.**

All four warnings in this range (687, 714, 720, 726) execute on the notify thread. `RedirectablePrint::vprintf` formats into a shared, unguarded `static char printBuf[512]` (`src/RedirectablePrint.cpp:55`) and then writes to the shared `Print` stream / logfile. Nothing else in meshtasticd is concurrent, so there is no lock protecting any of it.

It's a bit ironic given the header comment on this block correctly explains that the thread exists so `handleReceived()` never blocks — logging is the one thing the worker does that reaches back into main-thread state.

Either stash a failure code in an atomic and log it from `runOnce()`, or use a bare `::fprintf(stderr, ...)` here.

(The `LOG_WARN` on line 674 is fine — that one runs on the caller's thread.)

---

src/mesh/NodeDB.cpp
```diff
@@ -1447,6 +1447,13 @@ void NodeDB::installDefaultModuleConfig()
     moduleConfig.external_notification.alert_message_buzzer = true;
 #endif // HAS_I2S
 
+#if ARCH_PORTDUINO
```
jp-bennett: **Minor — this gates on `ARCH_PORTDUINO`, but the comment describes `HAS_LIBNOTIFY`.**

Native builds where `pkg-config` didn't find libnotify still get `external_notification.enabled = true` for a module that can't do anything. The comment right below already states the intent ("the module is what raises desktop notifications"), so `#if HAS_LIBNOTIFY` matches it.

To be clear this is config noise, not a hardware hazard: on native `EXT_NOTIFICATION_MODULE_OUTPUT` is `0` (`ExternalNotificationModule.cpp:62`) and `setup()` guards `pinMode`/`digitalWrite` behind `if (output > 0)`, so no GPIO gets driven.

---

jp-bennett (commented):
Review of the libnotify path at `38e1ea8`. Every item below was confirmed by running something, not by reading alone.

## Blocking: this doesn't compile on any target

The new `#if HAS_LIBNOTIFY` block in `ExternalNotificationModule.cpp` was inserted before the `#endif` that closed `#if !MESHTASTIC_EXCLUDE_INPUTBROKER`, leaving that conditional open. CI is red on all ~25 build jobs for exactly this:

```
src/modules/ExternalNotificationModule.cpp:630: error: unterminated #if
  630 | #if !MESHTASTIC_EXCLUDE_INPUTBROKER
*** [.pio/build/rak3172/src/modules/ExternalNotificationModule.cpp.o] Error 1
```

One-line fix, inline on line 642. `origin/develop` balances at depth 0; this head ends at depth 1.

## Should fix

**1. `LOG_WARN` from the worker thread races the main loop.** Inline on 687–726. The worker is the one place that reaches back into unguarded main-thread state (`RedirectablePrint`'s shared `static printBuf`).

**2. The permanent latch-off will trip on the deployment that actually ships.** After three failed shows, `notifyDisabled` latches for the process lifetime with no re-arm — and the packaged daemon is exactly the case that hits it. `bin/meshtasticd.service` sets `User=meshtasticd` / `Group=meshtasticd`, and the rpm spec creates that account with `-s /sbin/nologin`; there's no `DBUS_SESSION_BUS_ADDRESS`, so `notify_notification_show()` fails. A stock deb/rpm install therefore logs three warnings and then silently does nothing forever — while every install now carries the enabled-by-default config from the `NodeDB` change. A desktop user restarting their notification daemon has the same one-way effect.

I realize the latch was added in response to [#discussion_r3114144959](https://github.com/meshtastic/firmware/pull/10216#discussion_r3114144959) (don't retry and spam per message), which was the right call — it's just overshot to permanent. Backoff-and-retry would satisfy both, and/or gate the default-on on a session bus actually being reachable.

## Minor

1. **`NodeDB.cpp:1450`** gates on `ARCH_PORTDUINO` rather than `HAS_LIBNOTIFY` — inline.

2. **Mesh payload bytes reach libnotify unvalidated.** `notificationBody` is raw bytes off the wire. Checked against GLib 2.80: this does *not* crash — `g_variant_new_string()` returns NULL on invalid UTF-8 with `GLib-CRITICAL ... assertion 'g_utf8_validate (string, -1, NULL)' failed`, and `g_variant_new("(s)", bad)` substitutes the literal body `[Invalid UTF-8]`. So the real cost is one GLib CRITICAL on stderr per malformed message, a useless notification body, and an `abort()` under `G_DEBUG=fatal-criticals`. An embedded NUL also truncates the body at `.c_str()`. `g_utf8_make_valid()` plus stripping NULs covers both. (Distinct from the still-open thread about gating on `TEXT_MESSAGE_APP` — that one's about *which* packets get here, this is about the bytes themselves.)

3. **Body isn't markup-escaped.** GNOME Shell and other daemons parse limited Pango markup in the body, so a message containing `<` or `&` renders wrong or gets dropped. `g_markup_escape_text()`.

4. **`variants/native/portduino/platformio.ini`**: `[env:native-macos]` and `[env:native-windows]` both `extends = native_base`, so they inherit the libnotify probe. A Mac with Homebrew libnotify will compile the feature in where it can never work. Harmless, but the probe may belong in the Linux envs only.

5. **`notify_uninit()` is never called.** Trivial.

## What's good

The threading shape is right: sender and body are resolved on the caller's thread so `nodeDB` is only touched there, the mutex is dropped across the synchronous D-Bus round trip, the queue is bounded, the thread starts lazily, and the destructor joins. The `!pkg-config ... || :` probe follows the existing convention in that file, including the `;`-comment-inside-`build_flags` style already used by `[env:native]` — that parses fine under PlatformIO's `ConfigParser(inline_comment_prefixes)`. Packaging is updated consistently across `.devcontainer/Dockerfile`, `Dockerfile`, `Dockerfile.test`, `alpine.Dockerfile`, `debian/control`, `debian/meshtasticd.install`, `meshtasticd.spec.rpkg` and `.github/actions/setup-native`, and `bin/org.meshtastic.meshtasticd.svg` already exists in develop, so the new install lines resolve.

---

PR #10224, opened 2026-04-21
feat(radar): add radar minimap overlay to the Distance/Bearings frame

egorsiniaev:
## What this PR does

Adds a **radar minimap view** as an alternative rendering of the existing Distance/Bearings frame — no new screen slot in the rotation.

From the node-list long-press menu, a new **"Tracking View"** picker (clock-face style) lets the user switch between the regular Bearings list and the Radar overlay. While in radar mode, long-pressing opens a radar-specific menu (Tracking View, N-UP/HDG-UP toggle, Favorites Only toggle, Zoom In/Out). Both choices are **persistent** across reboots (`bearings_view_radar` and `radar_favorites_only` in `DeviceUIConfig`, proto fields 20 and 21).

### Radar layout

- **Left panel**: up to 5 closest nodes — per-row marker symbol, short name, distance
- **Right panel**: circular minimap with 3 range rings and 2 px padding so the circle never touches the display edge
- **Header**: shows the current outer-ring range as `Radar 5km` / `Radar 1.5mi` (units follow `display.units`)
- Own position is a single pixel at the centre
- Five unique markers (■ + × □ ◆) — sort-position-based, so the dot on the radar always matches the row in the list
- **Heading-up** when IMU (BMX160/RAK12034) is present; falls back to north-up
- **Favorites-only filter** for use at large gatherings — toggled from the radar long-press menu
- **No GPS fix** state shows a clear message instead of an empty radar
- No impact on e-ink devices — the entire bearings/distance frame this view lives in is already inside a `#ifndef USE_EINK` block in `NodeListRenderer.cpp`, so the radar code isn't compiled in for e-ink builds

### Files changed

- `src/graphics/draw/RadarRenderer.cpp/.h` — radar overlay renderer; owns its own header so the title can carry the current range
- `src/graphics/draw/NodeListRenderer.cpp` — delegates to `RadarRenderer::drawRadarOverlay` when `bearings_view_radar` is set
- `src/graphics/draw/MenuHandler.cpp/.h` — adds `trackingViewPicker()` and `radarBearingsMenu()`; wires "Tracking View" into the node-list menu
- `src/graphics/Screen.cpp` — routes long-press on the bearings frame to `radarBearingsMenu` when radar is active
- `src/mesh/generated/meshtastic/device_ui.pb.h` — adds `bool bearings_view_radar` (field 20) and `bool radar_favorites_only` (field 21), both backwards-compatible

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [x] lilygo lora32 v2.1
  - [x] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [x] RAK WisBlock 4631

**Note:** Happy to get help testing on other hardware from the community.

image

Video: https://photos.app.goo.gl/Tj3y2PKypFPwczE58

## Summary by CodeRabbit

* **New Features**
  * Added a radar-style location overlay with north-up vs heading-up display, zoom in/out, favorites-only filtering, and a side list of active nearby nodes.
  * Added a “Tracking View” chooser to switch between bearings and radar, with consistent Back navigation.
* **Bug Fixes**
  * Improved menu routing so the selected tracking mode opens correctly from the nodelist/location screens (GPS-aware option included).
* **Chores**
  * Standardized the mesh node online/offline timeout threshold constant used for node status decisions.


egorsiniaev:
> [CLA assistant check](https://cla-assistant.io/meshtastic/firmware?pullRequest=10224) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/meshtastic/firmware?pullRequest=10224) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/meshtastic/firmware?pullRequest=10224) it.

I have 500 Internal Server Error after signing it.

caveman99:
We have 3 nodelist displays with different additonal things. This adds a 4th list with a graphical representation. Also overlaps with the functionality of the GPS Compass. Maybe we can have this panel as an option instead of the GPS Compass or even merge it with the compass.

caveman99:
> I have 500 Internal Server Error after signing it.

The CLA server seems to have issues. Can you try again maybe tomorrow?

egorsiniaev:
> We have 3 nodelist displays with different additonal things. This adds a 4th list with a graphical representation. Also overlaps with the functionality of the GPS Compass. Maybe we can have this panel as an option instead of the GPS Compass or even merge it with the compass.

I thought about it, but because this is my first PR and contribution - to suggest such merge without long term community vision is a strange step. That's why I decided to have clear separate section that easy to review + easy to disable if people don't need it.

caveman99:
Problem is, we rather would like to make the interface more clear, not pile on it. I see the value in your contribution, i just think i t would really benefit the existing screens rather than open up a new one.

fifieldt:
This is cool :) I want to see how it looks on a larger screen, like the sensecap indicator - will try when I get time.

Also, appreciate your collaborative approach.

egorsiniaev:
> Problem is, we rather would like to make the interface more clear, not pile on it. I see the value in your contribution, i just think i t would really benefit the existing screens rather than open up a new one.

I updated. Removed separate Radar view and add option to switch Compass mode Radar mode. Plus changed a bit orientation (nodes on the left, radar on the right) to match compass layout. 

Please take a look. Should I need to update screenshots? They are still showing old layout.

egorsiniaev:
> This is cool :) I want to see how it looks on a larger screen, like the sensecap indicator - will try when I get time.
> 
> Also, appreciate your collaborative approach.

Will appreciate for checking it. I have no such devices yet.

egorsiniaev:
Hey @caveman99 Thanks for your suggestion. I'v made changes and pushed them. Today made photos and record video for better representation. In short - I extended Compass, add long-press menu to switch to the radar view + changed a bit layout to match compass. Please take a look and let me know if anything else is missing.

egorsiniaev:
Hey @caveman99 @fifieldt 
Could you please tell me - how to move forward with this PR? There are no reviewers added automatically. It's waiting about 2 weeks. Maybe I missed something to do or somebody need to approve CI to start checking changes. 
Thank you.

Xaositek:
Reviewing the video - I do like the idea and concept. It's well executed and provides a display not so dissimilar to what is in InkHUD. Need to load it up on a couple devices to give it a run through.

Xaositek (changes requested):
Tested it out on my Cardputer and I can't see it actually working. I tried Zooming In and Out, verified I have a GPS lock, and nodes are pretty close by.

I feel like we need some refinements like "What is my current Zoom level?", "Do I have GPS lock"?

Also the toggle between Compass and Radar likely needs to be similar to what we have on the Clock screen where you click on "Clock Face" and then choose Digital or Analog.

Xaositek:
https://github.com/user-attachments/assets/ef43559a-935c-44b4-9696-a7aff080a08a

Video of my attempt

HarukiToreda:
I had a similar idea while creating the distance and bearing screen, I ultimately opted to not implement due to the following:

- Adds another screen frame into an already extensive list of frames users have to cycle through, this was prior to the action menus (I see you solved this nicely using action menu but I would suggest making it swap the distance and bearing screen instead of the position screen since the position screen is not related to tracking nodes)

- Showing the closest random nodes does not add much value other than a wow factor, something the position and bearing screen already do with arrows and distance markers, (However if you can limit it to show favorite nodes only, now that adds a lot of value for people since that is a deliberate action useful for large events and gatherings "Favorite Radar View")

- Markers to differentiate the nodes was very limited (I see from the video you're reusing the x for the first 3, would you be able to make them different?)

egorsiniaev:
Hey @Xaositek @HarukiToreda - big thank you for review and feedback! Really appreciate! I've pushed an update to address it. Also rebased onto the latest develop + added new photo and video.

@HarukiToreda - all three of your suggestions are in:
- Radar now replaces the bearings/distance frame instead of the position screen, since this is about tracking nodes. 100% agree that this is more relevant.
- Added a "Show Favorites Only" toggle in the radar long-press menu. Agree, that this is totally make sence.
- Markers are now unique per row (■ + × □ ◆) instead of random symbols, so the dot on the radar always matches the row in the list.

@Xaositek - addressed each of your review points:
- Compass \ Radar toggle now uses a clock-face-style picker ("Tracking View" → Bearings / Radar). Reachable from both the node-list menu and the radar long-press menu.
- GPS-lock indicator: when there's no fix, the radar is replaced with a clear "No GPS fix" message - no rings, no markers, nothing ambiguous.
- Zoom level: I tried showing it on the radar face, but it just turned into visual noise on a 128×64 OLED. Instead, the header now reads "Radar 5km" - the current outer-ring range, which is what the zoom number actually means in practice. Open to better ideas.
- About the empty radar you saw on Cardputer - I think I found the bug. Your screenshot's list showed nodes at 5ft, 26ft, 59ft, 68ft, 77ft - all within ~23m. But the old auto-scale was computed from every valid-position node in your NodeDB, not just the 5 closest that get plotted. A single far node (a stale MQTT contact a few km away, for example) would push the outer ring out to ~6km. Your close-by nodes then plotted at <1 pixel from the centre, all overlapping the own-node marker. The list looked fine because it just prints the distance. That's fixed in this push - scale now only uses the 5 nodes that are actually plotted. Could you give it another try? A screenshot of what you see would be great, even if it works — I want to confirm the spread looks right at your screen size.

egorsiniaev:
Updated PR a bit. Fixed layout to better show radar and node list, especially when bluetooth connected. Plus fixed a bug with wrong order of nodes and showing not correct ones.

HarukiToreda:
Looks good, got a chance to test it today

https://github.com/user-attachments/assets/c5b0f42a-d56e-41d6-9985-48e43631bc1b

HarukiToreda (approved):

egorsiniaev:
@HarukiToreda Thanks for checking! 
I think I need to order some device with TFT to test it properly. UI now is too far from ideal, it could show 8-10 nodes + radar could be more readable.

HarukiToreda:
yes, if you can fit it, allowing more nodes to show to take advantage of the higher resolution and space would be great.

egorsiniaev:
@HarukiToreda could you tell me your device name and screen resolution? 

I also think - let's use Meshtastic community you to test it properly. I wrote it for 128×64 OLED, but there are some TFT and eInk screens. Any idea how & where to populate it to test? 
We need: 
- T-Display S3 / T-Lora Pager (170×320 TFT)
- T-Echo or t114 with EInk (200×200 / 250×122 EInk)
- T-Deck / T-Deck Pro (240×320 TFT )
- M5Stack Cardputer (240×135 TFT)
- Elecrow / T5-S3 / large EInk  (20×480+)
- Unit C6L  (64×48 OLED)

HarukiToreda:
I was using a Heltec T114 with a large 240x320 screen I personally added, a good one to try with a large enough screen like that is the Tdeck, i think what you have now is good to merge. Fine tuning for larger screens could be a new PR

egorsiniaev:
Hey @Xaositek. Could you please take a look on this PR?

src/mesh/generated/meshtastic/device_ui.pb.h
```diff
@@ -277,6 +281,8 @@ extern "C" {
 #define meshtastic_DeviceUIConfig_screen_rgb_color_tag 17
 #define meshtastic_DeviceUIConfig_is_clockface_analog_tag 18
 #define meshtastic_DeviceUIConfig_gps_format_tag 19
+#define meshtastic_DeviceUIConfig_bearings_view_radar_tag 20
```
fifieldt, line 284: Can you please make a PR in the meshtastic/protobufs repo to add these there?

egorsiniaev: Sure. This is the PR - https://github.com/meshtastic/protobufs/pull/935

egorsiniaev:
While waiting for PR - I ordered T-Deck and did some tweaks for highres screens:
- Now it could 5 \ 8 \ 10 nodes depends on screen size
- Optimised paddings for different screens
- Added ring labels to show distance. Works only on highres. 


FYI @HarukiToreda @HarukiToreda @fifieldt Let me know what do you think.

egorsiniaev:
@HarukiToreda @fifieldt Could you take a look on this PRs, guys?

egorsiniaev:
Hey hey @fifieldt @Xaositek Friendly ping. Please review this PR.

Xaositek (approved):

egorsiniaev:
Sorry for the ping again, @Xaositek  could you trigger CI again? I got some feedback from CodeRabbit and fixed them.

egorsiniaev:
Hey @Xaositek @fifieldt @caveman99 @HarukiToreda could you please trigger CI?

egorsiniaev:
Hey @Xaositek @fifieldt @caveman99 @HarukiToreda 
Builds are green. Could you please help me to understand how to merge this PR? 

FYI. This PR https://github.com/meshtastic/protobufs/pull/935 need to be merged before this.

HarukiToreda:
we don't merge PRs ourselves, we can only approve it, which I believe all have by now, looks good. Merging decision is usually done by @thebentern and other admin maintainers. I do believe the need for Protos is what has delayed this decision, even my PRs get delayed when protos are involved.

egorsiniaev:
Ok. Thank you @HarukiToreda for explaining this. Could you take a look on Protos PR too? =) 
Then if all will be done, I will just wait.

egorsiniaev:
Hey @thebentern. It's a been a while that this PR created (April) and ready. Could you please tell - what's need to be done to make it merged? 
Also - I can't get any reaction in protobuf PR to move this PR forward.

src/mesh/generated/meshtastic/device_ui.pb.h
caveman99, line 1: This file needs to be removed before the proto merge or it will conflict badly. it will be regenerated automatically.

caveman99 (changes requested):

src/graphics/draw/RadarRenderer.cpp
```diff
@@ -0,0 +1,507 @@
...
+    // -----------------------------------------------------------------------
+    if (currentResolution == ScreenResolution::High) {
+        display->setFont(FONT_SMALL_LOCAL);
+        display->setTextAlignment(TEXT_ALIGN_CENTER);
+        const int kRingFontH = _fontHeight(FONT_SMALL_LOCAL);
+        const float oppNBrg = -headingRad + static_cast<float>(M_PI); // 180° from N
+        for (int ring = 1; ring <= 3; ring++) {
+            const int ringR = (radarRadius * ring) / 3;
+            char ringLabel[12];
+            formatDistNum(ringLabel, sizeof(ringLabel), scale * ring / 3.0f);
```
caveman99, line 409: **Ring labels mix units within one radar and print no suffix.**

`formatDistNum` (line 96) picks its unit *per value*: metric switches from metres to km at 1000 m, imperial from feet to miles at 0.1 mi. It emits a bare number with no suffix, so the three rings can't be told apart.

With a 1.5 km outer ring, the three rings are 500 m / 1000 m / 1500 m and label as:

```
500    1.0    1.5
```

The first is metres, the other two kilometres — visually identical. Imperial is worse: a 1 km outer ring gives `328` (feet), `0.1`, `0.2` (miles).

Two ways out: pick the unit once from `scale` and format all three rings in it, or append the suffix. Given the header already carries the range as `Radar 5km`, deriving the unit from `scale` and keeping the labels bare would stay consistent with it.

src/graphics/draw/RadarRenderer.cpp
```diff
@@ -0,0 +1,507 @@
...
+        else
+            snprintf(name, sizeof(name), "%04X", (uint16_t)(e.node->num & 0xFFFF));
+
+        char dist[10] = "";
+        formatDistM(dist, sizeof(dist), e.distM);
+
+        display->setTextAlignment(TEXT_ALIGN_LEFT);
+        display->drawString(x + 11, rowY, name); // 3 px gap after marker right edge
+        display->setTextAlignment(TEXT_ALIGN_RIGHT);
+        display->drawString(x + listRight, rowY, dist);
```
caveman99, line 482: **Frame offset `x` is counted twice here.**

`listRight` already includes `x`:

```cpp
const int radarCX  = x + sw - pad - radarRadius;   // line 274
const int listRight = radarCX - radarRadius - 4;   // line 278 — carries x
```

so `x + listRight` applies the frame offset a second time. At `x == 0` (the normal static frame) this is invisible, which is why it doesn't show in the screenshots — but `drawRadarOverlay` is called with a non-zero `x` during the UI library's slide transitions, and there the distance column travels at twice the speed of everything else and detaches from the frame.

`display->drawString(listRight, rowY, dist);` is the fix.

The same shape appears at line 492 in the empty-list branch:

```cpp
const int panelCX = (x + radarCX - radarRadius) / 2;
```

`radarCX - radarRadius` is the list panel's right edge and already contains `x`, so adding `x` and halving makes the block drift by `x/2` instead of `x`. That one wants `x + (radarCX - radarRadius - x) / 2`, or equivalently `(x + listRight + 4) / 2` reworked to keep `x` outside the division.

src/graphics/draw/RadarRenderer.cpp
```diff
@@ -0,0 +1,507 @@
...
+    const int listFooterH = (connection_icon_height * footerScale) + 1;
+
+    const int contentH = sh - headerH;               // full-height area for the radar
+    const int listContentH = contentH - listFooterH; // shorter area for list rows
+    const int pad = (currentResolution == ScreenResolution::High) ? 9 : 4;
+
+    // -----------------------------------------------------------------------
+    // Radar circle - right side, 2 px padding on all sides.
+    // -----------------------------------------------------------------------
+    const int radarDiam = contentH - 2 * pad;
```
caveman99, line 272: **Radar diameter is derived from height alone, so the list column collapses on square or narrow panels.**

`radarDiam = contentH - 2 * pad` never consults `sw`, and `listRight` is then whatever is left over:

```cpp
const int radarCX   = x + sw - pad - radarRadius;
const int listRight = radarCX - radarRadius - 4;   // = sw - 2*pad - radarDiam - 4
```

On a 128×64 OLED that's fine (radar ≈ 48 px, list ≈ 72 px). On the wide-and-short or square panels it goes bad in both directions:

- **240×240** (T-Watch S3): `contentH` is nearly 240, so the circle eats almost the full width and `listRight` drops to roughly 20 px — while names still draw at `x + 11` (line 480), so short names and the distance column print straight over the circle.
- **64×48**: same collapse, with even less to give.

Clamping the diameter to the width as well would keep the list panel viable everywhere:

```cpp
const int radarDiam = std::min(contentH - 2 * pad, sw / 2);
```

This is the mirror image of the large-screen tuning @HarukiToreda asked about earlier — that thread was about using *more* space; this is about panels where the aspect ratio leaves the list none.

src/graphics/draw/RadarRenderer.cpp
```diff
@@ -0,0 +1,507 @@
...
+
+    // -----------------------------------------------------------------------
+    // Node list (left panel) - up to 10 closest nodes.
+    //
+    // Each row: marker symbol (matches the radar dot) | short name | distance.
+    // -----------------------------------------------------------------------
+    display->setFont(FONT_SMALL);
+
+    constexpr int kListTopPad = 5;
+    const int rowPitch = (listContentH - kListTopPad) / kMaxPlotted;
```
caveman99, line 455: **`rowPitch` is never checked against the font height, so rows overlap on compact panels.**

`kMaxPlotted` rows are spread across whatever `listContentH` happens to be, with no floor:

```cpp
const int rowPitch = (listContentH - kListTopPad) / kMaxPlotted;
```

On a 72×40 panel `listContentH` works out to roughly 20 px; with `kMaxPlotted == 5` that's a 3–4 px pitch against a 13 px `FONT_HEIGHT_SMALL`, so all five rows print on top of each other.

Deriving the row count from the space actually available degrades gracefully instead:

```cpp
const int maxRows = std::min<int>(kMaxPlotted, (listContentH - kListTopPad) / FONT_HEIGHT_SMALL);
```

and then bounding the loop at line 462 by `std::min(plottedCount, maxRows)`. Fewer nodes listed on a tiny screen is a reasonable outcome; five illegible ones isn't. The radar circle itself is unaffected either way.

src/graphics/draw/RadarRenderer.cpp
```diff
@@ -0,0 +1,507 @@
...
+ * Draw just the BT/API connection icon glyph at the bottom-left, without the
+ * full-width black wipe that drawCommonFooter performs.  The wipe was erasing
+ * the radar circle's bottom arc and the descender of the last list row even
+ * though the icon's actual 5×5 footprint (x=0..4 at scale=1) doesn't overlap
+ * the radar (x≈80..126) or the list text (x≥7).
+ *
+ * Replicates the icon-rendering half of SharedUIDisplay::drawCommonFooter so
+ * this overlay can own its own footer behaviour without touching shared UI.
+ */
+static void drawConnectionIconNoWipe(OLEDDisplay *display)
```
caveman99, line 203: **This copy drops two behaviours from the `drawCommonFooter` original it replicates.**

The comment above says it "replicates the icon-rendering half of `SharedUIDisplay::drawCommonFooter`", but two things didn't come across:

1. **The `isCompactPanel()` early return.** `drawCommonFooter` skips the connection icon entirely on compact panels because there is no room for it. Without that guard, the icon is painted over radar content on exactly the screens that can least afford it.
2. **The `TFTColorRole::ConnectionIcon` registration.** Colour TFT builds tint the icon through that role; unregistered, it renders in the default colour and looks out of place next to every other screen's footer.

Both are one-liners lifted from the original. Alternatively, since the stated reason for the copy is avoiding the background wipe, a `bool wipe = true` parameter on the shared helper would let this path reuse it and stay in sync as the shared footer evolves.

caveman99 (commented):
Reviewed at 346291d (+712/−5 across 10 files) against `develop` at 80cfa52 — read every hunk plus the surrounding renderers, the menu plumbing, the nanopb save/load paths, and the pinned `esp8266-oled-ssd1306` library to check framebuffer and transition semantics. Nice feature, and the iteration history on this PR shows in the code.

src/mesh/generated/meshtastic/device_ui.pb.h
egorsiniaev: Will do, but to delete it, I need to cover proto PR first. Please help me to find a solution there.

src/graphics/draw/RadarRenderer.cpp
egorsiniaev: Good catch. Fixed: the unit is now picked once from scale and all three rings use it, so they line up with the header. formatDistNum became formatRingNum(..., scale).

src/graphics/draw/RadarRenderer.cpp
egorsiniaev: Fixed both. Worth noting the second one is visible: since the incoming frame is drawn at x = width, the halved offset put the "No active nodes" label in the middle of the screen for a frame during page changes.

src/graphics/draw/RadarRenderer.cpp
egorsiniaev: Done, clamped to sw / 2 as suggested.

One consequence to flag: on a 320x240 T-Deck this takes the circle from 204 px to 160 px and grows the list from 103 px to 147 px. It fixes 240x240 properly (the list was 23 px there), but it does cost radar area on the big screens. Happy to keep it as is, or to try a clamp based on the list's actual text width if you would rather protect the circle on wide panels. Your call.

I also had to centre the circle vertically as a follow-up, since once it is shorter than the content area it was left pinned under the header with all the slack at the bottom.

src/graphics/draw/RadarRenderer.cpp
egorsiniaev: Fixed, though I used a different threshold than suggested and wanted to explain why.

FONT_HEIGHT_SMALL is 13 on a 128x64 OLED, so (listContentH - kListTopPad) / FONT_HEIGHT_SMALL gives 2 rows there and the main display would drop from 5 to 2. The current layout works because the glyph ink is roughly half the font bbox, which the existing comments call out.

So I floored the pitch at FONT_HEIGHT_SMALL / 2 + 1 and derived the row count back from it. Checked numerically:

128x64    pitch  8 -> 8   rows  5 -> 5   (unchanged)
72x40     pitch  3 -> 7   rows  5 -> 2
64x48     pitch  5 -> 7   rows  5 -> 3
320x240   pitch 20 -> 20  rows 10 -> 10  (unchanged)

I have no small screen devices, so could not really test it.

src/graphics/draw/RadarRenderer.cpp
egorsiniaev: Took the second option. drawConnectionIconNoWipe is gone and drawCommonFooter now takes bool wipe = true; the radar calls it with false. That picks up the isCompactPanel guard and the ConnectionIcon role, and it will not drift out of sync from here. All existing callers are untouched.

I also added a null guard on service in the shared function, which the copy had and the original did not. Shout if you would rather I left that alone.

egorsiniaev:
@caveman99 Thank you for all comments! I fixed all issues and suggestions (except device_ui.pb.h). Please take a look. I will remove device_ui.pb.h when we will solve protobuf PR.

PR #10227, opened 2026-04-21
NextHopRouter: fix 49.7-day millis() rollover in retransmission timing

nightjoker7:
## Summary

Fix the `// FIXME, handle 51 day rolloever here!!!` in `NextHopRouter::doRetransmissions()` by switching the retransmission-due check to a signed-difference comparison.

## Problem

```cpp
// FIXME, handle 51 day rolloever here!!!
if (p.nextTxMsec <= now) {
```

`millis()` wraps every ~49.7 days (2^32 ms). Both `p.nextTxMsec` and `now` are `uint32_t`. When `now` wraps past the stored `nextTxMsec`, the plain `<=` comparison silently does the wrong thing in two ways:

1. Just before rollover (large `nextTxMsec`, small `now`): all pending retransmissions stall until `now` catches up — minutes to days depending on how close to the wrap the packets were queued.
2. Just after rollover (small `nextTxMsec`, large `now` before wrap): the entire retransmission queue becomes due simultaneously, causing a burst of airtime at the rollover boundary.

The packet-retransmission table is long-lived on routers and infrastructure nodes that stay up for weeks, so this FIXME is reachable in normal operation.

## Fix

Standard Arduino/embedded idiom for rollover-safe "deadline has passed" checks — subtract first, then compare as signed:

```cpp
if ((int32_t)(p.nextTxMsec - now) <= 0) {
```

`(p.nextTxMsec - now)` is computed in `uint32_t` and yields a value whose two's-complement interpretation as `int32_t` is the true signed time delta, provided the actual delta is within ±2^31 ms (~24.8 days). Retransmission deadlines are milliseconds to tens of seconds in the future, so this is always satisfied.

This is the same pattern used elsewhere in the Arduino ecosystem for rollover-safe timing (e.g. `(long)(millis() - target) >= 0`).

## Testing

- Builds cleanly against `develop`.
- Running on a 4-node fleet (RAK4631 / Station G2 / T114 / Heltec V3). Reliable-DM retransmissions continue to behave correctly under normal millis() values; the rollover case is by construction impossible to exercise in a reasonable test window but the replacement expression is a drop-in for the classic Arduino rollover-safe compare.

## Risk

Minimal. One-line change. On non-rollover timelines `(int32_t)(future - now)` is positive and `<= 0` is false exactly when `nextTxMsec <= now` was false — identical behavior. Difference only manifests across the wrap boundary, which is the bug being fixed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message retransmission timing across system clock rollover.
  * Prevented retransmissions from firing prematurely or becoming delayed when the timer wraps around.

### DIFF
--- a/src/mesh/NextHopRouter.cpp
+++ b/src/mesh/NextHopRouter.cpp
@@ -291,8 +291,10 @@ int32_t NextHopRouter::doRetransmissions()
 
         bool stillValid = true; // assume we'll keep this record around
 
-        // FIXME, handle 51 day rolloever here!!!
-        if (p.nextTxMsec <= now) {
+        // Use signed-difference comparison so retransmission timing stays correct across the
+        // ~49.7 day millis() wraparound (previously this FIXME would stall all retx for the
+        // duration of the wrap or fire them all at once immediately after).
+        if ((int32_t)(p.nextTxMsec - now) <= 0) {
             if (p.numRetransmissions == 0) {
                 if (isFromUs(p.packet)) {
                     LOG_DEBUG("Reliable send failed, returning a nak for fr=0x%x,to=0x%x,id=0x%x", p.packet->from, p.packet->to,
--- a/src/mesh/NextHopRouter.cpp
+++ b/src/mesh/NextHopRouter.cpp
@@ -405,10 +405,16 @@ int32_t NextHopRouter::doRetransmissions()
 
         bool stillValid = true; // assume we'll keep this record around
 
-        // Use signed-difference comparison so retransmission timing stays correct across the
+        // Use unsigned half-range comparison so retransmission timing stays correct across the
         // ~49.7 day millis() wraparound (previously this FIXME would stall all retx for the
         // duration of the wrap or fire them all at once immediately after).
-        if ((int32_t)(p.nextTxMsec - now) <= 0) {
+        //
+        // Casting an unsigned difference to int32_t for a "time passed" test is
+        // implementation-defined in C++ when the value exceeds INT32_MAX. The unsigned
+        // half-range form below is fully well-defined: nextTxMsec is in the past (or is now)
+        // iff (now - nextTxMsec) has not wrapped past 2^31 ms. Anything further in the
+        // future wraps into the top half and reads as "not yet."
+        if ((uint32_t)(now - p.nextTxMsec) < 0x80000000u) {
             if (p.numRetransmissions == 0) {
                 if (isFromUs(p.packet)) {
                     LOG_DEBUG("Reliable send failed, returning a nak for fr=0x%08x,to=0x%08x,id=0x%08x", p.packet->from,
--- a/src/mesh/NextHopRouter.cpp
+++ b/src/mesh/NextHopRouter.cpp
@@ -405,8 +405,10 @@ int32_t NextHopRouter::doRetransmissions()
 
         bool stillValid = true; // assume we'll keep this record around
 
-        // FIXME, handle 51 day rolloever here!!!
-        if (p.nextTxMsec <= now) {
+        // Use signed-difference comparison so retransmission timing stays correct across the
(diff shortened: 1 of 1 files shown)


nightjoker7:
Addressed in [1cd825bce](https://github.com/meshtastic/firmware/pull/10227/commits/1cd825bce): switched from signed-cast of unsigned subtraction to the fully well-defined unsigned half-range form — `(now - nextTxMsec) < 0x80000000u`. Same semantics on every two's-complement platform, but portable to any conforming C++ impl.

cvaldess:
Tested on Nordic nRF54L15-DK (Zephyr 4.2.1 + arm-none-eabi-gcc).

- Cherry-picked both commits clean on top of upstream/develop.
- Builds without warnings, no footprint change.
- `doRetransmissions()` runs as expected during ~5 min of mesh traffic;
  no behavioural regressions observed (LoRa RX/TX, BLE pair, config stream
  all OK).

Can't repro the 49.7-day rollover in a smoke test obviously, but the
unsigned half-range form is well-defined per the standard whereas the
prior `(int32_t)(unsigned - unsigned) <= 0` was implementation-defined
when the result exceeded INT32_MAX — so this is also a portability win,
not just a correctness one. LGTM.

Tested-by: cvaldess

fifieldt (approved):

NomDeTom:
Sorry - got carried away rebasing with clod. Might need a hard reset at the next pull.

NomDeTom:
Tom says to write "Clod here, with a polite note"

@nightjoker7 — a heads-up from adopting this PR rather than a criticism of it. We think the fix is
correct, and we've taken it as a baseline for a wider `millis()` rollover cleanup (tracking issue
#11281). While running the full native suite on top of it, one existing test started failing.

`test_packet_signing` → `test_C3_invalid_repeated_packet_cannot_ack_or_change_retry_state`:

```
test/test_packet_signing/test_main.cpp:1224: Expected 4294967295 Was 6247
```

That test parks a pending packet at `addPending(prior, UINT32_MAX)` to mean *"never retransmit"*, then
asserts an invalid repeated packet leaves the retry state untouched. Under the old
`p.nextTxMsec <= now`, `UINT32_MAX` read as the far future. Under the new
`(uint32_t)(now - p.nextTxMsec) < 0x80000000u`, `now - 0xFFFFFFFF == now + 1` — a small positive delta
— so the sentinel now reads as roughly 1 ms **in the past**, the retransmission fires, and
`nextTxMsec` gets rewritten. The assertion is doing its job.

Two things worth saying about it:

- **It looks like a test-only problem, not a production regression.** In `src/`, `nextTxMsec` is only
  ever written as `millis() + d` (`NextHopRouter.cpp:513`); the `UINT32_MAX` value comes from a
  test-harness helper, so nothing shipped parks a pending packet at "never".
- **The test post-dates this PR.** It came in with `d6b12ea3f` (packet authenticity policies), long
  after this branch was opened in April — which is why it only surfaced after the rebase onto current
  `develop`.

Suggested fix, for whatever it's worth: change the *test's* "never" value to something genuinely in
the future and inside the half-range, e.g. `millis() + 3600000UL`, rather than teaching
`doRetransmissions()` to special-case `UINT32_MAX`. A sentinel that reads as "expired" under any
wrap-correct comparison is exactly the hazard the rest of this cleanup is trying to remove, and the
retransmit path seems like a poor place to keep one.

One CI note: the only checks currently showing on this PR are CodeRabbit and `license/cla`. The
`native-tests` job (`pr_tests.yml` → `test_native.yml`) doesn't appear to have reported against the
current head `1d97a29b5`, so the green state here isn't covering the suite that catches this. Might be
worth a maintainer triggering a run before it merges.

Happy to open a small PR with just the one-line test change if that's easier than doing it here.

NomDeTom:
closed and incorporated into https://github.com/meshtastic/firmware/pull/11291 with author credit

PR #10235, opened 2026-04-22
[draft] pioarduino: fix tbeam (ESP32 original) support

cpatulea:
DRAFT ONLY

Current state of pioarduino branch: https://github.com/meshtastic/firmware/actions/runs/24730901810/job/72346464441

As expected: https://github.com/meshtastic/firmware/pull/9122#issuecomment-4241813349

We asked why ESP32 original still uses Bluedroid: https://github.com/espressif/esp32-arduino-lib-builder/pull/296#issuecomment-4292724957 answer is there.

Let's try to handle ESP32 original with external NimBLE-Arduino on Meshtastic side.

State after this PR (still many build errors):

```
src/nimble/NimbleBluetooth.cpp: In member function 'void BluetoothPhoneAPI::runOnceHandleFromPhoneQueue()':
src/nimble/NimbleBluetooth.cpp:327:35: error: invalid conversion from 'char*' to 'const uint8_t*' {aka 'const unsigned char*'} [-fpermissive]
  327 |             handleToRadio(val.data(), val.size());
      |                           ~~~~~~~~^~
      |                                   |
      |                                   char*
In file included from src/mesh/StreamAPI.h:3,
                 from src/SerialConsole.h:4,
                 from src/DebugConfiguration.h:34,
                 from src/configuration.h:560,
                 from src/nimble/NimbleBluetooth.cpp:1:
src/mesh/PhoneAPI.h:116:47: note:   initializing argument 1 of 'virtual bool PhoneAPI::handleToRadio(const uint8_t*, size_t)'
  116 |     virtual bool handleToRadio(const uint8_t *buf, size_t len);
      |                                ~~~~~~~~~~~~~~~^~~
src/nimble/NimbleBluetooth.cpp: At global scope:
src/nimble/NimbleBluetooth.cpp:415:10: error: 'void NimbleBluetoothToRadioCallback::onWrite(NimBLECharacteristic*)' marked 'override', but does not override
  415 |     void onWrite(BLECharacteristic *pCharacteristic) override
      |          ^~~~~~~
In file included from .pio/libdeps/tbeam/NimBLE-Arduino/src/NimBLEService.h:28,
                 from .pio/libdeps/tbeam/NimBLE-Arduino/src/NimBLEDevice.h:294,
                 from src/nimble/NimbleBluetooth.cpp:16:
[...]

PR #10244, opened 2026-04-22
[pioarduino] Build ESP32 original with NimBLE

vidplace7:
Testing the "force nimble on ESP32" approach to getting original ESP32 working.


vidplace7:
Whoops this is a dupe of
- #10235

PR #10247, opened 2026-04-22
Fix INA226 detection for non-TI compatible chip (Silergy)

theKorzh:
#### Problem

Popular cheap INA226-compatible current/power monitor chip Silergy SQ52201 is incorrectly detected as `SHTXX (SHT2X)`.
(These chips are commonly found on modules sold as “INA226” or even misrepresented as TI parts, so users may be unaware they are using Silergy silicon instead of Texas Instruments).

This happens due to strict Manufacturer ID check (only TI `0x5449` accepted)

---

#### Root cause

INA2xx-compatible chips do not always use TI's Manufacturer ID:

| Device        | MFG ID | DIE ID |
|--------------|--------|--------|
| INA226 (TI)  | 0x5449 | 0x2260 |
| SQ52201 (Silergy) | 0x190F | 0x0000 |

Current logic only accepts `0x5449`, causing valid compatible devices to fall through to SHT detection.

---

#### Changes

- Extended INA detection to support known compatible MFG IDs

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other: Heltec VisionMaster E213


theKorzh:
Closing this PR for now.

It appears that the Silergy SQ52201 is not fully compatible with INA226.

I will revisit logic.

theKorzh:
Reopening this PR — closed it by mistake.

During additional testing of the SQ52201 support and compatiblity, I accidentally used a defective module unit, which produced incorrect readings and made it look like the changes were broken. The code itself is ready for review.

fifieldt (approved):
Thanks for the detailed work!

PR #10251, opened 2026-04-23
PositionModule::sendLostAndFoundText: use stack buffer, eliminate heap alloc

nightjoker7:
## Problem

`src/modules/PositionModule.cpp::sendLostAndFoundText()` uses an unnecessary 60-byte heap allocation + unbounded `sprintf`:

```cpp
char *message = new char[60];
sprintf(message, "🚨I'm lost! Lat / Lon: %f, %f\a", (lastGpsLatitude * 1e-7), (lastGpsLongitude * 1e-7));
...
delete[] message;
```

Two issues with that:

1. **60 bytes is too small in worst cases.** The `%f` specifier renders IEEE-754 doubles at default precision — each `%f` alone can exceed 10–15 characters in the pathological case, plus the emoji is 4 UTF-8 bytes, plus the `\a` BEL, plus the `\0` terminator. With two `%f`s plus the surrounding text, a typical render is ~40–50 bytes but the upper bound is not controlled. And `sprintf` (not `snprintf`) has no bounds check.

2. **Unnecessary heap allocation in a GPS-callback path.** This code runs in a heap-sensitive context. A stack buffer is trivially available here.

## Fix

Replace the heap buffer with a 128-byte stack buffer and use bounded `snprintf`:

```cpp
char message[128];
snprintf(message, sizeof(message), "...", ...);
```

128 bytes is comfortably above the practical maximum and matches the sizing convention used elsewhere in PositionModule (`canned_message_module` etc.). Drop the matching `delete[]` since there's nothing to delete.

## Risk

- **Happy path**: byte-identical output.
- **Overflow path**: was undefined-behavior heap corruption, is now safe truncation at 127 bytes + NUL.
- **Stack usage**: +128 bytes in a function that's called infrequently (only when `sendLostAndFoundText()` is explicitly invoked), not in any tight loop.

## Related

Part of a broader "eliminate unnecessary heap allocs in module hot-paths" cleanup I've been working through on my deployment.


nightjoker7:
Addressed in [c7bdfa0bd](https://github.com/meshtastic/firmware/pull/10251/commits/c7bdfa0bd): cleaner form using `std::min(static_cast<size_t>(written), sizeof(message) - 1)` for the size clamp. Handles snprintf's 'would-have-written' return semantics and the encoding-error (<0) path.

thebentern:
@copilot trunk fmt

PR #10258, opened 2026-04-23
KeyVerificationModule: null-check getMeshNode() before dereference

nightjoker7:
## Summary

Six sites in `KeyVerificationModule.cpp` dereferenced `nodeDB->getMeshNode(currentRemoteNode)->user` (or `->bitfield`) without a null check. Four of these carry a literal `// should really check for nulls, etc` comment from the original commit, acknowledging the bug was known at merge time.

`getMeshNode()` returns `NULL` when the node number isn't in the NodeDB — which can happen between key-verification state transitions if the DB evicts the remote node (eg. DB fills up while we wait for the user to punch in a Security Number on the banner overlay). When that happens, the firmware crashes instead of cancelling verification cleanly.

## Sites fixed

1. `handleAdminMessageForModule` — `DO_VERIFY` path: null-check before setting `NODEINFO_BITFIELD_IS_KEY_MANUALLY_VERIFIED_MASK`.
2. `handleReceivedProtobuf` — hash2-received path: null-check before copying `user.long_name` into the Security Number Request ClientNotification.
3. `handleReceivedProtobuf` — hash1-match banner callback: null-check inside the lambda before setting the verified bitfield.
4. `handleReceivedProtobuf` — hash1-match notification: null-check before copying `user.long_name` into the Final ClientNotification.
5. `allocReply` — null-check before copying `user.long_name` into the Number Inform ClientNotification.
6. `processSecurityNumber` — reuse the `remoteNodePtr` already null-checked at the top of the function, instead of calling `getMeshNode()` a second time.

When the node is missing, the long_name copy falls back to an empty string so the notification still fires — blank remote name is better than a crash.

## Test plan

- [x] Grep confirms no remaining `nodeDB->getMeshNode(currentRemoteNode)->` dereferences without a preceding null check in this file
- [x] The existing null-check at line 222 in `processSecurityNumber` already guards site #6's `remoteNodePtr` reuse
- [ ] Build verification (existing CI)

---

This is part of an ongoing null-deref sweep across the codebase; related PRs in flight: #10226 (Router PKI decode), #10229 (MeshService has_user check).


nightjoker7:
Addressed in [bfaa1cfed](https://github.com/meshtastic/firmware/pull/10258/commits/bfaa1cfed): at each of the 4 `strncpy` sites, copy `sizeof(dst) - 1` and explicitly set the last byte to `'\0'`. Guarantees NUL-termination even when `user.long_name` fills the buffer.

jp-bennett (approved):
Looks OK to me

PR #10261, opened 2026-04-23
Router::allocForSending: null-check packet pool allocation

nightjoker7:
## Summary

[Router::allocForSending()](https://github.com/meshtastic/firmware/blob/develop/src/mesh/Router.cpp#L208-L221) calls \`packetPool.allocZeroed()\` and dereferences the result (\`p->which_payload_variant = ...\`) **before** any null check. When the static MemoryPool is exhausted, \`allocZeroed()\` returns nullptr (with a LOG_WARN) rather than asserting — so we crash reading \`p->which_payload_variant\`.

From [MemoryPool.h:149-160](https://github.com/meshtastic/firmware/blob/develop/src/mesh/MemoryPool.h#L149-L160) — static pools return nullptr when no free slot is available. That is a real failure mode under burst conditions (traceroute flood response, simultaneous module sends, ACK/NAK storms). On a ROUTER_LATE relaying traffic for a large neighborhood, the \`MAX_PACKETS\` limit can be saturated.

## Fix

Null-check the allocation and return nullptr on failure. Every upstream caller already handles a nullptr return (eg. \`RoutingModule::sendAckNak\`, PositionModule, TelemetryModule, admin reply paths) — we just need to let them.

## Test plan

- [x] Grep confirms the callers of \`allocForSending()\` all handle nullptr
- [ ] Simulate pool exhaustion (fill txQueue) and confirm graceful degradation instead of crash

## Related

This is the same bug class as #10252 (RadioLib RX path hardening, which included a similar allocZeroed NULL-check). Ongoing pool-allocation safety sweep.

### DIFF
--- a/src/graphics/niche/InkHUD/Applets/System/Menu/MenuApplet.cpp
+++ b/src/graphics/niche/InkHUD/Applets/System/Menu/MenuApplet.cpp
@@ -1993,6 +1993,10 @@ uint16_t InkHUD::MenuApplet::getSystemInfoPanelHeight()
 void InkHUD::MenuApplet::sendText(NodeNum dest, ChannelIndex channel, const char *message)
 {
     meshtastic_MeshPacket *p = router->allocForSending();
+    if (!p) {
+        LOG_WARN("MenuApplet::sendText: packet pool exhausted, dropping message");
+        return;
+    }
     p->decoded.portnum = meshtastic_PortNum_TEXT_MESSAGE_APP;
     p->to = dest;
     p->channel = channel;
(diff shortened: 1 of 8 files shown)


thebentern:
@nightjoker7 I'm not sure every module actually handles nullptr gracefully. I believe I've seen this crash before when I have implemented an early return before. We need unit tests for each one before I will feel remotely comfortable with this change.

nightjoker7:
@thebentern done full caller audit in [43bb5841f](https://github.com/meshtastic/firmware/pull/10261/commits/43bb5841f). All 10 call sites of `allocForSending()` now either early-return on NULL or forward NULL up their own call chain. See commit message for per-site classification. The three that were already NULL-safe (TraceRouteModule x2, TrafficManagement) unchanged; the seven that weren't now are.

PR #10262, opened 2026-04-23
Null-check pool.allocZeroed() at 13 sites across 9 files

nightjoker7:
## Summary

13 sites across 9 files called \`pool.allocZeroed()\` and immediately dereferenced the result without null-checking. When the static MemoryPool is exhausted, \`allocZeroed()\` returns nullptr (with a LOG_WARN in Allocator::allocZeroed) — every dereference that follows is a crash.

From [MemoryPool.h:149-160](https://github.com/meshtastic/firmware/blob/develop/src/mesh/MemoryPool.h#L149-L160): static pools return nullptr when no free slot is available. That's a real failure mode under burst load.

## Sites fixed

| File | Line | Pool | Trigger |
|---|---|---|---|
| \`main.cpp\` | 911 | clientNotification | Low-entropy key warning on boot |
| \`mesh/NodeDB.cpp\` | 1883 | clientNotification | Duplicate public key warning |
| \`mesh/PhoneAPI.cpp\` | 767 | clientNotification | \`sendNotification()\` helper |
| \`mesh/Router.cpp\` | 327 | clientNotification | Duty-cycle-exceeded notification |
| \`modules/PositionModule.cpp\` | 399 | clientNotification | Tracker role power-save notify |
| \`modules/SerialModule.cpp\` | 98 | clientNotification | Invalid serial config |
| \`modules/Telemetry/AirQualityTelemetry.cpp\` | 418, 432 | clientNotification | Sensor role power-save notify (2 sites) |
| \`modules/Telemetry/EnvironmentTelemetry.cpp\` | 646 | clientNotification | Sensor role power-save notify |
| \`mqtt/MQTT.cpp\` | 484, 505 | mqttClientProxyMessage | MQTT publish proxy (2 sites) |
| \`mqtt/MQTT.cpp\` | 687, 703 | clientNotification | Invalid-config notifications (2 sites) |
| \`mqtt/MQTT.cpp\` | 888 | packet | Map report MeshPacket alloc |

## Style

Matches the existing pattern used in already-null-checked sites:
- [AdminModule::sendWarning](https://github.com/meshtastic/firmware/blob/develop/src/modules/AdminModule.cpp#L1553-L1555): \`if (!cn) return;\`
- [RadioInterface::sendErrorNotification](https://github.com/meshtastic/firmware/blob/develop/src/mesh/RadioInterface.cpp#L773-L775): \`if (!cn) return;\`
- [MQTT.cpp:671-678](https://github.com/meshtastic/firmware/blob/develop/src/mqtt/MQTT.cpp#L671-L678): \`if (cn) { ... }\`

Used \`if (!cn) return;\` for single-exit contexts and \`if (cn) { ... }\` where the surrounding code needs to run regardless (eg. \`sleepOnNextExecution = true;\` still happens in the telemetry modules even when the notification can't be sent).

## Related

- #10261 — same bug class in \`Router::allocForSending\` (separate PR, narrower scope)
- #10252 — same bug class in \`RadioLibInterface\` RX path (already filed)

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -482,8 +482,10 @@ bool MQTT::publish(const char *topic, const char *payload, bool retained)
 {
     if (moduleConfig.mqtt.proxy_to_client_enabled) {
         meshtastic_MqttClientProxyMessage *msg = mqttClientProxyMessagePool.allocZeroed();
-        if (!msg)
+        if (!msg) {
+            LOG_WARN("MQTT proxy publish skipped: message pool exhausted");
             return false;
+        }
         msg->which_payload_variant = meshtastic_MqttClientProxyMessage_text_tag;
         strncpy(msg->topic, topic, sizeof(msg->topic));
         msg->topic[sizeof(msg->topic) - 1] = '\0';
@@ -505,8 +507,10 @@ bool MQTT::publish(const char *topic, const uint8_t *payload, size_t length, boo
 {
     if (moduleConfig.mqtt.proxy_to_client_enabled) {
         meshtastic_MqttClientProxyMessage *msg = mqttClientProxyMessagePool.allocZeroed();
-        if (!msg)
+        if (!msg) {
+            LOG_WARN("MQTT proxy publish skipped: message pool exhausted");
             return false;
+        }
         msg->which_payload_variant = meshtastic_MqttClientProxyMessage_data_tag;
         strncpy(msg->topic, topic, sizeof(msg->topic));
         msg->topic[sizeof(msg->topic) - 1] = '\0'; // Ensure null termination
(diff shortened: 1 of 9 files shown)


src/mqtt/MQTT.cpp
```diff
@@ -878,6 +886,10 @@ void MQTT::perhapsReportToMap()
 
     // Allocate MeshPacket and fill it
     meshtastic_MeshPacket *mp = packetPool.allocZeroed();
+    if (!mp) {
+        LOG_WARN("MQTT Map report: packet pool exhausted");
```
thebentern: Should be LOG_ERROR if we can't allocate enough memory to send a map report

nightjoker7:
Addressed all 5 review points in [055bf7b52](https://github.com/meshtastic/firmware/pull/10262/commits/055bf7b52): map-report LOG_ERROR, proxy-publish LOG_WARN, hasWarned unconditional, level-parameter honored, sprintf→snprintf for public-key warning. See commit message for per-issue detail.

PR #10271, opened 2026-04-23
PhoneAPI: rate-limit handleStartConfig to prevent reconnect-storm heap fragmentation

nightjoker7:
## Summary

Add a 2000 ms minimum interval between full config-dump handshakes in [PhoneAPI::handleStartConfig](https://github.com/meshtastic/firmware/blob/develop/src/mesh/PhoneAPI.cpp#L50). A misbehaving or flapping client that reconnects faster than a config dump can complete is currently able to exhaust the internal SRAM heap via fragmentation.

## Evidence

Measured on a Station G2 (ESP32-S3, 248 KB internal SRAM heap, 8 MB PSRAM) running `2.7.23.ef4693d` with full DEBUG_HEAP instrumentation:

- Meshtastic-python \`TCPInterface\` client with a reader thread that gets \`Connection reset by peer\` from a remote under heap pressure → Python side tears down and reconnects → full \`want_config_id\` re-issued → new \`handleStartConfig\` every ~1 s.
- **466 complete reconnect cycles over 85 minutes.**
- **Net heap loss: 108 KB** (avg 232 B / median 192 B per cycle).
- \`heap_free\` trajectory: 150 KB → 50 KB over the 85 min, crashed into reboot.
- Post-reboot with the storm stopped: heap rock-stable at 145 KB for 11 hours (no continued leak).

Stage-by-stage heap delta per cycle (from log parse over 398 cycles):

| Segment | Avg Δ | Median Δ |
|---|---|---|
| \`wants_config → got_files\` | +2,815 | +2,052 |
| \`force_close → incoming_api\` | +1,472 | +1,164 |
| \`send_myinfo → send_metadata\` | **-3,556** | **-3,692** |
| \`send_cfg_device → send_cfg_sec\` | -501 | -568 |

Most of the loss is heap fragmentation from the repeated alloc/free of mid-sized protobuf encode buffers (MY_INFO, OWN_NODEINFO, METADATA, NodeDB entries) interleaved with longer-lived queue entries. Once the storm stops, the free total does not recover until reboot.

## Fix

Add \`PhoneAPI::lastStartConfigMs\` as a static member (shared across all transport subclasses — BLE, WiFi/TCP, Serial, HTTP, Ethernet) and reject handshakes within 2000 ms of the previous one:

- 2000 ms chosen to be just longer than a full config dump takes to complete on a saturated (250-node) NodeDB, so legitimate reconnect-after-disconnect flows still work.
- Well-behaved clients only issue one \`want_config_id\` per reconnect so are unaffected.
- The throttled path returns early *before* touching state, observers, or \`getFiles\` — no resources are leaked per throttled call.
- The client either waits and retries or times out and recovers naturally.

## Test plan

- [x] Builds clean on heltec-v4 (ESP32-S3) — local verification
- [ ] CI build across all targets
- [ ] Repro: run a meshtastic-python TCP reconnect loop against a Station G2, confirm heap_free no longer bleeds out

## Related work

The underlying fragmentation vulnerability is an architectural issue — large per-session protobuf buffers land in internal SRAM even on boards with 8 MB of idle PSRAM. Separately scoping a PSRAM-routing change for the large allocators; this PR is the narrow defensive fix.

### DIFF
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -47,8 +47,28 @@ PhoneAPI::~PhoneAPI()
     close();
 }
 
+uint32_t PhoneAPI::lastStartConfigMs = 0;
+
+// Minimum interval between full config-dump handshakes, shared across all PhoneAPI
+// transports (BLE, WiFi/TCP, Serial, HTTP, Ethernet). Chosen to be just longer than
+// a nominal config dump takes to complete (~1.5-2 seconds on a 250-node DB), so
+// legitimate reconnect-after-disconnect flows still work but a flapping client
+// can't drive one handshake per second and exhaust internal SRAM.
+static constexpr uint32_t MIN_CONFIG_HANDSHAKE_INTERVAL_MS = 2000;
+
 void PhoneAPI::handleStartConfig()
 {
+    // Rate-limit full config dumps. Each dump exercises the protobuf encode path for
+    // all channels, configs, the entire NodeDB, and a full filesystem walk — back-to-back
+    // dumps fragment the internal SRAM heap and can wedge the device on static-pool
+    // allocators. Well-behaved clients only issue one want_config_id per reconnect.
+    if (lastStartConfigMs != 0 && (millis() - lastStartConfigMs) < MIN_CONFIG_HANDSHAKE_INTERVAL_MS) {
+        LOG_WARN("Config handshake throttled (last one %u ms ago, min interval %u ms)",
+                 (unsigned)(millis() - lastStartConfigMs), (unsigned)MIN_CONFIG_HANDSHAKE_INTERVAL_MS);
+        return;
+    }
+    lastStartConfigMs = millis();
+
     // Must be before setting state (because state is how we know !connected)
     if (!isConnected()) {
         onConnectionChanged(true);
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -55,6 +55,9 @@ std::atomic<uint32_t> PhoneAPI::lastStartConfigMs{0};
 // legitimate reconnect-after-disconnect flows still work but a flapping client
 // can't drive one handshake per second and exhaust internal SRAM.
 static constexpr uint32_t MIN_CONFIG_HANDSHAKE_INTERVAL_MS = 2000;
+// Rate-limit the throttled-handshake WARN itself so a flapping client can't turn this
+// defensive code into its own log-flood problem.
+static constexpr uint32_t HANDSHAKE_THROTTLE_LOG_INTERVAL_MS = 5000;
 
 void PhoneAPI::handleStartConfig(uint32_t newConfigNonce)
 {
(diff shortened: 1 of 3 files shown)


nightjoker7:
Thanks @Copilot — all three points addressed in [179821e00](https://github.com/meshtastic/firmware/pull/10271/commits/179821e00):

1. **Nonce race** — Changed `handleStartConfig` to take the incoming nonce as a parameter and only commit it to `config_nonce` **after** the throttle check passes. An in-flight dump keeps its original nonce so `sendConfigComplete` matches what the client asked for. Side benefit: this also fixed a latent bug in `PacketAPI` where `uint32_t config_nonce = mr->want_config_id;` declared a local that shadowed the member — the member was never being updated, so handshakes from that path used whatever stale value was left over.

2. **Data race on `lastStartConfigMs`** — Changed to `std::atomic<uint32_t>` with `compare_exchange_weak` so concurrent handshakes across transports (BLE callback task vs APIServerPort OSThread vs serial-console task) can't both pass the check.

3. **millis() multiple times** — Already fixed by @thebentern's pre-applied autofix in [4c8f555a5](https://github.com/meshtastic/firmware/pull/10271/commits/4c8f555a5).

Built clean on heltec-v4. Ready for re-review.

nightjoker7:
Caught a state-machine bug in my own throttle while hardware-testing today on a Heltec V4 TFT: the original early-return path didn't advance the state machine, so when a client issued a second `want_config_id` while `state` was either `STATE_SEND_NOTHING` (fresh client) or `STATE_SEND_PACKETS` (post-dump steady state), the throttle would silently swallow the request and the client stalled on "reading config" forever.

Reproduced 2026-04-27 during a BLE → TCP transport switch on the Android app: BLE handshake completed → static `lastStartConfigMs` captured timestamp → switched to TCP → second handshake hit at `state=11` (POST_DUMP) within 570ms → throttle returned early → state never advanced → Android stuck.

Fixed in [6d9641fc9](https://github.com/meshtastic/firmware/pull/10271/commits/6d9641fc9): restrict the throttle to states `[SEND_UIDATA..SEND_COMPLETE_ID]` only — i.e. only when a dump is actually in-flight, which is what the heap-fragmentation guarantee the throttle was meant to enforce actually cares about. Fresh clients and post-dump re-handshakes now pass through unconditionally. The original anti-fragmentation property is preserved: concurrent overlapping dumps from a flapping client still hit the throttle.

Hardware-verified on V4 TFT both before and after the fix:
- **Pre-fix**: second handshake hit at `state=11`, throttled at 570ms, Android stalled on every BLE→TCP cycle (4× attempts captured)
- **Post-fix**: second handshake passes through, dump completes cleanly, Android moves past "reading config" to the main UI

Also added `state=%d` to the throttle WARN log so future debugging shows exactly which state the throttle fired in.

cvaldess:
Tested on Nordic nRF54L15-DK with the iOS Meshtastic app — and the
throttle / lightweight-request path actually fires in normal use, not
just under a synthetic reconnect storm.

Trace from the smoke test (uptime in seconds):
```
INFO  | 25 [BleDeferred] Client wants config, nonce=69420
INFO  | 25 [BleDeferred] Start API client config millis=25231
DEBUG | 26 Send config: device
...                              ← full config + module config stream
DEBUG | 27 FromRadio=STATE_SEND_FILEMANIFEST
DEBUG | 27 FromRadio=STATE_SEND_PACKETS
INFO  | 28 [BleDeferred] Client wants config, nonce=69421
INFO  | 28 [BleDeferred] Start API client config millis=28050
INFO  | 28 [BleDeferred] Client only wants node info, skipping other config  ← new path
```

iOS's second `wantConfig` arrived ~2.8 s after the first START and was
correctly identified as a node-info-only request, so the heavy dump was
bypassed. Heap pressure on this port is real (256 KB RAM, ~50 % used
at idle) so any reconnect-storm protection is welcome.

All 5 commits cherry-picked clean (only conflict on rebase was an include
swap `<cstdint>` → `<atomic>` in `PhoneAPI.h`, trivially resolved). MITM
pair, encrypted ATT, MTU 185, full config + 2nd lightweight wantConfig
all green. LGTM.

Tested-by: cvaldess

PR #10276, opened 2026-04-23
power.cpp/h cleanup

jp-bennett:
Move defines and such into power.h, deduplicate a bit of code. I'm not sure it's any prettier, but I find power.cpp slightly more readable. This standardizes the behavior introduced on the t-watch-s3, where clicking the power button sleeps and wakes the screen. Brings that behavior to the t-beam-s3, and also makes that device respond immediately to USB plug and unplug.


fifieldt (approved):

jp-bennett:
I also wonder if it would be useful to split power.cpp into power and powerstatus. There's a lot in the one file currently.

jp-bennett:
Closing this in favor of the separated PRs.

PR #10283, opened 2026-04-24
HWIdentity: derive Curve25519 keypair from per-chip UID + NVS-persistent salt

### LINKED ISSUE #8211: [Bug]: DMing can be permanently impaired if a device resets its keys
### Category

Other

### Hardware

Rak4631

- [ ] InkHUD ePaper

### Firmware Version

2.6.11.60ec05e

### Description

It seems there is a gap in the current Meshtastic implementation of Trust On First Use for direct message encryption keys. If a node is operating in an area with a public mesh and its keys are ever reset, that device may be permanently unable to direct message with nodes that have seen it on the network before.

From this point forward, the device will be unable to send or receive DMs with nodes that previously wrote down the old key. Since the key is associated with the device's fixed Node ID based on the hardware MAC address, the problem is permanent.

The only solution is to get the owner of every node that you might want to DM (or that might want to DM you) to manually delete your device from their NodeDB. This is complicated by the fact that you can't DM them to ask them to do this, and the apps don't make it clear when this may be needed.

Restoring the old key isn't a solution because the problem will exist for any nodes that first saw this node when it was using the new key (and the old key may have been reset for a good reason).

A common solution in other Trust On First Use systems is to have a UI that informs the sender or recipient that the peer they are communicating with has changed keys and confirms whether they want to trust the new key.

Another option is to let nodes reset the Node ID along with its keys. That way, if you find yourself in this situation you could at least "start fresh" by resetting both your ID and your keys. This seems helpful for other reasons, like when selling or giving away used hardware the new owner wouldn't have to use your identity on the mesh and then the old and new owners wouldn't be able to (theoretically) decrypt each other's communications.

There are various common reasons why keys may need to be reset, including when addressing the "low entropy key" issue, when factory resetting, when buying or being given used hardware, or when a private key has been leaked/compromised.

It would be great if there was a means for devices in this situation to become fully viable on the network again.

This topic was previously [discussed here](https://github.com/meshtastic/firmware/discussions/7101#discussion-8480336).

### Relevant log output

```Shell

```

nightjoker7:
Addresses #8211 ("DMing permanently impaired if device resets keys") by
deterministically deriving the node's Curve25519 keypair from values tied
to the silicon rather than generating a random keypair that is lost on
factory reset.

    seed = SHA256(device_id || salt || "meshtastic-id-v1")
    priv = clamp25519(seed)
    pub  = Curve25519::eval(priv)

Where:
  device_id : the 16-byte per-chip UID already populated into
    myNodeInfo.device_id at NodeDB init time on ESP32-S3/C3/C6 (via eFuse
    BLK2 OPTIONAL_UNIQUE_ID) and nRF52840 (via FICR.DEVICEID + DEVICEADDR).
    No new hardware reads required — the code that reads it already lives
    in NodeDB at the top of installDefaultNodeDatabase().

  salt : 32 random bytes generated on first boot via the platform HW RNG
    and stored at /hwid_salt.bin on LittleFS/InternalFS. The path lives
    OUTSIDE /prefs on purpose: NodeDB::factoryReset() only removes /prefs,
    so the salt survives factory reset and the derived key remains stable
    across reset. This is the core property that fixes #8211.

## Security properties (Tier A — no eFuse burn)

| Property | Tier A |
|---|---|
| Factory reset preserves identity | ✅ |
| Firmware flash preserves identity | ✅ (filesystem untouched by flash) |
| Different chips produce different keys | ✅ (per-die UID differs) |
| No irreversible hardware provisioning step | ✅ (ships on existing fleet) |
| Resistant to firmware dump attack | ❌ — same as status quo (today the
    private key is in /prefs; Tier A stores salt beside it, so a full
    LittleFS dump reveals equivalent info). Not a regression. |
| Resistant to chip swap | ✅ (key is tied to UID, not salt alone) |

Tier B (future follow-up, out of scope for this PR) would burn a secret
into an RD_DIS'd HMAC_UP eFuse slot on ESP32-S3; that key material
physically cannot leave the die even with a full firmware dump. Requires
a one-time irreversible provisioning step and is intentionally NOT in
this patch.

## Why this matters for the XEdDSA rollout

In my separate full-stack testing of PRs #7602 + #9610, I observed that
flashing signing-enabled firmware can regenerate a node's identity on the
keygen path. In a production fleet, deploying signed firmware without
this patch means:

  - upgraded node gets a new pubkey
  - peers' first-seen-wins NodeDB cache (NodeDB.cpp:1892) rejects the
    "new" identity as a pubkey collision
  - upgraded node looks like a fresh impersonator
  - mass peer-rejection across the fleet

With this patch landed first, flashing signing firmware preserves the
pre-existing deterministic key. Peers' caches stay valid. XEdDSA rollout
becomes safe to deploy without out-of-band re-trust coordination.

So while this PR stands alone (it fixes #8211 by itself), it also
unblocks a smooth rollout of the upcoming signing work.

## Test evidence

Branch built clean on both platforms tested:
  - heltec-v4 (ESP32-S3)
  - heltec-mesh-node-t114 (nRF52840)

**Standalone persistence test (V4):**

Issued factoryReset + LoRa region config → read pubkey → repeat. Both
cycles produced bit-identical pubkey `i0eE078KATsXTJfd9XIV0CHS15bckbXnFH72KSIywCE=`,
demonstrating the salt survives factory reset on LittleFS/InternalFS and
the derivation is deterministic.

**Combined test (cf8f T114, full-stack firmware including XEdDSA):**

Same two-cycle test on a combined firmware (XEdDSA + #9610 fix-ups +
shape-detection defenses + this patch). Pubkey
`IbHJ4IIp0I7pn1rIpI2j1j1HYfSs6q+maPn5MmnDUiA=` stable across 2 factory
resets AND across a 10-minute live attack exposure. Different chip than V4
→ different pubkey, confirming per-chip isolation.

Full evidence (test harness script, raw logs, firmware MD5s):
https://github.com/nightjoker7/meshtastic-spoof-research/blob/main/evidence/phase_3/SUMMARY.md
https://github.com/nightjoker7/meshtastic-spoof-research/blob/main/evidence/phase_3/COMBINED_5LAYER_TEST.md

## Falls back gracefully

HWIdentity::deriveKey returns false on any failure path:
  - UID unavailable (e.g. bare native/portduino build)
  - FSCom write fails
  - random-source unavailable

Falls back to the existing `crypto->generateKeyPair()` random-key path
exactly as before. Zero behavior change on platforms without the
prerequisites.

## Diff summary

- `src/mesh/HWIdentity.h` — public API surface (`deriveKey`, `isAvailable`).
  ~60 lines with doc comments explaining the security model.
- `src/mesh/HWIdentity.cpp` — implementation. ~140 lines. Uses existing
  SHA256 from Crypto library, Curve25519::eval from the same source as
  the existing `regeneratePublicKey()`, and FSCom for salt persistence.
  No new dependencies.
- `src/mesh/NodeDB.cpp` — 1 include line + ~15-LOC change in
  `generateCryptoKeyPair()` to try HWIdentity::deriveKey first, fall
  back to random on failure.
- `src/modules/AdminModule.cpp` — 1 include line + 8-LOC symmetric change
  in the security config handler so admin-initiated key regen lands on
  the same stable per-chip identity.

Opt-out mechanism: since the fallback to random is always available on
failure, no explicit "disable" flag is needed — if a platform doesn't
want HWIdentity, it can ensure `myNodeInfo.device_id.size == 0` and the
code returns false, falling through to the existing path.

Happy to iterate on anything — log levels, label string, salt path,
API shape, etc.

### DIFF
--- a/src/mesh/HWIdentity.cpp
+++ b/src/mesh/HWIdentity.cpp
@@ -1,0 +1,159 @@
+#include "HWIdentity.h"
+#include "FSCommon.h"
+#include "configuration.h"
+#include "mesh/generated/meshtastic/mesh.pb.h"
+
+#include <Curve25519.h>
+#include <SHA256.h>
+#include <string.h>
+
+#ifdef ARCH_ESP32
+#include <esp_system.h>
+#endif
+
+// Forward-declare the global populated at NodeDB init time.
+extern meshtastic_MyNodeInfo myNodeInfo;
+
+namespace
+{
+// The salt lives OUTSIDE /prefs on LittleFS/InternalFS so it survives
+// NodeDB::factoryReset() which only removes /prefs. Path is short + fixed so
+// operator tools can audit it if needed.
+constexpr const char *kSaltPath = "/hwid_salt.bin";
+constexpr const char *kLabel = "meshtastic-id-v1";
+constexpr size_t kSaltSize = 32;
+
+bool readHwRandom(uint8_t *out, size_t len)
+{
+#ifdef ARCH_ESP32
+    // esp_fill_random is a true HW RNG once radio is up (entropy from ADC noise
+    // of the RF front-end). Before radio is up the RNG falls back to a
+    // ring-oscillator — still acceptable per Espressif but weaker. Our caller
+    // runs from NodeDB init which is AFTER radio is up.
+    esp_fill_random(out, len);
+    return true;
+#else
+    // Fall back to whatever Arduino random() provides. Not cryptographic on
+    // all platforms; caller should treat HWIdentity as unavailable if this
+    // path is hit on hardware that matters.
+    for (size_t i = 0; i < len; ++i) {
+        out[i] = (uint8_t)random(0, 256);
+    }
+    return true;
+#endif
+}
+
+bool loadOrGenerateSalt(uint8_t salt[kSaltSize])
+{
+#ifdef FSCom
+    // Try to read existing salt.
+    if (FSCom.exists(kSaltPath)) {
+        auto f = FSCom.open(kSaltPath, FILE_O_READ);
+        if (f) {
+            size_t n = f.read(salt, kSaltSize);
+            f.close();
+            if (n == kSaltSize) {
+                return true;
+            }
+            LOG_WARN("HWIdentity: salt file size mismatch (%u != %u), regenerating", (unsigned)n,
+                     (unsigned)kSaltSize);
+        } else {
+            LOG_WARN("HWIdentity: could not open existing salt file, regenerating");
+        }
+    }
+
+    // First-boot path: generate new salt.
+    if (!readHwRandom(salt, kSaltSize)) {
+        return false;
+    }
+    auto f = FSCom.open(kSaltPath, FILE_O_WRITE);
+    if (!f) {
+        LOG_ERROR("HWIdentity: failed to open salt file for write");
+        return false;
+    }
+    size_t written = f.write(salt, kSaltSize);
+    f.close();
+    if (written != kSaltSize) {
+        LOG_ERROR("HWIdentity: failed to persist salt (wrote %u)", (unsigned)written);
+        return false;
+    }
+    LOG_INFO("HWIdentity: generated and stored new salt at %s", kSaltPath);
+    return true;
+#else
+    (void)salt;
+    LOG_DEBUG("HWIdentity: no filesystem available, cannot persist salt");
+    return false;
+#endif
+}
+
+// Curve25519 scalar clamping per RFC 7748 §5. Callers of Curve25519::eval are
+// expected to do this on the scalar before multiplying.
+void clampCurve25519Scalar(uint8_t s[32])
+{
+    s[0] &= 248;
+    s[31] &= 127;
+    s[31] |= 64;
+}
+
+bool haveDeviceId()
+{
+    // myNodeInfo.device_id is populated during NodeDB init from the per-chip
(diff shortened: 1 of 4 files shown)


jp-bennett:
This is definitely clever, but I don't think it'll ever be acceptable to generate the x25519 keys from anything other than randomness. That device ID get's logged by manufacturers.

nightjoker7:
Thanks @jp-bennett — fair point, and it lands. Manufacturer-logged UIDs put a hard ceiling on the entropy regardless of clamping or salt, so UID-derived is the wrong primitive for a long-lived x25519 key.

Since your comment I've been on a different approach locally:

- **`PersistentIdentity`** (replacement for `HWIdentity`): on first boot, pull 32 bytes from the platform TRNG (`esp_fill_random` / `nRFCrypto.Random` / `rp2040.hwrand32` / `getrandom`), RFC 7748 §5 clamp, derive the pubkey, write the keypair to `/persistent_identity.bin` — outside `/prefs` so it survives `NodeDB::factoryReset()` but gets wiped by `factory_reset_device` (which already deletes the full FS). No UID, no KDF, no hardware-serial in the chain.
- Existing `/prefs` keys preserved as today; `loadOrGenerate` only runs when neither source has a key. Bench-tested: 3 full `factory_reset` cycles on a T114, same pubkey returns all three times.

Happy to close this PR and open a fresh one for the pure-random variant, or push a v2 onto this branch — whichever is easier for you to review. Thanks for the pushback.

PR #10288, opened 2026-04-24
fix(inkhud): scale map applet markers with font size

giannoug:
The `MapApplet` (used by the Positions applet) had all marker sizes hardcoded in pixels: an 11 px hop-count box, a radius-8 own-node bullseye, and a 12 px labeled-marker cross. On displays with a small font (6 pt, ~8 px line height) these looked fine. On the LilyGo T5-S3-ePaper-Pro, which uses a 12 pt `fontSmall` (~17 px line height), the hop-count digit overflowed its box entirely and all markers appeared as tiny specks on the 960×540 panel.

All three sizes now derive from `fontSmall.lineHeight()`, so the applet scales correctly across display variants without any per-variant tuning. Before / after screenshots attached.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other: LilyGo T5-S3-ePaper-Pro

### DIFF
--- a/src/graphics/niche/InkHUD/Applets/Bases/Map/MapApplet.cpp
+++ b/src/graphics/niche/InkHUD/Applets/Bases/Map/MapApplet.cpp
@@ -43,8 +43,8 @@ void InkHUD::MapApplet::onRender(bool full)
 
         // Add white halo outline first
         constexpr int outlinePad = 1;
-        int boxSize = 11;
-        int radius = 2; // rounded corner radius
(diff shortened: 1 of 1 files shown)


HarukiToreda:
I checked the other targets like Heltec Wireless Paper and L1 Pro Eink, no regressions. Good to commit

HarukiToreda (approved):

PR #10291, opened 2026-04-24
platformio-custom: mtjson fix dependency to prevent fake-success

vidplace7:
@mverch67 pointed out that tlora-c6 is showing "Successful" even though it's a fail.

This PR should address that! Check the CI :pray:


vidplace7:
Ayy it works :)

PR #10297, opened 2026-04-25
t5s3-epaper: Move variant.cpp -> extra_variants/variant.cpp ...again

vidplace7:
#10241 was undone by #10286 so... fixing it again :fist_raised: 

Fixes `t5s3-epaper` issues with #includes inherited from `configuration.h` when building for pioarduino (`variants/` is not included by the LDF there).

Aligns t5s3_epaper with other variants like t_deck_pro.

This is intentionally targeting `develop` even though it fixes an issue with `pioarduino` builds. Trying to keep that PR from getting too huge.


HarukiToreda:
this works

PR #10305, opened 2026-04-25
fix: new PKI_DECRYPT_FAILED response instead of NO_CHANNEL when PKI decryption fails

### LINKED ISSUE #6074: [Bug]: No channel while trying to send DM
### Contact Details

Dandcodes

### What happened?

I have a friend I DM with frequently, my application was saying I was connected to my Heltec Meshtastic device. Anytime I send a DM the cloud icon shows up, but quickly turns to a cloud with an X through it, and after clicking on it the error says "no channel" which isn't a very helpful as it doesn't describe how I can fix the potential issue. 

### App Version

2.5.16

### Phone

Pixel 6 Android 15

### Device

Heltec v3

### Firmware

2.5.20

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

llamafilm:
Fixes #6074 and discussions #6079 and #10210.

This change depends on a new protobuf definition: https://github.com/meshtastic/protobufs/pull/901

When a PKI message fails to decrypt, it previously responded with NO_CHANNEL, which is misleading at best, and arguably inaccurate.  Now it responds with PKI_DECRYPT_FAILED instead.

The current iOS app displays this unknown error code in the UI as "Empty Ack Error".  It will need an update to handle the new error code.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


jp-bennett:
This is probably the right idea. In the past we hadn't done this because hash 0x0 could be a regular channel. But now that we have completely deprecated legacy DMs, we would be even more confident it was intended to be a DM.

PR #10307, opened 2026-04-26
Update variant.h to force small icons for Heltec Wireless Tracker  V1.1

vasques666:
Force small icons for Heltec Wireless Tracker  V1.1
Default icons are too big for that screen


fifieldt:
I wonder if this is because the screen params aren't set properly - the screen is  160*80

Issue #9776, opened 2026-02-28
[Feature Request]: True multi-client support for TCP Client API (per-connection state)


mammuthus:
### Platform

ESP32

### Description

Right now the node may accept multiple TCP connections, but the Client API/PhoneAPI state is effectively shared (singleton), so multiple clients fight each other and one client silently breaks. This makes it impossible to run common setups like a persistent connection for monitoring/metrics/automation while occasionally connecting from a laptop/phone (TCP) for debugging/configuration.

A workaround like “stop advertising via mDNS” / “stop listening while connected” (see #7582) only enforces single-client behavior. It avoids surprise, but it doesn’t solve the underlying limitation.

**Request:**
Implement true multi-client TCP support by partitioning client state per connection (e.g., move “most recently read messageNum / fromnum” and any other per-client cursors out of global MeshService/PhoneAPI singleton state and into per-connection client objects). Each TCP client should be able to:
- Connect, request NodeDB/config (want_config_id / startConfig flow), and receive a consistent stream of FromRadio packets
- Maintain its own read cursor / backlog handling without impacting other clients
- Not interfere with other active clients (no silent breaks)

**Clarification:**
If full multi-client support is out of scope for resource reasons, please state the official stance explicitly in docs (single-client only) and have the firmware reject additional connections deterministically. But ideally, TCP should support multiple clients reliably.

Since #526 was closed as “enhancement” years ago, can you clarify what the final decision is today?
- Is the official stance still “single-client only” for PhoneAPI/TCP (and thus we should intentionally reject/disable additional connections as the best solution)?
- Or is true multi-client support (per-connection state) still considered a goal for the firmware at some point?

**Related:**
- #526 (single-client limitation / shared state discussion)
- #7582 (mitigation request: prevent duplicate TCP connections)

garthvh:
Has always been out of scope for microcontrollers, is already pretty clear.

mammuthus:
I see. For ESP32, multi-connection TCP is feasible: lwIP/ESP-IDF supports multiple simultaneous TCP sockets with per-connection state. The blocker here isn’t "microcontrollers", it’s PhoneAPI design + resource tradeoffs.
Right now the read cursor / flow state (messageNum/fromnum, want_config_id/startConfig/NodeDB) lives in singleton service state, so concurrent clients inevitably interfere and one can silently stop receiving.

Please pick and document one deterministic policy:
[...]

PR #10315, opened 2026-04-27
fix(api): use std::mutex for PhoneAPI nodeInfoMutex + StreamAPI streamLock

nightjoker7:
### What this changes

Replaces `concurrency::Lock` with `std::mutex` for two per-instance lock members:

- `PhoneAPI::nodeInfoMutex` — protects `nodeInfoForPhone` + `nodeInfoQueue` against concurrent BLE-callback access
- `StreamAPI::streamLock` — serializes `stream->write()` between the main packet path (`emitTxBuffer`) and the log emission path (`emitLogRecord`)

Net diff: 4 files, 22 insertions / 16 deletions.

### Why

`concurrency::Lock` has no destructor. The class was introduced in commit [`0a6059ba1`](https://github.com/meshtastic/firmware/commit/0a6059ba1) (2020-07-05) wrapping `xSemaphoreCreateBinary()` with no paired `vSemaphoreDelete()`. Each `Lock` instance that gets destructed leaks its underlying FreeRTOS semaphore handle — about 80 B per destruction.

This is harmless for `Lock` instances that live in singletons (most of them — they only destruct at program shutdown). The leak surfaces under one specific lifecycle: **TCP API stream connect/disconnect**. Each new TCP client constructs `PhoneAPI` + `StreamAPI` state including their `Lock` members, and each disconnect destructs them. Tools that reconnect on a timer (MeshMonitor's `STALE_CONNECTION_TIMEOUT=5min`, CLI loops, the web flasher) drive this lifecycle hundreds of times per hour.

Both Locks promoted here are functionally process-lifetime singletons by usage anyway:

- The mesh-wide nodeInfo dump is single-flight (project policy "officially single client only," #9776).
- The stream serializer guards a single per-process output stream.

The companion fix #10311 (Router::p_encrypted local-scope, already merged) closed the dominant leak path on the same lifecycle. This PR closes the second contributing path.

### Why std::mutex (not file-scope statics)

The earlier revision of this PR promoted both Locks to file-scope `static concurrency::Lock` members. That closed the leak by ensuring the locks never destruct, but introduced cross-transport contention that @copilot flagged in review:

> `g_streamLock` is shared across *all* `StreamAPI` instances (TCP `ServerAPI`, `SerialConsole`, `SerialModule`, etc.), not just a single TCP connection. That means a potentially blocking `stream->write()/flush()` on one stream (e.g., a slow TCP client) can stall unrelated stream outputs (e.g., serial console).

Hardware A/B on Heltec V4 (USB-attached, MediumFast channel) confirmed the regression — 8× `wantResponse=true NODEINFO_APP` requests at 12s spacing to 1-hop neighbors:

| Build | Avg response latency | Best | Pure success |
|---|---|---|---|
| File-scope-static workaround (earlier revision) | **26.5s** | 7.9s | 25% |
| Per-instance `concurrency::Lock` (the leak we're trying to fix) | 7.5s | 0.9s | 37.5% |
| **`std::mutex` (this revision)** | **1.1s** | 1.0s | 37.5% |

`std::mutex` is the right primitive for this job. The C++ standard library contract guarantees a proper RAII destructor that releases the underlying synchronization handle. No leak, no contention, no need for a workaround pattern. Per-instance semantics are preserved exactly as they were before — each `PhoneAPI` / `StreamAPI` instance has its own mutex.

### Hardware validation

- **Heltec V4 TFT** (ESP32-S3 + 2 MB PSRAM): A/B above (1.1s avg latency, 24× faster than file-scope-static workaround, ~7× faster than per-instance `Lock`).
- **Station G2** (ESP32-S3 + 8 MB PSRAM): DM round-trips clean, traceroute V4↔G2 returned promptly, no observed regressions.
- **Heltec V3** (ESP32-S3, no PSRAM, ~70 KB free heap idle — the canonical tight-RAM target): boots cleanly, normal init path, free heap healthy at ~206 KB after boot, `--info` succeeds. The build was **clean origin/develop + only the 4-file std::mutex diff**, so this confirms the change is well-behaved on the hardware tier this PR is most meant to help.

### Risk

`std::mutex` on Arduino-ESP32 maps to a FreeRTOS mutex via the C++ standard library wrapper, with proper init in the constructor and destroy in the destructor — exactly the RAII pattern we need. The semantics for the use sites here (basic critical sections, no recursive takes, no ISR access) are identical between `concurrency::Lock` and `std::mutex`.

Other `concurrency::Lock` instances in the codebase remain unchanged. They sit on non-churning lifecycles (singletons that only destruct at program shutdown) so the missing-destructor leak doesn't fire for them in practice. Eventually adding a proper `~Lock()` destructor on the class itself is the right long-term fix and would let those Locks be RAII-correct too — separate, larger PR.

### What's NOT included here

- A destructor on `concurrency::Lock` itself. That deserves its own focused synchronization-design PR (release-fence + deferred-delete to handle in-flight takers).
- Any change to `Lock` users on non-churning lifecycles (SPI, crypto, nrf52SAADC, etc).
- The platform pin discussion that came up during testing — that's an unrelated upstream USB-Serial-JTAG bug to be filed separately.

### DIFF
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -15,7 +15,7 @@
 #include "Router.h"
 #include "SPILock.h"
 #include "TypeConversions.h"
-#include <mutex>
+#include "concurrency/LockGuard.h"
 #include "main.h"
 #include "modules/NodeInfoModule.h"
 #include "xmodem.h"
@@ -78,7 +78,7 @@ void PhoneAPI::handleStartConfig()
     LOG_INFO("Start API client config millis=%u", millis());
     // Protect against concurrent BLE callbacks: they run in NimBLE's FreeRTOS task and also touch nodeInfoQueue.
     {
-        std::lock_guard<std::mutex> guard(nodeInfoMutex);
+        concurrency::LockGuard guard(&nodeInfoMutex);
         nodeInfoForPhone = {};
         nodeInfoQueue.clear();
     }
@@ -117,7 +117,7 @@ void PhoneAPI::close()
         toRadioScratch = {};
         // Clear cached node info under lock because NimBLE callbacks can still be draining it.
         {
-            std::lock_guard<std::mutex> guard(nodeInfoMutex);
+            concurrency::LockGuard guard(&nodeInfoMutex);
             nodeInfoForPhone = {};
             nodeInfoQueue.clear();
         }
@@ -291,14 +291,14 @@ size_t PhoneAPI::getFromRadio(uint8_t *buf)
             info.has_hops_away = false;
             info.is_favorite = true;
             {
-                std::lock_guard<std::mutex> guard(nodeInfoMutex);
+                concurrency::LockGuard guard(&nodeInfoMutex);
                 nodeInfoForPhone = info;
             }
             fromRadioScratch.which_payload_variant = meshtastic_FromRadio_node_info_tag;
             fromRadioScratch.node_info = info;
             // Should allow us to resume sending NodeInfo in STATE_SEND_OTHER_NODEINFOS
             {
-                std::lock_guard<std::mutex> guard(nodeInfoMutex);
+                concurrency::LockGuard guard(&nodeInfoMutex);
                 nodeInfoForPhone.num = 0;
             }
         }
@@ -502,7 +502,7 @@ size_t PhoneAPI::getFromRadio(uint8_t *buf)
 
         meshtastic_NodeInfo infoToSend = {};
         {
-            std::lock_guard<std::mutex> guard(nodeInfoMutex);
+            concurrency::LockGuard guard(&nodeInfoMutex);
             if (nodeInfoForPhone.num == 0 && !nodeInfoQueue.empty()) {
                 // Serve the next cached node without re-reading from the DB iterator.
                 nodeInfoForPhone = nodeInfoQueue.front();
@@ -532,7 +532,7 @@ size_t PhoneAPI::getFromRadio(uint8_t *buf)
             prefetchNodeInfos();
         } else {
             LOG_DEBUG("Done sending %d of %d nodeinfos millis=%u", readIndex, nodeDB->getNumMeshNodes(), millis());
-            std::lock_guard<std::mutex> guard(nodeInfoMutex);
+            concurrency::LockGuard guard(&nodeInfoMutex);
             nodeInfoQueue.clear();
             state = STATE_SEND_FILEMANIFEST;
             // Go ahead and send that ID right now
@@ -659,7 +659,7 @@ void PhoneAPI::prefetchNodeInfos()
     bool added = false;
     // Keep the queue topped up so BLE reads stay responsive even if DB fetches take a moment.
     {
-        std::lock_guard<std::mutex> guard(nodeInfoMutex);
+        concurrency::LockGuard guard(&nodeInfoMutex);
         while (nodeInfoQueue.size() < kNodePrefetchDepth) {
             auto nextNode = nodeDB->readNextMeshNode(readIndex);
             if (!nextNode)
@@ -717,7 +717,7 @@ bool PhoneAPI::available()
         return true;
 
     case STATE_SEND_OTHER_NODEINFOS: {
-        std::lock_guard<std::mutex> guard(nodeInfoMutex);
+        concurrency::LockGuard guard(&nodeInfoMutex);
         if (nodeInfoQueue.empty()) {
             // Drop the lock before prefetching; prefetchNodeInfos() will re-acquire it.
             goto PREFETCH_NODEINFO;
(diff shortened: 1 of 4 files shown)


nightjoker7:
Hardware A/B on a Heltec V4 (USB-attached, talking to the local mesh on a busy MediumFast channel):

| Build | NodeInfo-request avg response latency | Best response | Pure success | Silent miss |
|---|---|---|---|---|
| **This PR (lock-as-globals)** | **26.5s** | 7.9s | 25% (2/8) | 62.5% (5/8) |
| **Per-instance Locks (revert)** | **7.5s** | 0.9s | 37.5% (3/8) | 50% (4/8) |

Same firmware, same workload (8× `wantResponse=true` `NODEINFO_APP` requests at 12 s spacing to 1-hop neighbors HILL, 2860, MagicMattOBS), only difference is the lock-as-globals revert. Logs: `_lockglobals_ab_BEFORE_*.log` / `_lockglobals_ab_AFTER_*.log`.

**Copilot's cross-transport-contention concern in the review is real and measurable** — 3.5× latency regression on responses that *do* arrive. The phone-side timeout fires before the response gets through the contended `g_streamLock` / `g_nodeInfoMutex`, which surfaces as `MAX_RETRANSMIT` in the Android log even though the response actually arrives a few seconds later.

Two ways to address this without giving up the leak fix:

1. **Narrow the scope to `ServerAPI` (TCP only).** That's the only `StreamAPI`/`PhoneAPI` subclass with a churning lifecycle — `SerialConsole`, `SerialModule`, `BluetoothPhoneAPI`, `HttpAPI`, `PacketAPI` are all singletons that never destruct, so they don't actually leak. A Lock-pointer member with a per-instance default and a `ServerAPI` subclass override would close the leak path where it fires and remove the contention on the singleton paths. ~25 lines, but it introduces a pattern (pointer + override) that doesn't exist elsewhere in the tree.

2. **Close this PR and pursue the principled fix instead** — a deferred-delete `~Lock()` destructor on `concurrency::Lock` itself, with a release-fence so concurrent takers drain safely. Separate, larger PR. The merged #10311 already closed the dominant portion of the original leak; the residual ~80 B/cycle from the missing Lock destructor is real but small for typical use, and the destructor PR is the right place to fix it for good.

Happy to do either. Option 1 I have ready locally; option 2 I'd start as a follow-up after #10311 settles. Which way do you want to go?

nightjoker7:
Force-pushed a clean replacement based on the A/B feedback above. Summary of the pivot:

**Original approach** (file-scope-static `g_nodeInfoMutex` / `g_streamLock`): closed the leak but caused cross-transport contention. Hardware A/B on V4 measured **3.5× latency regression** on NodeInfo responses. @copilot's review concern was correct.

**New approach** (this revision): replace `concurrency::Lock` with `std::mutex` at the same two per-instance use sites. C++ standard library contract guarantees a proper RAII destructor → no leak, no need for the file-scope workaround, no cross-transport contention.

**Hardware A/B comparison** (8 × `wantResponse=true NODEINFO_APP` requests at 12s spacing to 1-hop neighbors on busy MediumFast channel):

| Build | Avg latency | Best | Pure success |
|---|---|---|---|
| File-scope-static (earlier revision) | 26.5s | 7.9s | 25% |
| Per-instance `concurrency::Lock` (no fix) | 7.5s | 0.9s | 37.5% |
| **`std::mutex` (this revision)** | **1.1s** | 1.0s | 37.5% |

24× faster than the file-scope-static workaround, ~7× faster than per-instance `concurrency::Lock`. The std::mutex implementation on Arduino-ESP32 is just a thin wrapper around FreeRTOS mutex with proper init/destroy.

**Validated on three hardware tiers:**
- Heltec V4 TFT (ESP32-S3 + 2 MB PSRAM) — the A/B above
- Station G2 (ESP32-S3 + 8 MB PSRAM) — DM/traceroute clean
- Heltec V3 (ESP32-S3, no PSRAM, ~70 KB free heap — the canonical tight-RAM target this PR is most meant to help) — clean develop + this PR's 4-file diff only, boots cleanly, free heap healthy ~206 KB, `--info` succeeds

The `concurrency::Lock` class itself is untouched — adding a proper `~Lock()` destructor with release-fence + deferred-delete is its own focused PR (cancel of the file-scope workaround approach).

nightjoker7:
Closing — investigation since the last force-push revealed the locking infrastructure I was trying to add (`nodeInfoMutex`, `streamLock`, their `LockGuard` call sites) is already in `develop` via #8354 and 8e635bc2f. After reverting the `std::mutex` pivot (which broke cross-platform compile because `std::mutex` isn't available in the bare-metal ARM-GCC newlib STL on nRF52/STM32/RP2), this PR's diff against develop was 7 lines of comments and zero functional change.

The residual issue is the per-destruction `~80 B` `concurrency::Lock` semaphore-handle leak. `Lock.cpp` documents that closing it requires lifecycle work in PhoneAPI/StreamAPI `close()` (drain holders before destruction so `vSemaphoreDelete` is safe). That's a different, larger PR — happy to revisit if the lifecycle drain ever lands first.

Sorry for the noise.

Issue #4410, opened 2024-08-07
[BUG] Infinite reboot loop after configuring as REPEATER


vk2diy:
### Category

Other

### Hardware

T-Beam

### Firmware Version

2.4.0.46d7b82

### Description

1. Get new device.
2. Update to latest stable firmware with web flasher.
3. Install app on phone.
4. Can't see anyone, so enable up/down and MQTT
5. Successful interaction
6. Try turning off phone proxying, up/down and MQTT and ignoring MQTT.
7. Can't see anyone.
8. Try changing system over to either REPEATER mode
9. System gets locked in infinite reboot/crash loop as follows.

```
DEBUG | ??:??:?? 36 [Router] Forwarding to phone (id=0x68ab6268 fr=0x18 to=0x18, WantAck=0, HopLim=3 Ch=0x0 Portnum=6 requestId=91fb59fe priority=70)
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x400d36b9  PS      : 0x00060230  A0      : 0x80119ff2  A1      : 0x3ffd85f0  
A2      : 0x3ffe3164  A3      : 0x3ffc8d4c  A4      : 0x3ffc8468  A5      : 0x000009bc  
A6      : 0x000009bc  A7      : 0x00000000  A8      : 0x00000100  A9      : 0x3ffd85c0  
A10     : 0x00000001  A11     : 0x00000000  A12     : 0x00000004  A13     : 0x3ffe329c  
A14     : 0x00000000  A15     : 0x00000000  SAR     : 0x00000019  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000161  LBEG    : 0x40093e14  LEND    : 0x40093e2a  LCOUNT  : 0x00000000  


Backtrace: 0x400d36b6:0x3ffd85f0 0x40119fef:0x3ffd8610 0x4011778d:0x3ffd8630 0x4012ae65:0x3ffd8690 0x4012b0fd:0x3ffd86e0 0x4013b4f6:0x3ffd8700 0x401324f9:0x3ffd8720 0x400e2df5:0x3ffd8760




ELF file SHA256: 5daac71e23ca1618

E (40526) esp_core_dump_flash: Core dump flash config is corrupted! CRC=0x7bd5c66f instead of 0x0
Rebooting...
```
10. Re-flash the thing
11. Infinite reboot continues.

### Relevant log output

```Shell
[...]

Issue #1355, opened 2022-04-05
[Bug]: Crash in JSON MQTT bridge


caveman99:
### Category

Hardware Compatibility

### Hardware

T-Lora v2 1.6

### Firmware Version

1.3.5.998c90d3

### Description

While running on Wifi and unencrypted MQTT bridge, the JSON code crashes from time to time. Most of the encodes go fine.

### Relevant log output

```Shell
13:49:12 182 [Router] Should encrypt MQTT?: 0
13:49:12 182 [Router] publish msh/1/c/LongFast/!25c28464, 84 bytes
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.
Core 1 register dump:
PC      : 0x400eeba7  PS      : 0x00060430  A0      : 0x800ef2b0  A1      : 0x3ffd5880
A2      : 0x00000000  A3      : 0x00000004  A4      : 0x3ffecab8  A5      : 0x00000043
A6      : 0x00000000  A7      : 0x3ffd5880  A8      : 0x800eeba0  A9      : 0x3ffd5850
A10     : 0x00000001  A11     : 0x00000011  A12     : 0x3f405508  A13     : 0x3ffd5a74
A14     : 0x00000008  A15     : 0x00000002  SAR     : 0x00000020  EXCCAUSE: 0x0000001c
EXCVADDR: 0x00000004  LBEG    : 0x4000c46c  LEND    : 0x4000c477  LCOUNT  : 0x00000000

ELF file SHA256: 0000000000000000

Backtrace: 0x400eeba7:0x3ffd5880 0x400ef2ad:0x3ffd5c00 0x400e2191:0x3ffd5c60 0x400de255:0x3ffd5c80 0x400e1b43:0x3ffd5ca0 0x400ebafc:0x3ffd5cc0 0x400ebaae:0x3ffd5ce0 0x400de731:0x3ffd5d40 0x400e2221:0x3ffd5d70 0x400e22f1:0x3ffd5d90 0x400e2311:0x3ffd5db0 0x400de0b5:0x3ffd5de0 0x400d515e:0x3ffd5e00 0x400f56c1:0x3ffd5e20 0x400dd415:0x3ffd5e50 0x40127591:0x3ffd5e70

stack:
0x400eeba7: MQTT::downstreamPacketToJson(_MeshPacket*) at /root/fw/Meshtastic-device/src/mqtt/MQTT.cpp:293
[...]

PR #10316, opened 2026-04-27
fix: null-check packetPool / queueStatusPool allocCopy across 11 router/module callers

nightjoker7:
### What this changes

Adds null-checks at 11 call sites that take a pointer from `packetPool.allocCopy()` or `queueStatusPool.allocCopy()` and immediately dereference (or pass to a function that derefs) without verifying the allocation succeeded. `Allocator::allocCopy` is documented to return `nullptr` on pool exhaustion — these call sites assumed "always succeeds."

Total diff: 6 files, 49 insertions / 17 deletions. No behavioral change on the happy path.

| File:line | Pre-fix |
| --- | --- |
| `src/mesh/MeshService.cpp:113` | `sendToPhone(packetPool.allocCopy(*mp))` |
| `src/mesh/MeshService.cpp:209` | `auto a = packetPool.allocCopy(p); sendToMesh(a, RX_SRC_USER);` |
| `src/mesh/MeshService.cpp:233` | `meshtastic_QueueStatus *copied = queueStatusPool.allocCopy(qs); copied->res = res;` |
| `src/mesh/MeshService.cpp:271` | `auto a = packetPool.allocCopy(*p); sendToPhone(a);` |
| `src/mesh/NextHopRouter.cpp:35` | `startRetransmission(packetPool.allocCopy(*p));` |
| `src/mesh/NextHopRouter.cpp:151` | `tosend = packetPool.allocCopy(*p); tosend->hop_limit = 0;` |
| `src/mesh/NextHopRouter.cpp:329, 331, 336` | three retransmission `allocCopy` sites passed straight to `*::send(...)` |
| `src/mesh/ReliableRouter.cpp:21` | `auto copy = packetPool.allocCopy(*p); startRetransmission(copy, NUM_RELIABLE_RETX);` |
| `src/mesh/Router.cpp:376` | `p_decoded = packetPool.allocCopy(*p); ... mqtt->onSend(*p, *p_decoded, chIndex);` |
| `src/modules/NodeInfoModule.cpp:63` | `packetCopy = packetPool.allocCopy(mp); packetCopy->decoded.payload.size = ...` |
| `src/modules/AtakPluginModule.cpp:199` | `decompressedCopy = packetPool.allocCopy(mp); decompressedCopy->decoded.payload.size = ...` |

### Affected hardware

The bug fires when `packetPool` is exhausted — i.e., when the heap-backed allocator can't fulfill a `MeshPacket` allocation (~400 B with metadata). Pool size is bounded by `MAX_PACKETS = MAX_RX_TOPHONE(32) + MAX_RX_FROMRADIO(4) + 2*MAX_TX_QUEUE(16) + 2 = 70` slots, but the practical exhaustion trigger is **free heap dropping below the per-allocation cost under burst load**, not slot count.

Boards most exposed are those without PSRAM and with tight idle free heap:

- **Heltec V1 / V2 / V3 LoRa-32** — ESP32 / ESP32-S3 with no PSRAM, ~70-80 KB free heap idle
- **T-Beam classic** — ESP32 with ~80 KB free idle
- **T-Lora v2 1.6** — ESP32 with ~90 KB free idle
- **RAK4631** — nRF52840 with 256 KB total RAM, ~60 KB free idle

PSRAM boards (Heltec V4 TFT, Tracker, Station G2, T-Beam-S3 family, T-Watch-S3) are harder to exhaust because the allocator can back onto 8 MB PSRAM, but the same crash mechanism applies if pool exhaustion ever happens.

### Backtraces this addresses

Two existing field reports have decoded backtraces that land directly on the call sites this PR is guarding:

- **#1355** — `Crash in JSON MQTT bridge`. `Guru Meditation Error: LoadProhibited`, `EXCVADDR=0x00000004` (textbook null-deref of a struct field at offset 4). Decoded backtrace frame 0 is `MQTT::downstreamPacketToJson`, called via `MQTT::onSend` from `Router::send` — the exact path guarded by the `Router.cpp:376` fix here (`if (p_decoded && ...)`). Filed by @caveman99.

- **#4410** — `Crash/reboot loop entering 'Settings/Device' using Android app after configuring as REPEATER`. @GUVWAF decoded the backtrace in-thread; frame 0 is `MeshService::sendToPhone`, which is the call site of `packetPool.allocCopy(*mp)` at `MeshService.cpp:113`. GUVWAF noted "not sure why it's crashing" — the unguarded null deref on pool exhaustion is a plausible cause, though not definitively proven from the trace alone.

I'm not claiming this PR "fixes" either issue (both are closed); I'm claiming the call graphs in their decoded backtraces match the call sites being guarded here. If pool exhaustion is the trigger, the symptom shifts from a panic to a logged drop after this change.

### Reproduction

Used a deterministic null-injection harness on **Heltec V3** (ESP32-S3, no PSRAM, the same hardware tier as the cited reports): a 1-line edit in `MemoryPool.h::allocCopy` returning `nullptr` every 5th call. Drove sendText load via `meshtastic-python SerialInterface` for 90 s on two devices flashed identically except for these patches.

| Build | rebootCount before / after | Verdict |
| --- | --- | --- |
| Baseline (`origin/develop` + injection) | 4 → 26 | **+22 panics in 90 s** |
| This patch (`develop` + 11 sites + injection) | 3 → 4 | **+1 (post-test reopen artifact, identical on baseline)** |

Same hardware, same workload, same injection rate, same channel. The only difference is the 11 null-checks.

### Why the diff is small

Mirrors the merge pattern of #10262 (13 `allocZeroed` null-checks in one focused PR). Same shape, adjacent class. Sites that already had null-checks are left alone — `Router.cpp:744` (legacy guard from #9136), `Router.cpp:813` (existing `if (p_encrypted_new)`), and the Telemetry consumers (`if (!lastMeasurementPacket)`).

### Risk

Minimal. Each fix is an additive `if (auto *c = ...) ...;` or `if (!ptr) { LOG_WARN; early-return; }`. The only runtime change vs baseline is "pool exhaustion now logs and drops one packet instead of crashing the radio task." No public-API surface change.

### Follow-up

A round-2 follow-through covers helper-level fixes (`Router::allocForSending`, `MeshModule::allocAckNak`, `SinglePortModule::allocDataPacket`, `ProtobufModule::allocDataProtobuf`) plus 10 module callers (RoutingModule, NeighborInfoModule, PositionModule, StatusMessageModule, the 5 Telemetry modules + HostMetrics). Held for a separate PR to keep this diff focused; the 11 sites here are the load-bearing ones in the cited backtraces.


nightjoker7:
Pushed `67bc2665d` addressing the Copilot review on `0984d68`:

- **`AtakPluginModule.cpp:205`** — fixed `sizeof(payload)` → `sizeof(payload.bytes)`. `PB_BYTES_ARRAY_T` is `{size + bytes[N]}`, so the previous form overstated the buffer passed to `pb_encode_to_bytes` by the size-field width and could let it write past `payload.bytes` into adjacent fields under a worst-case TAKPacket re-encode. Now matches the canonical form already used at `NodeInfoModule.cpp:67`. Nice catch by Copilot.
- 4× caller-side `LOG_WARN` → `LOG_DEBUG` (`MeshService.cpp:116`, `MeshService.cpp:235`, `AtakPluginModule.cpp:201`, `NodeInfoModule.cpp:71`). The allocator and the underlying static pool already log WARN on failure (`MemoryPool.h:24,45,158`), so each pool exhaustion was producing 3 lines. Caller-side breadcrumb is preserved for debug builds.
- `MeshService.cpp:281` — reworded the cc-to-phone exhaustion comment. Previous wording implied future re-delivery via the RX path, which isn't guaranteed for non-broadcast / non-local sends.

Verified locally: `pio test -e coverage -f test_atak -f test_default -f test_meshpacket_serializer` → 41/41 pass. Diff is +9/-7 across 3 files.

PR #10319, opened 2026-04-27
Actions: Build MacOS binary

vidplace7:
Preliminary CI for the MacOS builds

Currently building for
- MacOS 26 (ARM64)
- MacOS 15 (ARM64)

MacOS Intel is ignored :skull:


jp-bennett:
Might as well YOLO this in. No real way to test it otherwise.

PR #10330, opened 2026-04-29
fix: MQTT connection on Portduino/Linux native nodes

WB3IHY:
## Problem
When running meshtasticd as a software-only node (no radio hardware) 
on Linux/Portduino, MQTT never connects at boot despite correct 
configuration. The node shows "MQTT server on a private IP" but never 
attempts a connection.

## Root Cause
`isConnectedToNetwork()` in MQTT.cpp always returns `false` on 
ARCH_PORTDUINO because none of HAS_WIFI, HAS_ETHERNET, or USE_WS5500 
are defined for Linux native builds. This causes `wantsLink()` to 
always return `false`, which causes `runOnce()` to immediately disable 
the MQTT thread on every boot.

## Fix
Return `true` for ARCH_PORTDUINO since Linux always has network 
access available through the host OS.

## Testing
Tested on Ubuntu 24.04 running meshtasticd as a cloud-based 
software-only MQTT node. MQTT now connects automatically at boot 
without any manual intervention.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -350,6 +350,8 @@ inline bool isConnectedToNetwork()
     return WiFi.isConnected();
 #elif HAS_ETHERNET
     return Ethernet.linkStatus() == LinkON;
+#elif defined(ARCH_PORTDUINO)
+    return true;
 #else
     return false;
 #endif


WB3IHY:
Hi, I'm a first-time contributor. Could a maintainer please take a look at the CI failures? I'm not sure if they're related to my change or environment issues on your end. Happy to make any adjustments needed.

jp-bennett:
I'm still unsure why only some users run into this bug... But this seems like a straightforward fix.

PR #10342, opened 2026-04-29
Increase QoL by adding subtle animation between screens.

CodeKing547:
Adds a subtle animation when scrolling through screens. Looks really nice.

  - [x] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [x] LilyGo T-Beam
  - [x] RAK WisBlock 4631
   (tested on L1 Pro tracker as well)

### DIFF
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -677,7 +677,7 @@ void Screen::setup()
     displayWidth = dispdev->width();
     displayHeight = dispdev->height();
 
-    ui->setTimePerTransition(0);           // Disable animation delays
+    ui->setTimePerTransition(200);         // Subtle slide between manually selected frames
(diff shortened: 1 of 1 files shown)


HarukiToreda:
Sorry, but this PR cannot be approved.

When BaseUI was designed, it was built to support the current displays (OLED, TFT, and eInk). To avoid compatibility issues, animations were intentionally disabled. Early testing showed multiple crashes, increased memory usage, and UI elements not behaving correctly since they were not designed for animation.

Enabling this would require a more comprehensive refactor of BaseUI, which is not currently planned. We can’t compromise stability for visual changes at the moment.

https://github.com/user-attachments/assets/81c0bc0d-a193-45bf-9d06-c187f079867e

PR #10351, opened 2026-04-30
Initial implementation of serialHal

jp-bennett:
Depends on https://github.com/meshtastic/protobufs/pull/904

This adds two big chunks of code, half of which only lives inside the Portduino platform. It gives handheld nodes the ability to work as a proxy for radiolib running on Native. In other words, a node can effectively emulate a ch341 device.

There's also a pretty obvious path here to build a Meshtastic-micro firmware, that *only* provides this functionality. This should run on devices that were previously completely unworkable, and at least provide some use-case for them.

Lots of copilot usage here, and probably a few bits where code can be cleaned up as a result, but overall seems pretty sane. 

## Summary by CodeRabbit

* **New Features**
  * Added a serial-based LoRa backend for Portduino with configurable device, baud rate, and timeout.
  * Added framed Serial HAL support for GPIO, interrupts, and SPI operations.
  * Added controls to suppress logging during Serial HAL transactions.

* **Bug Fixes**
  * Improved Serial HAL frame detection, routing, and response handling.
  * Updated serial-only startup to skip unnecessary radio initialization and errors.
  * Improved GPIO and SPI setup when using the serial backend.


src/mesh/SerialHalDevice.cpp
```diff
@@ -0,0 +1,380 @@
...
+static inline void setResponseError(meshtastic_SerialHalResponse &response, meshtastic_SerialHalResponse_Result result,
+                                    const char *error = nullptr)
+{
+    response.result = result;
+    if (error != nullptr) {
+        snprintf(response.error, sizeof(response.error), "%s", error);
+    }
+}
+
+void SerialHalDevice::handleCommand(const uint8_t *buf, size_t len, StreamAPI *streamApi)
```
jp-bennett, line 183: It would be possible to just check for an existing radiolib instance here, and delete it if found. This could be a parallel or alternative approach to the config.lora config entry.

SpudGunMan:
`There's also a pretty obvious path here to build a Meshtastic-micro firmware,`

Why not boot into the micro mode like tdek boots into ble/basic mode?

madeofstown:
I would love to see this feature! Just looking up last night how to emulate a ch341 with an esp32, glad to know it's already in the process of being implemented!

jp-bennett:
> `There's also a pretty obvious path here to build a Meshtastic-micro firmware,`
> 
> Why not boot into the micro mode like tdek boots into ble/basic mode?

The idea is to use a micro firmware for devices we can't support with the full-fat Meshtastic firmware.

NomDeTom:
Flash, head merged with develop `3468af94a` (CI runner): rak4631 **+3,480 B**, nrf52_promicro_diy_tcxo **+3,512 B**, static RAM +144 B — nearly all of it the device-side `SerialHalDevice.cpp` + StreamAPI hooks compiled into every target; behind a `HAS_SERIAL_HAL_DEVICE` build flag (default off) it measures **+80 / +152 B** ([branch](https://github.com/NomDeTom/MeshtasticFirmware/tree/serialhal-device-gate)).

A13xB0:
bump

PR #10356, opened 2026-04-30
Fix MAC_from_string to use input parameter instead of global config for MAC address parsing

thebentern:
This seems like a pretty big oversight. Ran into it while doing testing my TCP connection on the local MCP server on MacOS native


jp-bennett (approved):

jp-bennett:
This was a definite case of it-works-on-my-machine. Good hardening

PR #10365, opened 2026-05-01
fix: INA226 sensor detection

Krokodilushka:
## Problem
The firmware attempted to read data from an INA226 power sensor as if it were an INA219. Due to an incorrect conditional, any supported INA* sensor was incorrectly identified as an INA219.

Related logs:
```
INFO  | ??:??:?? 0 Scan for i2c devices
DEBUG | ??:??:?? 0 Scan for I2C devices on port 1
DEBUG | ??:??:?? 0 Register value from 0x40: 0x5449
DEBUG | ??:??:?? 0 Register MFG_UID: 0x5449
DEBUG | ??:??:?? 0 Register value from 0x40: 0x2260
DEBUG | ??:??:?? 0 Register DIE_UID: 0x2260
INFO  | ??:??:?? 0 INA226 found at address 0x40
INFO  | ??:??:?? 0 INA219 found at address 0x40
INFO  | ??:??:?? 0 Device found at address 0x47 was not able to be enumerated
INFO  | ??:??:?? 0 1 I2C devices found
```
## Fix
Fixed the incorrect conditional used for sensor detection.

## Hardware
ESP32-C6 + E22900M30S + INA226 + BMP580

### DIFF
--- a/src/detect/ScanI2CTwoWire.cpp
+++ b/src/detect/ScanI2CTwoWire.cpp
@@ -442,15 +442,17 @@ void ScanI2CTwoWire::scanPort(I2CPort port, uint8_t *address, uint8_t asize)
                         type = INA260;
                     }
                 }
+                if (type == NONE) {
 #if HAS_TELEMETRY && !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR
-                if (type == NONE && detectSHT21SerialNumber(i2cBus, (uint8_t)addr.address)) {
-                    logFoundDevice("SHTXX (SHT2X)", (uint8_t)addr.address);
-                    type = SHTXX;
-                }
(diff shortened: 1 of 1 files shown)


oscgonfer:
Note to self: check with #10482

Finn10111:
Thanks, this seems to have fixed my issue. I didn't check my logs but the sensors sometimes reported wrong values (like 1V instead of around 3V).

oscgonfer:
I have done INAs now with me, but not so much time. Is this one solved @Finn10111 ? With this PR or with #10482 ?

Finn10111:
I once manually merged this one and built the firmware, it fixed my issue. But in recent firmware versions it already works again because of the other PR I think

oscgonfer:
I'll test with the other one and see if there's anything else! Thanks for quick feedback @Finn10111 !

Issue #10084, opened 2026-04-05
[Feature Request] Generic nRF52 LPCOMP solar wake-up, strict external radio power management, and user-defined voltage thresholds


EA2OY:
### Platform

NRF52

### Description

This is a feature request to implement a robust power management architecture for nRF52-based boards utilizing external LoRa modules (like the EBYTE E22P series) and solar harvesting. It aims to move away from hardcoded low-voltage thresholds to a fully dynamic approach applicable to any radio module with a sleep/enable pin.

1. Strict External Radio Power Management:
Currently, modules with active internal pull-ups (like the E22P) can suffer from backpowering or phantom boot-ups when the nRF52 enters SYSTEM_OFF if the enable pin is left floating (INPUT).

Proposal: Introduce a generic EXTERNAL_RADIO_ENABLE_PIN macro. Board designers/DIYers would simply wire their radio's EN/RXEN pin to a free GPIO and define this macro in their variant.h.

Modify cpuDeepSleep() to force this pin to OUTPUT and LOW, actively pulling it to GND during deep sleep to completely eliminate parasitic drain from external PA/LNA modules.

2. Dynamic LPCOMP Solar Resurrection:

Proposal: Integrate the nRF52 LPCOMP hardware to wake the node from SYSTEM_OFF via the DETECT_UP event on a designated analog pin (e.g., BATTERY_LPCOMP_INPUT).

Include a helper function that dynamically maps a target mV wake threshold to the closest nrf_lpcomp_ref_t fraction (1/8 to 15/16) based on the specific board's resistor divider ratio (defined in variant.h).

3. User-Defined Voltage Thresholds (App Integration):

Proposal: Replace the hardcoded OCV_ARRAY shutdown threshold with user-configurable values via the App: config.power.deep_sleep_voltage_mv and config.power.wake_voltage_mv (would require a PR in the protobufs repo).

The low_voltage_counter logic would evaluate against the user-defined deep_sleep_voltage_mv before triggering EVENT_LOW_BATTERY.

4. Safe Regulatory Overrides for High-Power Modules:

Proposal: Allow boards with sensitive external PAs to securely define SX126X_MAX_POWER and LORA_DUTY_CYCLE_PERCENT in their variant.h to override regional defaults natively (e.g., EU_868) without breaking global compliance for other standard boards.

Motivation and Context:
[...]

Issue #9911, opened 2026-03-15
[Feature Request]: Heltec power management. Sleeping Beauty mode.


roberthadow:
### Platform

ESP32

### Description

There is lots of talk and speculation on Reddit about power management for the Heltec ESP32 devices.  They got this far, but I cannot swear this is all true.

When the Heltec unit runs down to 2.7 V, they say, it boot loops, running down to 2.5 V, then it turns off.  Even if the sun comes out and the solar panel starts charging again, the device needs a manual reboot to start again.

Some people suggested that it's a power issue.  No matter the size of the battery, the small wires to a PH1.25 plug aren't substantial enough for the current at the start of a V4 transmit cycle.  A capacitor added across the battery leads will supply the instantaneous current needed.  One suggested a 6000 mAh battery as a minimum for V4. That's ludicrous.

For my part, I can tell you that the V4 won't run all night.  When I retrieve the unit after the sun comes up and charging starts, the unit still needs a manual reboot.

A few months ago, I posted a feature request called the "Sleeping Beauty" mode.  It should be relatively easy to realize in firmware.  When the battery voltage drops below Vcritical, the unit shuts off all display, Lora, WiFi, and Bluetooth functions.  It operates a minimum power configuration that does only one thing.  It polls the battery voltage every n seconds.  Default should be 3600 seconds. Only when V increases to Voperating it repowers all the other functions.  Sleeping Beauty is awoken. 

An indication of charging should not be enough.  A monentary car headlight could trigger the charging indication.  The unit still doesn't have enough power to come back to life. Only when you know you have enough juice to reboot and run for an hour (or whatever) the unit wakes up.

linuxslate:
Right now, my HelTec V4 is 30 ft up a pine tree, and is unresponsive after depleting the battery on the first night.

I agree with the above post.  The V4 has no hardware LVC.  The CN3165 is a charge controller, not a battery manager.  The datasheet for the CN3165 does not mention low voltage or depletion.

In the current firmware, battery voltage is known, but not acted upon.

The ESP32-S3 does have a very deep sleep mode, with draw on the order of a few uA.  With the option to set a Vcritical, the device would enter the deepest possible sleep mode, with periodic wakes to check voltage, as @roberthadow describes.

The Vcritical should be user configurable so that the user can determine if they want to prioritize staying up as long as possible, or prioritize the ability to reliably return from a shutdown without physical access to the device.

Nerdtality:
[...]

Issue #9699, opened 2026-02-20
nRF52 Solar Nodes: Inconsistent wake-up behavior and brownout risks


JBAU92:
### Category

Other

### Hardware

DIY

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.20

### Description

Currently, nRF52-based targets (especially XIAO nRF52840 and RAK4631) suffer from a "Deep Sleep Coma" when powered by solar panels. The nodes deplete the battery down to ~3.1V, where the LDO often fails to maintain a stable 3.3V rail, leading to brownouts or unrecoverable SYSTEM_OFF states.

**Technical Analysis**

1. OCV Curve: The default OCV_ARRAY is too aggressive for unattended solar nodes, leaving zero safety margin for the regulator before shutdown. 
2. LPCOMP: Hardware wake-up via LPCOMP is either missing (XIAO) or not correctly aligned with the discharge curve, preventing the node from resetting when solar power returns.

**Proposed Solution**

I have validated a fix in my fork that:Redefines the 0% battery threshold to 3.4V to ensure a clean shutdown and leave "start-up" energy for the next day.
Enables/Aligns LPCOMP (NnRF52 hardware comparator) to trigger a wake-up only when $V_{bat} > 3.7V$.

I have already compiled and tested these changes for RAK4631 and XIAO nRF52840 with successful autonomous recovery. I am willing to submit a PR if the team is interested in this power management optimization.

### Relevant log output

```Shell

```

fifieldt:
@HarukiToreda ,power god, how do you feel about diff OCV curves for solar variants?

JBAU92:
Different OCV curves for solar variants are necessary, but they are only half the solution.

Here is the technical reality of why these nodes are falling into a "Deep Sleep Coma":

1. The Brownout Trap (OCV Curve)
The default OCV curve pushes the hardware down to ~3.10V. For unattended solar nodes, this is fatal. At this level, the LDO struggles to maintain a stable 3.3V rail. A LoRa TX spike causes an immediate voltage drop, leading to brownouts and bootloops before the PowerFSM can trigger a clean SYSTEM_OFF. I've set my solar OCV 0% at 3.4V, which guarantees a safe software shutdown and leaves a minimal buffer for the morning sun.

[...]

Issue #5199, opened 2024-10-31
[Feature Request]: App configurable deep sleep voltage


Dude5101:
### Platform

ESP32

### Description

Heltec V3 and other Heltec boards become unstable when battery voltage drops below 3.3v
Current firmware goes into deep sleep when voltage is less than 3.1v which causes boot loop/brownout issues on Heltec boards.

Please add an option to change deep sleep voltage to a user-configurable value.

Current workaround I am using is to adjust ADC override down to 4.6 on Heltec boards, deep sleep then kicks in at 3.3v as measured at battery terminals and shows 3.09v on meshtastic app. Obviously battery level is then wrong and never reaches 4.2v in the app.

tteskac:
[+1](https://github.com/meshtastic/firmware/issues/5259#issuecomment-3146848544) For this feature 👍

blind-oracle:
Can we reopen this?

blind-oracle:
Context: when using e.g. 18650/21700 cells to power the device, they can go down to 2.5v or even lower. With default deep sleep voltages we don't use a lot of battery capacity.

This, of course, depends on the dropout voltage of the LDO used in the given device, but still would be nice to allow to override it w/o rebuilding the source.

msmuenchen:
Please reopen, still relevant

PR #10391, opened 2026-05-05
Add SafeBoot pre-init guard for solar/low-battery nodes

Mickyleitor:
## Summary

Adds a pre-init battery guard for solar and battery nodes. Checks if we have enough juice to boot *before* lighting up the radio, scanning I²C, and writing to flash.

Currently the firmware only evaluates battery voltage after expensive init is done. On depleted solar nodes, that's too late — brownout mid-boot, restart, repeat forever, sometimes corrupting the filesystem along the way.

This guard runs very early in `setup()`, reads Vbat via quick ADC (no I²C, no PMU), and either continues boot if voltage is stable, or goes back to sleep with hysteresis and backoff until conditions improve.

## Related issues

- #9911 — "Sleeping Beauty" (stuck after brownout) — @roberthadow
- #9699 — nRF52 solar wake / LPCOMP — @JBAU92 @sandos
- #5199 — configurable low-battery thresholds
- #4555 — FS corruption from early-boot power loss
- #10084 — generic LPCOMP + user-defined thresholds — @EA2OY @sandos @roberthadow

Personal note: I built an MSP430 watchdog to force-wake nodes that got stuck after brownouts: https://github.com/Mickyleitor/meshtastic_watcher — it works, but the real fix belongs in firmware.

## What's included

- Early boot power guard for ESP32 and nRF52
- ADC-based voltage check (~80ms, no dependencies)
- Hysteresis + exponential backoff (120s → 600s cap)
- LPCOMP wake on nRF52 when available (~1-3µA)
- Small FS retry after unclean resets

## What's NOT included

- No protobuf changes
- No runtime config yet (follow-up)
- No power subsystem refactor

## What this does NOT fix (but may help indirectly)

To be clear: this guard prevents boot-loop drain when the battery is too low to complete init. It does **not** directly fix nodes that boot successfully but then get stuck in an unresponsive state at runtime.

However, it's very likely these issues are related. In my experience, nodes that end up in these "zombie states" — where they don't respond to the user button but come back to life after pressing RESET — probably got there because they booted with marginal power. The system came up, but with invalid or unhealthy RAM state from a partial/unstable init. The device is technically "running" but stuck. Pressing RESET clears everything and the node works again.

By preventing boots with insufficient power in the first place, SafeBoot should reduce the chances of ending up in that zombie state. If you never let the node attempt to boot until voltage is actually stable, you avoid the marginal-power init that leaves things broken.

## Testing

TBD.


jp-bennett:
This is really interesting. The first thing that comes to mind is whether we can reuse the existing battery curve to set the minimum voltages. (And make a custom curve for the devices with questionable LDO chips.) This way the battery percentages actually make sense with what the device is going to do. Going to sleep at 0% instead of randomly rebooting at 5% like it does now.

Mickyleitor:
> This is really interesting. The first thing that comes to mind is whether we can reuse the existing battery curve to set the minimum voltages. (And make a custom curve for the devices with questionable LDO chips.) This way the battery percentages actually make sense with what the device is going to do. Going to sleep at 0% instead of randomly rebooting at 5% like it does now.

@jp-bennett That's actually a good idea... I've just pushed that so SafeBoot now falls back to the existing battery OCV curve when a board doesn't provide explicit boot thresholds, which keeps early-boot protection aligned with the same voltage model already used for battery percentage and runtime low-battery handling.

Mickyleitor:
Unfourtunately, there is some bad news with the first initial SafeBoot test results:

Two Heltec V4 nodes powered by LiPo + solar panel were tested after a full or near-full battery depletion scenario.

One node was running stock Meshtastic firmware.
The second node was running the SafeBoot modification, which blocks peripheral initialization during boot until the battery voltage exceeds a safe threshold.

Test procedure:

* Both nodes were left without solar input until they stopped communicating with the mesh.
* Solar charging was restored afterwards.
* Batteries successfully recharged to approximately 4.2V.
* 3V3 rail measured stable at 3.3V on the affected node.
* Neither node automatically rejoined the mesh after battery recovery.

Current conclusions:

* SafeBoot alone does not appear to solve the autonomous mesh rejoin issue after battery depletion.
* The issue may not be limited to early boot brownout conditions.
* The node may instead be entering a deeper low power or sleep state from which it does not recover automatically.
* This could involve Deep Sleep, wake source configuration, PowerFSM behavior, or platform-specific low power handling.

Important:

* The SafeBoot change may still help prevent cyclic boot loops or unstable early boot behavior under low battery conditions.
* However, this has not yet been fully validated.
* Further investigation is ongoing.

Current status:
This PR should still be considered experimental and investigation-oriented rather than a confirmed fix.

Mickyleitor:
By the way, it looks like someone pulled the latest `develop` into this branch. I do not want to do that yet.

For testing, I want to keep the exact same stable baseline this change was originally based on. I do not want to validate it against the latest alpha build yet.

I will try to revert that change, or force push a rewritten history, so this branch is based again on the `2.7.15` tag.

Once the fix is confirmed and validated against that stable baseline, we can validate it again against the latest `develop`. At that point, we can rebase or merge as needed before pushing it upstream.

EA2OY:
For me, with my code its fully working on nrf52840, tried with promicro and xiao succesfully. when battery its lower from 3.4 goes to sleep left lpcomp to wake Up the node when solar MPPT charging goes Up 3.7v . So far so good!!!! i have a few nodes with this implemented sollution and one of this, left in a bad solar light zone to try this. the node goes to sleep and wake Up few times (bad weather and low capacity battery ) without problems and no brownouts, like in the lab power supply

Mickyleitor:
> For me, with my code its fully working on nrf52840, tried with promicro and xiao succesfully. when battery its lower from 3.4 goes to sleep left lpcomp to wake Up the node when solar MPPT charging goes Up 3.7v . So far so good!!!! i have a few nodes with this implemented sollution and one of this, left in a bad solar light zone to try this. the node goes to sleep and wake Up few times (bad weather and low capacity battery ) without problems and no brownouts, like in the lab power supply

Glad to hear your nRF52 solution is working solidly in the field, that's exactly the kind of validation we need, and the LPCOMP approach in #10084 looks like the right answer for that platform.

The catch is that ESP32 / ESP32-S3 doesn't have an LPCOMP equivalent, the closest would be programming the ULP coprocessor to sample the ADC during deep sleep, which is a much heavier lift and not something Meshtastic uses today. So a straight port of your approach isn't possible, and I'd rather avoid landing a fix that only works on nRF52 and silently leaves ESP32 nodes in the same broken state.

I'm looking at a more generic recovery layer (timer-based wake with hysteresis as a baseline, with LPCOMP/ULP as platform-specific optimizations on top) so SafeBoot + autonomous recovery behave consistently across platforms. Give me a few days. Thanks @EA2OY !

Mickyleitor:
Quick follow-up:

I’ve been chasing what looked like a single bug and it’s turning out to be a tangled mess of overlapping failure modes, battery discharge, slow voltage ramps, Deep Sleep entry, wake-up, and reset loops all seem to be poking at the same sore spot. Looking back at the field log, the reboot counter only moved when I physically pressed the button, the device went silent for 11 hours sitting on a fully charged 4.2 V battery, and when I finally hit RESET it came up as a cold boot, not as a wake from sleep. So whatever is happening, it doesn’t look like Deep Sleep behaving badly, it looks more like the chip never re-arming after a slow recovery ramp, which on a Heltec V4 with no external supervisor is a known soft spot.

That changes how I see this PR. The SafeBoot change in here may help with the reset-loop symptom, but I’m honestly not sure anymore that it touches the original problem I opened it for, and I can’t even claim it fully kills the reset loop yet. It’s starting to feel like a partial mitigation aimed at one corner of a bigger problem.

I only have one validation left to do before being able to say anything firmer: reproduce the same voltage steps and ramps on a programmable power supply on the bench, with probes on VBat, 3V3 and CHIP_PU. Until I run that, I’d rather not promote any of this to a conclusion. Holding the PR for now.

PR #10396, opened 2026-05-06
Add support for SY6970 fot tdisplay s3 pro

jp-bennett:
Minimal changes to make the existing LipoCharger also work with the SY6970 chip


jp-bennett:
It does come up and init. However the device powers off immediately when USB is unplugged. So something is still missing, I suspect. -- Or my battery is just dead. Based on further tests, I suspect my battery has been smoked.

fifieldt:
Sooo ... safe to merge? ;)

PR #10403, opened 2026-05-06
Refactor meshtasticd service files and update installation scripts for multiple instances

### LINKED ISSUE #9892: [Bug]: meshtasticd@.service template instances fail to start
### Hardware

Linux Native

### Firmware Version

2.7.20

### Description

The new feature introduced in https://github.com/meshtastic/firmware/issues/8676 to start multiple meshtasticd services fails.  I've poked around but am not familiar enough with systemd to submit a  fix.

`systemctl start meshtasticd@my_instance.service` fails as we have no installed `meshtasticd@.service`.  If we rename meshtasticd.service to meshtasticd@.service we need to give a name to the default instance and refer to it as such (or use a symlink).

Another potential issue I see is we're [using](https://github.com/meshtastic/firmware/blob/0ed537a336adfa4ebd2b132e407a53565e08075d/bin/meshtasticd-start.sh#L4) CONF_DIR="/etc/meshtasticd/config.d" but anything in this folder overrides /etc/meshtasticd/conf.yaml So our original instance would seem to use the additional instance config as well.  Unless I'm mistaken, which happens frequently.  An easy fix here is to set and use `CONF_DIR="/etc/meshtasticd"`

While I was poking around I noted this [log message](https://github.com/meshtastic/firmware/blob/da808cb43bd8806f2af2734bb3d35483138d36d8/bin/meshtasticd-start.sh#L8) is likely supposed to be contained within the IF block that follows.

### Relevant log output

```Shell
sudo systemctl start meshtasticd@mp.service
Failed to start meshtasticd@mp.service: Unit meshtasticd@mp.service not found.
```

jessm33:
Making using linux multiple instances easier by:

 - Including template service file `meshtasticd@.service` which allows creation of new instances out of the box
 - Updated `meshtasticd-start.sh` to use improved locations for config files and VFS, config files are out of config.d to not conflict with the default instance,  VFS folders are under `/var/lib/meshtasticd` so the script actually has permissions to automatically create them.  The script will check the old location and continue to use it for instances created before this change.  

The original `meshtasticd.service` has been changed to no longer use the startup script as it serves no purpose there.

After this change creating additional instances of meshtasticd should be as simple as creating the config file, and the system enable and start the service.

Braking Change note, config files must be moved out of `config.d` and to `/etc/meshtasticd` manually to be able to start after this change, this is a necessary step to get these config files out of config.d.

Fixes #9892

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - Linux

## Summary by CodeRabbit

* **New Features**
  * Added support for running the daemon as a systemd template instance for multi-instance setups.
  * Instance-based startups now use dedicated storage/config paths and automatically reuse existing legacy storage when present.

* **Bug Fixes**
  * Improved service startup reliability and exit handling for instance and non-instance launches.

* **Chores**
  * Updated native installs and Debian/RPM packaging to include and manage the new templated systemd unit.


jessm33:
All three valid findings fixed. Here's a summary:

[bin/meshtasticd-start.sh](vscode-webview://0f4p9dvogvsgke52ei04370rq0nu49p8c43dn92d6os5qo3k7olf/bin/meshtasticd-start.sh) — Finding 1 (line 23): Changed mkdir "${VFS_DIR}" to mkdir -p "${VFS_DIR}" || exit 1. The -p creates any missing parent directories, and || exit 1 halts the script if creation fails so meshtasticd never starts with an invalid --fsdir.

[bin/meshtasticd-start.sh](vscode-webview://0f4p9dvogvsgke52ei04370rq0nu49p8c43dn92d6os5qo3k7olf/bin/meshtasticd-start.sh) — Finding 3 (lines 33-38): Replaced the if/else around the meshtasticd invocation with a direct call + captured $?. The script now exits with meshtasticd's actual exit code instead of echo's 0, while preserving the same ok/failed logging.

[meshtasticd.spec.rpkg](vscode-webview://0f4p9dvogvsgke52ei04370rq0nu49p8c43dn92d6os5qo3k7olf/meshtasticd.spec.rpkg) — Finding 2 (lines 149-156): Made all three systemd scriptlets template-aware:

%post: adds %systemd_post meshtasticd@.service to register the template on install
%preun: on full uninstall ($1 -eq 0), disables and stops all meshtasticd@*.service instances
%postun: on upgrade ($1 -ge 1), restarts all running instances via try-restart

PR #10404, opened 2026-05-06
Enhance GPS search failure handling backoff logic

thebentern:


b8b8:
Can this be updated so the double button click (of screenless devices) not only forces the standard position ping but ALSO forces a new lock if within the new cool down period. Basically you know youve been spelunking all day and as soon as you exit the cave, you want to manually force a new position lock. :)

PR #10401, opened 2026-05-06
LR2021 radio on NRF_Promicro

NomDeTom:
This pull request adds support for the LR2021 radio chipset, introducing a new `LR2021Interface` and a generic `LR20x0Interface` base class, based on the LR11x0 and LR1110.

**LR2021 radio support:**

* Added `LR2021_RADIO` to the `LoRaRadioType` enum and updated radio detection logic to instantiate and initialize `LR2021Interface` when appropriate. [[1]](diffhunk://#diff-d1ba735d815f635321eba3c5db8a06f8b6d0895ddb3937121a83705b147d4f42L14-R15) [[2]](diffhunk://#diff-ad55475cca6c487a8b020a5f8036f160cd1fc92cce2ea887bfe83f284a6fc855R482-R496)
* Added and implemented `LR2021Interface` and generic `LR20x0Interface` classes for LR2021 radio support, including initialization, configuration, and runtime methods. [[1]](diffhunk://#diff-91d47153dacad14a3f94052b064120dbac7a816068a1096e7f0836a99699ce42R1-R17) [[2]](diffhunk://#diff-ec6a64f75dc17b35988c85ddd1df8c45b718d1fdc49ab2769119c93c0a82515dR1-R15) [[3]](diffhunk://#diff-e277e25860d272402102115b1dd5c82ea69df1d2f7b5394109b4177e2be460bfR1-R75) [[4]](diffhunk://#diff-f10fb3492f1ccf65f44c09d8723b0d6bb691186864a20df9ce0b4275e87c83d5R1-R381)
* Integrated LR2021 interface into build system and template instantiations, including necessary includes in source files. [[1]](diffhunk://#diff-93922d13b20758873efd32f4ebb5ba7c63cbc5a058460911af3f55ecc88d59d1R3-R4) [[2]](diffhunk://#diff-93922d13b20758873efd32f4ebb5ba7c63cbc5a058460911af3f55ecc88d59d1R26-R28) [[3]](diffhunk://#diff-ad55475cca6c487a8b020a5f8036f160cd1fc92cce2ea887bfe83f284a6fc855R8)

**Radio initialization and configuration improvements:**

[...]

PR #10424, opened 2026-05-08
fix(mesh): radio + router reliability hardening

DatanoiseTV:
## Summary

A bundle of small, independently-revertable fixes in the radio and routing layers. Each commit addresses one specific class of bug uncovered while reading the recent reconfigure-return-type churn (#10407 / #10411) and tracing the surrounding error paths.

## Fixes (one logical commit per item)

- **`radio: return real status from reconfigure() and stop panicking on SPI errors`** — extends #10407. SX126x/SX128x/LR11x0/LR20x0/RF95 all returned `true` unconditionally even when individual SPI setters failed, so the reboot-on-bad-reconfigure path in `RadioInterface.cpp:515` never tripped. Track failures in a local `bool ok` and `return ok`. Same drivers also `assert(err == RADIOLIB_ERR_NONE)` on standby/setSyncWord/setCurrentLimit/setPreambleLength/setOutputPower — converted to log-and-fail so a momentary SPI glitch doesn't panic-reset the firmware. LR20x0 (added in #10401) had the same bug as the others.
- **`mesh: null-check getMeshNode in handleFromRadio`** — `MeshService.cpp:97` dereferenced `nodeDB->getMeshNode(mp->from)->has_user` without a null check. `getMeshNode` legitimately returns nullptr under heap pressure (see `NodeDB::isFull` and `updateFrom` early-out), so every received decoded packet from a new node could crash on a full device. Restructure with an init-statement so the predicate short-circuits.
- **`radio: drop oversized RX packets instead of asserting`** — `handleReceiveInterrupt` bounded the encrypted-payload memcpy with `assert((uint32_t)payloadLen <= sizeof(mp->encrypted.bytes))`. `assert` vanishes in NDEBUG, and even when on, panicking on a malformed (or simply oversized future-modem) packet is the wrong response. Replace with a runtime length check that releases the pool entry, bumps `rxBad`, and returns.
- **`mesh: add exponential backoff and jitter to retransmits`** — `NextHopRouter::setNextTx` scheduled every retry at the same offset returned by `getRetransmissionMsec` (driven by channel utilization, not attempt history). Two nodes that simultaneously NAK each other re-collide on near-identical schedules. Introduce 1×/2×/4× backoff capped at 8× plus random half-base jitter.
- **`router: skip pre-encryption packet copy when MQTT does not need it`** — extends the spirit of #10411. `Router::send` unconditionally `allocCopy`/`release`'d a clone of every relayed packet to feed `MQTT::onSend`, even when MQTT is excluded at compile time, runtime-disabled, or the packet wasn't from us. Hoist the alloc behind the same predicate the call site already used. Halves the pool-alloc rate on the hot path for non-MQTT devices.
- **`router: don't evict ACKs when a background-priority packet arrives full`** — `fromRadioQueue` is a plain FIFO of size 4. When full, the prior behaviour dropped the oldest entry — usually a routing ACK or reply — to make room for whatever just arrived, even if it was BACKGROUND priority. Reverse that for the low-priority case: BACKGROUND or below is dropped at the door; higher priorities still evict. `PointerQueue` doesn't expose iteration so a full priority refactor is bigger than needed here.
- **`radio: poll TX_DONE and stuck-TX timeout from main loop`** — the 60 s stuck-TX guard in `canSendImmediately` only fires when more packets enter the queue, because that's its only call site. If the queue empties after the radio wedges, the device sits in `busyTx` forever. `pollMissedIrqs` is already invoked unconditionally from the main loop — extend it to poll `TX_DONE` when `sendingPacket != NULL` and trip the same reboot path independently of queue activity.
- **`mesh: bound the user-facing notification sprintf calls`** — two `sprintf(cn->message, …)` sites in `Router.cpp` and `NodeDB.cpp` have no bound check. Today's format strings fit, but a later edit without re-checking the buffer would silently overrun. Switch to `snprintf` with `sizeof`.

## Build verification

- `pio run -e t-deck-tft` — SUCCESS, 38.4% RAM / 55.3% flash. No new warnings introduced (LSP diagnostics seen during dev are local PlatformIO-include-path misconfig, not real).
- Did **not** run on hardware — please request community testing before merge. The reconfigure() and TX_DONE-poll changes especially deserve a packet-capture eye.

## Attestations

- [x] I have tested that my proposed changes behave as described — review-level / static-analysis only; not on-air.
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Other — build-verified `t-deck-tft` only

## Notes for reviewers

- Each commit is independent; happy to split into 8 PRs if you'd prefer atomic review. Kept together because they share the radio/router context and were uncovered as a set.
- Edits-by-maintainers should be enabled.

### DIFF
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -173,21 +173,33 @@ template <typename T> bool LR11x0Interface<T>::reconfigure()
     // set mode to standby
     setStandby();
 
-    // configure publicly accessible settings
+    // The caller reboots the device when reconfigure returns false; track
+    // every SPI setter so we don't silently report success after a glitch.
+    bool ok = true;
+
     int err = lora.setSpreadingFactor(sf);
-    if (err != RADIOLIB_ERR_NONE)
+    if (err != RADIOLIB_ERR_NONE) {
         RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
+        ok = false;
+    }
 
     err = lora.setBandwidth(bw, wideLora() && (getFreq() > 1000.0f));
-    if (err != RADIOLIB_ERR_NONE)
+    if (err != RADIOLIB_ERR_NONE) {
         RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
+        ok = false;
+    }
 
     err = lora.setCodingRate(cr, cr != 7); // use long interleaving except if CR is 4/7 which doesn't support it
-    if (err != RADIOLIB_ERR_NONE)
+    if (err != RADIOLIB_ERR_NONE) {
         RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
+        ok = false;
+    }
 
     err = lora.setSyncWord(syncWord);
-    assert(err == RADIOLIB_ERR_NONE);
+    if (err != RADIOLIB_ERR_NONE) {
+        LOG_ERROR("LR11x0 setSyncWord %s%d", radioLibErr, err);
+        ok = false;
+    }
 
     if (config.lora.region == meshtastic_Config_LoRaConfig_RegionCode_LORA_24) { // clamp if wide freq range
         limitPower(LR1120_MAX_POWER);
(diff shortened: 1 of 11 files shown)


thebentern:
Hi. Please don't include so many unrelated changes in one PR

NomDeTom:
I would check the jitter on packets - I believe there is already some random delay on rebroadcast.

Other question: how does a node know the reason for failure of its own packet?

DatanoiseTV:
Thanks @thebentern — fair point. Re-opened as 8 atomic PRs, one per fix:

- #10430 — radio: return real status from reconfigure() and stop panicking on SPI errors (extends #10407)
- #10431 — mesh: null-check getMeshNode in handleFromRadio
- #10432 — radio: drop oversized RX packets instead of asserting
- #10433 — mesh: add exponential backoff and jitter to retransmits
- #10434 — router: skip pre-encryption packet copy when MQTT does not need it (extends spirit of #10411)
- #10435 — router: don't evict ACKs when a background-priority packet arrives full
- #10436 — radio: poll TX_DONE and stuck-TX timeout from main loop
- #10437 — mesh: bound the user-facing notification sprintf calls

Each is independent and revertable; happy to revise / drop any of them on their own thread.

Issue #167, opened 2020-06-09
wait to sleep bug


geeksville:
```
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
NOTE! Recording critical error 2, address=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
radio wait to sleep, txEmpty=0
```

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/ttgo-lora32-v2-0-appears-to-sleep-forever/339/6

geeksville:
Oh!  Root cause is that the wait to sleep loop is consuming the main thread - and therefore starving the handler that would have sent the message (see the last "delaying " line in this smoking gun trace):

```
Telling client we have new packets 3
Telling client we have new packets 3
Starting low level send from=0xabdddf38, to=0xffffffff, id=250959435, want_ack=0
Completed sending to=0xffffffff, id=250959435
tx complete - starting timer
processUBX: counter hit MAX_PAYLOAD_SIZE
new gps pos lat=0.000000, lon=0.000000, alt=0
processUBX: counter hit MAX_PAYLOAD_SIZE
Ignoring invalid GPS time
processUBX: counter hit MAX_PAYLOAD_SIZE
Ignoring invalid GPS time
processUBX: counter hit MAX_PAYLOAD_SIZE
[...]

PR #10425, opened 2026-05-08
fix(power): sleep/battery reliability hardening

DatanoiseTV:
## Summary

A bundle of small power-management fixes uncovered while auditing the sleep state machine and battery-presence logic. All cross-cutting (no board-specific pin changes); each commit independently revertable.

## Fixes

- **`sleep: arm deep-sleep hold so per-pin gpio_hold_en actually holds`** — `doDeepSleep` calls `gpio_hold_en` on `BUTTON_PIN` and `LORA_CS`, but the ESP-IDF API contract is that `gpio_hold_en` alone only retains pin state during *light* sleep. Deep sleep additionally needs a one-time `gpio_deep_sleep_hold_en()` before `esp_deep_sleep_start` to arm the holds across the boundary. Without it the pins float in DSLP, which on some boards leaks current through `LORA_CS` and on others triggers spurious wake from a noisy button line.
- **`gps: soft-sleep the receiver during CPU light-sleep`** — `GPS` only subscribed to `notifyDeepSleep`. During light sleep (the default non-router path) the receiver kept drawing ~25 mA the entire time the MCU was asleep — the dominant drain on devices that spend most of their time in LIGHT_SLEEP. Subscribe to `notifyLightSleep` / `notifyLightSleepEnd` and use `GPS_SOFTSLEEP` instead of full `disable()` so the state machine survives across the short window without paying re-init / re-acquire on every cycle.
- **`power: latch low-battery decision in RTC memory and add recovery hysteresis`** — the previous handler counted 11 consecutive sub-OCV[min] readings and triggered `EVENT_LOW_BATTERY`, but on SDS wake `low_voltage_counter` reset to 0. With the cell still flat, the count climbed back to 11 and re-deep-slept — wake-thrash burning boot energy each cycle. There was also no recovery hysteresis: a brief spike across threshold cleared the count even with the cell effectively flat. Persist the latch in `RTC_DATA_ATTR` (plain `static` on non-ESP32 — same effective semantics since SDS doesn't reset there). Once latched, clearing requires K consecutive readings above OCV[min] + 100 mV. If the device wakes still latched-flat, it immediately re-enters SDS.
- **`sleep: replace assert(0) on stuck preflight with controlled restart`** — `waitEnterSleep` panicked with `assert(0)` when an observer vetoed sleep for >30 s (FIXME #167). On a field device that turns into a noisy crash log every 30 s rather than a clean reboot. Use the same per-arch restart primitives `Power::reboot` already uses (`ESP.restart` / `NVIC_SystemReset` / `rp2040.reboot` / `HAL_NVIC_SystemReset`).
- **`power: use BQ25896 ADC for both battery voltage and presence`** — on the BQ27220+BQ25896 combo, `getBattVoltage` returned the fuel-gauge reading while `isBatteryConnect` derived presence from the BQ25896 charger ADC. The two could disagree on a freshly-attached cell before the gauge had learned: hasBattery would say false but voltage 4.0 V, and the upstream PowerFSM would treat the unit as externally powered. Read voltage from the same source as presence so both views always agree.

## Build verification

- `pio run -e t-deck-tft` — SUCCESS, 38.4% RAM / 55.3% flash. No new warnings.
- Not yet run on hardware — please request community testing for the GPS soft-sleep behaviour and the low-battery latch in particular.

## Deferred (scope kept tight)

- T-Deck Pro v1.1 board-specific power gating (KB_BL pin 42, GPS_EN pin 39, LORA_EN pin 46, DRV2605 pin 2, BMS IRQ wiring) — separate follow-up branch when desired.
- I²C-keyboard polling-rate reduction in DARK/LS — needs a small refactor to expose PowerFSM state to `kbI2cBase`, didn't want to land a half-fix here.
- `setCPUFast` carve-outs on T-Deck — risks TFT/WiFi instability without bench testing.

## Attestations

- [x] I have tested that my proposed changes behave as described — review/static-analysis only.
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Other — build-verified `t-deck-tft` only

## Notes for reviewers

- Each commit is independent; happy to split into 5 PRs if preferred.
- The low-battery latch is the highest-impact fix in this set.
- Edits-by-maintainers should be enabled.

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -848,6 +848,17 @@ void Power::shutdown()
 #endif
 }
 
+// Persisted across SDS / deep-sleep on ESP32 so a flat-cell device that has
+// already chosen to deep-sleep doesn't wake on the timer, see still-low
+// voltage, and burn boot energy on a doomed full init only to deep-sleep
+// again. On non-ESP32 the SDS path doesn't reset this anyway, so a plain
+// static gives the same effective semantics.
+#ifdef ARCH_ESP32
+RTC_DATA_ATTR static bool low_battery_latched = false;
+#else
+static bool low_battery_latched = false;
+#endif
+
 /// Reads power status to powerStatus singleton.
 //
 // TODO(girts): move this and other axp stuff to power.h/power.cpp.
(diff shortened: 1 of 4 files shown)


DatanoiseTV:
Thanks for the review — all three points addressed in `ab77d4dba`:

1. **NUM_CELLS scaling.** Good catch. `getBattVoltage()` returns pack voltage and `OCV[]` is per-cell; the comparison was wrong on multi-cell variants (chatter2, station-g1). Scaled both the low-battery latch threshold and the recovery hysteresis offset by `NUM_CELLS` so behaviour matches the existing percentage-calc convention at `Power.cpp:890`.
2. **`getBattVoltage` docstring.** Fixed — returns `uint16_t` with `0` as the unknown sentinel, not `NAN`.
3. **`notifyReboot` from `waitEnterSleep`.** Added the `notifyReboot.notifyObservers(NULL)` call before the platform reset primitives, matching `Power::reboot`'s contract so observers like InkHUD get the last-second cleanup pass.

Build still clean on `pio run -e t-deck-tft`. No new warnings.

DatanoiseTV:
Splitting this proactively per the same atomic-PR feedback @thebentern gave on #10424. Re-opened as 5 separate PRs, one fix per PR (all incorporate the Copilot review feedback from this thread):

- #10438 — power: arm deep-sleep hold so per-pin gpio_hold_en actually holds
- #10439 — power: soft-sleep the GPS receiver during CPU light-sleep
- #10440 — power: latch low-battery decision in RTC memory and add recovery hysteresis (incl. NUM_CELLS scaling)
- #10441 — power: replace assert(0) on stuck preflight with controlled restart (incl. notifyReboot)
- #10442 — power: use BQ25896 ADC for both battery voltage and presence (incl. doc fix)

Each is independent and revertable.

PR #10430, opened 2026-05-09
fix(mesh): return real status from reconfigure() and stop panicking on SPI errors

DatanoiseTV:
reconfigure() in all five chip drivers (SX126x, SX128x, LR11x0, LR20x0,
RF95) was returning true unconditionally, even when individual SPI
setters failed. The caller in RadioInterface::cppInit() reboots the
device when reconfigure returns false, so the failure-recovery path was
silently disabled. Track failures in a local bool and return its
accumulated state.

Same drivers also asserted on remote SPI calls in reconfigure() and
setStandby() — a momentary glitch on the SPI bus would panic-reset the
firmware. Replaced with log-and-continue; the new return value lets the
caller make the reboot decision instead of crashing inline.

## Build verification

## Attestations

### DIFF
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -173,21 +173,33 @@ template <typename T> bool LR11x0Interface<T>::reconfigure()
     // set mode to standby
     setStandby();
 
-    // configure publicly accessible settings
+    // The caller reboots the device when reconfigure returns false; track
+    // every SPI setter so we don't silently report success after a glitch.
+    bool ok = true;
+
     int err = lora.setSpreadingFactor(sf);
-    if (err != RADIOLIB_ERR_NONE)
+    if (err != RADIOLIB_ERR_NONE) {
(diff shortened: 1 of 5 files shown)


caveman99:
@DatanoiseTV could you address the copilot findings here?

cvaldess:
Tested on Nordic nRF54L15-DK with EBYTE E22-900M30S (SX1262) on EU_868,
stacked on top of #10227 / #10432 / #10217 / #10441 / #10271 / #10252 in
the same experiment branch.

- Builds clean across all SX/LR/RF95 interfaces touched (5 files, no
  warnings).
- SX1262 init came back result 0 — the new error path didn't fire, but
  the previous behaviour was the bigger concern on this port: with
  `CONFIG_RESET_ON_FATAL_ERROR=n` our fault handler does a
  `sys_reboot(SYS_REBOOT_COLD)`, so a transient SPI hiccup during
  bring-up used to silently reboot the device instead of surfacing a
  status code we could log.
- ~5 min of normal LoRa RX/TX traffic with no regressions; rxGood counts
  correctly, no spurious resets.

LGTM.

Tested-by: cvaldess

PR #10432, opened 2026-05-09
fix(mesh): drop oversized RX packets instead of asserting

DatanoiseTV:
handleReceiveInterrupt bounded the encrypted-payload memcpy with an
assert, which a future modem with a longer max payload would turn into
a panic-reset on a malformed (or simply larger) air packet. Replace
with a runtime length check that releases the pool entry, bumps rxBad,
and returns. assert() also vanishes in NDEBUG builds, so there was no
actual bound check in release.

## Build verification

## Attestations

### DIFF
--- a/src/mesh/RadioLibInterface.cpp
+++ b/src/mesh/RadioLibInterface.cpp
@@ -530,7 +530,16 @@ void RadioLibInterface::handleReceiveInterrupt()
 
             mp->which_payload_variant =
                 meshtastic_MeshPacket_encrypted_tag; // Mark that the payload is still encrypted at this point
-            assert(((uint32_t)payloadLen) <= sizeof(mp->encrypted.bytes));
+            if ((uint32_t)payloadLen > sizeof(mp->encrypted.bytes)) {
+                LOG_WARN("Drop oversized RX packet (%d > %u)", payloadLen, (unsigned)sizeof(mp->encrypted.bytes));
+                packetPool.release(mp);
+                // The outer else branch already incremented rxGood for this packet;
+                // undo that and reclassify as rxBad so num_packets_rx isn't double-counted.
+                rxGood--;
+                rxBad++;
+                airTime->logAirtime(RX_ALL_LOG, rxMsec);
+                return;
+            }
             memcpy(mp->encrypted.bytes, radioBuffer.payload, payloadLen);
             mp->encrypted.size = payloadLen;
 


DatanoiseTV:
Good catch — fixed in 277817b8d. The outer else branch increments `rxGood` before any of the inner validity checks, so the cleanest minimal fix is to decrement it back when reclassifying the packet as `rxBad` on this path. Matches the existing pattern that `from == 0` (which also early-returns from this same outer-else block) leaves untouched, while keeping our new oversized branch counted exactly once.

src/mesh/RadioLibInterface.cpp
```diff
@@ -530,7 +530,16 @@ void RadioLibInterface::handleReceiveInterrupt()
 
             mp->which_payload_variant =
                 meshtastic_MeshPacket_encrypted_tag; // Mark that the payload is still encrypted at this point
-            assert(((uint32_t)payloadLen) <= sizeof(mp->encrypted.bytes));
+            if ((uint32_t)payloadLen > sizeof(mp->encrypted.bytes)) {
+                LOG_WARN("Drop oversized RX packet (%d > %u)", payloadLen, (unsigned)sizeof(mp->encrypted.bytes));
+                packetPool.release(mp);
+                // The outer else branch already incremented rxGood for this packet;
+                // undo that and reclassify as rxBad so num_packets_rx isn't double-counted.
+                rxGood--;
```
GUVWAF, line 538: These statistics are really about whether it was correctly decoded when receiving over the air, which is the case (the CRC must have passed if you reach here), hence "rxGood" is still correct.

caveman99: i'm with @GUVWAF here. don't touch the RX statistics at all, just make it fail gracefully instead of asserting. Also aomething to note: this was in place originally as a 'safeguard' against people stuffing too much into a meshpacket, so more of a sanity check. Arguably a happy path decision.

caveman99 (changes requested):
Please have a look at the failing tests. Can Probably ignore the 502 but the rest is of importance.

cvaldess:
Tested this on a Nordic nRF54L15-DK with EBYTE E22-900M30S (SX1262)
on an EU_868 mesh.

- Builds clean (RAM 50.9%, Flash 32.4% — no footprint change).
- Boots OK, LoRa RX flowing from neighbour nodes (rxSNR 11–13).
- The defensive branch didn't fire in ~5 min of normal traffic (no oversized
  packets to provoke it), but the assert() removal alone is meaningful on
  this port: with `CONFIG_RESET_ON_FATAL_ERROR=n` our crash handler does a
  `sys_reboot(SYS_REBOOT_COLD)`, so the previous bound check would have
  rebooted the device on the first malformed/oversized frame in release
  builds (where `assert()` already vanishes).

LGTM as a port-side reviewer.

Tested-by: cvaldess

PR #10433, opened 2026-05-09
fix(mesh): add exponential backoff and jitter to retransmits

DatanoiseTV:
setNextTx scheduled every retry at the same base offset returned by
getRetransmissionMsec, which is driven by channel utilization rather
than per-attempt history. Two nodes that simultaneously NAK each other
re-collide on near-identical schedules and waste airtime. Use a 1x/2x/4x
backoff (with the multiplier capped at 8x for safety) plus random
half-base jitter so concurrent attempts spread out.

## Build verification

## Attestations

### DIFF
--- a/src/mesh/NextHopRouter.cpp
+++ b/src/mesh/NextHopRouter.cpp
@@ -349,9 +349,16 @@ int32_t NextHopRouter::doRetransmissions()
 void NextHopRouter::setNextTx(PendingPacket *pending)
 {
     assert(iface);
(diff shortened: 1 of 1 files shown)


GUVWAF:
I don't see why we need yet another random backoff. The schedule of the retransmission is just for when it is placed in the Tx queue. Before handling the packet in front of the Tx queue, we already apply random backoff. I don't see what this additional random offset based on the number of attempts done adds to it.

PR #10435, opened 2026-05-09
fix(mesh): don't evict ACKs when a background-priority packet arrives full

DatanoiseTV:
fromRadioQueue is a plain FIFO of size 4. When it fills, the prior
behaviour was to drop the oldest packet — usually a routing ACK or
reply — to make room for whatever just arrived, even if the new packet
was BACKGROUND priority. Reverse that for the low-priority case: if
the incoming packet's priority is at or below BACKGROUND, drop it
instead. Higher-priority arrivals still evict to fit. PointerQueue
doesn't expose iteration, so a full priority-queue refactor would be a
larger change; this addresses the worst case.

## Build verification

## Attestations

### DIFF
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -168,13 +168,22 @@ int32_t Router::runOnce()
  */
 void Router::enqueueReceivedMessage(meshtastic_MeshPacket *p)
 {
-    // Try enqueue until successful
-    while (!fromRadioQueue.enqueue(p, 0)) {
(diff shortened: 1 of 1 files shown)


GUVWAF:
Are you observing that the FIFO ever gets full? 
This should really not happen: it means we're too slow handling incoming packets. If we're really seeing this (often), IMO we should increase the FIFO size rather than throwing away the "least important one".

PR #10439, opened 2026-05-09
fix(power): soft-sleep the GPS receiver during CPU light-sleep

DatanoiseTV:
GPS only subscribed to notifyDeepSleep, so during light-sleep the
receiver kept drawing ~25 mA on a non-router device — the dominant
drain on devices that spend most of their time in LIGHT_SLEEP.
Subscribe to notifyLightSleep/notifyLightSleepEnd as well, but use
GPS_SOFTSLEEP instead of disable()+enable() since the state machine
survives across the short light-sleep window and full power-cycle
re-acquisition would cost more than it saves.

## Build verification

## Attestations

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -819,6 +819,10 @@ bool GPS::setup()
     }
 
     notifyDeepSleepObserver.observe(&notifyDeepSleep);
+#ifdef ARCH_ESP32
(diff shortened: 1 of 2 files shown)


NomDeTom:
This needs to be reviewed against the main GPS power system. Tying it to CPU sleeping doesn't necessarily make sense.

PR #10441, opened 2026-05-09
fix(power): replace assert(0) on stuck preflight with controlled restart

DatanoiseTV:
waitEnterSleep panicked with assert(0) when an observer vetoed sleep
for >30s (FIXME #167). On a field device that turns into a noisy
crash log every 30s rather than a clean reboot. Use the same
per-arch restart primitives Power::reboot uses, fall back to assert
only for archs we don't have a primitive for.

## Build verification

## Attestations

### DIFF
--- a/src/sleep.cpp
+++ b/src/sleep.cpp
@@ -195,7 +195,26 @@ static void waitEnterSleep(bool skipPreflight = false)
             if (!Throttle::isWithinTimespanMs(now,
                                               THIRTY_SECONDS_MS)) { // If we wait too long just report an error and go to sleep
                 RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_SLEEP_ENTER_WAIT);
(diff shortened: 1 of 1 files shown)


cvaldess:
Tested on Nordic nRF54L15-DK (stacked with the other Tier-1 hardening
PRs on a single experiment branch).

- Builds clean.
- No stuck preflight observed during normal bring-up, so the new restart
  path wasn't exercised in this smoke test.
- Same reasoning as on #10430 / #10432 applies on this port: replacing
  `assert(0)` is a real upgrade for us because `CONFIG_RESET_ON_FATAL_ERROR=n`
  routes fatal halts through our crash handler's `sys_reboot(SYS_REBOOT_COLD)`,
  which is noisier and harder to diagnose than the controlled restart this
  PR introduces.

LGTM.

Tested-by: cvaldess

PR #10442, opened 2026-05-09
fix(power): use BQ25896 ADC for both battery voltage and presence

DatanoiseTV:
getBattVoltage on the BQ27220+BQ25896 combo returned the fuel-gauge
reading while isBatteryConnect derived presence from the BQ25896
charger ADC. The two could disagree on a freshly-attached cell before
the fuel gauge had learned: hasBattery would say false but voltage
4.0V, and the upstream PowerFSM would treat the unit as externally
powered. Read voltage from the same source as presence so both views
move together.

## Build verification

## Attestations

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -1668,9 +1668,14 @@ class LipoCharger : public HasBatteryLevel
     }
 
     /**
-     * The raw voltage of the battery in millivolts, or NAN if unknown
+     * The raw voltage of the battery in millivolts, or 0 if unknown.
+     * Use the BQ25896 charger ADC for both voltage and presence detection so
+     * the two readings can never disagree. The BQ27220 fuel gauge can report
+     * 0 / stale on a freshly-attached cell before its initial learn cycle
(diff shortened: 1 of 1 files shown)


mverch67:
@DatanoiseTV you wrote:
>
image
</p>

The T-Deck **does not have** a BQ27220/BQ25896 charger/gauge, it uses GPIO ADC battery pin 4.

Please use T-Lora Pager or T-Deck Pro for tests with BQ27220/BQ25896.

Also, the BQ27220 has a more accurate voltage reporting as it implements the calibrated charging curve and automatically learns the charge capacity from previous full battery cycles.

NomDeTom:
Be very careful - the bq25896 is currently linked to another bq module for "fuel gauging" and also shares an i2c address with some other common sensors.

mverch67:
yeah, T-Deck I2C keyboard and BQ27220 also have the **same I2C address 0x55**. Maybe that's where the confusion came from.

PR #10459, opened 2026-05-12
Remove gradient sync nonce and simplify replay handling

thebentern:
I'm willing to admit, the gradient sync was a good idea, but it was a mistake to add new nonces. We should just always split the telem and position sync post complete node sync. Tested and it's working great on iOS. I'll do some Android testing too.


thebentern:
@copilot apply changes based on the comments in [this thread](https://github.com/meshtastic/firmware/pull/10459#pullrequestreview-4268270689)

thebentern:
@copilot apply changes based on the comments in [this thread](https://github.com/meshtastic/firmware/pull/10459#pullrequestreview-4268270689)

PR #10478, opened 2026-05-14
Packet Signing via XEdDSA

jp-bennett:
New PR after a particularly ugly merge.


Xaositek:
Should https://github.com/meshtastic/firmware/pull/7602 be closed in favor of this?

thebentern:

src/mesh/CryptoEngine.cpp
```diff
@@ -65,13 +71,102 @@ bool CryptoEngine::regeneratePublicKey(uint8_t *pubKey, uint8_t *privKey)
...
     } else {
         LOG_WARN("X25519 key generation failed due to blank private key");
         return false;
     }
     return true;
 }
 
+/**
+ * Build a signing buffer that covers packet metadata and payload:
+ *   [fromNode(4) | packetId(4) | portnum(4) | payload(N)]
```
adam-scott-thomas, line 92: This is a really solid signing boundary. Covering fromNode, packetId, portnum, and payload means this catches more than just payload tampering: replay under a new packet id, sender reattribution, and port redirection all get folded into the same check. Nice call.

src/mesh/Router.cpp
```diff
@@ -530,6 +530,36 @@ DecodeState perhapsDecode(meshtastic_MeshPacket *p)
...
+                if (p->xeddsa_signed) {
+                    // Mark this node as a signer so future unsigned packets from it are rejected
+                    nodeInfoLiteSetBit(node, NODEINFO_BITFIELD_HAS_XEDDSA_SIGNED_MASK, true);
+                    LOG_DEBUG("Verified XEdDSA signature from 0x%08x", p->from);
+                } else {
+                    LOG_WARN("XEdDSA signature verification failed from 0x%08x, dropping", p->from);
+                    return DecodeState::DECODE_FAILURE;
+                }
+            } else {
+                LOG_DEBUG("No public key for 0x%08x, cannot verify XEdDSA signature", p->from);
```
adam-scott-thomas, line 578: One provenance-semantics thought: a packet with a signature present but no available public key is different from both “verified signed” and “unsigned.” Even if accepting it is the right compatibility choice, it may be worth surfacing that state distinctly for apps/admin tooling: verified signed, unsigned, and signed-but-unverifiable.

src/mesh/CryptoEngine.cpp
```diff
@@ -65,13 +71,102 @@ bool CryptoEngine::regeneratePublicKey(uint8_t *pubKey, uint8_t *privKey)
...
+ * This prevents replay, reattribution, and portnum redirection attacks.
+ */
+static size_t buildSigningBuffer(uint8_t *buf, size_t bufSize, uint32_t fromNode, uint32_t packetId, uint32_t portnum,
+                                 const uint8_t *payload, size_t payloadLen)
+{
+    const size_t headerLen = sizeof(uint32_t) * 3;
+    size_t totalLen = headerLen + payloadLen;
+    if (totalLen > bufSize)
+        return 0;
+    memcpy(buf, &fromNode, sizeof(uint32_t));
```
adam-scott-thomas, line 103: Small deterministic-preimage concern: because this buffer is now the exact bytes being signed, it effectively becomes wire-contract material. Are we guaranteed consistent endian/layout across supported targets here? If not, it may be safer to encode fromNode, packetId, and portnum explicitly in a fixed byte order before signing.

src/mesh/Router.cpp
```diff
@@ -602,6 +632,16 @@ meshtastic_Routing_Error perhapsEncode(meshtastic_MeshPacket *p)
             p->decoded.has_bitfield = true;
             p->decoded.bitfield |= (config.lora.config_ok_to_mqtt << BITFIELD_OK_TO_MQTT_SHIFT);
             p->decoded.bitfield |= (p->decoded.want_response << BITFIELD_WANT_RESPONSE_SHIFT);
+            // Sign broadcast packets if payload + signature fits within the max Data payload.
+            // The actual encoded size is checked after pb_encode (TOO_LARGE).
+            if (!p->pki_encrypted && isBroadcast(p->to) &&
+                p->decoded.payload.size + XEDDSA_SIGNATURE_SIZE < meshtastic_Constants_DATA_PAYLOAD_LEN) {
```
adam-scott-thomas, line 668: This size gate makes sense given the MTU. Longer term, this feels like the natural seam for an opt-in custody/continuity layer rather than more per-packet crypto: a small custody flag plus an 8–16 byte rolling hash/tag on custody-enabled messages, with periodic signed Merkle/batch anchors instead of attaching heavy proof material to every packet.

The Merkle side can work fully offline too: roots can be signed by a local field box with monotonic counters/GPS time, then externally anchored/timestamped later when an internet-connected path is available.

adam-scott-thomas (commented):
Really solid direction overall. The signing boundary looks right to me, especially binding the signature to sender/packet/app context instead of just payload. My comments are mostly around making the signature preimage deterministic across ports, preserving provenance semantics for signed-but-unverifiable packets, and where a future opt-in custody/continuity layer could fit without fighting the MTU constraints.

src/mesh/CryptoEngine.cpp
jp-bennett: Added a comment calling this out.

thebentern (approved):

Issue #7553, opened 2025-08-05
[Feature Request]: RGB LED for node status


CypressXt:
### Platform

Cross-Platform

### Description

Hello there 👋, 

For devices that support RGB, would you be interested in being able to use the LED as a device's status indication?
As a first implementation I thought about those status: 
* Charging (breathing blue LED)
* Low battery (blinking red LED)
* Lost GPS lock (Blinking orange LED)
* Default color set by the ambient lighting module
* Temporary override the LED status by the external notification module when needed

This would be in addition of the current ambient lighting and external notification module features and you should be able to use the LED status in addition of the 2 already existing modules.

Let me know what you think, if that make sense to everyone I can gladly prepare a PR with it 🚀

jp-bennett:
I have a NeoPixel hanging off my Raspberry Pi, and I've been thinking about something very similar. Watching this.

Mhowser:
I would also like to see different colors for radio TX / RX activity as well.

markbirss:
Yes, i would like to see RGB LED support

radiolee:
Yes, I was run ws2812 on an ESP32C3 mini with success, but this rgb is usless this time.

<img width="950" height="654" alt="Image" src="https://github.com/user-attachments/assets/4cfb1bf8-62b3-4d46-9c64-fda7e2e51e06" />

melkati:
Hello, is there any progress on this? 
I would like to use the onboard LED of the ESP32 S3 Supermini.

CypressXt:
> Hello, is there any progress on this? I would like to use the onboard LED of the ESP32 S3 Supermini.

Hi Hi @melkati, on my side not yet as it's not as straight forward as I expected due to the fact that the LED is controlled by different modules at the same time (power, external notifications, etc.). I didn't give up on anything, but I haven't had time to address it yet.

shafr:
Isn't this (at least partially) implemented by https://github.com/meshtastic/firmware/pull/3647/changes ?

CypressXt:
I'll jump back at it starting next week 👌

jp-bennett:
@CypressXt also take a look at what's new in StatusLEDModule.cpp.

Something I've just barely started working on is a behavior document to outline the standard Meshtastic LED/button behaviors. 

For instance, we have a primary (usually red) heartbeat. The blips last longer when plugged in and charging, and a recent addition is that the red LED goes solid when fully charged. (Possibly some bugs in that behavior based on reports.)

[...]

PR #10489, opened 2026-05-17
Splitted RGB / NON-RGB in Device status LED module

CypressXt:
Follow up on #7553

Added support for RGB LP5562 and NCP5623 devices


CypressXt:
Hiiii  :wave:, sorry for the delay. This is still relevant and I'll update it swiftly

PR #10528, opened 2026-05-21
Update ESP32-CH390 to v1.1.0

vidplace7:
Library updated for idf5 / arduino 3.x compat (should retain backwards compat with idf4 / arduino 2.x)
https://github.com/meshtastic/ESP32-CH390/pull/2


vidplace7:
Still compiles against master 👍

thebentern (approved):

PR #10531, opened 2026-05-21
Add a way to have MQTT uplinks uplink multiple copies of packets

ianmcorvidae:
... when they're heard multiple times (hidden behind compile flag + userPrefs)

This is entirely for MQTT-based analytics. Uplinking packets multiple times means analytics tools and users can see SNRs for rebroadcasts after the first heard by each uplink, and it's especially useful for traceroutes, where entire paths that might otherwise not appear can come by way of later-heard copies of the packet.

My experience level with C++ is not amazing so I'm suspecting there are things here that aren't needed, despite my best efforts to rein in the generated code. The primary parts of this PR are just moving MQTT uplink logic into a dedicated `Router::publishReceivedToMqtt` and then calling it from a variety of places to catch packets that otherwise wouldn't make it to the regular uplink location.

Some specific question areas are the necessity of lines 870-874 in Router.cpp, serving as a dummy implementation I guess, and whether the gate on MQTT being enabled at all should be moved to live outside the newly created function. I'll try to test it on a few more devices, but it's also not a huge change and it works on at least one esp32 and nrf52, so I'm not too worried it'll turn up issues.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other (please specify below) (LilyGo t3s3 v1)

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -88,6 +88,11 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define MESHTASTIC_PREHOP_DROP 1
 #endif
 
+// When enabled, uplink every seen packet copy to MQTT, including duplicates.
+#ifndef USERPREFS_MQTT_UPLINK_ALL_SEEN
+#define USERPREFS_MQTT_UPLINK_ALL_SEEN 0
+#endif
+
 /// Convert a preprocessor name into a quoted string
 #define xstr(s) ystr(s)
 #define ystr(s) #s
--- a/src/mesh/FloodingRouter.cpp
+++ b/src/mesh/FloodingRouter.cpp
@@ -34,7 +34,7 @@ bool FloodingRouter::shouldFilterReceived(const meshtastic_MeshPacket *p)
     // Handle hop_limit upgrade scenario for rebroadcasters
     if (wasUpgraded && perhapsHandleUpgradedPacket(p)) {
 #if USERPREFS_MQTT_UPLINK_ALL_SEEN
-        publishReceivedToMqtt(const_cast<meshtastic_MeshPacket *>(p), DecodeState::DECODE_FAILURE);
+        publishReceivedToMqtt(p, DecodeState::DECODE_FAILURE);
 #endif
         return true; // we handled it, so stop processing
     }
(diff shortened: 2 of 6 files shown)


fifieldt:
Code looks fine, but is this feature appropriate for the general firmware, as opposed to a private fork? It seems to be a specific use case, since to do something useful with it additional software on the other end would be required too.

Maybe could be re-written as a module, so it gets packets out of the pipeline, rather than changes to the core router code?

ianmcorvidae:
I'm not sure, I guess. Most tools that ingest MQTT data for analytics that I've seen already do fine with this (e.g. MeshView), not least of all since we had an accidental version of this in the hop-upgrade code for a few releases. I made it gated behind compile flags/userPrefs since it's inappropriate for the public MQTT probably, though only in a "bit more traffic" way, not a "cause problems for nodes" sort of way. Given it requires compiling custom anyway, I'm not terribly opposed to it being kept separate if that's more of the firmware dev consensus (despite of course preferring not to need to maintain a fork!), but it would make the fork maintenance easier if we could still bring in the refactor splitting off the reusable uplink logic.

I have it on my todo list to go through the copilot review comments soon as well, so I'll try to keep an eye on the possibility of doing it in a module, though my feeling is the MQTT stuff is rather deeply embedded right now and it might need hook points we don't currently have.

ianmcorvidae:
I should also mention that testing this the past week or so I've spotted a gap in my logic, which is that it still only ever uplinks one copy of any packet sent by itself (so no first-rebroadcasts heard by the MQTT uplink of its own packets). This might not be too bad to leave but if it's not more intrusive than what's already here I figured I'd try to patch that part.

vidplace7:
First off, I love this! ❤️ 

I'd like to see this change become configurable at *runtime* (not hidden behind a userPref), in addition to adding some guards to keep it from being used on the public Meshtastic MQTT broker.

Jckf:
Supporting a runtime config option for this. Would be very useful to monitor and debug routing.

PR #10542, opened 2026-05-24
feat(http): expose IPv6 address in /json/report

Xavierhorwood:
Add ipv6_local and ipv6_global string fields to the wifi object emitted by handleReport() to match IPv4.

if IPv6 is off both fields are omitted from the response.
if IPv6 is on and has a IPV6 address it will return the global address and link local.
if IPv6 is on and does not have a IPV6 global address it will return "::".

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [x] LilyGo T-Beam - MESHTASTIC_EXCLUDE_WEBSERVER was set to 0 for this test
  - [ ] RAK WisBlock 4631
  - [x] Other Heltec (Lora32) V4.2

## Summary by CodeRabbit

* **New Features**
  * Device status and report responses now include IPv6 details when IPv6 is enabled on supported ESP32 builds.
  * Reports can display both global and local IPv6 addresses, providing more complete network information for device diagnostics and connectivity monitoring.


Simplycissmus:
I read this one from the consumer side - we poll `/json/report` from a dashboard that watches a fixed node - so the shape of the object matters to me as much as the values.

First, the thing that would have blocked it if it were true, and is not: the guards look asymmetric (`>= VAL(3,0,0)` at the use site, `<= VAL(3,0,0)` on the new declaration), but they work out. `GlobalIPv6()` is defined in `WiFiAPClient.cpp` under exactly the same `<=` guard, so the declaration matches its definition, and the only overlap is `== VAL(3,0,0)`, where the use site takes the other branch and the symbol is simply unused. It compiles on all three sides of the boundary.

Worth knowing while reviewing it: with the framework the repo currently builds against (arduino-esp32 3.x) the `#else` branch is dead, so both the legacy call and the new header declaration exist only for a core no supported build uses. That is not a reason to change anything, but it does mean the interesting half of this diff has never been compiled in CI.

Two API points, and they are the ones that get expensive after the fact:

**The keys are conditional.** `ipv6_global` and `ipv6_local` appear only when `config.network.ipv6_enabled`. Everything else in the `wifi` object is unconditional - `ip` is emitted even when there is nothing to report. A consumer now has to handle two shapes of the same object and cannot tell "IPv6 is off" from "this firmware predates the feature". Emitting both keys always, empty when disabled, costs a few bytes and keeps the schema single-shaped.

**`ipv6_local` under-describes what it holds.** It is the link-local address - the 3.x branch calls `WiFi.linkLocalIPv6()` for it. `ipv6_link_local` would say so. Once this ships, a rename breaks every consumer, so it is worth spending the extra characters now.

Neither is a correctness problem; both are the kind of thing that is free to fix today and awkward in a year.

PR #10135, opened 2026-04-12
feat: add Raspberry Pi Pico 2 + W5500 + E22-900M30S variant

cvaldess:
## Summary

- Adds community variant for **Raspberry Pi Pico 2** (RP2350, 4 MB flash) with an external **WIZnet W5500** Ethernet module and an **EBYTE E22-900M30S** LoRa module (SX1262, 30 dBm PA, 868/915 MHz)
- Reuses existing `WIZNET_5500_EVB_PICO2` code paths in `ethClient.cpp` for W5500 Ethernet initialization on SPI0
- DHCP timeout reduced from 60 s to 10 s to avoid blocking LoRa startup when no DHCP server is present

## Hardware

| Component | Model | Notes |
|-----------|-------|-------|
| MCU | Raspberry Pi Pico 2 | RP2350 @ 150 MHz, 512 KB RAM, 4 MB flash |
| Ethernet | W5500 module | Any WIZnet W5500 breakout, SPI0 |
| LoRa | EBYTE E22-900M30S | SX1262 + 30 dBm PA, 868/915 MHz, SPI1 |

## Pin assignments

- **LoRa SPI1**: SCK=GP10, MOSI=GP11, MISO=GP12, CS=GP13, RST=GP15, DIO1=GP14, BUSY=GP2, RXEN=GP3
- **Ethernet SPI0**: MISO=GP16, CS=GP17, SCK=GP18, MOSI=GP19, RST=GP20
- **GPS UART1/Serial2**: TX=GP8, RX=GP9

## RF configuration

- `SX126X_ANT_SW 3` — GP3 (RXEN) driven HIGH at init and held there permanently (LNA always active)
- `SX126X_DIO2_AS_RF_SWITCH` — SX1262 drives DIO2 HIGH during TX; a DIO2→TXEN bridge on the E22 module activates the PA automatically (no extra GPIO needed)
- `SX126X_DIO3_TCXO_VOLTAGE 1.8` — TCXO controlled via DIO3

> **Note:** The E22-900M30S requires a short wire or solder bridge between its DIO2 and TXEN pins. Without this bridge the module will not transmit.

## Test plan

- [x] Firmware builds cleanly (`pio run -e pico2_w5500_e22`)
  - RAM: 18.1% (94 KB / 512 KB)
  - Flash: 27.3% (978 KB / 3.58 MB)
- [x] LoRa RX/TX verified on hardware (868 MHz, E22-900M30S)
- [x] Ethernet DHCP acquisition verified
[...]

PR #10552, opened 2026-05-25
feat: add WIZnet W5500-EVB-Pico2 + E22P-868M30S variant

cvaldess:
## Summary

Adds a community variant for the **WIZnet W5500-EVB-Pico2** development board paired with an external **EBYTE E22P-868M30S** LoRa module. Hardware twin of [`variants/rp2350/diy/pico2_w5500_e22`](https://github.com/meshtastic/firmware/tree/develop/variants/rp2350/diy/pico2_w5500_e22) (added in #10135) — same LoRa pinout, same W5500 pin mapping — reuses its `variant.h` verbatim via `-I variants/rp2350/diy/pico2_w5500_e22`, so there is no duplicated pinout file.

## Why a separate env

Two differences vs `pico2_w5500_e22` make the existing env unsafe to use for this board:

1. **Board target.** The WIZnet W5500-EVB-Pico2 PCB carries a 2 MB Q-SPI flash chip; a stock Raspberry Pi Pico 2 has 4 MB. Selecting ` board = wiznet_5500_evb_pico2` makes the linker fail fast on overflow instead of producing a UF2 that gets silently truncated when flashed to the device.
2. **W5500 PHY is integrated** on the PCB (hard-wired SPI0 GP16-20) rather than supplied as an external module the user solders manually.

## E22P vs E22

The newer E22**P**-868M30**S** replaces the older E22-900M30S's split RXEN/TXEN pins with a single combined **RFEN** pin (LNA + PA enable). Firmware drives this with:

- `SX126X_ANT_SW 3` — GP3 (RFEN) held HIGH at boot, never toggled afterwards;
- `SX126X_DIO2_AS_RF_SWITCH` — SX1262 drives DIO2 HIGH during TX, enabling TXEN via the on-module DIO2 → TXEN bridge.

Both behaviours are inherited verbatim from the `pico2_w5500_e22` `variant.h`.

## Files

- `variants/rp2350/diy/wiznet_5500_evb_pico2_e22p/platformio.ini` — env definition; reuses the shared `variant.h` via `-I`.
- `variants/rp2350/diy/wiznet_5500_evb_pico2_e22p/README.md` — board overview, pinout, build/flash instructions.

No changes to shared code or other variants.

## Test plan

- [x] Build `pio run -e wiznet_5500_evb_pico2_e22p` SUCCESS (RAM 19.2 %, Flash 65.6 % of 1.5 MB app partition on the 2 MB chip).
- [x] Verified on hardware: flashed UF2 to a WIZnet W5500-EVB-Pico2, board boots, Ethernet up via DHCP, LoRa TX/RX via E22P-868M30S module confirmed end-to-end on a 868 MHz mesh.
- [x] No impact on `pico2_w5500_e22` build (variant.h shared but unchanged); existing builds keep working.


cvaldess:
@thebentern, thanks :)

PR #10568, opened 2026-05-28
TCA8418 keypad interface optimisations

Slankydudl:
Altered the 12 key multitap implementation to make it work more intuitively and contextually aware. Allowing for fairly efficient UI navigation and text entering with no additional interface, broadly mimicking a trackball when in UI and keypad in free text.

## Changes
- Keypad layout starts with alpha to make entering text faster.
- Keypad defaults to a navigation keymap when not in a free text entering state (navigation retained on long press).
- Navigation keys repeatedly fire in all contexts until released.
- Reply with freetext destination defaults to last viewed message context.
- Added key to enter reply with freetext without having to naviagate.
- Added more long press key functions in free text, to quickly change dest, exit free text, or send the message.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [X] Heltec V4
  - [X] Heltec T114
  - [X] Seed XIAO nRF52840 & Wio-SX1262 Kit


Slankydudl:
need to see how this can be merged with #7989, so far seems not to break the existing keypad implementation

PR #10571, opened 2026-05-28
feat: Skip Bluetooth wait when Bluetooth is disabled

h3lix1:
Make the PowerFSM DARK-to-LS timeout immediate when Bluetooth support is compiled out or config.bluetooth.enabled is false. The configured wait_bluetooth_secs default behavior is unchanged when Bluetooth is enabled.

Without this change, the firmware still waits wait_bluetooth_secs even when config.bluetooth.enabled == false

This is part of the ESP32 Power Saving series of diffs. 

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
xiao wio esp32-s3


mverch67:
Note: this PR breaks the PowerFSM _wait_for_secs_ transitions which cause other side effects like alert sound is cut off

PR #10136, opened 2026-04-12
feat: add Ethernet OTA support for RP2350/W5500 boards

cvaldess:
## Summary

- Adds over-the-air firmware update via Ethernet (W5500) for RP2350-based boards
- New TCP server on port 4243 using the **MOTA protocol**: SHA256 challenge-response auth + CRC32-verified transfer
- Python upload tool (`bin/eth-ota-upload.py`) — stdlib-only, auto GZIP-compresses firmware before upload
- Watchdog (8s) enabled in `rp2040Loop()` so the device recovers if an OTA write stalls

> **Depends on:** #10135 (Raspberry Pi Pico 2 + W5500 + E22-900M30S variant) — please review that PR first.

## Protocol

1. Device sends a 32-byte random nonce
2. Client responds with `SHA256(nonce || PSK)` — constant-time comparison on device side
3. Client sends 12-byte header: `"MOTA"` magic + firmware size (LE u32) + CRC32 (LE u32)
4. Firmware transferred in 1 KB chunks; device verifies CRC32 and calls `Update.end()` (picoOTA)
5. Device reboots — bootloader applies the update from LittleFS

The PSK defaults to `meshtastic_ota_default_psk_v1!!!` and can be overridden per-build via `USERPREFS_OTA_PSK`.

## Files changed

| File | Change |
|------|--------|
| `src/mesh/eth/ethOTA.cpp` | New — OTA TCP server implementation |
| `src/mesh/eth/ethOTA.h` | New — public API: `initEthOTA()`, `ethOTALoop()` |
| `bin/eth-ota-upload.py` | New — Python upload tool |
| `src/mesh/eth/ethClient.cpp` | Hook `initEthOTA()` after connect, `ethOTALoop()` every 5s |
| `src/platform/rp2xx0/main-rp2xx0.cpp` | Enable watchdog 8s in `rp2040Loop()` |
| `variants/rp2350/pico2_w5500_e22/platformio.ini` | `filesystem_size=0.75m`, `-D HAS_ETHERNET_OTA` |

## Usage

```bash
# Build
pio run -e pico2_w5500_e22

# Upload OTA (device must be connected to Ethernet with IP assigned)
python bin/eth-ota-upload.py --host 192.168.1.100 .pio/build/pico2_w5500_e22/firmware-*.bin

# With custom PSK
python bin/eth-ota-upload.py --host 192.168.1.100 --psk mySecretKey firmware.bin
```

## Test plan

[...]

PR #10573, opened 2026-05-28
feat(eth): port mesh/http to RP2350 + W5500 (HTTP/80 + HTTPS/443)

cvaldess:
## Summary

Ports the phone-API endpoints that today only ship on ESP32 (`mesh/http/ContentHandler`) over to RP2350 boards with W5500 Ethernet. Adds:

- HTTP server on TCP/80 exposing `/api/v1/{fromradio,toradio}` over the existing arduino-libraries/Ethernet stack.
- HTTPS server on TCP/443 reusing the same handlers via a `IStreamReadWrite` abstraction and an mbedTLS source-compiled TLS engine with a self-signed ECDSA P-256 cert (SAN=IP) persisted under LittleFS.

The new feature is **off by default** — gated by `HAS_ETHERNET_API` (HTTP/80) and `HAS_ETHERNET_TLS_API` (HTTPS/443). No upstream variant enables these flags in this PR; W5500 variants can opt-in independently.

## Why two servers and not Mongoose / esp32_https_server?

The existing `mesh/http/` implementation depends on `esp32_https_server`, which is not portable to arduino-pico, and Mongoose's built-in TCP/IP stack conflicts with the active `EthernetServer` used by OTA / MQTT / NTP. The cleanest path was to keep using arduino-libraries/Ethernet for the socket layer and bring our own HTTP/TLS code on top, sharing handlers between the two transports.

## Architecture

```
                +----------+         +----------+
                | TCP/80   |         | TCP/443  |
                +----+-----+         +----+-----+
                     |                    | mbedtls_ssl_read/write
                     v                    v
       EthernetClientStream         TlsClientStream
                     \                  /
                      v                v
                   IStreamReadWrite (Print)
                              |
                              v
                       parseRequest()
                              |
                              v
                    handleApiClient()
                       |          |
                       v          v
                handleFromRadio  handleToRadio  (-> PhoneAPI)
```

- ` ethApiHandlers.{h,cpp}` — Request struct, parser (CORS + Content-Length), `EthHttpAPI : PhoneAPI` (`api_type = TYPE_HTTP`, `checkIsConnected = true`).
- `ethApiServer.{h,cpp}` — listener on TCP/80, accept loop on its own OSThread (20 ms tick).
- `ethTlsApiServer.{h,cpp}` — listener on TCP/443, same OSThread pattern, wraps `EthernetClient` in an `mbedtls_ssl_context`.
- `ethCert.{h,cpp}` — first-boot ECDSA P-256 keygen + self-signed cert with `KeyUsage(digitalSignature + keyEncipherment, critical)` + `ExtendedKeyUsage(serverAuth, critical)`. Generation runs once on a dedicated OSThread; result persisted under `/prefs/eth_*.der` and reloaded on subsequent boots.
- `scripts/add_mbedtls_sources.py` — vendors mbedTLS 3.6.2 (the copy already in pico-sdk) via `BuildSources()` so the build doesn't depend on a precompiled lib that arduino-pico doesn't ship.
- `src/mbedtls_user_config.h` — strips TIME / NET_IO / FS_IO / TIMING / PSA storage, defines `MBEDTLS_NO_PLATFORM_ENTROPY` (the platform check in `entropy_poll.c` is not flag-gated). Keygen and the f_rng path use pico-sdk's `get_rand_64()` directly, bypassing `mbedtls_entropy`.
- `rp2040Loop()` — adds an RP2xx0 hardware watchdog (`watchdog_enable(8000, true)`) pet from the main loop. Required because the TLS handshake on Chrome can hold the OSThread inside `mbedtls_ssl_read()` past 8 s; the netRecv poll loop pets the watchdog every 2 ms so a quiet client cannot starve it.

## Performance

HTTP/1.1 keep-alive across both servers — `handleApiClient` loops on the same connection until the peer closes or `parseRequest` hits its 3 s idle timeout. Measured against `client.meshtastic.org` doing its initial sync (~80 sequential `/api/v1/fromradio` polls = MyInfo + every Config_* + every ModuleConfig_* + every NodeInfo):

- Before keep-alive: 80 × ~625 ms ECDSA P-256 handshake = ~50 s of pure handshake overhead.
- After: one handshake (641 ms, CHACHA20-POLY1305), all 80 requests pipelined, full config + 40 NodeInfos in ~6-8 s. Per-request post-handshake latency ~10-25 ms.

## Validation

Validated end-to-end against a WIZnet W5500-EVB-Pico2 + E22-900M30S board on the local network:

- `OPTIONS /api/v1/fromradio` → 204 with CORS + `X-Protobuf-Schema`.
- `PUT /api/v1/toradio` (ToRadio{want_config_id=1}) → 200 + echo.
- `GET /api/v1/fromradio?all=true` → 200 + ~3500 B (config + channels + nodeinfos).
- `GET /api/v1/nonexistent` → 404.
- HTTPS handshake via `curl -kv` and via Firefox / Chrome / Edge after accepting the self-signed cert; `curl` with two URLs reports `Reusing existing https: connection` and the server returns `Connection: keep-alive` explicitly.
- Watchdog reboot stress: held a Chrome `Test Connection` open through idle periods > 8 s — no watchdog reset, no socket exhaustion across back-to-back requests.

Cert generation timing: ~430 ms ECDSA P-256 cold boot on the dedicated OSThread (off the Periodic stack, off the Ethernet reconnect path).

## Build impact

Numbers from the wiznet variant on a downstream build that enables both flags:

- HTTP-only (`HAS_ETHERNET_API`): +~11 KB flash, +1.8 KB RAM.
- + HTTPS (`HAS_ETHERNET_TLS_API`): +~600 KB flash (mbedTLS + cert path), +~50 KB RAM. The cert key + DER blobs are heap-allocated only during gen and freed after persist.

## Not included

- No variant in this PR enables `HAS_ETHERNET_API` / `HAS_ETHERNET_TLS_API`. Activation lives in the W5500 variant configs (separate PRs).
- No OTA endpoints over HTTP/HTTPS. The existing `HAS_ETHERNET_OTA` path lives in #10136 and is unchanged. A follow-up commit can route OTA through `handleApiClient` once that lands.

## Files

- New: `scripts/add_mbedtls_sources.py`, `src/mbedtls_user_config.h`, `src/mesh/eth/{ethApiServer,ethApiHandlers,ethCert,ethTlsApiServer}.{cpp,h}`.
- Modified: `src/mesh/eth/ethClient.cpp` (2 init/loop hooks gated by the new flags), `src/main.cpp` + `src/main.h` (RP2040 loop hook), `src/platform/rp2xx0/main-rp2xx0.cpp` (watchdog enable + poke).

Net diff: 14 files changed, 1353 insertions(+), 1 deletion(-).

## Summary by CodeRabbit

* **New Features**
  * Added a TCP Ethernet HTTP API with keep-alive, bounded request parsing, CORS support, and `GET /api/v1/fromradio` / `PUT /api/v1/toradio` endpoints.
  * Added an HTTPS (TLS) API on RP2040 using self-signed certificates generated and cached on LittleFS, with automatic listener rebind when the certificate refreshes.
  * Added a bare-metal mbedTLS configuration to enforce TLS 1.2-only behavior.
* **Bug Fixes**
  * Fixed a missing conditional compilation terminator that could mis-route code for RP2040 builds.
* **Chores**
  * Added an automated build script to import required mbedTLS C sources into the firmware build.


markbirss:
@cvaldess

was able to compile and test over https with similiar hardware

cvaldess:
Thanks for the review. Addressed the build-portability points (1–6); details below, plus a note on the concurrency comment.

### Applied

**API handlers compile for HTTPS-only builds** (`ethApiHandlers.{h,cpp}`)
The TLS server depends on `handleApiClient` / `IStreamReadWrite` from this header, but the header was gated on `HAS_ETHERNET_API` alone, so a `HAS_ETHERNET_TLS_API`-only variant would have lost the symbols. Gate is now:
```cpp
#if HAS_ETHERNET && (defined(HAS_ETHERNET_API) || defined(HAS_ETHERNET_TLS_API))
```

**TLS/cert modules are RP2040-only** (`ethTlsApiServer.{cpp,h}`, `ethCert.{cpp,h}`)
These use Pico SDK APIs (`get_rand_64()`, `watchdog_update()`), so enabling `HAS_ETHERNET_TLS_API` on a non-arduino-pico Ethernet target (ESP32-S3 / nRF52) would have failed with cryptic missing-header errors. Tightened to:
```cpp
#if HAS_ETHERNET && defined(HAS_ETHERNET_TLS_API) && defined(ARCH_RP2040)
```
The TLS include + call-site guards in `ethClient.cpp` were tightened to match, so enabling the flag off-RP2040 is now a clean no-op rather than a link error. (`ARCH_RP2040` is also defined for RP2350 under arduino-pico, so the supported boards are unaffected — verified with a clean `wiznet_5500_evb_pico2_e22p` build that still compiles the TLS server.)

### Note on the `handleApiClient` concurrency comment

That one is a false positive for this codebase. `EthApiServerThread` and `EthTlsApiServerThread` are `concurrency::OSThread`s, which are **cooperatively** scheduled — their `runOnce()` methods are invoked sequentially from the main loop on a single core, never preemptively or in parallel. `handleApiClient()` runs to completion *inside* one `runOnce()` (it blocks for the whole client session — that's why the loop's `watchdog_update()` doesn't run while a client is being served), so the static `webAPI` `PhoneAPI` instance is only ever touched by one session at a time. There's no concurrent mutation of the state machine / protobuf scratch, and a mutex would be a no-op without preemption. The single shared `PhoneAPI` is also consistent with the existing `mesh/http` design. Left as-is.

PR #10576, opened 2026-05-29
Save Frame Visibility Actions

### LINKED ISSUE #10566: [Bug]: BaseUI 'Frame Visibility Toggles' settings don't persist after device restart (Wio L1)
### Category

Other

### Hardware

Seeed Wio Tracker 1110

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [x] BaseUI

### Firmware Version

2.7.15.567b8ea

### Description

It seems that the 'Frame Visibility Toggles' settings don't persist after device restarts on my Wio L1 Tracker Pro using BaseUI, either by restarting with the physical power button or a software restart using the UI. All other settings persist after restart such as Notifications (buzzer actions), Screen Options (brightness), Bluetooth toggle, GPS toggle, GPS Format, Compass mode. On the Wio L1 with BaseUI, these settings are configured on-device only (not using Bluetooth and phone app) using the joystick and selection buttons. 

The 'Frame Visibility Toggles' settings not sticking after a restart is annoying because I use this to minimize some of the menu options that I don't use and to hide them, such as Favorites, Node List, and Bearings; upon restart, I have the cluttered UI and must manually disable these again. It seems like a bug because all of the other settings persist after a restart.

### Relevant log output

```Shell

```

Xaositek:
Let's actually save frame visibility actions; tested and works with a reset or restart/shutdown action as it saves at the time of the change. Tested across some devices and all seem to work well.


src/graphics/Screen.cpp
```diff
@@ -1454,6 +1462,167 @@ bool Screen::isFrameHidden(const std::string &frameName) const
     return false;
 }
 
+// ---------------------------------------------------------------------------
```
fifieldt, line 1465: Is this better in a separate file?

Xaositek: It is Screen related so it makes sense to be there but open to discussion on it.

HarukiToreda (approved):

src/graphics/Screen.cpp
```diff
@@ -1454,6 +1462,167 @@ bool Screen::isFrameHidden(const std::string &frameName) const
...
+
+inline bool getBit(uint32_t mask, uint8_t bit)
+{
+    return (mask & (1UL << bit)) != 0;
+}
+} // namespace
+
+uint32_t Screen::packHiddenFrames() const
+{
+    uint32_t mask = 0;
```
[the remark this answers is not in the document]
Xaositek: Completely valid - but probably not going to be something anyone ever moves between devices?

PR #10582, opened 2026-05-30
Power airtime monotonic windows due to light sleep. 

h3lix1:
# Power Airtime Monotonic Windows

## Summary

This change updates airtime accounting to advance from monotonic uptime instead of the `AirTime` thread's once-per-second scheduler tick.

Before this change, `AirTime::runOnce()` incremented an internal `secSinceBoot` counter by one each time the thread ran. On devices using light sleep, the scheduler can be paused while real time continues. That means the channel-utilization and TX-utilization windows could decay according to awake scheduler time, not elapsed wall time.

After this change, airtime windows are synchronized from `millis()` before airtime is logged or read. Packets observed after wake are counted in the current elapsed-time bucket, and buckets crossed during sleep are cleared or rotated immediately.

## Motivation

The power-saving work makes ESP32 nodes enter light sleep quickly after packet processing. That exposed a mismatch in airtime accounting:

- Channel utilization is intended to be a rolling 60-second view.
- TX utilization is intended to be a rolling 60-minute view.
- Historical airtime reports are intended to use hour-sized periods.
- The old implementation advanced those windows only when the `AirTime` thread ran.

When a node slept frequently, old airtime could remain in the active channel-utilization window for longer than 60 real seconds. This could make the channel appear busy after the channel had actually been quiet from the node's point of view.

## Code Changes

The change is isolated to:

- `src/airtime.cpp`
- `src/airtime.h`

The new private helper is:

```cpp
void AirTime::syncNow();
```

`syncNow()` uses `millis()` as the time source. It intentionally does not use RTC, GPS, NTP, or mesh time, because wall-clock time can move backward or forward. Airtime windows should be based on monotonic uptime.

The helper is called from:

- `logAirtime()`
- `airtimeReport()`
- `getSecondsSinceBoot()`
- `channelUtilizationPercent()`
- `utilizationTXPercent()`
- `runOnce()`
- `airtimeRotatePeriod()`

Calling it from both readers and writers matters because a node may wake, receive a packet, or make a send-gating decision before the periodic `AirTime` thread runs again.

## Window Behavior

Channel utilization keeps the existing six 10-second buckets:

```cpp
#define CHANNEL_UTILIZATION_PERIODS 6
```
If one or more 10-second boundaries were crossed while the scheduler was paused, `syncNow()` clears each crossed bucket. If the elapsed time is large enough to cover the full 60-second window, it clears the whole channel-utilization window.
TX utilization keeps the existing 60 one-minute buckets:
```cpp
#define MINUTES_IN_HOUR 60
```

If one or more minute boundaries were crossed, `syncNow()` clears those buckets. If the elapsed time covers the full hour, it clears the full TX-utilization window.

Historical airtime reports keep the existing hour-sized periods:

```cpp
#define SECONDS_PER_PERIOD 3600
#define PERIODS_TO_LOG 8
```
If one or more hour boundaries were crossed, `syncNow()` rotates the report periods. If the elapsed time covers all retained periods, it clears the full report window.
## Expected Runtime Impact
The main behavioral change is that channel utilization and TX utilization decay according to elapsed monotonic uptime, including time spent in light sleep.
This can affect send-gating decisions that use:
```cpp
airTime->isTxAllowedChannelUtil(...)
airTime->isTxAllowedAirUtil()
```
## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Heltec v4.3

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved airtime tracking accuracy after light sleep or other scheduling pauses.
  * Ensured transmitted and received packets are attributed to the correct current time period.
  * Updated airtime and utilization reporting to consistently reflect elapsed time.
  * Added reliable catch-up behavior for rolling airtime/utilization windows after extended inactivity.

### DIFF
--- a/src/airtime.cpp
+++ b/src/airtime.cpp
@@ -12,6 +12,8 @@ uint32_t air_period_rx[PERIODS_TO_LOG];
 
 void AirTime::logAirtime(reportTypes reportType, uint32_t airtime_ms)
 {
+    // A packet may be logged immediately after waking from light sleep. Sync first so
+    // the packet is counted in the current wall-time bucket, not a stale awake-time bucket.
     syncNow();
 
     if (reportType == TX_LOG) {
@@ -50,11 +52,14 @@ uint8_t AirTime::getPeriodUtilHour()
 
 void AirTime::airtimeRotatePeriod()
 {
+    // Preserve the public helper while keeping all rotation logic in one monotonic-time path.
     syncNow();
 }
 
 void AirTime::syncNow()
 {
+    // Use monotonic uptime rather than RTC/network time; user, GPS, or NTP clock changes
+    // must not move airtime accounting backward or forward.
     uint32_t nowMsec = millis();
 
     if (firstTime) {
@@ -67,6 +72,7 @@ void AirTime::syncNow()
         memset(air_period_rx, 0, sizeof(air_period_rx));
 
         this->secSinceBoot = nowMsec / 1000;
+        // Keep the checkpoint on a whole-second boundary so elapsedSecs advances predictably.
         this->lastSyncMsec = nowMsec - (nowMsec % 1000);
         this->lastUtilPeriod = this->getPeriodUtilMinute();
         this->lastUtilPeriodTX = this->getPeriodUtilHour();
@@ -85,6 +91,8 @@ void AirTime::syncNow()
     this->secSinceBoot += elapsedSecs;
     this->lastSyncMsec += elapsedSecs * 1000;
 
+    // Historical airtime reports use 1-hour buckets. If multiple hours elapsed while
+    // asleep, rotate each crossed bucket or clear the whole report window.
     uint32_t elapsedAirtimePeriods = (this->secSinceBoot / SECONDS_PER_PERIOD) - (oldSecSinceBoot / SECONDS_PER_PERIOD);
     if (elapsedAirtimePeriods >= PERIODS_TO_LOG) {
         memset(this->airtimes.periodTX, 0, sizeof(this->airtimes.periodTX));
@@ -112,6 +120,8 @@ void AirTime::syncNow()
     }
     this->airtimes.lastPeriodIndex = this->currentPeriodIndex();
 
+    // Channel utilization is a rolling 60-second view split into six 10-second buckets.
+    // Clear every bucket crossed while asleep so old airtime decays by real elapsed time.
     uint32_t elapsedUtilPeriods = (this->secSinceBoot / 10) - (oldSecSinceBoot / 10);
     if (elapsedUtilPeriods >= CHANNEL_UTILIZATION_PERIODS) {
         memset(this->channelUtilization, 0, sizeof(this->channelUtilization));
@@ -122,6 +132,7 @@ void AirTime::syncNow()
     }
     this->lastUtilPeriod = this->getPeriodUtilMinute();
 
+    // TX utilization is a rolling 60-minute view used by duty-cycle checks.
     uint32_t elapsedUtilTXPeriods = (this->secSinceBoot / 60) - (oldSecSinceBoot / 60);
     if (elapsedUtilTXPeriods >= MINUTES_IN_HOUR) {
         memset(this->utilizationTX, 0, sizeof(this->utilizationTX));
@@ -135,6 +146,7 @@ void AirTime::syncNow()
 
 uint32_t *AirTime::airtimeReport(reportTypes reportType)
 {
+    // Reports may be requested before runOnce() executes after wake.
     syncNow();
 
     if (reportType == TX_LOG) {
@@ -159,12 +171,14 @@ uint32_t AirTime::getSecondsPerPeriod()
 
 uint32_t AirTime::getSecondsSinceBoot()
 {
+    // Keep HTTP/debug reporting aligned with the same monotonic clock used by the buckets.
     syncNow();
     return this->secSinceBoot;
 }
 
 float AirTime::channelUtilizationPercent()
 {
+    // Gate decisions should see buckets that have decayed across light-sleep time.
     syncNow();
 
     uint32_t sum = 0;
@@ -177,6 +191,7 @@ float AirTime::channelUtilizationPercent()
 
 float AirTime::utilizationTXPercent()
 {
+    // Duty-cycle checks use this value, so keep it current even outside the periodic thread.
     syncNow();
 
     uint32_t sum = 0;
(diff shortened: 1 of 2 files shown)


m1nl:
need to review your code more carefully, but isn't that duplicate of [9778](https://github.com/meshtastic/firmware/pull/9778), which looks like has been closed as I didn't receive feedback / review from maintainers?

h3lix1:
@m1nl I guess we'll see what the devs decide with #9778  - I didn't see you made this change before submitting this one, but my clanker has shown concerns.

The main it sees with #9778 is the use of buckets that could potentially roll and not get cleared correctly if 13 periods elapsed (for example, with CHANNEL_UTILIZATION_PERIODS = 6) and it won't clear the values for all the other buckets.

This PR tracks elapsed time instead instead of buckets, so it avoids this issue.

Also, secSinceBoot is only updated after runOnce, but logAirtime could choose the wrong bucket since a node can wake from sleep before it runs another runOnce loop, causing secSinceBoot to be stale.

Using syncNow() helps avoid these issues, but it is a bigger change.

I don't have much skin in this game - but the other patch could introduce some issues as-is.

Exact text from the Clod:

### 1. The window-clearing only handles a single wrap

A gap larger than one full window (exactly what light sleep produces) clears the wrong buckets. Trace, `channelUtilization` (6 buckets × 10 s): say `lastUtilPeriod = 2`, the device light-sleeps ~130 s, so on the next run `utilPeriod = (2 + 13) % 6 = 3`.

```
lastUtilPeriod = (2 + 1) % 6 = 3
3 > 3 ? no
for i = 3..3: clear [3]      // clears ONE bucket
```

After a 130-second sleep it clears a single bucket and leaves `[0,1,2,4,5]` holding minute-old airtime. And when the gap is an exact multiple of the window (`elapsed ≡ 0 mod 6`, e.g. ~60 s), `lastUtilPeriod == utilPeriod`, the `if` is skipped entirely, and **nothing** is cleared — the whole stale window is reported as live.

Because `channelUtilizationPercent()` drives CSMA contention-window sizing (`getTxDelayMsec`) and the `isTxAllowedChannelUtil` TX gate, this can make a freshly-woken node back off or refuse to send on an idle channel — silently, with no log. The same flaw affects `utilizationTX` (60 × 1 min) for sleeps > 60 min.

The root cause is that `lastUtilPeriod` stores a *mod* value, so you can't recover how many periods actually elapsed. Track the **absolute** period count and clamp the clear to the window size, and the wrap-detection block disappears entirely:

```cpp
} else {
    this->airtimeRotatePeriod();

    // Clear every 10-second channel-utilization bucket crossed since the last run.
    // Clamping to the window size means a long gap (light sleep) clears the whole
    // window instead of mis-wrapping.
    uint32_t curUtilPeriod = this->secSinceBoot / 10;
    uint32_t crossed = curUtilPeriod - lastUtilPeriod;
    if (crossed > CHANNEL_UTILIZATION_PERIODS)
        crossed = CHANNEL_UTILIZATION_PERIODS;
    for (uint32_t i = 1; i <= crossed; i++)
        this->channelUtilization[(lastUtilPeriod + i) % CHANNEL_UTILIZATION_PERIODS] = 0;
    lastUtilPeriod = curUtilPeriod;

    // Same for the 1-minute TX-utilization buckets.
    uint32_t curUtilPeriodTX = this->secSinceBoot / 60;
    uint32_t crossedTX = curUtilPeriodTX - lastUtilPeriodTX;
    if (crossedTX > MINUTES_IN_HOUR)
        crossedTX = MINUTES_IN_HOUR;
    for (uint32_t i = 1; i <= crossedTX; i++)
        this->utilizationTX[(lastUtilPeriodTX + i) % MINUTES_IN_HOUR] = 0;
    lastUtilPeriodTX = curUtilPeriodTX;
}
```

This needs the two trackers widened to absolute counters in `airtime.h`:

```cpp
uint32_t lastUtilPeriod = 0;    // absolute count of 10s periods, not (… % CHANNEL_UTILIZATION_PERIODS)
uint32_t lastUtilPeriodTX = 0;  // absolute count of 1-min periods, not (… % MINUTES_IN_HOUR)
```

and the `firstTime` branch should seed them with the absolute value (`lastUtilPeriod = secSinceBoot / 10;` / `lastUtilPeriodTX = secSinceBoot / 60;`) instead of the `% …` values.

### 2. State is only refreshed inside `runOnce`, so reads/writes can still be stale after wake

Separately from the clearing math: both `secSinceBoot = millis()/1000` *and* the bucket clearing happen **only** in `runOnce`. Nothing on the read or write paths self-refreshes, so until the thread next ticks after wake:

- **Packet logging** — `logAirtime()` selects its bucket via `currentPeriodIndex()` / `getPeriodUtilMinute()`, which derive from `secSinceBoot`. A packet RX'd/TX'd after wake but before `AirTime::runOnce` runs is written into the *pre-sleep* bucket, and may then be clobbered when `runOnce` rotates a moment later — the airtime is mis-attributed or lost.
- **Duty-cycle / CSMA decisions** — the buckets aren't decayed until `runOnce`, so `isTxAllowedChannelUtil`, `isTxAllowedAirUtil`, and the CSMA reads in `getTxDelayMsec`/`getRetransmissionMsec` can see un-decayed (too-high) utilization right after wake and wrongly skip a send or oversize the backoff on an idle channel.
- **HTTP/debug** — `/json/report` momentarily shows a stale `seconds_since_boot` and utilization (informational only).

Note this isn't bounded to "≤1 s": OSThreads run cooperatively in registration order each loop pass, and there's no guarantee `AirTime::runOnce` runs before `RadioLibInterface::runOnce` (which calls `logAirtime`) or before a `Router::send` in the *same* first post-wake iteration. Fixing the clearing math in (1) doesn't close this — the refresh has to move onto the read/write paths (e.g. a small `sync()` helper called at the top of `logAirtime`, the percent getters, and `getSecondsSinceBoot`) rather than living only in `runOnce`.

### Also (minor, separate)

`airtimeRotatePeriod()` still shifts the hourly `periodTX/RX` arrays by at most one slot per call, so a multi-hour gap rotates only once and leaves stale hours in the report. The same clamped-elapsed approach fixes it there too.

h3lix1:
@fifieldt @thebentern One of these need to make it into develop soon (either this or #9778) If we're moving to more tick-free power saving modes. It's very obvious with light sleep that utilization numbers are off when the CPU is not ticking away. Mine was completely vibe coded - maybe @m1nl is less vibe coded, but there are edge cases with #9778 that aren't show stoppers.

NomDeTom:
sorry, got carried away with Clod and added an uneccesary rebase.

NomDeTom:
Closed as https://github.com/meshtastic/firmware/pull/11291 has landed. Authorial credit is maintained - many thanks!

PR #10482, opened 2026-05-15
Fix SHT2x detection for INA219 addresses

balya:
## Summary

Fixes #10110.

This PR fixes false-positive SHT2x/SHTxx detection on I2C addresses shared with INA219/INA-series sensors.

INA219 can respond on `0x40`/`0x41`, which overlaps with SHT2x probing. The previous SHT2x serial-number probe only checked that enough bytes were returned, so an INA219 could be incorrectly registered as `SHTXX`. That caused environment telemetry initialization to fail and prevented INA219 voltage readings from being reported correctly.

This change validates the SHT2x serial-number response using the Sensirion CRC-8 checksum before accepting the device as SHT2x. If the CRC validation fails, detection falls through to the existing INA detection path.

## Testing
Tested on an ESP32-S3 E22 W5500 node with INA219 at I2C address `0x41`.

Before this change, boot logs showed:
```text
SHTXX (SHT2X) found at address 0x41
Init sensor: SHTXX
SHTXX: init(): failed
```

After this change, the same hardware is detected correctly:
```text
Init sensor: INA219
Opened INA219 sensor on i2c bus
Using INA on I2C addr 0x41 for charging detection
DeviceTelemetry Send: ... battery_level=100, voltage=4.968000
```

## 🤝 Attestations
- [x] I have tested that my proposed changes behave as described.
- [x] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [x] Other: ESP32-S3 E22 W5500 with INA219 on I2C address 0x41

PR #10587, opened 2026-05-30
Add SHT2x CRC check to fix i2c scan hang

ascendr:
Commit 2f19a1d7a introduced SHT2x probing on INA shared address 0x40 and the associated regression.

The seeed_xiao_nrf52840_kit_i2c startup scan was hanging at the shared 0x40 I2C address because the new SHT2x/SHT21 fallback probe preceded and not specific, stealing the INA219 address. The probe logic could loop unsafely while draining responses, and it treated any returned bytes as a successful SHT match, which let an INA219 be misidentified as SHT2X preventing INA telemetry from being registered.

I've fixed it by making the SHT probe validate the Sensirion CRC-framed serial-number reads instead of accepting arbitrary bytes, and by keeping the INA219 fallback only when the SHT probe does not genuinely match.

Validate on Seeed Studio nRF52840 kit with build env seeed_xiao_nrf52840_kit_i2c

### DIFF
--- a/src/detect/ScanI2CTwoWire.cpp
+++ b/src/detect/ScanI2CTwoWire.cpp
@@ -179,8 +179,28 @@ String readSEN5xProductName(TwoWire *i2cBus, uint8_t address)
 #endif
 
 #if HAS_TELEMETRY && !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR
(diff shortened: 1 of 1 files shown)


oscgonfer:
Duplicate of https://github.com/meshtastic/firmware/pull/10482?

caveman99:
fixed with #10482

PR #10390, opened 2026-05-04
Simplify tracking of BLE connection handle & improve thread safety.

cpatulea:
- Redefine isConnected in terms of nimbleBluetoothConnHandle. isConnected was not safe because BLEServer::getConnectedCount is not thread-safe (https://github.com/espressif/arduino-esp32/issues/12538) while isConnected is called from various threads. Now we can avoid checking bleServer every time before calling isConnected.
- getRssi: don't try to "populate nimbleBluetoothConnHandle", that requires calling BLEServer::getPeerDevices which is not thread-safe (same issue as above).

The code simplifications are intended to facilitate merging of the pioarduino branch (https://github.com/meshtastic/firmware/pull/9122), which makes many changes to NimBLE code.

## 🤝 Attestations

- [ ] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)

PR #10588, opened 2026-05-30
Fix null pointer dereference in updateBatteryLevel function

vidplace7:
Fixes crashes observed on T-Beam BPF at first-boot (possibly others?). Possibly caused by changes in #10390 @cpatulea 

This pull request makes a minor but important change to the Bluetooth battery level update logic. It adds a safety check to ensure that `nimbleBluetooth` is not a null pointer before accessing its methods, which helps prevent potential crashes if the object is not initialized.

- Added a null pointer check for `nimbleBluetooth` in the `updateBatteryLevel` function to prevent dereferencing a null pointer.


cpatulea:
Apologies, stupid mistake. This fix LGTM.

PR #10592, opened 2026-05-31
Report on-chip temperature as environment telemetry

Komzpa:
## Summary

- Adds an on-chip temperature telemetry sensor that reports die temperature through the existing environment telemetry temperature field.
- Enables the provider by default at compile time with `MESHTASTIC_ENABLE_ONCHIP_TEMPERATURE_SENSOR=0` available for builds that want to remove it.
- Runs the on-chip provider after detected external environmental sensors so it only fills the temperature field when no real ambient-temperature sensor already did.
- Supports MCU die-temperature reads on ESP32, nRF52, and RP2040/RP2350, plus LR11xx radio die-temperature reads through the Meshtastic `RadioLibInterface` adapter.
- Rejects implausible chip-temperature readings outside `(-45, 130)` C or non-finite values; ESP32 reads use five quick samples, require at least three plausible samples, and yield between samples instead of using timed sleeps.
- Emits debug logs with the raw on-chip reading, accepted telemetry value, and ESP32 accepted sample count.

## Chip coverage

Implemented MCU paths:

- ESP32-family Arduino `temperatureRead()`
- nRF52 SoftDevice `sd_temp_get()`
- RP2040/RP2350 Arduino-Pico `analogReadTemp()`

Implemented modem path:

- LR11xx `getTemp()` via a Meshtastic-side wrapper because RadioLib keeps the calibrated accessor protected.

Not reported from this PR:

- SX127x/RF95 raw temperature, because RadioLib exposes only an uncalibrated raw read that changes radio operating mode.
- SX126x/SX128x/LLCC68, because the pinned RadioLib and Semtech SX126x driver APIs do not expose a calibrated temperature accessor for them.

## Validation

- Head `d762e9707`
- `git diff --check`
- `/home/kom/.platformio/penv/bin/platformio run -e coverage`
- `/home/kom/.platformio/penv/bin/platformio run -e heltec-v3`
- Flashed only the `heltec-v3` app partition to an attached Heltec V3 without erasing config/NVS/filesystem.
- Confirmed the flashed Heltec V3 booted `2.8.0.5bd5969`/`heltec-v3` with environment telemetry still enabled after app-only flashing.
- Heltec V3, USB powered, local environment telemetry request loop:
  - idle/stable baseline: 10/10 replies, MCU die temperature reported `52.3`, `51.9`, `52.9`, `52.9`, `52.9`, `52.9`, `53.3`, `53.3`, `53.7`, `53.9` C.
  - response/activity window: 14/14 replies, value moved from the stable `53.9` C plateau to `54.7` C and then `54.9` C.
  - cooldown/recovery window after a quiet delay: 8/8 replies, value came down from `55.7` C to the `54.9`-`55.1` C band.
  - after replacing ESP32 inter-sample timed sleeps with scheduler yields: 5/5 replies, all `53.9` C.
- Inspected the installed ESP32 Arduino core for `temperatureRead()` failure behavior and reject non-finite readings.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/mesh/LR11x0Interface.h
+++ b/src/mesh/LR11x0Interface.h
@@ -6,6 +6,15 @@
  * \brief Adapter for LR11x0 radio family. Implements common logic for child classes.
  * \tparam T RadioLib module type for LR11x0: SX1262, SX1268.
  */
+// RadioLib keeps getTemp() protected; this exposes it only inside the Meshtastic adapter.
+template <class T> class TemperatureReadableLR11x0 : public T
+{
+  public:
+    explicit TemperatureReadableLR11x0(Module *mod) : T(mod) {}
+
+    bool readChipTemperature(float &temperature) { return this->getTemp(&temperature) == RADIOLIB_ERR_NONE; }
+};
+
 template <class T> class LR11x0Interface : public RadioLibInterface
 {
   public:
@@ -27,6 +36,15 @@ template <class T> class LR11x0Interface : public RadioLibInterface
 
     bool isIRQPending() override { return lora.getIrqFlags() != 0; }
 
+    bool getChipTemperature(float &temperature) override
+    {
+        if (isSending() || isReceiving) {
+            return false;
+        }
+
+        return lora.readChipTemperature(temperature);
+    }
+
 #ifdef LR11X0_AGC_RESET
     void resetAGC() override;
 #endif
@@ -35,7 +53,7 @@ template <class T> class LR11x0Interface : public RadioLibInterface
     /**
      * Specific module instance
      */
-    T lora;
+    TemperatureReadableLR11x0<T> lora;
 
     int16_t getCurrentRSSI() override;
 
--- a/src/mesh/RadioLibInterface.h
+++ b/src/mesh/RadioLibInterface.h
@@ -238,6 +238,11 @@ class RadioLibInterface : public RadioInterface, protected concurrency::Notified
      */
     bool randomBytes(uint8_t *buffer, size_t length);
 
+    /**
+     * Read radio die temperature in Celsius, if the active modem exposes a calibrated reading.
+     */
+    virtual bool getChipTemperature(float &temperature) { return false; }
+
   private:
     uint8_t getNoiseFloorSampleCountInternal() const;
     int32_t getAverageNoiseFloorInternal() const;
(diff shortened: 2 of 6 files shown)


phaseloop:
Well, every 6 months someone tries to implement it (myself included) but those PR's are rejected claiming that on-chip temp is not an environment reading (which is true).

This is useful, for example on remote nodes to get a grasp how cold or hot the setup is in general (like a stratospheric baloon) but AFAIK it would make more sense to have this as separate telemetry attribute like system temp - configurable to use either on-chip or i2c sensor inside the case. And this must be switchable in UI (default off) so people don't spam RF with system temp no one cares about.

FFAMax:
>on-chip temp is not an environment reading (which is true)

That is not true. Measurements are providing some correlation with external environments and the only the question - design and accuracy. So please do not influent on the public who can treat your comment seriously. Internal sensor is not dummy device.

>And this must be switchable in UI (default off) 

100% agree

>so people don't spam RF with system temp no one cares about.

@phaseloop, look, I am not blaming you and your point, but let's try to see the entire picture. Imaging the node with internal sensor in snow environment with almost 0 watt power consumptions (hight temp precision) and ~0 solar insolation - do you see it more useful now in environments where no other sensors at all? We never know how and for whom it can be useful.

basement-labs:
@FFAMax - is "internal environment" temperature and should be marked as such while general telemetry is expected to show regular air temp. 

"So please do not influent on the public who can treat your comment seriously. Internal sensor is not dummy device." 

Yes, I'm planning to influent public and development cycle indeed. And my comments should be treated seriously. Internal sensor is not a dummy device but it's measurement should be marked properly to avoid confusion.

It is useful but it should be designed and marked properly.

FFAMax:
>Internal sensor is not a dummy device but it's measurement should be marked properly to avoid confusion.

100%


So, to keep it clear for others who not that deep in the topic: If locate external temperature sensor (yellow arrow) close by main NRF chip (lower power consumption = low heat)(on the top) it will make almost no sense in close environments like a closed box. But it will cost more money, it will cost more waste for the planet, it will waste more space on the PCB.
So it literally make almost 0 sense how to call it - internal or external. Make sense how it used and what the purpose of it.
image

100% we should not trust to any sensor inside the case if aggressive charging is ongoing, or during and after active power consumption.

PR #10617, opened 2026-06-03
fix: FEM Sleep/Wake Fix - Enable PA after sleep

h3lix1:
# Heltec v4.3 (and others) FEM Sleep/Wake Fix

## Summary

Fixes Heltec v4.3/KCT8103L FEM handling when waking from MCU sleep so transmit mode can drive the PA path at full power.

## Root Cause

The MCU sleep path RTC-held the FEM GPIOs in a receive state, but the active TX/RX paths did not release those holds before changing the pins. On ESP32-S3 this can leave KCT8103L control pins latched from the sleep receive state, so wake-from-sleep transmissions may not enter the full TX PA state.

## Changes

- Release ESP32 GPIO/RTC holds on FEM control pins before entering active TX/RX or changing FEM sleep state.
- Power the KCT8103L FEM rail back on before active TX/RX, with a short settle delay of 5ms only when the rail was off.
- Turn off `LORA_PA_POWER` for KCT full radio sleep to avoid wasting power.
- Keep the rail on only for the explicit LoRa wake/listen sleep path.
- Put KCT MCU-sleep receive into RX bypass (`CTX=HIGH`) instead of keeping the external LNA active.
- Add FEM type log lines for Heltec v4 runtime detection and fixed FEM variants.

## Expected States

Full radio sleep:

```text
LORA_PA_POWER = 0
CSD           = 0
CTX           = don't care
```

MCU sleep with LoRa wake/listen:

```text
LORA_PA_POWER = 1
CSD           = 1
CTX           = 1   # KCT RX bypass
CPS/DIO2      = 0   # SX1262 RX path
```

Active TX:

```text
LORA_PA_POWER = 1
CSD           = 1
CTX           = 1
CPS/DIO2      = 1
```

## Validation

- Flashed a diagnostic build to a connected Heltec v4.3.
- Confirmed runtime FEM detection logs `Detected KCT8103L LoRa FEM`.
- Confirmed simulated sleep RX to wake TX transition drove `CTX` from `0` to `1` after releasing holds.
- Built `heltec-v4`.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
Heltec V4


h3lix1:
@thebentern Testing is going swimmingly well with this one.

PR #10625, opened 2026-06-03
Add MVP control point routing module for Meshtastic

dck5mfs74p-star:
## Experimental control point routing

This fork contains an experimental MVP implementation of a control-point-based routing approach for Meshtastic firmware.

### Implemented changes

- Added a custom protobuf message for control point metric exchange.
- Added `ControlPointModule` for receiving, storing, and broadcasting routing-related metrics.
- Integrated control point relay selection into `NextHopRouter`.
- Preserved fallback behavior to the default Meshtastic next-hop logic.

### Status

This implementation is an academic prototype created for research and thesis work. It is not a production-ready routing feature.


caveman99 (changes requested):
- don't check in generated files.
- dont check in unrelated scripts
- don't broadcast anything periodically on a ressource contrained network. Been there, done that. AKA Neighbourinfo. Realistically we'll never acceppt a patch that includes this behaviour. You can derive the route quality metrics from existing traffic passing through the node. This would be totally acceptable. but don't create additional traffic to improve traffic

caveman99:
Clearly doesn't listen to feedback. Closing burning dumpster PR.

Issue #6658, opened 2025-04-23
[Feature Request]: Native WireGuard VPN Support for Remote Access


mauro-midolo:
### Platform

ESP32

### Description

I would like to propose the addition of native WireGuard VPN client support within the Meshtastic firmware.​

### Use Case

The primary goal is to enable secure remote access to Meshtastic devices over the internet. By integrating WireGuard, users could establish encrypted tunnels directly to their devices, facilitating:​

Remote configuration and management without relying on intermediary gateways.

Secure data transmission over public networks.

Enhanced privacy and security for deployments in sensitive environments.​

### Current Workaround

At present, achieving remote access requires setting up an external gateway (e.g., a Raspberry Pi) that connects to the Meshtastic device via USB or Bluetooth and handles the VPN connection. While functional, this setup adds complexity and potential points of failure.​

### Benefits

- Simplified remote management of Meshtastic nodes.

- Reduced reliance on additional hardware.

- Improved security posture for deployments.​

fifieldt:
Ciao, and thanks for the idea.

It looks like the wireguard kernel module is about ~35kB and the userspace tools are another ~27kB. By meshtastic firmware standards, that's a non-trivial amount of space, so this would require careful consideration.

TheWISPRer:
I plan to test a rough implementation of this using the "XIAO ESP32S3 & Wio-SX1262 Kit for Meshtastic & LoRa" as I have several laying around and the platform should have more than enough space to handle this.

I do believe that having this as an option on hardware that can support the additional overhead would be a great benefit for the reasons OP mentions above. However, I do also agree that including this in base-implementation could be too bulky.

We have already run across the need for similar functionality between several of our disparate mesh networks and have successfully implemented WireGuard on other ESP-based projects that have higher overhead on the device. 

I will hopefully have some testing done over the next few days and will update with any findings.

TheWISPRer:
I have a rough implementation working, runs quite smoothly on my test hardware. Still uses static values for the WireGuard client config (set in the WireConfig.h file), so need to finish getting it fully integrated - but there is a fork available on my GitHub with a branch that contains the current working version

fifieldt:
@TheWISPRer , looks like it will integrate fairly cleanly, nicely done!

TheWISPRer:
[...]

PR #10626, opened 2026-06-04
Add configurable WireGuard client support

TheWISPRer:
## Summary

Adds WireGuard client support for ESP32-based devices that can be configured at runtime. Feature set added to firmware based on a flag during compile to prevent bulk when feature is not needed. (Addresses issue #6658)

This includes:
- WireGuard module config fields in the admin protobuf path.
- Firmware-side WireGuard config apply/status handling.
- Admin get/set support for `WIREGUARD_CONFIG`.
- WireGuard module config readback through the normal phone/API config stream.
- A Python configuration utility that can import a standard WireGuard `.conf` file.
- A Windows PlatformIO build path fix for this branch.
- (Separate) A standalone GUI/Windows helper path for configuring WireGuard outside official clients while client support catches up.

## Notes

During hardware validation, the WireGuard `enabled` field persisted while lower-numbered string fields did not. The root cause was the generated nanopb field layout for `WireGuardConfig`; fields are now ordered by protobuf tag so firmware decodes and persists the full message correctly.

The `status` and `last_error` fields are firmware-reported readback fields and should be treated as read-only by clients.

## Protobufs

This PR updates the `protobufs` submodule pointer for the WireGuard config field-order fix. The matching protobufs commit needs to be available upstream or through a visible protobufs PR before this firmware PR is ready to merge.

## Validation

Tested on Seeed XIAO ESP32S3:
- `platformio run -e seeed-xiao-s3`
- Serial flash using the generated firmware image
- Runtime WireGuard config import from a standard `.conf` file using the Python configurator
- Device readback confirmed persisted WireGuard fields and `RUNNING` status
- Connection resiliency against all discovered failure modes over the past year of testing 

No real WireGuard endpoint or key material is included in this PR description.

## Attestations

  - [x] Seeed XIAO ESP32S3


vidplace7:
Closing for now. This work needs to be rebased on `develop` branch (ESP32 family has gone through a large/breaking upgrade to Arduino 3.x in that branch). Please rebase / resubmit.

TheWISPRer:
The 2.8 / Arduino 3.x Wireguard branch is now functional on the XIAO S3, but it depends on narrow compatibility shims around WireGuard-ESP32 because that library still expects older ESP32 networking APIs. The stable 2.7.24-era WireGuard branch remains available separately. Before proposing this upstream, we should keep soaking hardware, document the shim risk clearly, and consider evaluating a native ESP-IDF WireGuard component as a future replacement

TheWISPRer:
For anyone following, I have started a discussion on this subject: https://github.com/meshtastic/firmware/discussions/10716

PR #10655, opened 2026-06-08
NRF52 LTO flags

thebentern:
Please test heavily, but this saves a ton of flash

### DIFF
--- a/extra_scripts/nrf52_lto.py
+++ b/extra_scripts/nrf52_lto.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+# trunk-ignore-all(ruff/F821)
+# trunk-ignore-all(flake8/F821)
+#
+# Whole-image LTO for nrf52840, EXCEPT the interrupt handlers.
+#
+# Every interrupt/exception handler is referenced only from the assembly vector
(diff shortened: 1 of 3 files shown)


Xaositek (approved):
Tested on a RAK4631 - while seeking for a GPS module, the whole thing certainly feels slower than previously noticed. Once it gives up around the 1 minute mark, performance improves significantly. Moving GPS to "Not Present" prevents this all together.

thebentern:
> Tested on a RAK4631 - while seeking for a GPS module, the whole thing certainly feels slower than previously noticed. Once it gives up around the 1 minute mark, performance improves significantly. Moving GPS to "Not Present" prevents this all together.

I noticed the GPS probing seems to be more sluggish lately, even prior to the LTO. Time to merge @HarukiToreda's GPS probe caching change ;-)

Xaositek:
T114 with LTO:
RAM:   [=====     ]  46.4% (used 115508 bytes from 248832 bytes)
Flash: [========  ]  84.3% (used 687112 bytes from 815104 bytes)

RAK4631 with LTO:
RAM:   [====      ]  44.0% (used 109468 bytes from 248832 bytes)
Flash: [========= ]  86.5% (used 705232 bytes from 815104 bytes)

PR #10656, opened 2026-06-08
UDP Unicast for Meshtasticd

jp-bennett:
Retreads our UDP multicast support for a unicast endpoint.

## Summary by CodeRabbit

* **New Features**
  * Added optional UDP unicast connectivity for Portduino devices.
  * Devices can send and receive encrypted mesh packets through a configured unicast peer.
  * Added configuration support for specifying the UDP unicast peer.
  * Made channel cryptography configuration available for broader integration.

* **Bug Fixes**
  * Improved decryption of encrypted packets received through UDP unicast connections.


Jorropo (commented):
1. please use dtls or some other authenticated encryption, don't be part of the problem of poorly secured iot devices, it's not too hard here.
Imo something like making the connection key: `1.2.3.4:4403@327d6a5dafc9ac5e5459f93225ee3996a66c9a4749e64e61c6ab4ff51df4104d` with raw Ed25519 or ECC keys.
Ping me if you want me to claude it up and review it.

2. It's kinda weird that you are implementing this only on meshtasticd.
An ESP32 is way good enough to also do this. I think it would make more sense if you implement the AsyncUDP unicast API in portduino, then write this code to use AsyncUDP, thus it would also work on ESP32s.

jp-bennett:
> 1. please use dtls or some other authenticated encryption,

How about AES-256 and AES-CCM based on x25519? *trollface*
Seriously, I question the usefulness of adding yet another encryption layer. We're pushing this data over the network with the same encryption as over the airwaves.



> 2\. It's kinda weird that you are implementing this only on meshtasticd.

Somehow I didn't think AsyncUDP had a unicast option. If we test this out and find it useful, moving it to that library makes sense, as a future PR.

Jorropo:
The point isn't to encrypt the data again,
it's that UDP is trivial to forge.

I want there to be some authentication that means someone random who finds an open meshtastic UDP port can't easily forge a couple packets to spam the mesh with garbage.
Imo using DTLS is the easiest way to achieve this but you could come up with something else.

jp-bennett:
The meshtasticd instance is the sender, and filters the UDP port for the connect address. While this isn't bulletproof security, it does mean that you can't just do a portscan for listening meshtasticd nodes.

Some authentication would eventually be handy. There's no published client yet, this is still a WIP. It's just a pain to juggle a bunch of patches to try to test things out.

PR #10658, opened 2026-06-09
Fix BLE advertising management to prevent crashes during reconnections

thebentern:
@vidplace7 give this a test run

### DIFF
--- a/src/nimble/NimbleBluetooth.cpp
+++ b/src/nimble/NimbleBluetooth.cpp
@@ -166,8 +166,8 @@ class BluetoothPhoneAPI : public PhoneAPI, public concurrency::OSThread
         // re-enter the GAP API while the host is mid-reset -- the crash that happens when a
         // phone reconnects with a stale bond to a freshly-wiped device.
         if (pendingStartAdvertising) {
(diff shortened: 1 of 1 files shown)


src/nimble/NimbleBluetooth.cpp
```diff
@@ -666,7 +700,15 @@ class NimbleBluetoothServerCallback : public BLEServerCallbacks
 
         nimbleBluetoothConnHandle = BLE_HS_CONN_HANDLE_NONE;
 
-        ble->startAdvertising();
```
cpatulea, line 670: Looks like Arduino-ESP32's native 'advertiseOnDisconnect' does make this call directly from the callback, like us: https://github.com/espressif/arduino-esp32/blob/0d1440d1be38ab530d274fe87ee88565fe167392/libraries/BLE/src/BLEServer.cpp#L588

@lucasssvaz in case he has any comment

cpatulea (approved):
Yowch, what a mess. Looks reasonable. Have not tested on a device.

thebentern:
Superceded by #10741

PR #10663, opened 2026-06-09
BaseUI: First attempt at Ham Mode implementation

Xaositek:
First attempt at Ham Mode implementation

- When choosing an ITU Region, an additional prompt will be displayed to confirm "I confirm I am a licensed amateur radio operator"
- If in "is_licensed" mode, an additional prompt will be displayed is choosing a normal mode.

### DIFF
--- a/src/graphics/draw/MenuHandler.cpp
+++ b/src/graphics/draw/MenuHandler.cpp
@@ -183,15 +183,11 @@ static void applyLoraRegion(meshtastic_Config_LoRaConfig_RegionCode region, bool
     if (isHam && adminModule) {
         meshtastic_HamParameters hamParams = meshtastic_HamParameters_init_zero;
         strncpy(hamParams.call_sign, "N0CALL", sizeof(hamParams.call_sign) - 1);
-        strncpy(hamParams.short_name, "N0CL", sizeof(hamParams.short_name) - 1);
+        strncpy(hamParams.short_name, "N0CL", sizeof(hamParams.short_name));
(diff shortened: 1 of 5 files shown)


src/graphics/draw/MenuHandler.cpp
```diff
@@ -232,6 +267,16 @@ void menuHandler::LoraRegionPicker(uint32_t duration)
...
 
+            bool hamMode = false;
+
+            if (selectedRegion == meshtastic_Config_LoRaConfig_RegionCode_ITU1_2M ||
+                selectedRegion == meshtastic_Config_LoRaConfig_RegionCode_ITU2_2M ||
+                selectedRegion == meshtastic_Config_LoRaConfig_RegionCode_ITU3_2M ||
+                selectedRegion == meshtastic_Config_LoRaConfig_RegionCode_ITU2_125CM) {
+                LOG_INFO("User chose an amateur radio mode region");
+                hamMode = true;
+            }
```
vidplace7: You should be able to use
```
selectedRegion->profile->licensedOnly
```
instead of creating a static list here.

Xaositek: This has been addressed.

vidplace7 (approved):
Tested this locally on T-Beam Supreme. Happy with it 👍

vidplace7:
Going ahead and shipping this since it's so helpful in troubleshooting ham-only regions (clients can't set them yet and most clients dont support setHamMode properly).

PR #10666, opened 2026-06-10
Short circuit invalid nodenum check in ham mode.

vidplace7:
Currently when ham_mode is enabled, the nodenum is declared invalid and regenerated upon every reboot because `isOurOwnEntry` uses the publicKey (which doesn't exist in ham mode).

This PR short circuits the invalid nodenum check when ham mode is enabled. Until signing #10478 is implemented, this will prevent the nodenum from being regenerated upon each reboot.

## 🤝 Attestations

  - [x] `tbeam-s3-core` 144mhz

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -1542,6 +1542,8 @@ void NodeDB::pickNewNodeNum()
 
     // Identity check via public key (or "empty slot?" when no keys yet);
     // macaddr no longer lives on the slim header.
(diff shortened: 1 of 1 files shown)


Xaositek (approved):

vidplace7:
This is still needed even after the fix in 309d51a3e8c01aa3ffb2a57280cbf6819c990776 @thebentern

caveman99:
@vidplace7 #10478 is in firmware now - is this still needed or does it need to be updated to cater for signing?

PR #10700, opened 2026-06-12
Stm32 general

jp-bennett:
This adds a general ARCH_STM32, and splits some of the ifdefs out to use this define rather than the more specific, ARCH_STM32WL
Also includes a trunk run

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -48,7 +48,7 @@
 #include "concurrency/LockGuard.h"
 #endif
 
-#if defined(ARCH_STM32WL) && defined(BATTERY_PIN)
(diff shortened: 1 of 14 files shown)


src/Power.cpp
```diff
@@ -864,7 +864,7 @@ void Power::shutdown()
 #endif
     }
 #endif
-#if !defined(ARCH_STM32WL)
+#if !defined(ARCH_STM32)
```
thebentern: Do we really need to exclude all STM32 from beeps?

jp-bennett: Yeah, that could conceivably go the other way on a more performant MCU.

PR #10724, opened 2026-06-15
Use upstream fusion library

jp-bennett:
We had previously just copied our sensor fusion implementation from an upstream library. I've opened a PR to that library, adding a library.properties and library.json, and pointed to that PR hash in our platformio.ini.

This includes a couple minor changes to compile against the latest version of that library.

Assuming the upstream PR is accepted, we can pin to the right version and add renovate support when they make their next release.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated compass heading handling on supported sensors for more consistent orientation results.
  * Improved heading calculations when the display compass orientation is rotated past 270°.

* **Chores**
  * Removed several unused legacy components from the bundled motion library.


jp-bennett:
No movement on the upstream PR, so for now fork it and use the Meshtastic repo.

caveman99 (approved):

Issue #10687, opened 2026-06-10
[Feature Request]: Move portable nodes to a later contention window


luivicur:
### Platform

Cross-Platform

### Description

Currently portable nodes share the same CW as a client_base, this interacts with the managed flooding algorithm preferring weaker connections and results in portable nodes having priority.

Portable nodes tend to be at ground level, surrounded by obstacles (often the very user), and seldom have antennas of more than 2.5dBi (i.e. dipole or whip). As a result they make poor relays.
Client_base nodes are commonly located in more advantageous positions like roofs or short masts. On average client_base nodes are better suited to forwarding traffic than mobile nodes. 
In my own experience my mobile node has a 5dB SNR disadvantage compared to my roof node with comparable antenna, but the roof node is hands down the better relay.

Given the physical limitations, re-transmissions from portable nodes are a last ditch effort to extend range. Ideally these nodes should skip relaying while covered by infrastructure nodes, and resume relaying in sparse coverage.  The existence of the client_mute role is a recognition of the negative effect clients can have on a dense network.

I propose to delay all mobile roles re-transmissions (everything except router, router_late, and client_base) to allow infrastructure nodes to propagate first. The result should be:
-More cancelled re-transmissions in mobile nodes when in dense coverage
-Slight delay for re-transmissions when working in ad-hoc meshes.

While client_mute would still be the ideal role in large meshes, this  change would mean a lesser impact from new users since client is the default and commonly recommended role.
For users frequently moving between dense and sparse regions it would allow a single role to work well enough.

NomDeTom:
@korbinianbauer any chance you can put something into your jupyter notebook simulators and see what the effect of pulling client_base forward in the window (method TBD) would be? I imagine if they were considered to be nodes with X dB more signal than those adjacent, that might help model the "better placed" aspect.

shalberd:
"Currently portable nodes share the same CW contention window as a client_base"

Not true / I do not think that is the case. If you mean the role Client, that role is in the medium rebroadcast window.

Client_base, for nodes that is has favorited, is using the late rebroadcast window. 

We have a mesh here with 600 nodes or more and have long switched to recommending role Client only for really well-placed nodes that benefit the entire mesh and, yes, making Client_Mute the new recommended default for all portable nodes, as they do not contribute anything in a dense mesh.

"While client_mute would still be the ideal role in large meshes, this change would mean a lesser impact from new users since client is the default and commonly recommended role."

[...]

PR #10725, opened 2026-06-15
feat(mesh): Shift mobile roles contention window to lower priority (10687)

luivicur:
This is the proposed code for issue [10687](https://github.com/meshtastic/firmware/issues/10687).

It shifts all non infrastructure role's contention window to a later frame. This should allow infrastructure to have priority and reduce client retransmissions in dense meshes. 

I tried a couple of other approaches but settled on this. It provides a predictable behavior with no (theoretical) impact on the existing retry logic.

- [X] Heltec (Lora32) V3
    Muzi R1 Neo    

Using the serial debug output I was able to confirm that the expected wait times were upheld even at extreme SNRs. With an SDR I was able to confirm that retransmissions still happen. Finally I have been using the modified code on my daily driven R1 Neo with no obvious problems.

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -707,9 +707,18 @@ uint32_t RadioInterface::getTxDelayMsecWeighted(meshtastic_MeshPacket *p)
     if (shouldRebroadcastEarlyLikeRouter(p)) {
         delay = random(0, 2 * CWsize) * slotTimeMsec;
         LOG_DEBUG("rx_snr found in packet. Router: setting tx delay:%d", delay);
+    } else if (config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT_BASE) {
+        delay = (2 * CWmax * slotTimeMsec) + random(0, pow_of_2(CWsize)) * slotTimeMsec;
+        LOG_DEBUG("rx_snr found in packet. Client_base: setting tx delay:%d", delay);
     } else {
         // offset the maximum delay for routers: (2 * CWmax * slotTimeMsec)
-        delay = (2 * CWmax * slotTimeMsec) + random(0, pow_of_2(CWsize)) * slotTimeMsec;
+        // plus another 8*CWmax to cede priority to client_base
+        uint8_t routerWindow = 2 * CWmax;
+        delay = (8 * CWmax) + random(0, pow_of_2(CWsize));
+        delay = delay > pow_of_2(CWmax) ? pow_of_2(CWmax) : delay;
+        delay += routerWindow;
+
+        delay *= slotTimeMsec;
         LOG_DEBUG("rx_snr found in packet. Setting tx delay:%d", delay);
     }
 


src/mesh/RadioInterface.cpp
```diff
@@ -705,9 +707,18 @@ uint32_t RadioInterface::getTxDelayMsecWeighted(meshtastic_MeshPacket *p)
...
+    } else if (config.device.role == meshtastic_Config_DeviceConfig_Role_CLIENT_BASE) {
+        delay = (2 * CWmax * slotTimeMsec) + random(0, pow_of_2(CWsize)) * slotTimeMsec;
+        LOG_DEBUG("rx_snr found in packet. Client_base: setting tx delay:%d", delay);
     } else {
         // offset the maximum delay for routers: (2 * CWmax * slotTimeMsec)
-        delay = (2 * CWmax * slotTimeMsec) + random(0, pow_of_2(CWsize)) * slotTimeMsec;
+        // plus another 8*CWmax to cede priority to client_base
+        uint8_t routerWindow = 2 * CWmax;
+        delay = (8 * CWmax) + random(0, pow_of_2(CWsize));
+        delay = delay > pow_of_2(CWmax) ? pow_of_2(CWmax) : delay;
```
luivicur, line 718: `pow_of_2(CWmax)` is calculated twice. The compiler should optimize this away but it does look fugly. Suggestions are welcome.

GUVWAF, line 718: Capping it to `pow_of_2(CWmax)` will lead to several random values all leading to the same delay, which would result into a high chance of collision.

GUVWAF (commented):
This has a large impact on messaging in general. It means *all* messages transmitted by mobile roles (which is the default!) will be delayed (even more), only for `CLIENT_BASE` nodes to get priority. Given that `CLIENT_BASE` is a very specific role, I don't think this is worth it. It might simply be you need to use another role then. 

Alternatively, `CLIENT_BASE` could get a slightly less delay than mobile nodes, e.g. `delay = (1 * CWmax) + random(0, pow_of_2(CWsize))`, so it doesn't impact the delay of mobile nodes.

src/mesh/RadioInterface.cpp
luivicur: The cap only comes into play when SNR is +10dB, which is pretty much never*. Even then the random will essentially be from 64 to 256. The original code uses 0 to 256. In practice the current code will most commonly use the 0 to 64 range and often smaller.  
At lower SNRs the range used in the random remains the same. It is simply shifted and so the same probability of collision applies.

*: The sx1262 using LongFast settings appears to saturate around +6dB SNR. It never reads higher. Still, even at +9dB,  2^CWSize drops to 128, leaving the 128-256 range empty.

luivicur:
> This has a large impact on messaging in general. It means _all_ messages transmitted by mobile roles (which is the default!) will be delayed (even more), only for `CLIENT_BASE` nodes to get priority. Given that `CLIENT_BASE` is a very specific role, I don't think this is worth it. It might simply be you need to use another role then.
> 
The conversation goes a little deeper into  the why's: https://github.com/meshtastic/firmware/issues/10687
The proposed delay is 1 second in long fast. I consider the benefits worth the delay.
Client_base should be the most common infrastructure role in a large mesh.


> Alternatively, `CLIENT_BASE` could get a slightly less delay than mobile nodes, e.g. `delay = (1 * CWmax) + random(0, pow_of_2(CWsize))`, so it doesn't impact the delay of mobile nodes.

This does impact the probability of collision with router nodes.

caveman99:
Don't assume everyone runs a large infrastructure mesh. Or change timing values without proper simulation. This impacts use cases (popular ones) that are not covered by 'large infrastructure mesh'.

luivicur:
Understood and taken into account. I don't believe an extra 1s delay (in longfast) is a deal breaker. The impact is only in rebroadcasts, direct connections are not affected. The proposal allows tuning by lowering the wait time.

I played around with an artificial play on the SNR but this provided the most consistent results.

Other ideas on how to make Client better behaved under both infrastructure and ad-hoc are welcome in the issue https://github.com/meshtastic/firmware/issues/10687

luivicur:
Going over my numbers again, using 64 slotTimes pretty much guaranteed that clients would transmit after client_base in all scenarios. 
Reducing it to 32 slotTimes (0.5s in LongFast) still clears most cases while significantly reducing the impact.

It looks to me like a good compromise to make client a more flexible role.

PR #10730, opened 2026-06-16
Fix PKC on portduino sim by working around region blocks keygen and packet length

thebentern:
This pull request improves the handling of PKI-encrypted direct messages (DMs) and simulator behavior in the Meshtastic codebase. The changes ensure that encrypted messages are properly carried and delivered between simulator nodes, and that key generation works as expected in simulation environments. Additionally, the handling of multicast UDP packets is clarified to preserve essential fields for correct decryption.

**Simulator PKI DM Handling:**

* In `SimRadio::startSend`, packets that are PKI-encrypted (such as DMs) and cannot be decrypted by the sender are now carried as ciphertext, allowing the receiving simulator node to decrypt them as if they were received over the air. This prevents spurious "no suitable channel" logs and avoids payload overflows.
* In `SimRadio::unpackAndReceive`, when a received packet is marked as carrying ciphertext (via a sentinel portnum), the code restores the encrypted payload so the router can attempt decryption, maintaining correct end-to-end encryption semantics in simulation.
* Ensured that all reads of the encrypted payload are complete before re-encoding the packet, due to union aliasing in the protobuf structure.

**Simulator Key Generation:**

* Modified `NodeDB::generateCryptoKeyPair` to allow keypair generation in the simulator environment even if the LoRa region is unset, ensuring that PKI-encrypted DMs work between sim nodes. This is done by checking for the simulator radio module and overriding the region restriction.

**UDP Multicast Handling:**

* Added a comment in `UdpMulticastHandler` to clarify that the full `MeshPacket` is preserved (including channel and PKI-encrypted fields) so that the receiver can attempt PKI decryption, rather than falling back to channel-based decoding.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -3047,8 +3047,15 @@ bool NodeDB::checkLowEntropyPublicKey(const meshtastic_Config_SecurityConfig_pub
 bool NodeDB::generateCryptoKeyPair(const uint8_t *privateKey)
 {
 #if !(MESHTASTIC_EXCLUDE_PKI_KEYGEN || MESHTASTIC_EXCLUDE_PKI)
-    // Only generate keys for non-licensed users and if LoRa region is set
-    if (owner.is_licensed || config.lora.region == meshtastic_Config_LoRaConfig_RegionCode_UNSET) {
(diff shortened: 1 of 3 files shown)


thebentern:
@copilot trunk fmt

PR #10752, opened 2026-06-20
Reset noise floor samples after frequency changes

RCGV1:
## Summary
- Reset RadioLib noise-floor calibration when `reconfigure()` applies a different effective receiver frequency.
- Keep the change local to `RadioLibInterface`; all RadioLib chip drivers already call this base reconfigure path before applying chip-specific settings.
- Clear the noise-floor sample throttle timestamp during reset so collection can restart immediately on the new frequency.
- Add `test_radio` coverage that seeds RadioLib noise-floor state, verifies it is cleared when frequency changes, and verifies it is preserved when frequency is unchanged.

## Why this is needed
LoRa config changes do not generally reboot the node. `AdminModule::handleSetConfig()` explicitly sets `requiresReboot = false` for LoRa config, then `saveChanges()` calls `MeshService::reloadConfig()`, which notifies the radio config observer and reconfigures the radio live. Channel changes also save without reboot and flow through reloadConfig; this matters because the default frequency slot can be derived from the primary channel name.

## Testing
- `trunk fmt src/mesh/RadioInterface.h src/mesh/RadioInterface.cpp src/mesh/RadioLibInterface.h src/mesh/RadioLibInterface.cpp test/test_radio/test_main.cpp`
- `~/.platformio/penv/bin/python -m platformio test -e native-macos -f test_radio`: 17 passed
- `pio run -e heltec-v3`: SUCCESS

## Notes
- Attempted `~/.platformio/penv/bin/python -m platformio test -e native -f test_radio` on macOS, but it failed before test execution because LovyanGFX includes Linux-only `<malloc.h>`. The repo has `native-macos` for this case, and that target passed.

### DIFF
--- a/src/mesh/RadioInterface.h
+++ b/src/mesh/RadioInterface.h
@@ -240,8 +240,8 @@ class RadioInterface
   protected:
     int8_t power = 17; // Set by applyModemConfig()
 
-    float savedFreq;
-    uint32_t savedChannelNum;
+    float savedFreq = 0.0f;
+    uint32_t savedChannelNum = 0;
 
     /***
      * given a packet set sendingPacket and decode the protobufs into radiobuf.  Returns # of bytes to send (including the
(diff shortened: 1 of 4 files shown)


thebentern:
This probably should target master

thebentern:
Now that it targets master, it needes to be rebased. We don't want to merge develop into master yet

PR #10754, opened 2026-06-20
fix(phone-api): skip manifest scan for node-info-only config requests

jeremiah-k:
## Overview

This change fixes a firmware crash I hit while testing the Android WiFi/TCP handshake recovery work that landed in meshtastic/Meshtastic-Android#5856.

During Android Stage 2, the app sends a node-info-only config request. On my tbeam running `v2.7.25.104df5f`, the firmware logged that the client only wanted node info, then crashed/rebooted before completing the node-info response.

The main fix is to skip filesystem manifest loading entirely for the `SPECIAL_NONCE_ONLY_NODES` path. That path does not use the file manifest, so avoiding the SPI lock and filesystem walk removes unnecessary memory pressure during node-info-only config starts.

This also hardens filesystem manifest enumeration by making it bounded and less allocation-heavy.

## Key Changes

**PhoneAPI**
- Skips filesystem manifest loading for `SPECIAL_NONCE_ONLY_NODES`.
- Releases any existing manifest data when starting the node-info-only path.
- Keeps the node-info-only flow intact: send node info, then send config complete.
- Uses explicit manifest scan limits for normal config requests:
  - depth: `3`
  - max entries: `64`
- Logs when manifest enumeration was limited.

**Filesystem Enumeration**
- Extends `getFiles()` with optional `maxCount` and `wasLimited` parameters.
- Defaults `getFiles()` to at most 64 entries.
- Adds bounded recursive traversal so enumeration stops once enough entries are collected.
- Uses one accumulator vector instead of building temporary vectors for recursive calls.
- Uses `strlcpy` instead of `strcpy` for file names.
- Avoids temporary `String` allocation for dot-suffix filtering.
- Compile-gates the `std::bad_alloc` / `std::length_error` reserve fallback so no-exception targets still build.

## Validation

Firmware built from `86f062657c95a988492d35979e972ed8d3a4a350` has been running well for about two days across multiple tbeam, rak4631, and T1000-E devices.

With that firmware, Android can connect without triggering the crash/reboot loop I was seeing before.

The latest branch includes a few additional review-suggested filesystem traversal hardening changes, which I am loading up for more testing now.

This was tested together with the latest Android-side WiFi/TCP recovery changes in meshtastic/Meshtastic-Android#5856.

## Review Notes

This patch deserves careful maintainer review. I have done the best I can to review the C/C++ side with AI tooling and device testing, but I do not have much experience with C yet.

The main behavior change to call out is that `getFiles()` now limits results to 64 entries by default. Existing callers still compile because the new parameters have defaults, but callers that need a larger or complete listing should pass a larger `maxCount` and can use `wasLimited` to detect truncation.

## Breaking Changes / Migration

`getFiles()` now limits results by default to at most 64 entries.

Existing callers continue to compile due to default arguments, but they may now receive a truncated listing. Callers that require more entries should pass a larger `maxCount` and optionally check `wasLimited`.

## Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [x] LilyGo T-Beam
  - [x] RAK WisBlock 4631
  - [ ] Other

### DIFF
--- a/src/FSCommon.cpp
+++ b/src/FSCommon.cpp
@@ -101,6 +101,7 @@ bool renameFile(const char *pathFrom, const char *pathTo)
 
 #include <cstring>
 #include <new>
+#include <stdexcept>
 #include <vector>
 
 #ifdef FSCom
@@ -108,13 +109,36 @@ namespace
 {
 bool pathEndsWithDot(const char *path)
 {
+    if (!path)
+        return false;
+
     size_t length = strlen(path);
     return length > 0 && path[length - 1] == '.';
+}
+
+bool copyFilePath(char *dest, size_t destSize, const char *path, bool *wasLimited)
+{
+    if (!path || destSize == 0) {
+        if (wasLimited)
+            *wasLimited = true;
+        return false;
+    }
+
+    if (strlcpy(dest, path, destSize) >= destSize) {
+        if (wasLimited)
+            *wasLimited = true;
+        return false;
+    }
+
+    return true;
 }
 
 void collectFiles(const char *dirname, uint8_t levels, size_t maxCount, std::vector<meshtastic_FileInfo> &filenames,
                   bool *wasLimited)
 {
+    if (!dirname)
+        return;
+
     File root = FSCom.open(dirname, FILE_O_READ);
     if (!root)
         return;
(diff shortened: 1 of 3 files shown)


thebentern:

Issue #10746, opened 2026-06-19
NodeDB warmstore (#10705) causes hard reboot on RP2350/W5500 at full-DB nodes.proto save


cvaldess:
### Category
Bug report

### Hardware
RP2350 (Raspberry Pi Pico 2) + WIZnet W5500 Ethernet + Ebyte E22 (SX1262).
DIY variant `wiznet_5500_evb_pico2_e22p`. USB-powered, **no battery connected**.

### Firmware version
`develop` after #10705 ("Pr1 nodedb warmstore", `79a7dcc46`, merged 2026-06-18).

### Description
On RP2350/W5500, the device hard-reboots (no panic, no "Rebooting" log — abrupt output cutoff) **exactly** at `Save /prefs/nodes.proto`, once the node DB fills up and an eviction fires. On a busy mesh the DB reaches capacity in ~3 minutes, so the board enters a reboot loop and never stays online.

Bisected to #10705: a build from the immediately-preceding commit (pre-warmstore) runs the identical save path with a full DB and survives indefinitely (15+ min, many full-DB saves). The only functional difference is the warmstore feature.

### Steps to reproduce
1. Flash `develop` (incl. #10705) onto an RP2350 + W5500 board.
2. Place it on a populated mesh so the node DB fills to capacity.
3. Watch the serial log: it reboots at the first `Save /prefs/nodes.proto` after `Node database full ... Erasing oldest entry`.

### Expected
Saving a full NodeDB completes without resetting the device.

### Actual (failing build, #10705 — reboots at uptime ~193 s)
```
INFO  | [Router] Node database full with 120 nodes and 332168 bytes free. Erasing oldest entry
INFO  | [Router] Adding node to database with 120 nodes and 332216 bytes free!
DEBUG | [Router] Save to disk 16
DEBUG | [Router] Opening /prefs/nodes.proto, fullAtomic=0
INFO  | [Router] Save /prefs/nodes.proto
<<< abrupt cutoff / reboot — no further output >>>
```

### Control (pre-warmstore build — survives, 15+ min uptime)
```
INFO  | 01:15:17 918 [Router] Node database full with 100 nodes ... Erasing oldest entry
INFO  | 01:15:17 918 [Router] Adding node to database with 100 nodes ...
DEBUG | 01:15:17 918 [Router] Module 'routing' wantsPacket=1
... txGood=16, txRelay=13, rxGood=43, rxBad=0    (continues normally)
```
[...]

PR #10759, opened 2026-06-22
fix: right-size warm tier for constrained platforms (#10746, #10705)

caveman99:
Constrained platforms fall into the generic `WARM_NODE_COUNT 320` catch-all, which is too large for them.

- **RP2040/RP2350 (#10746):** `nodes.proto` and `warm.dat` are written back-to-back in one blocking stretch, so the loop never feeds the 8s HW watchdog and it resets on the second write. Fix: `watchdog_update()` between the two writes, plus an explicit `ARCH_RP2040` branch (150).
- **No-PSRAM ESP32-S3 (#10705 BLE reboots):** `WARM_NODE_COUNT` 320→150 and `TRAFFIC_MANAGEMENT_CACHE_SIZE` 1000→500 (~12 KB heap recovered).

Warm tier stays enabled on all platforms; generic ESP32 unchanged.

### DIFF
--- a/src/mesh/mesh-pb-constants.h
+++ b/src/mesh/mesh-pb-constants.h
@@ -128,13 +128,15 @@ static inline int get_max_num_nodes()
 // chain. Backed by the raw-flash ring below LittleFS — see WarmNodeStore.h.
 #define WARM_NODE_COUNT 200
 #elif (defined(CONFIG_IDF_TARGET_ESP32S3) && defined(BOARD_HAS_PSRAM)) || defined(ARCH_PORTDUINO)
-#define WARM_NODE_COUNT 2000 // PSRAM-equipped ESP32-S3 / native host; warm.dat ~80 KB
-#elif defined(CONFIG_IDF_TARGET_ESP32S3)
-#define WARM_NODE_COUNT 150 // ESP32-S3 without PSRAM: ~6 KB heap, keeps BLE headroom (#10705)
+#define WARM_NODE_COUNT 2000 // PSRAM-equipped ESP32-S3 / native host; warm cache in PSRAM (~80 KB)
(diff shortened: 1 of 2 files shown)


src/mesh/mesh-pb-constants.h
```diff
@@ -164,8 +170,12 @@ static inline int get_max_num_nodes()
 #define TRAFFIC_MANAGEMENT_CACHE_SIZE 0
 #elif (defined(CONFIG_IDF_TARGET_ESP32S3) && defined(BOARD_HAS_PSRAM)) || defined(ARCH_PORTDUINO)
 #define TRAFFIC_MANAGEMENT_CACHE_SIZE 2048 // PSRAM-equipped ESP32-S3 / native host
+#elif defined(CONFIG_IDF_TARGET_ESP32S3) || defined(CONFIG_IDF_TARGET_ESP32C6) || defined(CONFIG_IDF_TARGET_ESP32P4)
+#define TRAFFIC_MANAGEMENT_CACHE_SIZE 500 // 512 KB+ SRAM, no PSRAM (S3/C6/P4): ~5 KB heap (#10705)
+#elif defined(ARCH_ESP32)
+#define TRAFFIC_MANAGEMENT_CACHE_SIZE 400 // classic ESP32 / S2 / C3: tightest free heap, ~4 KB (#10705)
 #else
-#define TRAFFIC_MANAGEMENT_CACHE_SIZE 1000 // Generic ESP32, ESP32-S3 without PSRAM
+#define TRAFFIC_MANAGEMENT_CACHE_SIZE 1000 // RP2040 / other
```
NomDeTom: Mention NRF52?

NomDeTom:
Overall it seems eminently sensible. I've lost track of all the different esp32 flavours that are supported.

caveman99:
The real fix is to trigger the watchdog in between saves. BUT when researching this i stumbled across other crashes that really hit heap exhaustion. I think those values are sensible unless told otherwise :-)

thebentern (approved):

PR #10766, opened 2026-06-23
 Fix for explicit vs. implicit channel name hashing

### LINKED ISSUE #10764: [Bug]: Apparently identical channel names can hash differently
### Category

Other

### Hardware

DIY

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [ ] BaseUI

### Firmware Version

2.7.25

### Description

When a Meshtastic channel's name field is stored as an explicit string (e.g. "LongFast"), the channel hash is computed from that literal string. When the name field is left empty, the firmware substitutes the modem preset's display name for human-readable purposes, but the hash that is actually computed and stamped into outgoing packets — and used to gate incoming packet decryption — derives from the empty string directly. Because XOR-hashing an empty string yields a different result than hashing "LongFast", the two nodes end up with different channel hashes even though they are on the same modem preset with the same PSK.

The consequence is that packets sent by a node with an explicit channel name will carry a hash that the blank-named node's channel does not recognise, and vice versa. The router's hash gate rejects the mismatch immediately and no decryption is attempted. There is no log warning on the receiving side beyond the generic "no suitable channel found" message, so to a user the network simply appears silent. Two nodes that appear to be on the same channel — same preset, same PSK — are in practice partitioned from each other by a naming convention that the UI never surfaces.

It is also noteworthy that the channel name and the modem preset are stored and mutated independently, with no coupling between them. If a node's channel name has been explicitly set and the modem preset is subsequently changed, the stored name persists unchanged in flash. The node will then be transmitting on the new preset's radio parameters while still hashing as the old name — silently incompatible with both the nodes that follow the new preset's implicit name and those that followed the old one. A blank-named node that changes preset will correctly re-derive its hash from the new preset name, but an explicitly named node will not, and there is no mechanism in the firmware or the client apps to detect or warn about this divergence.

### Relevant log output

```Shell
  10:24:11 182s INF [ChanDump] [ChanDump] slot=0 role=PRIMARY stored_name='' resolved_name='LongFast' hash=0x08
  10:24:41 212s INF [ChanDump] [ChanDump] slot=0 role=PRIMARY stored_name='LongFast' resolved_name='LongFast' hash=0x0a
```

NomDeTom:
fixes #10764 and replaces #10765

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
NRF promicro DIY (my beloved)


NomDeTom:
@ianmcorvidae is correct - and I think this is garbage.

PR #10776, opened 2026-06-24
Update native to exit when configured with an unknown module

jessm33:
Update native to exit when configured with an unknown module instead of going in to sim mode.  If a lora module is set in config.yaml and that module isn't one recognized by meshtasticd, meshtasticd will now exit with an error instead of going in to sim mode. This will prevent confusion by not going in to sim mode when a radio is explicitly confined by the user.  Sim mode can still be entered by command line option, or by setting the module to sim.

## 🤝 Attestations

    - [X] Linux native, code change only effects code used in native.


jp-bennett (approved):
Haven't tested yet, but I approve of the concept.

thebentern (approved):

PR #10786, opened 2026-06-25
BaseUI changes to support t-watch-s3 rounded screen

jp-bennett:
Not pixel perfect yet, but a significant improvement.


src/graphics/draw/UIRenderer.cpp
```diff
@@ -1814,17 +1831,22 @@ void UIRenderer::drawNavigationBar(OLEDDisplay *display, OLEDDisplayUiState *sta
         lastFrameIndex = frameToHighlight;
         lastFrameChangeTime = millis();
     }
-
+#ifdef OLED_HUGE
+    const int iconSize = 24;
+#else
```
Xaositek, line 1836: Can we add a new ScreenResolution to cover this?

jp-bennett: We could, though it'll take quite a few changes to update every time we check for ScreenResolution::High. Also worth mentioning that doing it in a define tends to save flash space vs doing it in an if() block.

jp-bennett:
Going to go ahead and merge this into the t-watch PR

PR #10802, opened 2026-06-28
Refactor SerialConsole to simplify timeout handling

jp-bennett:
Removed conditional compilation for timeout check on supported architectures.

Just for testing for now. Start of effort to kill the ugly #if defined() chains.

## Summary by CodeRabbit

* **Bug Fixes**
  * Serial connection startup now waits for the port to become available across all builds, helping avoid early startup failures and improving consistency.
  * Minor formatting adjustment was made with no user-facing behavior change.


jp-bennett:
I find the firmware size changes a bit hard to believe.

PR #10809, opened 2026-06-28
Fix `spiLock` deadlocks / frequent watchdog restarts in `WarmNodeStore::save()`

compumike:
Fixes frequent watchdog restarts on platforms with `FSCom`.

I was seeing frequent (every few minutes to tens of minutes) hangs and then watchdog reboots. I tracked it down to `WarmNodeStore::save()` which holds `spiLock` while calling `SafeFile f(filename.c_str(), false);`. The constructor for `SafeFile` itself tries to grab `spiLock`, which is already held, so this causes a deadlock, until the watchdog timer kicks in and reboots.

This PR makes it so `WarmNodeStore::save()` adds locking **only** around the calls that aren't already covered by an `spiLock`: the `mkdir` and the `write` calls.

(Also found and fixed similar in `MessageStore::clearAllMessages`, but that's not very commonly used.)

It appears that this logic was introduced about two weeks ago in https://github.com/meshtastic/firmware/pull/10705/changes#diff-fd0d79c004a2b4611b3eee806f16eef5d803bebd2e1a13f7b891e6b4bc312b2dR523-R534 @NomDeTom @thebentern 

My logs looked like:

```
DEBUG | 16:07:05 264 [Router] Saved Pubkey:  ...
DEBUG | 16:07:05 264 [Router] Update changed=1 user Redacted/RDCT, id=0x12345678, channel=1
DEBUG | 16:07:06 264 [Router] Node status update: 78 online, 250 total
DEBUG | 16:07:06 264 [Router] Save to disk 16
DEBUG | 16:07:06 264 [Router] Opening /prefs/nodes.proto, fullAtomic=0
INFO  | 16:07:06 264 [Router] Save /prefs/nodes.proto
E (355189) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:
E (355189) task_wdt:  - loopTask (CPU 1)
E (355190) task_wdt: Tasks currently running:
E (355190) task_wdt: CPU 0: IDLE0
E (355190) task_wdt: CPU 1: IDLE1
E (355190) task_wdt: Aborting.
E (355210) task_wdt: Print CPU 1 (current core) backtrace
...
```

## 🤝 Attestations

Testing is in progress; would appreciate other testers!

### DIFF
--- a/src/MessageStore.cpp
+++ b/src/MessageStore.cpp
@@ -354,10 +354,16 @@ void MessageStore::clearAllMessages()
     resetMessagePool();
 
 #ifdef FSCom
-    concurrency::LockGuard guard(spiLock);
     SafeFile f(filename.c_str(), false);
     uint8_t count = 0;
-    f.write(&count, 1); // write "0 messages"
+
+    // SafeFile already does its own spiLock in its constructor and close().
+    // Avoid nesting spiLocks, as this will hang until watchdog reset!
+    {
+        concurrency::LockGuard guard(spiLock);
+        f.write(&count, 1); // write "0 messages"
+    }
+
     f.close();
 #endif
 
--- a/src/mesh/WarmNodeStore.cpp
+++ b/src/mesh/WarmNodeStore.cpp
@@ -589,12 +589,22 @@ bool WarmNodeStore::save()
     h.entrySize = sizeof(WarmNodeEntry);
     h.crc = crc32Buffer(packed.data(), h.count * sizeof(WarmNodeEntry));
 
-    concurrency::LockGuard g(spiLock);
-    FSCom.mkdir("/prefs");
+    // SafeFile already does its own spiLock in its constructor and close().
+    // Avoid nesting spiLocks, as this will hang until watchdog reset!
+
+    {
+        concurrency::LockGuard g(spiLock);
+        FSCom.mkdir("/prefs");
+    }
 
     auto f = SafeFile(warmFileName, false);
-    f.write((const uint8_t *)&h, sizeof(h));
-    f.write((const uint8_t *)packed.data(), h.count * sizeof(WarmNodeEntry));
+
+    {
+        concurrency::LockGuard g(spiLock);
+        f.write((const uint8_t *)&h, sizeof(h));
+        f.write((const uint8_t *)packed.data(), h.count * sizeof(WarmNodeEntry));
+    }
+
     bool ok = f.close();
     if (!ok)
         LOG_ERROR("WarmStore: can't write %s", warmFileName);


cvaldess:
Confirmed on hardware — this fixes the RP2350/W5500 reboot-loop tracked in #10746.

Your `spiLock` deadlock diagnosis matches every symptom we saw there. `WarmNodeStore::save()` takes `LockGuard g(spiLock)` and then constructs `SafeFile(warmFileName, false)`, whose constructor calls `openFile()` which grabs `spiLock` again — non-recursive, so it hangs until the watchdog fires. That explains why our captures cut off right after `Save /prefs/nodes.proto` with no `warm.dat` line (the deadlock happens at the `LockGuard` *before* the `Opening …` `LOG_DEBUG`), why the post-reboot filesystem had `nodes.proto` but never `warm.dat`, and why `WARM_NODE_COUNT=0` avoided it while #10759's `watchdog_update()` (=150) did not.

### Test setup
- Board: WIZnet W5500-EVB-Pico2 (RP2350, 2 MB) — the exact board from #10746
- `develop` + this PR's diff, **warm store active** (override removed, so it takes the `ARCH_RP2040` branch → `WARM_NODE_COUNT = 150`)
- Forced the failing path: hot store full at 120, eviction triggered by an incoming NodeInfo (portnum 4)

### Result — survives, with `warm.dat` written
```
[Router] Node database full with 120 nodes ... Erasing oldest entry
[Router] Save /prefs/nodes.proto
[Router] Opening /prefs/warm.dat, fullAtomic=0
[Router] WarmStore: saved 8 warm nodes to /prefs/warm.dat
```

The `Opening /prefs/warm.dat` debug line — which never appeared in the broken build because the deadlock hit the `LockGuard` ahead of it — now shows, the warm write completes, and the board rides through repeated full-DB evictions with `rebootCount = 0`, no watchdog resets, and monotonic uptime. Previously this same sequence reboot-looped every ~60–100 s.

Thanks for tracking this down — happy to keep it on a longer soak if useful.

compumike:
Works for me here -- now getting multi-hour uptime on a Heltec V4, when previously it would reboot in < 20 minutes.

NomDeTom:
I'm trying to work out when these additional lock guards were added. The automated tools were very aggressive in demanding that spilock was guarded, but it hadn't occurred to me that it was being done twice in some places.

I'll see if I can recover one of my older histories and work it out.

NomDeTom:
@compumike I've spent a bit of time looking at this, and I bow to your superior knowledge.

Is it worth adding some commentary to SafeFile to note the fact that there's different levels of locking needed?

compumike:
> Is it worth adding some commentary to SafeFile to note the fact that there's different levels of locking needed?

Yes, it could be worth adding a comment inside `SafeFile`, similar to the ones I added in this PR. Feel free to do so if you'd like!

I don't have enough historical context to know why `SafeFile` was implemented the way it was, or what the other callers do! It seems "cleaner" to either have all `SafeFile` methods do the locking, or none (and have the caller always be responsible for that), but that's a bigger investigation and refactor.

NomDeTom:
Just putting this note here for future-us: SafeFile isn't completely symmetrical - some operations are internally guarded and some require the user to be cogniscent of the guards required. However, when I looked at incorporating the guards completely within the abstraction it turns out that there are several operations which require a rapid succession of guarded operations which leads to what Clod called "guard flag churn". So, maybe we just make a note and move on.

PR #10841, opened 2026-07-01
XEdDSA packet signing UI (BaseUI)

### LINKED ISSUE #10800: [2.8.0] XEdDSA packet signing UI (BaseUI)
Surface XEdDSA signing status in **BaseUI** — the firmware's built-in on-device display UI (the classic OLED screens) — per the cross-platform spec in [meshtastic/design#113](https://github.com/meshtastic/design/issues/113) (source of truth, read it first).

> BaseUI is the on-device screen UI rendered by firmware. The richer **MUI** (device-ui / standalone-ui) is tracked separately ([device-ui#331](https://github.com/meshtastic/device-ui/issues/331), [standalone-ui#35](https://github.com/meshtastic/standalone-ui/issues/35)).

## Summary

Indicate when a node and its **broadcast** packets are cryptographically **signed and verified** (XEdDSA). The radio already computes this on-device — BaseUI just needs to render the indicator. **Affirm the good state only; never warn on unsigned traffic** (legacy nodes, DMs, and oversized broadcasts are legitimately unsigned).

## Signals (already available on-device)

- `MeshPacket.xeddsa_signed` — this packet's signature is verified (primary signal).
- `NodeInfo.has_xeddsa_signed` — this node signs its packets.

## What to build

- [ ] **Node screen:** a small shield glyph when `has_xeddsa_signed` ("this node signs its packets").
- [ ] **Message/text screen:** a shield glyph on a verified (`xeddsa_signed`) broadcast. **Never on DMs** — the flag is never set on DMs, so gating on the bool alone is safe.

## OLED adaptation

- BaseUI is monochrome and space-constrained, so the spec's "green & prominent" can't carry over literally — rely on a clear **shield glyph** in a consistent slot rather than color.
- Keep the existing lock/encryption indication separate from signing (shield = authentic, lock = private).

## v1 decisions (from #113)

- Affirm-only: unsigned is **silent** (no icon, no warning).
- No "unverifiable" state — use the verified bool only.
- Don't reuse the lock for signing; nothing here is an error state.

**Parent:** meshtastic/design#113

Xaositek:
- [x] **Node screen:** Added to Favorite frame to identify a node can sign packets
- [x] **Message/text screen:** Added to messages list so a "*" will be added to signed messages ahead of the node name

image
image
image

## Summary by CodeRabbit

* **New Features**
  * Signed messages are now visually marked with an XEdDSA badge/icon in message headers and favorite-user rows, with improved spacing and truncation to keep the display aligned.

* **Bug Fixes**
  * Signed-message status is now persisted and restored reliably, so the badge/icon stays accurate after app restarts.


HarukiToreda (approved):
Looks good

thebentern (approved):

jp-bennett (approved):

PR #10844, opened 2026-07-01
Fix audio module availability

### LINKED ISSUE #10522: Audio module should be excluded from non-2.4GHz capable radios
## Description

The Audio (Codec2) module is currently available on all radios regardless of whether they support 2.4 GHz operation. The module should only be available on radios that have a 2.4 GHz capable radio (LORA_24 region).

## Current Behavior

The audio module configuration is exposed on all hardware variants, even sub-GHz-only radios where Codec2 audio cannot function.

## Expected Behavior

The audio module should be added to the excluded modules list for all radios that do not support 2.4 GHz operation. It should only be available when:

1. The hardware has a 2.4 GHz capable radio
2. The region is set to `LORA_24`

## Additional Context

On the Apple client side, we've already gated the Audio module config screen to only show when the node's LoRa region is set to `LORA_24` (region code 13). The firmware should enforce the same restriction so the module cannot be enabled on incompatible hardware.

Reference: `meshtastic/Meshtastic-Apple` branch `copilot/add-audio-module-config-screen`

RCGV1:
## Summary

- Add a shared audio availability helper for ESP32/SX1280/LORA_24 checks.
- Mark Audio as excluded in device metadata unless the build supports Audio and the current LoRa region is `LORA_24`.
- Reject Admin get/set of `AUDIO_CONFIG` with `BAD_REQUEST` when Audio is unavailable.

Fixes #10522.

## Root Cause

Audio metadata was gated only by the build-time `MESHTASTIC_EXCLUDE_AUDIO` macro, so radios and regions that cannot support Audio could still advertise Audio config availability.

## Verification

- `git diff --check`
- `~/.platformio/penv/bin/python -m platformio test -e native-macos -f test_module_availability` - PASS, 4/4

## Summary by CodeRabbit

* **Bug Fixes**
  * Audio module settings are now only shown and accepted when supported for the current region and hardware.
  * Viewing or updating unsupported audio settings now returns a clear “bad request” error instead of continuing.
  * Device metadata now accurately marks the audio module as unavailable when it isn’t supported for the selected region.
* **Tests**
  * Added unit coverage for audio-module availability logic, including region-based behavior and build-configuration expectations.


RCGV1:
Updated the split audio availability PR onto current `develop` and re-verified locally.\n\nVerification:\n- `~/.platformio/penv/bin/python -m platformio test -e native-macos -f test_module_availability` passes 4/4.\n- `git diff --check origin/develop...HEAD` passes.\n- `~/.cache/trunk/launcher/trunk fmt` reports no issues.\n- Bench env build sweep: `muzi-base`, no-MUI `t-deck`, and `tbeam-s3-core` pass. `tlora-t3s3-v1` still fails the final PlatformIO flash-size check, but the same target also fails on unmodified `origin/develop` (develop: 2,451,967 > 2,424,832; this branch: 2,452,231 > 2,424,832), so I am treating that as a pre-existing board size budget issue rather than part of this fix.\n\nThe PR remains scoped to issue #10522: Audio config is advertised/accepted only for ESP32 + SX1280 builds when the active LoRa region is `LORA_24`; incompatible requests return BAD_REQUEST.

RCGV1:
Addressed the CodeRabbit review items in 93641dd8d:

- Preserved `pki_encrypted` on the unavailable-audio BAD_REQUEST path before the early return.
- Added direct coverage for `isAudioModuleAvailableForRegion()`.
- Updated `test/native-suite-count` to 26 for the new native suite.

Verification rerun locally:
- `~/.cache/trunk/launcher/trunk fmt`
- `git diff --check origin/develop...HEAD`
- `~/.platformio/penv/bin/python -m platformio test -e native-macos -f test_module_availability` (6/6)

PR #10858, opened 2026-07-02
Harden XEdDSA unsigned-packet policy and add coverage

thebentern:
## Summary

Audit of the XEdDSA packet-signing implementation (#10478) turned up several issues in *when* unsigned packets are accepted on receive or emitted on send. This PR fixes them and adds regression coverage.

## Fixes

- **Unicast NodeInfo exchange broke against signer nodes.** `NodeInfoModule` dropped *any* unsigned NodeInfo from a node whose signer bit was set — but unicast NodeInfo (want_response replies, phone-initiated exchanges) is never signed by the sender, so request and reply both died. The drop is now gated to broadcasts.
- **Sign/TOO_LARGE dead band.** The old sender gate (`payload + 64 < DATA_PAYLOAD_LEN`) admitted 167–168 B broadcasts whose *signed* encoding overflowed the LoRa frame, so they were signed and then failed `TOO_LARGE` — undeliverable, though they'd have been fine unsigned. Replaced with an exact encoded-size gate (`signedDataFits`), and the receive-side downgrade predicate now mirrors it byte-for-byte.
- **Plaintext-MQTT decoded downlink bypassed the policy.** Already-decoded packets skip `perhapsDecode`'s crypto path, so signature verification *and* downgrade protection were skipped entirely on plaintext brokers — a rogue peer could impersonate a signing node. The policy was extracted into `checkXeddsaReceivePolicy()` and is now applied at MQTT ingress.
- **Malformed-signature downgrade bypass.** A crafted signature of length 1–63 landed in the unsigned branch while its bytes inflated the size estimate past the fit threshold, letting a forged unsigned broadcast from a known signer dodge the drop. Signatures whose length is neither 0 nor 64 are now rejected as malformed.
- **Missing lock on the MQTT verify path.** `xeddsa_verify` mutates a shared Ed25519 key cache; the RF path holds `cryptLock` but the new MQTT path did not (races the BLE/proxy task on nRF52). Now locked.
- **Client-preset signatures** on packets we originate are cleared on all builds (previously only inside the XEdDSA guard, so excluded builds could transmit stale signatures that fail verification everywhere).
- **Randomized (hedged) signing.** XEdDSA per the Signal spec is randomized — the nonce is `r = hash1(a ‖ M ‖ Z)` with `Z` caller-supplied randomness, giving hedged signatures. The vendored library was previously a deterministic Ed25519-style variant that ignored `Z`. This PR bumps the `meshtastic/Crypto` pin to [meshtastic/Crypto#3](https://github.com/meshtastic/Crypto/pull/3) (which makes `XEdDSA::sign` mix the first 32 bytes of the signature buffer into the nonce as `Z`) and seeds those bytes in `xeddsa_sign` from `HardwareRNG::fill` with a checked return and a seeded-CSPRNG fallback. Signing the same content twice now yields different, both-valid signatures; a weak/repeated `Z` still produces a signature safe against nonce reuse, so signing never fails over RNG quality.

## Tests

- `test_packet_signing`: groups A (receive matrix), B (send policy incl. exhaustive payload-size sweeps over two Data shapes), C (NodeInfo backstop), D (wire-format invariants), E (decoded-ingress policy incl. malformed-signature cases).
- `test_mqtt`: four end-to-end plaintext-downlink cases (drop unsigned-from-signer, accept non-signer, verify + learn signer bit, drop bad signature).
- `test_crypto`: deterministic-nonce pin.

All native suites pass locally: **521/521** in the Docker coverage/ASan environment and via the `native-macos` toolchain.

## Notes

- No behavior change for nodes that don't sign (legacy/STM32/ham/region-unset) or for DMs.

## Summary by CodeRabbit

* **New Features**
  * Improved XEdDSA receive/transmit policy across mesh and MQTT using exact protobuf-encoded sizing to decide when signing and downgrade protection apply.
* **Bug Fixes**
  * Prevented unsigned broadcasts from previously recognized signers when they could still fit signed encoding limits.
  * Improved handling of malformed/partial/tampered signatures and ensured valid packets (including unsigned NodeInfo unicast) are treated correctly.
  * Added a safer signing randomness fallback when hardware entropy is unavailable.
* **Documentation**
  * Clarified XEdDSA signature field size/encoding overhead expectations.
* **Tests**
  * Expanded packet-signing, routing, and decoded-ingress coverage for boundary and downgrade scenarios.

### DIFF
--- a/src/mesh/CryptoEngine.cpp
+++ b/src/mesh/CryptoEngine.cpp
@@ -116,8 +116,11 @@ bool CryptoEngine::xeddsa_sign(uint32_t fromNode, uint32_t packetId, uint32_t po
     size_t sigLen = buildSigningBuffer(sigBuf, sizeof(sigBuf), fromNode, packetId, portnum, payload, payloadLen);
     if (sigLen == 0)
         return false;
-    // the XEdDSA::sign function requires at least the first 32 bytes of signature to be pre-filled with randomness
-    HardwareRNG::fill(signature, 32);
+    // XEdDSA::sign mixes signature[0..31] into the nonce as the spec's random Z (meshtastic/Crypto#3)
+    // for hedged signatures, so seed it - hardware RNG, else the seeded CSPRNG. A weak Z is still
+    // safe against nonce reuse (defense-in-depth only), so we never fail signing over it.
+    if (!HardwareRNG::fill(signature, 32))
+        CryptRNG.rand(signature, 32);
     XEdDSA::sign(signature, xeddsa_private_key, xeddsa_public_key, sigBuf, sigLen);
     return true;
 }
--- a/src/mesh/Router.cpp
+++ b/src/mesh/Router.cpp
@@ -12,6 +12,7 @@
 #include "mesh-pb-constants.h"
 #include "meshUtils.h"
 #include "modules/RoutingModule.h"
+#include <pb_encode.h>
 #if HAS_TRAFFIC_MANAGEMENT
 #include "modules/TrafficManagementModule.h"
 #endif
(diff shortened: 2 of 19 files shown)


src/mesh/CryptoEngine.cpp
```diff
@@ -116,8 +116,16 @@ bool CryptoEngine::xeddsa_sign(uint32_t fromNode, uint32_t packetId, uint32_t po
     size_t sigLen = buildSigningBuffer(sigBuf, sizeof(sigBuf), fromNode, packetId, portnum, payload, payloadLen);
     if (sigLen == 0)
         return false;
-    // the XEdDSA::sign function requires at least the first 32 bytes of signature to be pre-filled with randomness
-    HardwareRNG::fill(signature, 32);
```
jp-bennett, line 120: Just putting a pin in this, the RNG fill needs to come back.

src/mesh/Router.cpp
```diff
@@ -542,35 +593,11 @@ DecodeState perhapsDecode(meshtastic_MeshPacket *p)
...
-            if (node && nodeInfoLiteHasXeddsaSigned(node) && isBroadcast(p->to) &&
-                p->decoded.payload.size + XEDDSA_SIGNATURE_SIZE < meshtastic_Constants_DATA_PAYLOAD_LEN) {
-                LOG_WARN("Dropping unsigned broadcast from 0x%08x that previously signed", p->from);
-                return DecodeState::DECODE_FAILURE;
-            }
-        }
+        // rawSize is the size of the encoded Data exactly as the sender built it (the PKI branch's
+        // MESHTASTIC_PKC_OVERHEAD subtraction preserves that, and PKI packets are unicast so the
+        // downgrade predicate ignores them anyway).
+        if (!checkXeddsaReceivePolicy(p, rawSize))
```
jp-bennett, line 601: Good idea, to move the logic into a function.

test/test_crypto/test_main.cpp
```diff
@@ -263,20 +263,29 @@ void test_XEdDSA_max_payload(void)
     TEST_ASSERT_FALSE(crypto->xeddsa_verify(pub, fromNode, packetId, portnum, payload, len, signature));
 }
 
-// Signing the same message twice yields signatures that both verify. This XEdDSA implementation is
-// deterministic in practice (the two signatures are typically byte-identical, even though
-// HardwareRNG::fill provides real entropy on this platform), so we assert only the security-relevant
-// property - every produced signature verifies - rather than asserting (non-)determinism.
+// Signing the same message twice yields byte-identical signatures. XEdDSA per the Signal spec is
```
jp-bennett: These comments need updated after the crypto PR landed

test/test_crypto/test_main.cpp
```diff
@@ -263,20 +263,29 @@ void test_XEdDSA_max_payload(void)
...
+// has begun consuming sign-time bytes (e.g. a spec-compliant Z) and xeddsa_sign must then feed it
+// vetted entropy from a checked RNG.
 void test_XEdDSA_repeated_sign_verifies(void)
 {
     uint8_t pub[32], priv[32], sig1[64], sig2[64];
     uint8_t message[] = "same message";
     uint32_t fromNode = 0x9, packetId = 0x9, portnum = 9;
 
     crypto->generateKeyPair(pub, priv);
+    // Different garbage prefills: if sign consumed the caller's buffer contents, the two signatures would differ.
```
jp-bennett: Good test, let's make sure it resolves differently with the PR

jp-bennett (changes requested):

jp-bennett (approved):

PR #10862, opened 2026-07-02
Harden against crafted-packet crashes + adversarial fuzzing

thebentern:
## Summary

Hardens the RF-packet processing path against crafted packets that could crash a node or a paired phone app — the surface beyond the XEdDSA authenticity work. Two memory-safety crashes were found and reproduced under AddressSanitizer, then fixed. Also adds defense-in-depth and two ASan fuzz suites.

## Crash fixes (reproduced under AddressSanitizer)

- **GeoCoord UTM/MGRS out-of-bounds read** — extreme `latitude_i`/`longitude_i` from a received `Position` drove the band/zone lookups (`latBands`, `e100kLetters`, `n100kLetters`) out of bounds. Clamp the UTM zone to 1–60 and guard the band/column/row indices. Regression: `test_geocoord_extreme_coords_no_oob`.
- **Telemetry `String(float)` stack overflow** — attacker-controlled telemetry floats near `FLT_MAX` overflow Arduino `String(float)`'s fixed `char[33]` (`dtostrf`) on nRF52/RP2040/STM32/portduino (ESP32 is unaffected — it allocates dynamically). New `UnitConversions::displaySafeFloat` bounds the rendered metrics; unit-tested in `test_type_conversions`.

## Defense-in-depth & robustness

- `TraceRouteModule::printRoute`: fix an out-of-bounds read caused by the wrong count in a guard condition (debug builds only).
- `WaypointModule` / `NodeDB`: sanitize untrusted strings before the OLED renderer and the phone-facing `ClientNotification` (belt-and-suspenders alongside the global `PB_VALIDATE_UTF8`).
- `MeshService::sendToPhone`: withhold NodeInfo/Waypoint packets whose nested string field won't cleanly decode, protecting strict phone protobuf decoders — without affecting mesh relay (which forwards the encrypted form on a separate path).

## Tests

- New `test_fuzz_decode` (protobuf-decode + UTF-8-sanitizer fuzz over the mesh-facing message types) and `test_fuzz_packets` (`perhapsDecode` / module-handler / traceroute / phone-gate fuzz), all under AddressSanitizer with a deterministic seeded RNG for reproducible failures.
- `native-suite-count` 25 → 27. Full native suite: **515/515 green**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes from extreme/unvalidated coordinate inputs during location conversions.
  * Improved telemetry display robustness by safely handling non-finite and very large float values.
  * Sanitized waypoint and device/user text before rendering to the OLED for better UI reliability.
  * Hardened forwarding to the phone by safely rejecting payloads that can’t be decoded.
  * Corrected Trace Route diagnostics to show the proper final-hop SNR.

* **Tests**
  * Added fuzzing and regression coverage for decoding, UTF-8 sanitation, and numeric clamping.


thebentern:
Thanks for the review — addressed both actionable comments and the nitpicks:

**Actionable**
- **GeoCoord double→int narrowing (float-cast-overflow UB):** confirmed with UBSan (`-fsanitize=float-cast-overflow`) — the actual site was upstream at the UTM `easting`/`northing` assignments to their `uint32_t` fields (extreme lat/lon can drive them negative/oversized). Added a `clampMeters()` narrowing helper (finite + `[0, 4e9]`) before the cast, which also makes the MGRS `col`/`row` casts provably in-range. The extreme-coord sweep now runs clean under **ASan + UBSan**.
- **TraceRoute origin-hop SNR:** the `INT8_MIN` "unknown" sentinel is no longer formatted as a dB value.

**Nitpicks:** condensed the over-length comments to the 1–2 line guideline (MeshService `.h`/`.cpp`, NodeDB, UnitConversions, EnvironmentTelemetry, WaypointModule, test comment) and added the `makeData` payload-size/`NOT_NULL` asserts.

Full native suite remains green (515/515 under AddressSanitizer).

PR #10863, opened 2026-07-03
nrf52_lto: fix variant LTO exclusion

NomDeTom:
The CI-fix follow-up in #10829 anchored _is_board_variant() to <PROJECT_DIR>/variants/ using node.get_abspath() -- but build middleware receives nodes from the SCons variant-dir mirror, whose abspath is $BUILD_DIR/variants/.../variant.cpp. The anchor never matched, so the variant silently went back into whole-image LTO: the shipped ELF has initVariant() resolved to the core's weak empty stub (bare `bx lr`), PIN_3V3_EN is never driven, and the SX1262 probe fails with CHIP_NOT_FOUND again -- while the build stays green, because the post-link guard only checked IRQ handlers.

Fix: match against node.srcnode().get_abspath(), which undoes the variant-dir mirror (the same trick piobuild.py uses for middleware pattern matching).

Add a second post-link guard so this failure mode is a red build, not a field failure: the linked variant.cpp.o must contain no .gnu.lto_* sections (proves the -fno-lto recompile fired), and any override the object defines strong (initVariant) must resolve strong in the ELF. Verified both ways on nrf52_promicro_diy_tcxo: clean build is green with T _Z11initVariantv and a real initVariant body in the ELF; re-breaking the matcher turns the build FAILED with both guard checks firing.

clod helped a lot

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of board variant detection during the build by matching against the true variant source location.
  * Ensured board variant recompilation uses the correct build settings, preserving generated headers and required version information.
  * Added stronger post-link validation to confirm critical variant override symbols remain intact in the final output, and to stop the build if LTO affects the variant objects.


thebentern:
Pushed a fix for the three failing nrf52840 variant builds (`rak_wismeshtag`, `seeed_xiao_nrf52840_kit`, `seeed_mesh_tracker_X1`), all failing with `#error APP_VERSION must be set` during the `-fno-lto` variant recompile.

**Root cause:** the variant middleware snapshots `projenv["CCFLAGS"]` when it fires (during `$BUILD_SCRIPT`), but `-DAPP_VERSION…` is appended to `projenv` by `bin/platformio-custom.py` — an unprefixed `extra_scripts` entry, which PlatformIO runs as a **post** script, i.e. *after* the middleware. So the frozen per-node override never contained `APP_VERSION`. Only these three boards tripped it because their `variant.cpp` includes `configuration.h` (via `sleep.h` or directly); other boards' variants don't.

**Fix:** make the variant CCFLAGS override a deferred callable so SCons reads `projenv`'s CCFLAGS at command-substitution time (after post-scripts), by which point `APP_VERSION` is present. `-fno-lto` is still appended last.

Verified clean builds of `rak_wismeshtag`, `seeed_xiao_nrf52840_kit`, and `rak4631` (regression) — all green, with your `variant guard OK — board variant kept out of LTO` firing on each (so the variant is still correctly excluded from LTO).

NomDeTom:
I've tested the latest commit and it still works on the promicro.

PR #10878, opened 2026-07-04
fixed tlora-v2-1-1_16 led heartbeat not blinking and stuck at single s…

lasek0:
since I hate waste and trash old hardware (which could still works) I am still using tlora-v2-1-1_16 usb powered
formware for tlora-v2-1-1_16 has a bug that caused the status LED to not blinking,
conflict setup LED_POWER for blinking and power status caused that.

fixed by using LED_HEARTBEAT instead of LED_POWER and force blinking regardless power status

## Summary by CodeRabbit

* **New Features**
  * Enabled the heartbeat status LED by default for the ESP32 build, with early startup GPIO initialization.
* **Bug Fixes**
  * Updated heartbeat LED behavior so it can toggle independently of power/charging states on supported builds, based on a build-time setting and the user’s “heartbeat disabled” option.
* **Configuration**
  * Updated the ESP32 variant LED/buzzer build settings: heartbeat LED now uses pin 25 (and the prior LED_POWER pin assignment is removed), and the buzzer pin definition was removed for the affected environment.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -348,6 +348,12 @@ void setup()
     // initialize power HAL layer as early as possible
     powerHAL_init();
 
+#if defined(ARCH_ESP32) && defined(LED_HEARTBEAT)
+    pinMode(LED_HEARTBEAT, OUTPUT);
+    digitalWrite(LED_HEARTBEAT, LED_STATE_ON);
+#endif
+
+
 #ifdef LED_POWER
     pinMode(LED_POWER, OUTPUT);
     digitalWrite(LED_POWER, LED_STATE_ON);
--- a/src/modules/StatusLEDModule.cpp
+++ b/src/modules/StatusLEDModule.cpp
@@ -140,7 +140,11 @@ int32_t StatusLEDModule::runOnce()
     }
 // If we want a LED to be dedicated to the simple hearbeat, we can use that instead of the charge LED
 #if defined(LED_HEARTBEAT)
+    #ifndef USE_HEARTBEAT_REGARDLESS_POWER_STATE
     if (power_state != charging && power_state != charged && !doing_fast_blink && !config.device.led_heartbeat_disabled) {
+    #else
+    if (!config.device.led_heartbeat_disabled) {
+    #endif
         if (HEARTBEAT_LED_state == LED_STATE_ON) {
             HEARTBEAT_LED_state = LED_STATE_OFF;
             my_interval = 999;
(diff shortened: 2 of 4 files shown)


jp-bennett:
The intent for the power status is for the LED to do the quick blink normally, do the slow blink while charging, and go solid to indicate charge completed. 

The LED_HEARTBEAT define is supposed to ignore the charged state, and do the fast or slow blink even when fully charged. That macro may need a bit more work, if it's not working as intended.

lasek0:
> The intent for the power status is for the LED to do the quick blink normally, do the slow blink while charging, and go solid to indicate charge completed.
> 
> The LED_HEARTBEAT define is supposed to ignore the charged state, and do the fast or slow blink even when fully charged. That macro may need a bit more work, if it's not working as intended.

In general from you description the `LED_HEARTBEAT` should ignore the power-state and blink (in 1Hz period) but current implementation does not work as you describe. Even if I define the `LED_HEARTBEAT` and undefine `LED_POWER`  status led on board goes solid indicating power status.

consider the following :
```
        if (POWER_LED_starttime + 30000 < millis() && !doing_fast_blink) {
            doing_fast_blink = true;
            POWER_LED_starttime = millis();
        }
        if (doing_fast_blink) {
            PAIRING_LED_state = LED_STATE_OFF;
            CHARGE_LED_state = !CHARGE_LED_state;
            my_interval = 250;
            if (POWER_LED_starttime + 2000 < millis()) {
                doing_fast_blink = false;
                CHARGE_LED_state = LED_STATE_OFF;
            }
        }
```
this code snippet indicate that the `doing_fast_blink` influence only power state and charge state but NOT heartbeat state. 

- Since the tlora-va-1-1_16 has on board led for charge status it is pointless to define the `LED_POWER` and duplicate the charging status.
- when `LED_HEARTBEAT` will be defined the expected behavior will be to just blink and not indicate anything else.

currently the only place when  `HEARTBEAT_LED_state` is change is that this code, and in my opinion it should be independent on other statuses

```
#if defined(LED_HEARTBEAT)
    if (power_state != charging && power_state != charged && !doing_fast_blink && !config.device.led_heartbeat_disabled) {
        if (HEARTBEAT_LED_state == LED_STATE_ON) {
            HEARTBEAT_LED_state = LED_STATE_OFF;
            my_interval = 999;
        } else {
            HEARTBEAT_LED_state = LED_STATE_ON;
            my_interval = 1;
        }
        digitalWrite(LED_HEARTBEAT, HEARTBEAT_LED_state);
    } else {
        HEARTBEAT_LED_state = LED_STATE_OFF;
        digitalWrite(LED_HEARTBEAT, HEARTBEAT_LED_state);
    }
#endif
```

lasek0:
any suggestions, welcome

PR #10884, opened 2026-07-05
Implement SD card backup/restore of preferences; add SD support for Cardputer Advanced

termax:
Implements the two `// TODO: After more mainline SD card support` stubs in `NodeDB::backupPreferences` / `NodeDB::restorePreferences`, so the existing `backup_preferences` / `restore_preferences` admin messages work with `BackupLocation_SD`, and wires up the SD card slot on the M5Stack Cardputer Advanced (which has a physical slot the firmware previously didn't use).

## Why

Flash-location backups don't survive reflashes or flash repartitioning. Devices with an SD slot can keep the same `BackupPreferences` proto on removable storage, which makes full config recovery (including channels and security keys) possible after a wipe — particularly useful on multi-firmware devices (e.g. Cardputer users switching firmwares via a boot launcher).

## What

- `NodeDB.cpp`: SD branches for backup and restore, mirroring the FLASH implementation. Same proto, same `/backups/backup.proto` path on the card. Guarded by `HAS_SDCARD && !SDCARD_USE_SOFT_SPI` (same condition as `FSCommon`'s SD support); devices without SD support now log an explicit error instead of silently returning false. SD access is wrapped in `spiLock` like other filesystem/radio bus users.
- `variants/esp32s3/m5stack_cardputer_adv/variant.h`: SD pin definitions (`SPI_SCK 40 / SPI_MISO 39 / SPI_MOSI 14 / SDCARD_CS 12`). The slot shares the SPI bus with the SX1262 (separate CS); pins match the working configuration used by community firmware that drives this slot (M5Launcher).

## Testing

- **Hardware-tested on two M5Stack Cardputer Advanced units** with this code backported onto v2.7.26: SD card mounts at boot (SDHC, correct size), backup writes verified (`Saved backup preferences to SD card`, ~600-700 bytes), restore verified end-to-end (config + module config + channels/PSKs + owner all applied and persisted; node identity/keys recovered after a full flash wipe). LoRa RX/TX confirmed unaffected by the shared SPI bus (messages exchanged with other mesh nodes after restore).
- This branch compiles clean for `m5stack-cardputer-adv` against current master.
- **Not tested**: master on hardware (our devices run the 2.7.26 backport of this exact code), and other `HAS_SDCARD` devices (e.g. T-Deck) — the code paths only execute when a client explicitly requests the SD location, which previously no-op'd, but community testing on other SD-equipped boards would be welcome.

## Transparency

This contribution was developed with substantial AI assistance (Anthropic's Claude, in an agentic coding session), with a human directing the work and performing all hardware testing. Please review accordingly. It is offered as-is under the project's license, without warranty of any kind — we're sharing it because it filled a real gap for us and may help others, and we're happy to revise per maintainer feedback.

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described. (On M5Stack Cardputer Advanced, via a v2.7.26 backport of this exact code — see Testing above.)
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Regression testing done on: M5Stack Cardputer Advanced (×2). No other hardware available to us; the changed code paths are compile-gated to `HAS_SDCARD` builds and runtime-gated to explicit `BackupLocation_SD` requests, so regression surface elsewhere should be minimal — but independent verification on T-Deck (which also defines an SD card) would be valuable.

## Summary by CodeRabbit

* **New Features**
  * Added SD card support for backing up and restoring device preferences on supported hardware.
  * Enabled SD card configuration for a supported device variant.

* **Bug Fixes**
  * Improved reliability of backup and restore operations with better error handling and file validation.
  * Restores now apply settings more completely and save them back safely after recovery.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -2179,23 +2179,52 @@ bool NodeDB::checkLowEntropyPublicKey(const meshtastic_Config_SecurityConfig_pub
 }
 #endif
 
+#ifdef FSCom
+// Shared by the FLASH and SD backup locations so the two paths can't drift apart
+static meshtastic_BackupPreferences buildBackupPreferences()
+{
+    meshtastic_BackupPreferences backup = meshtastic_BackupPreferences_init_zero;
+    backup.version = DEVICESTATE_CUR_VER;
+    backup.timestamp = getValidTime(RTCQuality::RTCQualityDevice, false);
+    backup.has_config = true;
+    backup.config = config;
+    backup.has_module_config = true;
+    backup.module_config = moduleConfig;
+    backup.has_channels = true;
+    backup.channels = channelFile;
+    backup.has_owner = true;
+    backup.owner = owner;
+    return backup;
+}
+
+static void applyRestoredPreferences(const meshtastic_BackupPreferences &backup, int restoreWhat)
+{
+    if (restoreWhat & SEGMENT_CONFIG) {
+        config = backup.config;
+        LOG_DEBUG("Restored config");
+    }
+    if (restoreWhat & SEGMENT_MODULECONFIG) {
+        moduleConfig = backup.module_config;
+        LOG_DEBUG("Restored module config");
+    }
+    if (restoreWhat & SEGMENT_DEVICESTATE) {
+        devicestate.owner = backup.owner;
+        LOG_DEBUG("Restored device state");
+    }
+    if (restoreWhat & SEGMENT_CHANNELS) {
+        channelFile = backup.channels;
+        LOG_DEBUG("Restored channels");
+    }
+}
+#endif
+
 bool NodeDB::backupPreferences(meshtastic_AdminMessage_BackupLocation location)
 {
     bool success = false;
     lastBackupAttempt = millis();
 #ifdef FSCom
     if (location == meshtastic_AdminMessage_BackupLocation_FLASH) {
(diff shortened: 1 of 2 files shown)


termax:
@coderabbitai review

termax:
Both review suggestions are implemented in 975eecf: extracted `buildBackupPreferences()` and `applyRestoredPreferences()` shared by the FLASH and SD branches (as file-local helpers rather than methods — the state they touch is global, so no header changes needed).

caveman99:
please review the CLA. We won't accept contributions without it.

termax:
> please review the CLA. We won't accept contributions without it.

You have agreed to the CLA for multiple repositories or organizations
 is that it or need to do something else as well

termax:
> > please review the CLA. We won't accept contributions without it.
> 
> You have agreed to the CLA for multiple repositories or organizations is that it or need to do something else as well

i have done it 3 times already, it still shows unsigned

termax:
What do you need from me to move this forward ? Your CLA automation was done 3+ times and still shows unsigned.

mverch67:
I see a few issues with this PR:
- have to include FSCommon.h instead of SD.h. FSCommon handles architecture specific SD card usage
- DO NOT use global SD directly. Always use FSCom instead. FSCom maps to the architecture specific definitions.
- backup/restore should make use of versioning, i.e. in case of known incompatibilities increase the version in code and reject the restore of older versions.

PR #10885, opened 2026-07-05
Fix GAT562 mesh time sync before GPS lock

gat-iot:
## Summary

Fixes GAT562 time sync behavior when the GPS module is connected but has not provided valid RTC time yet.

On GAT562, `hasGPS()` can be true as soon as the GPS UART module is detected. `PositionModule::hasQualityTimesource()` then treats the device as already having a quality time source and ignores time from mesh position packets. This prevents the terminal from syncing time over the mesh before GPS lock.

This change keeps the existing behavior for other targets, and for GAT562 only treats the device as having a quality time source once RTC quality is actually at least `RTCQualityFromNet`, or a hardware RTC / recent phone-NTP source is available.

## Validation

Built successfully with:

```bash
python -m platformio run -e gat562_mesh_trial_tracker
```

Result: `SUCCESS`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved time source detection on supported devices, making it more likely the app recognizes a valid clock source when mesh quality, hardware RTC, or recent phone/NTP time is available.
  * Preserved existing time source behavior on other builds, including configurations that exclude GPS.

### DIFF
--- a/src/modules/PositionModule.cpp
+++ b/src/modules/PositionModule.cpp
@@ -149,12 +149,21 @@ bool PositionModule::hasQualityTimesource()
 {
     bool setFromPhoneOrNtpToday =
         lastSetFromPhoneNtpOrGps == 0 ? false : Throttle::isWithinTimespanMs(lastSetFromPhoneNtpOrGps, SEC_PER_DAY * 1000UL);
+#if defined(GAT562)
+    // A connected GAT562 GPS means the UART module is present, not that it has
+    // provided valid time yet. Let mesh position packets seed RTC until this
+    // device actually has a valid clock from mesh, phone/NTP, RTC, or GPS.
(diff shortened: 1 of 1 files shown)


src/modules/PositionModule.cpp
```diff
@@ -149,12 +149,21 @@ bool PositionModule::hasQualityTimesource()
 {
     bool setFromPhoneOrNtpToday =
         lastSetFromPhoneNtpOrGps == 0 ? false : Throttle::isWithinTimespanMs(lastSetFromPhoneNtpOrGps, SEC_PER_DAY * 1000UL);
+#if defined(GAT562)
```
caveman99, line 152: Is this behaviour specific to GAT562 or will this benefit all GPS modules?

caveman99 (changes requested):
Please sign the CLA.

PR #10886, opened 2026-07-05
Set GAT562 GPS pins from board defaults

gat-iot:
## Summary

Sets the GAT562 GPS enable pin and GPS UART pins from the board definition, and migrates existing saved configs that still have missing or stale GPS pin values.

GAT562 uses the controllable 3V3 rail on GPIO 34 for GPS power. If `gps_en_gpio` remains unset in saved config, GPS can be reported as present but not actually powered as expected. This change:

- defines `PIN_GPS_EN` as `PIN_3V3_EN` for the GAT562 variant
- marks the GPS enable pin active high
- restores GAT562 GPS RX/TX/EN config from board defaults at boot when needed
- avoids applying the WisBlock canned-message EN-pin conflict guard to GAT562

## Validation

Built successfully with:

```bash
python -m platformio run -e gat562_mesh_trial_tracker
```

Result: `SUCCESS`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPS hardware handling so the GPS power/control pin now works correctly on the affected device variant.
  * Ensured GPS-related settings are automatically corrected during startup when needed, and saved for future boots.
  * Allowed GPS pin control to proceed on the supported hardware even when certain interrupt features are enabled.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -638,6 +638,34 @@ NodeDB::NodeDB()
         config.position.gps_mode = meshtastic_Config_PositionConfig_GpsMode_ENABLED;
         config.position.gps_enabled = 0;
     }
+#if defined(GAT562)
+#if defined(GPS_RX_PIN)
+    if (config.position.rx_gpio != GPS_RX_PIN) {
+        config.position.rx_gpio = GPS_RX_PIN;
+        if (!configDecodeFailed)
+            saveWhat |= SEGMENT_CONFIG;
+    }
+#endif
+#if defined(GPS_TX_PIN)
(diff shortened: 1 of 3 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -638,6 +638,34 @@ NodeDB::NodeDB()
         config.position.gps_mode = meshtastic_Config_PositionConfig_GpsMode_ENABLED;
         config.position.gps_enabled = 0;
     }
+#if defined(GAT562)
```
caveman99, line 641: This is not how you do it - for a specific board just define the presets in the variant.h like every other board does. No need to blow up the code.

caveman99 (changes requested):
Sign the CLA.

PR #10887, opened 2026-07-05
InkHUD: persist incoming broadcasts on all channels

drmzperx:
## Summary

InkHUD instantiates a `ThreadedMessageApplet` only for channels 0 and 1 (hardcoded in every `variants/*/nicheGraphics.h`), and those applets were the only path storing **incoming broadcasts** into the shared global `messageStore`. As a result, broadcasts on any other channel (2–7), or on channel 1 while its applet is inactive (`defaultActive=false`, or disabled via the menu), were never persisted — they existed only in the transient `latestMessage` cache and were lost on reboot / absent from message history. DMs were unaffected (already stored directly by `Events::onReceiveTextMessage`).

## Fix

Move persistent storage of incoming **text** broadcasts into `Events::onReceiveTextMessage` — the single InkHUD text-message observer, which is invoked for every channel regardless of which applets exist — mirroring how DMs are already stored there. `ThreadedMessageApplet::handleReceived` now stores only our own **outgoing/loopback** messages (which `Events` short-circuits at the top), so incoming broadcasts on channels 0/1 are not double-stored.

Storage is now partitioned by **direction** (Events = incoming, applet = outgoing), which are disjoint sets, so exactly one `addFromPacket` runs per message.

Only `TEXT_MESSAGE_APP` broadcasts are persisted. Other ports that also reach the text observer via `MeshService::isTextPayload` (detection sensor, alert, range test) continue to update only the transient cache, so periodic range-test traffic can't evict real message history.

Docs (`src/graphics/niche/InkHUD/docs/README.md`) updated to match the direction-partitioned flow.

**Changed files**
- `src/graphics/niche/InkHUD/Events.cpp`
- `src/graphics/niche/InkHUD/Applets/User/ThreadedMessage/ThreadedMessageApplet.cpp`
- `src/graphics/niche/InkHUD/docs/README.md`

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

**Testing status — please read:** I do not currently have an InkHUD e-ink device, so per the contributing guidance I'm flagging this for community testing rather than claiming hardware verification. What has been done:

- **Compile-verified**: builds clean for `t-echo-inkhud` (nRF52840), no errors/warnings on the changed files.
- **Logic-reviewed**: traced all four `addFromPacket` call sites and the observer/module wiring to confirm no double-store and no new missing-store, and that non-text ports (detection sensor / alert / range test) are excluded from persistence.
- I could not run `trunk fmt` locally (tooling unavailable in my environment); happy to push a formatting fixup if CI flags anything.

Testing help on any InkHUD e-ink variant (t-echo, Heltec Vision Master E213, T114-InkHUD, etc.) would be appreciated — specifically:
1. Send a broadcast on a channel with index ≥ 2, reboot, and confirm it now appears in message history.
2. Confirm channel 0/1 broadcasts are **not** duplicated.
3. Confirm detection-sensor / alert / range-test broadcasts do **not** flood the stored message history.

Allow edits by maintainers is enabled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate broadcast entries by changing what gets added to message history.
  * Text messages are now persisted through the main receive path, while other broadcast types update only the transient latest-message state.
  * Kept auto-show and notification behavior consistent for incoming messages.

* **Documentation**
  * Updated the InkHUD README to describe the new direction-partitioned persistence flow for outgoing/loopback versus incoming text and non-text broadcasts.

### DIFF
--- a/src/graphics/niche/InkHUD/Applets/User/ThreadedMessage/ThreadedMessageApplet.cpp
+++ b/src/graphics/niche/InkHUD/Applets/User/ThreadedMessage/ThreadedMessageApplet.cpp
@@ -190,12 +190,17 @@ ProcessMessage InkHUD::ThreadedMessageApplet::handleReceived(const meshtastic_Me
     if (mp.to != NODENUM_BROADCAST)
         return ProcessMessage::CONTINUE;
 
-    // Store in the global messageStore - this handles sender, timestamp, channel, text, and ack status
-    messageStore.addFromPacket(mp);
-
-    // If this was an incoming message, suggest that our applet becomes foreground, if permitted
-    if (getFrom(&mp) != nodeDB->getNodeNum())
+    // Incoming broadcasts are stored centrally by Events::onReceiveTextMessage (for every
+    // channel, regardless of which applets exist). Only store our own outgoing/loopback
+    // messages here - Events short-circuits outgoing - to avoid double-storing incoming
+    // broadcasts on channels 0/1.
+    if (getFrom(&mp) == nodeDB->getNodeNum()) {
+        // Outgoing (e.g. canned messages generated on this node)
+        messageStore.addFromPacket(mp);
+    } else {
+        // Incoming: suggest that our applet becomes foreground, if permitted
         requestAutoshow();
+    }
 
     // Redraw the applet, perhaps.
     requestUpdate(); // Want to update display, if applet is foreground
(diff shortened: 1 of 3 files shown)


HarukiToreda:
There will be a revamp of the Channel Applets so they can display as many channels as the device is configured to use. Once that's in place, simply enabling the desired channel applets will solve this issue.

There's also the matter of storage. The current message store only holds 20 messages before it starts deleting the oldest ones. The reasoning behind the current behavior is that storing messages for applet channels the user has chosen not to display would still consume those limited message slots, potentially pushing out messages from the channels the user actually cares about.

drmzperx:
Thanks for the context — that makes sense, and I hadn't properly weighted the storage-slot constraint. Given the message store only keeps ~20 messages, blanket-persisting broadcasts on channels the user hasn't chosen to display would compete for those slots and could evict messages from the channels they actually care about. So the current behavior is a deliberate tradeoff, not just an oversight — understood.

The Channel Applet revamp (enabling as many channels as the device is configured for) sounds like the right fix — it closes the visibility gap without the storage downside. I'll close this in favor of that. Happy to help test the revamp on InkHUD hardware once it lands, if useful.

Thanks again!

PR #10899, opened 2026-07-06
CI: track RAM (.data+.bss) in size reports and gate on per-env budgets

thebentern:
## Motivation

The 2.8.0 nRF52840 heap regression (field reports of 99% heap usage) shipped invisibly: CI tracks flash sizes, but nothing tracks RAM. On nRF52840 the heap arena is the linker gap after `.bss`, so **every byte of static `.data`+`.bss` growth shrinks the usable heap 1:1**. rak4631 is also at 95.8% of its app flash region (`0x27000..0xEA000` = 798,720 B). This PR gives RAM the same CI guardrails flash already has, plus a hard budget gate for the tightest target.

### 1. RAM in the size manifests and PR size report

- `bin/platformio-custom.py` now emits `ram_bytes` (`.data` + `.bss` summed from `size -A` on the ELF, via the toolchain's `$SIZETOOL`) into each `.mt.json` manifest. Heap/stack placeholder sections (e.g. the nRF52 `.heap` section) are deliberately excluded - the heap arena is the linker gap, not a static allocation.
- `bin/collect_sizes.py` records `{"flash_bytes": ..., "ram_bytes": ...}` per env; `bin/size_report.py` grows **RAM** and **RAM-delta** columns. Older artifacts without `ram_bytes` (including legacy int-schema baseline artifacts from previous develop/master runs) degrade gracefully to `n/a` - never crash.

Sample output (synthetic manifests; heltec-v3's manifest lacks `ram_bytes`, baseline is mixed legacy/new schema):

**2 targets | vs `develop`: 2 increased, net +2,528 (+2.5 KB)**

| Target | Flash | vs `develop` | RAM | RAM vs `develop` |
|--------|------:|----------:|----:|----------:|
| `heltec-v3` | 1,310,720 | 📈 +1,720 (+1.7 KB) | n/a | n/a |
| `rak4631` | 766,000 | 📈 +808 | 111,460 | 📈 +512 |

### Size budgets

| Env | Metric | Measured | Budget | Used |
|-----|--------|---------:|-------:|-----:|
| `rak4631` | RAM (.data+.bss) | 111,460 | 113,000 | 98.6% |
| `rak4631` | flash | 766,000 | 786,000 | 97.5% |

### 2. Budget gate (`bin/ram_budgets.json` + `size-budget-gate` job)

- `{"<env>": {"ram_bytes": <budget>, "flash_bytes": <budget>}}`; enforced **only for envs listed in the file** (and only when that env was built in the run). `"_comment"` keys document the semantics.
- Seeded with **rak4631 only**: RAM 113,000 (current 110,948 + ~2 KB slack), flash 786,000 (current 765,192 + ~20 KB; region cap is 798,720 and the image must also stay clear of the warm-store record-ring guard in `extra_scripts/nrf52_warm_region.py`).
- A new `size-budget-gate` CI job runs `size_report.py --budgets bin/ram_budgets.json --enforce-budgets` and **fails the build** on violation - it is deliberately separate from `firmware-size-report`, which is informational and `continue-on-error`. That job now additionally renders the budget-usage table into the PR comment.
- Failure output names the env, measured value, budget, and overage, and points at the fix:

```
RAM/flash budget check FAILED:
  - rak4631 RAM (.data+.bss): 115,200 bytes exceeds the budget of 113,000 bytes (over by 2,200)

On nRF52840 the heap arena is the linker gap after .bss, so every
byte of static RAM growth shrinks the usable heap 1:1. Budgets are
raised deliberately, never automatically: if this growth is intended
and reviewed, raise the env's limit in bin/ram_budgets.json in this
same PR and justify the increase in the PR description.
```

**Budgets are raised deliberately, never automatically**: bump the limit in `bin/ram_budgets.json` in the same PR that needs the headroom and justify it in the PR description.

### 3. Boot heap watermark flag (firmware, opt-in)

At the very end of `setup()`, under `#ifdef MESHTASTIC_HEAP_WATERMARK_CHECK`, log `LOG_ERROR` when less than 20% of the heap is free at boot. Off by default; CI/test builds enable with `-DMESHTASTIC_HEAP_WATERMARK_CHECK`. Platforms without heap accounting (`getHeapSize()` returning `UINT32_MAX`/0, see `src/memGet.cpp`) are skipped.

## Verification

- `python3 bin/test_size_scripts.py`: **23/23 pass**, including new cases for `ram_bytes` collection, malformed/missing `ram_bytes`, legacy-schema baselines, the `n/a` fallback, and the budget gate (over-RAM, over-flash, under, env-not-built, missing-metric, render-only, `--enforce-budgets` without `--budgets`).
- Section parser validated against real `size -A` shapes: nRF52 output yields exactly 110,948 (`.data` 2,504 + `.bss` 108,444, `.heap`/`.stack_dummy` excluded); ESP32 `.dram0.data`/`.dram0.bss` are picked up, `.flash.rodata`/debug sections excluded.
- End-to-end sanity run of `collect_sizes.py` → `size_report.py` against a synthetic manifest pair (with/without `ram_bytes`) and a mixed-schema baseline - the tables above are that run's output; enforce mode verified exit 1 over budget, exit 0 under.
- `trunk fmt` / `trunk check` clean on touched files (no new issues).
- No firmware builds run locally; the workflow changes are exercised by this PR's own CI (this PR builds rak4631 in the matrix, so the new gate runs here).

## Summary by CodeRabbit

* **New Features**
  * Added size reporting with separate Flash and RAM visibility, including optional budget details in generated reports.
  * Introduced a CI gate that can enforce RAM/flash size budgets and fail the build when limits are exceeded.
* **Bug Fixes**
  * Improved size summary robustness by supporting both new and legacy manifest formats and ignoring malformed RAM entries.
* **Changes**
  * Added an optional startup guard to warn about low available heap during initialization (when enabled).

### DIFF
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -387,18 +387,18 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
+      # No continue-on-error / empty-dir fallback: the gate must fail closed when
+      # the data it enforces on cannot be fetched (size_report.py additionally
+      # fails on missing budgeted envs under --enforce-budgets).
       - name: Download current manifests
         uses: actions/download-artifact@v8
-        continue-on-error: true
         with:
           path: ./manifests/
           pattern: manifest-*
           merge-multiple: true
 
       - name: Collect current firmware sizes
-        run: |
-          mkdir -p ./manifests/
-          python3 bin/collect_sizes.py ./manifests/ ./current-sizes.json
+        run: python3 bin/collect_sizes.py ./manifests/ ./current-sizes.json
 
       - name: Enforce RAM/flash budgets
         run: python3 bin/size_report.py ./current-sizes.json --budgets bin/ram_budgets.json --enforce-budgets
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -337,7 +337,7 @@ jobs:
         if: github.event_name == 'pull_request'
         id: report
         run: |
-          ARGS="./current-sizes.json"
+          ARGS="./current-sizes.json --budgets bin/ram_budgets.json"
           if [ -f ./develop-sizes.json ]; then
             ARGS="$ARGS --baseline develop:./develop-sizes.json"
           fi
@@ -374,6 +374,34 @@ jobs:
             ./size-report.md
             ./pr-number.txt
           retention-days: 5
+
+  # RAM/flash guardrails: fails CI when an env listed in bin/ram_budgets.json
+  # exceeds its static RAM (.data+.bss) or flash budget. Kept separate from
+  # firmware-size-report, which is informational and continue-on-error.
+  size-budget-gate:
+    permissions:
+      contents: read
+      actions: read
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Download current manifests
+        uses: actions/download-artifact@v8
+        continue-on-error: true
+        with:
+          path: ./manifests/
+          pattern: manifest-*
+          merge-multiple: true
+
+      - name: Collect current firmware sizes
+        run: |
+          mkdir -p ./manifests/
+          python3 bin/collect_sizes.py ./manifests/ ./current-sizes.json
+
+      - name: Enforce RAM/flash budgets
+        run: python3 bin/size_report.py ./current-sizes.json --budgets bin/ram_budgets.json --enforce-budgets
 
   release-artifacts:
     permissions: # Needed for 'gh release upload'.
--- a/bin/collect_sizes.py
+++ b/bin/collect_sizes.py
@@ -1,6 +1,15 @@
 #!/usr/bin/env python3
 
-"""Collect firmware binary sizes from manifest (.mt.json) files into a single report."""
+"""Collect firmware binary sizes from manifest (.mt.json) files into a single report.
+
+Output schema (consumed by bin/size_report.py):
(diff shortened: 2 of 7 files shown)


thebentern:
All five review findings addressed in 08ae317:

1. **Gate fails closed** — removed `continue-on-error` + the empty-dir fallback from `size-budget-gate`, and `--enforce-budgets` itself now fails on every missing-data path: empty collected sizes, a budgeted env absent from the run, or a manifest lacking the budgeted metric. Report-only mode still renders those as `n/a` without failing (the informational job keeps working on partial data).
2. **Empty-sizes silent pass** — covered by the same change plus a dedicated test.
3. **RAM section matching** — `compute_ram_bytes()` now counts RISC-V small-data (`.sdata`/`.sbss`) and excludes ESP-IDF `.rtc.*` sections. Kept name-based rather than per-target: the ESP32-C3/C6 linker scripts actually fold small-data into `.dram0.data`, which was already matched, so the `.sdata`/`.sbss` names are a safety net for toolchains that emit them as separate output sections.
4. **Zero/negative budget crash** — `load_budgets()` rejects non-positive/non-integer budgets with a clear error (previously a `0` would have crashed the render with `ZeroDivisionError`, and a quoted `"113000"` would have been silently skipped — both fail loudly now); the percentage render keeps a defensive guard for direct callers.
5. **Comment length** — trimmed the heap-watermark comment to two lines.

`bin/test_size_scripts.py`: **27/27** — the two tests that codified the fail-open behavior are flipped to assert fail-closed, with new report-only counterparts, plus cases for empty-sizes-under-enforcement and malformed budgets (`0`, `-5`, `"113000"`, `true`).

PR #10900, opened 2026-07-06
Add MemAudit: per-subsystem heap accounting in the boot log

thebentern:
## Motivation

The 2.8.0 nRF52840 field reports of ~99% heap use had to be diagnosed by hand: reconstructing which subsystem owned how much heap from source, build flags, and guesswork, one report at a time. This PR turns that community heap breakdown into a first-class feature — every future report self-diagnoses straight from the serial log.

## What it does

A tiny fixed-size registry (`src/memory/MemAudit.h/.cpp`, namespace `memaudit`) that subsystems report their large long-lived allocations into, printed as a single log line at the end of `setup()` and again alongside the existing periodic `Heap free:` line:

```
INFO  | MemAudit[boot]: tmm=2500 warm=4000 pkthist=5824 nodedb=13440 msgstore=2200 pktpool(live)=3270 total=31234
```

- `add(tag, delta)` / `set(tag, bytes)` / `snapshot(out, max)` / `logBreakdown(when)`
- Tags are string literals, registered on first use (pointer compare fast path, `strcmp` fallback), up to 16 tags, no heap allocation in the registry itself
- Counters are relaxed 32-bit atomics: single instructions on ARM Cortex-M, no locks, safe from the one hot path (per-packet pool alloc/release)

## Instrumented subsystems

| Tag             | What it measures                                           | Where                          |
| --------------- | ---------------------------------------------------------- | ------------------------------ |
| `nodedb`        | hot NodeInfoLite vector (`MAX_NUM_NODES * sizeof`)         | `NodeDB::nodeDBSelfCare()`     |
| `satmaps`       | position/telemetry/environment/status satellite maps (rb-tree overhead estimated at 44 B/entry) | `NodeDB::enforceSatelliteCaps()` |
| `warm`          | warm node store entry array                                | `WarmNodeStore` ctor/dtor      |
| `pkthist`       | packet history records + hash index                        | `PacketHistory` ctor           |
| `tmm` / `tmm_ni`| TrafficManagement unified cache / NodeInfo payload cache   | `TrafficManagementModule` ctor/dtor |
| `msgstore`      | message text pool                                          | `MessageStore` pool init       |
| `display`       | TFT line + repaint chunk buffers                           | `TFTDisplay`                   |
| `pktpool(live)` | live in-flight `meshtastic_MeshPacket`s (the only hot-path counter; atomic add) | `MemoryPool.h` via an optional allocator tag set in `Router.cpp` |

## Cost

- **RAM:** 128 B static table (16 tags x 8 B, measured in the rak4631 map) + one tag pointer per packet-pool instance; ~600 B transient stack while formatting the log line
- **Flash:** ~340 B of MemAudit code in the rak4631 ELF (`logBreakdown` 192 B, `findOrRegister` 132 B, `set` 14 B, `add` inlined by LTO) plus the one-line call sites and tag literals — well under 1 KB total
- **Escape hatch:** `MESHTASTIC_MEM_AUDIT=0` compiles the whole thing out to inline no-op stubs (call sites need no `#ifdef`s). Default is on everywhere except STM32WL, which is already flash-tight and excluded by default.

## Verification

- New native suite `test_mem_audit` (add/set/snapshot math, tag reuse incl. cross-TU duplicate literals, null/unknown tags, table-full behavior); `test/native-suite-count` bumped 27 -> 28
- Full native suite in Docker: GREEN — `542 test cases: 542 succeeded` (28/28 suites, `test_mem_audit` PASSED)
- `pio run -e rak4631`: SUCCESS

## Summary by CodeRabbit

* **New Features**
  * Added a memory audit registry and tag-based memory accounting across multiple subsystems.
  * Added boot-time and periodic heap breakdown logging to surface per-area memory usage.
* **Bug Fixes**
  * Improved reliability of live memory accounting for pools, buffers, and message/text/packet caches.
* **Tests**
  * Added unit tests validating audit registration, accumulation, snapshots, and full-capacity behavior.
* **Chores**
  * Updated the WebAssembly build to include the memory audit sources.


thebentern:
**On hardware (rak4631, 120-node NodeDB):** the boot line renders exactly as designed and the arithmetic checks out against the ELF and live heap telemetry:

```
INFO | MemAudit[boot]: msgstore=2200 warm=4000 nodedb=11040 satmaps=17864 pkthist=4800 tmm=2500 total=42404
```

It already earned its keep twice during validation: it exposed that the satellite maps (`satmaps=17864`) are roughly double the prior estimate and now the largest single boot-time consumer, and it made the per-PR savings verifiable from a single serial line — which is precisely the field-report workflow this PR is for.

PR #10898, opened 2026-07-06
Right-size nRF52 heap tiers after 2.8.0 heap-exhaustion field reports

thebentern:
## Problem

Field reports on 2.8.0 (see Discord) show nRF52840 devices at **99% heap — 114/115 KB — within minutes of boot** with a client connected. On nRF52 `operator new` asserts on OOM, so an affected device is one allocation away from a reboot loop.
The actual growth came from the 2.8.0 cache tier ladders, which gave nRF52 the **largest** non-PSRAM tiers on the assumption that a BLE-only part has a roomy heap. The nRF52840 arena is ~125 KB (256 KB minus 24 KB SoftDevice minus ~110 KB static, and shared with the FreeRTOS task stacks), and boot-time caches were consuming ~30 KB of it.

## Changes (all one-line tier constants; no formats, no features removed)

| Change | Before → After | Saves (nRF52840) |
|---|---|---|
| TMM unified cache (`mesh-pb-constants.h`) | 1000 → **250** entries; unclassified fallthrough → 400 | 7.5 KB |
| Warm node store | 200 → **100**; non-XXAA fallthrough 320 → **100** | 4.0 KB (+12.8 KB latent) |
| `MESSAGE_HISTORY_LIMIT` | 20 → **10** (tier classic ESP32 already ships) | 2.2 KB |
| `MAX_RX_TOPHONE` | 32 → **16** → static packet pool 70 → 54 slots | ~6.6 KB .bss → heap |
| PacketHistory hash index (`nrf52.ini`) | off arch-wide; O(n) over 240 records is µs at LoRa rates | 1.0 KB |
| `OLEDDISPLAY_REDUCE_MEMORY` (`nrf52.ini`) | drops ThingPulse back buffer; the 5 TFT variants `-U` it (TFTDisplay.cpp needs `buffer_back`) | 1.0 KB (OLED) |
| T1000-E stale "for testing" TMM override | 1024 → tier default | 10.2 KB on that variant |

ESP32 targets are untouched (already right-sized by #10741). The TMM fallthrough change also lowers RP2040/RP2350 from 1000 → 400, matching the classic-ESP32 tier.

## Measured result (rak4631, `arm-none-eabi-size -A`)

- Heap arena (`.heap` section): **124,572 → 131,180 B** (+6,608 B from the static packet pool cut)
- Boot-time heap allocations: **−15.7 KB** (TMM −7.5K, warm −4K, message pool −2.2K, hash −1K, OLED −1K)
[...]

PR #10901, opened 2026-07-06
Add central memory-class ladder (MemClass.h) with fail-safe-small defaults

thebentern:
> Stacked on #10898 (base is that branch; GitHub will retarget to `develop` when it merges). Review just this PR's two-file diff.

## Problem

The 2.8.0 nRF52840 heap exhaustion (#10898) had a structural cause beyond the numbers: every RAM-sized cache picks its per-platform tier from its own chip `#ifdef` ladder, and every ladder's fallthrough default was its *largest* non-PSRAM tier. nRF52 was never explicitly named in the TMM ladder, so it silently inherited 1000-entry caches on a ~115 KB arena. Nothing stops the next feature from repeating this.

## Change

**New `src/memory/MemClass.h`** — a single `MESHTASTIC_MEM_CLASS` (`TINY` / `SMALL` / `MEDIUM` / `LARGE`) ranked by *usable app heap after platform overheads* (SoftDevice, WiFi+BLE stacks), which is why classic ESP32 shares a class with nRF52840. Two deliberate properties:

1. **Unclassified chips land in SMALL.** A new target boots with small caches until someone opts it up in one visible place — the structural fix for the fallthrough-to-largest bug.
2. **A per-class `MESHTASTIC_BOOT_CACHE_BUDGET`** is `static_assert`ed in `mesh-pb-constants.h` over the three big boot-allocated caches (TMM cache + warm store + packet history), so a future cache-adding PR that would blow a small platform's budget **fails to compile** instead of exhausting heap in the field. (Skipped where `MAX_NUM_NODES` is runtime-resolved: ESP32-S3 and portduino.)

The TMM cache, warm store, `MAX_RX_TOPHONE`, and `MAX_SATELLITE_NODES` ladders now key off the class. Branches pinned by something other than RAM stay explicit and say why (nRF52840's SoftDevice-shared arena → TMM 250; RP2040's warm.dat watchdog bound → warm 150). `MAX_NUM_NODES` intentionally stays separate — it's flash-shaped (nodes.proto vs LittleFS), not heap-shaped. All ladders remain `#ifndef`-guarded, so existing variant overrides keep working.

## No-op proof

Rebuilt one env per class on this branch vs its base and compared `size` output:

| Env | Class | text/data/bss |
|---|---|---|
| rak4631 | SMALL (nRF52840) | **byte-identical** |
| tbeam | SMALL (classic ESP32) | **byte-identical** |
| rak11310 | SMALL (RP2040) | **byte-identical** |
| wio-e5 | TINY (STM32WL) | **byte-identical** |

MEDIUM/LARGE (S3/C6/P4, PSRAM, portduino) are covered by the value-by-value mapping in the diff (each class value equals the branch it replaced) and by CI.

## Verification

- Docker native suite: **536/536 passed**
- Builds green: rak4631, tbeam, rak11310, wio-e5

Follow-up candidates deliberately *not* changed here (kept historical values, flagged inline): `MESSAGE_HISTORY_LIMIT` normalization, ESP32-C3's `MAX_RX_TOPHONE=32`.

## Summary by CodeRabbit

* **New Features**
  * Improved memory-aware sizing across devices by introducing a compile-time memory class ladder and tier-based cache budgeting.
  * Added stricter packet history limits derived from node count to keep mesh caching within the configured boot-cache budget.

* **Bug Fixes**
  * Reduced memory-related instability by enforcing compile-time size consistency and validating packet history sizing against the new caps.
  * Enabled larger display buffers on select nRF52840 variants by unsetting the display memory reduction flag.


thebentern:
Both review findings addressed in 85a3c2f, and the merge conflict is resolved (the base #10898 landed on develop as a squash, which is what tripped this stacked branch — develop was merged in; its `mesh-pb-constants.h` is byte-identical to this branch's base, so the ladder is unchanged):

1. **RP2350 class-table doc** — the table now matches the ladder: RP2350 rides with RP2040 in `SMALL` (deliberately, to keep this header a behavioral no-op), with an inline note that it's a `MEDIUM` candidate whenever someone wants to tune it up.
2. **PacketRecord "pinned by static_assert"** — the assert the budget comment referenced didn't actually exist; it does now (`static_assert(sizeof(PacketRecord) == 20, …)` in `PacketHistory.h`), so a future layout change breaks the build instead of silently skewing the budget math.

Verified: rak4631 and wio-e5 build green post-merge; `test_packet_history` 47/47.

thebentern:
CodeRabbit feedback addressed in 08476c2:

1. **`PACKETHISTORY_MAX` hoisted** (the actionable finding) — the definition now lives in `mesh-pb-constants.h` next to the `MAX_NUM_NODES` it derives from, and both consumers use it: PacketHistory's constructor clamp and the boot-cache budget `static_assert`. No more hand-mirrored arithmetic to drift. The expression remains valid on targets where `MAX_NUM_NODES` resolves at runtime (ESP32-S3 flash-size probe, portduino config) — verified by building heltec-v3 alongside rak4631 — and the old `* 2.0` double math is now plain integer.
2. **Comment-length nitpicks** — MemClass.h's header prose cut from 36 to 16 lines and the sizing-policy/budget comments trimmed to guideline size. I kept the class table, the fail-safe-small rule, the override mechanism, and the include-order constraint in the header deliberately: they're the API contract a contributor needs at the point of use; the 2.8.0 backstory now lives only in the commit history.

Verified: rak4631 + heltec-v3 build green, `test_packet_history` 47/47.

PR #10902, opened 2026-07-06
BLE lifecycle memory: true BLE-off on nRF52, NimBLE host pools to PSRAM on S3

thebentern:
> **Draft: needs on-hardware validation before review** (checklist below). Part of the heap-recovery series (#10898, #10901, #10900, #10899).

## nRF52: being "off" should be free

When `config.bluetooth.enabled == false`, the firmware previously **fully initialized SoftDevice + Bluefruit and then muted advertising** — the #4055-era workaround ("device hangs several days after boot", root cause never found). Being off cost ~3.5 KB of Bluefruit heap plus ~6 KB of BLE/SOC FreeRTOS task stacks, out of a ~115 KB arena.

This PR leaves the SoftDevice entirely off in that case:
- `setBluetoothEnable()` short-circuits and clears `useSoftDevice`, so `checkSDEvents()` stops polling.
- The deep-sleep `GPREGRET` writes (`DFU_MAGIC_SKIP`) now mirror `lfs_assert()`'s existing pattern: try `sd_power_gpregret_*`, fall back to a direct `NRF_POWER->GPREGRET` write when the SD isn't enabled.
- `NRF52Bluetooth::startDisabled()` is removed. `clearBonds()` keeps its lazy bring-up for explicit bond maintenance, and re-enabling bluetooth applies via reboot, so no runtime re-init state machine is needed.

**Why this is safe:** with the SD disabled, the Adafruit BSP flash driver checks `sd_softdevice_is_enabled()` per operation and takes its **synchronous** path — the exact path every boot already uses for all filesystem work before `Bluefruit.begin()` runs. `HardwareRNG` uses the CC310, not the SD. `sd_*` calls with the SD disabled return an error code rather than faulting (the `lfs_assert` fallback has relied on this for years). The failure mode this deliberately revisits is the un-root-caused #4055 hang — hence draft + soak test below, and the change is a one-branch revert if the ghost reappears.

*(Note: the nrf54l15 platform carries its own copy of the `startDisabled()` workaround — left untouched here, flagged as follow-up.)*

## ESP32-S3 + PSRAM: NimBLE host pools out of internal DRAM

[...]

PR #10903, opened 2026-07-06
nrf52840: right-size the SoftDevice RAM reservation (+8 KB heap arena)

thebentern:
Part of the heap-recovery series (#10898, #10901, #10900, #10899, #10902).

## Problem

Both nrf52840 linker scripts hard-code the app RAM ORIGIN at `0x20006000` — a 24 KB SoftDevice reservation dating to their introduction. `sd_ble_enable()` computes the *actual* requirement from the Bluefruit configuration (one peripheral link, `BANDWIDTH_MAX`/MTU 247, `0x1000` attribute table) and it is substantially below that. The SoftDevice can't use the gap and the app is linked above it, so the difference is pure waste — on the platform where 2.8.0 field reports show the heap arena (which on this BSP is exactly the linker gap) at 99% use.

## Change

- `RAM ORIGIN 0x20006000 → 0x20004000` in both `nrf52840_s140_v6.ld` and `nrf52840_s140_v7.ld`.
- `Bluefruit.begin()`'s return value — previously discarded — is now checked: if a future SoftDevice/Bluefruit config bump raises the requirement past the reservation, the node records a **critical error with instructions** instead of silently running without BLE.

## Measured effect

**+8,192 B of heap arena on every nRF52840 board**, BT on or off:

| Build | `.heap` section | `.data` addr |
|---|---|---|
| rak4631 (S140 v6 script) develop | 124,572 B | 0x20006000 |
| rak4631 this PR | **132,764 B** | 0x20004000 |
| tracker-t1000-e (S140 v7 script) this PR | links clean | 0x20004000 |

## Hardware validation (rak4631, S140 6.1.1)

Measured the actual SoftDevice requirement via `CFG_DEBUG=1`:

```
[CFG] SoftDevice's RAM requires: 0x200038A0
```

| | Address | vs. SD floor |
|---|---|---|
| SoftDevice requirement (floor) | `0x200038A0` | — |
| Old ORIGIN | `0x20006000` | +10,080 B wasted |
| **This PR** | `0x20004000` | **+1,888 B (1.84 KB) margin** ✅ |

`0x20004000` reclaims the full +8 KB while keeping a ~1.8 KB margin for SD-version/config variance across the v6/v7 fleet. (Measurement also ruled out tightening to `0x20003800`, which would sit **below** the 0x200038A0 floor.) BLE bring-up confirmed on the new base: `Init Bluefruit → services (DIS/BAS/Mesh) → Advertise`, `Bluefruit.begin()` returns success (the new critical-error guard never fires), client reconnects over serial; a v7-script board links at the new base too.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bluetooth startup reliability by detecting initialization failures early and stopping setup cleanly when Bluetooth cannot start.
  * Updated device memory settings to better match Bluetooth requirements, which should help avoid startup issues and improve available memory headroom.

### DIFF
--- a/src/platform/nrf52/NRF52Bluetooth.cpp
+++ b/src/platform/nrf52/NRF52Bluetooth.cpp
@@ -4,6 +4,7 @@
 #include "HardwareRNG.h"
 #include "PowerFSM.h"
 #include "configuration.h"
+#include "error.h"
 #include "main.h"
 #include "mesh/PhoneAPI.h"
 #include "mesh/mesh-pb-constants.h"
@@ -275,7 +276,16 @@ void NRF52Bluetooth::setup()
     LOG_INFO("Init the Bluefruit nRF52 module");
     Bluefruit.autoConnLed(false);
     Bluefruit.configPrphBandwidth(BANDWIDTH_MAX);
-    Bluefruit.begin();
+    if (!Bluefruit.begin()) {
+        // sd_ble_enable() rejected our RAM base: the linker RAM ORIGIN
+        // (src/platform/nrf52/nrf52840_s140_v*.ld) is below what the SoftDevice needs for the
+        // current Bluefruit config. Without this check the node would silently run without BLE.
+        // Rebuild with -DCFG_DEBUG=1 to get "SoftDevice's RAM requires: 0x..." in the log, then
+        // raise the ORIGIN accordingly.
+        LOG_ERROR("Bluefruit.begin failed - SoftDevice RAM reservation too small for this config");
+        RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_UNSPECIFIED);
+        return;
+    }
     // Clear existing data.
     Bluefruit.Advertising.stop();
     Bluefruit.Advertising.clearData();
--- a/src/platform/nrf52/nrf52840_s140_v6.ld
+++ b/src/platform/nrf52/nrf52840_s140_v6.ld
@@ -21,8 +21,17 @@ MEMORY
    * - Max ATT MTU
    * - Concurrent connection peripheral + central + secure links
    * - Event Len, HVN queue, Write CMD queue
+   *
+   * With our fixed Bluefruit config (1 peripheral link, BANDWIDTH_MAX / MTU 247,
+   * 0x1000 attr table) sd_ble_enable() reports a requirement well below the old
+   * 0x20006000 ORIGIN; every byte of gap is unusable RAM. 0x20004000 keeps a
+   * ~2+ KB margin over the reported base. NRF52Bluetooth::setup() now checks
+   * Bluefruit.begin() and records a critical error if the SoftDevice ever
+   * rejects this base (e.g. after an SD/Bluefruit config change) - re-measure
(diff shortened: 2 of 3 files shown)


thebentern:
**Hardware validation (rak4631, S140 6.1.1, v6 linker script):**

- ✅ `sd_ble_enable()` **accepts the 0x20004000 base**: Bluefruit init, GATT services, PIN pairing setup, and advertising all come up normally; the new `Bluefruit.begin()` failure check never fires. BLE also re-initializes cleanly after being disabled and re-enabled via config.
- ✅ `dbgHeapTotal` on-device rose **124,572 → 139,236 with the sizing PR stack** (this PR contributes exactly +8,192; `.data` links at 0x20004000, confirmed in the ELF).
- ⚠️ The exact `SoftDevice's RAM requires: 0x…` figure is still uncaptured — `-DCFG_DEBUG=1` did not take effect through either flash path (no Bluefruit config dump in the boot stream both attempts). Since begin() success already bounds the requirement ≤ 0x20004000 with our BANDWIDTH_MAX config, I'd treat the exact value as part of the future tightening pass rather than a blocker. A v7-script board (t1000-e) also links at the new base but hasn't been booted on hardware yet.

Remaining before undraft: phone pairing smoke (FIXED_PIN/RANDOM_PIN + config download) and one v7-board boot.

thebentern:
**Measured on hardware — the outstanding item is closed.** rak4631 (S140 6.1.1, v6 linker script), `CFG_DEBUG=1`:

```
[CFG] SoftDevice's RAM requires: 0x200038A0
```

Why it wasn't showing up before: on nRF52 this firmware overrides `printf` to route to **SEGGER RTT** (J-Link channel), and the BSP's `LOG_LV1`/`ADALOG` uses that same `printf` — so the line was emitted correctly all along, just to a debugger I don't have wired up, not USB serial. Captured it by temporarily mirroring the BSP printf to USB CDC (throwaway, reverted).

**The number validates the choice and corrects the draft's speculative target:**

| | Address | vs. SD floor (0x200038A0) |
|---|---|---|
| SoftDevice requirement (floor) | `0x200038A0` | — |
| Old ORIGIN | `0x20006000` | +10,080 B wasted |
| **This PR** | `0x20004000` | **+1,888 B (1.84 KB) margin** ✅ |
| Draft's speculative "tighten to" | `0x20003800` | **−160 B — below the floor, would fail** ❌ |

So `0x20004000` is confirmed as the right, safe choice: it reclaims the full +8 KB, keeps a healthy ~1.8 KB margin for SD-version/config variance across the v6/v7 fleet, and the earlier speculative tighten to `0x20003800` is now ruled out. (There's ~1 KB more theoretically available at `0x20003C00`/864 B margin, but that thin a margin isn't worth it fleet-wide — noting it, not doing it.)

**BLE bring-up on the 0x20004000 build:** `Init Bluefruit → services (DIS/BAS/Mesh) → Advertise`, `Bluefruit.begin()` returns success (the new critical-error guard never fires), and the client reconnects over serial. On a v7-script board the image also links at the new base (verified earlier). Taking this out of draft.

PR #10906, opened 2026-07-06
fix(power): correct charge/power detection preprocessor bugs

ndoo:
## Summary

1. `isCharging()`'s `#elif EXT_CHRG_DETECT` tests the macro's value, not whether it's defined. Pin numbering is architecture-defined (`PA0`=0 on STM32, `P0.00`=0 on NRF52, `GPIO0`=0 on ESPx), so any board wiring CHRG to pin 0 silently loses this check. Fixed with `defined(...)`.

2. `isVbusIn()` trusts `EXT_PWR_DETECT` alone as "power present," but the macro name isn't an enforced contract — some boards wire it to a charge-complete pin instead, so it reads false for the entire time a device is actively charging. A charger IC's CHRG output can only assert while drawing input power, so falling back to `EXT_CHRG_DETECT` when defined is always a valid extra check.

## Regression scope

- Fix 1 does not affect any current variants, as no variant currently uses a 0-equivalent pin for `EXT_CHRG_DETECT`.
- Fix 2 only changes behavior for the three boards defining both macros (`ELECROW-ThinkNode-M6`, `tracker-t1000-e`, `seeed_mesh_tracker_X1`), and only their standalone `isVbusIn()`/`usbPowered` — `isCharging()` is unchanged for all three (M6's own `CHRG || isVbusIn()` simplifies algebraically to the same result; the other two don't call `isVbusIn()` from `isCharging()` at all). The widening makes "power present" also recognize non-USB charging (e.g. solar), which it previously missed.

## Testing

Build-verified: `rak3172`, `russell`, `tracker-t1000-e`, `thinknode_m4`, `thinknode_m6`. Hardware-verified (flash/boot/charge detection) on a new RAK3172-based board using `EXT_CHRG_DETECT=PA0`.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other: build-only for the boards above. Requesting community testing on ThinkNode M6/tracker-t1000-e/seeed_mesh_tracker_X1 (the only boards where `isVbusIn()` behavior changes).


ndoo:
Hmm, this has the rather annoying side effect that 101% gets reported while the device is charging, so now we don’t know the actual % while charging.

I’ll have to raise a PR to not report 101% while charging.

PR #10911, opened 2026-07-07
Add Meshnology W10 AIOT Dev Kit (SX1262 via MCP23017 I2C expander)

thebentern:
## Summary

Adds support for the **Meshnology W10 "LoRa AIOT Dev Kit"** — an ESP32-S3R8 board with an EBYTE E22-900MM22S (SX1262) radio, AXP2101 PMIC, Quectel L76KB GPS, PCF85063 RTC, SHT41 + QMI8658 sensors, and an SPI TFT. `MESHNOLOGY_W10 = 140` already exists in the `HardwareModel` enum.

The interesting part of this board is that the SX1262's control lines (**NRESET, DIO1, BUSY**) and the LCD reset are wired to an **MCP23017 I2C GPIO expander** rather than native ESP32 GPIOs — and the expander's `/INT` is not routed to the MCU, so there is no hardware DIO1 interrupt available. Two reusable pieces handle that, split into their own commit:

1. **`Add software LoRa IRQ polling and an MCP23017 expander HAL …`**
   - `ExtensionIOMCP23017` + `MCP23017LockingArduinoHal`: a RadioLib HAL that maps virtual pins (`MCP23017_VPIN_BASE … +15`) onto the expander's GPIO, so RESET/BUSY/DIO1 become I2C transactions while the real SPI pins (SCK/MOSI/MISO/CS) pass straight through.
   - `LORA_DIO1_SOFTWARE_POLL`: with no DIO1 interrupt, the SX126x interface polls the radio IRQ status register from the radio thread and synthesizes the `ISR_TX`/`ISR_RX` events, filtering noisy preamble/header IRQs so they can't starve TX. A 1 ms `ISR_POLL_TICK` drives the poll; TX timers may overwrite the pending tick.
   - **Everything is gated behind `USE_MCP23017` / `LORA_DIO1_SOFTWARE_POLL`; boards that don't define them are byte-for-byte unaffected** (`txTimerOverwrite` defaults to `false`, the normal `setDio1Action` interrupt path is unchanged).

2. **`Add Meshnology W10 AIOT Dev Kit variant …`** — the variant + a one-line `HW_VENDOR` mapping.

## Hardware verification

Brought up on a physical W10 AIOT Dev Kit (US / LongFast):

- **AXP2101 PMU** init, **MCP23017** at 0x20 (`GPIOA=0xEA`, A0=LOW)
- **SX1262**: `init success`, then a **full over-the-air TX/RX round trip** (`Received an implicit ACK`) — RESET/DIO1/BUSY through the expander, DIO1 emulated by the IRQ-register poll
- **PCF85063 RTC**, **SHT41** (0x44) and **QMI8658 IMU** (0x6B) auto-detected; **ES8311** codec at 0x18
- **L76KB GPS** fix; board reports **HardwareModel 140 (MESHNOLOGY_W10)**

The vendor's own Arduino demo firmware can't complete a LoRa TX on this same unit (its bit-banged driver finds the E22 control lines through the expander marginal); RadioLib via `MCP23017LockingArduinoHal` handles it reliably.

## Notes

- The expander is an **MCP23017** despite the V1.1 schematic labelling it "TCA9555" — the V1.2 placement diagram and all of the vendor's firmware/demo code use the MCP23017 register map (address 0x20).
- The E22-900MM22S has no TCXO and does RF switching in hardware (DIO2→TXEN), so no TXEN/RXEN GPIOs and no `SX126X_DIO3_TCXO_VOLTAGE`.
- `board = esp32-s3-devkitc-1` with a local `pins_arduino.h` that drops `RGB_BUILTIN` (it otherwise pulls the Arduino RMT HAL into the link and fails against this build's trimmed FreeRTOS config).
- The SPI TFT panel is the one thing I couldn't visually verify from the bench — init runs cleanly and the variant defaults to the ST7789 1.54" 240×240 (build `-D MESHNOLOGY_W10_LCD_ST7796_35` for the 3.5" ST7796).

- **Audio**: the ES8311 codec + NS4150 amp + speaker are wired up for notification tones / ringtones over the I2S buzzer path (`use_i2s_as_buzzer`), the same way the sub-GHz M5 Cardputer uses its ES8311 — verified on hardware (a test tune played through the speaker). Codec2 voice is SX1280-only, so it does not apply here.

## For reviewers (shared-core touch points)

The software-poll path is fully behind `USE_MCP23017` / `LORA_DIO1_SOFTWARE_POLL`; boards keeping the hardware DIO1 interrupt are unaffected (`txTimerOverwrite` defaults to `false`, `setDio1Action` path unchanged). One known corner, documented in the code: because the poll tick and the TX-delay timer share the single `NotifiedWorkerThread` slot, a TX queued while the radio is actively receiving can delay poll servicing until `pollMissedIrqs()` (~1 s) recovers it — bounded added latency under contention, not a lost event. Happy to restructure that if you'd prefer a dedicated timer.

## Summary by CodeRabbit

* **New Features**
  * Added Meshnology W10 ESP32-S3 support, including MCP23017 I/O expander bring-up during boot and MCP23017-based LoRa control wiring.
  * Introduced software-poll scheduling for LoRa DIO1 interrupt handling to better manage transmit-heavy timing.
  * Extended W10 audio startup with amplifier enable control and enhanced ES8311/codec initialization.
* **Bug Fixes**
  * Improved LoRa interrupt processing to reduce TX scheduling starvation and maintain correct RX terminal-event behavior.
  * Updated light-sleep wake handling to match extended/virtual DIO1 configurations.


thebentern:
Thanks for the review — addressed the feedback in the latest commit (rebased onto the current `develop` merge):

- **MCP23017 concurrency** (major): `ExtensionIOMCP23017` now serializes register access with a `concurrency::Lock`, since the radio HAL (BUSY/DIO1/RESET) and `AudioThread` (amp enable) reach the expander from different threads and the read-modify-write paths aren't atomic.
- **`enablePinChangeInterrupt`** (major): switched to the checked `readReg(reg, val)` overload and skip-on-failure, matching the other RMW helpers so a glitched read can't clobber the rest of the bank.
- **`digitalRead` fail-safe** (minor): returns HIGH on a failed read instead of LOW, so a transient I2C error on the LoRa BUSY line can't read as "ready" and let RadioLib start an SPI transaction early. (DIO1 is polled via the radio IRQ register, not this pin.)
- **`es8311_write_reg`** (minor): logs a warning if a codec register write NACKs instead of silently leaving the ES8311 half-configured.
- **`RadioInterface.cpp`** (minor): guarded the MCP23017 HAL branch with `ARCH_ESP32` to match the include and driver-file guards.

Builds clean; the MCP23017 radio path and the ES8311 audio were re-verified on hardware.

PR #10917, opened 2026-07-07
Firmware for Seeed XIAO ESP32C6 and WIO-SX1262 for XIAO

reb1995:
Confirmed LED blinking as expected, TX and RX working on both public channels and DMs. Working over bluetooth connection and WiFi. I2C confirmed working with oled screen.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Seeed ESP32C6 XIAO device with Wio SX1262. Tested working with communication between my C6 and my RAK devices.

## Summary by CodeRabbit

* **New Features**
  * Added support for the Seeed XIAO C6 (ESP32-C6) board with updated build, debug, and upload defaults.
  * Included board-specific pin mapping for USB identification, LED/button, battery sensing, I2C, and SPI.
  * Enabled LoRa support (SX1262) with board-appropriate control, RF switch, and TX/RX behaviors; Wi-Fi and Bluetooth are supported as well.


reb1995:
Working I2C test.
image

Working tx/rx.
image

Device details.
image

thebentern:
Looks good. Can you please run trunk fmt for our linting?

reb1995:
> Looks good. Can you please run trunk fmt for our linting?

`trunk fmt` was run. Collapsed some of the arrays into single lines. Amended my commit and pushed it back up. Please let me know of any other issues.

reb1995:
Build seems to be failing due to a missing partition scheme csv. The app size, ~2.5MB, will not fit in the 4MB flash with OTA enabled. In my local branch, I was using a default huge_app.csv partition scheme which had 3MB for the app. This wasn't included in the repo and the CI/CD didn't have the partition. Sorry for my oversight.

As a solution, we could add a new partition for the 4MB devices that doesn't include OTA capabilities. I have tested the following partition table to build and work. Not sure if it is something that you want to do. I'll leave that up to you all if that is worth doing or needs to be done before this PR. 

```
RAM:   [==        ]  21.3% (used 111928 bytes from 524288 bytes)
Flash: [========  ]  81.8% (used 2574396 bytes from 3145728 bytes)
```

```
partition-table-4MB-no-OTA.csv 
# Can be used by devices with only 4 MB of flash. 
# Allows for building of and flashing of firmware on space constained devices.
# Will, obviously, prevent OTA updates.
# Name, Type, SubType, Offset, Size, Flags
nvs, data, nvs, 0x9000, 0x5000,
otadata, data, ota, 0xe000, 0x2000,
app0, app, ota_0, 0x10000, 0x300000,
spiffs, data, spiffs, 0x310000, 0xF0000,
```

thebentern:
> Build seems to be failing due to a missing partition scheme csv. The app size, ~2.5MB, will not fit in the 4MB flash with OTA enabled. In my local branch, I was using a default huge_app.csv partition scheme which had 3MB for the app. This wasn't included in the repo and the CI/CD didn't have the partition. Sorry for my oversight.
> 
> As a solution, we could add a new partition for the 4MB devices that doesn't include OTA capabilities. I have tested the following partition table to build and work. Not sure if it is something that you want to do. I'll leave that up to you all if that is worth doing or needs to be done before this PR.
> 
> ```
> RAM:   [==        ]  21.3% (used 111928 bytes from 524288 bytes)
> Flash: [========  ]  81.8% (used 2574396 bytes from 3145728 bytes)
> ```
> 
> ```
> partition-table-4MB-no-OTA.csv 
> # Can be used by devices with only 4 MB of flash. 
> # Allows for building of and flashing of firmware on space constained devices.
> # Will, obviously, prevent OTA updates.
> # Name, Type, SubType, Offset, Size, Flags
> nvs, data, nvs, 0x9000, 0x5000,
> otadata, data, ota, 0xe000, 0x2000,
> app0, app, ota_0, 0x10000, 0x300000,
> spiffs, data, spiffs, 0x310000, 0xF0000,
> ```

Try excluding modules like Audio and webserver?

reb1995:
> > Build seems to be failing due to a missing partition scheme csv. The app size, ~2.5MB, will not fit in the 4MB flash with OTA enabled. In my local branch, I was using a default huge_app.csv partition scheme which had 3MB for the app. This wasn't included in the repo and the CI/CD didn't have the partition. Sorry for my oversight.
> > As a solution, we could add a new partition for the 4MB devices that doesn't include OTA capabilities. I have tested the following partition table to build and work. Not sure if it is something that you want to do. I'll leave that up to you all if that is worth doing or needs to be done before this PR.
> > ```
> > RAM:   [==        ]  21.3% (used 111928 bytes from 524288 bytes)
> > Flash: [========  ]  81.8% (used 2574396 bytes from 3145728 bytes)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > partition-table-4MB-no-OTA.csv 
> > # Can be used by devices with only 4 MB of flash. 
> > # Allows for building of and flashing of firmware on space constained devices.
> > # Will, obviously, prevent OTA updates.
> > # Name, Type, SubType, Offset, Size, Flags
> > nvs, data, nvs, 0x9000, 0x5000,
> > otadata, data, ota, 0xe000, 0x2000,
> > app0, app, ota_0, 0x10000, 0x300000,
> > spiffs, data, spiffs, 0x310000, 0xF0000,
> > ```
> 
> Try excluding modules like Audio and webserver?

Sure. I can try that. Will report back with my findings.

reb1995:
Did quite a bit of test compiles and measurements (~15 different builds). Disabling the audio doesn't seem to change the size at all. Finally cut enough by disabling the Webserver and WiFi components to get it small enough to be able to OTA (~1.70 MB), however the ESP32C6 doesn't do bluetooth OTAs. So with no WiFi it wouldn't be able to OTA at all.

I think I was overthinking this and I think the only way to do this is to go forward is with a non OTA 4MB ESP32 partition table. 

Not sure if this is something you guys want to do, but the partition table I posted earlier is working for the ESP32C6. I believe it would also work for the Seeed ESP32C3 (and other C3/C6's) as well. If this gets accepted, I'd be happy to write and test configurations for that.

Let me know if we want to go the non OTA route with a new partition table or if we want to close without merging. Or maybe there's a 3rd option I'm missing. 

Sorry again to run in circles on this. I thought it would be a simple win but maybe missed the mark.

| Build Configuration | Binary Size (Bytes) | Fits OTA Partition? (Max ~1.70MB) | Includes WiFi | OTA Result |
|---|---|---|---|---|
| Baseline (Fully Featured) | 2,574,396 | No | Yes | Fails (Binary too large) |
| Exclude Webserver | 2,358,962 | No | Yes | Fails (Binary too large) |
| Exclude GPS | 2,506,048 | No | Yes | Fails (Binary too large) |
| Exclude SCREEN | 2,378,610 | No | Yes | Fails (Binary too large) |
| Exclude SCREEN, GPS | 2,309,748 | No | Yes | Fails (Binary too large) |
| Exclude WIFI | 1,697,334 | Yes | No | Fails (No network to receive payload) |
| Exclude WIFI, Webserver | 1,697,070 | Yes | No | Fails (No network to receive payload) |


Proposed 4MB flash partition table
```
partition-table-4MB-no-OTA.csv 
# Can be used by devices with only 4 MB of flash. 
# Allows for building of and flashing of firmware on space constained devices.
# Will, obviously, prevent OTA updates.
# Name, Type, SubType, Offset, Size, Flags
nvs, data, nvs, 0x9000, 0x5000,
otadata, data, ota, 0xe000, 0x2000,
app0, app, ota_0, 0x10000, 0x300000,
spiffs, data, spiffs, 0x310000, 0xF0000,
```

AlexandrZloy:
> Proposed 4MB flash partition table
> 
> ```
> partition-table-4MB-no-OTA.csv 
> # Can be used by devices with only 4 MB of flash. 
> # Allows for building of and flashing of firmware on space constained devices.
> # Will, obviously, prevent OTA updates.
> # Name, Type, SubType, Offset, Size, Flags
> nvs, data, nvs, 0x9000, 0x5000,
> otadata, data, ota, 0xe000, 0x2000,
> app0, app, ota_0, 0x10000, 0x300000,
> spiffs, data, spiffs, 0x310000, 0xF0000,
> ```

Hello, maybe the default [huge_app.csv](https://github.com/espressif/arduino-esp32/blob/master/tools/partitions/huge_app.csv) partition table will fit, instead of creating a custom one.

Simplycissmus:
I have a hand-wired XIAO ESP32-C6 with an SX1262-class module on the bench, so I went through this one pin by pin against the tree. Two things I would change, and a list of what I checked and found sound - so nobody has to redo it.

**`HAS_GPS` is not overridden, and the ESP32 default is 1.**

`src/platform/esp32/architecture.h:22` sets `#define HAS_GPS 1` for the whole architecture, and `main.cpp:1046` gates GPS setup on it. The Wio-SX1262 carrier has no GNSS, so this variant boots into the full probe sequence - the common baud table, then `GPS_PROBETRIES` rounds, then the rare table - looking for a receiver that is not on the board. `variants/esp32s3/seeed_xiao_s3` sets `HAS_GPS 1` deliberately because that one can carry a GPS; here `#define HAS_GPS 0` is the one line that avoids the whole detour. (Related: #11470 is a bounds fix in exactly that probe path.)

**`BUTTON_PIN 0` is the outlier for a C6.**

Both C6 variants already in the tree use pin 9 - `variants/esp32c6/heltec_rcc6/variant.h:5` actively, `m5stack_unitc6l` commented out at the same number - and 9 is the C6's own boot/strapping pin, which is why neither needs `BUTTON_NEED_PULLUP`. My hand-wired board uses 9 as well. Is D0 an actual button on the Wio-SX1262 carrier? If it is, this is right and worth a comment saying so, because it reads like a typo otherwise. If it is not, the button silently never triggers: `BUTTON_NEED_PULLUP` keeps the input high forever and nothing ever presses it.

**Checked and sound**, in case it saves a reviewer the trip:

- `SX126X_RXEN 23` together with `SX126X_DIO2_AS_RF_SWITCH` is supported, not a conflict: `SX126xInterface.cpp` calls `setDio2AsRfSwitch(true)` and then `setRfSwitchPins(SX126X_RXEN, SX126X_TXEN)` unconditionally, and the comment right above says `RADIOLIB_NC` on the TX side is a no-op. DIO2 drives the TX path, pin 23 the RX side.
- `BATTERY_PIN -1` matches the existing convention - `variants/esp32s3/seeed_xiao_s3/variant.h:39` and four other variants do the same for boards without a divider.
- The SPI map uses the standard XIAO pins (D8 SCK / D9 MISO / D10 MOSI), which is what a stacking carrier should use, and every LoRa pin stays inside the D0-D10 set.
- `board_level = pr` and `board_check = true` mean it actually gets built in the PR matrix rather than only in `extra`.

Simplycissmus:
Following up on the partition question from July, because the numbers have moved and the premise deserves a re-check against today's tree.

The July discussion assumed OTA needs the application to fit twice. Meshtastic's 4 MB scheme does not work that way: the default `partition-table.csv` (set for every ESP32 variant by `variants/esp32/esp32-common.ini`) gives the app `0x250000` (2,424,832 bytes) in `ota_0`, a 640 KB `flashApp` slot in `ota_1` that holds the small OTA *updater* rather than a second image, and 1 MB spiffs. The updater receives the new firmware and writes it into `ota_0`.

**Does a C6 build fit that? Yes, today, measurably.** `tlora-c6` - the in-tree 4 MB C6, `board_level = pr` - builds against exactly that table in last night's develop run:

```
Flash: [==========]  96.5% (used 2340860 bytes from 2424832 bytes)
```

(run 33460192851, job `build (tlora-c6, esp32c6)`.) So the ~2.5 MB measured here in July either predates some slimming or was read against a different limit; today the standard table closes with about 82 KB to spare.

To be fair to both sides of the trade:

- **No C6 OTA exists today either way.** The release pipeline pulls the updater from `meshtastic/firmware-ota`, whose latest release ships `firmware.bin`, `firmware-c3.bin`, `firmware-s3.bin` - no C6. So `huge_app.csv` costs nothing *right now*.
- **But a partition table is per-device, forever.** A XIAO C6 flashed with `huge_app.csv` can never adopt OTA without coming back to a USB port, even after a C6 updater appears - and the manifest tooling already looks for `mt-esp32c6-ota.bin` by name, so the slot in the naming scheme exists. Units on the standard table pick it up the day the image does.
- **82 KB headroom is thin.** That cuts the other way: at 96.5% the in-tree C6 is close to the cliff regardless of this PR, and if the firmware grows past it, both boards have the same problem at the same time.

Concrete suggestion that settles it without anyone arguing from memory: drop the `board_build.partitions = huge_app.csv` line for one push and let CI print this variant's real size against the default table. If it fits like its sibling does, the override can stay gone and the board keeps the same future every other 4 MB ESP32 has; if it does not, the thread finally has the number that justifies the exception.

PR #10928, opened 2026-07-07
Add chunked PhoneAPI transport payloads

kernel-oops:
## Summary

This adds a compact chunked PhoneAPI transport so clients with a 20-byte GATT payload limit can exchange complete Meshtastic client API protobufs without ATT long reads or writes.

The immediate use case is Garmin Connect IQ BLE. Garmin limits application writes to 20 bytes and does not provide ATT long reads, while normal `ToRadio` and `FromRadio` messages can be substantially larger.

Depends on the corresponding protobuf schema PR: https://github.com/meshtastic/protobufs/pull/980

## What changed

- Adds a distinct `ClientApiChunkedPayload` transport wrapper with `uint32 payload_size = 1` and `bytes payload_chunk = 2`.
- Adds `ToRadio.chunked_payload = 8` and `FromRadio.chunked_payload = 20`.
- Keeps the existing four-field `ChunkedPayload` and `ChunkedPayloadResponse` messages unchanged for compatibility with their current users.
- Reassembles incoming chunked `ToRadio` messages before passing the original serialized message to the existing `PhoneAPI` handler.
- Emits chunked `FromRadio` messages only after that client has successfully completed a chunked `ToRadio` transfer. Legacy clients retain the existing unchunked behavior.
- Tracks inbound and outbound state for two simultaneous `PhoneAPI` clients in one lock-protected static table.
- Reclaims incomplete incoming transfers after 30 seconds and clears all state when a client disconnects.
- Regenerates the nanopb `mesh.pb.cpp` and `mesh.pb.h` sources for the schema change.

## Wire format

The transport assumes ordered, reliable delivery on one client connection. This matches Garmin's serialized BLE writes-with-response and reads, and avoids spending bytes on transfer IDs and chunk indices in every packet.

Only the first chunk carries the total serialized payload size. Every chunk carries the next payload bytes.

For a 20-byte GATT payload:

- `ToRadio` first chunk: up to 13 payload bytes.
- `ToRadio` continuation: up to 16 payload bytes.
- `FromRadio` first chunk: up to 12 payload bytes.
- `FromRadio` continuation: up to 15 payload bytes.

The `FromRadio` encoder writes the fixed canonical protobuf wire representation directly. Boundary tests compare it byte-for-byte with nanopb output.

## Safety and compatibility

- The existing `ChunkedPayload` wire schema remains unchanged.
- Chunking is negotiated per connection; clients that do not opt in continue receiving normal `FromRadio` messages.
- Incoming first chunks are rejected when the declared size exceeds `MAX_TO_FROM_RADIO_SIZE` or is smaller than the bytes already supplied.
- Continuations are rejected and their state cleared if they would exceed the declared total.
- Empty chunks, continuations without an active transfer, and clients beyond the bounded two-slot table fail closed.
- Reassembled messages are copied out of shared state and passed to `handleToRadio()` only after releasing the mutex.
- Outbound messages are size-checked before entering the fixed 512-byte buffer.
- Pending outbound chunks are drained before another normal `FromRadio` message is generated.

This is intentionally not a retransmission protocol. It relies on the ordered, reliable behavior of the underlying client connection.

## Flash footprint

Measured on `nrf52_promicro_diy_tcxo` against the same `develop` base:

| Build | Flash | RAM | Warm-store headroom |
| --- | ---: | ---: | ---: |
| `develop` | 800,816 B | 102,612 B | 2,000 B |
| Initial integrated chunk transport | 802,816 B | 104,724 B | 0 B |
| Final optimized chunk transport | 801,584 B | 104,708 B | 1,232 B |

The final patch adds 768 bytes of flash and 2,096 bytes of RAM relative to `develop`. Consolidating the client tables and locks, removing redundant state and chunk-only log strings, avoiding unnecessary buffer clearing, and directly encoding the small outbound wrapper recovered 1,232 bytes of flash from the initial integrated implementation.

The target's nominal flash limit is 815,104 bytes, but its reserved WarmNodeStore region starts earlier at `0xEA000`. The final image ends at `0xE9B30`, 1,232 bytes below that effective boundary, and passes the warm-region guard.

No unrelated Pro Micro functionality or target configuration was removed.

## Validation

- Built `nrf52_promicro_diy_tcxo`; warm-region, ISR, and LTO guards passed.
- Built `heltec-v4` from the PR branch: 2,314,411 bytes flash and 103,760 bytes RAM.
- Regenerated nanopb output and verified it is reproducible.
- Compared direct outbound encoding byte-for-byte with nanopb at total payload sizes 21, 127, 128, 511, and 512 bytes; every wrapper was at most 20 bytes and reconstructed the original payload exactly.
- Verified the pre-existing `ChunkedPayload` serialization vector remains `080110021803220104`.
- Ported the same transport to current `master`, built and flashed a Heltec V4 app image, and completed a live BLE round trip with a Garmin D2 Mach 2.
- Live validation covered chunked `FromRadio` messages including 116-byte node info and a 171-byte MQTT proxy message, plus a chunked 34-byte `ToRadio` message; all BLE writes completed successfully and config transfer completed without errors.
- Ran `git diff --check`.

The remaining hardware test gap is simultaneous use by two physical PhoneAPI clients; the bounded two-client paths have been reviewed and tested at the serialization/state-machine level.

## Summary by CodeRabbit

* **New Features**
  * Added automatic chunked transport for large `ToRadio`/`FromRadio` protobuf payloads, including outbound chunk sequencing and inbound reassembly.
  * Enabled chunked messaging for clients that advertise chunked support, with proper per-connection state management.

* **Bug Fixes**
  * Improved validation and recovery for incomplete, invalid, oversized, or timed-out fragments to avoid processing corrupted data.

* **Chores**
  * Updated referenced protobuf sources and generated artifacts.


kernel-oops:
Local validation after rebasing onto current `develop`:

- Resolved the protobuf submodule conflict by updating to `kernel-oops/protobufs@90d0af5` from meshtastic/protobufs#980.
- `./.venv/bin/pio` has a stale shebang in this checkout, so I ran PlatformIO via the venv Python instead.
- Command: `.venv/bin/python .venv/bin/pio run -e heltec-v4`
- Result: `heltec-v4 SUCCESS` (`00:37:57.748`, including first-run tool/dependency downloads).

The protobuf PR is now ready for review as well: meshtastic/protobufs#980.

kernel-oops:
Flash footprint update after optimizing only the chunked PhoneAPI implementation:

| `nrf52_promicro_diy_tcxo` build | Flash | RAM | Headroom before reserved warm-store region |
| --- | ---: | ---: | ---: |
| Current `develop` baseline | 800,816 B | 102,612 B | 2,000 B |
| Initial integrated chunk transport | 802,816 B | 104,724 B | 0 B |
| Final optimized chunk transport | 801,584 B | 104,708 B | 1,232 B |

The optimization saves **1,232 bytes of flash** compared with the initial integrated implementation. The final feature cost relative to the same `develop` baseline is **768 bytes of flash** and **2,096 bytes of RAM**.

The board definition's nominal flash limit is 815,104 bytes, which would suggest 13,520 bytes remain. The meaningful constraint on this target is tighter, however: the 12 KiB WarmNodeStore reservation starts at `0xEA000`. The final image ends at `0xE9B30`, so it is **1,232 bytes under the effective protected boundary** and the nRF warm-region guard passes.

The savings came from combining inbound/outbound per-client state and locks, removing redundant state and buffer clearing, removing chunk-only log strings, and replacing nanopb encoding of the small outbound wrapper with a fixed canonical wire encoder. Boundary tests at 21, 127, 128, 511, and 512-byte payload totals matched nanopb byte-for-byte and reconstructed the original payload exactly.

Protocol behavior is unchanged: two clients remain supported, inbound timeout and bounds checks remain, outbound chunking is still opt-in per client, and legacy clients still receive unchunked messages. No unrelated Pro Micro functionality or target configuration was removed.

PR #10934, opened 2026-07-07
Re-enable wantresponse and other fields in simradio when on the decoded path

ianmcorvidae:
This was regressed by Copilot due to https://github.com/meshtastic/firmware/pull/10730/changes/c92e69df63f0d912e91066a89bdce41dcbcef5fe#r3420608957

My machine's assessment suggests that this change was & is correct on the `carryEncrypted` branch added by that PR (https://github.com/meshtastic/firmware/pull/10730) generally, but on the existing side it clears out all the fields on the SIMULATOR_APP packet that are needed -- want_response is the one that mattered for my testing (running traceroutes over a simulated network) but other fields are also cleared by it.

Let me know if the machine has been too verbose in the commenting, heh.

Found due to work on https://github.com/meshtastic/python/pull/946

## 🤝 Attestations

  - [x] Portduino (native-tft)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed simulator loopback so decoded packet details are only cleared for encrypted transmissions; non-encrypted sends now preserve previously decoded fields instead of unintentionally resetting them.
  * Improved consistency of how simulator packets are prepared before receive handling, resulting in more accurate message fields during simulation.

### DIFF
--- a/src/platform/portduino/SimRadio.cpp
+++ b/src/platform/portduino/SimRadio.cpp
@@ -252,7 +252,17 @@ void SimRadio::startSend(meshtastic_MeshPacket *txp)
 
     // pb_encode_to_bytes writes into decoded.payload, which aliases `encrypted` in the union, so all
     // reads of p->encrypted above must be complete before this point.
-    p->decoded = meshtastic_Data_init_zero;
+    if (carryEncrypted) {
+        // On the encrypted path, `decoded` aliases the ciphertext we just copied into c.data;
+        // the remaining `Data` fields hold ciphertext bytes that would serialize as spurious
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
I rather suspect that flash usage report isn't accurate given this PR is adding an `if` statement in portduino-only code, heh

thebentern (approved):

thebentern:
> I rather suspect that flash usage report isn't accurate given this PR is adding an `if` statement in portduino-only code, heh

It's totally off. Tom is investigating. Every PR shouldn't be growing 25k 😅

PR #10939, opened 2026-07-08
Fix SENSOR power saving deep sleep truncating TX and skipping sleep on failed reads

### LINKED ISSUE #10890: [Bug]: EnvironmentTelemetry packet is truncated/lost because deep sleep is triggered before the LoRa transmission actually completes (SENSOR role + Power Saving)
### Category

Other, Hardware Compatibility

### Hardware

Rak4631, Seeed Card Tracker T1000-E

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [ ] BaseUI

### Firmware Version

2.7.26

### Description

Summary

On devices running the SENSOR (or TRACKER) role with Power Saving enabled, the EnvironmentTelemetry packet is reliably not received by neighboring nodes, even though the sending node's own log shows Send packet to mesh and Completed sending. DeviceTelemetry packets sent by the same node in the same boot session are received correctly.

Root cause: EnvironmentTelemetryModule::runOnce() puts the radio to sleep (doDeepSleep()) before the asynchronous LoRa transmission of the just-queued EnvironmentTelemetry packet has actually finished going out over the air.

Environment

Board: RAK4631 (nRF52840)
Role: SENSOR
Power Saving: enabled
Sensor: BME680 (BSEC2)
Firmware: custom build based on master (also reproduced on a build based on release v2.7.26)
Region: EU_868, also reproduced with SHORT_FAST preset (to rule out channel congestion as the cause, see below)

Steps to reproduce

Flash a RAK4631 with an environmental I2C sensor (BME680) attached.
Set role to SENSOR.
Enable power.is_power_saving.
Let the device boot, send its first EnvironmentTelemetry packet, and go to deep sleep.
Observe on a second node in direct radio range (0 hops) whether the EnvironmentTelemetry packet was received.

Result: The second node consistently receives DeviceTelemetry from the sensor node, but never receives the corresponding EnvironmentTelemetry packet sent moments later in the same boot session.

Log evidence

DeviceTelemetry send (uninterrupted, completes cleanly):

INFO  | [DeviceTelemetry] Send packet to mesh
DEBUG | [DeviceTelemetry] enqueue for send (id=0xcfaceaf5 ...)
DEBUG | [RadioIf] Started Tx (id=0xcfaceaf5 ...)
DEBUG | [RadioIf] Packet TX: 101ms
DEBUG | [RadioIf] Completed sending (id=0xcfaceaf5 ...)

EnvironmentTelemetry send (interrupted by the sleep sequence):

INFO  | [EnvironmentTelemetry] Send packet to mesh
DEBUG | [EnvironmentTelemetry] enqueue for send (id=0xd24086f6 ...)
DEBUG | [EnvironmentTelemetry] Start next execution in 5s, then sleep
DEBUG | [RadioIf] Started Tx (id=0xd24086f6 ...)
DEBUG | [RadioIf] 0 packets remain in the TX queue
DEBUG | [EnvironmentTelemetry] Sleep for 1800000ms, then awake to send metrics again
INFO  | [EnvironmentTelemetry] Enter deep sleep for 1800 seconds
INFO  | [EnvironmentTelemetry] Disable NRF52 bluetooth
DEBUG | [EnvironmentTelemetry] SX126x entering sleep mode      <-- radio put to sleep here
DEBUG | [EnvironmentTelemetry] Packet TX: 101ms                <-- "TX complete" logged AFTER the radio was told to sleep
DEBUG | [EnvironmentTelemetry] Completed sending (id=0xd24086f6 ...)

Note that Start next execution in 5s, then sleep and Enter deep sleep are logged in the same second — the intended 5-second delay before sleeping is never actually honored.

Root cause analysis

In src/modules/Telemetry/EnvironmentTelemetry.cpp, sendTelemetry() sets a delay before the next run of the module, intended to give the radio time to finish transmitting before the device goes to sleep:

```
service->sendToMesh(p, RX_SRC_LOCAL, true);

if (config.device.role == meshtastic_Config_DeviceConfig_Role_SENSOR && config.power.is_power_saving) {
    ...
    sleepOnNextExecution = true;
    LOG_DEBUG("Start next execution in 5s, then sleep");
    setIntervalFromNow(FIVE_SECONDS_MS);
}
```

However, at the very end of runOnce(), the function's own return value overrides this:

`return min(sendToPhoneIntervalMs, result);`

result comes from the per-sensor polling loop earlier in the same function:

```
for (TelemetrySensor *sensor : sensors) {
    uint32_t delay = sensor->runOnce();
    if (delay < result) result = delay;
}
```

For BME680, BME680Sensor::runOnce() unconditionally returns 35 (ms):

```
int32_t BME680Sensor::runOnce()
{
    if (!bme680.run()) {
        checkStatus("runTrigger");
    }
    return 35;
}
```

Since OSThread schedules the next call to runOnce() based on this return value, the module's own return statement effectively overrides setIntervalFromNow() (and the 5-second — or manually increased 15-second — delay we intended). runOnce() gets called again almost immediately (~35ms later), sees sleepOnNextExecution == true, and calls doDeepSleep() right away — while the radio is still asynchronously transmitting the EnvironmentTelemetry packet queued moments earlier. doDeepSleep() puts the SX126x radio to sleep, truncating/aborting the in-flight transmission.

I confirmed this by manually increasing the delay from FIVE_SECONDS_MS to 15 * 1000 in setIntervalFromNow() — this had no effect on the timing shown in the log, confirming that this value is not what actually governs the reschedule delay; the trailing return min(sendToPhoneIntervalMs, result) is.

This bug affects any telemetry sensor whose runOnce() returns a short polling interval (not just BME680) in the SENSOR + Power Saving combination, since the override happens at the module level, not the sensor level.

Suggested fix

Make the function's return value respect the pending sleep instead of letting the sensor polling delay silently override it, e.g.:

```
if (sleepOnNextExecution) {
    return FIVE_SECONDS_MS; // Radio needs time to finish transmitting before we sleep
}
return min(sendToPhoneIntervalMs, result);
```

placed just before the final return statement in EnvironmentTelemetryModule::runOnce().

Alternatively/additionally, doDeepSleep() could actively check/wait for the radio's TX-in-progress state before proceeding, rather than relying purely on a fixed delay.

Impact

For SENSOR-role devices with Power Saving enabled, this means environmental telemetry is essentially never reliably delivered to the mesh, even though the device believes it sent it successfully (misleading log output: Send packet to mesh / Completed sending both appear, giving no indication that the packet was cut short on air). DeviceTelemetry, which is not followed by an immediate sleep trigger, is unaffected and works normally.

### Relevant log output

```Shell
←[36m←[0m←[34mDEBUG ←[0m| ??:??:?? 1907 [EnvironmentTelemetry] ←[34mStart next execution in 5s, then sleep←[0m←[0m
←[36m←[0m←[34mDEBUG ←[0m| ??:??:?? 1907 [RadioIf] ←[34mStarted Tx (id=0xd24086f6 fr=0x68a531eb to=0xffffffff, transport = 0, WantAck=0, HopLim=7 Ch=112 encrypted len=117 hopStart=7 relay=0xeb priority=70)←[0m←[0m
←[36m←[0m←[34mDEBUG ←[0m| ??:??:?? 1907 [RadioIf] ←[34m0 packets remain in the TX queue←[0m←[0m
←[36m←[0m←[34mDEBUG ←[0m| ??:??:?? 1907 [EnvironmentTelemetry] ←[34mSleep for 1800000ms, then awake to send metrics again←[0m←[0m
←[37m←[0m←[32mINFO  ←[0m| ??:??:?? 1907 [EnvironmentTelemetry] ←[32mEnter deep sleep for 1800 seconds←[0m←[0m
←[37m←[0m←[32mINFO  ←[0m| ??:??:?? 1907 [EnvironmentTelemetry] ←[32mDisable NRF52 bluetooth←[0m←[0m
**←[36m←[0m←[34mDEBUG ←[0m| ??:??:?? 1907 [EnvironmentTelemetry] ←[34mSX126x entering sleep mode←[0m←[0m
←[36m←[0m←[34mDEBUG ←[0m| ??:??:?? 1907 [EnvironmentTelemetry] ←[34mPacket TX: 101ms←[0m←[0m
←[36m←[0m←[34mDEBUG ←[0m| ??:??:?? 1907 [EnvironmentTelemetry] ←[34mCompleted sending (id=0xd24086f6 fr=0x68a531eb to=0xffffffff, transport = 0, WantAck=0, HopLim=7 Ch=112 encrypted len=117 hopStart=7 relay=0xeb priority=70)←[0m←[0m**
←[36m←[0m←[34mDEBUG ←[0m| ??:??:?? 1907 [EnvironmentTelemetry] ←[34mSave to disk 31←[0m←[0m
```

### LINKED ISSUE #10932: [Bug]: BME680 BSEC2 "call timing violation" (code 100) under heavy mesh traffic prevents EnvironmentTelemetry from being sent and blocks deep sleep on SENSOR/TRACKER role
### Category

Other, Hardware Compatibility

### Hardware

Seeed Card Tracker T1000-E, Rak4631

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [ ] BaseUI

### Firmware Version

2.7.26

### Description

Summary

On a SENSOR-role (and TRACKER-role) node with Power Saving enabled and a BME680 (BSEC2) sensor attached, running on a busy mesh (many neighboring nodes, high channel activity), the sensor intermittently fails to produce a valid reading because bme680.run() reports a BSEC2 "call timing violation" (status code 100). When this happens, EnvironmentTelemetryModule::getEnvironmentTelemetry() never returns valid data for that wake cycle — and because of how the SENSOR + Power Saving deep-sleep trigger is currently implemented, the node then fails to enter deep sleep for that entire cycle, staying awake until the next successful telemetry interval.

Environment

Board: RAK4631 (nRF52840)
Role: SENSOR
Power Saving: enabled
Sensor: BME680 (BSEC2)
Firmware: v2.7.26
Region: EU_868

Steps to reproduce

Flash a RAK4631 with a BME680 sensor attached.
Set role to SENSOR, enable power.is_power_saving.
Place the node in a mesh with meaningful channel activity (many neighboring nodes, e.g. LONG_FAST preset in a dense urban deployment).
Observe the serial log during a wake cycle.

Under low traffic (verified by switching to SHORT_FAST, which effectively isolates the node from most neighboring traffic on the radio layer), the sensor reads and sends reliably every cycle. Under normal LONG_FAST traffic in a busy mesh, the following warning appears intermittently:
```
WARN | 208 [EnvironmentTelemetry] runTrigger BSEC2 code: 100
WARN | 259 [EnvironmentTelemetry] runTrigger BSEC2 code: 100
```

Root cause

```
BME680Sensor::runOnce():

int32_t BME680Sensor::runOnce()
{
    if (!bme680.run()) {
        checkStatus("runTrigger");
    }
    return 35;
}
```

BSEC2 status code 100 is BSEC_W_SC_CALL_TIMING_VIOLATION:

"Difference between actual and defined sampling intervals of bsec_sensor_control() greater than allowed"

The BME680 is subscribed with BSEC_SAMPLE_RATE_LP (~3-second expected sampling interval). Under heavy mesh traffic, the main loop can be busy long enough handling LoRa RX, routing, and TrafficManagement processing that BME680Sensor::runOnce() isn't invoked closely enough to the expected 3-second cadence, causing BSEC to report a timing violation for that measurement cycle. When this happens, bme680.getData(BSEC_OUTPUT_RAW_PRESSURE).signal stays at 0, and BME680Sensor::getMetrics() returns false:

```
bool BME680Sensor::getMetrics(meshtastic_Telemetry *measurement)
{
    if (bme680.getData(BSEC_OUTPUT_RAW_PRESSURE).signal == 0)
        return false;
    ...
}
```

which in turn makes EnvironmentTelemetryModule::getEnvironmentTelemetry() and sendTelemetry() return false for that cycle — no packet is sent, which is expected/acceptable behavior on its own.

The actual bug: the node never goes to sleep for that cycle

The problem is that the SENSOR + Power Saving deep-sleep trigger currently lives exclusively inside the success branch of sendTelemetry():

```
bool EnvironmentTelemetryModule::sendTelemetry(NodeNum dest, bool phoneOnly)
{
    ...
    if (getEnvironmentTelemetry(&m)) {
        ...
        service->sendToMesh(p, RX_SRC_LOCAL, true);

        if (config.device.role == meshtastic_Config_DeviceConfig_Role_SENSOR && config.power.is_power_saving) {
            ...
            sleepOnNextExecution = true;
            setIntervalFromNow(FIVE_SECONDS_MS);
        }
        return true;
    }
    return false;
}
```

If getEnvironmentTelemetry() returns false (e.g. due to the BSEC timing violation above), sleepOnNextExecution is never set. There is no other code path in the SENSOR role that triggers doDeepSleep() — this is the only place it happens. As a result, on a wake cycle where the BME680 reading fails due to a timing violation, the node stays awake until the next telemetry cycle instead of simply skipping telemetry interval and sleeping as usual. This is confirmed in serial logs: on affected cycles, uptime continues climbing (600s, 1200s, 1800s+...) with periodic housekeeping log lines (TrafficManagement Maintenance, Power Battery, SX126x AGC reset), but no Enter deep sleep ever appears, until the mesh traffic decreases and the node manages to get a clean reading in the next cycle.

Under a busy mesh (dense urban deployment with many neighboring nodes), this timing violation is not a rare edge case — it recurs periodically depending on channel activity at the exact moment the module wakes up, making the node's deep sleep behavior in the SENSOR role unreliable in practice.

Suggested fix

Decouple the sleep trigger from telemetry send success, so a SENSOR + Power Saving node reliably returns to sleep even when the sensor reading fails for that cycle:

```
bool EnvironmentTelemetryModule::sendTelemetry(NodeNum dest, bool phoneOnly)
{
    meshtastic_Telemetry m = meshtastic_Telemetry_init_zero;
    m.which_variant = meshtastic_Telemetry_environment_metrics_tag;
    m.time = getTime();

    bool gotTelemetry = getEnvironmentTelemetry(&m);

    if (gotTelemetry) {
        // ... existing LOG_INFO / packet build / sendToMesh / sendToPhone logic unchanged ...
    } else if (!phoneOnly) {
        LOG_WARN("Environment telemetry unavailable this cycle, skipping send");
    }

    // Regardless of whether we had valid data, make sure a SENSOR + Power Saving node still sleeps.
    if (!phoneOnly && config.device.role == meshtastic_Config_DeviceConfig_Role_SENSOR && config.power.is_power_saving) {
        if (gotTelemetry) {
            // existing client notification logic
        }
        sleepOnNextExecution = true;
        LOG_DEBUG("Start next execution in 5s, then sleep");
        setIntervalFromNow(FIVE_SECONDS_MS);
    }

    return gotTelemetry;
}
```

This way, a failed reading simply means environmental telemetry is skipped for that one interval — the node still returns to deep sleep as expected, instead of remaining awake (and draining battery) until manual intervention.

Note: this fix should be applied together with the related sleep-timing race condition described in #10890  ( runOnce()'s final return min(sendToPhoneIntervalMs, result) currently overrides the setIntervalFromNow() delay set above, which can otherwise cause doDeepSleep() to run before sleepOnNextExecution is properly honored ). Combining both fixes ensures the sleep trigger is both reachable regardless of sensor read success, and actually respected with the intended delay.

Impact

For any SENSOR/TRACKER-role, Power-Saving device using a BSEC2-based sensor (BME680) deployed in a mesh with non-trivial channel activity, this bug causes the node to occasionally get "stuck awake" for an entire wake cycle whenever a timing violation coincides with the telemetry read attempt — significantly undermining the battery-saving purpose of the SENSOR + Power Saving combination, with no telemetry sent and no indication to the user other than the node failing to sleep as expected.

### Relevant log output

```Shell
WARN | 208 [EnvironmentTelemetry] runTrigger BSEC2 code: 100
WARN | 259 [EnvironmentTelemetry] runTrigger BSEC2 code: 100
```

caveman99:
Fixes #10890
Fixes #10932

## Summary by CodeRabbit

* **New Features**
  * Improved sleep handling to better distinguish between light sleep and deep sleep, helping devices pause more safely during active radio traffic.

* **Bug Fixes**
  * Telemetry updates now avoid starting deep sleep too early, reducing the chance of interrupted transmissions.
  * Air quality, environment, health, and power telemetry now handle missing readings more gracefully and preserve a short grace period before sleeping.
  * Sleep retries are now deferred for a limited time when the radio is still busy.


caveman99:
Addressed in 12de139: the radio-busy deferral and its counter moved into BaseTelemetryModule (shouldDeferDeepSleep), together with a shared isPowerSavingSensor() helper. This also removes the telemetry-specific counter from OSThread. The sleep arming block in sendTelemetry() and the grace-period return stay per module: they access protected OSThread members (sleepOnNextExecution, setIntervalFromNow) that BaseTelemetryModule cannot reach without restructuring the class hierarchy, which is out of proportion for this change.

thebentern (approved):

PR #10947, opened 2026-07-08
Fixed pin state setup when SX126X_ANT_SW is defined

AlexandrZloy:
Fixed the order of pinMode and digitalWrite calls if SX126X_ANT_SW is defined.

## 🤝 Attestations

- [V] I have tested that my proposed changes behave as described.
- [V] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - Wio-SX1262 for XIAO LoRa Module with the Waweshare ESP32-S3-Zero as the MCU
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SX126x startup reliability by configuring the antenna switch GPIO correctly before enabling it.


AlexandrZloy:
Agreed with **coderabbitai** that `portduino_config.lora_sx126x_ant_sw_pin.pin` case was overlooked and should be fixed as well.

PR #10949, opened 2026-07-09
Make Safe File actually safe

jp-bennett:
SafeFile has been something of a landmine, and several uses of it were failing to manage the spilock correctly. These changes eliminate the need to touch FSCom or spiLock directly for users of SafeFile.

These are pretty core changes, so lots of testing needed.

If we like this approach, I'd recommend adding file read routines to safe file, and work to eliminate raw FSCom access everywhere.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of saving settings, messages, caches, and calibration data to flash.
  * Reduced the chance of storage write failures by handling file access locking more consistently.
  * Persisted data now creates needed folders automatically in more cases, avoiding missing-directory errors.
  * Updated several save flows to avoid lock conflicts during file operations.

### DIFF
--- a/src/MessageStore.cpp
+++ b/src/MessageStore.cpp
@@ -291,23 +291,16 @@ static inline bool readMessageRecord(File &f, StoredMessage &m)
 void MessageStore::saveToFlash()
 {
 #ifdef FSCom
-    // Ensure root exists
-    spiLock->lock();
-    FSCom.mkdir("/");
-    spiLock->unlock();
-
     SafeFile f(filename.c_str(), false);
 
-    spiLock->lock();
     uint8_t count = static_cast<uint8_t>(liveMessages.size());
     if (count > MAX_MESSAGES_SAVED)
         count = MAX_MESSAGES_SAVED;
-    f.write(&count, 1);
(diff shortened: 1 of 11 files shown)


thebentern:
In theory I agree with all of this. It's a risky 11th hour change I'd like to hold off on for now though

jp-bennett:
> In theory I agree with all of this. It's a risky 11th hour change I'd like to hold off on for now though

Yes, agreed. It is worth noting that there are a couple instances where FSCom is currently getting called without taking the spiLock first.

thebentern:
> > In theory I agree with all of this. It's a risky 11th hour change I'd like to hold off on for now though
> 
> Yes, agreed. It is worth noting that there are a couple instances where FSCom is currently getting called without taking the spiLock first.

Maybe we can peel those off into a separate PR

oscgonfer:
Hi @jp-bennett ,

Is this the right approach? If not I can change it for this (and any other sensor).

https://github.com/meshtastic/firmware/blob/develop/src/modules/Telemetry/Sensor/SEN5XSensor.cpp#L390

PR #10951, opened 2026-07-09
fix(modules): null-check packetPool/clientNotificationPool/mqttClientProxyMessagePool allocations

ndoo:
Follow-up to #10948, which fixed the core routing path (`Router`/`NextHopRouter`/`ReliableRouter`/`MeshService`/`RadioLibInterface`) after reproducing a HardFault on STM32WL hardware under real mesh traffic. This covers the same `allocCopy()`/`allocZeroed()` unchecked-return pattern in lower-frequency paths that were out of scope for that PR: `SimRadio.cpp` (portduino sim RX/TX), `NodeInfoModule`, the Telemetry modules' power-saving-sleep notifications, `MQTT` (map report, client-proxy messages, config-validation notifications), `PositionModule`, `SerialModule`, and `KeyVerificationModule`.

Sites where the allocation result was already read back through an existing null check downstream (e.g. Telemetry's `lastMeasurementPacket`, SimRadio's `receivingPacket`, `AdminModule::sendWarning`) were verified safe and left untouched. `Router::allocForSending()` (and everything that funnels through it, e.g. `allocDataProtobuf()`) remains unguarded and out of scope here too — it's called from 30+ sites across the codebase and needs its own audit.

**Important:** this PR alone is not a complete fix for meshpacket null-dereference crashes on memory-constrained platforms (e.g. STM32WL) — #10948 needs to land too. The two PRs cover disjoint call sites of the same unchecked-`allocCopy()`/`allocZeroed()` pattern; without #10948, the core routing/radio path (`RadioLibInterface`, `Router`, etc.) is still exposed. See the hardware soak test below, which reproduced this exact crash class on this branch without #10948 and confirmed it goes away once #10948's guard is present.

Also includes an incidental fix flagged by CodeRabbit's automated review: four `sprintf(cn->message, ...)` calls in `KeyVerificationModule.cpp` wrote into the notification's fixed-size buffer without bounds checking. Switched to `snprintf(cn->message, sizeof(cn->message), ...)`, matching the pattern already used for the same field in `SerialModule.cpp` and `MQTT.cpp` elsewhere in this PR.

## Hardware soak test (wio-e5)

wio-e5's dev-board USB-serial is not wired to the STM32WL UART bootloader, so it can only be flashed via ST-Link/SWD (`STM32_Programmer_CLI -c port=SWD -w <hex> -v -rst`), not the usual meshtastic UART DFU flow.

By default this target ships with `DEBUG_MUTE` set and no `PIO_FRAMEWORK_ARDUINO_NANOLIB_FLOAT_PRINTF`, so there's no serial log output at all — needed both flipped to get logs for this test. That pushed the build over the flash budget, so a few features wio-e5 normally ships with (GPS, I2C, environmental sensor) were also temporarily excluded to make room for the log strings + float-printf support. We also temporarily raised `MAX_NUM_NODES` from the STM32WL default of 10 to 50 to force sustained NodeDB churn/heap pressure — the default cap wasn't enough to reliably reproduce memory pressure. None of this was committed; it's local-only test-rig configuration, applied on top of this PR's commits:

```diff
diff --git a/variants/stm32/stm32.ini b/variants/stm32/stm32.ini
index 5726fd369..940d4b3d5 100644
--- a/variants/stm32/stm32.ini
+++ b/variants/stm32/stm32.ini
@@ -28,8 +28,8 @@ build_flags =
   -DMESHTASTIC_EXCLUDE_XEDDSA=1 ; The Ed25519 signing code does not fit in the 256KB flash. Packets are sent unsigned, like pre-XEdDSA firmware.
   -DSERIAL_RX_BUFFER_SIZE=256 ; For GPS - the default of 64 is too small.
   -DHAS_SCREEN=0 ; Always disable screen for STM32, it is not supported.
-  ;-DPIO_FRAMEWORK_ARDUINO_NANOLIB_FLOAT_PRINTF ; Enable this if enabling debugg logging. It is REQUIRED for at least traceroute debug prints - without it the length returned by printf ends up uninitialized.
-  -DDEBUG_MUTE ; You can #undef DEBUG_MUTE in certain source files if you need the logs.
+  -DPIO_FRAMEWORK_ARDUINO_NANOLIB_FLOAT_PRINTF ; Enable this if enabling debugg logging. It is REQUIRED for at least traceroute debug prints - without it the length returned by printf ends up uninitialized.
+  ;-DDEBUG_MUTE ; You can #undef DEBUG_MUTE in certain source files if you need the logs.
   -fmerge-all-constants
   -ffunction-sections
   -fdata-sections
diff --git a/variants/stm32/wio-e5/platformio.ini b/variants/stm32/wio-e5/platformio.ini
index 8c7579aa3..6a28325b9 100644
--- a/variants/stm32/wio-e5/platformio.ini
+++ b/variants/stm32/wio-e5/platformio.ini
@@ -18,5 +18,9 @@ build_flags =
   -DHAS_GPS=1
   -DGPS_SERIAL_PORT=Serial2
   -DMESHTASTIC_EXCLUDE_AIR_QUALITY_SENSOR=1
+  -DMESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR=1 ; TEMP: soak-test-only, freeing flash for debug logging. Do not commit.
+  -DMESHTASTIC_EXCLUDE_I2C=1 ; TEMP: soak-test-only, freeing flash for debug logging. Do not commit.
+  -DMESHTASTIC_EXCLUDE_GPS=1 ; TEMP: soak-test-only, freeing flash for debug logging. Do not commit.
+  -DMAX_NUM_NODES=50 ; TEMP: soak-test-only, intentionally raising NodeDB memory pressure. Do not commit.

 upload_port = stlink
```

(The raw serial captures aren't attached — they contain live mesh traffic from a real MQTT-bridged mesh, including other people's node names, IDs, and public keys. Only the aggregate timing/heap findings below are shared.)

#10948 should be merged together with this PR to form a complete fix for meshpacket null-dereference crashes. For this soak test, `ad0d36905` (#10948's null-check commit) was cherry-picked onto this branch to get a clean read on this PR's own changes.

With that in place: 46 minutes uptime, zero resets. NodeDB stayed pinned at the 50-node cap with as little as 3316 bytes free, continuously cycling eviction/insertion, without incident.

This confirms this PR's own changes are stable under sustained memory pressure. The soak test heavily exercised the NodeDB/Router/RadioLibInterface hot path, but the captured mesh traffic didn't specifically trigger allocation failures in the lower-frequency paths this PR actually touches (`KeyVerificationModule`, `SerialModule`, `MQTT`, `PositionModule`, `SimRadio`, `NodeInfoModule`) — those still need dedicated exercise once #10948 lands and this branch is rebased onto it.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Seeed Wio-E5 (see hardware soak test above)

Compiles clean on `rak3172` and `wio-e5`. See the hardware soak test section above for wio-e5 runtime results — no regressions from this PR's own changes found, but the reset loop uncovered during testing depends on #10948 landing first. Will re-test the specific modules this PR touches once this branch is rebased onto #10948.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability across key verification, node info forwarding, position/telemetry, serial validation, MQTT publishing, and simulated radio sending/receiving.
  * Added safeguards so client notifications, proxy messages, and packet copies are only populated and sent when allocations succeed.
  * Prevented rare crashes/invalid behavior when the system is under low-memory conditions, allowing operations to fail gracefully instead of dereferencing null pointers.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1053,10 +1053,12 @@ void setup()
     if (nodeDB->keyIsLowEntropy && !nodeDB->hasWarned) {
         LOG_WARN(LOW_ENTROPY_WARNING);
         meshtastic_ClientNotification *cn = clientNotificationPool.allocZeroed();
-        cn->level = meshtastic_LogRecord_Level_WARNING;
-        cn->time = getValidTime(RTCQualityFromNet);
-        sprintf(cn->message, LOW_ENTROPY_WARNING);
-        service->sendClientNotification(cn);
+        if (cn) {
+            cn->level = meshtastic_LogRecord_Level_WARNING;
+            cn->time = getValidTime(RTCQualityFromNet);
+            sprintf(cn->message, LOW_ENTROPY_WARNING);
+            service->sendClientNotification(cn);
+        }
         nodeDB->hasWarned = true;
     }
 #endif
--- a/src/modules/KeyVerificationModule.cpp
+++ b/src/modules/KeyVerificationModule.cpp
@@ -83,7 +83,7 @@ bool KeyVerificationModule::handleReceivedProtobuf(const meshtastic_MeshPacket &
         meshtastic_ClientNotification *cn = clientNotificationPool.allocZeroed();
         if (cn) {
             cn->level = meshtastic_LogRecord_Level_WARNING;
-            sprintf(cn->message, "Enter Security Number for Key Verification");
+            snprintf(cn->message, sizeof(cn->message), "Enter Security Number for Key Verification");
             cn->which_payload_variant = meshtastic_ClientNotification_key_verification_number_request_tag;
             cn->payload_variant.key_verification_number_request.nonce = currentNonce;
             copyNodeLongNameOrUnknown(cn->payload_variant.key_verification_number_request.remote_longname,
@@ -118,7 +118,7 @@ bool KeyVerificationModule::handleReceivedProtobuf(const meshtastic_MeshPacket &
             meshtastic_ClientNotification *cn = clientNotificationPool.allocZeroed();
             if (cn) {
                 cn->level = meshtastic_LogRecord_Level_WARNING;
-                sprintf(cn->message, "Final confirmation for incoming manual key verification %s", message);
+                snprintf(cn->message, sizeof(cn->message), "Final confirmation for incoming manual key verification %s", message);
                 cn->which_payload_variant = meshtastic_ClientNotification_key_verification_final_tag;
                 cn->payload_variant.key_verification_final.nonce = currentNonce;
                 copyNodeLongNameOrUnknown(cn->payload_variant.key_verification_final.remote_longname,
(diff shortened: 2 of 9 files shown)


ndoo:
@coderabbitai review

ndoo:
@coderabbitai review

thebentern (approved):

PR #10973, opened 2026-07-10
tlora-t3s3-epaper: e-ink startup and build cleanup

HarukiToreda:
## Summary

This PR cleans up the `tlora-t3s3-epaper` / InkHUD build, reduces flash usage on this target, and fixes the T3-S3 e-ink SPI setup so both `InkHUD` and `BaseUI` can share the SD-card bus correctly.

## What changed

- Cleaned up the InkHUD build configuration so it no longer pulls in unnecessary `BaseUI` renderer/support sources.
- Reused the SD card `HSPI` bus for the T3-S3 e-ink display instead of creating a second `HSPI` instance.
- Updated the `BaseUI` e-ink path to also reuse the shared `HSPI` bus and preconfigure the panel control pins before display init.
- Disabled unused subsystems for this variant:
  - `MESHTASTIC_EXCLUDE_AUDIO=1`
  - `MESHTASTIC_EXCLUDE_WEBSERVER=1`
- Reduced flash usage for this target from about `96%` to about `86%`.
- Increased `CONFIG_ESP_IPC_TASK_STACK_SIZE` to `2048` for the `tlora-t3s3-epaper-inkhud` environment to avoid NimBLE init stack overflow on this target.

## Why

These changes are targeted cleanup and stabilization for the T3-S3 e-ink configuration:

- InkHUD should build only the NicheGraphics/InkHUD path it actually uses, not extra `BaseUI` pieces.
- The T3-S3 display and SD card share `HSPI`, so the display should reuse the bus initialized by `setupSDCard()` rather than reinitializing the same host separately.
- Reusing the shared bus also keeps the SD card path fully usable for future features instead of competing with a second display-owned SPI setup.
- The `BaseUI` e-ink path was also reinitializing `HSPI`, which caused duplicate SPI setup and panel init failures on this board. Updating it to use the shared bus makes both display paths behave consistently.
- Audio and webserver are not needed for this variant, and excluding them helps bring flash usage down from about `96%` to about `86%`.
- The larger IPC task stack prevents a NimBLE init overflow seen on this board/environment.

## Notes

This PR is focused on the `tlora-t3s3-epaper` target and its e-ink display paths. It does not change general UI behavior outside of the InkHUD build filtering, variant-specific size reduction, and shared SPI setup.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved e-ink display initialization and SPI communication on the T-Lora T3S3 E-Paper device.
  - Improved display control and startup reliability for supported e-ink hardware.

- **Configuration**
  - Streamlined the InkHUD graphics build to include only required components.
  - Disabled audio and web server features for T-Lora T3S3 E-Paper builds to better fit device capabilities.


variants/esp32s3/tlora_t3s3_epaper/platformio.ini
```diff
@@ -39,6 +41,9 @@ extends = esp32s3_base, inkhud
 board = tlora-t3s3-v1
 board_check = true
 upload_protocol = esptool
+custom_sdkconfig =
+  ${esp32s3_base.custom_sdkconfig}
+  CONFIG_ESP_IPC_TASK_STACK_SIZE=2048 ; default 1024 overflows during NimBLE init off USB power
```
thebentern, line 46: Have you seen this issue occur on other boards? Might be a good default up a level

PR #10976, opened 2026-07-10
Fix serial protobuf corruption on short USB CDC writes

### LINKED ISSUE #10975: [Bug]: ESP32-S3 serial protobuf frames are corrupted during full NodeDB sync
## Category

Serial

## Hardware

- Heltec Wireless Tracker V2
- M5Stack Cardputer ADV

No firmware UI component is involved.

## Firmware version

- Reproduced on `2.8.0.b3ddeca`
- Confirmed in `develop` source at `91043face`
- Client: meshtastic-python 2.7.10

## Description

A serial API client downloading the full NodeDB from an ESP32-S3 device can receive malformed `FromRadio` frames. Depending on which frame is damaged, the connection either times out waiting for `config_complete_id` or reports success with a silently incomplete NodeDB.

I reproduced two independent firmware-side corruption mechanisms.

### 1. Raw logs can interleave with framed protobuf output

After `SerialConsole` enters protobuf mode, `log_to_serial()` still falls back to `RedirectablePrint::log_to_serial()` when `security.debug_log_api_enabled=false`. Those unframed ANSI/debug bytes share the same serial stream but do not share the framed-write lock.

A Heltec Tracker V2 byte capture contained raw ANSI `DEBUG` text inside the declared payload of a `FromRadio.node_info` frame. meshtastic-python then raised `DecodeError: Truncated message`, never received the matching `config_complete_id`, and timed out after 30 seconds.

### 2. A short HWCDC write abandons the remainder of a frame

`StreamAPI::writeFrame()` performs one `stream->write(buf, totalLen)`. ESP32-S3 `HWCDC::write()` can return fewer bytes than requested after bounded backpressure. SerialConsole does not override `onFrameWriteFailed()`, so the already-dequeued frame tail is discarded and the next `0x94 0xC3` header is written immediately.

Cardputer ADV captures contain exactly this byte shape:

```text
partial FromRadio.node_info or FromRadio.log_record payload
94 c3 <next-frame-length> <next FromRadio frame>
```

The receiver has already accepted the first frame's declared length, so it consumes the nested header and next frame as payload.

Enabling `debug_log_api_enabled` is therefore not a workaround. It removes raw ANSI output but increases framed traffic and made the short-write failure substantially more frequent.

## Reproduction

Use an ESP32-S3 USB CDC device with a large NodeDB and repeatedly open a full serial API session. Do not use `--no-nodes`:

```sh
PORT=/dev/cu.usbmodemXXXX
for attempt in $(seq 1 100); do
  echo "attempt $attempt"
  meshtastic --port "$PORT" --nodes >"/tmp/nodes-$attempt.log" 2>&1 || break
done
```

A debugOut/raw-byte capture is useful because a session can complete while returning fewer nodes than `myInfo.nodedb_count`.

## Baseline results

Cardputer ADV, 200-node NodeDB:

```text
debug_log_api_enabled=false
attempt 43: 175/200 nodes, 6 malformed FromRadio frames
attempt 92: 193/200 nodes, 2 malformed FromRadio frames
100 opens: 2 incomplete NodeDBs, 8 decode errors

debug_log_api_enabled=true
28/100 attempts contained malformed frames
85 malformed FromRadio frames total
0 raw ANSI records
```

Representative parser errors:

```text
google.protobuf.message.DecodeError: Wire format was corrupt
google.protobuf.message.DecodeError: String field had bad UTF-8
google.protobuf.message.DecodeError: Truncated message
```

The Heltec Tracker V2 independently reproduced the raw-log insertion path during a 45-node full sync. The Cardputer ADV reproduced the short-write/nested-header path.

## Impact

- Full NodeDB/config downloads can time out.
- Successful connections can silently expose an incomplete node list.
- The parser cannot safely resynchronize on a nested `0x94 0xC3`, because those bytes are legal inside protobuf byte fields.
- `--no-nodes` avoids the high-volume replay and is the current safe workaround, but does not provide the full NodeDB.

This is distinct from #10955/#10956. That issue fixed a post-disconnect blocking log write that starved the task watchdog. This corruption happens while a host is actively connected and draining the serial API. Any fix must preserve #10956's bounded/no-host behavior rather than retrying indefinitely.

A firmware fix and native regression suite have been tested on both affected devices; I will link the PR to this issue.

p0ns:
Fixes #10975

Implementation: `p0ns:fix/serial-stream-frame-integrity`, head `a9d5be531` (five commits), based on `develop` at `f52e2ea8e`.

ESP32-S3 serial clients could receive malformed protobuf streams during full NodeDB/config replay through two independent paths:

1. `SerialConsole` emitted unframed debug text after switching the port to protobuf mode when `debug_log_api_enabled` was false.
2. `HWCDC::write()` can return a short write under backpressure. The old generic `StreamAPI` path abandoned the remaining bytes after PhoneAPI had already advanced, allowing the next `0x94 0xC3` header to land inside the previous declared payload.

This can either time out the config handshake or silently return fewer nodes than `myInfo.nodedb_count`.

#### Changes

- Suppress all unframed serial output once protobuf mode is active.
- Honor `pauseBluetoothLogging` during config/NodeDB replay.
- Add a transport hook that completes retained output before `getFromRadio()` can advance to another packet.
- Continue ESP32-S3 USB CDC short writes over later loop passes using pointers into the existing persistent `txBuf`/`txBufLog`; no additional frame-sized copy is allocated.
- Keep framed logs best-effort: do not start one unless its complete frame fits in the HWCDC TX ring.
- Preserve an already-encoded required frame behind a synchronous partial log instead of dropping either frame or interleaving them.
- Retain an incomplete frame across disconnects. HWCDC may still hold its accepted prefix; clearing metadata alone could put the next header inside that frame's declared payload.
- Do not call `HWCDC::flush()` for serial framed output; its no-progress path can discard queued bytes.
- Keep generic StreamAPI/TCP behavior unchanged: a short write still invokes `onFrameWriteFailed()` so TCP transports can close the connection.
- Extract the pending/deferred state machine into `StreamFrameWriter` so native CI tests the same production logic used by SerialConsole.

Each continuation call makes at most one bounded `Stream::write()` attempt. Zero-byte progress returns to the main loop rather than spinning, so this does not reintroduce the 90-second task-watchdog failure fixed by #10956.

Frame storage in `StreamFrameWriter` is non-owning. SerialConsole only passes its persistent main and log buffers, and holds `streamLock` around every helper call.

#### Native regression tests

Added `test_stream_api` coverage for:

- continuing exactly the unwritten frame tail across multiple short writes;
- completing a retained old frame before a replacement client's new frame;
- preserving wire order when a required main frame is deferred behind a partial log;
- dropping best-effort logs before writing when the complete frame cannot fit;
- making one bounded attempt per zero-progress call;
- unchanged generic StreamAPI framing, flush, and short-write failure-hook behavior;
- completing pending output before PhoneAPI dequeues/encodes another packet;
- gating log encoding and marking framed logs best-effort;
- suppressing real SerialConsole raw bytes and logs after protobuf mode begins.

```text
coverage: test_stream_api 10/10 passed
full native suite: GREEN 31/31 suites, 583/583 tests
```

Firmware builds pass for:

```text
m5stack-cardputer-adv
heltec-wireless-tracker-v2
tbeam
rak4631
```

#### Hardware validation

Tested with meshtastic-python 2.7.10 on macOS:

```text
Cardputer ADV:        200-node NodeDB
Heltec Tracker V2:     45-node NodeDB
both debug_log_api_enabled=false and true
```

Initial candidate matrix:

```text
Cardputer false: 100/100 complete, 0 malformed frames
Cardputer true:  100/100 complete, 0 malformed frames
Heltec false:    100/100 complete, 0 malformed frames
Heltec true:     100/100 complete, 0 malformed frames
```

Final candidate matrix:

```text
Cardputer: 50 false + 20 true, all complete
Heltec:    50 false + 20 true, all complete
```

Additional backpressure/lifecycle validation:

- 20/20 forced 2-5 second mid-download reader stalls resumed with complete NodeDBs.
- 2/2 forced 20-second stalls completed before the client's 30-second timeout.
- 10/10 stalled-client SIGKILL/replacement cycles completed on the replacement client.
- Two 150-second post-close checks kept reboot counts flat: Cardputer `74 -> 74`, Heltec `137 -> 137`.

Before rebasing, the exact `2.8.0.94d0294` fix binaries were flashed to both devices and passed another 20/20 full-NodeDB sessions each: Cardputer 199/199 nodes and Heltec 45/45 nodes, with no parser errors. Current head `a9d5be531` is rebased onto `f52e2ea8e`, includes the reconnect-tail regression and review clarifications, passes the focused coverage suite (10/10) and full native suite (583/583), and builds for both affected targets.

Total: 580 normal full-NodeDB sessions plus 32 forced-stall/replacement sessions, with zero malformed frames, incomplete NodeDBs, config timeouts, or watchdog reboots.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other: M5Stack Cardputer ADV
  - [x] Other: Heltec Wireless Tracker V2

LilyGo T-Beam and RAK WisBlock 4631 builds pass, but those targets were compile-tested rather than hardware-tested, so their hardware checkboxes remain unchecked.

Maintainer edits are enabled on the PR branch.

## Summary by CodeRabbit

- **New Features**
  - Added stateful framed output for protobuf and USB serial logging, including retained-frame resume and best-effort diagnostic log framing.

- **Bug Fixes**
  - Prevented unframed console output from mixing with protobuf/framed traffic.
  - Ensured pending framed tails complete before sending new frames to avoid corruption/reordering.
  - Improved connection-mode TX transitions and gated framed log encoding to avoid unintended raw emission.

- **Tests**
  - Added/expanded unit tests for frame writer and stream framing flow, including protobuf-mode console suppression and best-effort log gating.


p0ns:
@coderabbitai review

p0ns:
@coderabbitai full review

p0ns:
@coderabbitai full review

Issue #10982, opened 2026-07-10
[2.8.0] Clarify XEdDSA material on keygen-excluded identity restore


RCGV1:
## Problem

In builds with `MESHTASTIC_EXCLUDE_PKI_KEYGEN` but without `MESHTASTIC_EXCLUDE_PKI`, boot restores an existing 32-byte Curve25519 keypair through `CryptoEngine::setDHPrivateKey()` without first calling `regeneratePublicKey()`. The in-memory XEdDSA signing material is therefore not rebuilt.

Normal keygen-capable builds call `regeneratePublicKey()` before installing the private key and already derive this material. This issue is limited to the keygen-excluded restore path.

## Decision needed

Confirm whether XEdDSA signing is expected to operate on keygen-excluded builds that restore an existing PKI identity. If yes, loading the valid persisted identity should rebuild its matching in-memory XEdDSA private/public material and a focused regression should cover that configuration. If not, the proposed fix should not proceed.

Related: #10966
Design: meshtastic/design#122

RCGV1:
I am actively working on this fix and its regression coverage. This contributor account cannot self-assign the issue, so please treat this comment as the ownership claim.

jp-bennett:
I don't immediately see a code path where this is a problem, though it's quite possible. Can we write a test case that currently fails, that demonstrates this problem?

RCGV1:
Reproduction is now covered by a focused native crypto regression:

```cpp
void test_XEdDSA_persisted_private_key_reload(void)
{
    uint8_t curvePublic[32], persistedPrivate[32];
    uint8_t expectedEdPrivate[32], expectedEdPublic[32];
    crypto->generateKeyPair(curvePublic, persistedPrivate);
    XEdDSA::priv_curve_to_ed_keys(persistedPrivate, expectedEdPrivate, expectedEdPublic);

    CryptoEngine firstLoad;
    CryptoEngine secondLoad;
    firstLoad.setDHPrivateKey(persistedPrivate);
    secondLoad.setDHPrivateKey(persistedPrivate);

    TEST_ASSERT_EQUAL_MEMORY(expectedEdPrivate, firstLoad.xeddsa_private_key, sizeof(expectedEdPrivate));
    TEST_ASSERT_EQUAL_MEMORY(expectedEdPublic, firstLoad.xeddsa_public_key, sizeof(expectedEdPublic));
[...]

Issue #10966, opened 2026-07-09
[2.8.0] Sign plaintext licensed/ham traffic


RCGV1:
## Description

Implement the authenticated-plaintext licensed/ham-mode design from [meshtastic/design#122](https://github.com/meshtastic/design/issues/122) as a separate follow-up to the packet authenticity policy stack.

Licensed traffic must remain unencrypted and readable on air, while publishing an identity public key and carrying verifiable XEdDSA signatures on signable broadcast and directed packets.

## Required changes

- Allow identity key generation/restoration for licensed users without enabling PKI encryption.
- Stop stripping `User.public_key` from licensed NodeInfo.
- Use the normal key-derived `NodeNum = crc32(public_key)` identity and migration path.
- Sign all locally originated licensed, non-PKI `Data` packets that fit, including plaintext unicast/direct messages.
- Keep channel PSKs removed, the admin channel disabled, and `shouldUsePkiEncryption()` false for licensed traffic.
- Preserve callsign fields and existing licensed rebroadcast restrictions.
- Update the packet-policy warnings/tests so verified licensed traffic works in Strict.

## Safety and compatibility

- Existing licensed nodes without keys will undergo a one-time NodeNum migration. Preserve valid keys and warn/document the impact.
- A signature provides origin/integrity authentication, not confidentiality. Direct-message payloads must remain plaintext on the wire.
- A signed callsign-bearing NodeInfo self-binds that label to a key; it does not validate the amateur licence or replace required station identification.
- Operators remain responsible for local licence conditions and permitted modes.

## Acceptance criteria

- Licensed NodeInfo publishes a 32-byte public key and verifies through first-contact identity binding.
- Licensed broadcast and directed packets are signed whenever the encoded `Data` fits.
- Packet captures demonstrate that licensed direct-message payloads are not PKI- or channel-encrypted.
- Strict accepts verified licensed traffic and rejects unsigned/invalid/unverifiable licensed traffic.
- PKI encryption is unreachable for licensed senders even when keys are available.
- Tests cover licensed-mode entry/exit, identity migration, key preservation/regeneration, all relevant portnums/destinations, oversized packets, and plaintext wire semantics.
- Draft PR remains unmerged until representative hardware capture and cross-version testing are recorded.

## Dependencies

- Design: meshtastic/design#122
- Packet authenticity schema/policy: meshtastic/design#121 and meshtastic/protobufs#983

RCGV1:
[...]

PR #10984, opened 2026-07-10
Draft: Restore XEdDSA material in keygen-excluded identity loads

RCGV1:
Refs #10982

Related: #10966
Design: meshtastic/design#122

## Summary

- rebuild the in-memory XEdDSA private/public material when a valid persisted Curve25519 identity is installed through the keygen-excluded boot restore path;
- use the same guarded deterministic conversion already used during key generation and public-key regeneration;
- leave normal Curve25519 PKI behavior and XEdDSA-excluded targets unchanged;
- add a regression that loads one persisted identity into two fresh crypto engines, compares the exact derived XEdDSA material, then signs with one instance and verifies with the other.

Normal keygen-capable builds already call `regeneratePublicKey()` before installing the private key. This draft is intentionally limited to deciding and covering behavior for `MESHTASTIC_EXCLUDE_PKI_KEYGEN` builds that restore an existing PKI identity.

## Validation

- native macOS crypto suite: 12/12 passed;
- native macOS packet-signing suite: 29/29 passed;
- T-Echo XEdDSA-enabled build: passed, 37.5% RAM, 92.6% flash, 46 KiB warm-region clearance;
- Wio-E5 with `MESHTASTIC_EXCLUDE_XEDDSA=1`: passed, 40.7% RAM, 96.8% flash;
- independent dedicated security review approved exact head `521c62015d39ca57b8180cf7e78e62ae48a2184b`;
- Trunk and `git diff --check`: clean.

The regression fails on the pre-fix baseline because a fresh engine has no reconstructed XEdDSA material after loading a persisted Curve private key, and passes with this change.

## Draft / decision status

Keep this PR draft pending maintainer direction on whether XEdDSA signing is intended for the keygen-excluded restore path. No internal bench logs or captures are included.

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described within the validation limits above.
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Other (build-only validation on T-Echo and Wio-E5)

### DIFF
--- a/src/mesh/CryptoEngine.cpp
+++ b/src/mesh/CryptoEngine.cpp
@@ -289,6 +289,9 @@ bool CryptoEngine::decryptCurve25519(uint32_t fromNode, meshtastic_NodeInfoLite_
 void CryptoEngine::setDHPrivateKey(uint8_t *_private_key)
 {
     memcpy(private_key, _private_key, 32);
+#if !(MESHTASTIC_EXCLUDE_XEDDSA)
+    XEdDSA::priv_curve_to_ed_keys(private_key, xeddsa_private_key, xeddsa_public_key);
+#endif
 }
 
 /**
--- a/src/mesh/CryptoEngine.cpp
+++ b/src/mesh/CryptoEngine.cpp
@@ -290,6 +290,11 @@ void CryptoEngine::setDHPrivateKey(uint8_t *_private_key)
 {
     memcpy(private_key, _private_key, 32);
 #if !(MESHTASTIC_EXCLUDE_XEDDSA)
+    if (memfll(private_key, 0, sizeof(private_key))) {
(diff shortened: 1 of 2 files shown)


RCGV1:
@coderabbitai review

RCGV1:
The label check requires an allowed repository label. This contributor account cannot apply labels; a maintainer will need to add the appropriate bug label.

RCGV1:
Added a guard for blank persisted private keys so loading one clears XEdDSA signing material instead of retaining or deriving a signing identity. The native Linux simulator test now covers valid-key load followed by zero-key reload; `test_crypto` passes (13/13).

PR #10988, opened 2026-07-11
Add MEDIUM_TURBO modem preset

thebentern:
## Add the MEDIUM_TURBO modem preset

`MEDIUM_TURBO` (enum value 16) already existed in the protobufs but was never wired into firmware — selecting it silently fell through to the `LONG_FAST` default and rendered an `Invalid` display name (which also feeds channel-frequency-slot hashing). This PR wires it up.

Per the proto definition, it "performs similarly to MEDIUM_FAST, but with 500kHz bandwidth. It is not legal to use in all regions due to this wider bandwidth."

### Spec

**500 kHz** bandwidth (1625 kHz on 2.4 GHz `wideLora` regions), **SF9**, **CR 4/5** — i.e. MEDIUM_FAST's spreading factor and coding rate at double the bandwidth.

Where it lands among the sub-GHz presets by on-air throughput:

| Preset | BW kHz | SF | CR | ~Data rate |
|---|--:|--:|---|--:|
| Short Turbo | 500 | 7 | 4/5 | 21.9 kbps |
| Short Fast | 250 | 7 | 4/5 | 10.9 kbps |
| **★ Medium Turbo (new)** | **500** | **9** | **4/5** | **7.0 kbps** |
| Short Slow | 250 | 8 | 4/5 | 6.25 kbps |
| Medium Fast | 250 | 9 | 4/5 | 3.5 kbps |
| Medium Slow | 250 | 10 | 4/5 | 2.0 kbps |
| Long Turbo | 500 | 11 | 4/8 | 1.3 kbps |
| Long Fast *(default)* | 250 | 11 | 4/5 | 1.1 kbps |

3rd-fastest preset — ~2× the throughput of Medium Fast (same SF9, double bandwidth), true medium range.

### 500 kHz "wide" handling

It's added to `PRESETS_STD` (US and other wide-capable regions) but deliberately **not** to `PRESETS_EU_868` — EU868's band is only 250 kHz wide. The two existing backstops enforce this automatically: preset-list membership (`supportsPreset`) and the region-span physical-fit check. Verified: it resolves to 500/SF9/CR4:5 in US and is rejected + clamped to `LONG_FAST` in EU_868 and UNSET.

### Changes

- **Params** — `modemPresetToParams()` in `MeshRadio.h`
- **Display name** — `MediumTurbo` / `MedT` in `DisplayFormatters.cpp`
- **Availability** — appended to `PRESETS_STD` (turbo cluster) in `RadioInterface.cpp`
- **SNR grading** — MEDIUM bucket in `UIRenderer.cpp` and `CannedMessageModule.cpp`
- **Tests** — new positive coverage in `test_radio`; EU868-reject + US-accept in `test_admin_radio` and `test_mesh_beacon`; preset count 9→10; fuzz range extended to include the new value
- **Docs** — `lora_region_preset_compatibility_client_spec.md`

The InkHUD preset menu picks up the new preset automatically via its dynamic region-list path (same as `LONG_TURBO`), so no InkHUD change is needed.

### Testing

Native test suite green: **578/578 test cases, 0 failures, 0 ignored** (all 30 suites, via the Docker native-test path).

## Summary by CodeRabbit

* **New Features**
  * Added support for the `Medium Turbo` LoRa modem preset, including parameter mapping.
  * Updated signal-quality labeling and preset display naming for the new preset.
* **Documentation**
  * Updated LoRa region→preset compatibility docs to reflect a 10-preset standard group, including `MEDIUM_TURBO`.
* **Bug Fixes**
  * Improved preset validation/clamping so `MEDIUM_TURBO` is accepted in US configurations and rejected/cleared in EU 868 configurations.
* **Tests**
  * Added/updated unit tests and expanded fuzz coverage for `MEDIUM_TURBO`.


shalberd:
Hi, is it ensured somehow that the 500khz TURBO presets are not selectable on EU_868? 
Background https://www.nxp.com/docs/en/supporting-information/RDRKEBANDPL.pdf
250 Khz only

also, here on page 22, section P, band number 54 https://www.etsi.org/deliver/etsi_en/300200_300299/30022002/03.02.01_60/en_30022002v030201p.pdf
that is band from 869,400 MHz to
869,650 MHz  -> 250 Khz difference, with the default value 869.525 Mhz used by Meshtastic being right at mid-band.

Frequency slot calculator also shows no frequency slot available. 

On the Android App, I can select the TURBO presets from the dropdown list when selecting region EU_868.
It is only after SAVE that the preset is instead set to LONG_FAST.

Is it technically possible to remove some frequency presets based on the region?
In this case, for EU_868: no selectable TURBO presets?
Conditional UI rendering, I mean.

PR #10991, opened 2026-07-11
feat(gps): cache negative probe verdict — skip 76s baud scan on GPS-less boards

jamesarich:
## Why

Boards that define `HAS_GPS` + GPS pins for an *optional* module (e.g. seeed-xiao-s3, where the L76K is an expansion board) run the full exhaustive GPS baud probe on **every boot** when no module is attached: ~76 seconds of `Set GPS Baud to X` / `Trying $PDTINFO...` / `No GNSS Module` cycles before `GPS not detected; marked not present for this boot`. This delays boot-to-usable and floods the serial console (on ESP32-S3 native USB-Serial/JTAG we observed byte-level corruption in the stream during the flood).

The probe cache (`/prefs/gps_probe_cache.dat`) already persists *positive* verdicts (model+baud) — but `saveProbeCache()` explicitly skips saving when `gnssModel == GNSS_MODEL_UNKNOWN`, so the expensive negative outcome is rediscovered every boot.

## What

- `GPSProbeCacheRecord`: split the `uint16_t reserved` into `reserved` + `negative_verdict` (same 16-byte layout), bump cache version 1→2 (old caches auto-invalidated).
- `saveNegativeProbeCache()`: after the probe exhausts all bauds with no module found, persist the negative verdict.
- `loadProbeCache()`: a valid negative verdict skips the probe entirely on subsequent boots (GPS stays `GNSS_MODEL_UNKNOWN` → disabled, same end state as today, just ~76s sooner).
- Invalidation: any **position config write via AdminModule** clears the cache file (`GPS::clearProbeCacheFile()`, guarded by `#if !MESHTASTIC_EXCLUDE_GPS` like the neighboring GPS references) — so attaching a module later just requires touching position config (which reboots anyway) or deleting the file. Version bump covers upgrades.

## Testing

- `pio run -e seeed-xiao-s3` clean.
- Flashed to a physical seeed-xiao-s3 with no GPS module: first boot ran the full probe once; every subsequent observed boot (4+, including DTR-reset boots) showed zero probe activity, firmware healthy (scheduler/modules running normally).
- Positive-verdict path untouched apart from the struct/version change.

Found while running e2e sessions against develop with an MCP-driven bench harness.

## Summary by CodeRabbit

* **New Features**
  * GPS detection outcomes are now persisted across reboots, including “no GPS detected” results.
  * Devices can avoid repeated GPS probing when a prior negative result is available.
* **Bug Fixes**
  * GPS probe cache format was updated to improve reliability, and corrupt/invalid cached data is cleared automatically to allow recovery.
  * Changing GPS-related configuration now clears the cached result so fresh detection runs on the next boot.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -79,7 +79,7 @@ namespace
 {
 // Versioned on-disk record for persisted GPS probe results.
 constexpr uint32_t GPS_PROBE_CACHE_MAGIC = 0x47504348UL; // "GPCH"
-constexpr uint16_t GPS_PROBE_CACHE_VERSION = 1;
+constexpr uint16_t GPS_PROBE_CACHE_VERSION = 2;
 constexpr const char *GPS_PROBE_CACHE_FILE = "/prefs/gps_probe_cache.dat";
 constexpr int MIN_PLAUSIBLE_GPS_YEAR = 2020;
 constexpr int MAX_PLAUSIBLE_GPS_YEAR = 2100;
@@ -91,7 +91,8 @@ constexpr uint32_t T1000_E_AIROHA_WAKE_INTERVAL_MS = 40;
 struct GPSProbeCacheRecord {
     uint32_t magic;
     uint16_t version;
-    uint16_t reserved;
+    uint8_t reserved;
+    uint8_t negative_verdict;
     uint32_t baud;
     uint8_t model;
 };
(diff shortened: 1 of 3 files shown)


jamesarich:
Withdrawing this one. It came out of an e2e session where a GPS-less seeed-xiao-s3 couldn't complete the serial API handshake, and the boot-time probe was suspected as a contributor. Follow-up diagnosis showed the handshake failures were actually (a) the stream-framing corruption already fixed by #10976 and (b) client-side DTR/RTS line-state semantics on the S3 native USB-Serial/JTAG (opening with DTR asserted parks the chip in ROM download mode — being addressed on the client side). The ~76s exhaustive probe on every boot for boards without a GPS module is still real, but with the primary motivation resolved I'd rather not push extra probe-cache state machinery upstream right now. Branch stays available if the boot-time cost is worth revisiting.

PR #10964, opened 2026-07-09
fix(stm32wl): add TCXO-optional support and fix hardcoded TCXO voltage

ndoo:
## Summary

Adds TCXO-optional support to the STM32WL radio path (`STM32WLE5JCInterface`), so one `rak3172` build works on both non-T RAK3172 (XTAL) and RAK3172-T (3.0V TCXO). Also fixes a bug that silently ignored any STM32WL variant's declared TCXO voltage, hardware-confirmed on `wio-e5`.

## Problem

1. **Hardcoded 1.7V, ignoring `SX126X_DIO3_TCXO_VOLTAGE`.** Every variant except `rak3172` got `setTCXOVoltage(1.7)` unconditionally. `russell` declares `SX126X_DIO3_TCXO_VOLTAGE 3.0`, but it was silently overridden to 1.7V.
2. **`rak3172` configured no TCXO at all.** RAK3172-T (populated 3.0V TCXO) failed radio init outright.

Confirmed on TCXO-equivalent test hardware: RadioLib error `-707` (`RADIOLIB_ERR_SPI_CMD_FAILED`) on stock `develop`. RAK's own [`radio_conf.h`](https://github.com/RAKWireless/RAK-STM32-RUI/blob/e5a28be8fab1a492bd9223dd425ca33a8a297d90/variants/WisDuo_RAK3172-T_Board/radio_conf.h#L91) confirms RAK3172-T needs 3.0V.

```mermaid
flowchart TD
    A["STM32WLE5JCInterface::init()"] --> B{"_VARIANT_RAK3172_ defined?"}
    B -- "yes (rak3172)" --> C["No TCXO configured"]
    B -- "no (every other variant)" --> D["setTCXOVoltage(1.7), hardcoded"]
    D -.-> N["russell/milesight_gs301 declare 3.0V,\nget 1.7V instead — outcome unverified"]
    C --> E["lora.begin()"]
    D --> E
    E --> F{"Success?"}
    F -- "yes" --> G["Radio up"]
    F -- "no" --> H["Fails, no retry.\nRAK3172-T: confirmed fails (-707)"]
```

## Fix

- `STM32WLE5JCInterface::init()` reads `SX126X_DIO3_TCXO_VOLTAGE` per variant instead of hardcoding 1.7V. No macro → no TCXO call (XTAL).
[...]

PR #10961, opened 2026-07-09
stm32wl: add hardware RTC support (rak3172)

ndoo:
## Summary

Adds STM32WL internal hardware RTC support, wired into `perhapsSetRTC()`/`readFromRTC()` (`src/gps/RTC.h`/`RTC.cpp`) alongside the existing I2C RTC chip branches. Time survives a reset on `rak3172` without GPS/NTP/phone re-sync, since the STM32 RTC calendar lives in the battery/VDD-backed domain.

- `HAS_LSE` / `STM32WL_LSE_DRIVE` are new `variant.h` flags, following the existing `HAS_*` convention. Only `rak3172` opts in here.
- LSE bring-up avoids a known library hang:
  - LSE is started and polled manually, with a bounded timeout, before calling into `STM32RTC`.
  - The library's own internal LSE startup has no bounded fallback — it hangs via `Error_Handler()` (`while(1){}`) if the crystal never locks.
  - Doing our own bounded check first avoids that hang on a missing/faulty crystal.
  - On timeout, LSE is explicitly turned off (`RCC_LSE_CONFIG(RCC_LSE_OFF)`), not left running unlocked.
- Other STM32WL variants are not opted in:
  - Insufficient Flash headroom:
    - `wio-e5` (GPS): 96.8% flash, ~8KB headroom.
    - `russell` (GPS + BME280): 98.3% flash, ~4KB headroom.
  - `CDEBYTE_E77-MBL`/`milesight_gs301` are out of scope here due to lack of hardware or time to test
    - Opting any of them in later is the same two-line variant change.

## Commits

1. `stm32wl: add hardware RTC support infrastructure`
   `HAS_LSE` scaffolding, LSE bring-up (`stm32wlSetup()`), `readFromRTC()`/`perhapsSetRTC()` wiring. `HAS_LSE` defaults to `0`, so this is inert everywhere until a variant opts in.
2. `gps: qualify RTC.h includes to avoid case-insensitive filesystem collision with STM32RTC`
   - `STM32RTC` ships its own `src/rtc.h`.
   - On case-insensitive filesystems, PlatformIO's LDF orders `lib_deps` include paths ahead of `-Isrc/gps`.
   - So an unqualified `#include "RTC.h"` outside `src/gps/` resolves to the library's header once `STM32RTC` is linked in anywhere in the project.
   - Qualified all ~46 call sites to `gps/RTC.h`. Mechanical, no behavior change on case-sensitive filesystems (Linux CI).
[...]

PR #10993, opened 2026-07-11
stm32wl(rak3172): implement cpuDeepSleep()/shutdown() via STM32LowPower

ndoo:
## Summary

Implements deep-sleep/shutdown for STM32WL (`rak3172`) using the `STM32LowPower` Arduino library, building on #10961's `HAS_LSE`/`STM32RTC` infrastructure.

- `cpuDeepSleep()` (`src/platform/stm32wl/main-stm32wl.cpp`) was an empty stub, so the SDS deep-sleep PowerFSM state and low-battery shutdown never actually slept.
- `Power::shutdown()` didn't include STM32WL in its arch list, so an explicit shutdown command just logged a FIXME warning.

Both gaps are closed for `rak3172`, gated behind the existing `HAS_LSE` flag (default 0, only set for `rak3172`). The main reason to only implement it in RAK3172 is because the remaining FLASH memory is too tight on other variants, including the RTC and LowPower libraries overflow on most other STM32WL variants.

In future, this can (and should) be rolled out to all STM32WL variants that have a hardware LSE crystal (most of them, it seems).

## Dependency

This PR is based on #10961 (open, not yet merged). `stm32wl-hardware-rtc` (its head branch) only exists on the `mesh-malaysia` fork, not upstream, so this PR targets `develop` directly — the diff currently includes #10961's commits too, until #10961 merges.

## Implementation

Both `cpuDeepSleep()` paths use `LowPower.shutdown()` (Standby mode) — with an RTC alarm for a finite wake (SDS/low-battery), without one for a forever shutdown (`Power::shutdown()`).

## Flash budget

| | Flash |
|---|---|
| `rak3172` before | 76.7% (190,100 / 247,808) |
| `rak3172` after | 77.1% (191,452 / 247,808) |
| `wio-e5` before/after | 96.8%, byte-identical |

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Tested on a RAK3172-T unit, with PR #10964's TCXO fixes cherry-picked in for that testing session (not included in this PR — a RAK3172-T needs #10964 to boot at all, but a base RAK3172 does not, and this PR's changes don't depend on it). Confirmed on hardware:
- Explicit shutdown (`Power::shutdown()`, Standby mode, no alarm): device saves state, sleeps the radio, enters Standby, and stays silent until manually reset.
- Low-battery deep sleep (Standby mode with RTC alarm, finite wake): `low_voltage_counter` debounce fires `EVENT_LOW_BATTERY`, device enters deep sleep, then resets and rejoins the mesh on its own once `sds_secs` elapses.

Build-verified no regression on `wio-e5`, `russell`, `CDEBYTE_E77-MBL` (`HAS_LSE` defaults to `0`, all new code compiled out). `milesight_gs301` fails to build on `develop` independent of this PR (pre-existing `NANOLIB_FLOAT_PRINTF`/debug-print config issue, confirmed via a clean checkout).

## Summary by CodeRabbit

* **New Features**
  * Added reliable deep-sleep support for STM32WL devices with hardware RTC availability checks.
  * STM32WL devices now support timed wake-up and indefinite deep sleep.
* **Bug Fixes**
  * Shutdown now consistently turns off indicator LEDs before entering deep sleep on supported STM32WL hardware.
  * Added safeguards to reset the device if deep-sleep recovery does not complete as expected.

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -906,7 +906,7 @@ void Power::shutdown()
 #if HAS_SCREEN
     messageStore.saveToFlash();
 #endif
-#if defined(ARCH_NRF52) || defined(ARCH_ESP32) || defined(ARCH_RP2040)
+#if defined(ARCH_NRF52) || defined(ARCH_ESP32) || defined(ARCH_RP2040) || defined(ARCH_STM32WL)
 #ifdef PIN_LED1
     ledOff(PIN_LED1);
 #endif
--- a/src/platform/stm32wl/main-stm32wl.cpp
+++ b/src/platform/stm32wl/main-stm32wl.cpp
@@ -7,6 +7,7 @@
 #include <stm32wlxx_hal.h>
 
 #if HAS_LSE
+#include <STM32LowPower.h>
 #include <STM32RTC.h>
 
 // LSEDRV is a 2-bit RCC_BDCR field where every combination is a legal drive level, so this covers all 4 values.
@@ -120,6 +121,7 @@ void stm32wlSetup()
         rtc.setClockSource(STM32RTC::LSE_CLOCK);
         rtc.begin();
         stm32wlRtcValid = true;
+        LowPower.begin();
         LOG_INFO("STM32WL: LSE locked, hardware RTC available");
     } else {
         // Don't leave a failed oscillator burning current.
@@ -138,7 +140,24 @@ bool stm32wlRtcAvailable()
 void stm32wlSetup() {}
 #endif
 
-void cpuDeepSleep(uint32_t msecToWake) {}
+void cpuDeepSleep(uint32_t msecToWake)
+{
+#if HAS_LSE
+    if (!stm32wlRtcAvailable())
+        return;
+
+    if (Serial)
+        Serial.end();
+
+    if (msecToWake != portMAX_DELAY) {
+        LowPower.shutdown(msecToWake);
+    } else {
+        LowPower.shutdown();
+    }
+    // RTC wakes from shutdown into MCU reset, so this code should never be reached
+    HAL_NVIC_SystemReset();
+#endif
+}
 
 // Hacks to force more code and data out.
 
--- a/src/platform/stm32wl/main-stm32wl.cpp
+++ b/src/platform/stm32wl/main-stm32wl.cpp
@@ -143,11 +143,21 @@ void stm32wlSetup() {}
 void cpuDeepSleep(uint32_t msecToWake)
 {
 #if HAS_LSE
-    if (!stm32wlRtcAvailable())
-        return;
+    if (!stm32wlRtcAvailable()) {
+        // Hardware can't shutdown, but firmware has already prepared itself for shutdown
+        // Do not leave the device unresponsive, reset instead
+        LOG_WARN("STM32WL: hardware RTC failed, cannot deep sleep/shutdown");
+        if (Serial) {
+            Serial.flush();
+            Serial.end();
+        }
+        HAL_NVIC_SystemReset();
+    }
 
-    if (Serial)
+    if (Serial) {
+        Serial.flush();
         Serial.end();
+    }
 
     if (msecToWake != portMAX_DELAY) {
         LowPower.shutdown(msecToWake);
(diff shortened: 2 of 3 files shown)


ndoo:
@coderabbitai review

Only commit 79d126add6de98827fe47182d0749cb50dc30f9b should be reviewed (IMHO) because the commits preceding it were reviewed in #10961

ndoo:
> One thing worth a second look, in `cpuDeepSleep()` (`src/platform/stm32wl/main-stm32wl.cpp`, Line 134-onward):
> 
> ```c++
> if (!stm32wlRtcAvailable())
>     return;
> ```
> 
> If the LSE crystal fails to lock at boot (e.g., missing/faulty crystal on a given `rak3172` unit), this makes every later `cpuDeepSleep()` call — both explicit `Power::shutdown()` and low-battery SDS — a silent no-op. `doDeepSleep()`/`Power::shutdown()` still run through their full flow (LEDs off, radio off, `nodeDB->saveToDisk()`, `notifyObservers()`, etc.), but the MCU never actually enters Standby or resets — it just falls through and keeps running as if nothing happened. `stm32wlSetup()` does log a `LOG_WARN` once at boot when the LSE fails, but there's nothing logged at the actual moment a shutdown/deep-sleep is attempted and silently skipped, which could be confusing when debugging a unit that "shut down" but stayed alive.
> 
> This isn't a regression (the previous empty stub was equally silent), and it's an edge case gated behind a boot-time hardware fault, so it's not blocking. A small improvement would be to add a `LOG_WARN` right before the early `return` so it's visible in the logs at the time of the failed shutdown attempt, not just at boot:
> 
> ```c++
> if (!stm32wlRtcAvailable()) {
>     LOG_WARN("STM32WL: hardware RTC unavailable, cannot deep sleep/shutdown");
>     return;
> }
> ```

1. Added the logging as suggested, thank you
2. Added Serial flushes before both Serial.end() to ensure logging from here or elsewhere successfully make it to the receiving device.
3. Reset the MCU to avoid the user ending up with a stalled device

@coderabbitai review

ndoo:
Pushed

@coderabbitai review

ndoo:
> The three changes you mentioned are all there and look good:
> 
> * `LOG_WARN("STM32WL: hardware RTC failed, cannot deep sleep/shutdown")` is now logged right at the failed shutdown/deep-sleep attempt (Line 138), in addition to the boot-time warning in `stm32wlSetup()`. This resolves the "silent no-op" concern I raised.
> * `Serial.flush()` calls before `Serial.end()` in both branches should help ensure any pending log output (including the new warning) actually reaches the receiving terminal before the UART is torn down.
> * The no-RTC path now calls `HAL_NVIC_SystemReset()` instead of just returning, so a unit with a failed LSE won't get stuck "silently alive" after a shutdown attempt — it'll cycle instead.
> 
> Two small nits, neither blocking:
> 
> 1. In the no-RTC branch (Lines 138-141), `Serial.flush()`/`Serial.end()` are called unconditionally, while the normal path a few lines below guards the flush with `if (Serial)` (Line 144). For consistency (and to avoid touching an uninitialized/disabled `Serial` object on boards where it's compiled out), you could mirror the same guard in the no-RTC branch.
> 2. Like the final fallback reset at Line 154, the reset in the no-RTC branch relies on `HAL_NVIC_SystemReset()` never returning; a short comment (or restructuring as an `else`) would make it clearer that falling through into `LowPower.shutdown()` without a working RTC is unintended, matching the documentation style already used for the bottom fallback.

Force-pushed for a few fixes:

1. Done and fixed an indentation/scope error for the second Serial.end()
2. Comment added

@coderabbitai review

thebentern:
Please fix the merge conflicts

ndoo:
> Please fix the merge conflicts

#10961 must merge before this PR btw, so I will fix the conflicts there too first

ndoo:
> Please fix the merge conflicts

fixed the merge conflicts in #10961 then rebased this PR on that cleanly

ndoo:
rebased this on HEAD of develop now that #10961 has been merged

PR #10996, opened 2026-07-12
Reduce flash usage of GPS variants by deferring unused GeoCoord conversions

ndoo:
## Motivation

`wio-e5` is at 97.0% flash usage on current `develop` (7,508 B / 3.03% free of the 247,808 B budget). This PR is part of a search for flash reductions on that platform that don't cost an existing feature. `GeoCoord`'s eager UTM/MGRS/OSGR/OLC computation was the largest lever found: none of that math is reachable on `wio-e5` today (`NMEAWPL.cpp` only reads DMS), so it's paid for on every boot without ever being used there.

## Change

`GeoCoord`'s constructor unconditionally computes all five coordinate representations (DMS, UTM, MGRS, OSGR, OLC) via `setCoords()`, regardless of which one a caller actually reads. The UTM conversion pulls in a chain of libm trig functions (`atan`, `__kernel_tan`, `__ieee754_acos`, `__ieee754_pow`, `__kernel_rem_pio2`/`__ieee754_rem_pio2`) that then get linked in even when nothing is ever displayed.

The only screen consumer (`UIRenderer.cpp`'s GPS coordinate display) dispatches on a single configured format and reads exactly one representation per call.

This PR keeps DMS eager and defers UTM/MGRS/OSGR/OLC to first access via their own getters, tracked with per-representation mutable dirty flags. A caller that never touches a given representation never pulls in its conversion code. No getter's output changes — values are computed on demand instead of upfront.

This applies to any headless (`HAS_SCREEN=0`), GPS-enabled board where `NMEAWPL.cpp`'s NMEA/CalTopo serial export is the only `GeoCoord` consumer, since it only reads the DMS getters. Measured against current `develop`:

| Board | Flash before | Flash after | Flash saved | RAM delta |
|---|---|---|---|---|
| `wio-e5` | 240,300 / 247,808 B (97.0%) | 232,008 / 247,808 B (93.6%) | **8,292 B** | unchanged |
| `tracker-t1000-e` | 506,784 / 815,104 B (62.2%) | 499,528 / 815,104 B (61.3%) | **7,256 B** | unchanged |
| `EBYTE_ESP32-S3` | 2,075,695 / 2,424,832 B (85.6%) | 2,064,923 / 2,424,832 B (85.2%) | **10,772 B** | +16 B (~0.003%, negligible) |

Updated `test_geocoord_extreme_coords_no_oob` (regression test for a prior UTM/MGRS out-of-bounds crash on extreme lat/lon) to explicitly call each representation's getter, since it previously relied on the constructor triggering all four conversions eagerly.

## 🤝 Attestations

  - Native test suite: 586/586 pass, including the updated OOB test under ASan
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card (build-verified only, see flash numbers above — not hardware-flashed)
    - [x] `wio-e5`: flashed over SWD, clean first boot, no crash loop
    - [x] Seeed Wio Tracker L1 Pro (screen-equipped, where the deferred getters are actually reachable): flashed via UF2, clean boot; fixed test position, all six coordinate formats (DEC/DMS/UTM/MGRS/OSGR/OLC) cycled on-device via joystick — all render correctly, no crash (OSGR correctly shows "Out of Boundary" outside its UK-only coverage)

## Summary by CodeRabbit

* **Performance Improvements**
  * Coordinate conversions (UTM, MGRS, OSGR, OLC) are now generated on-demand the first time they’re requested, avoiding unnecessary work.

* **Bug Fixes**
  * Improved reliability of coordinate handling for extreme latitude/longitude values across UTM, MGRS, OSGR, and OLC.

* **Tests**
  * Updated extreme-coordinate test coverage to explicitly exercise each coordinate representation through its public getters, improving safety checks (including out-of-bounds protection).

### DIFF
--- a/src/gps/GeoCoord.cpp
+++ b/src/gps/GeoCoord.cpp
@@ -36,17 +36,50 @@ GeoCoord::GeoCoord(double lat, double lon, int32_t alt) : _altitude(alt)
     GeoCoord::setCoords();
 }
 
-// Initialize all the coordinate systems
+// Initialize DMS eagerly (cheap, most commonly used); UTM/MGRS/OSGR/OLC are computed lazily on
+// first access via their getters (see ensure*() below), since most callers never touch them.
 void GeoCoord::setCoords()
 {
     double lat = _latitude * 1e-7;
     double lon = _longitude * 1e-7;
     GeoCoord::latLongToDMS(lat, lon, _dms);
-    GeoCoord::latLongToUTM(lat, lon, _utm);
-    GeoCoord::latLongToMGRS(lat, lon, _mgrs);
-    GeoCoord::latLongToOSGR(lat, lon, _osgr);
-    GeoCoord::latLongToOLC(lat, lon, _olc);
+    _utmValid = false;
+    _mgrsValid = false;
+    _osgrValid = false;
+    _olcValid = false;
     _dirty = false;
+}
+
+void GeoCoord::ensureUTM() const
+{
+    if (!_utmValid) {
+        GeoCoord::latLongToUTM(_latitude * 1e-7, _longitude * 1e-7, _utm);
+        _utmValid = true;
+    }
+}
+
+void GeoCoord::ensureMGRS() const
+{
+    if (!_mgrsValid) {
+        GeoCoord::latLongToMGRS(_latitude * 1e-7, _longitude * 1e-7, _mgrs);
+        _mgrsValid = true;
+    }
+}
+
+void GeoCoord::ensureOSGR() const
+{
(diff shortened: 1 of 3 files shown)


ndoo:
Back to draft for further build-testing and to address the nitpicks

ndoo:
@coderabbitai review, nitpicks possibly addressed

PR #10999, opened 2026-07-12
Fix double free in virtual keyboard

jp-bennett:
Found via portduino. As there are two separate references to this pointer, they both could call free

Actual credit goes to my daughter for triggering the crash. =D

Potential backport to 2.7

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved virtual keyboard cleanup when input sessions time out or are canceled.
  * Helps prevent inconsistent keyboard state and ensures related input handling continues reliably.

### DIFF
--- a/src/graphics/draw/NotificationRenderer.cpp
+++ b/src/graphics/draw/NotificationRenderer.cpp
@@ -1038,10 +1038,9 @@ void NotificationRenderer::drawTextInput(OLEDDisplay *display, OLEDDisplayUiStat
             // Cancel virtual keyboard - call callback with empty string to indicate timeout
             auto callback = textInputCallback; // Store callback before clearing
 
-            // Clean up first to prevent re-entry
-            delete virtualKeyboard;
-            virtualKeyboard = nullptr;
-            textInputCallback = nullptr;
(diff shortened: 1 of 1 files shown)


src/graphics/draw/NotificationRenderer.cpp
```diff
@@ -1038,10 +1038,9 @@ void NotificationRenderer::drawTextInput(OLEDDisplay *display, OLEDDisplayUiStat
             // Cancel virtual keyboard - call callback with empty string to indicate timeout
             auto callback = textInputCallback; // Store callback before clearing
 
-            // Clean up first to prevent re-entry
-            delete virtualKeyboard;
-            virtualKeyboard = nullptr;
-            textInputCallback = nullptr;
+            // Clean up first to prevent re-entry. The keyboard belongs to OnScreenKeyboardModule; only stop()
+            // may free it, and it clears virtualKeyboard/textInputCallback for us.
+            OnScreenKeyboardModule::instance().stop(false);
```
thebentern, line 1043: Any null checks needed on this instance?

jp-bennett: It's a static object, so I don't *think* so? Probably also a bigger refactor than this straightforward fix. This instance call was already used a few lines above, so it's as least as safe as it already was.

PR #11019, opened 2026-07-16
fix: heap leaks on display wake and BLE re-enable; remove dead PacketCache

thebentern:
## Why

Field reports on 2.8 nightlies (and going back to 2.7.12) of Heltec Wireless Tracker V2 heap climbing steadily — one tester saw utilization go from a stable 78% to 92% within a day. A deep audit of packet handling, TMM, GPS, NodeDB, PhoneAPI, and the display stack traced the dominant cause to the display wake path, plus two smaller instances of the same bug class found in a follow-up sweep.

## Root cause 1: TFT driver instance leaked per screen wake

On variants where `Screen::handleSetOn()` re-runs `ui->init()` on every display wake (`HELTEC_TRACKER_V1_X`, `HELTEC_WIRELESS_TRACKER_V2` at Screen.cpp:635, plus the `USE_ST7789`+`VTFT_LEDA` and `USE_ST7796` gates, and MUZI/Cardputer via `dispdev->init()`), the OLED library's `init()` unconditionally calls `connect()` — and `TFTDisplay::connect()` did an unguarded `tft = new LGFX`, orphaning the previous driver instance every OFF→ON transition. PowerFSM wakes the screen on every received text, so leak rate scales with channel activity.

**Bench measurement (Heltec Wireless Tracker V2, self-driving wake loop, passive serial monitor):**

| build | free heap per wake cycle |
|---|---|
| develop, unfixed | **−1,260 bytes every wake** (9 samples, zero variance; 59,248 → 49,168 over 8 cycles) |
| develop + this fix | **flat at 67,712 across 13 wakes** |

~30 field wakes/day ≈ 38KB/day — matching the tester's 78→92% climb. With the guard, `connect()` still runs `tft->init()` on each wake and the panel re-initializes and renders normally (verified on hardware: `Do TFT init` logged per wake, display works, heap flat).

## Root cause 2: BLE re-enable leaks (`NimbleBluetooth`)

`setup()`/`setupService()` re-run when Bluetooth is re-enabled after `deinit()` (reachable without reboot). `BLEDevice::deinit()` frees the GATT objects but not our caller-owned ones, which were re-allocated unguarded each cycle: `bluetoothPhoneAPI` (~3.5KB, and the orphan is an OSThread that stays scheduled — a duplicate thread servicing the same static queues), the four callback objects, and a `BLESecurity` shim that leaked even on first boot. Reuse the objects; `setCallbacks()` still runs each time since the characteristics are new.

## Cleanup: remove dead `PacketCache`

Landed in #8341 with no consumer ever wired up (only self-references repo-wide). It also mallocs per packet with no cap or eviction, so it needs a bounds re-review before any future use — history preserves it.

## Verification

- Native tests: 600/600 pass (native-macos) + Docker suite green
- `heltec-wireless-tracker-v2` target build green
- Hardware A/B on a bench Tracker V2 as above; device runs the fixed build with normal config now
- Changed hunks verified clean against trunk's clang-format 16.0.3 config

## Follow-ups (out of scope here)

- `EInkDisplay::connect()` has the same latent pattern in every branch (no wake gate reaches it today; needs its own careful pass since branches mix construction with per-board power/SPI bring-up)
- Vendored OLED lib: `OLEDDisplay::allocateBuffer()` nulls `logBuffer` without freeing on re-init, which also permanently disables on-screen log lines on wake-reinit boards — fix belongs in meshtastic/esp8266-oled-ssd1306

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved display wake/reconnect stability by initializing the platform display driver only once and reusing the existing instance on subsequent connects.
  * Enhanced Bluetooth reliability by centralizing BLE/PhoneAPI session teardown, preventing late notifications after teardown, and tightening connection/RSSI handling.
* **Refactor**
  * Removed the legacy packet caching system and its associated cache logic and data structures.


caveman99 (approved):

thebentern (commented):
## Self-review (Copilot-style pass over the diff)

Ran an adversarial multi-angle review of this PR before requesting human review. Findings and dispositions:

### Fixed in c7a1bbc6e / 7c072d418

1. **`NimbleBluetooth.cpp` — reused `bluetoothPhoneAPI` could carry stale session state (correctness, the important one).** The only full cleanup path (`onDisconnect`) is skipped when `deinit()`'s 2s disconnect wait expires before the event lands — and that 2s cap equals the connection supervision timeout set in `onConnect`, so a phone that walked out of range makes the skip a coin flip. The pre-PR code accidentally masked this by giving each session a fresh (leaked) object; the reuse in this PR would have handed the next session a mid-config state machine, stale `toPhoneQueue` frames served to the new phone's first reads, a stale `lastToRadio` that drops the new session's first write as a duplicate, and a stale conn handle that defeats `checkConnectionTimeout`. Fixed by factoring `onDisconnect`'s cleanup into `resetBleSessionState()` and running it on the reuse path.
2. **`NimbleBluetooth.cpp:863` — `getRssi()` connection check inverted** (pre-existing, adjacent): returned 0 for every active connection, then queried `ble_gap_conn_rssi()` with `BLE_HS_CONN_HANDLE_NONE` when disconnected. BLE RSSI always read 0 on S3/C6. One-character fix.
3. **Guarded heap pointers → function-local static objects** for the four stateless callback objects (matches the by-value idiom in `NRF52Bluetooth.cpp`); verified against the *resolved* BLE library (pioarduino framework wrapper — `NimBLE-Arduino` is in `lib_ignore`) that `setCallbacks`/`setSecurityCallbacks` store plain pointers and never free them, so static instances are safe.
4. **Comment overclaimed** "deinit() frees the GATT objects": it deletes only the `BLEServer` object; the services/characteristics/`BLE2904` created in `setupService()` still leak per enable cycle inside the framework wrapper (`BLEServer` has no destructor cascade). Corrected the comment; that residual is a framework-level issue this PR can't cleanly fix.
5. **`TFTDisplay.cpp` — single construct-once guard around the whole `#if` ladder** instead of per-branch guards (a future `#elif` can't reintroduce the leak by copying an unguarded branch), `bus` made local (was a write-only global), comment now points at the `Screen::handleSetOn` gates instead of a board list that would rot.

### Reviewed and intentionally not changed

- **Warm-path wake cost** (`tft->init()` + `fillScreen` re-run per wake on the reused instance): identical to pre-PR behavior on the fresh instance — kept for exact behavior parity in a leak fix. Skipping panel re-init on non-power-cut boards is a possible follow-up but needs per-board testing (these gates exist precisely because some panels need the full "kick").
- **HACKADAY `begin()` retry on the same objects after a transient first-boot failure**: that variant has no wake gate, so `connect()` runs once per boot there; `Arduino_GFX::begin()` is re-runnable regardless.
- **Deeper altitude** (make wake re-init a first-class `reinitPanel()` instead of re-running library `init()`): tracked as a follow-up along with the vendored OLED lib's `logBuffer` orphan (`allocateBuffer()` nulls it without freeing, which also permanently kills on-screen logs on wake-reinit boards — fix belongs in meshtastic/esp8266-oled-ssd1306) and `EInkDisplay::connect()` idempotency hardening.

Verification after review fixes: heltec-wireless-tracker-v2 build green; native suites unaffected (ESP32-only code paths); changed hunks clean against trunk's clang-format 16.0.3 config.

thebentern:
## Follow-up: use-after-free found in the BLE teardown path (fixed in 8c0b5764f)

A deeper lifecycle audit of the re-enable changes surfaced a real UAF in the *teardown* counterpart, `deinit()`:

- `deinit()` nulled `bleServer`/`BatteryCharacteristic` but left **`fromNumCharacteristic` dangling** after `BLEDevice::deinit(true)` frees the GATT graph, and never detached the PhoneAPI `fromNumChanged` observer.
- When `deinit()`'s bounded 2 s disconnect wait expires before `onDisconnect` runs (the same stale-bond/host-reset race `resetBleSessionState()` was built for), `close()` is skipped, so the observer stays attached with `state == STATE_SEND_PACKETS`.
- After BLE is off, the next mesh packet drives `MeshService fromNumChanged → PhoneAPI::onNotify → onNowHasData → fromNumCharacteristic->notify()` on freed memory — and the framework BLE wrapper's `notify()` dereferences the freed server via `getConnectedCount()`. Unlike `sendLog()`, `onNowHasData()` had no `isConnected` guard.
- Reachable on **all ESP32/S3/C6 NimBLE boards** via admin disable-bluetooth or a sleep transition while a phone is connected.

Fix: `deinit()` now calls `resetBleSessionState()` (detaches the observer, resets session state, forces the conn handle to `NONE` so `checkConnectionTimeout` can't be fooled by the stale handle), nulls the two freed characteristic pointers like the others, and `onNowHasData()` bails on a null characteristic. Build green on heltec-wireless-tracker-v2.

### Other lifecycle findings from the same audit — filed separately, not in this PR

- **[security, medium]** Lockdown builds: admin authorization persists across serial-console client swaps (the `SerialConsole` PhoneAPI singleton is reused and only re-locks on a 15-min timeout). Needs a per-link re-lock **and a security review** — kept out of this PR deliberately.
- **[low]** Live-toggling `double_tap_as_button_press` / `wake_on_tap_or_motion` off never stops `accelerometerThread` (runs until reboot, becomes un-restartable).
- **[low, nRF52]** `NRF52Bluetooth::shutdown()` latches the refuse-pairing callback; `resumeAdvertising()` never restores it (only reachable today via the PowerStress test opcodes).

Audit cleared as safe: nRF52 BLE enable/disable (never deletes the singleton), MQTT reconnect (unique_ptr + reused client), TCP/WiFi/Eth API (fresh PhoneAPI per connection), all `.observe()` sites (each runs once/boot), and light-sleep attachInterrupt pairing.

PR #11020, opened 2026-07-16
WarmNodeStore: carry the XEdDSA signer flag through the warm tier

caveman99:
The warm tier already caches an evicted node's role and protected category, but not
whether we ever verified an XEdDSA signature from it. That flag is learned from
verified traffic rather than from NodeInfo, so a node that was evicted and later
re-admitted came back with it clear until the next signed packet arrived.

Cache it in the same packed word: `last_heard` now carries 7 metadata bits (role 4,
protected 2, signer 1) and `getOrCreateMeshNode()` restores the flag on re-admission
alongside the role and key. `WarmNodeEntry` stays 40 B.

Bit 6 was part of the quantised timestamp under the v2 record format, so a v2 record
read with the new layout would report a signer about half the time. The record magic
moves to WRN3/WRM3 and `load()` clears that bit on v2 records on both backends
(raw-flash ring and warm.dat), rewriting them in the current format. v2 keeps its
role/protected bits; v1 handling is unchanged (identity + key kept, last_heard
discarded). The ring forces a rotation when the head page predates v3 so new records
never land in an older-format page.

Tests: signer round trip through absorb/take with a cleared-flag control; a v2
snapshot migration asserting a stale timestamp bit does not read as a signer; and an
eviction/re-admission round trip through NodeDB with a non-signer control.

Note for #10967, which touches the same metadata: it adds `WarmProtected::XeddsaSigner`
as a fourth category value for this purpose. Since `WarmProtected` holds one value,
that makes the flag exclusive with the existing categories, and `warmProtectedCategory()`
returns it ahead of the favourite/ignored/manually-verified check, so a favourite or a
TRACKER that has signed comes back without its protected category. A dedicated bit keeps
the two independent. `hasSeenXeddsaSigner()` from that PR works unchanged on top of this
via `lookupMeta`; only the category value needs dropping on rebase.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved XEdDSA signer status across hot-to-warm and warm-to-hot node transitions.
  * Ensured warm-node persistence/reload correctly retains or clears the signer bit depending on the stored format.
  * Enhanced warm storage migration for older on-disk layouts to prevent signer-bit misreads from timestamp metadata.
* **Tests**
  * Added warm-tier migration coverage to verify signer-bit survival without affecting non-signer nodes.
  * Added warm store round-trip and legacy-format migration tests for signer-bit behavior.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -1852,7 +1852,8 @@ void NodeDB::cleanupMeshDB()
                 // Keep any key we learned (e.g. via a DM before the NodeInfo
                 // exchange completed) rather than losing it with the purge.
                 if (n.public_key.size == 32)
-                    warmStore.absorb(gone, n.last_heard, n.public_key.bytes, n.role, warmProtectedCategory(n));
+                    warmStore.absorb(gone, n.last_heard, n.public_key.bytes, n.role, warmProtectedCategory(n),
+                                     nodeInfoLiteHasXeddsaSigned(&n));
 #endif
 
                 eraseNodeSatellites(gone);
@@ -2036,7 +2037,7 @@ void NodeDB::demoteOldestHotNodesToWarm()
         // Keep the public key if we have one (40 B warm record); keyless nodes
         // still get a placeholder so re-admission restores last_heard.
         warmStore.absorb(n.num, n.last_heard, n.public_key.size > 0 ? n.public_key.bytes : nullptr, n.role,
-                         warmProtectedCategory(n));
+                         warmProtectedCategory(n), nodeInfoLiteHasXeddsaSigned(&n));
         // Demotion drops the node from the header table, so drop its satellites
         // too (the eviction chokepoint) - they'd otherwise orphan until the next
         // enforceSatelliteCaps pass.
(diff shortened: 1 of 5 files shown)


caveman99:
Both addressed in 990297afb.

**Comment length**: condensed every comment this PR adds or touches to two lines, per the one-or-two-line guideline. Skipped one site from the list: the corrupt-header rationale in `WarmNodeStore.cpp` load() is pre-existing and unchanged here (only the adjacent `bool legacy` -> `WarmFormat fmt` line moved), so shortening it would be unrelated churn.

**Timestamp assertion**: added, good catch. The test claimed time carries over but only checked the flag bits, so a mask error in the upper bits would have gone unnoticed. Note the fixture happens to make this sharper than it looks: 123456 has bit 6 set in its own timestamp, so the assertion also pins that `warmPackLastHeard()` masks the time before OR-ing the signer flag in.

thebentern (approved):

caveman99:
Copilot findings addressed in 87a94c6f3:

- **`warmSignerOf` implicit bool→bool** (WarmNodeStore.h): now returns an explicit `!= 0` comparison.
- **`legacy` duplicating `fmt`** (WarmNodeStore.cpp load()): `legacy` is now derived as `fmt != WarmFormat::Current`, and a separate `known` flag drives the reject, so the format decision lives in one place.
- **Hard-coded `<< 6`** (test_warm_store): the LRU-spacing test now shifts by `WARM_META_BITS`, so it tracks the quantum instead of a stale literal.
- **Unchecked short read/write** (v2 migration test): the read and write byte counts are now asserted against `buf.size()`.

Left the identical read/write pattern in the pre-existing v1-migration and oversized-snapshot tests as-is to keep this diff to the signer change; happy to tighten those in a follow-up if you'd like them consistent.

NomDeTom (approved):
Lgtm

PR #11022, opened 2026-07-16
NodeDB: clear only the slots removeNodeByNum() vacates

caveman99:
`removeNodeByNum()` compacts the survivors down and then clears the slot the compaction
left free. It always cleared exactly one, whether or not anything was removed:

```c
numMeshNodes -= removed;
std::fill(nodeDatabase.nodes.begin() + numMeshNodes,
          nodeDatabase.nodes.begin() + numMeshNodes + 1, meshtastic_NodeInfoLite());
```

`nodeDatabase.nodes` is sized to MAX_NUM_NODES, so when the requested node is not present
`numMeshNodes` is unchanged, and on a full store that clear lands one entry past the last
slot. It was also short in the other direction: removing several nodes with the same num
vacated `removed` slots but only ever cleared one, leaving stale copies behind.

Size the clear from `removed` instead, so nothing is written when nothing was removed, all
vacated slots are cleared when they are, and the range is clamped to the vector.

The sibling clear in `cleanupMeshDB()` is already correct and is left alone: its range ends
at `numMeshNodes + removed`, which is the pre-compaction count and therefore never past the
end.

Tests: removal of an absent node from a full store (the coverage env builds with ASan, so a
write past the last slot fails the run), plus a control removing a present node from a full
store to show compaction still works.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node removal and table compaction to correctly clear all freed slots, including edge cases at maximum capacity.
  * Prevented accidental overwrite of entries when removing a missing node.
  * Ensured satellite data tied to removed nodes is always cleaned up after compaction.

* **Tests**
  * Added new coverage for `removeNodeByNum()` when the database is full: handling both present and absent node IDs, verifying correct compaction and tail cleanup.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -1603,8 +1603,10 @@ void NodeDB::removeNodeByNum(NodeNum nodeNum)
         const size_t first = numMeshNodes;
         const size_t last = std::min(first + removed, nodeDatabase.nodes.size());
         std::fill(nodeDatabase.nodes.begin() + first, nodeDatabase.nodes.begin() + last, meshtastic_NodeInfoLite());
-        eraseNodeSatellites(nodeNum);
     }
+    // Drop the node's satellite stores and warm-tier copy regardless of which tier it lived in, so
+    // an explicit removal fully forgets it.
+    eraseNodeSatellites(nodeNum);
 #if WARM_NODE_COUNT > 0
     // Explicit user removal: don't let the warm tier resurrect the node
     warmStore.remove(nodeNum);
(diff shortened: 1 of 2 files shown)


caveman99:
Addressed in 721afddb6. `eraseNodeSatellites(nodeNum)` now runs unconditionally, matching this function's documented contract (it "drops the lite entry, its satellite stores, and the warm-tier copy"); only the slot-clear (the actual out-of-bounds fix) stays guarded by `removed`.

In practice a warm-only node has no satellites - every demotion path already erases them - so this is a no-op on today's code paths, but it drops the reliance on that invariant so an explicit removal fully forgets the node regardless of which tier it lived in.

PR #11024, opened 2026-07-16
AdminModule: only accept admin responses to requests we sent

caveman99:
Admin `*_response` messages skip both the sender-authorization branch and the
session-passkey check that gate every other admin message:

```c
if (messageIsResponse(r)) {
    LOG_DEBUG("Allow admin response message");   // no auth, no session key
}
...
if (mp.from != 0 && !messageIsRequest(r) && !messageIsResponse(r)) { checkPassKey... }
```

The module is bound to no channel, so on any channel it can decode, a node can deliver
a `get_module_config_response` it never asked for. `handleGetModuleConfigResponse` then
writes the responder's node_num and pin list into `devicestate.node_remote_hardware_pins`.

Accept a response only from a node we sent the matching request to. `MeshService::handleToRadio`
now records the destination of each outgoing admin request (per remote, with the pinned PKC key
when the client set one), and the response branch checks for a live matching request within the
same window as the session passkey. Local admin (`from == 0`) is unchanged - PhoneAPI already
gates that connection.

Second, smaller fix in the same path: the response dispatch compared
`get_module_config_response.which_payload_variant` (a ModuleConfig oneof tag, `remote_hardware` = 9)
against `meshtastic_AdminMessage_ModuleConfigType_REMOTEHARDWARE_CONFIG` (an AdminMessage enum, = 8).
They never matched, so the handler was unreachable. It's compared against the oneof tag now, the
same way the request-side switch already builds the response.

Tests (test_admin_session_repro): the pairing gate directly - unsolicited response rejected,
solicited accepted, a request to one node does not admit another, and a setter does not arm the
gate - plus two end-to-end cases through handleReceivedProtobuf asserting the pin table is left
alone for an unsolicited response and updated for a solicited one (the latter also covering the
dispatch fix).

Note: touches AdminModule alongside the in-flight signing series (#10967, #10969, #10972); no
overlap on the response path, but they will need a trivial rebase around the switch.

## Summary by CodeRabbit

* **Security Improvements**
  * Admin “responses” are now accepted only if they match a recently recorded, correctly typed outgoing request.
  * Unsolicited or mismatched responses (including from different senders or wrong payload types) are rejected.
  * When applicable, responses must satisfy cryptographic key pinning and timeout rules.
  * Unsolicited module configuration replies no longer affect remote hardware pin ownership.

* **Bug Fixes**
  * Improved detection/handling for remote hardware module configuration responses to ensure correct pairing behavior.

* **Tests**
  * Added deterministic regression tests covering solicitation tracking and end-to-end pin update behavior.

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -17,6 +17,7 @@
 #include "main.h"
 #include "mesh-pb-constants.h"
 #include "meshUtils.h"
+#include "modules/AdminModule.h"
 #include "modules/NodeInfoModule.h"
 #include "modules/PositionModule.h"
 #include "modules/RoutingModule.h"
@@ -259,6 +260,14 @@ void MeshService::handleToRadio(meshtastic_MeshPacket &p)
                   const StoredMessage &sm = messageStore.addFromPacket(p);
                   graphics::MessageRenderer::handleNewMessage(nullptr, sm, p); // notify UI
               })
+#if !MESHTASTIC_EXCLUDE_ADMIN
+    // Note admin requests on their way out: AdminModule only accepts a response from a remote we
+    // actually asked. Runs before encryption, while the payload is still readable.
+    if (adminModule && p.which_payload_variant == meshtastic_MeshPacket_decoded_tag &&
+        p.decoded.portnum == meshtastic_PortNum_ADMIN_APP)
+        adminModule->noteOutgoingAdminRequest(p);
+#endif
+
     // Send the packet into the mesh
     DEBUG_HEAP_BEFORE;
     auto a = packetPool.allocCopy(p);
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -10,6 +10,7 @@
 #include "input/InputBroker.h"
 #include "meshUtils.h"
 #include <FSCommon.h>
+#include <Throttle.h>
 #include <ctype.h> // for better whitespace handling
 #if defined(ARCH_ESP32) && !MESHTASTIC_EXCLUDE_WIFI
 #include "MeshtasticOTA.h"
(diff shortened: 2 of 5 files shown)


caveman99:
Both addressed in 21a8f9697.

**millis() rollover**: the expiry now uses `Throttle::isWithinTimespanMs(sentAtMs, 300000)`, which does the rollover-safe `(now - sentAtMs) < window` unsigned subtraction. The record stores raw `millis()` rather than `millis()/1000` sums, so a request spanning the ~49.7-day wrap no longer reads as permanently live.

**Response-type binding**: each outstanding record now carries a bitmask of the response variants its requests authorize (`adminResponseForRequest` maps request tag to response tag, `adminResponseBit` maps that to a bit). A response is accepted only if its variant's bit is set, so a `get_owner` request no longer admits a `get_module_config` response from the same node. I used an accumulating per-remote mask rather than a single expected tag specifically to keep pipelined flows working: a client that fires several getter types at one node within the window still has every matching answer accepted. Added a test asserting the wrong variant is denied while the requested one from the same slot is still accepted.

jp-bennett:
Definitely approve of the concept

thebentern:

thebentern:

caveman99:
I have no idea what happened here but i cleaned it up :-)

caveman99:
Both new findings addressed in 8b73e68a6 (and the per-variant/rollover ones stay fixed by the per-request model in d8e5dd4).

**Module-config subtype (AdminModule.cpp:1904):** `remote_hardware` is the only `get_module_config_response` that mutates state (it's the only case in the switch, and it writes the pin table). Each outstanding entry now records the requested `ModuleConfigType`, and the gate requires a `remote_hardware` response to answer a `REMOTEHARDWARE_CONFIG` request specifically - a request for any other subtype (e.g. MQTT) no longer authorizes a pin-table update. I scoped the check to that one sink rather than duplicate the 17-case ConfigType→tag mapping, since the other subtypes have no state-mutating handler.

**Consume on accept (AdminModule.cpp:1925):** `responseIsSolicited` now clears the matched entry before returning true, so a node can't replay a state-mutating response across the 5-minute window. Since one request yields one response, I dropped the earlier request de-dup (a client's N indexed `get_channel` requests are now N entries, each consumed by its own response - de-dup + consume would have dropped the 2nd..Nth channel).

Tests: a non-remote-hardware request does not admit a `remote_hardware` response; a second copy of an answered response is rejected.

PR #11025, opened 2026-07-16
Stop accelerometer thread when double-tap/wake-on-motion disabled at runtime

thebentern:
## Problem

`double_tap_as_button_press` (device config) and `wake_on_tap_or_motion` (display config) are applied **live**, without a reboot — the `requiresReboot` equality gates in `handleSetConfig` deliberately exclude these two fields. On the OFF→ON edge, each handler calls `accelerometerThread->start()` and sets `enabled = true`.

But there was **no ON→OFF branch anywhere** (`grep 'accelerometerThread->'` shows only `calibrate`, `enabled = true`, and `start()`). So:

- Turning either flag off at runtime left the accelerometer thread running — polling I2C and drawing power — until the next reboot.
- Because `enabled` stayed stuck `true`, a later OFF→ON edge became a no-op (the `enabled == false` guard blocked the re-`start()`), leaving the feature **un-restartable** without a reboot.

## Fix

Add the symmetric ON→OFF branch to both handlers. When a flag transitions true→false **and** the other consumer of the shared thread is also off, call `accelerometerThread->disable()`:

- `OSThread::disable()` sets `enabled = false` and `setInterval(INT32_MAX)`, halting `runOnce()` (no more I2C polling) without deleting the sensor. Since `init()` is idempotent, a later `start()` cleanly resumes — no leak, fully restartable.
- Each branch checks the *other* flag first, so disabling one feature never stops the sensor while the other still needs it.

Chose the live `disable()` (symmetric with the existing live `start()`) over adding these fields to the `requiresReboot` gate, since the flags are intentionally applied live — a reboot gate would clash with the existing live-start.

Transition matrix (D = double-tap, W = wake-on-motion; only one config variant changes per `setConfig`):

| Transition | Other flag | After fix |
|---|---|---|
| D off→on | W off | `start()` |
| D on→off | W off | `disable()` (off, enabled cleared) |
| D on→off | W on | skip — W still needs it |
| W off→on | D off | `start()` |
| W on→off | D off | `disable()` |
| W on→off | D on | skip — D still needs it |

This is a cooperative `OSThread` scheduled on `mainController` in the same loop context as `setConfig`, so `disable()` is safe with no concurrency concern.

## Testing

- Compile-verified for the `rak4631` (nRF52840) target, where the accelerometer block is active (the native/portduino test build `#if`-excludes this region). `AdminModule.cpp` compiled cleanly.
- Low severity: power/behavior, not a crash. No hardware required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved power management by coordinating accelerometer start/stop across double-tap and wake-on-tap/motion configuration changes.
  * Prevented compass/heading freezing by detecting when the active sensor can directly provide heading and keeping motion updates running appropriately.
  * Reduced update inconsistencies by handling shared accelerometer-thread ownership more safely when motion processing availability changes.


jp-bennett:
Does this break the compass, particularly on devices that are 6+3 axis?

caveman99:
> Does this break the compass, particularly on devices that are 6+3 axis?

Not any more

PR #11033, opened 2026-07-16
PhoneAPI: gate local admin on the connection, not the wire from

caveman99:
Build config affected: nRF52 with `MESHTASTIC_ENABLE_LOCKDOWN=1` (which co-defines
`MESHTASTIC_PHONEAPI_ACCESS_CONTROL` and `MESHTASTIC_ENCRYPTED_STORAGE`).

The per-connection admin check in `handleToRadioPacket` was guarded by `if (p.from == 0 && ...)`:

```c
if (p.from == 0 && p.decoded.portnum == meshtastic_PortNum_ADMIN_APP) {
    ...
    if (!getAdminAuthorized()) return false;   // only reached when from == 0
}
```

`p.from` is a client-supplied wire field. The outer lockdown gate lets a decoded `ADMIN_APP`
packet addressed to us through without checking `from`, and `MeshService::handleToRadio`
rewrites `p.from = 0` before `AdminModule` runs. So a client that sets `from != 0` skips the
`!getAdminAuthorized()` drop entirely, the packet is normalized back to a local identity, and
`AdminModule` treats it as authorized local admin. An unauthorized local connection can then
run arbitrary admin once the device has been unlocked.

Every packet reaching `handleToRadioPacket` already originates from the local connection, so
whether it is "local admin" is a property of the connection, not of the wire `from`. The
decision now lives in `classifyLocalAdminPacket()`, which ignores `from` and keys only on the
admin variant and the connection's authorization: `lockdown_auth` is delivered inline (it is
itself the authentication), any other admin from an unauthorized connection is dropped, and an
authorized connection passes through. Authorized flows are unchanged.

The classifier is a static method compiled unconditionally, so it is unit-testable; the guarded
caller (built only in the lockdown config) calls it. No CI target builds
`MESHTASTIC_ENABLE_LOCKDOWN=1`, so the guarded caller is verified by inspection while the
security-critical decision is covered by the test.

Test (test_stream_api): an unauthorized connection's `ADMIN_APP` packet with `from != 0` is
classified `DropUnauthorized`; an authorized connection's identical packet passes through;
`lockdown_auth` is delivered inline regardless of `from`; a non-admin packet is outside the gate.

## Summary by CodeRabbit

* **Security Improvements**
  * Strengthened phone-to-radio administration gating by determining authorization from the active connection state rather than packet sender details.
  * Spoofed or unauthorized administrative payloads are dropped with warnings.
  * Lockdown authentication is still handled inline, and sensitive admin payload contents are cleared after processing.
* **Bug Fixes**
  * Improved handling of undecodable administrative payloads by routing them through the normal reject path.
* **Tests**
  * Added coverage for lockdown gate behavior, including sender-spoofing and undecodable admin packet scenarios.


caveman99:
All three addressed in 6bcc69945.

**Encoded passphrase wipe (Major)**: correct - `handleLockdownAuthInline` takes `admin.lockdown_auth` by reference and only zeroes that decoded copy; the still-encoded passphrase in `p.decoded.payload.bytes` was left in the packet buffer, and the comment I carried forward wrongly claimed otherwise. Now wiping both the decoded scratch and the encoded packet bytes, and fixed the comment.

**Two comment-length nitpicks**: shortened the `classifyLocalAdminPacket` docstring and the gate comment to two lines each, keeping the non-obvious rationale (ignore the wire `from`).

Note this whole block is in the `MESHTASTIC_ENCRYPTED_STORAGE && MESHTASTIC_PHONEAPI_ACCESS_CONTROL` guard, which no CI target compiles, so these changes are verified by inspection; the from-independent decision they wrap remains covered by the `classifyLocalAdminPacket` unit test.

NomDeTom (commented):
I've only reviewed on my phone but I ran it through clod and it made some sensible suggestions for once:

## Minor suggestions / nits
* Redundant decoded-scratch wipe. In the LockdownAuth case, handleLockdownAuthInline() already zeroes admin.lockdown_auth.passphrase.bytes[0..size] before it returns, then the caller zeroes the same buffer again (full capacity). Harmless and arguably defensive, but a one-line comment noting the caller wipe is the authoritative/full one would avoid a future reader "simplifying" it away. The encoded wipe is the one that must stay.
* Vestigial anonymous scope. With the if guard removed, the bare { … } block now exists only to scope admin. Fine, but slightly odd-looking; a short comment or dropping the braces would read cleaner.
* Stale p.decoded.payload.size. After wiping the encoded bytes you don't reset payload.size to 0. Irrelevant here (the packet returns true/false and is never forwarded), so not a bug — just noting it's not defense-in-depth.
* Test has no case for an undecodable/garbage ADMIN_APP payload → NotAdmin (the pb_decode failure branch). Cheap to add and would lock in the "fall through to the normal reject path" contract.
* The passphrase-wiping behavior in handleToRadioPacket isn't exercised (hard to unit-test through the transport stack) — acceptable, but it means the security-critical encodedVol wipe rides on code review alone.

caveman99:
Thanks, addressed in 25534efa2:

1. **Redundant decoded-scratch wipe** - comment now states the encoded-payload wipe is the security-critical one (nothing else clears the passphrase from the packet buffer) and the decoded-scratch re-zero is defensive full-capacity coverage over `handleLockdownAuthInline`'s size-bounded clear. Kept the encoded wipe.
2. **Vestigial anonymous scope** - kept the block (it deliberately bounds the lifetime of the decoded `admin`, which holds the plaintext passphrase) and added a one-line comment saying so.
3. **Stale `payload.size`** - now reset to 0 after the buffer wipe so the length stays consistent with the cleared bytes.
4. **Test gap** - added `test_lockdown_admin_gate_rejects_undecodable_admin`: a garbage ADMIN_APP payload must classify as `NotAdmin`. The control asserts the same for `adminAuthorized=true`, proving the decode-failure branch runs before the auth check and locking in the fall-through-to-reject contract.
5. **`handleToRadioPacket` wipe not exercised** - agreed, it's not reachable through the native transport stack, so that path stays on review coverage.

PR #11036, opened 2026-07-16
AudioModule: validate codec2 header and bound the RX decode reads

caveman99:
Two issues in the codec2 RX path, both reachable from a crafted AUDIO_APP payload (module is
ESP32 + 2.4 GHz / wide-LoRa only, so this affects those builds).

**Unvalidated mode byte -> NULL dereference.** When a received frame's header did not match
ours, the code built a temporary decoder from the frame's own mode byte:

```c
CODEC2 *tmp_codec2 = codec2_create(audioModule->rx_encode_frame[3]);
codec2_set_lpc_post_filter(tmp_codec2, ...);   // tmp_codec2 == NULL for an invalid mode
```

`codec2_create` returns NULL for an out-of-range mode, and the following calls dereference it.
Now only frames that carry our own header (magic + mode) are decoded; anything else is dropped,
so an untrusted mode byte never reaches `codec2_create`.

**Out-of-bounds read on a misaligned length.** The decode loop stepped by the frame size but only
checked `i < rx_encode_frame_index`, so a payload whose length was not a clean multiple of the
frame size made the final `codec2_decode` read past the received data and potentially past the
233-byte `rx_encode_frame`. Each read is now bounded by `i + frameSize <= rx_encode_frame_index`
clamped to the buffer size, and the receive-side `memcpy` is clamped to the buffer as well.

Behavior change: a frame whose codec2 mode differs from this node's configured mode is dropped
rather than decoded with a temporary codec. Interoperable audio already requires both ends to run
the same codec2 bitrate, so cross-mode frames were not usefully decodable.

No test: the module is ESP32/SX1280-only and has no native build or existing test coverage, so
these changes are verified by inspection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of incoming audio frames in RX mode before decoding, rejecting malformed or incompatible data.
  * Hardened TX-side audio capture to use byte-accurate buffering, improving reliability of captured frames.
  * Added strict bounds checking when copying received payloads to prevent overflows from crafted packets.
  * Removed fallback decoding on header mismatch to avoid unintended playback from invalid frames.

### DIFF
--- a/src/modules/esp32/AudioModule.cpp
+++ b/src/modules/esp32/AudioModule.cpp
@@ -68,23 +68,22 @@ void run_codec2(void *parameter)
             }
             if (audioModule->radio_state == RadioState::rx) {
                 size_t bytesOut = 0;
-                if (memcmp(audioModule->rx_encode_frame, &audioModule->tx_header, sizeof(audioModule->tx_header)) == 0) {
-                    for (int i = 4; i < audioModule->rx_encode_frame_index; i += audioModule->encode_codec_size) {
+                // Reject frames not carrying our own header (magic + mode); an invalid mode byte
+                // would else NULL-deref via codec2_create. The bound keeps reads inside the buffer.
+                const int headerLen = sizeof(audioModule->tx_header);
+                const int frameSize = audioModule->encode_codec_size;
+                int limit = audioModule->rx_encode_frame_index;
+                if (limit > (int)sizeof(audioModule->rx_encode_frame))
+                    limit = sizeof(audioModule->rx_encode_frame);
+                if (frameSize > 0 && limit >= headerLen &&
+                    memcmp(audioModule->rx_encode_frame, &audioModule->tx_header, headerLen) == 0) {
+                    for (int i = headerLen; i + frameSize <= limit; i += frameSize) {
                         codec2_decode(audioModule->codec2, audioModule->output_buffer, audioModule->rx_encode_frame + i);
                         i2s_write(I2S_PORT, &audioModule->output_buffer, audioModule->adc_buffer_size, &bytesOut,
                                   pdMS_TO_TICKS(500));
                     }
                 } else {
-                    // if the buffer header does not match our own codec, make a temp decoding setup.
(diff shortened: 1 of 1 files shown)


caveman99:
Addressed in acc421072. Confirmed and correct: `adc_buffer_size` is a sample count (`codec2_samples_per_frame`), and the TX path a few lines down already uses `2 * adc_buffer_size` for its byte-sized memcpy, so `i2s_write` was being handed half the byte count. Now passing `adc_buffer_size * sizeof(int16_t)`.

Two caveats: this is a pre-existing bug (the original `i2s_write` had the same sample count) that only shows up in this diff because the loop restructure moved the line, and I can't validate it on hardware - the module is ESP32/SX1280-only with no native build - so this is verified from the in-file sample/byte convention rather than by hearing corrected audio. Flagging in case you want to confirm on a real 2.4 GHz node.

caveman99:
Follow-up: the sample/byte fix wasn't complete with just the write, so 6d760be5d fixes the mirror on the capture path.

There were three i2s spots in play:
- **playback `i2s_write` (matched-header path)** - CR's finding, fixed in acc421072.
- **temp-codec `i2s_write`** - removed entirely with the NULL-deref branch in the security fix, so nothing to fix there.
- **capture `i2s_read`** - the mirror bug, and worse: `adc_buffer` is `uint16_t[]` and `adc_buffer_size` a sample count, but `i2s_read`'s size arg is bytes, `adc_buffer_index` accumulated bytes yet indexed a `uint16_t*` (advancing 2x), and the completion check compared bytes against samples. Net effect: it captured ~half a frame and copied the rest stale. Now byte-consistent throughout: `frameBytes = adc_buffer_size * sizeof(uint16_t)`, a `uint8_t*` cursor, remaining-bytes read size, and completion/`memcpy` in bytes.

Same caveat as before: ESP32/SX1280-only, no native build, so both directions are verified from the in-module sample/byte convention rather than on hardware. This one reworks the capture state machine, so it's worth a real 2.4 GHz round-trip check before relying on it.

thebentern (approved):

PR #11040, opened 2026-07-17
NRF52 - power management cleanups (NRF_APM)

phaseloop:
Small cleanup of redundant power management code:

- NRF_APM (VBUS detection) is present in all NRF52 chips and there is no point in defining it only in certain variants. It is now set globally in NRF52 architecture.

- USB power detection in NRF52 is already covered in PowerHAL implementation and is used by analogBatteryLevel class, so USB power and battery charging is now detected properly without any need for NRF52 specific code. That redundant part was removed.

- PowerHAL VBUS detection (powerHAL_isVBUSConnected()) in analogBatteryLevel class is now used for all devices. If specific platform does not have it implemented - it's just ignored.

- Enabling INA sensor will use it's readings and ignore NRF_APM VBUS detection as intended.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


basement-labs:
cc: @Jorropo - I think you added original code to fix some RAK4631 issues

ndoo:
This is a good idea, had to add it manually to Seeed Wio Tracker L1 but it worked out of the box after that, so I doubt there’s any NRF52 board where this wouldn't work.

Especially since Nordic docs already say VBUS needs to be connected to USB 5V to power the USB peripheral, not powered from elsewhere.

phaseloop:
I was busy with other things but I'll solve the merge conflict and make it ready to merge

PR #11048, opened 2026-07-17
TrafficManagement: gate role/NodeInfo cache writes on signer authenticity

caveman99:
The tier-3 role cache (`updateCachedRoleFromNodeInfo`) and the PSRAM NodeInfo response cache (`cacheNodeInfoPacket`) were refreshed from any received NodeInfo without an authenticity check. A spoofed NodeInfo (the sender field is forgeable) could therefore:

- set a tracked node's cached role, granting the tracker / lost-and-found position-dedup exceptions and pinning the entry against eviction, or
- overwrite the cached user that direct NodeInfo responses serve to requestors.

Both cache writes now use the same gate the identity-update path already applies: when a node known to sign its NodeInfo (`nodeInfoLiteHasXeddsaSigned`) sends an unsigned NodeInfo, the update is skipped. Nodes that have never signed are unaffected. No overlap with the existing direct-response `updateUser` gate; that guards a separate sink.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unauthenticated (unsigned) NodeInfo packets from updating trusted cached identity details.
  * Ensured NodeInfo processing consistently applies enhanced authentication gating, avoiding unintended updates to cached role data and related refresh behavior.
  * Updated the direct-response handling path to reuse the same authentication decision, preventing unsigned senders from triggering guarded identity overwrites.

### DIFF
--- a/src/modules/TrafficManagementModule.cpp
+++ b/src/modules/TrafficManagementModule.cpp
@@ -729,8 +729,15 @@ ProcessMessage TrafficManagementModule::handleReceived(const meshtastic_MeshPack
     // Learn NodeInfo payloads into the dedicated PSRAM cache, and refresh the tier-3
     // role cache for any node we already track (keeps the dedup role exception current).
     if (mp.decoded.portnum == meshtastic_PortNum_NODEINFO_APP) {
-        cacheNodeInfoPacket(mp);
-        updateCachedRoleFromNodeInfo(mp);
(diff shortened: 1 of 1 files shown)


NomDeTom (approved):
It will do the job. There's a couple of wider testing and overall performance points which I'll pick up in a separate PR that builds on this work.

PR #11061, opened 2026-07-18
Backlight: unify E-Ink backlight logic and fix ThinkNode M1 wake

### LINKED ISSUE #9768: [Bug]: Elecrow ThinkNode M1 - Rotary knob does not wake screen in v2.7.15
### Category

Hardware Compatibility

### Hardware

Other

- [x] InkHUD ePaper

### Firmware Version

2.7.15

### Description

Device: Elecrow ThinkNode M1 (nRF52840)
Prior to 2.7.15, turning the knob would wake the screen, from no backlight to full backlight. After flashing to 2.7.15, after the preset screen timeout (10 min), screen will no longer wake up. Text is visible, but no backlight. This issue happens with both the base UI or InkHUD ePaper. 
My temporary workaround is to set timeout to "Never", then manually turn brightness to zero.

### Relevant log output

```Shell

```

brianTrz:
fixes #9768

The E-Ink backlight control relied on a sentinel check `brightness == 1` to toggle the display's power rail. On devices like the ThinkNode M1 (which boots with a default brightness of 153 and uses a physical dial for dimming), this caused the backlight power rail to stay disabled when waking up from screen timeout.

Unify E-Ink backlight checks globally in Screen, MenuHandler, and InputBroker to check for `brightness > 0` instead of the sentinel. Backlight toggling now alternates between `0` (OFF) and `255` (ON). This resolves the ThinkNode M1 wake bug and allows the backlight to turn completely off by writing the pin fully LOW when set to 0.

To preserve the default-off battery-saving behavior of low-power E-Ink devices, initialize TTGO T-Echo family targets (T-Echo, T-Echo Lite, T-Echo Card, and T-Echo Plus) with a default screen brightness of `0` in NodeDB. All other targets retain the default of `153`.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Testing Note:
This change has been built and tested on an Elecrow ThinkNode M1. It has not yet been physically verified on other devices.

## Summary

* **Bug Fixes**
  * Updated screen backlight/E-ink enable logic so any `screen_brightness` value above 0 is treated as “on”, and turning on/off respects that threshold.
  * Fixed touch-button backlight toggling to use the current hardware pin state at press time and avoid disabling when the screen is on with brightness > 0.
  * Updated “Toggle Backlight” to use full brightness (255) when turning on, and disable when set to 0.

* **Chores**
  * Adjusted default screen brightness for specific display hardware variants (0 for echo variants; otherwise 153).


HarukiToreda (approved):
Looks good, I was able to wake the backlight after it timed out on the thinknode M1 as intended. Good to merge

PR #11062, opened 2026-07-18
fix(t5s3-epaper): 40MHz octal PSRAM; guard EInk init; esp32s3 sw-atomic shim

thebentern:
Gets the LilyGo T5-ePaper-S3 Pro building and running.

- **t5s3-epaper**: the board's 3.3V octal (AP_3v3) PSRAM is unreliable at the `qio_opi` default of 80MHz — `Total PSRAM` reads 0, FastEPD's PSRAM-only framebuffer alloc then fails and the node boot-loops. Clock octal PSRAM at 40MHz. No precompiled 40MHz-octal libs exist, so this forces a from-source build. Verified on hardware: Free PSRAM ~7.4MB, e-paper renders.
- **EInkParallelDisplay**: capture `initPanel()`'s rc and no-op display ops when the FastEPD framebuffer is NULL (PSRAM absent / init failed), so the node boots headless instead of hard-faulting (`StoreProhibited` in ROM `memset`).
- **esp32s3 sw-atomic shim**: weak software `__atomic_*_{1,2,4}`. `-mdisable-hardware-atomics` makes GCC emit these libcalls, but precompiled `libnewlib` ships only the `_8` variants and some toolchains' `libgcc` ship none, so esp32s3 fails to link locally on macOS. Marked weak, so a real `libgcc` definition wins where present (no-op on Linux/CI).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved e-paper display initialization by detecting missing framebuffer/failed panel setup and preventing unsafe refresh operations.
  * Display update actions are now skipped until the display is confirmed ready, reducing startup and refresh issues.
* **Compatibility**
  * Improved ESP32-S2/ESP32-S3 builds with software-based atomic operation support when required.
  * Enhanced reliability on supported ESP32-S3 e-paper hardware by overriding PSRAM speed settings for more consistent operation.

### DIFF
--- a/src/graphics/EInkParallelDisplay.cpp
+++ b/src/graphics/EInkParallelDisplay.cpp
@@ -77,12 +77,13 @@ EInkParallelDisplay::~EInkParallelDisplay()
 bool EInkParallelDisplay::connect()
 {
     LOG_INFO("Do EPD init");
+    int initRc = BBEP_SUCCESS;
     if (!epaper) {
         epaper = new FASTEPD;
 #if defined(T5_S3_EPAPER_PRO_V1)
-        epaper->initPanel(BB_PANEL_LILYGO_T5PRO, 28000000);
+        initRc = epaper->initPanel(BB_PANEL_LILYGO_T5PRO, 28000000);
 #elif defined(T5_S3_EPAPER_PRO_V2)
-        epaper->initPanel(BB_PANEL_LILYGO_T5PRO_V2, 28000000);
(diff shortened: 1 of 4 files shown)


thebentern:
Self-review (code-review, high): hardened the PSRAM-absent path — `connect()` now returns true so `OLEDDisplay` still allocates its base buffer (Screen ignores `init()`'s failure), avoiding a NULL base-buffer deref; `displayReady` keeps the FastEPD push paths no-op. Verified builds from-source at 40MHz.

PR #8248, opened 2025-10-07
Meshtastic build manifest

vidplace7:
Adds a simple json manifest to the firmware build process via `platformio-custom.py`.

The intent is to use this manifest to simplify flasher / install script board support, and to better support DIY boards (uploading zip to the flasher manually).

This will later be extended to simplify the `bin/install-*` scripts (they are a MESS).

---

Ex `t-deck`:
```json
{
    "version": "2.7.12.e68ab48e4",
    "build_epoch": 1759809600,
    "board": "t-deck",
    "mcu": "esp32s3",
    "repo": "vidplace7/meshtastic-firmware",
    "part_csv": "default_16MB.csv",
    "has_mui": false
}
```

Ex `t-deck-tft`:
```json
{
    "version": "2.7.12.e68ab48e4",
    "build_epoch": 1759809600,
    "board": "t-deck-tft",
    "mcu": "esp32s3",
    "repo": "vidplace7/meshtastic-firmware",
    "part_csv": "default_16MB.csv",
    "has_mui": true
}
```

Ex `seeed-sensecap-indicator-tft`:
```json
{
    "version": "2.7.12.e68ab48e4",
    "build_epoch": 1759809600,
    "board": "seeed-sensecap-indicator-tft",
    "mcu": "esp32s3",
    "repo": "vidplace7/meshtastic-firmware",
    "part_csv": "partition-table-8MB.csv",
    "has_mui": true
}
```

Ex `native`:
```json
{
    "version": "2.7.12.e68ab48e4",
    "build_epoch": 1759809600,
    "board": "native",
    "mcu": null,
    "repo": "vidplace7/meshtastic-firmware",
    "part_csv": null,
    "has_mui": false
}
```

Ex `rak4631`:
```json
{
    "version": "2.7.12.e68ab48e4",
    "build_epoch": 1759809600,
    "board": "rak4631",
    "mcu": "nrf52840",
    "repo": "vidplace7/meshtastic-firmware",
    "part_csv": null,
    "has_mui": false
}
```

Ex `rak3172`:
```json
{
    "version": "2.7.12.e68ab48e4",
    "build_epoch": 1759809600,
    "board": "rak3172",
    "mcu": "stm32wle5ccu",
    "repo": "vidplace7/meshtastic-firmware",
    "part_csv": null,
    "has_mui": false
}
```

PR #11078, opened 2026-07-19
Fix inverted .factory.bin filename check in device-install.bat

Cryptoteep:
## Summary

The filename validation added in #8248 is inverted: it errors out precisely when the filename **does** contain `.factory.bin`:

```bat
IF NOT "__!FILENAME:.factory.bin=!__"=="__!FILENAME!__" (
    CALL :LOG_MESSAGE ERROR "Filename must be a firmware-*.factory.bin file."
```

The substring replacement changes the string exactly when `.factory.bin` is present, so every valid `firmware-*.factory.bin` file is rejected with "Filename must be a firmware-*.factory.bin file.", making the `-f` flow of the batch installer unusable since that refactor — while names *without* `.factory.bin` pass the check.

This drops the stray `NOT`, matching the intent and the equivalent check in `device-install.sh`.

## Testing

On Windows 11 with a dummy `firmware-test-1.0.factory.bin` + `.mt.json` metadata:

- before: `ERROR ... Filename must be a firmware-*.factory.bin file.` and exit
- after: metadata is parsed and the script proceeds through the flash steps (`--debug` mode); a filename without `.factory.bin` is now the one that gets rejected

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed firmware filename validation during device installation.
  * The installer now correctly accepts only firmware files matching the `firmware-*.factory.bin` pattern.
  * Incorrectly named firmware files now reliably trigger an error instead of being misclassified.

### DIFF
--- a/bin/device-install.bat
+++ b/bin/device-install.bat
@@ -70,7 +70,7 @@ IF "__!FILENAME!__"=="____" (
         CALL :LOG_MESSAGE ERROR "Filename containing spaces are not supported."
         GOTO help
     )
-    IF NOT "__!FILENAME:.factory.bin=!__"=="__!FILENAME!__" (
+    IF "__!FILENAME:.factory.bin=!__"=="__!FILENAME!__" (
         CALL :LOG_MESSAGE ERROR "Filename must be a firmware-*.factory.bin file."
         GOTO help
     )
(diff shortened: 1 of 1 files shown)


Cryptoteep:
Good catch — fixed in 0bfca08: the check now validates the basename starts with `firmware-` and ends exactly with `.factory.bin`, rejecting names like `foo.factory.bin` or `firmware-x.factory.bin.bak`, matching device-install.sh.

NomDeTom:
@markbirss can you have a look?

caveman99:
CLA is not signed. Waiting 7 days

PR #11082, opened 2026-07-20
Remove dead code

caveman99:
Removes function definitions that have no call sites.

## Method

Definitions were extracted mechanically and every candidate was then verified individually against `src`, `test`, `variants`, `lib` and the external device-ui checkout, checking for virtual dispatch, function-pointer and callback registration, and platform ifdef guards. 158 candidates were reviewed; false positives were kept.

Deliberately kept:
- nanopb field callbacks and HTTP route handlers reached by function pointer
- virtual overrides invoked through a base pointer
- vendored code under `src/platform` and LovyanGFX
- driver API surface with no current caller (TCA8418KeyboardBase, INA3221Sensor, SGM41562, BBQ10Keyboard, RadioLibInterface noise floor accessors)
- `AdminModule::handleGetNodeRemoteHardwarePins`, `Channels::setDefaultPresetCryptoForHash` and `MeshService::getNodenumFromRequestId`, which are unused but would be protocol or feature changes rather than cleanup

## Notes

The OnScreenKeyboardModule popup chain is removed as a unit. `showPopup` was the only writer of `popupVisible`, and its only caller was already unreferenced, so `drawPopup`, `drawPopupOverlay` and `clearPopup` were reachable but permanently inert. This is fallout from the frame based UI being replaced by baseUI.

`htmlDeleteDir` became unreachable once `handleDeleteFsContent` was removed; it was only ever reached through that handler plus its own recursion.

`ThreadedMessageApplet::saveMessagesToFlash` is a vestigial wrapper, not a missing save. Persistence moved to `Events::beforeDeepSleep` and `Events::beforeReboot`.

## Verification

Local run used the `native-windows` env, which is headless (`HAS_SCREEN=0`), so the graphics and InkHUD changes are not covered locally and rely on CI. No deleted symbol appears in the local build log.

## Summary by CodeRabbit

* **New Features**
  * Long-press audio feedback now plays progressively, with support for restarting the lead-up sequence.
  * Canned messages can be launched with a selected destination and channel.

* **Changes**
  * Keyboard popup overlays and related transient notifications are no longer shown.
  * Web admin, hotspot, and filesystem management endpoints have been reduced.
  * Canned messages no longer support “repeat last destination” behavior.
  * Telemetry speed display now uses knots (instead of miles-per-hour).
  * Message persistence is handled centrally during shutdown and sleep flows.

* **Documentation**
  * Updated guidance on message ingestion and filtering behavior.


caveman99:
Addressed the review comment: dropped the `or manual input` wording, since `addFromString` was removed. Also noted the null-return contract, which is the part callers actually need.

Two points in the walkthrough do not match the diff. This PR adds no behavior; the only five added lines are the `tryAddFromPacket` declaration line and the InkHUD README correction. Specifically, it does not add "stateful one-note buzzer progression" (the `leadUpNotes` table and `playNextLeadUpNote` already existed and are untouched; only the unused `playLongPressLeadUp` was removed) and it does not replace the session remaining-time query with an expiration status (only the unused `getSessionRemainingSeconds` was removed). Both look like pure deletions being read as additions.

PR #11086, opened 2026-07-20
Null-check packet allocations in allocForSending and its callers

caveman99:
`Router::allocForSending()` dereferenced `packetPool.allocZeroed()` without a null check. Sibling of 0ae44d701.

The pool is heap-backed on ARCH_STM32WL and BOARD_HAS_PSRAM, so allocation can fail. Guarding `allocForSending()` alone moves the deref to its callers, so the `allocDataPacket()` wrappers and all unguarded call sites are guarded too. 31 call sites: 29 guarded, 2 return the pointer directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability by safely handling packet/protobuf allocation failures during outgoing mesh communication.
  * Prevented potential crashes and invalid sends across routing ACK/Nak, MQTT local ACK generation, and multiple telemetry and module message paths (including sensors, status, position/PLI, audio, games, neighbor info, admin replies, key verification, store-and-forward, and paxcounter).
  * Messages/telemetry now fail gracefully with early exit/backoff instead of dereferencing null allocations.

### DIFF
--- a/src/graphics/niche/InkHUD/Applets/System/Menu/MenuApplet.cpp
+++ b/src/graphics/niche/InkHUD/Applets/System/Menu/MenuApplet.cpp
@@ -2614,6 +2614,8 @@ uint16_t InkHUD::MenuApplet::getSystemInfoPanelHeight()
 void InkHUD::MenuApplet::sendText(NodeNum dest, ChannelIndex channel, const char *message)
 {
     meshtastic_MeshPacket *p = router->allocForSending();
+    if (!p)
+        return;
     p->decoded.portnum = meshtastic_PortNum_TEXT_MESSAGE_APP;
     p->to = dest;
     p->channel = channel;
--- a/src/mesh/MeshModule.cpp
+++ b/src/mesh/MeshModule.cpp
@@ -57,6 +57,8 @@ meshtastic_MeshPacket *MeshModule::allocAckNak(meshtastic_Routing_Error err, Nod
     // So we manually call pb_encode_to_bytes and specify routing port number
     // auto p = allocDataProtobuf(c);
     meshtastic_MeshPacket *p = router->allocForSending();
+    if (!p)
+        return nullptr;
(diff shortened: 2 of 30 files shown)


caveman99:
Both findings were valid and are fixed.

Rather than patch the two reported sites, I computed the transitive closure of functions that can now return `nullptr`. Twelve functions are nullable once the contract propagates: `allocForSending`, `allocDataPacket`, `allocReply`, `allocDataProtobuf`, `allocAckNak`, `allocErrorResponse`, `allocAtakPli`, `allocPositionPacket`, `preparePayload`, `gameAllocDataPacket`, `getForPhone`, `sendConditions`.

That surfaced 23 unguarded dereference sites at the second tier, including the two reported. All are now guarded: AdminModule x8, the seven Telemetry senders, PaxcounterModule, KeyVerificationModule x2, NeighborInfoModule, RoutingModule, StoreForwardModule, MQTT.

One reported site needed no change. At AdminModule.cpp:719 the only dereference is already guarded by `if (mp.pki_encrypted && myReply)`.

Totals across both commits: 55 call sites guarded, 0 unguarded dereferences remaining.

caveman99:
Valid, and fixed in 5076999.

The early `return false` skipped the pre-sleep block at the end of each sender, which the code explicitly documents must run even when nothing was sent, otherwise a power-saving SENSOR node stays awake until the next interval and drains its battery. Allocation failure now sets `validTelemetry = false` and skips only the packet setup and send, so control still reaches the sleep scheduling and the function still returns false. The existing LOG_WARN in that block already reads "unavailable this cycle, sleep without sending", which covers allocation failure correctly.

Scope checked: only these four senders have the sleep mechanism. DeviceTelemetry, HostMetrics and PaxcounterModule have no such block, so their early returns are left as they are.

`git diff -w` for the four files shows only the guard replacement and one closing brace, so the send path is unchanged apart from indentation. native-windows builds clean and all four objects compile.

PR #11097, opened 2026-07-20
Honour which_payload_variant on MQTT client-proxy ingress

caveman99:
`MQTT::onClientProxyReceive` read `msg.payload_variant.data.bytes/.size` without checking `which_payload_variant`.

`payload_variant` is a union of `data` (`PB_BYTES_ARRAY_T(435)`) and `text` (`char[435]`). On the text variant, `data.size` aliases the first bytes of the string, so a client sending text beginning `0xFF 0xFF` yields a length of 65535. That length is passed to `onReceive` and on to `DecodedServiceEnvelope`, reading up to ~64KB past a 501-byte by-value stack copy.

Reachable from any client that can send a ToRadio `mqttClientProxyMessage` (PhoneAPI.cpp:488) when MQTT and proxy_to_client are enabled.

Both variants are legitimate on this path: the firmware publishes text for JSON topics and data for protobuf, so the fix dispatches on the variant rather than dropping text. Text length comes from `strnlen` bounded by the buffer. A message with no variant set is ignored.

Tests added in `test_mqtt` for the text variant with a `0xFF`-filled payload, a text-variant envelope still being delivered, and an unset variant being ignored.

Note: `test_mqtt` does not build on `native-windows` (it includes `arpa/inet.h`), so these tests were not run locally. They are covered by the Linux native CI job.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MQTT client proxy payload handling by decoding according to the selected payload variant (binary vs text).
  * Prevented delivery when the proxy payload variant is missing or unrecognized.
  * Ensured valid text payloads continue to be processed and forwarded correctly.

* **Tests**
  * Added unit tests covering correct text-variant interpretation (not treated as raw bytes) and ignoring proxy messages with no selected payload variant.

### DIFF
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -335,7 +335,20 @@ void MQTT::mqttCallback(char *topic, byte *payload, unsigned int length)
 
 void MQTT::onClientProxyReceive(meshtastic_MqttClientProxyMessage msg)
 {
-    onReceive(msg.topic, msg.payload_variant.data.bytes, msg.payload_variant.data.size);
+    // payload_variant is a union: on the text variant, data.size aliases the first bytes of the
+    // string, so reading it unconditionally let a client name any length up to PB_SIZE_MAX.
+    switch (msg.which_payload_variant) {
+    case meshtastic_MqttClientProxyMessage_data_tag:
+        onReceive(msg.topic, msg.payload_variant.data.bytes, msg.payload_variant.data.size);
+        break;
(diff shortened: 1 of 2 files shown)


caveman99:
Note on test coverage: the text variant has no delivery control test. An encoded `ServiceEnvelope` always contains NUL bytes, so a text payload can never carry one — the text path exists for JSON topics, which do not decode as protobuf either way. An earlier version of this PR had such a test guarded by `TEST_IGNORE`, which meant it never ran; it has been removed rather than left as permanently-skipped coverage.

PR #11100, opened 2026-07-20
Budget the admin-key PKI decrypt fallback

caveman99:
`Router::perhapsDecode` tries each configured admin key when the sender's own key does not decrypt a PKI unicast. Each attempt is an X25519 scalar multiplication, and the AEAD tag that rejects a forged packet is only checked afterwards, so the expensive work happens before any authentication. An attacker needs a frame with `channel=0`, `to` set to the victim, 13+ bytes of random ciphertext and a fresh `id`; every such packet costs the full three operations because the loop only stops on success. Nothing on the path rate limited this.

`p->from` is unauthenticated header data, so a per-node throttle is evaded by varying it. The budget is therefore global: a token bucket of 8, refilling one per 250ms, spent only when the fallback actually runs.

Successful runs refund their token. A successful admin decrypt already persists the admin key into the sender's NodeDB entry, so subsequent packets from that admin take the existing single-operation fast path and never re-enter the loop. Authorized administration is low volume and unaffected.

Trying admin keys for senders that are not yet in NodeDB is deliberate, so that remote administration works against a node that has not learned the admin's key. That behavior is unchanged.

Adds a regression test covering both directions: the fallback stops running once the burst is drained, and it resumes after the bucket refills.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added global throttling for PKI administrative-key fallback decryption to reduce flood impact.
  * Failed fallback attempts are now rate-limited, and successful decryption continues to work normally once the correct key is used after the limit refills.
* **Tests**
  * Added a unit test validating rate-limited fallback behavior, including budget exhaustion, refill, and eventual success with the correct administrative key.


caveman99:
Condensed both comment sites. The rationale now lives in one line at the implementation and the test refers to it rather than repeating it. Tests still pass, 5 cases.

PR #11101, opened 2026-07-20
Bound remote-initiated key verification sessions

caveman99:
`KeyVerificationModule` holds a single verification slot. A remote packet opens it in `allocReply()`, which also raises a 30 second banner and a WARNING client notification. None of that required authentication: the peer supplies any 32 bytes in `hash1`, and `pki_encrypted` is not needed on the first message because bootstrapping a peer that does not yet hold our key is intentional.

Four problems, all remotely reachable:

- `updateState()` ran at the top of `handleReceivedProtobuf` before the `mp.from` check, so any node could refresh the deadline of someone else's session and keep the slot occupied indefinitely.
- `resetToIdle()` left no cooldown, so the slot could be reoccupied immediately and the banner raised again.
- The busy path returned `nullptr` without setting `ignoreRequest`, so `MeshModule` answered with a NAK.
- `currentNonceTimestamp < getTime() - 60` underflows while the clock is below 60, so the timeout behaved differently depending on the time source.

Changes:

- `sessionStartedMs` gives an absolute cap of 3 minutes that incoming packets cannot refresh. It is millis() based, so it still applies when the RTC is unset.
- The idle deadline is now extended only where the protocol actually advances, not on every arriving packet.
- `lastRemoteSessionMs` enforces 60 seconds between remote-initiated sessions, stamped when a session ends so the cooldown is a real gap. Locally initiated verification is not affected.
- Because the banner and notification are raised only when a session is opened, spacing sessions also bounds the notification spam. No separate notification throttle is needed.
- The busy and cooldown paths set `ignoreRequest`.
- The timeout comparison uses addition.

Builds clean on native-windows, which compiles this module.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened key verification session timing with explicit per-step deadlines and an absolute maximum session duration.
  * Prevented unauthenticated inbound traffic from extending session deadlines.
  * Added stronger rate limiting and a remote-initiated cooldown to reduce repeated attempts.
  * Rejected remote requests when a session is already active.
  * Improved session cleanup by recording remote session timing and resetting origin tracking after completion.

### DIFF
--- a/src/modules/KeyVerificationModule.cpp
+++ b/src/modules/KeyVerificationModule.cpp
@@ -71,8 +71,7 @@ AdminMessageHandleResult KeyVerificationModule::handleAdminMessageForModule(cons
 
 bool KeyVerificationModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshtastic_KeyVerification *r)
 {
-    // Do not refresh the deadline here: this runs before the sender is checked, so any node could keep
-    // the session alive indefinitely. The deadline is extended only when the protocol actually advances.
+    // No refresh here: this runs before the sender is checked, so any node could hold the session open.
     updateState(false);
     // Note: pki_encrypted is not required here. The first response (M2) may arrive channel-encrypted in
     // the bootstrap case; the follow-on hash1 packet (M3) is required to be PKI in its branch below.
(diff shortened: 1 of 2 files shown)


caveman99:
Both addressed.

Timeout comparison: adopted the elapsed-time form. It is strictly better than what I had. My addition version fixed the original underflow below 60 seconds but would wrap near uint32 max and expire a session early; `now - currentNonceTimestamp >= KEY_VERIFICATION_TIMEOUT_SECS` is correct in both directions.

Comment length: condensed.

PR #11102, opened 2026-07-20
Limit node database churn from unauthenticated packets

caveman99:
Two related problems on the unauthenticated NodeDB path.

**Ordering.** `updateUser()` called `getOrCreateMeshNode()` before the signature gate added in 5ded0ec8c. Creating an entry can evict another node and write the warm tier, so a refused identity update had already caused both side effects before returning false. The gate now runs against the existing record, and the entry is created only once it passes. Behaviour is otherwise identical: the gate only ever fired for nodes already present and already known to sign.

**Admission churn.** `updateFrom()` creates an entry for any decoded packet, and `mp.from` is unauthenticated header data. On a channel whose key is known, including the default public channel, invented node numbers churn the database at packet rate. Eviction prefers keyless nodes, so a flood pushes real keyless neighbours out, and each eviction also writes the warm tier.

Admission from the packet path is now spaced once the database is full. The limit is deliberately narrow:

- only the packet-driven path, so explicit admission and internal callers still evict freely
- only while the database is full, so discovery is untouched while slots remain
- existing nodes update normally, since the throttle only applies to creating an entry

An earlier revision put the throttle inside `getOrCreateMeshNode()`; `test_nodedb_blocked` correctly rejected it, because that also gated eviction paths that must succeed. Narrowing it to `updateFrom()` addresses the threat without changing those semantics.

The nRF52840 warm store was also examined. Its immediate `persistEntry` write is by design, buffered through the shared flash page cache with `saveIfDirty()` as the durability point, so it is left alone.

test_nodedb_blocked, test_packet_signing, test_warm_store and test_optin_migration pass, 75 cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary node database churn when updates arrive while storage is full.
  * Prevented rejected unsigned identity updates from creating or evicting node records after prior valid signing.
  * Added rate-limiting for admitting unknown nodes when the node database reaches capacity, improving stability and performance.
  * Improved handling of decoded node updates by only creating new records when needed, deferring fabricated/unknown node admissions when at capacity.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -3120,6 +3120,9 @@ size_t NodeDB::getNumOnlineMeshNodes(bool localOnly)
 #include "MeshModule.h"
 #include "Throttle.h"
 
+// Minimum spacing between evictions once the node database is full.
+#define NODEDB_FULL_EVICTION_INTERVAL_MS (2 * 1000UL)
+
 static constexpr uint32_t HOPSTART_DROP_LOG_INTERVAL_MS = 15000;
 
 void logHopStartDrop(const meshtastic_MeshPacket &p, const char *context)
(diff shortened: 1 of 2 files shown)


caveman99:
Both comment blocks condensed. No logic change; native-windows still builds clean.

PR #11104, opened 2026-07-20
Throttle NodeInfo direct responses

caveman99:
`nodeinfo_direct_response_max_hops` lets a bystander answer a NodeInfo request from cache. The reply is built with `reply->from = p->to` and `reply->to = getFrom(p)`, both taken from the incoming packet, and `p->from` is unauthenticated header data.

That makes it a reflector. One request with `from` set to a victim causes every neighbour holding the target in cache to transmit a NodeInfo at that victim, all at once, with no suppression between responders and no jitter. The trigger can be a bare header with an empty payload, and the reply carries a full `meshtastic_User`, so it amplifies by size as well as by responder count. The request is consumed with STOP, so the real target never sees it.

Nothing limited this. The module's existing `isRateLimited()` sits after the direct-response block returns, so it never ran on this path, and it keys on the spoofable `from` anyway.

Replies are now spaced per requester, which is what bounds the harm: an attacker can vary `from` freely, but the responses then scatter across different addresses rather than concentrating on one victim. A global floor additionally bounds the airtime this feature can consume regardless of how the requests are spread.

The check sits immediately before a reply is sent, so requests declined for other reasons, such as the target not being cached, do not consume the budget.

This path is opt-in and defaults to off, so this affects only meshes that enabled it.

test_traffic_management passes, 47 cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added throttling for direct NodeInfo responses to prevent too-frequent replies.
  * Enforced limits on reply frequency both globally and per requesting device.
  * Avoided unnecessary reply-budget consumption and packet buffering when requests come in rapidly.

### DIFF
--- a/src/modules/TrafficManagementModule.cpp
+++ b/src/modules/TrafficManagementModule.cpp
@@ -1092,8 +1092,7 @@ bool TrafficManagementModule::shouldRespondToNodeInfo(const meshtastic_MeshPacke
     if (!sendResponse)
         return true;
 
-    // Checked here, once we know a reply would actually go out, so requests we decline for other
-    // reasons do not consume the budget.
+    // Checked here, once a reply would actually go out, so declined requests do not consume the budget.
     if (!directResponseAllowed(getFrom(p), clockMs())) {
         TM_LOG_DEBUG("NodeInfo direct response throttled for 0x%08x", getFrom(p));
         return false;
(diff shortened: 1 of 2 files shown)


caveman99:
Race condition: valid, fixed. `directResponseAllowed` is reached from the packet path while the module is also an OSThread, which is why the cache has a lock. It now takes the same `cacheLock`, rather than a second lock, to avoid introducing a lock ordering relationship. It is only called after the earlier scoped guard has been released, so there is no recursion. test_traffic_management still passes, 47 cases.

Comment length: condensed.

Throttle::isWithinTimespanMs: skipping this one. It reads `millis()` internally, but TMM routes every time read through the injectable `clockMs()` so tests can advance `s_testNowMs` on a virtual timebase instead of sleeping. Using the helper here would bypass that clock and make the throttle untestable under test_traffic_management. The raw subtraction is deliberate for that reason and is rollover safe in the same way.

NomDeTom:
@caveman99 would you mind looking at https://github.com/meshtastic/firmware/pull/11050 ?
It contains some of the fixes you're putting in (or potentially does, anyway).

NomDeTom:
> This path is opt-in and defaults to off, so this affects only meshes that enabled it.

I think this should be on by default if the function is enabled. It may be useful to add this as a flag in one of the many nodeDBs we now have.

PR #11110, opened 2026-07-20
feat: preserve normal radio profiles across event firmware

RCGV1:
## Event-only scope
This PR does **not** add or change channel QR behavior. Existing QR `Replace` and `Add` remain the app workflow.

## What changes
- normal firmware continues to use `/prefs/config.proto`, `/prefs/channels.proto`, and `/backups/backup.proto`
- event firmware uses `/prefs/event-config.proto`, `/prefs/event-channels.proto`, and `/backups/event-backup.proto`
- on its first boot, event firmware preserves the normal identity and non-radio settings, but applies the event build LoRa defaults and creates event channels from the build defaults
- flashing back to normal firmware naturally resumes the prior channels and LoRa configuration, with no app required

## Storage safety
- first event-profile creation reserves 2,954 bytes: maximum config/channel payloads plus their full-atomic temporary files
- if the filesystem has less space, event firmware runs from its in-memory event defaults but does not write event config or channels; this avoids `saveToDisk()` formatting the filesystem on a failed write
- the next event boot retries profile creation when enough space is available
- no extra persistent RAM is used; only the active profile is loaded
- normal firmware removes the event config, channels, and event backup at boot, releasing all event-only storage; a later event flash starts a fresh event profile

## Safety
- event firmware never copies normal channels into the event table
- corrupt normal config remains on the existing degraded/no-new-identity recovery path
- encrypted-storage migration and lockdown disable cover both normal and event profile files and backups
- backups are profile-specific

## Verification
- `pio test -e native-macos -f test_event_profile_storage`
- `pio run -e native-macos`
- `pio run -e rak3172` (the 14 KiB STM32WL filesystem family)
- native simulator: eight distinct AES-256 normal channels plus custom 250 kHz/SF9/CR6 LoRa -> official Open Sauce 2026 two-channel US Short Turbo/channel-31 profile -> normal firmware; normal config and channel files had identical SHA-256 hashes before, during, and after the event boot
- full aggregate native suite was started but intentionally stopped because it rebuilds each of 38 test targets independently; focused coverage and both relevant target builds completed successfully

Design background: https://github.com/meshtastic/design/issues/120#issuecomment-5026489974

## Summary by CodeRabbit

* **New Features**
  * Added event-mode–specific configuration, channel, and backup storage, including preflight checks to ensure safe atomic event-profile writes.
  * Automatically derives an event configuration from the standard one while preserving identity/security details and applying event-specific radio settings.
* **Bug Fixes**
  * Improved boot persistence behavior by deferring automatic saves until configuration is successfully verified, and adding safe recovery when the event configuration is missing.
  * Enhanced encrypted-storage lockdown handling by migrating the relevant standard/event radio-profile files accordingly.
* **Tests**
  * Added embedded tests covering file isolation, identity preservation, capacity boundaries, and boot persistence deferral/recovery.


RCGV1:
Review follow-up: force-pushed `65f92fca` with three fixes before broad review:
- decrypt/migrate both normal and event profile files plus their backups across lockdown transitions
- preserve `DECODE_FAILED` when the normal config is corrupt during first event boot, retaining the existing degraded/no-new-identity behavior
- isolate event backups at `/backups/event-backup.proto` so normal restore cannot consume event data

The draft remains blocked only on hardware verification and the existing macOS native-test `argp.h` host-toolchain failure.

RCGV1:
Updated to `0bc11766`: applied the repository Trunk formatter; `trunk fmt src/mesh/NodeDB.cpp` now reports no issues.

The remaining `check-label` failure is metadata-only. My fork token cannot add labels to the upstream PR; could a maintainer add `enhancement`? The build/test matrix is still running.

RCGV1:
macOS native simulator verification on `0bc11766` (same persistent VFS, no erase between firmware images):

- regular firmware first boot created `/prefs/config.proto` and `/prefs/channels.proto`
- event firmware boot loaded the ordinary config, logged `Initialized event config without modifying /prefs/config.proto`, created `/prefs/event-config.proto` and `/prefs/event-channels.proto`, and ran the compiled `US` / `EventSim` / `SHORT_FAST` radio profile
- regular firmware booted again and loaded the original ordinary config and channels, resuming `UNSET` / `LongFast`
- the ordinary config and channel SHA-256 values were identical before the event boot and after the return:
  - config: `9582708236b98ecd5ba5b4db7e74a2850cbe09c4661bc2531666cc890795a299`
  - channels: `d2b27729b144786829d0175c6e6dbc029a7bd735baf20ef49d310cc3f33e8bdf`

The stable PKI-derived node number also remained `0x3678de25`; the different number at fresh normal first boot was the expected pre-PKI MAC-derived provisional ID. No source changes from this verification.

RCGV1:
Follow-up verification found and fixed a real degraded-boot persistence bug.

When the normal config was corrupt on the first event boot, the original change kept config.proto and channels.proto intact but still rewrote device.proto and nodes.proto. That could change the saved node identity despite the intended "freezing identity" behavior.

This update defers automatic protobuf writes during initialization until config loading is verified. A decode-failed boot now remains degraded and radio-silent without writing normal files, event profile files, device state, or the node database. Explicit runtime configuration saves remain available after initialization.

Verification:
- TDD: added the focused boot-persistence regression test; pio test -e native-macos -f test_event_profile_storage passes (4 tests).
- Built both normal and event-configured native-macos firmware successfully.
- Native simulator: normal -> event -> normal retains the separate profiles and expected radio settings.
- Corrupt normal config before the first event boot: config, channels, device state, and node database hashes stayed unchanged; no event config or channels file was created.
- Corrupt existing event config: normal config/channels/device/node hashes stayed unchanged, and the corrupt event config was not replaced.

The event-only userPrefs file used for the simulator remains a local test fixture and is not part of the PR.

RCGV1:
@jp-bennett @vidplace7 Please review

src/mesh/NodeDB.h
```diff
@@ -112,11 +112,65 @@ extern meshtastic_Position localPosition;
...
+
+// Reserve event files and their atomic temporaries before seeding, preventing retry formatting on full storage.
+static constexpr size_t EVENT_PROFILE_STORAGE_RESERVATION_BYTES = 2 * (meshtastic_LocalConfig_size + meshtastic_ChannelFile_size);
+
+constexpr bool hasEventProfileStorageSpace(size_t totalBytes, size_t usedBytes)
+{
+    return usedBytes <= totalBytes && totalBytes - usedBytes >= EVENT_PROFILE_STORAGE_RESERVATION_BYTES;
+}
+
+constexpr bool shouldDiscardEventProfile(bool eventMode)
```
jp-bennett: This seems hilariously overcooked.

jp-bennett: (that function in particular)

src/mesh/NodeDB.cpp
```diff
@@ -2546,6 +2614,21 @@ void NodeDB::loadFromDisk()
                 EncryptedStorage::migrateFile(fn);
             }
         }
+
+        // Migrate inactive radio files explicitly: they may contain PSKs and are outside saveToDisk().
+#ifdef FSCom
```
jp-bennett, line 2652: Probably only needs to run in event mode?

jp-bennett:
In general I really think this is the right approach.

PR #11114, opened 2026-07-21
SEN5X: fix version parsing, VOC index reporting, and read-buffer handling

Musashii-kun:
## What this PR does

Fixes several bugs in the SEN5X air-quality sensor driver, plus hardening
against reads of uninitialized buffer memory.

### Behavioral fixes

- **`getVersion()` parsed mostly-unread memory.** It requested only 3 raw
  I2C bytes (= 2 data bytes) but parsed `versionBuffer[0..6]`, so hardware
  and protocol versions were always 0. It now requests the full 12-byte
  reply (8 data bytes) and validates the length actually consumed.
- **Minor version numbers were always truncated to .0.** The version math
  used uint8_t division (`minor / 10`), so firmware 2.2 was reported and
  stored as 2.00. Now divides by `10.0f`.
- **SEN54 never reported its VOC index.** `getMetrics()` gated
  `pm_voc_idx` on `noxIndex` (NOx is SEN55-only) instead of `vocIndex`.
- **The 10-second fan-cleaning wait could exit immediately.**
  `startCleaning()` truncated `millis()` into a `uint16_t` before the
  elapsed-time comparison.
- **"Unknown" sensor readings were published as plausible-looking garbage.**
  The sensor reports unavailable values as 0xFFFF (unsigned) / 0x7FFF
  (signed) per the datasheet; nothing filtered them, so e.g. an
  uninitialized VOC index was reported as 3276.7. `readValues()` /
  `readPNValues()` now map them to the UINT16_MAX / UINT32_MAX / FLT_MAX
  sentinels that `getMetrics()` checks, making those gates functional for
  the first time.
- **PN unit conversion dropped a digit of resolution.** Converting #/cm3
  to #/0.1l as `raw / 10 * 100` truncated the 0.1-resolution digit
  (up to 90 #/0.1l); now computed as `raw * 10`.

### Hardening / cleanup

- `readBuffer()` callers only rejected a return of 0 before indexing fixed
  offsets in the destination buffer. All five call sites now require the
  byte count they actually consume, and destination buffers are
  zero-initialized (also silences maybe-uninitialized warnings).
- Removed dead code: `isnan()` on integer intermediates (always false, so
  the sentinel fallback branches never ran) and an always-true state check
  in `wakeUp()`.
- Added missing `return false` in the non-`FSCom` branches of
  `loadState()`/`saveState()`, which fell off the end of a bool function.

## Testing

Changes are confined to `src/modules/Telemetry/Sensor/SEN5XSensor.{cpp,h}`
(only compiled when the air-quality sensor is enabled), so regression
surface on unrelated devices is limited.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

I do not have hardware to test on at the moment; community testing appreciated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of sensor communication by validating command/reply sizes, rejecting invalid payloads, and treating short reads as failures to prevent partial/corrupted measurements.
  * Corrected firmware/hardware/protocol version parsing, and tightened product-name reply validation.
  * Prevented VOC state corruption by isolating updates until reads fully succeed.
  * Fixed VOC index availability reporting in metrics and improved handling of “unavailable” readings.

* **Documentation**
  * Clarified expected command payload/CRC byte handling for the sensor interface.


src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -22,16 +22,18 @@ bool SEN5XSensor::getVersion()
...
     delay(20); // From Sensirion Datasheet
 
-    uint8_t versionBuffer[12]{};
-    size_t charNumber = readBuffer(&versionBuffer[0], 3);
-    if (charNumber == 0) {
-        LOG_ERROR("%s: Error getting data ready flag value", sensorName);
+    // The version reply is 8 data bytes (12 raw bytes with CRCs):
+    // fw major/minor, fw debug, hw major/minor, protocol major/minor, padding
+    uint8_t versionBuffer[8]{};
+    size_t charNumber = readBuffer(&versionBuffer[0], 12);
```
caveman99: Why do you ignore the CRC bytes? Clearly 12 is correct here though 8 is technically not wrong.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -108,6 +110,13 @@ bool SEN5XSensor::sendCommand(uint16_t command)
 
 bool SEN5XSensor::sendCommand(uint16_t command, uint8_t *buffer, uint8_t byteNumber)
 {
+    // The payload is consumed in 2-byte pairs (each followed by a CRC byte);
+    // an odd size would read past the caller's buffer and overrun toSend
+    if (byteNumber % 2 != 0) {
```
caveman99: 2 byte plus CRC byte is 3. Same reasoning as with the version string, why artificially ignore the CRC?

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -163,6 +172,13 @@ bool SEN5XSensor::sendCommand(uint16_t command, uint8_t *buffer, uint8_t byteNum
 
 uint8_t SEN5XSensor::readBuffer(uint8_t *buffer, uint8_t byteNumber)
 {
+    // Replies arrive in 3-byte groups (2 data bytes + 1 CRC); any other size
+    // would underflow the read loop below and overrun the destination buffer
+    if (byteNumber == 0 || byteNumber % 3 != 0) {
```
caveman99: Now it is 3 when you read stuff?

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -367,15 +383,18 @@ bool SEN5XSensor::vocStateFromSensor()
...
+    uint8_t stateBuffer[SEN5X_VOC_STATE_BUFFER_SIZE]{};
+    size_t receivedNumber = readBuffer(&stateBuffer[0], SEN5X_VOC_STATE_BUFFER_SIZE + (SEN5X_VOC_STATE_BUFFER_SIZE / 2));
     delay(20); // From Sensirion Datasheet
 
-    if (receivedNumber == 0) {
+    if (receivedNumber < SEN5X_VOC_STATE_BUFFER_SIZE) {
         LOG_DEBUG("%s: Error getting VOC's state", sensorName);
         return false;
     }
+    memcpy(vocState, stateBuffer, SEN5X_VOC_STATE_BUFFER_SIZE);
```
caveman99, line 382: If you are worried about CRC errors you need to actually compute and compare the CRC, not copy what you received. This way you could as well dump the buffer alltogether.

caveman99 (changes requested):
This review was hand-made by an actual human. Please be so kind not to let AI answer it.

oscgonfer:
I'll check this in the coming days with hardware. Thanks for submitting!

PD: I don't use AI tools myself, so this feedback will not be quick.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
Musashii-kun: This was confusing to me too when I first read the code.  If you jump to the definition of `readBuffer()` in your IDE and carefully read the function body, you'll see that the CRC bytes are checked during the read operation and then stripped out before the function writes to its output buffer, which only receives 2/3 of the specified read size.

Essentially the function exposes some of its internal logic to the caller in the form of different read vs output / write sizes.

Your comment inspired me to better document this behavior at the function's declaration (in the header).  I also checked the remaining two call sites and their output buffers are actually oversized.

The two discrete units used in this API is genuinely confusing.  I've intentionally not changed the behavior of this API in this PR, only added genuine bug fixes.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
Musashii-kun: See my reply above - this `sendCommand()` API takes the input data size and inserts CRCs after every 2 bytes before sending its 3-byte internal buffer.  The API is indeed genuinely confusing.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
Musashii-kun: Yes, the internal implementation a few lines below this comment can explain everything.  This check just makes the existing code's behavior observable and self-documenting.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
Musashii-kun: This change establishes an all-or-nothing staging invariant so that the data from a failed read can never reach `vocState`.  This is important because the data eventually gets persisted to flash and uploaded back to the sensor.

Note that the CRC is calculated internally by `readBuffer()` and that we never get to see the CRC bytes at the call site.

Musashii-kun:
@caveman99 in response to your first comment: I know first-hand how frustrating it is to receive a copy-pasted AI code review.  If I wanted an AI to review my PRs I would ask it to do so.  Seeing someone copy-paste an AI's output into a review comment (or response!) is disrespectful to both the reviewer and the original code author.

I have been up front in the commit messages that I used Claude Fable to assist me in writing this PR.  But I assure you that I personally reviewed every line of the commits.

Thank you for your review comments.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -22,16 +22,18 @@ bool SEN5XSensor::getVersion()
     }
     delay(20); // From Sensirion Datasheet
 
-    uint8_t versionBuffer[12]{};
-    size_t charNumber = readBuffer(&versionBuffer[0], 3);
-    if (charNumber == 0) {
-        LOG_ERROR("%s: Error getting data ready flag value", sensorName);
+    // The version reply is 8 data bytes (12 raw bytes with CRCs):
+    // fw major/minor, fw debug, hw major/minor, protocol major/minor, padding
+    uint8_t versionBuffer[8]{};
```
oscgonfer: If we are moving from doing the buffer requests as before, I suggest we use defines like on `SEN5X_VOC_STATE_BUFFER_SIZE` and use the same format for all the requests (including that on the buffer definition, and then the + <buffer_size>/2)

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -48,12 +50,13 @@ bool SEN5XSensor::findModel()
     }
     delay(50); // From Sensirion Datasheet
 
-    const uint8_t nameSize = 48;
-    uint8_t name[nameSize];
-    size_t charNumber = readBuffer(&name[0], nameSize);
+    // The product name reply is 32 data bytes, carried in 48 raw bytes with CRCs
+    const uint8_t nameSize = 32;
```
oscgonfer: Same, use defines if what we are looking for is clarity.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
oscgonfer: This doesn't need a check in here, but a comment before the function indicating `@brief` and `@param`

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
oscgonfer: Same as previous.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
oscgonfer: This is fine by me, it's an additional check, although not needed. The CRC check would prevent any VOC state not being passed through. Note that read buffer returns 0 if CRC is not right.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -433,6 +453,7 @@ bool SEN5XSensor::loadState()
     return okay;
 #else
     LOG_ERROR("%s: Filesystem not implemented", sensorName);
+    return false;
```
oscgonfer, line 440: Good.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -478,6 +499,7 @@ bool SEN5XSensor::saveState()
...
```
oscgonfer, line 486: Good.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -539,7 +560,7 @@ bool SEN5XSensor::startCleaning()
     // This message will be always printed so the user knows the device it's not hung
     LOG_INFO("%s: Started fan cleaning it will take 10 seconds...", sensorName);
 
-    uint16_t started = millis();
+    uint32_t started = millis();
```
oscgonfer, line 547: Good.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -721,7 +744,7 @@ bool SEN5XSensor::readPNValues(bool cumulative)
 
     uint8_t dataBuffer[20]{};
```
oscgonfer: Same as above, while we are at it, I'd consolidate this.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -738,22 +761,29 @@ bool SEN5XSensor::readPNValues(bool cumulative)
...
-    sen5xmeasurement.pN2p5 = !isnan(uint_pN2p5) ? uint_pN2p5 / 10 * 100 : UINT32_MAX;
-    sen5xmeasurement.pN4p0 = !isnan(uint_pN4p0) ? uint_pN4p0 / 10 * 100 : UINT32_MAX;
-    sen5xmeasurement.pN10p0 = !isnan(uint_pN10p0) ? uint_pN10p0 / 10 * 100 : UINT32_MAX;
-    sen5xmeasurement.tSize = !isnan(uint_tSize) ? uint_tSize / 1000.0f : FLT_MAX;
+    // Convert values based on Sensirion Arduino lib.
+    // Raw PN values are #/cm3 with 0.1 resolution; multiplying by 10
+    // converts to #/0.1l without the truncation of dividing first.
+    // The sensor reports unavailable values as 0xFFFF; map those to the
+    // sentinels getMetrics() checks for
+    sen5xmeasurement.pN0p5 = (uint_pN0p5 != 0xFFFF) ? ((uint32_t)uint_pN0p5 * 10) : UINT32_MAX;
```
oscgonfer: Let's use something like `#define `SEN5X_UINT_INVALID 0xFFFF` and 
`#define SEN5X_INT_INVALID 0x7FFF`

oscgonfer: I think the order here matters - we should double check.

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -738,22 +761,29 @@ bool SEN5XSensor::readPNValues(bool cumulative)
...
+    sen5xmeasurement.tSize = (uint_tSize != 0xFFFF) ? (uint_tSize / 1000.0f) : FLT_MAX;
 
     // Remove accumuluative values:
     // https://github.com/fablabbcn/smartcitizen-kit-2x/issues/85
     if (!cumulative) {
-        sen5xmeasurement.pN10p0 -= sen5xmeasurement.pN4p0;
-        sen5xmeasurement.pN4p0 -= sen5xmeasurement.pN2p5;
-        sen5xmeasurement.pN2p5 -= sen5xmeasurement.pN1p0;
-        sen5xmeasurement.pN1p0 -= sen5xmeasurement.pN0p5;
+        if (sen5xmeasurement.pN10p0 != UINT32_MAX && sen5xmeasurement.pN4p0 != UINT32_MAX)
```
oscgonfer, line 763: Good

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -775,10 +805,10 @@ uint8_t SEN5XSensor::getMeasurements()
     }
     delay(20); // From Sensirion Datasheet
 
-    uint8_t dataReadyBuffer[3];
+    uint8_t dataReadyBuffer[2]{};
```
oscgonfer: Same as above, let's use define

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -917,7 +947,7 @@ bool SEN5XSensor::getMetrics(meshtastic_Telemetry *measurement)
                 measurement->variant.air_quality_metrics.has_pm_temperature = true;
                 measurement->variant.air_quality_metrics.pm_temperature = sen5xmeasurement.temperature;
             }
-            if (sen5xmeasurement.noxIndex != FLT_MAX) {
+            if (sen5xmeasurement.vocIndex != FLT_MAX) {
```
oscgonfer, line 934: Good

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
```diff
@@ -22,16 +22,18 @@ bool SEN5XSensor::getVersion()
...
+    size_t charNumber = readBuffer(&versionBuffer[0], 12);
+    if (charNumber < 7) {
+        LOG_ERROR("%s: Error getting device version value", sensorName);
         return false;
     }
 
-    firmwareVer = versionBuffer[0] + (versionBuffer[1] / 10);
-    hardwareVer = versionBuffer[3] + (versionBuffer[4] / 10);
-    protocolVer = versionBuffer[5] + (versionBuffer[6] / 10);
+    firmwareVer = versionBuffer[0] + (versionBuffer[1] / 10.0f);
```
oscgonfer, line 34: Not sure this matters.

oscgonfer:
Hi @Musashii-kun 

I have gone through the changes and have left some comments. Please, take a look and make necessary changes. Once those are done, I'll check with real hardware.

A couple of non-technical notes. The tone in the title and descriptions in the different comments is off-putting in many ways, and can be easily perceived as aggressive. I'd recommend vetting or making some adjustments to things like `unitialized read-hazards`... Also, as a side note, I am curious to know what the purpose of submitting this particular PR is, when you don't have the hardware for other sensor-related / meshtastic projects in your profile. Nothing wrong with it, but all in all, it makes me need to count to 10 and then respond.

Musashii-kun:
Hi @oscgonfer -

Thanks for taking a look. I don't see any inline comments on the PR yet, just your feedback on the main discussion thread.

Regarding the tone - point taken.  The commit messages were written as terse defect descriptions and I can see how they might land wrong.

Regarding the intent and hardware - I ran into a compiler warning in the SENS5X code while working on an unrelated project in the Meshtastic source tree.  I decided to take a side trip and started pulling on threads, which led to the discovery of these issues.  I'm just trying to be a good open-source citizen, nothing more.

I don't own a SEN5X, which is why every fix is sourced from the datasheet and/or Sensirion's own drivers.  This is also why your hardware pass is exactly what it needs.

Musashii-kun:
I've rewritten the commit history in a more neutral tone - hopefully this addresses at least part of the review feedback.

oscgonfer:
> Hi @oscgonfer -
> 
> Thanks for taking a look. I don't see any inline comments on the PR yet, just your feedback on the main discussion thread.

That's strange, I left the comments this morning in-line in the PR itself.

> Regarding the tone - point taken. The commit messages were written as terse defect descriptions and I can see how they might land wrong.

No worries! It's good to be able to talk through as humans and clarify :)

> Regarding the intent and hardware - I ran into a compiler warning in the SENS5X code while working on an unrelated project in the Meshtastic source tree. I decided to take a side trip and started pulling on threads, which led to the discovery of these issues. I'm just trying to be a good open-source citizen, nothing more.
> 
> I don't own a SEN5X, which is why every fix is sourced from the datasheet and/or Sensirion's own drivers. This is also why your hardware pass is exactly what it needs.

Súper! Thanks for opening the PR. Let me know if you see the comments, otherwise I can push some changes myself and then test.

oscgonfer (changes requested):

oscgonfer:
> > Hi @oscgonfer -
> > Thanks for taking a look. I don't see any inline comments on the PR yet, just your feedback on the main discussion thread.
> 
> That's strange, I left the comments this morning in-line in the PR itself.
> 
> > Regarding the tone - point taken. The commit messages were written as terse defect descriptions and I can see how they might land wrong.
> 
> No worries! It's good to be able to talk through as humans and clarify :)
> 
> > Regarding the intent and hardware - I ran into a compiler warning in the SENS5X code while working on an unrelated project in the Meshtastic source tree. I decided to take a side trip and started pulling on threads, which led to the discovery of these issues. I'm just trying to be a good open-source citizen, nothing more.
> > I don't own a SEN5X, which is why every fix is sourced from the datasheet and/or Sensirion's own drivers. This is also why your hardware pass is exactly what it needs.
> 
> Súper! Thanks for opening the PR. Let me know if you see the comments, otherwise I can push some changes myself and then test.

I think this should posted now, my bad! @Musashii-kun

src/modules/Telemetry/Sensor/SEN5XSensor.cpp
Musashii-kun: Functionally it doesn't matter; the only consumer is the `firmwareVer < 2` check, which only depends on the major digit. It just makes the logged version accurate.

Musashii-kun:
@oscgonfer thanks for your review.  I've addressed your comments, please have another look when you get a chance.

oscgonfer (approved):

caveman99 (approved):

PR #11116, opened 2026-07-21
Gate the PKI-decrypt key on signer-proven, not the opportunistic cache

caveman99:
Follow-up to #11050. That PR made NodeDB::copyPublicKey fall through to the ephemeral TrafficManagement NodeInfo cache, and Router::perhapsDecode used it to resolve the sender key on the inbound PKI-decrypt path, so a DM decrypted with an unverified TOFU cache key was stamped pki_encrypted and attributed to p->from.

New NodeDB::copyPublicKeyForDecrypt resolves authoritative keys (hot/warm, TOFU baseline unchanged) and accepts a cold-tier cache key only when it is signer-proven. A previously-authenticated node evicted to the cache still resolves; an unverified TOFU cache entry no longer backs authenticated DM attribution. The outbound encrypt-to path keeps the opportunistic copyPublicKey lookup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secure message decryption to resolve sender public keys only from verified, authoritative sources.
  * If a cached cold-tier key is used, it is now accepted only when it is signer-proven (avoids unverified TOFU cache keys for attribution).
  * Updated decryption gating so only genuine PKI-decrypt candidates trigger pending-key handling and any related fallback flow, reducing incorrect decrypt behavior.


caveman99:
Switched to option 1 from the review: copyPublicKeyForDecrypt accepts a cold-tier cache key only when signer-proven, keeping the hot/warm TOFU baseline and the outbound opportunistic path unchanged.

NomDeTom:
warm store doesn't differentiate manually verified & xeddsa properly. There's a piece of work to do to re-allocate the spurious protected flag, and then split it properly. The protected flag is used at migration, but it's unused after that.

PR #11126, opened 2026-07-21
mesh: suppress relayed broadcast responses

RCGV1:
## Summary
- suppress module replies, bound-channel errors, and fallback NAKs for relayed broadcast `want_response` packets
- retain the narrow fleet-tracker exception for `POSITION_APP` requests only
- cover suppression and Tracker exception behavior in `test_mesh_module`

## Why
A relayed broadcast reaches every receiver; allowing each receiver to reply causes mesh-wide response fan-out.

## Validation
- `pio test -e native-macos -f test_mesh_module` (22 passed)
- `pio run -e native-macos`
- launched `meshtasticd -s` for a five-second macOS SimRadio smoke test
- `git diff --check`

`trunk fmt` was unavailable on this host.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary replies to relayed broadcast requests to help prevent mesh-wide response storms.
  * Ensured suppressed relayed broadcasts no longer generate inappropriate NAKs or authorization error responses.
  * Preserved tracker behavior: relayed position requests still receive responses, while other relayed broadcasts do not.
* **Tests**
  * Added automated coverage for relayed broadcast scenarios, including role-based response/NAK expectations.

### DIFF
--- a/src/mesh/MeshModule.cpp
+++ b/src/mesh/MeshModule.cpp
@@ -108,11 +108,11 @@ void MeshModule::callModules(meshtastic_MeshPacket &mp, RxSource src)
     // Was this message directed to us specifically?  Will be false if we are sniffing someone elses packets
     auto ourNodeNum = nodeDB->getNodeNum();
     bool toUs = isBroadcast(mp.to) || isToUs(&mp);
-    const bool relayedBroadcastResponseRequest = isDecoded && mp.decoded.want_response && isBroadcast(mp.to) &&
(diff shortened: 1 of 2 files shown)


NomDeTom:
Why would you remove the response to the broadcast want_response packets, but not the mechanism that generates them?

NomDeTom:
Having reviewed this in some depth, I don't believe it has any merit and believe it actively wastes airtime.

PR #11130, opened 2026-07-21
Run native PlatformIO tests by area for readable failures

caveman99:
The native test job ran every `test_*` suite in a single `platformio test -e coverage` invocation. As the suite set grows, a failure deep in that one run can land past the log's viewable limit, so the failing assertion is not visible without downloading the raw log.

## Change

Same job, same runner, same coverage. The run is split by area:

- **Build once.** `platformio test -e coverage --without-testing` compiles src and every test program a single time.
- **Run by area.** The suites are grouped into areas (admin, crypto, routing, position, fuzz, packets, io, misc) and run in sequential `--without-building` invocations that reuse that build. Each area writes its own JUnit report (`testreport-<area>.xml`) and its own collapsible log, so a failure is in a small named section instead of buried.
- **Coverage unchanged.** The runs are sequential and share one build dir, so gcov counts still accumulate into `.pio/build/coverage/src`; a single capture holds the union. `generate-reports` merges base + integration + tests as before.

## Grouping

Areas are ordered `name:regex` rules, first match wins, with a catch-all `misc`, so a newly added suite always runs even before it is placed in an area. Add a suite to an area by extending its regex; add an area with one rule line. All 38 current suites map with none dropped.

Sequential, so no build isolation or gcov races - this trades nothing for the readability, unlike a concurrent or matrix split which would each cost extra builds or runners.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved native CI test reliability by splitting execution into ordered test areas that run sequentially while reusing a shared build.
  * Enhanced CI test reporting by generating per-area JUnit reports and then consolidating them into a single combined `testreport.xml`.
  * Updated code coverage handling to automatically merge all discovered `coverage_*.info` inputs into a unified coverage result.

### DIFF
--- a/.github/workflows/test_native.yml
+++ b/.github/workflows/test_native.yml
@@ -163,6 +163,26 @@ jobs:
           done
           exit $fail
 
+      - name: Merge per-area reports into testreport.xml
+        # Preserve the single-file JUnit contract that downstream consumers rely on
+        # (pr_tests.yml's summary and generate-reports' Test Report both read testreport.xml).
+        # The per-area split is only for readable logs; the report stays consolidated.
+        if: always() # run even when a chunk failed, so the report captures the failures
+        shell: bash
+        run: |
+          python3 - <<'PY'
+          import glob, xml.etree.ElementTree as ET
+          out = ET.Element('testsuites')
+          for f in sorted(glob.glob('testreport-*.xml')):
+              try:
+                  root = ET.parse(f).getroot()
+              except ET.ParseError:
+                  continue
+              # PlatformIO writes a <testsuites> root; fold in a bare <testsuite> too, just in case.
+              out.extend(root.findall('testsuite') if root.tag == 'testsuites' else [root])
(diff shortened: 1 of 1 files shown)


caveman99:
Good catch, fixed in the latest commit.

The per-area split was only meant to make the *logs* readable; I unintentionally also split the JUnit report, breaking the single-file `testreport.xml` contract that `pr_tests.yml` (and this workflow's own Test Report step) read.

The per-area runs still write `testreport-<area>.xml`, but a new merge step now folds them into a single `testreport.xml` (one `<testsuites>` root with every suite's `<testsuite>`), and that consolidated file is what gets uploaded. So `pr_tests.yml:84` finds `testreport.xml` and its detailed summary runs unchanged, and generate-reports is back to the singular `testreport.xml` for its sed + dorny steps. The merge runs `if: always()` so a failing run still produces the report.

Verified the merge preserves the downstream parse: two area files with three suites merge to one root, and `.//testsuite` sums to the right totals and failure count.

PR #11131, opened 2026-07-21
Limit NodeInfo discovery replies to direct neighbors

RCGV1:
## Summary
- send broadcast NodeInfo discovery requests with a zero hop limit
- accept NodeInfo discovery replies only for direct radio receptions and force their replies to zero hops
- keep unicast NodeInfo exchange behavior unchanged, and suppress NAKs for relayed or unknown-hop discovery copies

## Validation
- `trunk check`
- `pio test -e native -f test_mesh_module` *(blocked before test compilation on macOS: Portduino requires `argp.h`)*
- `./bin/run-tests.sh` *(blocked on macOS Bash 3.2 / BSD `find`: `mapfile` and `-printf` unavailable)*
- `pio run -e heltec-v3` *(blocked by incomplete local Xtensa toolchain install; compiler not found)*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of response hop-limit handling across modules, ensuring replies use the correct hop limit per request.
  * Updated NodeInfo behavior for broadcast discovery: only qualifying direct broadcast discovery requests now generate replies, which use a hop limit of zero; other broadcast discovery packets are ignored. Unicast NodeInfo requests keep normal routing behavior.

* **Tests**
  * Extended Unity tests to validate module-controlled reply hop limits and direct vs. relayed/unknown broadcast discovery qualification, including suppression/allowance sequencing.


RCGV1:
@coderabbitai review

RCGV1:
@coderabbitai review

RCGV1:
@coderabbitai review

RCGV1:
@coderabbitai review

jp-bennett:
For a while we talked about dynamically detecting we were on a busy mesh. This seems like another optimization that should be gated to busy meshes, EG, when the nodedb is full and nodes are regularly falling off the end of the DB.

RCGV1:
I think this should be on all meshes. No node should be able to request the node info of all nodes in the mesh. The only ones that make sense are direct neighbors.

PR #11141, opened 2026-07-22
Add RadioMaster Nomad primary 915 MHz support

h3lix1:
## Summary

This draft is now a small hardening delta on top of #5318's `nomad-gemini` branch by @caveman99. It keeps the original Nomad implementation and attribution intact while adding the three safety items called out in review:

- hold the unused second LR1121 in reset with chip-select inactive
- hold the unsupported ESP32-C3 backpack in reset
- mark the primary LR1121 path sub-GHz-only until dual-radio/2.4 GHz support exists

The PR also adds target metadata and focused board documentation. The broader external-PA and fan rewrites from the earlier draft have been removed.

## Implementation

- Adds `LR1121_SUBGHZ_ONLY` handling and enables it only for the Nomad target.
- Drives second-radio CS GPIO13 high and reset GPIO21 low before primary SPI initialization.
- Drives ESP32-C3 backpack reset GPIO19 low while that path is unsupported.
- Retains `JANUS_RADIO` as the future dual-radio feature marker while the second LR1121 remains inactive.
- Registers the prototype target as `PRIVATE_HW`, support level 3, and `board_level = extra`.
- Documents the primary and reserved pin assignments, RF-switch table, PA lifecycle, fan behavior, and current limitations.

## PA calibration

This stacked delta inherits #5318's external-PA implementation and provisional table unchanged, including APC2 120 at 10 mW. That value is not claimed as bench-validated; the table should be updated after the planned RF power-meter measurements.

### Software

- [x] `trunk fmt`
- [x] `pio run -e radiomaster_nomad_gemini`
- [x] Final comparison against `nomad-gemini`: 5 files, +56/-5
- [x] Final merge tree preserves the newer PKI and LR2021 autoLDRO changes carried by the base

### Hardware

- [x] Firmware `2.8.0.5081c59` flashed to the connected ESP32-PICO-D4 with on-device hash verification
- [x] Booted as `radiomaster_nomad_gemini` with US / MediumFast / 10 dBm
- [x] Received and decoded live 915 MHz mesh traffic
- [x] Completed outbound RF transmission; serial logs showed the traceroute request TX complete
- [x] Completed a multi-hop traceroute to `!625aa4fb`, including a return path
- [x] Joined the configured Wi-Fi network, obtained `192.168.1.252`, responded to ICMP, exposed TCP 4403, and completed a Meshtastic API configuration exchange
- [x] Fan disable and enable states both wrote and read back; final state is enabled
- [x] Broadcast validation text was accepted by the device API

## Suggested review focus

- second-LR1121 and ESP32-C3 reset polarity/timing
- scope of the sub-GHz-only guard
- metadata and documentation accuracy

### DIFF
--- a/src/mesh/LR1121Interface.cpp
+++ b/src/mesh/LR1121Interface.cpp
@@ -11,6 +11,10 @@ LR1121Interface::LR1121Interface(LockingArduinoHal *hal, RADIOLIB_PIN_TYPE cs, R
 
 bool LR1121Interface::wideLora()
 {
+#ifdef LR1121_SUBGHZ_ONLY
+    return false;
+#else
     return true;
+#endif
 }
 #endif
--- a/src/platform/extra_variants/radiomaster_nomad_gemini/variant.cpp
+++ b/src/platform/extra_variants/radiomaster_nomad_gemini/variant.cpp
@@ -1,0 +1,97 @@
+#include "configuration.h"
+
+#ifdef RADIOMASTER_NOMAD_GEMINI
+
+#include "RadioExternalPa.h"
+#include <Arduino.h>
+
+/**
+ * RadioMaster Nomad Gemini external analog PA driver.
+ *
+ * The Nomad's PA is biased by an analog control voltage on the APC2 pin
+ * (RADIO_PA_APC2_PIN = GPIO26 = ESP32 DAC channel 2), driven with dacWrite()
+ * (0-255 -> 0-3.3V). The DAC is essentially a fixed bias; the actual output power
+ * is set by driving the LR1121 chip at a (negative) output power and letting the PA
+ * add ~25-31 dB of gain.
+ *
+ * The calibration below is derived from the ExpressLRS hardware target
+ * "TX/Radiomaster Nomad.json" (power_values = APC2 DAC codes, power_values2 = chip
+ * output dBm). These values reproduce ELRS behaviour but have NOT been validated on
+ * a bench against a power meter for Meshtastic — verify before trusting at 1 W, both
+ * for PA safety and for regulatory compliance.
+ *
+ * TODO: dual-band (Gemini) operation uses ELRS power_values_dual and a second radio;
+ * once JANUS_RADIO is implemented, branch on freqHz / active radio here.
+ */
+
+// total dBm corresponds to the ELRS power presets 10/25/50/100/250/500/1000 mW.
+static const struct {
+    int8_t totalDbm; // desired total radiated power
+    int8_t chipDbm;  // LR1121 output power register (ELRS power_values2)
+    uint8_t apc2;    // APC2 DAC code (ELRS power_values)
+} NOMAD_PA_CAL[] = {
+    {10, -17, 120}, // 10 mW
+    {14, -16, 120}, // 25 mW
+    {17, -14, 120}, // 50 mW
+    {20, -11, 120}, // 100 mW
+    {24, -7, 120},  // 250 mW
+    {27, -3, 120},  // 500 mW
+    {30, 5, 95},    // 1000 mW (1 W)
+};
+static const size_t NOMAD_PA_CAL_N = sizeof(NOMAD_PA_CAL) / sizeof(NOMAD_PA_CAL[0]);
+
+// APC2 bias chosen for the currently-configured power level; applied on TX, dropped
+// to 0 for RX/idle/sleep so the PA only draws current while transmitting.
+static uint8_t nomadApc2Level = 0;
+
+// Runs at the start of setup(), before the LoRa radio is initialized.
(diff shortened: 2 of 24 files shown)


caveman99:
Thanks for crediting #5318 and for the useful additions in this draft, in particular holding the second LR1121 and the ESP32-C3 backpack in reset, and the LR1121_SUBGHZ_ONLY guard.

That said, this is a parallel slice of work that is still in flight. Measured against the respective develop merge bases (normalized, non-empty diff lines), 123 of the 197 lines added here (62 percent) are byte-identical to the #5318 branch, and a further share is the same content reformatted or reworded. The genuinely new parts are exactly the ones I would like to see folded into the original work rather than competing with it.

Request: please change this PR's base from develop to the nomad-gemini branch. That turns it into a reviewable delta on top of the existing work, avoids two competing Nomad targets, and keeps attribution clean. The base can be switched in the PR header without losing the branch or its history.

One technical note: the 10 mW APC2 bias differs between your mLRS-derived table (150) and the ELRS target values (120) used in #5318. I have an ImmersionRC RF power meter on the bench and will validate the full PA table against measured output shortly. Whichever numbers survive the bench measurement should go in, for both PRs' sake.

h3lix1:
@caveman99 Appologies for stepping on your PRs. I didn't realize you were working on the same thing.

This was on the path to check if the Normad can do dual channel as a proof of concept.. which it does.. and it works for RX and TX for both.. but I am not going to make any more progress on this beyond this PR and #11160. It was more of a conversation piece rather than a PR I was planning to merge. The ability to support dual primary channels will require a lot more effort than I have the time to implement properly.

Can I offer you a Nomad as a peace offering? Two 900mhz radios, 1w each, prime for being setup as a dual preset router.

variants/esp32/radiomaster_nomad_gemini/variant.h
```diff
@@ -35,12 +36,12 @@
 // 2.4G Part
 #define LR1120_MAX_POWER 5
 
-// not yet implemented
-#define JANUS_RADIO
```
caveman99, line 39: Please leave the Janus define in place. It's a future macro to enable the dual radio code. Right now it does nothing.

variants/esp32/radiomaster_nomad_gemini/variant.h
```diff
@@ -19,6 +19,7 @@
 
 // supported modules list
 #define USE_LR1121
+#define LR1121_SUBGHZ_ONLY
```
caveman99, line 22: I hope this is not needed any more once both PA curves are confirmed.

src/platform/extra_variants/radiomaster_nomad_gemini/variant.cpp
```diff
@@ -47,6 +47,12 @@ static uint8_t nomadApc2Level = 0;
 // Runs at the start of setup(), before the LoRa radio is initialized.
 void earlyInitVariant()
 {
+    pinMode(LR1121_SPI_NSS2_PIN, OUTPUT);
```
caveman99, line 50: Please add documentation you are holding the 2nd LR and the C3 under reset. By the way - did you measure the power consumption when doing this? might as well send them to deep sleep instead.

src/mesh/LR1121Interface.cpp
```diff
@@ -11,6 +11,10 @@ LR1121Interface::LR1121Interface(LockingArduinoHal *hal, RADIOLIB_PIN_TYPE cs, R
 
 bool LR1121Interface::wideLora()
 {
+#ifdef LR1121_SUBGHZ_ONLY
```
caveman99, line 14: As mentioned elsewhere i hope we don't need this any more once the PA curves are confirmed.

caveman99 (changes requested):

caveman99:
> @caveman99 Appologies for stepping on your PRs. I didn't realize you were working on the same thing.
> 
> This was on the path to check if the Normad can do dual channel as a proof of concept.. which it does.. and it works for RX and TX for both.. but I am not going to make any more progress on this beyond this PR and #11160. It was more of a conversation piece rather than a PR I was planning to merge. The ability to support dual primary channels will require a lot more effort than I have the time to implement properly.
> 
> Can I offer you a Nomad as a peace offering? Two 900mhz radios, 1w each, prime for being setup as a dual preset router.

Thanks for the offer, i have a device here, and finally got a matching power supply to eanble the PA. No need to apologize, this item has been on my back burner for quite some time, i initially planned for it to be exactly what you had in mind, a test mule for dual radio support. I coined it project Janus (Roman god with 2 faces), to go in the same tone as the Gemini product name. What do you say we finalize this work and then go on with #11160 to finish the dream? :-)

h3lix1:

PR #11145, opened 2026-07-22
ci: warn and skip the pio check job on transient network failures

thebentern:
## Problem

The `check` matrix job intermittently fails before it analyses anything. `pio check` has to fetch the platform package before cppcheck can start, and that download gets reset mid-flight often enough to be a nuisance.

Three of the last ~200 CI runs died this way — all identical, all ~1 second in:

| run | board | failure |
| --- | --- | --- |
| [29883921204](https://github.com/meshtastic/firmware/actions/runs/29883921204) | station-g3 | `ConnectionResetError(104)` |
| [29880715107](https://github.com/meshtastic/firmware/actions/runs/29880715107) | heltec-v3 | `ConnectionResetError(104)` |
| [29858944295](https://github.com/meshtastic/firmware/actions/runs/29858944295) | tbeam | `ConnectionResetError(104)` |

```
Checking station-g3 > cppcheck (board: station-g3; platform:
https://github.com/pioarduino/platform-espressif32/releases/download/55.03.39/platform-espressif32.zip)
requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))
```

cppcheck never ran, so there is no signal at all — just a red X someone has to re-run by hand.

## Fix

`pio check` exits 1 for both a dead socket and a real defect, so the exit code can't tell them apart — only the output can. `bin/check-all.sh` now tees the run and classifies the failure:

- **cppcheck reported something** (the summary table, or a `file:line: [severity]` defect line) → real, propagate the exit status. This is checked *first*, so no amount of network noise elsewhere in the log can mask a genuine defect.
- **Transport-level error and nothing else** → log a `::warning::` annotation naming the board and the underlying error, and exit 0 under CI. Locally it exits 2, so a local run never quietly reports itself clean.
- **Anything else** → propagate the exit status.

`404` / `403` / `UnknownPackageError` are deliberately *not* transient: a missing or forbidden package is a reproducible break from a bad pin in a `.ini`, not weather.

No workflow change is needed — `gh-action-firmware`'s entrypoint runs `/workspace/bin/check-all.sh` for `MT_TARGET=check`, so the whole fix lives in this repo and applies to local runs too.

Scope is `check` only. The `build` jobs use the same action but must keep failing loudly — they produce release artifacts.

## Verification

Replayed the captured CI logs through the script with a stubbed `pio`:

| case | expected | result |
| --- | --- | --- |
| station-g3 connection-reset log, CI | exit 0 + warning | ✅ |
| same log, local | exit 2, no warning | ✅ |
| rak11200 `Total 1 0 0` defect log | exit 1 | ✅ |
| that defect log with the reset traceback appended | exit 1 | ✅ |
| 404 / `UnknownPackageError` | exit 1 | ✅ |
| clean pass | exit 0 | ✅ |

The skipped case looks like this in the log, and shows up as an annotation on the run:

```
RESULT: SKIPPED station-g3 - transient network failure: ConnectionResetError: [Errno 104] Connection reset by peer
::warning title=Static analysis skipped (station-g3)::pio check could not download its packages: ConnectionResetError: [Errno 104] Connection reset by peer
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved static analysis to better separate real code defects from temporary network/service disruptions.
  * Added CI-friendly handling for transient check failures, emitting clear warnings while still reporting genuine analysis issues accurately.
  * Refined exit behavior so local and CI runs distinctly identify transient problems without masking legitimate failures.

### DIFF
--- a/bin/check-all.sh
+++ b/bin/check-all.sh
@@ -31,16 +31,18 @@ export APP_VERSION=$VERSION
 
 if [[ $# -gt 0 ]]; then
 	# can override which environment by passing arg
-	BOARDS="$*"
+	BOARDS=("$@")
 else
-	BOARDS="tlora-v2 tlora-v1 tlora_v1_3 tlora-v2-1-1.6 tbeam heltec-v2.0 heltec-v2.1 tbeam0.7 meshtastic-diy-v1 rak4631 rak4631_eink rak11200 t-echo canaryone pca10059_diy_eink"
+	BOARDS=(tlora-v2 tlora-v1 tlora_v1_3 tlora-v2-1-1.6 tbeam heltec-v2.0 heltec-v2.1 tbeam0.7 meshtastic-diy-v1 rak4631 rak4631_eink rak11200 t-echo canaryone pca10059_diy_eink)
 fi
 
-echo "BOARDS:${BOARDS}"
+echo "BOARDS:${BOARDS[*]}"
 
-CHECK=""
-for BOARD in $BOARDS; do
-	CHECK="${CHECK} -e ${BOARD}"
+# Array, not a string: board names reach pio as literal arguments, so an override containing
+# whitespace or a glob character cannot turn into extra pio arguments.
+CHECK=()
+for BOARD in "${BOARDS[@]}"; do
+	CHECK+=(-e "${BOARD}")
 done
 
 LOG=$(mktemp)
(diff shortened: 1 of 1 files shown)


thebentern:
Addressed in 876a89a5e — good catch, and it was hiding a second bug.

`BOARDS` and `CHECK` are arrays now, and `pio` is invoked with `"${CHECK[@]}"`, so board names arrive as literal arguments. While converting, the two skip messages needed `"${BOARDS[*]}"` — a bare `"${BOARDS}"` on an array expands to the first element only, which would have quietly named just one board in the warning annotation on a multi-board run.

`shellcheck -x bin/check-all.sh` is now clean (the SC2086 this line had carried since it was written is gone).

Re-verified with a stubbed `pio` that records its argv:

| case | result |
| --- | --- |
| `tb*` | reaches pio as one literal `-e tb*`, no globbing |
| `tbeam extra` | reaches pio as one literal argument |
| default list (no args) | still expands to 15 `-e` pairs |
| `tbeam rak4631` skip message | names both boards |

All six classification cases are unchanged: network/CI → exit 0 + warning, network/local → exit 2, real defects → exit 1, defect log with a reset traceback appended → exit 1, 404 → exit 1, clean pass → exit 0.

caveman99 (approved):

PR #11155, opened 2026-07-22
Fix/call modules

Ixitxachitl:
## Summary

`Router::sendLocal()` was calling `handleReceived()` synchronously for locally-addressed packets. When a module (e.g. an admin/module-config handler) generates a reply from inside `MeshModule::callModules()` and sends it via `service->sendToMesh()`, that synchronous call re-entered `callModules()` from within itself, which caused reply packets to phone requests to be dropped instead of delivered.

This PR restores queuing of locally-addressed packets through `fromRadioQueue` (avoiding the re-entrancy), while fixing the underlying reason queuing had been avoided before: queuing used to erase the packet's `RxSource`, making a local phone/module-generated packet indistinguishable from already-decoded remote ingress once it was dequeued.

### Changes

- `Router::fromRadioQueue` is now a `TypedQueue<QueuedFromRadio>` (packet + `RxSource` pair) instead of a `PointerQueue<meshtastic_MeshPacket>`, so origin is preserved across the queue.
- `Router::enqueueReceivedMessage()` and `Router::perhapsHandleReceived()` now take/forward an `RxSource` (defaulting to `RX_SRC_RADIO` for existing radio/MQTT/UDP callers).
- `Router::sendLocal()` queues locally-addressed packets via `enqueueReceivedMessage(p, src)` instead of calling `handleReceived()` in-line, avoiding re-entrant `callModules()` calls while still replaying the packet with its true origin.
- `MeshModule::callModules()` now explicitly passes `RX_SRC_LOCAL` (and `isToUs(currentReply)`) when sending a reply packet, so it reaches the phone instead of being silently dropped by the loopback guard.

Fixes reply packets (e.g. admin/module-config responses to the phone) being silently dropped.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Tested on a Seeed WIO Tracker L1. Have not been able to test on the other listed hardware — help testing on those devices would be appreciated.


src/mesh/Router.cpp
```diff
@@ -1425,6 +1427,6 @@ void Router::perhapsHandleReceived(meshtastic_MeshPacket *p)
 
     // Note: we avoid calling shouldFilterReceived if we are supposed to ignore certain nodes - because some overrides might
     // cache/learn of the existence of nodes (i.e. FloodRouter) that they should not
-    handleReceived(p);
+    handleReceived(p, src);
```
caveman99, line 1442: locally-addressed packets now traverse perhapsHandleReceived() which means they are dedup ed and checked against the ignore-list. They are also added to the PacketHistory. Confirm a module emitting rapid to-phone-only packets can't get deduped.

src/mesh/Router.cpp
```diff
@@ -327,10 +326,13 @@ ErrorCode Router::sendLocal(meshtastic_MeshPacket *p, RxSource src)
...
-        // phone/module packet indistinguishable from remote already-decoded ingress.
-        handleReceived(p, src);
-        return ERRNO_SHOULD_RELEASE;
+        // Queue rather than call handleReceived() synchronously: a reply generated from inside
+        // MeshModule::callModules() (e.g. an admin/module-config response) lands here via
+        // sendToMesh(), and calling handleReceived() in-line would re-enter callModules() from
+        // within itself. The queue carries src through so the packet is still replayed with its
+        // true origin instead of defaulting to RX_SRC_RADIO.
+        enqueueReceivedMessage(p, src);
+        return ERRNO_OK;
```
caveman99, line 335: return code changed from ERRNO_SHOULD_RELEASE to ERRNO_OK (the queue now owns/frees the packet). Every other sendLocal caller must be verified to no longer release on ERRNO_OK or else we have another double-free/leak.

caveman99 (changes requested):
Overall agreement, please see comments inline.

src/mesh/Router.cpp
Ixitxachitl: You're right, locally-generated replies were falling through the ignore-list/PacketHistory/MQTT/pre-hop checks meant for radio ingress. handleReceived() already special-cases RX_SRC_LOCAL internally (e.g. it only applies the routing-auth cache when src == RX_SRC_RADIO), so I've made perhapsHandleReceived() skip straight to handleReceived() when src == RX_SRC_LOCAL, bypassing the filter pipeline entirely for local packets, same as the pre-PR behavior, just routed through the queue instead of called inline. Pushed in a2c03e5.

src/mesh/Router.cpp
Ixitxachitl: Checked all three sendLocal() callers:

- MQTT.cpp:127 — if (router->sendLocal(pAck) == ERRNO_SHOULD_RELEASE) packetPool.release(pAck);
- RoutingModule.cpp:60 — same pattern
- MeshService.cpp:348 — same pattern

All three only release when the result is ERRNO_SHOULD_RELEASE, never on ERRNO_OK. Since the local/queued path now returns ERRNO_OK, none of them double-release, the queue owns the packet and it's freed once in perhapsHandleReceived() after processing. No double-free/leak.

src/mesh/Router.cpp
Ixitxachitl: Follow-up: that guard was too narrow. RX_SRC_USER (a phone/serial packet addressed to the local node, e.g. a config request) hits the same perhapsHandleReceived() path and wasn't exempted, so it still traversed the ignore-list/dedup/pre-hop/routing-auth filters meant only for radio ingress — the same risk you flagged, just on the request side instead of the reply side. In the pre-PR code RX_SRC_USER never went through perhapsHandleReceived() at all (synchronous handleReceived() call), so this was a regression from queuing it. Widened the check from src == RX_SRC_LOCAL to src != RX_SRC_RADIO to cover both trusted non-radio sources. Pushed.

Ixitxachitl:
This also fixes the same config-get/set phone-timeout regression that #11185 targets (both trace back to #10967 making sendLocal()'s isToUs branch call handleReceived() synchronously) — MeshModule.cpp's existing service->sendToMesh(currentReply, RX_SRC_LOCAL, isToUs(currentReply)) already forces phone delivery at the call site that matters, so no equivalent to #11185's MeshService.cpp change is needed here. Unlike #11185, this approach doesn't reintroduce the synchronous re-entry into callModules(), which I confirmed overflows the stack on nRF52840 (crash on save, e.g. Wio Tracker L1) — #11185 alone still has that crash since it keeps the synchronous handleReceived() call.

Rebasing onto current develop will need two test updates that aren't new regressions, just stale coupling to the old synchronous contract: test_packet_signing's test_C8 asserts the old ERRNO_SHOULD_RELEASE return, and test_mqtt's MockRouter::enqueueReceivedMessage override needs the widened RxSource parameter to compile. Happy to push both.

caveman99:
i opted for approving #11185 since it's the most surgical approach of fixing the regression. Regarding the stack overflow, that'as actually easier to fix now #11185 decouples phone delivery from handleReceived entirely. Add a depth counter to the router. if it is not zero, it's calld from within callModules and you enqueue a copy of the packet with its RxSource into a small local queue, then return ERRNO_SHOULD_RELEASE. Depth 0 has no change to today.

PR #11160, opened 2026-07-23
Prototype dual-radio channels on RadioMaster Nomad

h3lix1:
## Summary

This draft prototypes both LR1121s in the RadioMaster Nomad as independent Meshtastic PHYs:

- the primary radio follows `config.lora` and was tested as US MediumFast at 913.125 MHz
- logical channel 1 uses the second LR1121 with a fixed US ShortTurbo profile at 926.750 MHz
- both paths may transmit, with a nominal 10 dBm hard cap and one-at-a-time TX coordination
- the peer radio enters standby before TX, APC2 is serialized, and both paths return to receive afterward
- two-slot RadioLib ISR dispatch and an alternate LoRa transport tag support the two radio instances
- GPIO2 fan control remains available through `lora.pa_fan_disabled`

This is an experimental continuation of the Nomad bring-up in #5318 and the hardening delta in #11141.

## Why

Meshtastic currently exposes one router-facing radio interface, while logical channels are cryptographic namespaces rather than per-channel PHY configurations. This composite interface is a small test vehicle for assigning one logical channel to each physical LR1121 before attempting a true cross-channel bridge.

## Current limitations

- This does not bridge channel 0 traffic into channel 1. A bridge would need an explicit decrypt, clone, re-encrypt, and loop-control policy.
- Secondary selection uses Meshtastic's 8-bit encrypted channel hash, so a collision is possible.
- PKI direct messages still use channel zero.
- The external PA table and antenna-port isolation are not calibrated.
- No external ShortTurbo peer acknowledged the channel-1 test packet.
- Simultaneous TX and high-power testing are intentionally excluded.

## Validation

- [x] `trunk fmt`
- [x] `git diff --check`
- [x] Clean `pio run -e radiomaster_nomad_gemini` from commit `defe8b60f7973ee94950ebe893d88bb0ed3e06c2`
- [x] Factory image flashed to the ESP32-PICO-D4 with esptool hash verification
- [x] Exact artifact booted as `2.8.0.defe8b6`
- [x] Both LR1121s initialized: MediumFast 913.125 MHz and ShortTurbo 926.750 MHz
- [x] Live MediumFast packets received with RSSI/SNR data; node database reached 120 entries
- [x] Primary traceroute completed:
  - outbound: `!5868023a -> 0b45f21f -> a0352604 -> !4703f8ab`
  - return: `!4703f8ab -> 273c3874 -> !5868023a`
- [x] Channel-1 packet `0x78bac82c` was selected for ShortTurbo and reported TX-done on each retry
- [x] Wi-Fi rejoined the configured network, obtained `192.168.1.252`, answered ICMP, and served the TCP API on port 4403
- [x] Fan disable and enable settings both wrote and read back; final state commands the fan on

The repository test wrapper is blocked on macOS by Bash `mapfile` and BSD `find` lacking `-printf`. Direct native tests are independently blocked by existing macOS-incompatible VLAs and `linux/input.h`; neither failure reaches the Nomad-specific code.

## Review focus

- composite ownership and Router-facing behavior
- dual-instance ISR dispatch
- SPI, standby, APC2, and TX-arbiter sequencing
- logical-channel selection and relay-path preservation
- whether the prototype constraints are sufficient before any calibrated RF work


thebentern:
Super neat!

h3lix1:

PR #11166, opened 2026-07-23
lint: suppress intentional root USER in runtime Dockerfiles

Jorropo:
meshtasticd must run as root to reach GPIO/I2C/SPI hardware, so the final
USER root is deliberate. The existing file-level ignore used the old trivy
code (DS002); rename it to DS-0002 and add checkov/CKV_DOCKER_8 so the
last-USER-root findings are suppressed with an explicit justification.
---

## Summary by CodeRabbit

* **Chores**
  * Updated top-of-file security scan suppression comments in the container build files to reflect current Trivy rule formatting, including the updated Trivy ignore identifier.
  * Added a Checkov suppression entry for the documented requirement that the container runs as root.


Jorropo:
Merging that one.

vidplace7 (approved):

PR #11170, opened 2026-07-23
lint: drop dead CKV2_GHA_1 ignore in main_matrix workflow

Jorropo:
The trunk-ignore no longer matches any finding (the job sets its own
permissions), so it only produced an ignore-does-nothing note.
---

## Summary by CodeRabbit

* **Chores**
  * Updated the continuous integration workflow to remove an outdated inline security-scan suppression that was no longer needed during the artifact collection phase.

### DIFF
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -215,7 +215,6 @@ jobs:
       push: false
 
   gather-artifacts:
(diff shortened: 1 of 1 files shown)


Jorropo:
Merging that one.

vidplace7 (approved):

PR #11173, opened 2026-07-23
lint: drop redundant bandit ignores in c3 exception decoder

Jorropo:
B404 and B603 are already skipped globally in .trunk/configs/.bandit, so
the file-level ignore-all directives never suppressed anything.
---


Jorropo:
Merging that one.

vidplace7 (approved):

PR #11174, opened 2026-07-23
lint: ignore `userPrefs.jsonc`

Jorropo:
checkov/CKV_SECRET_6 flags the commented-out "large4cats" example, which
is the public default credential for the meshtastic.org MQTT broker rather
than a real secret. Suppress it with an explicit justification.
---

## Summary by CodeRabbit

* **Chores**
  * Updated linting configuration to exclude the user preferences file from automated checks.
  * Added context explaining the exception for documented example credentials.

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -102,6 +102,13 @@ lint:
     - linters: [gitleaks]
       paths:
         - test/fixtures/nodedb/seed_v25_*.jsonl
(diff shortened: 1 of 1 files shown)


vidplace7 (changes requested):
I'd rather not pollute this file with trunk comments.
Let's have trunk ignore `userPrefs.jsonc` entirely instead.

vidplace7 (approved):

PR #11176, opened 2026-07-23
lint: hoist SerialModule arch gate into HAS_SERIALMODULE

Jorropo:
The (ARCH_ESP32||ARCH_NRF52||ARCH_RP2040||ARCH_STM32WL) && !ESP32S2 &&
!ESP32C3 guard was duplicated verbatim in SerialModule.h, SerialModule.cpp
and Modules.cpp, and each copy tripped the too-many-defined ifdef-complexity
check. Define it once as HAS_SERIALMODULE in SerialModule.h (split across
nested directives to stay under the complexity limit) and reference the
macro at the three sites. Verified boolean-equivalent to the original
expression across arch/target combinations with the preprocessor.
---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved platform detection for serial module support.
  * Preserved serial functionality on supported platforms while correctly excluding unsupported ESP32 variants and other platforms.

### DIFF
--- a/src/modules/Modules.cpp
+++ b/src/modules/Modules.cpp
@@ -244,8 +244,7 @@ void setupModules()
         new PowerTelemetryModule();
     }
 #endif
-#if (defined(ARCH_ESP32) || defined(ARCH_NRF52) || defined(ARCH_RP2040) || defined(ARCH_STM32WL)) &&                             \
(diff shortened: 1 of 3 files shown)


NomDeTom:
Can you take a look at https://github.com/meshtastic/firmware/pull/11339 and this together, please?

PR #11177, opened 2026-07-23
lint: suppress ifdef-complexity on ad-hoc board lists

Jorropo:
NodeDB and EInkDisplay2 gate behaviour on explicit board-name lists (TFT
keyboard devices that default Bluetooth off; e-paper panels wired on HSPI).
Unlike the SerialModule arch gate, these are one-off enumerations with no
umbrella feature macro to fold them into, and rewriting them blind risks
silently dropping a board. Suppress the too-many-defined findings with an
explicit justification instead. Comments are kept on a single line so
clang-format's reflow does not detach them from the directive.
---

### DIFF
--- a/src/graphics/EInkDisplay2.cpp
+++ b/src/graphics/EInkDisplay2.cpp
@@ -211,6 +211,7 @@ bool EInkDisplay::connect()
         }
     }
 
+// trunk-ignore(too-many-defined/too-many-defined): explicit HSPI e-paper board list, no umbrella macro fits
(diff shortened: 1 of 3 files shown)


jp-bennett:
Honestly these are exactly the sorts of lists that lint rule was designed to target. nak

Jorropo:
@jp-bennett agreed I'm not gonna merge all of them.

thebentern:
We should just hoist HAS_TFT up to the targets that need it

PR #11180, opened 2026-07-23
Fix: ESP32 NimBLE deinit/reinit cycle breaks bonded reconnect (reason…

dl9sau:
… 531)

Two independent bugs combine to break bonded reconnects after a NimBLE deinit + reinit cycle on ESP32. Together with the SystemCommandsModule change, this also removes the "reboot required" behavior of the display- menu Bluetooth toggle on both ESP32 and nRF52.

## 1. LE Address Resolution defaults to OFF after controller reset

`NimBLEDevice::deinit(true)` tears down the NimBLE host and resets the Bluetooth controller. On re-init, the host syncs and programs the Identity Resolving Keys (IRKs) of bonded peers into the controller's resolving list, but the controller starts with LE Address Resolution DISABLED (HCI LE Set Address Resolution Enable = off, opcode 0x202D).

When a bonded phone reconnects it uses a Resolvable Private Address (RPA). Without address resolution the controller cannot map the RPA back to the peer's identity -> cannot look up the stored Long Term Key (LTK) -> encryption setup fails with BLE_ERR_PIN_OR_KEY_MISSING (disconnect reason 531) -> link drops immediately.

Fix: send `ble_hs_hci_cmd_tx(0x202D, enable=1, ...)` after `init()` to re-enable controller-side address resolution.

## 2. `ble_hs_misc_restore_irks()` races with controller readiness

`ble_hs_sync()` calls `ble_hs_misc_restore_irks()` to load IRKs from NVS, but at that point the controller may still be initialising -- the HCI commands that program the resolving list get silently dropped or queued. Re-issuing `ble_hs_misc_restore_irks()` after a short `delay(50)` ensures the IRKs actually reach the controller.

Both `ble_hs_hci_cmd_tx` and `ble_hs_misc_restore_irks` live in `nimble/host/src/ble_hs_priv.h` and are not exposed via the public `host/*.h` headers shipped with esp32-arduino. Forward-declared with `extern "C"`; the linker resolves them from the NimBLE host library. Happy to wrap them differently if the maintainers prefer.

## Symptoms observed

- iOS tends to keep the bond and retry; user sees "not connecting" but the bond survives.
- Android interprets the immediate disconnect (reason 531) as "peer lost my bond" and silently deletes the pairing from its own database. The next attempt presents as a fresh device with a new pairing dialog -- users think Bluetooth "forgot" the device.

Both are symptoms of the same root cause; the fix addresses both.

## Gating on first-boot vs post-deinit

A fresh power-on has the controller and host synchronised correctly by the initial `NimBLEDevice::init()`. Running the extra `restore_irks()`/`set_addr_res_en` there causes a boot-loop crash at ~5s. A file-static `s_didSleepSinceBoot` flag guards the fix so it only runs after a real deinit -- `deinit()` sets the flag, `setup()` checks and clears it.

## SystemCommandsModule: menu-toggle no longer reboots

The original firmware had:
> "For ESP32, no way to recover from bluetooth shutdown without reboot"

That was true because NimBLE-Arduino's `deinit(true)` + `init()` left the host in an inconsistent state (address resolution off, IRKs missing). With both bugs fixed, the full enable<->disable<->enable cycle works reliably on ESP32. The display-menu BLE toggle can now call `nimbleBluetooth->setup()` directly instead of scheduling a reboot.

nRF52 already had shutdown/reinit working, so the reboot on that branch was pure conservatism -- the analogous change (call `nrf52Bluetooth->resumeAdvertising()`) also removes the reboot on nRF52.

## Also: reset isDeInit on setup()

The `isDeInit` flag was set to `true` in `deinit()` and never reset. Without a reset, all disconnect events after a re-toggle would be silently short-circuited in the disconnect callback. Set it to `false` at the top of `setup()`.

## Motivation

Beyond the standalone reconnect-bug fix, this is a prerequisite for a periodic BLE-sleep feature I plan to contribute in a follow-up PR. My fork already runs a small state machine that puts the BLE stack through the deinit/reinit sequence periodically while no phone is connected -- significant idle-current savings on ESP32, especially valuable for handheld and battery-only nodes. That sleep cycle triggers the reconnect bug on every wake, so it only becomes a shippable feature once the deinit/reinit path here is correct. Landing this fix first keeps that follow-up as a clean opt-in feature on top of a working stack rather than a workaround stack on top of a broken one.

### ❌ (Please delete all these tips and replace them with your text) ❌

  to say "Hey, I think this idea X should be implemented and I'm starting work on it. My general plan is Y, any feedback
  is appreciated." This will allow other devs to potentially save you time by not accidentally duplicating work etc...
- We recommend using the [Visual Studio Code](https://platformio.org/install/ide?install=vscode) editor along with the ['Trunk Check' extension](https://marketplace.visualstudio.com/items?itemName=trunk.io) (In beta for windows, WSL2 for the Linux version),
- Please do not submit untested code.
- If you do not have the affected hardware to test your code changes adequately against regressions, please indicate this, so that contributors and community members can help test your changes.
- If your PR gets accepted you can request a "Contributor" role in the Meshtastic Discord

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/modules/SystemCommandsModule.cpp
+++ b/src/modules/SystemCommandsModule.cpp
@@ -58,19 +58,20 @@ int SystemCommandsModule::handleInputEvent(const InputEvent *event)
 #if defined(ARDUINO_ARCH_NRF52)
         if (!config.bluetooth.enabled) {
             disableBluetooth();
-            IF_SCREEN(screen->showSimpleBanner("Bluetooth OFF\nRebooting", 3000));
-            rebootAtMsec = millis() + DEFAULT_REBOOT_SECONDS * 2000;
+            IF_SCREEN(screen->showSimpleBanner("Bluetooth OFF", 3000));
         } else {
-            IF_SCREEN(screen->showSimpleBanner("Bluetooth ON\nRebooting", 3000));
-            rebootAtMsec = millis() + DEFAULT_REBOOT_SECONDS * 1000;
+            IF_SCREEN(screen->showSimpleBanner("Bluetooth ON", 3000));
+            if (nrf52Bluetooth)
+                nrf52Bluetooth->resumeAdvertising();
         }
 #else
         if (!config.bluetooth.enabled) {
             disableBluetooth();
             IF_SCREEN(screen->showSimpleBanner("Bluetooth OFF", 3000));
         } else {
-            IF_SCREEN(screen->showSimpleBanner("Bluetooth ON\nRebooting", 3000));
-            rebootAtMsec = millis() + DEFAULT_REBOOT_SECONDS * 1000;
+            IF_SCREEN(screen->showSimpleBanner("Bluetooth ON", 3000));
+            if (nimbleBluetooth)
+                nimbleBluetooth->setup();
         }
 #endif
         return 0;
--- a/src/nimble/NimbleBluetooth.cpp
+++ b/src/nimble/NimbleBluetooth.cpp
@@ -22,8 +22,12 @@
 
 #if defined(CONFIG_NIMBLE_CPP_IDF)
 #include "host/ble_gap.h"
+#include "host/ble_hs.h"
+#include "host/ble_store.h"
 #else
 #include "nimble/nimble/host/include/host/ble_gap.h"
+#include "nimble/nimble/host/include/host/ble_hs.h"
+#include "nimble/nimble/host/include/host/ble_store.h"
 #endif
 
 #if defined(CONFIG_IDF_TARGET_ESP32S3) || defined(CONFIG_IDF_TARGET_ESP32C6)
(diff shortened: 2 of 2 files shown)


NomDeTom:
Youve targeted your pr to master and not develop, and you've not deleted the hint text, which may mean you've not read it.

You can retarget without closing this pr, btw.

dl9sau:
> Am 23.07.2026 um 21:24 schrieb Tom ***@***.***>:
> 
> NomDeTom left a comment (meshtastic/firmware#11180)
> Youve targeted your pr to master and not develop, and you've not deleted the hint text, which may mean you've not read it.

Thanks for the pointer, I missed the branch hint ad was too lazy to not remove the textual informational part.
If I come back with this or another patch, I'll target develop. 
The fix is running fine on my device — logs show restore_irks + address resolution kick in exactly as intended after deinit cycles.

NomDeTom:
> If I come back with this or another patch, I'll target develop.

Understood.

PR #11185, opened 2026-07-24
Deliver locally-generated replies addressed to us to the phone

thebentern:
## The bug

Config get/set from a client times out on every device running current develop. The client
connects fine and downloads config at connect, but opening a config screen or changing a
setting (e.g. Device Role) times out.

Reported by an alpha tester on a Seeed Xiao nRF52840 Kit against `2.8.0.ef1aedd`, working on
`2.8.0.5ded0ec`. It is not board- or app-specific — both Android 2.8.0 and 2.7.14 behave the
same, and the mechanism is platform-independent.

## Regression

`d6b12ea3f` (#10967, packet authenticity policies) changed `Router::sendLocal`'s `isToUs`
branch:

```diff
-        enqueueReceivedMessage(p);
-        return ERRNO_OK;
+        handleReceived(p, src);
+        return ERRNO_SHOULD_RELEASE;
```

Preserving the `RxSource` is the right call — the queue round-trip used to relabel local
packets `RX_SRC_RADIO`, which is exactly the confusion the new policy gates need gone. But
module replies are sent through `MeshService::sendToMesh()` with the default `RX_SRC_LOCAL`,
and a reply to a phone-originated request is addressed to our own node (`setReplyTo` resolves
`from == 0` to `ourNodeNum`). So those replies now re-enter `callModules` as `RX_SRC_LOCAL`,
where the loopback gate skips every module with `loopbackOk == false`:

https://github.com/meshtastic/firmware/blob/develop/src/mesh/MeshModule.cpp#L121-L124

That includes `RoutingModule`, whose promiscuous sniff is the only code that moves a received
packet into `toPhoneQueue`:

https://github.com/meshtastic/firmware/blob/develop/src/modules/RoutingModule.cpp#L35-L38

The reply is released without ever being queued.

Requests still work, because the phone's own packets arrive as `RX_SRC_USER` and pass the
gate. So a `set_config` **is applied** and only its acknowledgement is lost — which is why a
client can connect and download config but times out on every config screen and every setter.
Testers who saw a timeout may find the setting actually took effect.

## The fix

Deliver the phone's copy from `sendToMesh()` for a local packet addressed to us. The loopback
gate is doing its job keeping the packet out of module re-dispatch; the phone copy is the part
that went missing.

Two alternatives I ruled out:

- `loopbackOk = true` on `RoutingModule` — would also echo every locally-generated *broadcast*
  (position, nodeinfo, telemetry) back to the phone through the promiscuous sniff.
- Relabeling module replies `RX_SRC_RADIO` in `sendLocal` — re-muddles the exact local/remote
  distinction #10967 introduced.

Also stops reporting `ERRNO_SHOULD_RELEASE` (35) to the phone in the `QueueStatus` for these
packets. It means "caller frees", not a send failure, and the same hunk changed it from the 0
the phone used to see.

## Tests

Two regression tests in `test/test_mesh_module`:

- `test_localReplyToSelf_isDeliveredToPhone` — a reply addressed to ourselves reaches the
  phone queue exactly once, with a success `QueueStatus` and no radio traffic.
- `test_phoneRequest_replyReachesPhone` — full loop through the real router: a `from == 0`
  `want_response` request produces a module reply that reaches the phone queue with the right
  `request_id`, and the request itself does not echo back.

Both fail on unpatched develop (empty phone queue) and pass with the fix.

## Validation

Native suites with the fix applied: **native-macos 577 cases, 573 passed + 4
environment-conditional ignores, 0 failed**; **Docker 751/751 passed**.

End-to-end on the portduino simulator over the TCP PhoneAPI (same client path as BLE), as a
minimal pair at the reporter's exact commit `ef1aedd56`, with only `MeshService.cpp` differing:

| | unpatched `ef1aedd56` | `ef1aedd56` + this fix |
|---|---|---|
| `--set device.role CLIENT_MUTE` | hangs >180 s (killed) | completes in 0.6 s, persists across reboot |
| `--get-ringtone` (admin round-trip) | hangs >180 s (killed) | completes in 0.4 s |
| `--get device.role` | works | works |

`--get device.role` passing on both is itself diagnostic — the Python CLI serves it from the
want_config cache rather than an admin round-trip, which is the same reason a client can
connect while every config screen times out.

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured locally addressed module replies are consistently delivered to the phone callback/queue.
  - Updated queue status forwarding so local delivery failures are no longer reported incorrectly.
  - Prevented phone-originated requests/replies from being echoed back or sent to the radio.
- **Tests**
  - Added regression tests covering local reply-to-phone delivery and phone request response routing.
  - Improved test teardown to properly drain pending phone deliveries and statuses.

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -324,13 +324,24 @@ void MeshService::sendToMesh(meshtastic_MeshPacket *p, RxSource src, bool ccToPh
     uint32_t mesh_packet_id = p->id;
     nodeDB->updateFrom(*p); // update our local DB for this packet (because phone might have sent position packets etc...)
 
+    // A locally-generated packet addressed to ourselves (e.g. a module's response to a
+    // phone-originated admin/config request) can only be consumed by local API clients:
+    // sendLocal() dispatches it with src preserved, and the loopback gate in
+    // MeshModule::callModules hides RX_SRC_LOCAL packets from RoutingModule, the phone
+    // forwarder. Deliver the phone's copy here or the client times out waiting for it.
+    const bool localDelivery = isToUs(p);
+    if (src == RX_SRC_LOCAL && localDelivery)
+        ccToPhone = true;
+
     // Note: We might return !OK if our fifo was full, at that point the only option we have is to drop it
     ErrorCode res = router->sendLocal(p, src);
 
     /* NOTE(pboldin): Prepare and send QueueStatus message to the phone as a
      * high-priority message. */
     meshtastic_QueueStatus qs = router->getQueueStatus();
-    ErrorCode r = sendQueueStatusToPhone(qs, res, mesh_packet_id);
+    // Local delivery reports ERRNO_SHOULD_RELEASE (caller frees) - that is success, not a
+    // failure the phone should surface in the QueueStatus.
+    ErrorCode r = sendQueueStatusToPhone(qs, (res == ERRNO_SHOULD_RELEASE && localDelivery) ? ERRNO_OK : res, mesh_packet_id);
     if (r != ERRNO_OK) {
         LOG_DEBUG("Can't send status to phone");
     }
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -324,11 +324,8 @@ void MeshService::sendToMesh(meshtastic_MeshPacket *p, RxSource src, bool ccToPh
     uint32_t mesh_packet_id = p->id;
     nodeDB->updateFrom(*p); // update our local DB for this packet (because phone might have sent position packets etc...)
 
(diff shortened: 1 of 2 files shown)


thebentern:
Both addressed in 6efce755d.

Comments condensed to the one-or-two-line house style — the rationale belongs in the commit message and PR body, which carry it.

The QueueStatus point was a real weakness, thanks: the `while` loop asserted `ERRNO_OK` per record but would have passed just as happily on an empty queue. It now counts and requires exactly two, one for the request and one for the reply — verified empirically rather than assumed. Re-confirmed both tests still fail on unpatched develop ("Expected Non-NULL") and pass with the fix, 21/21 for the suite.

mverch67 (approved):
Tested with ThinkNode M9 and the bug from 11094 is fixed.

PR #11191, opened 2026-07-24
Router: defer nested local sends to avoid handleReceived re-entrancy

caveman99:
Guards Router::handleReceived with a depth counter so a module sending from inside callModules defers its loopback instead of re-entering and overflowing the nRF52 task stack on a config save. Replaces draft #11155.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of locally addressed packets and broadcasts.
  * Prevented recursive packet processing when modules send packets during message handling.
  * Ensured deferred packets are processed reliably without duplicate local deliveries or crashes.
  * Added graceful handling when the deferred delivery queue reaches capacity.

* **Tests**
  * Added coverage for nested sends, chained replies, broadcasts, and queue overflow scenarios.


thebentern (approved):

caveman99:
A different approach to #11155 (credits to @Ixitxachitl for the initial analysis) and complementing the #11190 stack fix, this prevents the original recursive callModules polluting the stack in the first place.

Issue #10823, opened 2026-06-30
[Bug]: RAK4631 stops responding/shuts down using 2.7.26 after OTA and reboot. (2.7.15 is the last working version).


glith:
### Category

Hardware Compatibility

### Hardware

Rak4631

### Is this bug report about any UI (https://meshtastic.org/docs/configuration/device-uis/) component firmware?

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [ ] BaseUI

### Firmware Version

2.6.27

### Description

RAK4631 stops responding/shuts down using 2.7.26 after OTA and reboot. 
2.7.15 is the last working version. I have tried alpha 2.7.20 with the same issue both via OTA and USB.. and had to revert back to 2.7.15.

I updated my RAK4631 with OTA from 2.7.15 to 2.7.27. After the update I was still connected to it via BT.
After I sent a reboot to it, it never came back again. 

It is powered via solar and LTO batteried to the battery input with "MPPT Solar Battery Charger for IoT & Meshtastic - LTO" (https://www.etsy.com/se-en/listing/1609406536/mppt-solar-battery-charger-for-iot)

I am thinking it may be some power or shutdown feature included after 2.7.15 that makes it think it needs to shutdown. Some watchdog or whatnot. 

I cant really give more information since I cant reach it at this moment in time. :-(

### Relevant log output

```Shell

```

Deuteranomalous1:
I've been doing OTA updates with all the alpha versions since 2.7.15 and haven't experienced this issues. Are you using the updated OTAFIX bootloader or the stock bootloader? DFU update or the built in app update?

glith:
I did a "normal" OTA like I done before. Everything stock.
I dont think this problem is due to the OTA, but something has changed with battery sensing. 
I'm feeding it with a regulated 3.3v and it sent low battery notifications before I tweaked/calibrated the ADC multiplier.
Is it thinking battery is low and shutting down? ( It worked fine in 2.7.15, even after OTA to that version)

vialmarc-fr:
RAK4631 same problem too, after OTA update, impossible to reconnect by bluetooth, node is unreachable into the mesh.

thebentern:
OTA is not consistently reliable and can leave your device in a broken state. Use normal USB DFU process

glith:
I had the same bug using normal usb upgrade too using an alpha 2.7.20. 
So its not just OTA. (2.7.15 works 2.7.20 and after is broken somehow).

[...]

PR #11211, opened 2026-07-25
nRF52: add opt-in low-voltage boot protection

Musashii-kun:
Ported from the vendor's MeshCore BSP (meshcore-dev/MeshCore NRF52Board.cpp). When PWRMGT_VOLTAGE_BOOTLOCK is defined and the battery reads below it at boot (and USB power is absent), enter SYSTEM_OFF with the LPCOMP armed on the battery-sense input plus VBUS wake, so a dying cell sleeps at microamps instead of brownout-looping. Guarded entirely behind the variant define; enabled for the GAT562 with the vendor's 3300mV threshold and 5/8 VDD recovery point.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other: GAT562 30s Kit

## Summary by CodeRabbit

* **New Features**
  * Added low-battery boot protection for supported hardware.
  * Devices with critically low battery voltage now enter a low-power state instead of completing startup.
  * Recovery is enabled when battery voltage rises sufficiently or USB power is connected.
  * Configured low-voltage protection for the MeshCore trial tracker variant.

### DIFF
--- a/src/platform/nrf52/main-nrf52.cpp
+++ b/src/platform/nrf52/main-nrf52.cpp
@@ -184,6 +184,88 @@ void __attribute__((noreturn)) __assert_func(const char *file, int line, const c
     NVIC_SystemReset();
 }
 
+#if defined(PWRMGT_VOLTAGE_BOOTLOCK) && defined(BATTERY_PIN)
+/**
+ * Low-voltage boot protection, derived from MeshCore's nRF52 power management
+ * implementation (https://github.com/meshcore-dev/MeshCore,
+ * src/helpers/NRF52Board.cpp):
+ *
+ *   Copyright (c) 2025 Scott Powell / rippleradios.com
+ *   Released under the MIT License (see that project's license.txt).
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software, to deal in the Software without restriction; the
+ * above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software. This adaptation is
+ * distributed as part of Meshtastic under GPL-3.0, as permitted by the MIT
+ * license, with this notice preserved per its terms.
+ *
+ * If the battery is below
+ * PWRMGT_VOLTAGE_BOOTLOCK millivolts at boot and we are not on USB power,
+ * enter SYSTEM_OFF with the LPCOMP comparator armed on the battery-sense
+ * input, so the chip sleeps at microamps until the voltage recovers (or USB
+ * power appears) instead of brownout-looping on a dying cell.
+ *
+ * Runs before the SoftDevice is enabled, so registers are accessed directly.
+ */
+static uint16_t bootBatteryMillivolts()
+{
+    pinMode(BATTERY_PIN, INPUT);
+    analogReadResolution(12);
+    uint32_t raw = 0;
+    for (int i = 0; i < 8; i++)
+        raw += analogRead(BATTERY_PIN);
+    raw /= 8;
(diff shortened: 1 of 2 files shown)


NomDeTom:
would you mind looking at https://github.com/meshtastic/firmware/issues/10823 and my futile attempts to fix it here: https://github.com/NomDeTom/MeshtasticFirmware/tree/i-know-what-im-doing

we have a thread going in the discord: https://discord.com/channels/867578229534359593/1521972656717430794

Musashii-kun:
@NomDeTom I fetched your branch `i-know-what-im-doing` and took a look.  Basically its an experiment to completely disable power management in order to debug a hang after e.g. a firmware update or a reboot.  That branch appears to be related to an nRF52 connected to and powered by an external MPPT/LTO controller.The GAT562 30s Kit - which is what I am testing with - is a portable unit intended for handheld use.  I see in the channel that there is a proposed fix from which I infer that there is some board-level design defect causing the stated issue rather than a firmware problem.

NomDeTom:
> @NomDeTom I fetched your branch `i-know-what-im-doing` and took a look. Basically its an experiment to completely disable power management in order to debug a hang after e.g. a firmware update or a reboot. That branch appears to be related to an nRF52 connected to and powered by an external MPPT/LTO controller.The GAT562 30s Kit - which is what I am testing with - is a portable unit intended for handheld use. I see in the channel that there is a proposed fix from which I infer that there is some board-level design defect causing the stated issue rather than a firmware problem.

Ok, and what does the human think?

Musashii-kun:
@NomDeTom please give me a little bit of credit :) The previous message, on a second read, does sound a bit like AI feedback.  I've seen a lot of it recently so I may have picked up some of its linguistic habits, but rest assured that response was not machine generated.

The implication that I didn't explicitly state is that I suspect a hardware design fault rather than a software issue.  That said, I did notice on my GAT562 that it hangs on flash/reboot more often than I would expect.  Its a new piece of hardware so I had a suspicion it might be from a lower bin.  But with your context now in mind maybe there is a subtle fault in the Meshtastic code somewhere.

I don't have time this weekend to test further, but there are reports that the prior stable release worked so I may try that and see if the hangs go away.  If so then its more clearly a software issue.

NomDeTom:
Understood - I'll give more credit in future. In honesty, I read and responded to your message in bulk while children were jumping on me, so I may have skipped a few checks myself ;)

The reason I'm trying to direct you to my experimental branch and the main discussions is because I was worried you're repeating work that is already in flight, or actions that are already complete. To me the better course of action is to solve the issues with the existing functions that were put in place rather than make new ones.

Anyway, this is all hand-wavy as clod 5.0 might say. I'm around on the discord if you want to discuss it further.

Edit to add: the RAK kits are used in all sorts of setups - mobile or fixed. The main thing is that the code affects the bottom end of the battery, when you're trying to prevent bootloops and other disasters which are annoying on a personal node but fatal on a fixed and remote location.

phaseloop:
Bootloop protection already exists in PowerHAL NRF52 implementation - we read VDD voltage directly from chip so we don't rely on specific board configuration or battery setup. I don't think it would be wise to have similar features duplicated.

Currently we sleep on boot and power draw is minimal as no subsystem is yet initialized. If you want  you can add a new PowerHAL hook like sleep_until_power_recovers and put LPCOMP setup there. Or just call existing deep sleep entry as AFAIK it covers those setups already.

caveman99:
there's two ways to skin a cat, this is the one we don't use

PR #11212, opened 2026-07-25
GAT562: add measured PA gain curve so Tx power is honest

Musashii-kun:
The variant shipped with no TX_GAIN_LORA, so the firmware treated the radio setting as the actual output while the 30dBm PA added up to ~18dB on top; the stock tx_power=0 default drove the PA to a full watt. Use the vendor's measured 869MHz curve so requested dBm matches radiated dBm. The curve is truncated at the saturation knee (setting 15, 30dBm) so max-power requests use the minimum drive that reaches full output instead of pushing the PA deeper into compression.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Other: GAT562 30s Kit

## Summary by CodeRabbit

* **Improvements**
  * Added a 16-point LoRa transmission power gain configuration for improved radio output control on the GAT562 Mesh Trial Tracker variant.

### DIFF
--- a/variants/nrf52840/gat562_mesh_trial_tracker/variant.h
+++ b/variants/nrf52840/gat562_mesh_trial_tracker/variant.h
@@ -217,6 +217,15 @@ SO GPIO 39/TXEN MAY NOT BE DEFINED FOR SUCCESSFUL OPERATION OF THE SX1262 - TG
 #define SX126X_DIO2_AS_RF_SWITCH
 #define SX126X_DIO3_TCXO_VOLTAGE 1.8
 
+// Per-setting PA gain (actual output minus radio setting, dB) so requested Tx
+// power matches what the PA actually radiates. From the vendor's GAT562 30s
+// measurements at 869 MHz. The PA saturates at 30 dBm from setting 15 up, so
+// the curve is truncated there: driving harder wastes battery and linearity
+// for zero extra output, and max-power requests map to setting 15.
+// https://shop.mtoolstec.com/product/gat562-30s-kit-30dbm-mesh-device
+#define NUM_PA_POINTS 16
+#define TX_GAIN_LORA 14, 15, 16, 17, 17, 18, 18, 18, 18, 18, 17, 17, 17, 16, 16, 15
+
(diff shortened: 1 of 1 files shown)


NomDeTom:
is that curve applicable to the 915 band?
is that curve applicable to the 868 band?

Have you confirmed the power outputs yourself with e.g. power meter?

Musashii-kun:
@NomDeTom officially (according to the vendor's product page) the measurements are based on 869MHz (EU frequency):

https://shop.mtoolstec.com/gat562-30s-tx-power-settings-869mhz-output-reference.html

The vendor does note that their table is frequency-dependent: "These measurements are specific to 869 MHz. Output power may vary slightly at different frequencies (e.g., 915 MHz for US band) due to PA characteristics and filter response."

Table reproduced here for posterity:

image

Musashii-kun:
I do not personally own equipment capable of reproducing these results; I'd need access to an RF engineering lab and maybe an RF isolation chamber.  If I really wanted to I could probably get access to one but it would be a fairly significant effort and I'd prefer not to for a PR where I already have measurements directly from the vendor.

Note the knee at TX power setting 14->15.

NomDeTom:
> I do not personally own equipment capable of reproducing these results; I'd need access to an RF engineering lab and maybe an RF isolation chamber.

Where are you based? I'll settle for tests from an immersion RC meter. If you're in SF, I think noise bridge has a test setup, and in NY I know a PorkCube.

Musashii-kun:
I'm close enough to SF that I can probably make it up there, but its a bit of a haul, maybe 45 minutes by car.  I didn't know that there was such a hackerspace nearby.

NomDeTom:
There's makerspaces and hackerspaces all over. The RF meters themselves are quite inexpensive, and always uncalibrated.

That said, I wouldn't recommend using the OEM datasheet blindly for something like this - trust but verify.

PR #11213, opened 2026-07-25
GAT562 Mesh Trail Tracker: enable buzzer, RGB LED, and 5-way joystick; add opt-in nRF52 DC/DC

Musashii-kun:
## Summary

Rounds out support for the RAKwireless GAT562 Mesh Trial Tracker (30s Kit) by
enabling on-board hardware that was previously unwired, plus an opt-in nRF52
power optimization that the board is built for. All pins were verified against
the vendor's own MeshCore variant for this board
(`meshcore-dev/MeshCore` → `variants/gat562_30s_mesh_kit`).

## Changes

**Buzzer and WS2812 RGB LED** (`variants/nrf52840/gat562_mesh_trial_tracker/variant.h`)
- `PIN_BUZZER 33` - the pre-existing commented-out `21` was wrong for this board.
- `HAS_NEOPIXEL` with data on pin 29, single GRB/800kHz pixel.
- Defining `PIN_BUZZER` also turns on external notification with buzzer alerts
  by default, which pairs with the userPrefs ringtone.

**5-way joystick → input broker**
- Mapped through the existing trackball driver (`TB_UP/DOWN/LEFT/RIGHT/PRESS`,
  `TB_DIRECTION FALLING`) so all five directions emit input events.
- Makes canned messages, UI navigation, and the on-screen keyboard usable
  directly on the device, with pins baked into the firmware rather than
  requiring manual app-side configuration.
- `TB_THRESHOLD` is deliberately left undefined — the pulse accumulator is for
  rollers, not clicky switches.

**Opt-in DC/DC converter** (`src/platform/nrf52/main-nrf52.cpp`,
`src/platform/nrf52/NRF52Bluetooth.cpp`)
- New `NRF52_USE_DCDC` guard, enabled for GAT562, which has the DC/DC
  inductors fitted. Running the REG1 buck converter instead of the LDO cuts
  active and radio current.
- Enabled first thing in `nrf52Setup()` via a SoftDevice-aware helper: direct
  `NRF_POWER->DCDCEN` write pre-SoftDevice, `sd_power_dcdc_mode_set()` once the
  SoftDevice owns the POWER peripheral — the same stage and pattern as the
  vendor's `NRF52BoardDCDC::begin()`.
- Early enablement means the whole boot (GPS, screen, I2C scan, radio init)
  runs on the buck converter and Bluetooth-disabled builds are covered;
  `NRF52Bluetooth::setup()` re-asserts the mode (idempotent) under the
  SoftDevice.
- REG0 (`DCDCEN0`) is deliberately left on its LDO, matching the vendor.
- No behavior change for any board that does not define `NRF52_USE_DCDC`.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other: GAT562 30s Kit

## Summary by CodeRabbit

* **New Features**
  * Added support for the board’s five-way trackball input.
  * Configured the populated push-button for cancel actions.
  * Enabled buzzer and single NeoPixel hardware configuration.
  * Added nRF52 DC/DC regulator support for improved power efficiency.
  * Added regulator status reporting during startup and Bluetooth initialization.

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -120,15 +120,6 @@ void nrf52Setup(), esp32Setup(), nrf52Loop(), esp32Loop(), rp2040Setup(), rp2040
 void esp32ReleaseBluetoothMemoryIfUnused();
 #endif
 
-#ifdef ARCH_NRF52
-// DC/DC buck converter control. nrf52EnableDCDC() runs from powerHAL_platformInit(),
-// which is before consoleInit(), so the outcome is latched and reported later by
-// nrf52LogDCDCStatus(). nrf52ReassertDCDC() re-applies the mode once the SoftDevice
-// owns the POWER peripheral. All three are no-ops unless the board sets
-// NRF52_USE_DCDC_REG0 and/or NRF52_USE_DCDC_REG1.
-void nrf52EnableDCDC(), nrf52LogDCDCStatus(), nrf52ReassertDCDC();
-#endif
-
 meshtastic_DeviceMetadata getDeviceMetadata();
 #if !MESHTASTIC_EXCLUDE_I2C
 void scannerToSensorsMap(const std::unique_ptr<ScanI2CTwoWire> &i2cScanner, ScanI2C::DeviceType deviceType,
--- a/src/main.h
+++ b/src/main.h
@@ -120,6 +120,15 @@ void nrf52Setup(), esp32Setup(), nrf52Loop(), esp32Loop(), rp2040Setup(), rp2040
 void esp32ReleaseBluetoothMemoryIfUnused();
 #endif
 
+#ifdef ARCH_NRF52
+// DC/DC buck converter control. nrf52EnableDCDC() runs from powerHAL_platformInit(),
+// which is before consoleInit(), so the outcome is latched and reported later by
+// nrf52LogDCDCStatus(). nrf52ReassertDCDC() re-applies the mode once the SoftDevice
+// owns the POWER peripheral. All three are no-ops unless the board sets
+// NRF52_USE_DCDC_REG0 and/or NRF52_USE_DCDC_REG1.
+void nrf52EnableDCDC(), nrf52LogDCDCStatus(), nrf52ReassertDCDC();
+#endif
+
 meshtastic_DeviceMetadata getDeviceMetadata();
 #if !MESHTASTIC_EXCLUDE_I2C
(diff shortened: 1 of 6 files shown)


NomDeTom:
@phaseloop

phaseloop:
The best place to init DCDC is powerhal_init function which is executed as earliest as possible during boot and is designed to hold stuff like that. At that point SD is not initialized so it requires a registry setup.

AFAIK it would make sense to have separate flags for REG0 and REG1 - many boards support DCDC on both regulators.

src/main.h
```diff
@@ -120,6 +120,15 @@ void nrf52Setup(), esp32Setup(), nrf52Loop(), esp32Loop(), rp2040Setup(), rp2040
 void esp32ReleaseBluetoothMemoryIfUnused();
 #endif
 
+#ifdef ARCH_NRF52
+// DC/DC buck converter control. nrf52EnableDCDC() runs from powerHAL_platformInit(),
+// which is before consoleInit(), so the outcome is latched and reported later by
+// nrf52LogDCDCStatus(). nrf52ReassertDCDC() re-applies the mode once the SoftDevice
+// owns the POWER peripheral. All three are no-ops unless the board sets
+// NRF52_USE_DCDC_REG0 and/or NRF52_USE_DCDC_REG1.
+void nrf52EnableDCDC(), nrf52LogDCDCStatus(), nrf52ReassertDCDC();
```
phaseloop: Why define it here while those methods are only used internally in NRF52 platform?

I understand this needs to be shared between nrf52-main and bluetooth code files. It will be better to just create a new internal header file like nrf52-main.h and eventually include it in bluetooth part file.

phaseloop (commented):
looks good - left one comment as function definitions should not "leak" directly into main.h as our long term goal is to hide anything platform specific only inside platform directory and not use tons of #ifdefs

Musashii-kun:
@phaseloop moved as requested - this makes sense.

Musashii-kun:
I switched the user-button (marked S1, on the right of the PCB face) to perform BACK.

Any further feedback from the maintainers?  @whywilson - would appreciate your feedback here as well because you're the hardware manufacturer / dev.

PR #11215, opened 2026-07-25
Lr1121 tcxo optional tries xtal first, and get all my yamls in a row

NomDeTom:
On a board where a TCXO may or may not be fitted (DIY / socketed LR1121 modules), configuring a
DIO3 TCXO reference that isn't physically present makes RadioLib hang **forever** during
`LR11x0::config()`. The oscillator never locks, BUSY stays asserted, and the calibration wait is
an unbounded `while(digitalRead(busy))`. It happens in `setup()` before any watchdog is running,
so the board needs a physical reset. This defeats the existing `TCXO_OPTIONAL` fallback, which
can only work if `begin()` returns.

Upstream bug: <https://github.com/jgromes/RadioLib/issues/1844>.

This PR bounds tries xtal first, then tcxo, which works without any other fix.

## More stuff
- Added WIO LR1121 pins to promicro with build flag
- Corrected LR1121 yamls for femtofox

## Testing

Built for `nrf52_promicro_diy_tcxo`. Hang reproduced and traced on a bare (no-TCXO) GNiceRF
LR1121 on an nRF52840 Pro Micro; behaviour confirmed independent of the LR1121's own on-die
firmware version (retested after updating `0x0101` → `0x0104`).

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

nRF52840 Pro Micro DIY (`nrf52_promicro_diy_tcxo`) with LR1121, both bare/no-TCXO and
TCXO-equipped modules.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LoRa radio bring-up for optional XTAL/TCXO setups with clearer attempt logging and safer fallback behavior.
  * Added a BUSY-signal watchdog to prevent potential hangs when BUSY stays asserted.
  * Receive now starts only after successful initialization (including RX gain setup).
  * Enhanced diagnostics for oscillator/DIO3 IRQ usability.

* **Documentation**
  * Updated FEMTOFOX LR1121 configuration naming/formatting, module labeling, and RF switch tables for E80 and Wio-LR1121.
  * Refreshed nRF52840 DIY Promicro TCXO docs with correct build-flag guidance and Wio-LR1121 truth table.

### DIFF
--- a/bin/config.d/femtofox/femtofox_E80-900M2213S.yaml
+++ b/bin/config.d/femtofox/femtofox_E80-900M2213S.yaml
@@ -1,30 +1,31 @@
 ---
 Lora:
-## Ebyte E80-900M22S
-## This is a bit experimental
-## 
-##
+  ## Ebyte E80-900M22S
+  ## This is a bit experimental
+  ##
+  ##
   Module: lr1121
   gpiochip: 1 # subtract 32 from the gpio numbers
   DIO3_TCXO_VOLTAGE: 1.8
   CS: 16 #pin6 / GPIO48 1C0
-  IRQ: 23  #pin17 / GPIO55 1C7
+  IRQ: 23 #pin17 / GPIO55 1C7
   Busy: 22 #pin16 / GPIO54 1C6
   Reset: 25 #pin13 / GPIO57 1D1
 
-
   spidev: spidev0.0 #pins are (CS=16, CLK=17, MOSI=18, MISO=19)
   spiSpeed: 2000000
-  
-rfswitch_table:
-  pins: [DIO5, DIO6, DIO7]
-  MODE_STBY: [LOW, LOW, LOW]
-  MODE_RX: [LOW, HIGH, LOW]
-  MODE_TX: [HIGH, HIGH, LOW]
-  MODE_TX_HP: [HIGH, LOW, LOW]
-  MODE_TX_HF: [LOW, LOW, LOW]
-  MODE_GNSS: [LOW, LOW, HIGH]
-  MODE_WIFI: [LOW, LOW, LOW]
+
+  # CDEbyte do not follow the Semtech reference switching: DIO5/DIO6 are swapped.
+  # DIO7 is wired to the GNSS path. See variants/nrf52840/diy/nrf52_promicro_diy_tcxo/readme.md
+  rfswitch_table:
+    pins: [DIO5, DIO6, DIO7]
+    MODE_STBY: [LOW, LOW, LOW]
+    MODE_RX: [LOW, HIGH, LOW]
+    MODE_TX: [HIGH, HIGH, LOW]
+    MODE_TX_HP: [HIGH, LOW, LOW]
+    MODE_TX_HF: [LOW, LOW, LOW]
+    MODE_GNSS: [LOW, LOW, HIGH]
+    MODE_WIFI: [LOW, LOW, LOW]
 
 General:
   MACAddressSource: eth0
--- a/bin/config.d/femtofox/femtofox_E80-900M2213S.yaml
+++ b/bin/config.d/femtofox/femtofox_E80-900M2213S.yaml
@@ -1,6 +1,6 @@
 ---
 Lora:
-  ## Ebyte E80-900M22S
+  ## Ebyte E80-900M2213S
   ## This is a bit experimental
   ##
   ##
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -95,23 +95,50 @@ template <typename T> bool LR11x0Interface<T>::init()
     // Allow extra time for TCXO to stabilize after power-on
     delay(10);
 
-    int res = lora.begin(getFreq(), bw, sf, cr, syncWord, power, preambleLength, tcxoVoltage);
+#if defined(TCXO_OPTIONAL)
+    // Try XTAL first, TCXO as fallback - the reverse of the naive order. Measured on hardware
+    // (see .notes/hardware/lr1121-tcxo-hang/ and the RadioLib bug report referencing this):
+    // on a bare/non-TCXO module, a TCXO-first attempt either hangs RadioLib's calibration wait
+    // forever (unpatched upstream) or, even once that's fixed with a timeout, costs a full ~1.3s
+    // failed attempt before falling back - whereas XTAL succeeds immediately (~350ms). On a
+    // genuine TCXO module, XTAL fails fast and cleanly (~300ms, RADIOLIB_ERR_SPI_CMD_FAILED),
+    // so falling back to TCXO costs only that ~300ms. XTAL-first is a strict improvement for
+    // hang-avoidance either way, and sidesteps the RadioLib bug entirely rather than depending
+    // on a timeout fix landing upstream.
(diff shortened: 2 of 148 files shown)


NomDeTom:
I will test this properly tonight, but it worked in a brief check

src/mesh/LR11x0Interface.cpp
```diff
@@ -54,22 +53,28 @@ template <typename T> bool LR11x0Interface<T>::init()
...
     float tcxoVoltage = (float)portduino_config.dio3_tcxo_voltage / 1000;
 // FIXME: correct logic to default to not using TCXO if no voltage is specified for LR11x0_DIO3_TCXO_VOLTAGE
 #elif defined(LR11X0_DIO3_TCXO_VOLTAGE)
     float tcxoVoltage = LR11X0_DIO3_TCXO_VOLTAGE;
-    LOG_DEBUG("LR11X0_DIO3_TCXO_VOLTAGE defined, using DIO3 as TCXO reference voltage at %f V", LR11X0_DIO3_TCXO_VOLTAGE);
-    // (DIO3 is not free to be used as an IRQ)
 #elif defined(TCXO_OPTIONAL)
-    float tcxoVoltage = 1.6f; // TCXO_OPTIONAL: try default 1.6 V first, fall back to XTAL on failure
-    LOG_DEBUG("TCXO_OPTIONAL: no LR11X0_DIO3_TCXO_VOLTAGE defined, trying default TCXO Vref 1.6 V first");
+    float tcxoVoltage = 1.6f;
```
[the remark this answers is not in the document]
NomDeTom: addressed

bin/config.d/femtofox/femtofox_E80-900M2213S.yaml
```diff
@@ -1,30 +1,31 @@
 ---
 Lora:
-## Ebyte E80-900M22S
-## This is a bit experimental
-## 
-##
+  ## Ebyte E80-900M22S
```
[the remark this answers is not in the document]
NomDeTom: 13

src/mesh/LR11x0Interface.cpp
```diff
@@ -95,47 +121,67 @@ template <typename T> bool LR11x0Interface<T>::init()
...
+        delay(100);
+        radioLibResult = tryBegin(3, attemptVoltage);
+    }
+
+    lockingHal->disarmBusyWatchdog();
+
     // \todo Display actual typename of the adapter, not just `LR11x0`
-    LOG_INFO("LR11x0 init result %d", res);
-    if (res == RADIOLIB_ERR_CHIP_NOT_FOUND || res == RADIOLIB_ERR_SPI_CMD_FAILED)
+    LOG_INFO("LR11x0 init result %d", radioLibResult);
```
caveman99: Please don't change established code just for the heck of it. I am solely talking about the res rename.

src/mesh/RadioLibInterface.cpp
```diff
@@ -40,6 +40,36 @@ void LockingArduinoHal::spiTransfer(uint8_t *out, size_t len, uint8_t *in)
...
 
+void LockingArduinoHal::armBusyWatchdog(uint32_t pin, uint32_t timeoutMs)
+{
+    busyWatchPin = pin;
+    busyWatchTimeoutMs = timeoutMs;
+    busyHighSince = 0;
+    busyWatchArmed = true;
+}
+
+uint32_t LockingArduinoHal::digitalRead(uint32_t pin)
```
caveman99: i don't think this is neccesary when you probe the other way around. Init on an TCXO-LR2021 without TCXO voltage clean -707's. Probe for TCXO 0 first, then TCXO voltage.

caveman99 (changes requested):

src/mesh/LR11x0Interface.cpp
NomDeTom: It wasn't supposed to change all the `res`, just the extra one it made that duplicated but didn't shadow that one. I'm currently about 9 layers deep..... 🕳️

src/mesh/RadioLibInterface.cpp
NomDeTom: I did that first, but this way feels better. I'll go back to xtal-then-tcxo 🙃

NomDeTom: Done

NomDeTom:
@caveman99 are we confident that a TCXO LR1121 started with DIO3=0V will definitely fail rather than starting in a degraded mode?

caveman99:
> @caveman99 are we confident that a TCXO LR1121 started with DIO3=0V will definitely fail rather than starting in a degraded mode?

I was testing with LR2021 a lot and if the previous generation is anything like it (either runs on xtal or tcxo) the startup in the wrong mode will raise a -707 error

NomDeTom:
> I was testing with LR2021 a lot and if the previous generation is anything like it (either runs on xtal or tcxo) the startup in the wrong mode will raise a -707 error

In this case, it does not / is no longer doing that, and instead is going into a perma-busy state. LR2021 with XTAL fails correctly, but LR1121 does not for some reason.

NomDeTom:
@caveman99 are you happy for this one to go when the checks are finished?

NomDeTom:
@vidplace7 can you take a look at this? Blast radius is low, but I'd prefer another pair of eyes before I merge t.

joshbowyer:
My weigh-in as I have been somewhat brought in to help with a HAT design for the Luckfox Lyra Zero and an LR1121 radio:

**Core fix (`LR11x0Interface.cpp`)**: `lr11x0SpiFailed()` cleanly unifies `SPI_CMD_FAILED`/`SPI_CMD_TIMEOUT` handling, which matches the earlier BUSY-watchdog iteration in the commit history correctly surfacing either error depending on where RadioLib gave up. The attempt ordering (XTAL first when `TCXO_OPTIONAL`, TCXO fallback only if the chip actually answered and refused XTAL, then one settling retry) directly matches the hang described in the PR body and jgromes/RadioLib#1844.

**Low blast radius is structurally true, not just asserted**: when `TCXO_OPTIONAL` isn't defined, `attemptVoltage` is just the pre-existing configured `tcxoVoltage`, so every already-shipped board (Heltec, T-Deck, T-Beam, RAK, T-1000E - the unchecked regression boxes) hits the exact same single-attempt path as before, just wrapped in the new `tryBegin` lambda for logging. That's a real reason to be comfortable merging without re-testing every listed board, since the change is a no-op for that code path.

**One open question**: the femtofox YAML changes correctly move `rfswitch_table` to be nested under `Lora:` instead of a top-level sibling key. If that was actually being silently ignored at runtime before (rather than just a style/indentation preference), it'd be worth knowing whether existing femtofox LR1121 users were running with no RF switch control this whole time.

Otherwise this looks mergeable to me.

---
*This comment was drafted with the assistance of an LLM, but manually posted by myself*

NomDeTom:
> it'd be worth knowing whether existing femtofox LR1121 users were running with no RF switch control this whole time.

I'm almost certain noone has had it working with an lr radio

vidplace7 (approved):

caveman99 (approved):

Issue #53, opened 2020-03-21
sometimes we fail to complete lora message tx


spattinson:
version 0.1.8 -It was happening on previous version too, this time about ten minutes after flashing. I dont know if its always the same node. I will check this time it was my node ending 44.
```
canSleep, mode=3, isRx=0, txEmpty=0, txGood=4
canSleep, mode=3, isRx=0, txEmpty=0, txGood=4
canSleep, mode=3, isRx=0, txEmpty=0, txGood=4
Trigger powerFSM 9
Transition powerFSM transition=Contact from phone, from=DARK to=DARK
pbread from ea9f3f82-8dc4-4733-9452-1f6da28892a2 returns 30 bytes
canSleep, mode=3, isRx=0, txEmpty=0, txGood=4
canSleep, mode=3, isRx=0, txEmpty=0, txGood=4
canSleep, mode=3, isRx=0, txEmpty=0, txGood=4
canSleep, mode=3, isRx=0, txEmpty=0, txGood=4
canSleep, mode=3, isRx=0, txEmpty=0, txGood=4
canSleep, mode=3, isRx=0, txEmpty=0, txGood=4
```
previously I didn't see the message in the middle, its usually just the other lines scrolling very quickly. Cannot get screen to turn on or anything happening by pressing the button. Blue led is off, have to press RST button then it works again. I am tending to leave my units on all they time, this problem usually arise some minutes after boot. I will see if I can capture the preceeding message by running the terminal through tee, or make it log to file.

geeksville:
oh that is super interesting.  I'll write more on this bug later tonight (just got back from sickness)

geeksville:
ok, checked the code, this means we have a bug (likely a race condition) where the lora radio is staying stuck in transmit mode (rather than returning to rxmode after sending).  I'm now looking to try to find the cause and something to check for it and report the bug earlier.

geeksville:
@spattinson I'm releasing a build later today 0.1.9 that I think probably fixes this (though I'm having a hard time reproducing it).  I fixed one possible root cause, but also added a watchdog to see if this bug occurs, if it does it 'fixes' it and reports back to the mothership (eventually).  

Can you check with 0.1.9 if it prints "ERROR! Bug! Tx packet took too long to send, forcing radio into rx mode"? that will tell me if the root bug still exists?  (though it will also apply a workaround to make the bug not fatal)

spattinson:
[...]

Issue #52, opened 2020-03-20
Bluetooth Won't Pair


lburrowes:
I tried .7 and .8 alpha release and the TTGO V1 will not pair with the phone or tablets.  Shows the pin code on he screen.

girtsf:
I can reproduce with the T-BEAM v1.0. This is what I see in the debug log:

```
Turning on screen
showing bluetooth screen
onAuthenticationComplete -> fail 81
showing standard frames
```

I'll poke at it and see if I can figure out what that means.

lburrowes:
Other bluetooth based code I have loaded on the board works fine.  Only meshtastic has an issue with pairing...

PR #11217, opened 2026-07-25
CI: defuse pioarduino's 60s esptool install on esp32 builds

thebentern:
## Problem

The [2026-07-25 nightly](https://github.com/meshtastic/firmware/actions/runs/30149645315) lost **23 of 139 build legs**, all esp32/esp32s3, all with a byte-identical error — and because `build` failed, `gather-artifacts` and `publish-nightly` were skipped, so **no nightly was published**.

```
Error in package configuration: TimeoutExpired: Command '['/pio/core/penv/bin/uv', 'pip',
 'install', '--quiet', '--force-reinstall', '--python=/pio/core/penv/bin/python', '-e',
 '/pio/core/packages/tool-esptoolpy']' timed out after 60 seconds
TypeError: cannot unpack non-iterable NoneType object:
  File "/pio/core/platforms/espressif32/builder/main.py", line 52:
    PYTHON_EXE, esptool_binary_path = platform.setup_python_env(env)
```

This is not new — the same failure hit on 07-18, 07-21, and the 07-24 nightly (2 legs). It is the reason `UV_OFFLINE=1` and `bin/patch-esp32_uv_reinstall.sh` were added to gh-action-firmware; both were then removed by [gh-action-firmware#50](https://github.com/meshtastic/gh-action-firmware/pull/50), whose branch was cut 2h39m before the patch it deleted.

## Root cause

espressif32 `55.03.39` (pinned at `variants/esp32/esp32-common.ini:8`) guards its esptool install with a probe (`builder/penv_setup.py:785-804`) that passes only when `import esptool` resolves from inside the `tool-esptoolpy` package dir.

**Nothing satisfies that probe on a first pass** — the editable install is never performed at container-build time (during `pio pkg install` the function early-returns at `penv_setup.py:774-775`), and `python_deps` carries no esptool. So *every* esp32 build falls through to:

```python
uv pip install --quiet --force-reinstall --python=<penv> -e <tool-esptoolpy>   # timeout=60
```

`--force-reinstall` is a hidden alias of uv's `--reinstall`, which implies `--refresh` — so this re-resolves esptool plus its dependencies against PyPI instead of being cache-served. It's a latency coin-flip against a 60s budget.

`TimeoutExpired` is not caught at `penv_setup.py:817` (only `CalledProcessError` is), so `_setup_python_environment_core` never reaches its `return penv_python, esptool_binary_path`; `platform.py` swallows the exception; `setup_python_env` returns `None` implicitly. **The `TypeError` is a downstream symptom, not the fault.**

Load is the trigger, not the cause — 15 of the 38 first-wave esp32 legs survived the same race, and all 37 later-starting esp32 legs passed.

## Fix

`bin/ci/esp32-esptool-preflight.sh`, called from `bin/build-esp32.sh` between `pkg install` (line 15) and `pio run` (line 25):

- **Part A** — strip `--force-reinstall` and raise `timeout=60` → `600` in the vendored `penv_setup.py`, so the install is cache-served and residual slowness degrades to a slow success rather than a failed build.
- **Part B** — best-effort: uninstall the shadowing regular esptool and install `tool-esptoolpy` editable, so the probe MATCHes and the install is skipped outright.

**Deliberately fail-open.** Part A alone removes the hard failure, so an unrecognized upstream layout warns and continues rather than reddening 79 builds. `exit 0` on every path.

### Why here, and not in the container

gh-action-firmware's `entrypoint.sh` runs `/workspace/bin/build-esp32.sh` and its `action.yml` bind-mounts `$GITHUB_WORKSPACE:/workspace` — **this repo owns the script that executes inside the container.** So this takes effect on the very next job, with no image rebuild, and works against any image vintage.

That matters because a container-side fix is *not* sufficient today: `action.yml` uses a bare `docker run` (Docker default `--pull=missing`, no digest pin), and no failing job log shows a pull — the 06:53Z image republish provably never reached the runners before the 07:37Z nightly.

## Verification

Placement verified: all 23 failures fire during `pio run`, never during `pkg install`, and the traceback itself names `/pio/core/platforms/espressif32/builder/penv_setup.py`, so the target file exists by then. No `Platform Manager:` line appears between the two steps, so nothing re-extracts the platform and reverts the edit.

The seds were dry-run against the pinned 55.03.39 source (md5 `9448800d6ddc186e7966a7d2cae1634f`) — **exactly 4 hunks** (585, 588, 811, 814), correctly leaving the unrelated `timeout=60` at lines 414 and 887 untouched, zero `--force-reinstall` residue.

Both parts were exercised end-to-end against a synthetic core with a properly pip-installed shadowing esptool:

```
run 1: preflight: patched .../penv_setup.py
       preflight: esptool baked editable from .../tool-esptoolpy
       import esptool -> .../tool-esptoolpy/esptool  (5.3.0, was 4.9.0 shadow)
run 2: preflight: esptool already resolves under .../tool-esptoolpy (probe will MATCH)
```

Idempotent, and the fail-open path was confirmed by an unbakeable variant (warns, exits 0).

### How to confirm after merge

Grep a build log for `preflight: ` plus either `esptool already resolves` or `esptool baked editable`, and assert the **absence** of `TimeoutExpired ... tool-esptoolpy`. A green run alone proves little — the failure is a race that 15 of 38 legs won unaided.

## Follow-ups (not in this PR)

- gh-action-firmware: [#52](https://github.com/meshtastic/gh-action-firmware/pull/52) fixes this properly at image-build time, but is currently a **no-op for esp32** because [#53](https://github.com/meshtastic/gh-action-firmware/pull/53) stops copying `/pio/core/penv` into esp32 images. Separate PR follows.
- `esp32c6 × master` has been hard-404ing on a deleted release asset, which makes every `Build Release` run report `failure` and masks real image regressions.
- No `timeout-minutes` or retry exists anywhere in the firmware build path.

/cc @vidplace7

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ESP32 build reliability by preventing unnecessary network-dependent reinstalls during the build process.
  * Added a CI preflight step to verify and stabilize the required esptool installation.
  * Builds now proceed successfully (with warnings) even if the preflight adjustment or validation can’t be completed.

### DIFF
--- a/bin/build-esp32.sh
+++ b/bin/build-esp32.sh
@@ -14,6 +14,12 @@ rm -r $OUTDIR/* || true
 # Important to pull latest version of libs into all device flavors, otherwise some devices might be stale
 platformio pkg install -e $1
 
+# Defuse pioarduino's esptool install, which otherwise runs a network-bound
+# `uv pip install --force-reinstall` under a hard-coded 60s timeout on every esp32
+# build and fails the build when it loses that race. Must run after `pkg install`
+# (which extracts the platform) and before `pio run` (which triggers penv setup).
+bin/ci/esp32-esptool-preflight.sh
+
 echo "Building for $1 with $PLATFORMIO_BUILD_FLAGS"
 rm -f $BUILDDIR/firmware*
 
--- a/bin/ci/esp32-esptool-preflight.sh
+++ b/bin/ci/esp32-esptool-preflight.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+#
+# Defuse pioarduino's esptool install on esp32 builds.
+#
+# The espressif32 platform (pinned to 55.03.39 in variants/esp32/esp32-common.ini)
+# guards its esptool install with a probe that only passes when `import esptool`
+# resolves from inside the tool-esptoolpy package dir. Nothing ever satisfies that
+# probe on a first pass -- the editable install is never performed at container-build
+# time, and python_deps carries no esptool -- so every esp32 build falls through to:
+#
+#   uv pip install --quiet --force-reinstall --python=<penv> -e <tool-esptoolpy>
+#
+# under a hard-coded timeout=60 (builder/penv_setup.py:811-814). `--force-reinstall`
+# is an alias of uv's `--reinstall`, which implies `--refresh`, so that call re-resolves
+# esptool and its dependencies against PyPI instead of being served from cache. It is a
+# latency coin-flip against a 60s budget, and it loses whenever the runners are busy.
+#
+# When it loses, subprocess.TimeoutExpired is not caught (only CalledProcessError is,
+# at penv_setup.py:817), so _setup_python_environment_core never reaches its
+# `return penv_python, esptool_binary_path`. platform.py swallows the exception, and
+# setup_python_env then returns None implicitly, which surfaces as the real-world
+# symptom:
+#
+#   TypeError: cannot unpack non-iterable NoneType object   (builder/main.py:52)
+#
+# This script runs inside the build container -- gh-action-firmware bind-mounts the
+# workspace and its entrypoint invokes bin/build-esp32.sh -- so it takes effect on the
+# next job, against any container image vintage.
+#
+# It is deliberately fail-open. Part A alone removes the hard failure, so nothing here
+# is worth reddening a build over: an unrecognized upstream layout warns and continues.
+
+set -uo pipefail
+
+core="${PLATFORMIO_CORE_DIR:-$HOME/.platformio}"
+penv="$core/penv"
+pkg="$core/packages/tool-esptoolpy"
+
+# --- Part A: de-fang the install command itself -------------------------------------
+#
+# Drop `--force-reinstall` so the install is cache-served rather than re-resolved, and
+# raise the 60s cap to 600s so residual slowness becomes a slow success instead of a
+# build failure. Both `--force-reinstall` call sites are patched: only the one at line
+# 811 is reachable from the minimal (env is None) path this build takes, but the file is
+# vendored third-party code and a global edit is cheaper to reason about than a targeted
+# one. The `], timeout=60, env=uv_env)` anchor is specific enough to leave the unrelated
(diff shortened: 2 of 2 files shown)


thebentern:
Updated for #11219. `develop` now pins the [meshtastic pioarduino fork](https://github.com/meshtastic/pioarduino-platform-espressif32), whose `cfdf56e` replaced `--force-reinstall` with `--reinstall-package esptool` — so the flag-strip half of this patch had silently stopped matching (the timeout half still applied).

That change narrowed *what* gets reinstalled but not the network round trip: `--reinstall-package` implies uv's `--refresh-package`, and `timeout=60` was left in place. `develop` is still losing legs to it — 5 in [run 30173773316](https://github.com/meshtastic/firmware/actions/runs/30173773316), same `TimeoutExpired` → `TypeError` shape as before:

```
TimeoutExpired: Command '[... 'uv', 'pip', 'install', '--quiet',
 '--reinstall-package', 'esptool', ... '-e', '.../tool-esptoolpy']'
 timed out after 60 seconds
```

This now handles both spellings, and the warning only fires when a reinstall flag **and** the 60s cap both survive — so it goes quiet rather than noisy once the real fix lands.

Verified against all three platform vintages:

| vintage | reinstall flags left | `timeout=60` left | warning | py_compile |
| --- | --- | --- | --- | --- |
| upstream 55.03.39 (`--force-reinstall`) | 0 | 0 | – | OK |
| current fork (`--reinstall-package`) | 0 | 0 | – | OK |
| fork + meshtastic/pioarduino-platform-espressif32#1 | n/a | n/a | none | OK |

**The proper fix is meshtastic/pioarduino-platform-espressif32#1** — it raises the cap to 600s at both call sites and, more importantly, catches `subprocess.TimeoutExpired` so it can no longer escape and produce that misleading `NoneType` unpack three frames away in `builder/main.py`. Once that merges this PR becomes a harmless no-op (verified above), and it can be dropped if you'd rather not carry a patch-the-vendored-source step. I'd suggest keeping it until the fork fix has a nightly's worth of evidence behind it.

PR #11222, opened 2026-07-26
fix: defer direct messages while learning peer keys

RCGV1:
## Summary
- Hold a DM briefly while requesting NodeInfo when the sender needs a recipient key or the recipient may need the sender key.
- Concurrent DMs to the same peer share one directed NodeInfo request and resume with their original client message IDs.
- Replay `KEY_EXCHANGE` for a duplicate client submission while the original message is still pending, without a second RF transmission.
- Avoid a pointless key-exchange wait for licensed recipients, which cannot use PKI.
- Do not replace a previously signed identity with an unsigned NodeInfo update, even when its stored key is malformed.
- Allow a valid client-supplied recipient key to encrypt an explicit PKI DM after the radio has evicted that contact, without persisting or replacing NodeDB state. A conflicting stored key still fails closed.
- Keep `QueueStatus.state = KEY_EXCHANGE` associated with the original message ID for every deferred state, including a re-deferred retry.
- Emit the resumed send result against the original message ID.

## Client behavior
`QueueStatus.state = KEY_EXCHANGE` means the radio is waiting for a short NodeInfo/key-refresh exchange, not that the DM was delivered. Clients should retain the existing pending row for `mesh_packet_id` and update its state rather than create a second message.

## Dependency
Depends on [meshtastic/protobufs#1020](https://github.com/meshtastic/protobufs/pull/1020), which adds the additive QueueStatus lifecycle field.

## Testing
- `native-macos` `test_packet_signing`: 86/86 passed
- `native-macos` `test_mesh_module`: 25/25 passed
- `native-macos` `test_stream_api`: 13/13 passed
- `trunk fmt` and `git diff --check` passed
- T-Beam S3 Core flashed from this worktree and returned healthy serial metadata; firmware API accepted isolated test messages

## Hardware scope
The attached nRF test nodes are USB-visible but currently do not complete app-protocol discovery or serial DFU handshakes. This PR does not claim a two-node PKI DM delivery test on the physical bench.

## Summary by CodeRabbit

* **New Features**
  * Direct messages can now wait for required public keys, automatically request key information, and retry delivery.
  * Queue-status updates now indicate when a message is undergoing key exchange, including duplicate phone submissions.
  * Node information requests can be sent on demand to support message recovery.

* **Bug Fixes**
  * Improved handling of missing, invalid, or empty encryption keys.
  * More accurate delivery errors and status reporting for encrypted messages.

* **Tests**
  * Expanded coverage for deferred messages, key exchange, retries, and queue-status behavior.


RCGV1:
@coderabbitai review

RCGV1:
Merged current `develop` and completed the remaining review follow-up in `9706d404f`.

- Incoming 32-byte all-zero public keys are normalized as absent before persistence, while the rest of the NodeInfo still updates. A focused regression proves the peer name lands but `copyPublicKey()` remains false.
- Deferred-DM test shims now restore `nodeInfoModule`, `airTime`, and Portduino simradio state after every test.
- Deferred queue and key-exchange table constants now follow repository naming and use named bounds.
- The undecryptable-PKI response path now reads the sender key through `NodeDB::copyPublicKey()` and logs the reason as hex.
- The three inline findings remain covered by the existing M24/M25 and ownership-safe assertions.

Validation on the merged branch: `trunk fmt` and `trunk check` on the five touched files (no new issues), `git diff --check`, and `pio test -e native-macos -f test_packet_signing` (105/105 passed).

PR #11231, opened 2026-07-26
fix(router): release packets Router::send() declines instead of leaking them

ndoo:
## Problem

Two related packet-lifecycle bugs in `src/mesh/Router.cpp` and its callers, both against the STM32WL dynamic `packetPool` context from the ongoing memory-safety audit (#11230 and follow-ups):

1. `Router::send()` leaked `p` on the invalid-payload-variant `BAD_REQUEST` early return — every other early return in this function releases the packet first.
2. `assert(iface)` (both in `Router::send()` and separately in `NextHopRouter::setNextTx()`) hangs forever with no diagnostic on STM32WL instead of NAK'ing/logging like every other error path in the same file.
3. `ERRNO_SHOULD_RELEASE` is a documented contract: `RadioLibInterface::send()` can decline to send a packet and return this code *without* releasing it, expecting the caller to release (`MeshService::sendToMesh()` already does this correctly). `NextHopRouter::relayOpaquePacket()` and both `MeshBeaconModule::sendBeaconPacket()` call sites ignored the return value entirely and leaked `p` whenever `send()` declined — e.g. the `p->to == NODENUM_BROADCAST_NO_LORA` path in `RadioLibInterface::send()`.

I also traced `NextHopRouter::shouldFilterReceived()`'s rebroadcast path and all five call sites in `doRetransmissions()`, which looked like the same bug in an earlier audit pass — those turned out to already be fixed upstream (they already check for `ERRNO_SHOULD_RELEASE` and release correctly). Verified against a clean `upstream/develop` checkout rather than re-fixing already-fixed code.

## Fix

- `Router::send()`: release `p` before the `BAD_REQUEST` return; convert `assert(iface)` to the same `abortSendAndNak()` pattern already used for every other error in this function.
- `NextHopRouter::setNextTx()`: convert `assert(iface)` to a logged early return.
- `NextHopRouter::relayOpaquePacket()`, `MeshBeaconModule::sendBeaconPacket()` (both call sites): check the `send()` return value and release on `ERRNO_SHOULD_RELEASE`, matching `MeshService::sendToMesh()`'s existing correct handling.

## Test plan

- [x] `pio run -e wio-e5` / `pio run -e rak3172` — build clean.
- [x] Hardware (wio-e5): flashed, confirmed clean boot, set region, sent a text message, confirmed the device stays responsive throughout (`--info` healthy before/after).
- [ ] The specific leak triggers (`p->to == NODENUM_BROADCAST_NO_LORA` via `RadioLibInterface::send()`, opaque-relay mode, beacon broadcast under a declined send) are narrow, protocol-level conditions that weren't independently reproduced live in this pass — the fix is a mechanical return-value check mirroring the already-shipped-correct `MeshService::sendToMesh()` pattern in the same file.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/mesh/NextHopRouter.cpp
+++ b/src/mesh/NextHopRouter.cpp
@@ -27,7 +27,10 @@ bool NextHopRouter::relayOpaquePacket(const meshtastic_MeshPacket *p)
         return false;
     relay->hop_limit--;
     relay->relay_node = nodeDB->getLastByteOfNodeNum(getNodeNum());
-    return Router::send(relay) == ERRNO_OK;
+    ErrorCode res = Router::send(relay);
+    if (res == ERRNO_SHOULD_RELEASE)
+        packetPool.release(relay);
+    return res == ERRNO_OK;
 }
 
 PendingPacket::PendingPacket(meshtastic_MeshPacket *p, uint8_t numRetransmissions)
@@ -483,7 +486,10 @@ int32_t NextHopRouter::doRetransmissions()
 
 void NextHopRouter::setNextTx(PendingPacket *pending)
 {
-    assert(iface);
+    if (!iface) {
+        LOG_WARN("setNextTx: no interface, cannot schedule retransmission");
+        return;
+    }
     auto d = iface->getRetransmissionMsec(pending->packet);
     pending->nextTxMsec = millis() + d;
     LOG_DEBUG("Setting next retransmission in %u msecs: ", d);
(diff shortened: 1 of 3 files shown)


ndoo:
Closing — the `MeshBeaconModule::sendBeaconPacket()` fix, the `NextHopRouter::relayOpaquePacket()` fix, and the `Router::send()` `BAD_REQUEST`-leak/`assert(iface)` fixes here are all superseded by @t-miura's #11223 (opened before this one), which lands the same changes.

One fix in this PR isn't covered by #11223: `NextHopRouter::setNextTx()` has its own separate `assert(iface)` (a different call site than the one in `Router::send()`). I'll re-propose that one on its own in a follow-up PR.

PR #11232, opened 2026-07-26
fix(mesh): harden MemoryPool<T,MaxSize>::release() against bad pointers

ndoo:
## Problem

`MemoryPool<T,MaxSize>::release()` (`src/mesh/MemoryPool.h`) computed `int index = p - pool;` — pointer subtraction that's undefined behavior unless `p` already points into `pool`. A double-free hit `assert(used[index])`, which hangs forever with no diagnostic on STM32WL (`__wrap___assert_func` is `while(true);`).

This static pool backs `clientNotificationPool`/`staticQueueStatusPool`/`staticMqttClientProxyMessagePool` on every platform including STM32WL (unlike `packetPool`, which is `MemoryDynamic`-backed there) — so this is reachable, not theoretical.

## Fix

Validate via address arithmetic (`reinterpret_cast<uintptr_t>`) instead of raw pointer subtraction, rejecting a foreign/misaligned pointer before it's used as an index. Convert the double-free `assert()` into a logged, graceful return — matching the `MemoryDynamic::alloc()` precedent from #11197.

## Test plan

- [x] `pio run -e wio-e5` / `pio run -e rak3172` — build clean.
- [x] Hardware (wio-e5): temporarily patched in a one-shot test (not part of this diff) that allocates a `meshtastic_ClientNotification`, releases it, then deliberately releases the same pointer again and a misaligned pointer. Confirmed over serial: `"Double release of pool item 0 at 0x...`" and `"Pointer 0x... not from our pool!"` logged, no hang, boot continues normally (radio init succeeds). Test hook reverted before opening this PR.
- [x] Hardware (wio-e5): reflashed the clean committed fix afterward, confirmed normal boot/operation.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/mesh/MemoryPool.h
+++ b/src/mesh/MemoryPool.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <Arduino.h>
-#include <assert.h>
 #include <functional>
 #include <memory>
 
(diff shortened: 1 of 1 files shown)


ndoo:
Closing — this is fully superseded by @t-miura's #11223, which landed the same `MemoryPool<T,MaxSize>::release()` alignment/bounds validation and double-free detection (opened before this one). Thanks for getting to it first!

PR #11233, opened 2026-07-26
fix(radio): replace beginSending() asserts with graceful rejection

ndoo:
## Problem

`RadioInterface::beginSending()` guards four invariants with `assert()`: no send already in progress, payload already encrypted, sender node num set, and encrypted payload fits the radio buffer. On STM32WL these all hang forever with no diagnostic instead of dropping the one bad packet (`__wrap___assert_func` is `while(true);`).

## Fix

Convert each `assert()` to a checked early return: log, release `p`, return 0. Both callers (`RadioLibInterface::startSend()`, `SimRadio::startSend()`) are updated to check for a 0 return and bail out cleanly instead of proceeding to transmit/use the now-released packet.

## Test plan

- [x] `pio run -e wio-e5` / `pio run -e rak3172` / `pio run -e native-macos` — build clean (native-macos exercises the `SimRadio.cpp` caller).
- [x] Hardware (wio-e5): flashed, confirmed clean boot, sent a text message over LoRa and confirmed the device stays responsive — exercises the normal (non-failing) path through the new checks.
- [ ] The four specific invariant violations weren't independently reproduced live (they require corrupting internal router state to trigger) — this is a mechanical assert→checked-return conversion, same pattern as #11231/#11232.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -1462,10 +1462,18 @@ void RadioInterface::deliverToReceiver(meshtastic_MeshPacket *p)
  */
 size_t RadioInterface::beginSending(meshtastic_MeshPacket *p)
 {
-    assert(!sendingPacket);
-
-    // LOG_DEBUG("Send queued packet on mesh (txGood=%d,rxGood=%d,rxBad=%d)", rf95.txGood(), rf95.rxGood(), rf95.rxBad());
-    assert(p->which_payload_variant == meshtastic_MeshPacket_encrypted_tag); // It should have already been encoded by now
+    // These used to be asserts, which hang forever with no diagnostic on STM32WL. Reject and release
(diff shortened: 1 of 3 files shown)


ndoo:
Closing — this is superseded by @t-miura's #11223 (the `RadioInterface.cpp`/`RadioLibInterface.cpp` portions — null/size/variant checks in `beginSending()` and the `numbytes==0` handling in `startSend()`) and #11229 (the `SimRadio.cpp` portion, which handles the same `numbytes==0` case more thoroughly than this PR did — it changes `startSend()`'s signature to propagate success/failure back to `onNotify()` rather than just early-returning). Both were opened before this one. Thanks for the more complete fix!

PR #11234, opened 2026-07-26
fix(radio): don't hang on a transient SPI failure talking to the LoRa chip

ndoo:
## Problem

`assert(err == RADIOLIB_ERR_NONE)` / `assert(result != RADIOLIB_ERR_WRONG_MODEM)` guard SPI communication with the physical LoRa chip in `setStandby()`, `startReceive()`, `isChannelActive()` (plus a few config-time calls in `init()`) across all five RadioLib interface wrappers (`SX126xInterface.cpp`, `LR11x0Interface.cpp`, `SX128xInterface.cpp`, `RF95Interface.cpp`, `LR20x0Interface.cpp`). On STM32WL these hang forever with no diagnostic instead of a logged, recoverable error — a real risk since `setStandby()`/`startReceive()`/`isChannelActive()` run on every RX/TX state transition, i.e. continuously during normal operation, guarding against exactly the kind of transient SPI glitch (noise, marginal power, contention) that's a realistic field failure mode.

Tellingly, the `ARCH_PORTDUINO` branch sitting right next to each of these already treats the same condition as recoverable (`portduino_status.LoRa_in_error = true;` instead of asserting) — real hardware just never got the same treatment.

## Fix

Converted each `assert()` to `RECORD_CRITICALERROR()`: `INVALID_RADIO_SETTING` for the config-time calls (matching how every neighboring call in the same `init()` functions — `setSpreadingFactor`, `setBandwidth`, `setFrequency`, etc. — already handles a rejected setting), `RADIO_SPI_BUG` for the runtime SPI operations in `setStandby()`/`startReceive()`/`isChannelActive()`.

`SX126xInterface.cpp` is the only one of these actually compiled for STM32WL today (`RADIOLIB_EXCLUDE_{SX128X,SX127X,LR11X0,LR2021}=1`); the other four are identical copy-paste siblings. Fixed all five in the same pass — a future STM32WL variant using a different RadioLib-supported front-end is plausible given this board family's history of gaining hardware support over time (RTC, TCXO options), and it's the same mechanical fix either way.

## Test plan

- [x] `pio run -e wio-e5` / `pio run -e rak3172` — build clean (exercises `SX126xInterface.cpp`, the one compiled for STM32WL).
- [x] `pio run -e native-macos` — build clean, confirmed all 5 modified files actually compile (portduino builds every radio driver).
- [x] `pio run -e tbeam` (RF95/SX1276), `pio run -e heltec-wireless-bridge` (LR11x0 + LR2021), `pio run -e makerpython_nrf52840_sx1280_oled` (SX128x) — build clean on real hardware targets, exercising the non-`ARCH_PORTDUINO` branch (the actual new code path) for all four sibling files not compiled on STM32WL.
- [x] Hardware (wio-e5): flashed, confirmed clean boot (exercises `setStandby()`/`startReceive()` on every boot's radio init) and sent a text message over LoRa (exercises `isChannelActive()`'s CAD scan before transmit) — device stayed fully responsive throughout, confirming the success path through all six converted call sites in `SX126xInterface.cpp` is unaffected.
- [ ] The actual transient-SPI-failure condition wasn't reproduced live (would require physically glitching the SPI bus) — this is a mechanical assert→`RECORD_CRITICALERROR` conversion, matching the already-established pattern from #11230/#11231/#11232.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other (please specify below): wio-e5 — build + hardware verified per test plan above. rak3172, tbeam, heltec-wireless-bridge, makerpython_nrf52840_sx1280_oled, native-macos — build-verified only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Radio configuration and operation failures are now reported as critical errors instead of abruptly terminating the application.
  * Improved handling of standby, receive startup, modem detection, and invalid radio settings across supported radio hardware.
  * Portduino devices now flag radio failures through their status indicator for better diagnostics.

### DIFF
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -290,7 +290,8 @@ template <typename T> bool LR11x0Interface<T>::reconfigure()
         RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
 
     err = lora.setSyncWord(syncWord);
-    assert(err == RADIOLIB_ERR_NONE);
+    if (err != RADIOLIB_ERR_NONE)
+        RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
 
     if (config.lora.region == meshtastic_Config_LoRaConfig_RegionCode_LORA_24) { // clamp if wide freq range
         limitPower(LR1120_MAX_POWER);
@@ -299,14 +300,16 @@ template <typename T> bool LR11x0Interface<T>::reconfigure()
     }
 
     err = lora.setPreambleLength(preambleLength);
-    assert(err == RADIOLIB_ERR_NONE);
+    if (err != RADIOLIB_ERR_NONE)
+        RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
 
     err = lora.setFrequency(getFreq());
     if (err != RADIOLIB_ERR_NONE)
         RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
 
     err = lora.setOutputPower(power);
-    assert(err == RADIOLIB_ERR_NONE);
+    if (err != RADIOLIB_ERR_NONE)
+        RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING);
 
     // Apply RX gain mode - valid in STDBY, matches resetAGC() pattern
     err = lora.setRxBoostedGainMode(config.lora.sx126x_rx_boosted_gain);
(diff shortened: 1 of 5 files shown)


NomDeTom:
have you observed this in the wild?

I've got some issues with LR1121 of my own.

ndoo:
I don’t really know if this PR should be merged. It stops the firmware from intentionally crashing, but it leaves the node unreachable by LoRa anyway.

Pros for merging: You can still reach it by BLE/Wi-Fi/USB as available and reboot it from a Meshtastic client device.
Cons for merging: Device doesn’t provide external indication of a crash - which may be desirable if you are looking for a stuck screen/LED.

WDYT?

caveman99:
@ndoo please have a look at https://github.com/meshmy/meshtastic-firmware/pull/3 if that would work the way you intended it.

ndoo:
I’m not sure what the utility of this is, seems like it’s more likely indicative of poor SPI pin routing (electrically or physically) or a faulty SX/LRxxxx chip, or some other problem such as misconfigured TCXO. Just gonna close it for now.

NomDeTom:
Uh, no. If spi is responding then it's not wiring. I think this is probably a good fix.

ndoo:
> Uh, no. If spi is responding then it's not wiring. I think this is probably a good fix.

rebased on develop HEAD and reopened

PR #11236, opened 2026-07-26
fix: defensive buffer-write hardening (sprintf, mt_sprintf, NMEA checksum)

ndoo:
## Problem

A full-tree sweep for unsafe buffer writes (as part of an ongoing STM32WL memory-safety audit) found the codebase in good shape overall — this is the small remaining handful of loose ends. None are demonstrated as exploitable today; all are latent landmines worth closing while in this area.

- `src/modules/DropzoneModule.cpp`: 4 raw `sprintf()` calls into fixed-size buffers, safe today only because of current nanopb field-size invariants (no defense-in-depth if those change).
- `src/platform/stm32wl/STM32_LittleFS.cpp`: `dbg_strerr_lfs`'s `sprintf(errcode, "%ld", err)` into a 10-byte buffer — `INT32_MIN` needs 12 bytes (debug-build-only code path).
- `src/RedirectablePrint.cpp`: `mt_sprintf()` had a `strcpy()` immediately overwritten by the `vsnprintf()` right after it — pure dead code, and on an empty format string it underflows, writing 1 byte past a 0-size allocation.
- `src/gps/NMEAWPL.cpp`: `printWPL()`/`printGGA()` use `snprintf()`'s return value (which can exceed `bufsz` on truncation) directly as an offset/size for a second `snprintf()` call and a checksum loop — both need the length clamped first, or `bufsz-len` underflows into a huge `size_t`. Safe today only because of current field-size bounds.

## Fix

- Swap the remaining raw `sprintf()` calls for `snprintf()`.
- Remove the dead/unsafe `strcpy()` in `mt_sprintf()` — `vsnprintf()` alone already produces the correct result.
- Clamp `len` to `bufsz` (or 0) after each `snprintf()` call in `NMEAWPL.cpp` before using it as an offset or in the checksum loop.

## Test plan

- [x] `pio run -e wio-e5` / `pio run -e rak3172` — build clean, confirms all four touched files actually compile on the memory-constrained target these were audited for.
- [x] Hardware (wio-e5): flashed, confirmed clean boot.
- [ ] None of the specific overflow conditions were reproduced live (all require values outside current field-size invariants to trigger) — this is defensive hardening, not a fix for a demonstrated live bug.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/RedirectablePrint.cpp
+++ b/src/RedirectablePrint.cpp
@@ -386,7 +386,6 @@ std::string RedirectablePrint::mt_sprintf(const std::string fmt_str, ...)
     va_list ap;
     while (1) {
         formatted.reset(new char[n]); /* Wrap the plain char array into the unique_ptr */
-        strcpy(&formatted[0], fmt_str.c_str());
         va_start(ap, fmt_str);
         int final_n = vsnprintf(&formatted[0], n, fmt_str.c_str(), ap);
         va_end(ap);
--- a/src/gps/NMEAWPL.cpp
+++ b/src/gps/NMEAWPL.cpp
@@ -27,11 +27,19 @@ uint32_t printWPL(char *buf, size_t bufsz, const meshtastic_PositionLite &pos, c
                             (abs(geoCoord.getLatitude()) - geoCoord.getDMSLatDeg() * 1e+7) * 6e-6, geoCoord.getDMSLatCP(),
                             geoCoord.getDMSLonDeg(), (abs(geoCoord.getLongitude()) - geoCoord.getDMSLonDeg() * 1e+7) * 6e-6,
                             geoCoord.getDMSLonCP(), name);
+    // snprintf() returns the length that *would* have been written, which can exceed bufsz if
+    // truncated - clamp before using it as an index/size, or the checksum loop reads out of
+    // bounds and bufsz-len underflows into a huge size for the next snprintf().
+    if (len >= bufsz)
+        len = bufsz > 0 ? bufsz - 1 : 0;
+
     uint32_t chk = 0;
     for (uint32_t i = 1; i < len; i++) {
         chk ^= buf[i];
     }
     len += snprintf(buf + len, bufsz - len, "*%02X\r\n", chk);
+    if (len >= bufsz)
+        len = bufsz > 0 ? bufsz - 1 : 0;
     return len;
 }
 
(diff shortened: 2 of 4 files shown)


caveman99:
Thanks for this @ndoo, and sorry to close it out from under you.

#11293 started as the NMEA checksum offset fix from #11242, which lands on the same three lines in `NMEAWPL.cpp` that this PR clamps. CodeRabbit then raised your truncation finding there too. Rather than have two PRs editing the same statements and conflicting with each other, I pulled your remaining hunks into #11293: the two `sprintf` calls in `sendConditions()`, the dead `strcpy` in `mt_sprintf()`, and the `errcode` buffer size. That keeps it to one CI matrix run instead of two, and avoids whichever merged second having to rebase through the other.

You are credited as co-author on the commit. The two `matchCompare` `sprintf` calls you flagged in `DropzoneModule.cpp` were already converted in #11286, so only the two in `sendConditions()` were still outstanding.

Your analysis was right on every point, including the empty-format case where `n = fmt_str.size() * 2` is 0 and the `strcpy` writes past a zero-size allocation.

caveman99:
Superseded by #11293, which carries these changes with @ndoo credited as co-author.

ndoo:
> Thanks for this @ndoo, and sorry to close it out from under you.
> 
> #11293 started as the NMEA checksum offset fix from #11242, which lands on the same three lines in `NMEAWPL.cpp` that this PR clamps. CodeRabbit then raised your truncation finding there too. Rather than have two PRs editing the same statements and conflicting with each other, I pulled your remaining hunks into #11293: the two `sprintf` calls in `sendConditions()`, the dead `strcpy` in `mt_sprintf()`, and the `errcode` buffer size. That keeps it to one CI matrix run instead of two, and avoids whichever merged second having to rebase through the other.
> 
> You are credited as co-author on the commit. The two `matchCompare` `sprintf` calls you flagged in `DropzoneModule.cpp` were already converted in #11286, so only the two in `sendConditions()` were still outstanding.
> 
> Your analysis was right on every point, including the empty-format case where `n = fmt_str.size() * 2` is 0 and the `strcpy` writes past a zero-size allocation.

No worries. Tbh most of the groundwork was Claude analysis. Nonetheless just happy to see the fix ship. Cheers

PR #11243, opened 2026-07-26
ESP32: Don't run pkg install on build

vidplace7:
pioarduino runs this anyways. No need to do it twice.

## Summary by CodeRabbit

* **Chores**
  * Streamlined the ESP32 build process by removing the automatic package installation step.
  * Build and artifact generation steps remain unchanged.


vidplace7:
Holy crap this saves so much time!!

PR #11248, opened 2026-07-27
fix(t-echo-card): correct SSD1315 rendering and compact UI

### LINKED ISSUE #11221: [Bug]: T-Echo Card (t-echo-card) OLED boot logo and UI sheared/wrapped — 72x40 column mapping
**Category:** Hardware Compatibility
**Hardware:** LilyGO T-Echo Card (`t-echo-card` variant, `board_level = extra`)
**UI component:** OLED (SSD1315, 72x40, I2C @ 0x3C)
**Firmware Version:** 2.7.27.6de6c3d (also reproduced on 2.7.26.54e0d8d), self-built from source (no prebuilt artifact ships for this `extra` variant)

### Description
On the `t-echo-card` variant, the OLED renders incorrectly on both the boot logo and the runtime UI. The top ~2 pixel rows are placed roughly correctly, then each successive page shears progressively to the side, and the lower half collapses into diagonally-wrapped noise. The device is otherwise fully functional — BLE/serial connect, config, and mesh all work — so this is display-mapping only.

The symptom is consistent with a horizontal column-addressing mismatch: the driver appears to write beyond the panels real 72-column width, so each page wraps and walks the image sideways (same failure class as #9390 and discussion #338).

The variant declares:
```c
#define USE_SSD1306
#define OLED_GEOMETRY_OVERRIDE GEOMETRY_72_40
#define OLED_Y_OFFSET_PAGES 3
#define OLED_TINY
```
`OLED_Y_OFFSET_PAGES 3` handles the vertical page offset correctly; vertical placement looks right. The issue is horizontal — the variant relies on the display librarys built-in horizontal auto-centering to place the 72px window within the 128-column GDDRAM, and that centering does not appear to land correctly for this SSD1315 panel. A per-variant column offset (visible window start, e.g. col 28 for a centered 72px panel) is likely needed rather than relying on auto-centering.

### Steps to reproduce
1. Build `-e t-echo-card` from master (or the v2.7.26 tag).
2. Flash the resulting UF2 and boot.
3. Observe the boot logo shear diagonally; runtime UI shows the same.

**Reproducibility:** 100%, two firmware versions.

_cc @caveman99 (authored the variant in #10267)._

_Note: I will attach two boot-logo photos showing the shear._

h3lix1:
Fixes #11221.

Depends on meshtastic/esp8266-oled-ssd1306#29.

## What changed

- Update the shared OLED-driver dependency so `GEOMETRY_72_40` initializes the SSD1315 with its native 128×64 scan configuration while exposing the panel's visible 72×40 window at columns 28–99 and pages 3–7.
- Remove the firmware-local T-Echo Card display driver and use the standard `SSD1306Wire` plus `setYOffset()` path at the panel's validated 100 kHz I²C rate.
- Transfer exactly the remaining framebuffer bytes in the final I²C chunk in the shared driver.
- Add a T-Echo Card-only compact UI mode for shared headers and text rows, messages, navigation, notifications, system and position screens, clock rendering, and clock menu actions.
- Keep compact option-menu titles at y=0 so the packed font remains fully inside the framebuffer.
- Render Bluetooth pairing-PIN banners through the same tagged-font measurement and drawing path as ordinary text banners, and clamp alert-line copies to the visible compact-screen capacity.
- Continue using the standard Meshtastic UI components and data, with layouts adjusted to make the 72×40 display useful.

## Root cause

The generic `GEOMETRY_72_40` SSD1306 path configured the controller as a native 72×40 display. The T-Echo Card instead uses an SSD1315 with 128×64 GDDRAM and exposes a cropped 72×40 viewport.

For this geometry, the pinned OLED library also omitted required COMPINS and contrast parameters during initialization. Its fixed 16-byte transfer loop read eight bytes beyond the end of the 360-byte framebuffer on the final chunk. Together, the controller setup and transfer length caused the sheared and wrapped output.

Compact full-screen option menus additionally set `boxTop` to 0 while retaining a generic three-pixel title offset. That rendered the 13-pixel packed title font at y=-1, which discarded its first eight raster rows. Compact titles now start at y=0; normal display geometry is unchanged.

Bluetooth pairing messages include an `[M]` tag that selects the medium banner font. When pairing banners received the dedicated `pairing_pin` notification type, the two tag-aware renderer gates remained limited to `text_banner`. The PIN screen therefore measured and painted the raw `[M]` marker as visible text, which overflowed the 72-pixel-wide panel and appeared corrupt. Pairing banners now use the tag-aware path consistently for both measurement and rendering.

## User impact

The boot screen and normal Meshtastic UI now render cleanly on the T-Echo Card. Messages, node and GPS information, notifications, menus, navigation, LoRa and system status, and the digital clock fit within the available display area. Option-menu titles render completely, and the Bluetooth pairing screen shows only the intended heading and six-digit PIN rather than the internal font tag.

The compact behavior reuses `OLED_TINY` and is selected only for the exact 72×40 geometry; the 64×32 T-Impulse Plus and 64×48 UnitC6L retain their existing layouts. The pairing tag correction uses the existing notification behavior on all geometries.

## Validation

- Shared OLED driver:
  - `pio ci` with `examples/SSD1306SimpleDemo` for `d1_mini`
  - `pio ci` with `examples/SSD1306SimpleDemo` for `esp32doit-devkit-v1`
  - `git diff --check`
- Firmware:
  - `trunk fmt`
  - `trunk check`
  - `git diff --check`
  - `pio run -e t-echo-card`
    - RAM: 37.8%
    - Flash: 90.9%
    - nRF52 warm-region, ISR-handler, and variant guards passed
  - `pio run -e t-impulse-plus`
    - RAM: 44.7%
    - Flash: 91.2%
  - `~/.platformio/penv/bin/pio run -e m5stack-unitc6l`
    - RAM: 34.6%
    - Flash: 39.0%
  - Verified PlatformIO downloaded OLED commit `846e0b424cd5405f65bbe4168aa2dcfe1cbfb3ad`; the installed sources are byte-identical and contain both the SSD1315 initialization and bounded final transfer.
- The equivalent firmware-local driver sequence was previously flashed to a physical LilyGO T-Echo Card, where the user confirmed clean general display output. Reflashing the shared-driver integration remains pending because the device is not currently enumerating over USB.

The native test suite was attempted but did not reach test execution on this macOS host because of existing runner/toolchain assumptions (`mapfile`, GNU `find -printf`, and Linux-only native dependencies).

## Attestations

- [x] I have tested that my proposed changes do not cause any obvious regressions on:
  - [x] Other: LilyGO T-Echo Card

## Summary by CodeRabbit

* **New Features**
  * Added compact-panel UI layout across clock, navigation, messages, notifications, GPS, and system screens.
  * Compact clock rendering now centers a segmented time display (and optional period).
  * Pairing PIN alerts can use the tagged text banner style when applicable.
  * Tiny clock menu options are simplified to Time Format and Timezone.

* **Bug Fixes**
  * Improved compact layout spacing, including header/footer/underline and scrollbar placement.
  * Notification rendering now respects visible line limits to avoid overdraw.

* **Chores**
  * Updated the OLED display driver dependency and enabled compact UI configuration for the supported build variant.


src/graphics/TCardSSD1315Wire.h
```diff
@@ -0,0 +1,85 @@
+#pragma once
```
caveman99: Please don't clutter the OLED driver library over several repositories. this belongs in the oled driver.

caveman99 (changes requested):

src/graphics/TCardSSD1315Wire.h
h3lix1: Ah, yup.. added to https://github.com/meshtastic/esp8266-oled-ssd1306/pull/29

h3lix1:
@caveman99 let me know if there are more things you see that should be updated. CUrrently blocking on your review.

Thanks

Xaositek:
We already have OLED_TINY, what does OLED_COMPACT_UI do that OLED_TINY doesn't already? I feel like we should re-use as much as possible to avoid a "what does one do versus the other" scenario in the future.

h3lix1:
> We already have OLED_TINY, what does OLED_COMPACT_UI do that OLED_TINY doesn't already? I feel like we should re-use as much as possible to avoid a "what does one do versus the other" scenario in the future.

@Xaositek It does seem like it should fit OLED_TINY except I didn't have any other of the tiny displays to test at the time.  I'll swap things out to use OLED_TINY. A c6l just arrived today, but someone will need to test the impulse plus for sake of completeness.

Xaositek:
@HarukiToreda Do you have an Impulse? I thought you might have.

HarukiToreda:
I don't have a T-impulse but this falls into a fix I was already working on for the T-Echo card and the entire UI for all compact size screens like this one. I tested this on my node and it works.

> @HarukiToreda Do you have an Impulse? I thought you might have.

h3lix1:
> I don't have a T-impulse but this falls into a fix I was already working on for the T-Echo card and the entire UI for all compact size screens like this one. I tested this on my node and it works.

I don't think anybody has a t-Impulse yet, so we'll not need to worry about the inevitable bug report that will arrive in a couple of months. 😃 

If you were already working on a fix, I'm cool with withdrawing the merge.. or merge this for now with understanding it will be fully replaced later. I just wanted to fix my tiny screen.

caveman99:
@h3lix1 i merged the OLED driver fix. I think the rest of this is covered by the omnibus PR.

PR #11250, opened 2026-07-27
fix(t-echo-card): correct GNSS power and standby handling

h3lix1:
## Summary

- correct the T-Echo Card's active-low GNSS main-enable polarity
- model P0.29 as the external RF front-end enable instead of a reset line
- add optional RF front-end handling to the shared GPS power-state transitions
- keep the L76K powered and use its standby input between scheduled fixes
- park the wake and RF-enable pins low while the peripheral rail starts

## Root cause

The T-Echo Card variant treated the active-low GNSS enable as active-high and labeled the RF front-end enable as a reset pin. At longer polling intervals, the generic sleep heuristic could also cycle the L76K's main rail instead of retaining its warm state, preventing reliable reacquisition.

## Impact

Periodic GNSS polling now uses standby on this board, preserving the receiver's warm state across longer intervals. Explicitly disabling GNSS still uses the hard-off path. Boards without an external RF-enable definition retain the existing behavior, and position-broadcast policy is unchanged.

## Validation

- `trunk fmt`
- `trunk check`
- `pio run -e t-echo-card`
- `pio run -e rak4631`
- flashed and exercised a physical T-Echo Card at 8-, 30-, and 60-second polling intervals
- observed the L76K acquire a fix, enter soft sleep, wake, and publish subsequent valid positions at 60 seconds
- confirmed an indoor fix with 17 satellites

## Summary by CodeRabbit

- **New Features**
  - Improved GPS power management by automatically enabling or disabling the external RF front-end across active, idle, and sleep states.
  - Added configurable polarity for the RF front-end enable signal.
  - Added support for forcing soft-sleep mode when available.

- **Bug Fixes**
  - Updated early boot behavior to keep the GPS RF front-end disabled while powering up other hardware.
  - Adjusted GPS GPIO initialization to improve standby and low-power reliability.


variants/nrf52840/t-echo-card/variant.cpp
```diff
@@ -36,13 +36,13 @@ void initVariant()
...
-// examples/original_test/original_test.ino. Runs before Meshtastic touches
-// PIN_POWER_EN, so the RT9080 LDO gets a clean reset pulse and peripherals
-// whose EN pins must be LOW at boot (GPS_EN, GPS_RF_EN, BUZZER) aren't left
-// floating while the 3V3 rail is ramping.
+// Runs before Meshtastic touches PIN_POWER_EN. Hold the GPS RF front end
+// disabled before toggling the RT9080 rail.
 void earlyInitVariant()
 {
+    pinMode(PIN_GPS_RF_EN, OUTPUT);
+    digitalWrite(PIN_GPS_RF_EN, LOW);
```
caveman99, line 44: der Schaltplan sagt, das stimmt, der Beispielcode von Lilygo setzt HIGH. Was stimmt jetzt? Der Schaltplan oder der Code?

h3lix1: Die Schaltung, um den Pin auf Low zu setzen, funktioniert bei meinem Gerät.

caveman99:
these changes are included in #11342 - closing in favor and crediting

PR #11138, opened 2026-07-22
feat(lr2021): enable LR2021 LoRa radio support on native/portduino

a-li3n:
Add full LR2021 radio support for meshtasticd on Linux native builds, including USB (CH341) and SPI (GPIO) configurations.

Key changes:

PortduinoGlue (config plumbing):
- Add use_lr2021 to lora_module_enum and loraModules map
- Add lr2021_max_power, lr2021_max_power_hf, lr2021_irq_dio_num config fields
- Parse LR2021_MAX_POWER, LR2021_MAX_POWER_HF, IRQ_DIO_NUM (and LR2021_IRQ_DIO_NUM alias) from YAML
- Use LR2021::MODE_* enums for rfswitch_table when module is lr2021 (LR2021 has MODE_RX_HF at value 4 where LR11x0 has MODE_TX_HP; no MODE_GNSS/MODE_WIFI)
- Parse MODE_RX_HF YAML key for LR2021 configs instead of MODE_TX_HP
- Branch YAML serialization on use_lr2021 so configs round-trip correctly

LR20x0Interface (radio driver):
- Set irqDioNum BEFORE lora.begin() so config() programs the correct DIO for IRQ routing. Setting it after begin() is too late - config() has already configured DIO5 as IRQ, then setRfSwitchTable() overrides it to RF_SWITCH, breaking all radio interrupts. This is the fix for the DIO5 IRQ/RF-switch conflict that caused meshtasticd to crash (SIGABRT/segfault) on TX when DIO5 is used for both IRQ and RF switch control.
- Add ARCH_PORTDUINO path for reading irqDioNum from config
- Add LR20X0_RFSWITCH_NATIVE guard for variant-defined RF switch tables

LR11x0Interface (symbol namespacing):
- Namespace RF switch symbols as lr11x0_rfswitch_* to prevent collisions when LR11x0 and LR20x0 templates share a translation unit (InterfacesTemplates.cpp). Removes the old #undef hack.

RadioInterface:
- Add case use_lr2021 to construct LR2021Interface in portduino mode
- Guard hardcoded SPI pin fallback with !defined(ARCH_PORTDUINO)

ProMicro rfswitch.h:
- Add per-table include guards (RFSWITCH_NRF52TCXO_LR11X0_DEFINED, RFSWITCH_NRF52TCXO_LR20X0_DEFINED) to prevent redefinition
- Nest #ifdef USE_LR1121 / #ifdef USE_LR2021 inside include guards so tables are only emitted when their radio is built
[...]

PR #11252, opened 2026-07-27
add LR 2021 to portduino, and allow Framebuffer devices to rotate the screen from config

caveman99:
Requires https://github.com/meshtastic/device-ui/pull/355 and supersedes https://github.com/meshtastic/firmware/pull/10567 and https://github.com/meshtastic/firmware/pull/11138

Many thanks to the original authors https://github.com/a-li3n and https://github.com/jessm33

N.B. This PR enables meshtasticd with framebuffer MUI on Lilygo T-Display K230

## Summary by CodeRabbit

* **New Features**
  * Added Portduino support for selecting and initializing the LR2021 LoRa module (including native/WASM builds).
  * Added configurable LR2021 maximum power limits (standard and high-frequency) with YAML import/export.
  * Added framebuffer rotation configuration via an environment value derived from the display rotation settings.
* **Bug Fixes**
  * Corrected which LR2021/LR20x0 interfaces are compiled and selected for Portduino and native targets, improving build consistency.
  * Improved safety by handling failures when setting the framebuffer rotation environment value, using a default if needed.

### DIFF
--- a/src/graphics/tftSetup.cpp
+++ b/src/graphics/tftSetup.cpp
@@ -11,8 +11,8 @@
 
 #ifdef ARCH_PORTDUINO
 #include "PortduinoGlue.h"
-#include <cstdio>  // snprintf
-#include <cstdlib> // setenv
+#include <cstdio>
+#include <cstdlib>
 #include <thread>
 #endif
 
@@ -62,8 +62,7 @@ void tftSetup(void)
         } else
 #elif defined(USE_FRAMEBUFFER)
         if (portduino_config.displayPanel == fb) {
-            // Config-driven display rotation, amount from yaml Display.OffsetRotate
-            // (1=90 deg, 2=180 deg, 3=270 deg)
+            // Rotation from yaml Display.OffsetRotate: 1=90, 2=180, 3=270 deg
             char rbuf[4];
             snprintf(rbuf, sizeof(rbuf), "%d", portduino_config.displayRotate ? (portduino_config.displayOffsetRotate & 3) : 0);
             setenv("MESHTASTIC_FB_ROTATION", rbuf, 1);
--- a/src/graphics/tftSetup.cpp
+++ b/src/graphics/tftSetup.cpp
@@ -65,7 +65,8 @@ void tftSetup(void)
             // Rotation from yaml Display.OffsetRotate: 1=90, 2=180, 3=270 deg
             char rbuf[4];
             snprintf(rbuf, sizeof(rbuf), "%d", portduino_config.displayRotate ? (portduino_config.displayOffsetRotate & 3) : 0);
-            setenv("MESHTASTIC_FB_ROTATION", rbuf, 1);
+            if (setenv("MESHTASTIC_FB_ROTATION", rbuf, 1) != 0)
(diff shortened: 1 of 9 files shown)


vidplace7 (changes requested):
Currently breaks the `native-wasm` target at link-time.
> Linking .pio/build/native-wasm/meshnode.mjs
wasm-ld: error: .pio/build/native-wasm/src/mesh/RadioInterface.cpp.o: undefined symbol: LR2021Interface::LR2021Interface(LockingArduinoHal*, unsigned int, unsigned int, unsigned int, unsigned int)
em++: error: '/home/runner/work/firmware/firmware/emsdk-cache/emsdk-main/upstream/bin/wasm-ld @/tmp/emscripten_w_bs6_zi.rsp.utf-8' failed (returned 1)

src/graphics/tftSetup.cpp
```diff
@@ -60,6 +60,11 @@ void tftSetup(void)
         } else
 #elif defined(USE_FRAMEBUFFER)
         if (portduino_config.displayPanel == fb) {
+            // Config-driven display rotation:, amount from yaml Display.OffsetRotate
+            // (1=90�, 2=180�, 3=270�)
+            char rbuf[4];
+            snprintf(rbuf, sizeof(rbuf), "%d", portduino_config.displayRotate ? (portduino_config.displayOffsetRotate & 3) : 0);
+            setenv("MESHTASTIC_FB_ROTATION", rbuf, 1);
```
mverch67: Using env to pass parameters from one function to another...  :joy:
It would be much cleaner to pass this info within the existing DisplayDriverConfig interface, e.g.
```
displayConfig.device(DisplayDriverConfig::device_t::FB)
             .panel(DisplayDriverConfig::panel_config_t{.type = panels[portduino_config.displayPanel],
                                                        .panel_width = (uint16_t)portduino_config.displayWidth,
                                                        .panel_height = (uint16_t)portduino_config.displayHeight,
                                                        .offset_rotation = (uint8_t)portduino_config.displayOffsetRotate });
```
and access this in the device-ui FBDriver by adding a new rotation parameter to ctor and passing this parameter in the two calls of FBDriver in DisplayDriverFactory (_or alternatively if there are even more parameters expected in future then pass the entire DisplayDriverConfig to FBDriver ctor_).

src/graphics/tftSetup.cpp
```diff
@@ -60,6 +60,11 @@ void tftSetup(void)
         } else
 #elif defined(USE_FRAMEBUFFER)
         if (portduino_config.displayPanel == fb) {
+            // Config-driven display rotation:, amount from yaml Display.OffsetRotate
+            // (1=90�, 2=180�, 3=270�)
+            char rbuf[4];
+            snprintf(rbuf, sizeof(rbuf), "%d", portduino_config.displayRotate ? (portduino_config.displayOffsetRotate & 3) : 0);
```
mverch67, line 67: The option displayRotate is only [used](https://github.com/meshtastic/device-ui/blob/bcb327f058469282408eed93e27249d6447de9c0/include/graphics/LGFX/LGFXConfig.h#L62) for physical devices when width/height need to be swapped during rotation and otherwise does not match the UI canvas width/height. It's not meant as a precondition for offset rotation, so for FB this can be dropped.

PR #11259, opened 2026-07-27
Add optional NXP SE050 secure-element PKI key agreement

cvaldess:
## Summary

Adds an optional NXP SE050 secure-element backend for the existing PKI key-agreement path (X25519/Curve25519 ECDH). When present and enabled, the SE050 performs the ECDH for a node's PKI direct messages; when absent, the node behaves exactly as it does today.

This is deliberately scoped to the part that doesn't change any product semantics: the node's identity stays exactly where it lives today (`config.security.private_key`), exportable and backup-able as always. A copy of that key is mirrored into the chip so it can run the agreement; nothing about identity backup, remote key rotation, or NodeNum derivation changes. See "Not in this PR" below for the bigger, separate conversation this deliberately doesn't touch.

Everything is gated behind two build flags (`HAS_SE050`, `HAS_CUSTOM_CRYPTO_ENGINE`), both off by default and only defined for `pico2_w5500_e22` in this PR (the one board here that carries the chip). Every other board's binary is byte-for-byte unaffected — the new `.cpp` files compile to nothing without those flags.

## What's added

- `src/security/SE050.{h,cpp}` — the chip driver: T=1oI2C transport (block framing, CRC-16, ATR), APDU exchange with WTX handling, a PlatformSCP03 secure channel (the SE050 refuses key agreement outside an authenticated channel), and an X25519 identity that's either generated in the chip (`identityEnsure`, private half never leaves it) or imported to mirror a key the caller already holds (`identityImport`), backing one ECDH agreement at a time (`identityEcdh`).
- `src/security/SE050CryptoEngine.cpp` — a `CryptoEngine` subclass overriding only `setDHPublicKey` (the single method both `encryptCurve25519`/`decryptCurve25519` route through), so it's the one place that needs to know the agreement can run in hardware. Falls back to `CryptoEngine`'s own software Curve25519 if the mirror import ever fails for any reason — a node is never left unable to complete a key agreement because a secure element didn't cooperate.
- `src/main.cpp` — an ENA power-on-reset pulse before the I2C scan (the SE050 has no reset line and comes up however the last power cycle left it), probing + constructing the driver right after the scan if it found one, and a self-test call placed deliberately after the network is up rather than next to NodeDB where the identity actually loads — if it ever hangs, the board still answers over the network instead of needing a bootloader pull.
- `variants/rp2350/diy/pico2_w5500_e22/platformio.ini` — the 3 lines enabling it on this board.

Two implementation details worth flagging for review:

- **Buffers are object members, not stack locals.** A single key agreement nests `identityEcdh -> sessionApdu -> secureApdu -> transceive -> xfer`, and as locals those buffers came to ~2.5 KB in one call chain — fine entered from `setup()`'s shallow frame, not fine entered from deep inside a packet handler on a board with a 4 KB FreeRTOS task stack (RP2350 targets build with `-D__FREERTOS=1`; `setup()`/`loop()` run in a `1024`-word CORE0 task). Moving them into the object is what makes this safe from a real packet-processing context, not just from a boot-time self-test.
- **Re-entrancy guard.** `reentered()` (checked at every layer) refuses to start new work on the chip while a transaction is parked waiting for a response — a second operation half-way through the first would otherwise advance the SCP03 counter and overwrite buffers still in flight.

## Not in this PR

The mirrored key stays exportable and backup-able exactly as it is today. A mode where the key is generated in the chip and never mirrored to config at all is a real possibility (`identityEnsure()` already supports it, unused by anything in this PR), but it's a bigger product conversation — what happens to remote key rotation (an admin client can send its own chosen private key today; the chip can't be told "use exactly this value"), to local identity backup/export (today a locally-connected client can read its own private key back for backup purposes), to a lost or replaced chip. Not something to fold into a driver PR.

## On-hardware verification

Tested against a real SE050E2 (applet 7.2.0) on `pico2_w5500_e22`, built from this branch (fresh off `develop`, not a byte-copy of a prior working tree):

- Clean build: `pio run -e pico2_w5500_e22` — RAM 18.0%, Flash 25.4%.
- Cold boot (physical reset, serial-captured): `SE050: node key already mirrored` → `SE050: PKI key agreement now runs in hardware` → `SE050: self-test OK - PKI agreement through the chip matches software`.
- ECDH agreement through the chip matches software Curve25519 byte for byte (self-test compares both).
- A real PKI direct message, sent from a peer node and received on this one: `Attempt PKI decryption` → `SE050: key agreement leaves 1240 bytes of the 4 KB CORE0 task stack` → `PKI Decryption worked!` → message delivered intact. No reset, node kept operating normally on real mesh/MQTT traffic afterward.
- Timing: ~64ms per agreement against the chip (SCP03-wrapped: AES-CMAC + AES-CBC on both command and response, plus a nested UserID session, over 100kHz I2C) vs ~27ms for the bundled software Curve25519 on the same RP2350 — slower, but a LoRa packet's own airtime dwarfs the difference.

## 🤝 Attestations

  - Not applicable to Heltec/T-Deck/T-Beam/RAK/Seeed — this change is entirely gated behind `HAS_SE050`/`HAS_CUSTOM_CRYPTO_ENGINE`, both undefined everywhere except `pico2_w5500_e22` in this PR, so every other board's build is byte-for-byte unaffected.
  - [x] Other: `pico2_w5500_e22` (see on-hardware verification above)

## Summary by CodeRabbit

- **New Features**
  - Added support for the NXP SE050 secure element over I2C.
  - Added protected SCP03 secure-channel communication and secure APDU exchanges.
  - Added hardware-backed X25519/ECDH key agreement with automatic software fallback.
  - Added secure-element identity management and cryptographic self-tests.
- **Hardware Integration**
  - Added optional enable-pin handling and reliable device detection.
- **Configuration**
  - Enabled SE050 support for the Pico 2 W5500 E22 variant, including its enable pin and custom cryptography.


cvaldess:
Thanks for the review — went through all 6 actionable findings against the actual code (not just the diff) and fixed everything that was real. Pushed as a new commit rather than folding into the originals so the fix is easy to diff against the review that prompted it.

**Fixed (all 6 actionable):**
- `main.cpp`: null-check on `new SE050(...)` before calling `probe()`.
- SCP03 host challenge was using `random()` (non-cryptographic) instead of the platform's real entropy source — switched to `HardwareRNG::fill()`, the same facility `CryptoEngine` already uses for XEdDSA/Curve25519 keygen (RP2350's hardware ROSC RNG).
- `secureApdu()`'s failure exits after `scp.counter++`/`chainedCmac()` left the host's SCP03 counter/MCV advanced for commands the card never confirmed — every such exit now also closes `scp.open`/`sessionActive`, so a wedge fails loudly once instead of silently forever. (The crypto engine already falls back to software on any agreement failure, so this doesn't change reliability — it changes a stuck channel from "keeps failing the same way" to "fails once, the fallback takes over.")
- `bus.requestFrom()` takes `size_t` on this platform — the `(uint8_t)` casts around it were pure narrowing, and a real bug: LEN 253-255 computes a body of 255-257, which wraps through `uint8_t` into a request for far fewer bytes than the frame needs. Dropped.
- Two more truncation points you didn't call out by name but are the same class of bug: the `Lc` byte (`encLen + 8`) and the `LEN` byte (`apduLen`) in `transceive()` are both single-byte ISO7816/UM11225 fields, tighter than the surrounding buffer-capacity checks alone enforce — a ~240-283 byte payload would pass those checks and then silently wrap. Both now explicitly bounded to 255.

**Nitpicks taken:** deleted copy/move on `SE050` (it holds live session state + ~2KB of buffers), typed pointer instead of `static_cast` on the global `crypto`, `sw` now logged when the two idempotent setup APDUs return neither success nor "already exists", object-ID/address logging switched to the project's `0x%08x`/`0x%x` convention, and the 5-round ECDH timing benchmark in `probe()` moved behind an opt-in `-D SE050_BENCHMARK` (off by default — not something a shipped device needs to redo every boot).

Also reworded "chip authenticated" → "secure channel open": the SCP03 keys are NXP's public factory defaults (documented right above that line), so a successful handshake confirms a chip that speaks PlatformSCP03 correctly, not one trusted more than an attacker with physical I2C access could be. Rotating to per-device keys is real follow-up work, out of scope here (see "Not in this PR" in the description).

**Not changed, with reasoning:**
- The fixed 250ms ENA settle delay → bounded polling: the chip ACKs its I2C address well before it can actually answer the T=1oI2C soft reset (empirically confirmed during bring-up — 10ms wasn't enough, chip acked but returned nothing), so polling on ACK alone would just move the same race earlier, not remove it. `reset()` and `xfer()` already retry bounded on top of this floor (4 and up to 400 attempts respectively).
- Trimming the multi-paragraph comments: this is the one file in the tree explaining a chip and protocol nobody else here has touched, and every block is answering a specific "why is this necessary" (buffer placement, ENA timing, mirror-overwrite safety) rather than restating the code. You flagged these as your own lowest tier (trivial/low-value) — left as-is.
- The `mirrorReady()` explicit-warmup suggestion — already effectively provided by the `se050CryptoSelfTest()` call in `main.cpp`, which runs before any real packet could plausibly arrive (right after network bring-up, well before mesh join).

Re-verified on hardware after all of the above: clean build, and a second real PKI DM sent from a peer node and decrypted through the chip without incident — confirms `secureApdu()`'s (unchanged) success path still works, only the failure exits gained the new state reset.

cvaldess:
Pushed `ee325736c`, which fixes a bug I introduced in the review round above.

While addressing the earlier feedback I made `secureApdu()` close the SCP03 channel at every exit that could not confirm the card was still in step, and I included the "short response" path in that. It was too broad. A response shorter than 10 bytes is the card answering with a bare status word and no R-MAC, which is what it does for an ordinary application error: the command did travel the secure channel, the card accepted the C-MAC and advanced its counter in step with ours.

The effect was that the idempotent setup steps poisoned the session they were preparing:

* `WriteUserID` returns `6985` on a chip that already holds the authenticator object. `identitySession()` treats that SW as benign and logs nothing, so the teardown left no trace.
* The `CreateSession` right behind it then hit `!scp.open` and came back with the driver's internal `0xFFFF`.

A chip with no identity yet fails the same way from the other side, where `identityEnsure()` relies on a failed `ReadObject` leaving the channel usable so it can go on to generate the key.

The fix closes the channel only when the card cannot have processed the command - no status word at all, or one saying the secure messaging itself was refused (`6982`/`6987`/`6988`) - and logs it when that happens, since the silent teardown is what made this hard to spot.

Verified on a `pico2_w5500_e22` with an SE050E2 that was provisioned earlier.

Before:

```
SE050: SCP03 secure channel open
SE050: CreateSession SW=ffff
SE050: identity not available
SE050: self-test FAILED - key agreement is not running on the chip, PKI stays in software
```

After:

```
SE050: SCP03 secure channel open
SE050: reusing existing identity (objId 0x4d544944)
SE050: identity public key 0462BB05...32BD6D
SE050: ECDH matches software byte for byte - hardware X25519 is usable
SE050: node key already mirrored (objId 0x4d544b59)
SE050: PKI key agreement now runs in hardware
SE050: key agreement leaves 1816 bytes of the 4 KB CORE0 task stack
SE050: self-test OK - PKI agreement through the chip matches software
```

PR #11267, opened 2026-07-28
fix(router): clear failed reliable send retries

### LINKED ISSUE #11266: [Bug]: failed reliable sends remain scheduled for retransmission
### Category

Other

### Hardware

Not Applicable

### Is this bug report about any UI component firmware like InkHUD or Meshtastic UI (MUI)?

- [ ] InkHUD ePaper

### Firmware Version

2.8.0.dafa583 (develop)

### Description

When a locally originated packet requests an ACK, `ReliableRouter` creates its retransmission record before the first encode/send attempt. If that attempt fails synchronously, for example because its requested channel cannot be encoded or the radio cannot enqueue it, the normal immediate NAK is emitted but the retransmission record remains.

It can later be retransmitted and ultimately produce `MAX_RETRANSMIT` even though the original send was already rejected. This leaves clients with conflicting delivery state and may retry a packet that never entered the transmit queue.

Expected: an initial send failure must retain the immediate error and remove the pending reliable-send record.

### Reproduction

1. Create a local `want_ack` unicast with an out-of-range channel index.
2. Send through `ReliableRouter`.
3. Observe one immediate `NO_CHANNEL` client NAK.
4. Before the fix, the router still contains one pending retransmission entry.

### Acceptance Criteria

- Synchronous initial encode/interface failures do not leave a retransmission entry.
- The immediate client NAK is preserved.
- Successful reliable sends retain existing retry behavior.

RCGV1:
## Summary
- Remove a reliable-send retransmission entry when the initial downstream send fails.
- Add coverage for an out-of-range channel that reports `NO_CHANNEL` immediately and leaves no pending retry.

## Why
- A packet rejected before it reaches the radio should not later be retransmitted or produce a conflicting `MAX_RETRANSMIT` result.

## Validation
- `trunk fmt`
- `~/.platformio/penv/bin/python -m platformio test -e native-macos -f test_packet_signing` (70/70)
- `~/.platformio/penv/bin/python -m platformio test -e native-macos` (785/785)
- `~/.platformio/penv/bin/python -m platformio run -j 4 -e tbeam-s3-core`

Fixes #11266

## Summary by CodeRabbit

* **Bug Fixes**
  * Reliable mesh messages that fail during initial transmission are no longer retried unnecessarily.
  * Send failures are now reported back to the originating client more accurately, ensuring no retransmission remains pending when appropriate (including correct handling around duty-cycle limits).

* **Tests**
  * Added new Unity coverage for reliable send failure behavior (including an interface/send failure path) to confirm proper acknowledgements and that retransmissions are not left pending.
  * Extended routing/duty-cycle scenarios and updated mocks/fixture cleanup to validate queued status reporting correctly.

### DIFF
--- a/src/mesh/ReliableRouter.cpp
+++ b/src/mesh/ReliableRouter.cpp
@@ -38,7 +38,8 @@ ErrorCode ReliableRouter::send(meshtastic_MeshPacket *p)
     }
 
     ErrorCode result = isBroadcast(p->to) ? FloodingRouter::send(p) : NextHopRouter::send(p);
-    if (retransmitting && result != ERRNO_OK)
+    // Duty-cycle rejections may clear before the scheduled retry.
+    if (retransmitting && result != ERRNO_OK && result != meshtastic_Routing_Error_DUTY_CYCLE_LIMIT)
         stopRetransmission(key);
 
     return result;
--- a/src/mesh/ReliableRouter.cpp
+++ b/src/mesh/ReliableRouter.cpp
@@ -16,6 +16,9 @@
  */
 ErrorCode ReliableRouter::send(meshtastic_MeshPacket *p)
 {
+    const GlobalPacketId key(p);
+    const bool retransmitting = p->want_ack;
+
     if (p->want_ack) {
         DEBUG_HEAP_BEFORE;
         auto copy = packetPool.allocCopy(*p);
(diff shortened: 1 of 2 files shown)


caveman99:
what if the rejection is transient, e.g. based on duty cycle? that may have cleared the next time around.

caveman99:
fixes PR #11266

caveman99 (approved):

PR #11151, opened 2026-07-22
Actions: Surgical CI - Only run PRs upon actual change

vidplace7:
This pull request introduces several improvements to the CI workflow and makes minor configuration changes to two board variants. The main workflow now dynamically narrows the build and check matrices to only the files affected by a pull request, improving efficiency. It also adds a stable "ci-gate" job for branch protection, ensuring that all relevant builds and checks pass before merging. Additionally, two board variants are reclassified from "community" to "extra".

**CI Workflow Improvements:**

* The build matrix in `.github/workflows/main_matrix.yml` now uses `tj-actions/changed-files` to detect changed files and narrows the build and static analysis jobs to only the affected variants on pull requests, reducing unnecessary builds.
* Introduced a stable `ci-gate` job that aggregates the results of all build and check jobs, providing a single status check for branch protection. This ensures merges only happen when all relevant jobs succeed.
* The `gather-artifacts` job is now skipped on pull requests, as narrowed builds may not produce all expected artifacts, and this job is only needed for release and nightly workflows.
* Updated documentation in the manifest download step to clarify handling of missing or narrowed-out environments in size reporting.

**Board Variant Configuration:**

* Changed `board_level` from `community` to `extra` in `variants/rp2350/diy/pico2_w5500_e22/platformio.ini` and `variants/rp2350/diy/wiznet_5500_evb_pico2_e22p/platformio.ini`, reclassifying these variants. [[1]](diffhunk://#diff-7e140cb899f931243e571e8b8b6ea416dca0e321df15865310dc199a538068e3L4-R4) [[2]](diffhunk://#diff-f3c9b35c1327302ca93d401f11c6ee00b785886240ed810353510106feec0e8aL4-R4)

PR #11289, opened 2026-07-30
Package meshtasticd for Windows as an MSI

caveman99:
Installs the Windows build as a background service instead of shipping a bare binary.

- `--service` connects meshtasticd to the Service Control Manager. Dispatcher runs on its own thread from `parse_opt()`, so portduino's `main()` is unchanged. Stop sets `shutdownAtMsec` and reuses the existing `Power::shutdown()` save-and-exit path.
- MSI installs an auto-start LocalSystem service. Config at `%ProgramData%\Meshtastic\config.yaml`, `NeverOverwrite` and `Permanent` so an edited one survives upgrade and uninstall. `Vital="no"`.
- `bin/build-winget-package.ps1` builds the MSI and emits the winget manifests.
- CI builds the MSI on every Windows run and uploads it as an artifact.

WiX is pinned below v6, which refuses to build without accepting the Open Source Maintenance Fee EULA.

Not included, deliberately:

- No `main_matrix.yml` changes. Attaching the MSI to releases waits for the matrix rework in #11151, alongside MacOS.
- `package_winget.yml` validates manifests against a published release but does not submit them. Publishing to microsoft/winget-pkgs stays manual for now.

Verified: `pio run -e native-windows`, MSI service/config rows by decompile, `winget validate`, trunk.

Service lifecycle not exercised end to end, which needs an elevated msiexec and attached hardware.

## Summary by CodeRabbit

- **New Features**
  - Added Windows MSI packaging and Microsoft Store-compatible manifest generation for x64 and ARM64 releases.
  - Added a Windows service mode that reports startup and shutdown status more reliably.
  - Added a configuration validation mode to identify attempted configuration files and report issues without terminating early.
  - Improved WASM build coverage for supported mesh interfaces.

- **Bug Fixes**
  - Fixed Windows service startup and shutdown race conditions, improving reliability when managed by Windows Service Control Manager.


bin/build-winget-package.ps1
```diff
@@ -0,0 +1,292 @@
...
+
+<#
+.SYNOPSIS
+    Builds the meshtasticd MSI and the winget manifests that install it.
+
+.DESCRIPTION
+    Writes the MSI and the three winget manifests into -OutputDirectory. The installer manifest
+    pins the MSI's SHA256, so the file must be published unchanged at -InstallerUrl.
+
+    WiX is required, pinned below v6, which refuses to build without accepting the OSMF EULA:
```
thebentern, line 12: Seeing WiX triggers a lot of dayjob PTSD 😅

caveman99: I think the native service control interface is the real win here ;-)

thebentern:
LGTM overall. We may consider doing an arch specific build for arm64. It's fine for now because Meshtastic is not resource intensive and it should run fine through the x64 emulation layer.

thebentern (approved):

.github/workflows/main_matrix.yml
vidplace7, line 1: Leave the main matrix alone for now please :pray:
Windows is still early and this is actively being redone in #11151 

We can add it to the main matrix along with MacOS as a part of that (very soon).

vidplace7 (changes requested):

.github/workflows/main_matrix.yml
caveman99: This was incomplete anyway, the next step would be activating the Winget plumbings. :-)

vidplace7 (approved):

vidplace7:
> LGTM overall. We may consider doing an arch specific build for arm64. It's fine for now because Meshtastic is not resource intensive and it should run fine through the x64 emulation layer.

image
As far as I can see, we actually *can't* ship an arm64 build without switching msys2 toolchains (from gcc to clang/llvm)

PR #11291, opened 2026-07-30
Fix `millis()` rollover in deadline, interval, and timestamp handling

NomDeTom:
Every 49.7 days the 32-bit `millis()` counter wraps, and this firmware trusted it in four load-bearing ways. Comparisons made directly against it (`millis() > deadline`, `deadline < millis()`) invert while the deadline sits across the wrap, so one-shot actions fire immediately or stall for days — the nRF52 flash-corruption backoff among them. `getTime()` measured elapsed-since-time-set as a 32-bit delta, so a node up longer than one wrap reported a wall clock 49.7 days in the past, and every `last_heard`, `rx_time`, message and position stamp inherited it. The queued-packet `rx_time` reconciliation aliased past one wrap, backdating old placeholders to plausible-but-wrong recent epochs. And a node first heard before the clock was trusted never received a `last_heard` at all: the phone showed "Last heard: unknown" for a node it had just announced.

Deadline and interval checks now go through `Throttle` — `deadlinePassed()`/`deadlinePassedAt()` where a site stores an absolute deadline, `hasElapsed()` (the complement of `isWithinTimespanMs()`) where it stores the last event — with disarmed-sentinel values (`0`, `UINT32_MAX`) tested before the arithmetic, and a `millis-deadline-check` CI job plus allowlist keeping naive compares out of `src/`. Timestamps ride on `Time::getMillisMonotonic()`, a 64-bit wrap-counting read whose carry is advanced by exactly one writer — `Time::serviceMonotonic()`, called at the top of the main `loop()` — while every reader derives its answer from that published snapshot plus the unsigned elapsed time since it, which is exact across the wrap; readers never inspect the boundary and never write back, so no number of concurrent callers can miscount a wrap. It replaces three duplicate private wrap counters (`AirTime`, `DeviceTelemetry`, and a dead pair in `HostMetrics`) and is documented as not ISR-safe — `getMillis()` remains the ISR-safe read. An earlier commit here removes `getMillis64()`, which was the same shape read lazily by a single caller and so could miss a wrap outright; the restored clock differs precisely in that its publish is guaranteed. On that base, `getTime()` anchors in 64-bit monotonic milliseconds, so the wall clock is exact at any uptime; the untrusted-clock `rx_time` placeholder is monotonic uptime seconds, so reconciliation is exact at any age and a leaked placeholder needs ~50 years of uptime to pass for an epoch instead of 18.3 days; and nodes heard before time arrives keep their arrival instant in a RAM-only sidecar that is backfilled into `last_heard` as a real epoch once the clock becomes trusted — `last_heard` never persists anything but a real epoch or 0, and PhoneAPI re-reads it at nodeinfo send time so handshake ordering doesn't decide what the phone sees.

Two prior PRs are adopted as the baseline with authorship preserved: #10227 (NextHopRouter retransmission rollover; its half-range compare now lives in `Throttle::deadlinePassedAt()`, credited at the call site) and #10582 (AirTime monotonic windows, which now read the shared clock). `Throttle` and both clocks are injectable through the `UptimeClock` test seam; the native suites drive the wrap boundary, the wall clock across it, and the `last_heard` backfill directly. The conventions are documented in `.github/copilot-instructions.md` and mirrored to `AGENTS.md` and `CLAUDE.md`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timer and deadline handling across millisecond counter rollover, reducing premature or delayed actions.
  * Improved power scheduling, screen timeouts, notifications, pairing windows, LED behavior, and network retries.
  * Improved airtime and utilization reporting after sleep or delayed processing.
  * Improved timestamp handling before and after the device clock becomes trusted.
  * Corrected uptime tracking and GPS fix-hold behavior across clock rollover.

* **Tests**
  * Added coverage for timer rollover, uptime tracking, airtime windows, deadlines, GPS fix holds, and timestamp reconciliation.


NomDeTom:
@RCGV1 here's my time PR

.github/workflows/test_native.yml
```diff
@@ -54,6 +54,70 @@ jobs:
...
+            xargs -0 awk '
+              {
+                line = $0
+                sub(/\/\/.*/, "", line)
+                if (line ~ /(millis\(\)[ \t]*[<>]=?)|([<>]=?[ \t]*millis\(\))/) {
+                  code = line
+                  sub(/^[ \t]+/, "", code); sub(/[ \t]+$/, "", code)
+                  printf "%s\t%s\t%s\n", FILENAME, FNR, code
+                }
+              }' > /tmp/millis-hits.tsv
```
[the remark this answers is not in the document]
NomDeTom: Tom says to say "Clod here with a response"

Fix incoming

src/mesh/NextHopRouter.cpp
```diff
@@ -405,8 +405,16 @@ int32_t NextHopRouter::doRetransmissions()
...
+        // Use unsigned half-range comparison so retransmission timing stays correct across the
+        // ~49.7 day millis() wraparound (previously this FIXME would stall all retx for the
+        // duration of the wrap or fire them all at once immediately after).
+        //
+        // Casting an unsigned difference to int32_t for a "time passed" test is
+        // implementation-defined in C++ when the value exceeds INT32_MAX. The unsigned
+        // half-range form below is fully well-defined: nextTxMsec is in the past (or is now)
+        // iff (now - nextTxMsec) has not wrapped past 2^31 ms. Anything further in the
+        // future wraps into the top half and reads as "not yet."
+        if ((uint32_t)(now - p.nextTxMsec) < 0x80000000u) {
```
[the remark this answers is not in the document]
NomDeTom: Tom says to say "Clod here with a response"

Fix incoming

src/modules/Telemetry/Sensor/BME680Sensor.cpp
```diff
@@ -163,10 +164,12 @@ void BME680Sensor::updateState()
         }
     } else {
         /* Update every STATE_SAVE_PERIOD minutes */
-        if ((stateUpdateCounter * STATE_SAVE_PERIOD) < millis()) {
+        // Interval since the last save; counter * period overflows uint32 past ~198 saves.
+        if (Throttle::hasElapsed(lastStateSaveMs, STATE_SAVE_PERIOD)) {
             LOG_DEBUG("%s state update every %d minutes", sensorName, STATE_SAVE_PERIOD / 60000);
             update = true;
             stateUpdateCounter++;
+            lastStateSaveMs = millis();
```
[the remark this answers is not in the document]
NomDeTom: Tom says to say "Clod here with a response"

Fix incoming

test/test_packet_signing/test_main.cpp
```diff
@@ -1214,14 +1214,18 @@ void test_C3_invalid_repeated_packet_cannot_ack_or_change_retry_state(void)
     prior.hop_start = 2;
     prior.transport_mechanism = meshtastic_MeshPacket_TransportMechanism_TRANSPORT_LORA;
     pipelineRouter->remember(&prior);
-    pipelineRouter->addPending(prior, UINT32_MAX);
+    // "Far future, so no retransmission is due." Must be a representable future time, not
+    // UINT32_MAX: doRetransmissions() compares with an unsigned half-range test, under which
+    // UINT32_MAX is ~1ms in the *past* and would fire a retransmit and rewrite nextTxMsec.
+    const uint32_t notDueTxMsec = millis() + 3600000UL;
```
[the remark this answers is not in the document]
NomDeTom: Tom says to say "Clod here with a response"

Fix incoming

test/test_throttle/test_main.cpp
```diff
@@ -0,0 +1,220 @@
...
+    const uint32_t deadline = 0xFFFFFF00u + 500;
+
+    TEST_ASSERT_FALSE(Throttle::deadlinePassed(deadline)); // not yet
+    Time::advanceTestMillis(400);                          // 0x00000094 - wrapped, still not due
+    TEST_ASSERT_FALSE(Throttle::deadlinePassed(deadline));
+    Time::advanceTestMillis(100); // exactly due, past the wrap
+    TEST_ASSERT_TRUE(Throttle::deadlinePassed(deadline));
+    Time::advanceTestMillis(60000); // stays passed
+    TEST_ASSERT_TRUE(Throttle::deadlinePassed(deadline));
+}
```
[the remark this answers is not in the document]
NomDeTom: Tom says to say "Clod here with a response"

Fix incoming

src/modules/Telemetry/Sensor/BME680Sensor.cpp
```diff
@@ -163,10 +164,12 @@ void BME680Sensor::updateState()
         }
     } else {
         /* Update every STATE_SAVE_PERIOD minutes */
-        if ((stateUpdateCounter * STATE_SAVE_PERIOD) < millis()) {
+        // Interval since the last save; counter * period overflows uint32 past ~198 saves.
+        if (Throttle::hasElapsed(lastStateSaveMs, STATE_SAVE_PERIOD)) {
             LOG_DEBUG("%s state update every %d minutes", sensorName, STATE_SAVE_PERIOD / 60000);
             update = true;
```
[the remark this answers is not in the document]
NomDeTom: Tom says to say "Clod here with a response"

Fix incoming

src/airtime.cpp
```diff
@@ -33,47 +38,117 @@ void AirTime::logAirtime(reportTypes reportType, uint32_t airtime_ms)
...
+    uint32_t elapsedUtilTXPeriods = (this->secSinceBoot / 60) - (oldSecSinceBoot / 60);
+    if (elapsedUtilTXPeriods >= MINUTES_IN_HOUR) {
+        memset(this->utilizationTX, 0, sizeof(this->utilizationTX));
+    } else {
+        for (uint32_t i = 1; i <= elapsedUtilTXPeriods; i++) {
+            this->utilizationTX[((oldSecSinceBoot / 60) + i) % MINUTES_IN_HOUR] = 0;
+        }
     }
+    this->lastUtilPeriodTX = this->getPeriodUtilHour();
 }
```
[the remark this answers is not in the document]
NomDeTom: Tom says to say "Clod here with a response"

Declining this one: there is no preemption here to protect against.

Every `airTime->` consumer runs in the single cooperative loop. The telemetry, position, router and UI callers are all `OSThread`s, and on ESP32 the web server is `WebServerThread : concurrency::OSThread` (`src/mesh/http/WebServer.h:13`) - so `ContentHandler.cpp`s `airtimeReport()` / `channelUtilizationPercent()` reads are not on a separate task. The only `xTaskCreate` in `src/mesh/http/` is the SSL-certificate task, which never touches airtime. NimBLE callbacks do run in their own FreeRTOS task, but nothing on that path reaches `AirTime` either.

So `syncNow()`s read-modify-write of `secSinceBoot` / `lastSyncMsec` and the pointer returned by `airtimeReport()` are only ever touched from one context. Adding a `concurrency::Lock` would put a mutex in the RX/TX logging hot path (`logAirtime()` is called per packet) for no benefit, and this file is on the light-sleep wake path where the existing lock ordering is already delicate.

Worth noting the one real hazard in this area, which is not concurrency: `syncNow()` is now reachable from read paths, and its `firstTime` branch calls `getPeriodUtilMinute()` / `getPeriodUtilHour()` / `currentPeriodIndex()` before clearing `firstTime`. Those three helpers were changed in this PR to read `secSinceBoot` directly; had they kept calling `getSecondsSinceBoot()` - which now calls `syncNow()` - that would be unbounded recursion on the first packet.

src/motion/MotionSensor.cpp
```diff
@@ -259,8 +259,11 @@ void MotionSensor::drawFrameCalibration(OLEDDisplay *display, OLEDDisplayUiState
     const uint32_t now = millis();
     const uint32_t endCalibrationAt = screen->getEndCalibration();
     uint32_t timeRemaining = 0;
-    if (endCalibrationAt > now) {
-        timeRemaining = (endCalibrationAt - now + 999) / 1000;
+    // Signed delta, as in ::runOnce(): this needs the remaining magnitude, not just whether the
+    // deadline passed, so it cannot use Throttle::deadlinePassed().
+    const int32_t remainingMs = (int32_t)(endCalibrationAt - now);
+    if (remainingMs > 0) {
+        timeRemaining = ((uint32_t)remainingMs + 999) / 1000;
```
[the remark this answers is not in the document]
NomDeTom: Tom says to say "Clod here with a response"

Declining this one: the `0` case is not reachable from this frame.

`drawFrameCalibration` is only installed as an alert frame by `beginCalibrationDisplay()`, which runs while a calibration window is open. When the window ends, `finishCalibrationIfExpired()` calls `screen->setEndCalibration(0)` and `screen->endAlert()` as consecutive statements in the same cooperative thread, so no frame can be rendered while the screens value is `0` - there is no yield between them for the screen thread to interleave.

The signed-cast observation is fair in general, but the site that actually gates behaviour is `MotionSensor.cpp:150` in `finishCalibrationIfExpired()`: `if ((int32_t)(now - endCalibrationAt) < 0) return;`, reached from four sensor drivers. Converting that one needs an `endCalibrationAt != 0` guard first, since `Throttle::deadlinePassed(0)` reads as passed for the first half of each wrap cycle - which would end a calibration that never started. Left as-is here rather than half-converting a path this PR does not otherwise touch.

src/motion/MotionSensor.cpp
```diff
@@ -259,8 +259,11 @@ void MotionSensor::drawFrameCalibration(OLEDDisplay *display, OLEDDisplayUiState
...
     }
```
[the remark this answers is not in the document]
NomDeTom: Tom says to say "Clod here with a response"

Declining this one, same reasoning as the CodeRabbit comment on the line above.

On reachability: `drawFrameCalibration` is only installed as an alert frame while a calibration window is open (`beginCalibrationDisplay()`), and `finishCalibrationIfExpired()` clears the value with `screen->setEndCalibration(0)` immediately followed by `screen->endAlert()` in the same cooperative thread, so the deadline is never `0` while this frame is drawn.

On the cast: the conversion is implementation-defined only before C++20, and GCC defines it as modular wrap, so the behaviour is deterministic on this toolchain. It is also computing a magnitude for display, not a passed/not-passed verdict, which is why `Throttle::deadlinePassed()` is not a drop-in here - the unsigned half-range form would need the delta computed separately anyway.

The site worth converting is `MotionSensor.cpp:150`, which does gate behaviour and which neither review flagged; it needs an `endCalibrationAt != 0` guard first, since `deadlinePassed(0)` reads as passed for the first half of each wrap cycle.

NomDeTom:
@coderabbitai can you check if I've resolved your comments?

src/gps/GPS.cpp
```diff
@@ -1426,6 +1429,35 @@ void GPS::publishUpdate()
...
+/// you reach for silently picks one. That is how the re-arm site below broke: `fixHoldEnds != 0 &&`
+/// looks like the sentinel rule from AGENTS.md, but there it means "a hold that was never armed can
+/// never expire", so nothing re-armed and the receiver stayed powered until the search timeout.
+///
+/// The `!= 0` test is not redundant with the arithmetic, though it looks it. deadlinePassed() is an
+/// unsigned half-range test: past 2^31 ms of uptime `now - threadIntervalMs` lands in the top half,
+/// so the sentinel reads as a deadline ~24.9 days in the *future* rather than in the past.
+///
+/// Not static, and deliberately without a header: it lives beside its only caller, and the native
+/// test build compiles this file, so test/test_gps_fix_hold/ declares the two prototypes directly.
```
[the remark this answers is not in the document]
NomDeTom: booo! it's a difficult one - it's taken me and Clod 3 rounds to get this right, and one of us is a computer.

test/test_gps_fix_hold/test_main.cpp
```diff
@@ -0,0 +1,219 @@
...
+    Time::setTestMillis(50 * 1000);
+
+    TEST_ASSERT_FALSE_MESSAGE(fixHoldInForce(0, kThreadInterval), "a hold that was never armed is not in force");
+    TEST_ASSERT_TRUE_MESSAGE(shouldArmFixHold(true, 3, 0, kThreadInterval), "...so it is a reason to arm one");
+    TEST_ASSERT_FALSE_MESSAGE(holdJustExpired(0), "...but not a reason to publish and sleep");
+}
+
+// The sentinel guard inside holdJustExpired() is load-bearing on every cycle, not just past the
+// half-range: fixHoldInForce() already reports an unarmed hold as "not in force", so negating it
+// alone reads as "expired" and would call down() on a receiver that was never holding.
```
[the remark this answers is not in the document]
NomDeTom: Trimmed to two lines in the next push - kept the "on every cycle, not just past the half-range" point, since that is the part that is not obvious from the assertion. Thanks!

AGENTS.md
```diff
@@ -81,7 +81,16 @@ Key rotation to never trigger casually: only the **full** factory reset (`factor
...
+  - `Throttle::deadlinePassedAt(nowMs, deadlineMs)` - the same test against a caller-supplied `now`, for a loop that snapshots the clock once and tests many deadlines. Snapshot from `Time::getMillis()`.
+
+  Raw `millis() > deadline` or `deadline < millis()` is rollover-unsafe: the comparison inverts while the deadline sits on the far side of the 32-bit wrap, so the action fires immediately or blocks for roughly the interval it should have waited. All five helpers subtract first, so unsigned wraparound cancels out. `Throttle` reads the clock through `Time::getMillis()` (`src/UptimeClock.h`), so all ~94 of its call sites are time-injectable and a native test can drive the wrap with `Time::setTestMillis()`.
+
+  **Sentinel hazard.** If a deadline variable also encodes "inactive" (`0` for `rebootAtMsec`, `shutdownAtMsec`, `alertBannerUntil`, `fixHoldEnds`; `UINT32_MAX` for `nagCycleCutoff`), test that sentinel _before_ the elapsed comparison - every such value is arithmetically far in the past, so a correct comparison fires on it immediately. Write `if (deadline && Throttle::deadlinePassed(deadline))`.
```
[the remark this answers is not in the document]
NomDeTom: Good catch - AGENTS.md now spells the two conventions separately: `if (deadline && ...)` for the `0` family, and `deadline != UINT32_MAX` (or the `isNagging` armed flag) for `nagCycleCutoff`. Thanks!

src/modules/Telemetry/Sensor/BME680Sensor.cpp
```diff
@@ -163,7 +165,8 @@ void BME680Sensor::updateState()
...
             stateUpdateCounter++;
```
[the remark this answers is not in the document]
NomDeTom: The timestamp half is fixed: `lastStateSaveMs` is now stamped only inside the successful-write branch, so a failed write no longer checkpoints. The `stateUpdateCounter` increment and verifying the full blob length both predate this PR and are not rollover-related, so I would rather leave those to a separate change than widen this one.

src/modules/NodeInfoModule.cpp
```diff
@@ -35,16 +35,15 @@ bool NodeInfoModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mes
     if (mp.decoded.want_response && !isFromUs(&mp)) {
         const NodeNum sender = getFrom(&mp);
         // A local dedup window, not a wall-clock reading - uptime avoids RTC-quality jumps and
-        // replayed packets' stale rx_time perturbing it.
-        const uint32_t now = (uint32_t)(Time::getMillis64() / 1000);
+        // replayed packets' stale rx_time perturbing it. Seconds, not millis: entries live as long
+        // as the node stays in the DB, and a 32-bit millisecond stamp aliases back into the window
+        // once uptime passes 49.7 days, silently suppressing a legitimate reply for up to 12h.
```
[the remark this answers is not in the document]
NomDeTom: Condensed - it keeps the rollover reason (seconds, not millis, because entries outlive a 32-bit ms stamp) and drops the secondary detail. I have deliberately left it at a few lines: the aliasing failure is silent and this is the only place that records it.

NomDeTom:
@ianmcorvidae would you mind casting an eye over this if you get time (pun intended)?

I've tried to keep it focused on solving a few outstanding issues, and not drift off into refactor for the sake of it.

The thread-safety in particular is something I'm not confident I know what it should look like.

mcenderdragon:
out of curiousity, libs that use millis() as a timeout are also effected by this right ? eg: [Adafruit](https://github.com/adafruit/Adafruit_BME280_Library/blob/master/Adafruit_BME280.cpp#L287) used from https://github.com/meshtastic/firmware/blob/develop/src/modules/Telemetry/Sensor/BME280Sensor.cpp#L38

NomDeTom:
> out of curiousity, libs that use millis() as a timeout are also effected by this right ? eg: [Adafruit](https://github.com/adafruit/Adafruit_BME280_Library/blob/master/Adafruit_BME280.cpp#L287) used from https://github.com/meshtastic/firmware/blob/develop/src/modules/Telemetry/Sensor/BME280Sensor.cpp#L38

Possbily - I'm not intending to delve that deep into the hinterland. Your issue with the 280 might be related - the 680 was definitely affected.

test/test_uptime_clock/test_main.cpp
```diff
@@ -0,0 +1,356 @@
...
+    std::vector<std::thread> readers;
+    for (int i = 0; i < 4; i++) {
+        readers.emplace_back([&stop]() {
+            while (!stop.load(std::memory_order_relaxed))
+                (void)getTime(false); // what the BLE / web-server threads actually call
+        });
+    }
+
+    advanceAndService(0x800u);      // cross the wrap while the readers are running
+    advanceAndService(60u * 1000u); // and some ordinary time after it
```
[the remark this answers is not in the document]
NomDeTom: Clod Here — Tom says: Fixed, thank you — the arithmetic was exactly right. Rather than relocate the comment I moved the start to `0xFFFFF800`, so the first advance now lands on the wrap while the readers are running and the second is genuinely the time after it; total elapsed is unchanged, so the closing assertion still holds. Full native suite is GREEN 47/47 (872 cases); the two comment-length nitpicks are noted and not yet done.

caveman99 (changes requested):
Sensible approach. But for the love of god, keep anecdotal documentation and source code comments to a minimum.

thebentern (approved):

src/gps/GPS.cpp
```diff
@@ -1428,6 +1431,44 @@ void GPS::publishUpdate()
     }
 }
 
+/// Is a post-lock ephemeris hold currently in force?
```
thebentern: Condense the book

src/mesh/eth/ethClient.cpp
```diff
@@ -196,7 +197,10 @@ static int32_t reconnectETH()
     }
 
 #ifndef DISABLE_NTP
-    if (isEthernetAvailable() && (ntp_renew < millis())) {
+    // 0 here means "renew now" (forced at link-up). deadlinePassed(0) only reads as passed for the
```
thebentern, line 200: Poem to haiku

src/mesh/NodeDB.h
```diff
@@ -241,6 +241,19 @@ enum LoadFileResult {
 
 enum UserLicenseStatus { NotKnown, NotLicensed, Licensed };
 
+/**
```
thebentern: Readers digest version

src/mesh/eth/ethClient.cpp
NomDeTom: That comment is 2 lines, plus a TODO because I don't want to re-engineer all the cases where it's not ideal.
I'm condensing the bugged form into the current not-awful form.
Not-awful-to-bulletproof is for another PR.

src/mesh/NodeDB.h
NomDeTom: I can save 2 lines by removing the borders, but it's a fairly obvious footgun if this bit is messed with.

src/gps/GPS.cpp
NomDeTom: This could be more condensed, but Todd spent a long time trying to get this right and I don't want to undo it.

Next PR will condense further for the special cases, I promise.

NomDeTom: Fine, I've done it.

src/mesh/NodeDB.h
NomDeTom: removed the borders to save electrons

NomDeTom:
> Sensible approach. But for the love of god, keep anecdotal documentation and source code comments to a minimum.

On it, boss

NomDeTom:
@caveman99 now 21 lines lighter and less storytelling.

src/platform/nrf52/main-nrf52.cpp
```diff
@@ -288,17 +291,19 @@ void preFSBegin()
     if (!(NRF_POWER->RESETREAS == 0 && NRF_POWER->GPREGRET == NRF52_MAGIC_LFS_IS_CORRUPT))
         return;
     NRF_POWER->GPREGRET = 0;
-    millis_until_formatting_again = millis() + MULTIPLE_CORRUPTION_DELAY_MILLIS;
+    last_format_ms = millis();
```
caveman99, line 297: What happens if millis() hypothetically was 0 here at the time of assignment. "0 means nothing formatted yet this boot"

src/airtime.cpp
```diff
@@ -33,47 +38,112 @@ void AirTime::logAirtime(reportTypes reportType, uint32_t airtime_ms)
...
+    return (secSinceBoot / 60) % MINUTES_IN_HOUR;
 }
 
 void AirTime::airtimeRotatePeriod()
 {
+    // Preserve the public helper while keeping all rotation logic in one monotonic-time path.
+    syncNow();
+}
+
+void AirTime::syncNow()
```
caveman99, line 60: The calls this replaces were read only before, now they change secSinceBoot and the buckets. All callers are on the scheduler thread, but if that ever is calles asynchronously it may race. Better add locking.

src/modules/ExternalNotificationModule.cpp
caveman99: Looks like develop head dragged in a new one.

caveman99 (changes requested):

src/modules/ExternalNotificationModule.cpp
caveman99: ... which the test promptly caught!

NomDeTom: The first step towards making progress is not going backwards.

NomDeTom: fixed

src/airtime.cpp
NomDeTom: I'll add a follow-up PR. I know we're relying on luck for now, but better to put it out of this PR for complexity reasons.

src/platform/nrf52/main-nrf52.cpp
NomDeTom: I (clod) put in a bool to check.

thebentern (approved):

NomDeTom:
@thebentern putting this one back a bit - it was tripping CI. Now ignored.

NomDeTom:
I have a better fix for this, but it won't be possible until Wednesday night

PR #11305, opened 2026-07-31
Make board_level = release explicit instead of relying on an empty value

vidplace7:
## What & why

`generate_ci_matrix.py` treated **absence** of `board_level` as "this is a release board". That sentinel is an inheritance footgun: an env's level was decided by whatever it inherited (or failed to inherit) through `extends`, so a variant could silently land in — or drop out of — the release matrix without anything in its `platformio.ini` saying so, and forgetting the key entirely was indistinguishable from deliberately choosing release.

This makes the release level explicit. Every variant env now declares one of `pr`, `release`, or `extra`, and the matrix generator refuses to run if any env is missing the key or uses an unrecognized value.

## What changed

- **`variants/**/platformio.ini`** (97 files, +122 lines): added `board_level = release` to every env that had no level of its own and inherited none. Existing `board_level = pr` and `board_level = extra` values were left untouched. These are pure one-line additions — no other content changed in any variant file.
- **`bin/generate_ci_matrix.py`**:
  - Added a `BOARD_LEVELS` constant documenting the three levels.
  - The release branch now tests `board_level == "release"` instead of `not board_level`.
  - **New**: exits non-zero on a missing or unrecognized `board_level`, naming the offending env. This is the point of the change — a new variant can no longer omit the key and get a level by accident.
  - Collapsed the nested conditional in the `check` path into an equivalent single condition.
- **`.github/copilot-instructions.md`**: documented all three levels and the new hard requirement. Also corrected a stale claim that `custom_meshtastic_support_level` controls PR-vs-merge builds — it does not; nothing reads it for matrix filtering. It is variant metadata that `bin/platformio-custom.py` emits as `supportLevel` in the generated hardware list.

## Behavior

**The generated CI matrix is unchanged.** This is a refactor of how the release level is *spelled*, not which boards get built. The only intended behavior change is the new fail-fast validation, which cannot trigger on this branch precisely because all 249 envs now declare a level.

## Testing

Built two git worktrees — `develop` (before) and this branch (after) — and diffed the generator's output across **14 targets** (all 12 `variants/` platform dirs plus `all` and `check`) × **4 level forms** (bare, `--level pr`, `--level extra`, `--level pr extra`) = **56 invocations per tree, 112 runs**. Each pair was compared both as an exact stdout string and as an order-insensitive board set.

**All 56 pairs are byte-for-byte identical** — not merely same-set-different-order, which would still have reshuffled CI job ordering.

| Invocation | Boards | Result |
| --- | --- | --- |
| `all` | 141 | identical |
| `all --level pr` | 19 | identical |
| `all --level extra` | 249 | identical |
| `all --level pr extra` | 127 | identical |
| `check` | 80 | identical |
| `check --level pr` | 9 | identical |

`check` = 80 and `check --level extra` = 80 both match, which specifically confirms the rewritten `check` path still keeps `extra`-level `board_check` envs in the non-PR check matrix.

The new validation was verified separately by running the updated script against `develop`'s un-migrated variants, where it correctly fails:

```
Error: environment 'rp2040-lora' has no value for 'board_level'; expected one of: pr, release, extra
Error: environment 'rp2040-lora' has 'Release' for 'board_level'; expected one of: pr, release, extra
```

Also confirmed PlatformIO already strips trailing whitespace and inline `;` comments from the value, so `board_level = release ; note` parses fine and needs no special handling.

## Reviewer notes

- Two envs are easy to miss when sweeping this file set: `t5s3-epaper-v1` and `t5s3-epaper-v2` in `variants/esp32s3/t5s3_epaper/platformio.ini`. Their section headers carry trailing comments (`[env:t5s3-epaper-v1] ; H752`), which defeats a naive `^\[env:...\]$` match. Both are included here; the post-change audit confirms all **249** envs resolve to exactly one level (122 `release` / 108 `extra` / 19 `pr`), zero missing.
- Nothing under `src/` or `test/` is touched — there is no runtime/firmware change, so there is no device-level regression surface.
- `--level` still accepts only `pr` and `extra`. Adding `release` as a choice would be redundant (the bare form already means "pr + release") and ambiguous in combinations like `--level pr release`.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

On-device testing is not applicable: this changes only CI matrix metadata and the matrix generator. No firmware source is modified, and the produced build matrix is byte-identical to `develop`, so every board builds exactly as it did before.

## Summary by CodeRabbit

* **CI Improvements**
  * Standardized hardware classification into PR, release, and extra levels.
  * CI matrix generation now validates classifications and applies clearer default filtering.
  * Release-level hardware is now included in appropriate builds.

* **Documentation**
  * Clarified required hardware classification settings and how metadata is used.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -557,8 +557,9 @@ pio run -e native-macos       # Build headless macOS meshtasticd (Homebrew prere
 1. Create directory under `variants/<arch>/<name>/`
 2. Add `variant.h` with pin definitions and hardware capability defines
 3. Add `platformio.ini` with build config - use `extends` to reference common base (e.g., `esp32s3_base`)
-4. Set `custom_meshtastic_support_level = 1` (PR builds) or `2` (merge builds)
-5. For e-ink displays, add `nicheGraphics.h` for InkHUD configuration
+4. Set `board_level` (required - `release` for a normal variant; see "Build Matrix Generation")
+5. Set `custom_meshtastic_support_level` (1-3) and the other `custom_meshtastic_*` metadata
+6. For e-ink displays, add `nicheGraphics.h` for InkHUD configuration
 
 ### Adding a New Telemetry Sensor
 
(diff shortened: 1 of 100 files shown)


NomDeTom (approved):
looks good to me

PR #11317, opened 2026-08-01
fix(radio): validate 2.4 GHz bandwidths

RCGV1:
## Summary

- validate effective LoRa bandwidths against the selected SX128x, LR11x0, or LR20x0 radio before calling RadioLib
- accept only the exact bandwidth values each radio driver supports, including LR2021 modes through 1000 kHz
- repair invalid local or persisted settings to a radio-compatible default
- keep mixed-radio probes transactional: suppress speculative errors, restore configuration, and quiesce/reject candidates whose restored configuration cannot be applied

## Why

A 2.4 GHz node could receive a sub-GHz bandwidth such as 125 kHz. Generic validation accepted and persisted it, then RadioLib rejected the unsupported bandwidth. Firmware recorded `INVALID_RADIO_SETTING` (critical error 7) but continued reconfiguration and reported success, leaving the saved protobuf inconsistent with the radio's retained bandwidth state. Display-equipped nodes force the critical-fault frame into focus, which appears to the user as a frozen node even though the scheduler and serial API remain responsive.

Mixed T3S3 environments also need validation to follow the radio actually populated on the board rather than the shared PlatformIO target.

## Validation

- `pio run -e tlora-t3s3-v1 -e muzi-base -e meshnology_w12`: passed for the mixed SX128x/LR1121 target, Muzi LR1121, and LR2021 production target
- `trunk fmt`, `trunk check`, and `git diff --check`: passed
- pre-fix real Muzi Base/LR1121 (`2.8.0.d0e9d02`): custom LORA_24 125 kHz was accepted and repeatedly read back; RadioLib recorded critical error 7 at `LR11x0Interface.cpp:183`; serial/API remained responsive for 20 seconds
- pre-fix real TLORA_T3_S3/SX128x (`5bfad255e`): custom LORA_24 125 kHz was accepted and repeatedly read back; RadioLib recorded critical error 7 at `SX128xInterface.cpp:127`; every serial/API probe passed for 20 seconds
- source trace confirmed critical error 7 focuses the display's `Critical fault #7` frame without stopping the scheduler/API, explaining the reported apparent freeze
- real TLORA_T3_S3/SX128x on final commit `88e643e3b`: custom LORA_24 1600 kHz applied, survived reboot, and read back as 1600
- real TLORA_T3_S3/SX128x: invalid custom 125 kHz was corrected to 800 and the node remained responsive
- real TLORA_T3_S3/SX128x: 200, 400, 800, and 1600 kHz applied and read back successfully
- exact original LORA_24/SHORT_FAST configuration and node identity were restored and verified after reboot
- focused native test compilation is blocked locally on macOS before test execution by existing initialized VLA errors in `NotificationRenderer.cpp`; the added radio tests are included for Linux CI
- LR2021 is build- and unit-tested only; no LR2021 hardware was available on this bench

## Attestations

- [x] I have tested that my proposed changes do not cause obvious regressions on:
  - [x] Other: LilyGo TLORA T3S3 with SX128x

## Summary by CodeRabbit

* **New Features**
  * Added hardware-aware LoRa bandwidth validation for supported radio platforms.
  * Automatically adjusts unsupported bandwidth settings to suitable regional defaults.
  * Improved handling of wide-band and turbo radio configurations.

* **Bug Fixes**
  * Prevented radios from restarting reception after unsuccessful configuration.
  * Preserved valid regional settings during radio detection and configuration.
  * Reduced misleading errors during automatic radio probing.

* **Tests**
  * Added coverage for platform-specific bandwidth validation, correction, probing, and configuration scenarios.


NomDeTom:
should this also cover LR2021?

src/mesh/RadioInterface.cpp
```diff
@@ -581,10 +582,16 @@ std::unique_ptr<RadioInterface> initLoRa()
...
+        rIf->setConfigErrorReporting(false);
         if (!rIf->init()) {
             LOG_WARN("No LR1120 radio");
+            config.lora = loraConfigBeforeProbe;
             rIf = nullptr;
         } else {
+            config.lora = loraConfigBeforeProbe;
+            rIf->setConfigErrorReporting(true);
+            if (!rIf->reconfigure())
+                LOG_WARN("LR1120 reconfigure failed after probe");
```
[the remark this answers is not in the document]
RCGV1: Completed in b6d2ed26e. LR1120/LR1121/LR2021 now reject a candidate when the restored configuration cannot be applied, quiesce it through the radio sleep/IRQ cleanup path, and restore the pre-probe configuration again before fallback. The concrete LR11x0/LR20x0 reconfigure methods now propagate RadioLib setup failures and do not restart RX on failure. Final T3S3 hardware validation passed at this exact commit.

RCGV1:
@NomDeTom Yes. LR2021 now uses the full exact bandwidth set accepted by the bundled RadioLib driver, through 1000 kHz, while rejecting 1600 kHz and lossy near-value aliases. The `meshnology_w12` LR2021 production target builds successfully, and tests cover every accepted wire code plus invalid aliases.

The Portduino initialization path remains intentionally unchanged: it creates one explicitly configured radio and exits if that radio fails, so its validation result is not speculative fallback state. Suppressing/restoring that path would hide genuine invalid configuration on the selected Portduino radio.

RCGV1:
@coderabbitai review

Final implementation and hardware validation are complete at `88e643e3b`. The remaining `check-label` failure is administrative: this account cannot add repository labels. A maintainer will need to apply the `bugfix` label.

NomDeTom:
The `meshnology_w12` LR2021 production target builds successfully, and tests cover every accepted wire code plus invalid aliases.
> 
> The Portduino initialization path remains intentionally unchanged: it creates one explicitly configured radio and exits if that radio fails, so its validation result is not speculative fallback state. Suppressing/restoring that path would hide genuine invalid configuration on the selected Portduino radio.

I'm not hearing hardware confirmation.

RCGV1:
You are right: I do not have LR2021 hardware confirmation. The attached bench covers a T3S3 with SX128x and Muzi/T-Beam devices with LR1121; LR2021 is build- and unit-tested only.

The author of #11279 does have dual-band LR2021 hardware, so I opened a focused companion contribution against that branch for the relevant band-switch guards and tests: https://github.com/Ethan-chen1234-zy/firmware/pull/1. I will keep #11317’s hardware claim limited to SX128x/LR1121 unless the LR2021 owner reports a successful live matrix.

RCGV1:
Hardware reproduction follow-up for the original failure mode:

- Pre-fix Muzi Base/LR1121 and TLORA_T3_S3/SX128x both accepted and persisted custom `LORA_24` bandwidth `125`.
- RadioLib rejected the unsupported bandwidth and firmware recorded `INVALID_RADIO_SETTING` (critical error 7) at the radio-specific bandwidth call.
- Both nodes continued answering serial configuration reads throughout a 20-second probe window.
- Source tracing shows critical error 7 forces the display's `Critical fault #7` frame into focus. That explains the reported apparent freeze; the scheduler/API did not stop.
- Because RadioLib rejects the value before updating its internal bandwidth, the persisted protobuf can disagree with the radio's retained bandwidth state.

On the PR firmware, the same invalid 125 kHz setting is corrected to 800 before reaching RadioLib, with no critical error, and the T3S3 remains responsive. LR2021 remains build- and unit-tested only because no LR2021 hardware was available.

src/mesh/LR11x0Interface.cpp
```diff
@@ -210,9 +222,10 @@ template <typename T> bool LR11x0Interface<T>::reconfigure()
     if (err != RADIOLIB_ERR_NONE)
         LOG_WARN("LR11x0 setRxBoostedGainMode %s%d", radioLibErr, err);
 
-    startReceive(); // restart receiving
+    if (success)
+        startReceive(); // restart receiving
 
-    return true;
+    return success;
```
NomDeTom, line 228: Success bool may be false. What's the cost of a descriptive result?

src/mesh/MeshRadio.h
```diff
@@ -200,6 +200,27 @@ static inline uint16_t clampBandwidthCode(uint16_t bwCode)
     return bwCode;
 }
 
+static inline bool supportsSx128xLoRaBandwidth(float bandwidthKHz, bool wideBand)
+{
```
NomDeTom, line 204: This whole section looks like an anti-pattern to the rest of the file, or at least a horrible typo trap, even for an llm

src/mesh/MeshRadio.h
```diff
@@ -200,6 +200,27 @@ static inline uint16_t clampBandwidthCode(uint16_t bwCode)
     return bwCode;
 }
 
+static inline bool supportsSx128xLoRaBandwidth(float bandwidthKHz, bool wideBand)
```
thebentern, line 203: Use the `IS_ONE_OF` macro to clean up these chained comparisons in these 3 methods

src/mesh/RadioInterface.cpp
```diff
@@ -581,12 +582,23 @@ std::unique_ptr<RadioInterface> initLoRa()
...
             LOG_WARN("No LR1120 radio");
+            config.lora = loraConfigBeforeProbe;
             rIf = nullptr;
         } else {
-            LOG_INFO("LR1120 init success");
-            radioType = LR1120_RADIO;
+            config.lora = loraConfigBeforeProbe;
+            rIf->setConfigErrorReporting(true);
+            if (!rIf->reconfigure()) {
+                LOG_WARN("LR1120 reconfigure failed after probe");
```
NomDeTom, line 594: Why are you using a string for a known variable?

src/mesh/RadioInterface.cpp
```diff
@@ -609,12 +632,23 @@ std::unique_ptr<RadioInterface> initLoRa()
     if (!rIf) {
         rIf = std::unique_ptr<LR2021Interface>(
             new LR2021Interface(loraHal, LR2021_SPI_NSS_PIN, LR2021_IRQ_PIN, LR2021_NRESET_PIN, LR2021_BUSY_PIN));
+        rIf->setConfigErrorReporting(false);
         if (!rIf->init()) {
             LOG_WARN("No LR2021 radio");
+            config.lora = loraConfigBeforeProbe;
             rIf = nullptr;
         } else {
-            LOG_INFO("LR2021 init success");
```
NomDeTom, line 616: Strings for known variables. Tell sol to think harder.

NomDeTom (commented):
Get sol to think about flash usage growth. So much laziness.

src/mesh/RadioInterface.h
```diff
@@ -279,6 +287,8 @@ class RadioInterface
   protected:
     int8_t power = 17; // Set by applyModemConfig()
 
+    bool shouldReportConfigErrors() const { return reportConfigErrors; }
```
NomDeTom, line 290: There may be some value in transmission of errors to the phone.

@RCGV1 is this what you were thinking, or something else?

@thebentern is this a better option than my WebMD for nodes?

jamesarich:
Reviewed this alongside the two app PRs. The diagnosis is right, and I checked the bandwidth tables against vendored RadioLib — SX128x `{203.125, 406.25, 812.5, 1625}`, LR11x0 high `{203.125, 406.25, 812.5}` / low `{62.5, 125, 250, 500}`, and all twelve LR2021 values round-trip correctly through RadioLib's `bw_div2` truncation. Those are exact. The test fixture is clean too — `setUp` news a fresh `TestableRadioInterface`, so the per-test `bandwidthProfile` mutations aren't order-dependent.

My concerns are all about the blast radius of the hardening rather than the tables themselves.

### 1. A failed `reconfigure()` now costs the whole radio, and the failure sources are broader than what this PR validates

In `initLoRa`, an LR1120/LR1121/LR2021 candidate that fails `reconfigure()` is `disable()`d and dropped. But `success` goes false on any RadioLib error from SF, BW, CR **or frequency** — and only bandwidth is validated.

`override_frequency` is the gap that worries me. It's set in `AdminModule` (`config.lora.override_frequency = p.frequency`) with no validation against the chip's tuning range, and `checkOrClampConfigLora` only branches on whether it's zero. A frequency the radio can't tune makes `lora.setFrequency()` error → `success = false` → the candidate is dropped. That surfaces as `CriticalErrorCode_NO_RADIO` in `main.cpp`, so it's diagnosable rather than silent — but the node has no LoRa at all, and because `config.lora = loraConfigBeforeProbe` restores the offending config before nulling `rIf`, the next boot fails identically. Ham-mode users on LR11x0/LR20x0 are the exposed population. Previously that node ran (badly); now it doesn't run.

Suggest either scoping the drop to the mismatch actually being probed for, or validating frequency/SF/CR alongside bandwidth.

### 2. `reportErrors` defaults to `true`, so the speculative callers now record critical errors

`MeshBeaconModule::beaconTxConfigInvalid()` and the beacon validation in `AdminModule` call `validateConfigLora(probe)` with the default. The new bandwidth block does `RECORD_CRITICALERROR + sendErrorNotification`.

`PRESETS_STD` includes SHORT_TURBO / LONG_TURBO / MEDIUM_TURBO and LORA_24 uses `PROFILE_STD`, so all three are valid presets for the region — and all three map to 1625 kHz in wide mode, which LR11x0/LR20x0 reject. On such a node configured to beacon-broadcast a turbo preset, that's critical error 7 on every broadcast cycle. The `reportErrors` plumbing already exists; these speculative sites should pass `false`.

### 3. The critical-fault frame still fires on the case in the description

Pre-fix, `validateConfigLora` *passed* for LORA_24 + 125 kHz (the region spans 2400–2483.5 MHz, trivially wider than 0.125 MHz), so error 7 came only from RadioLib. Post-fix it is recorded by `validateConfigLora` and again by `clampConfigLora` before the repair lands.

The bench note says the node "remained responsive" — but responsiveness wasn't the reported symptom; the focused `Critical fault #7` frame was. Worth confirming on hardware whether it still appears. Since the config is being auto-repaired, a warning-level notification seems more fitting than a critical error.

### 4. Turbo presets are silently substituted, and only one is tested

All three turbo presets map to 1625 kHz in wide mode, so on LR1121/LR2021 all three get rewritten to the region default. Tests cover SHORT_TURBO only — MEDIUM_TURBO and LONG_TURBO have the same mapping.

To be fair this is loud-failure → silent-substitution rather than working → broken, since RadioLib rejected 1625 on these chips before too. But the node now quietly runs a different preset and channel hash than the one it was configured with.

### Nits

- Skipping `startReceive()` on failure applies to live reconfigures, not just probes. It self-heals on the next queued transmit since the event loop re-arms RX, so the cost is a missed RX window rather than permanent deafness — but `SX128xInterface`/`SX126xInterface::reconfigure` still `return true` unconditionally, so the behaviour is now chip-family-dependent. That includes the `initLoRa` tail-end `if (rIf && !rIf->reconfigure())` reboot path, which is live on LR-family boards and unreachable elsewhere.
- The snapshot/restore covers 3 of the probe branches. In practice the sub-GHz branches are guarded by `region != LORA_24` and the snapshot is taken once at the top of `initLoRa`, so a later transactional branch heals earlier damage — but a scope guard applied uniformly would be more obviously correct than reasoning about probe order.
- `supportsLr20x0LoRaBandwidth` discards `wideBand` via `(void)wideBand`. That correctly mirrors RadioLib (`LR2021::setBandwidth` takes no band argument), but it silently accepts 31.25 kHz on LORA_24 — worth a comment so nobody "fixes" it later.
- `useWideModemParameters()` (`region->wideLora || !supportsSubGhz()`) and the driver's `wideLora() && getFreq() > 1000.0f` are two predicates for one decision. They agree only because LORA_24 is currently the sole wide region.
- `computeSlotTimeMsec()` still keys off `myRegion->wideLora` while `applyModemConfig` now uses `useWideModemParameters` for `bw` and `preambleLength`.

### Cross-repo

See my note on the app PRs: the `pio_env` allowlists there statically mirror this repo's build matrix because the protocol doesn't expose the runtime radio chip. Worth settling whether that belongs in `DeviceMetadata` before three copies of the workaround land.

src/mesh/LR11x0Interface.cpp
RCGV1: Addressed in . Reconfigure failures now report the first failing operation plus the RadioLib code to the error log and client notification (for example, ). Probe-time checks remain quiet, and RX restarts only after a successful reconfigure.  covers the client payload.

src/mesh/RadioInterface.cpp
RCGV1: Addressed in : all LR1120/LR1121/LR2021 paths now call the same enum-based transactional probe helper, eliminating the duplicated string identifiers and branch-specific state handling.

src/mesh/RadioInterface.h
RCGV1: Implemented in . The phone receives one compact error containing the failing operation and RadioLib result, while speculative probes/validation log quietly and do not emit client notifications. The repair path is also pre-validated quietly so an auto-clamped config produces only one notification.

src/mesh/MeshRadio.h
RCGV1: Reworked in `defefdc58` to typed `constexpr std::array<float, N>` capability sets plus a single `containsBandwidth` helper. This removes the repeated comparison chains and keeps the exact RadioLib values in one auditable place.

src/mesh/MeshRadio.h
RCGV1: Addressed with a typed helper rather than `IS_ONE_OF`: that macro builds an `int[]`, so using it here would truncate fractional RadioLib bandwidths such as 203.125 and 406.25. `defefdc58` uses `std::array<float, N>` and exact float membership instead.

src/mesh/RadioInterface.cpp
RCGV1: Addressed in `defefdc58`. Probe identity is now carried as `LoRaRadioType`; a centralized `radioTypeName()` mapping is used only at the log boundary. The shared probe helper also owns rollback, reconfigure validation, and final `radioType` assignment.

RCGV1:
Review follow-up is complete in `defefdc58`, after merging current `origin/develop` at `4296d5d58`.

Key changes:
- Typed `std::array<float, N>` bandwidth capability sets replace comparison chains. `IS_ONE_OF` was not used because it constructs an `int[]` and would truncate fractional RadioLib values.
- One enum-based transactional probe helper now handles LR radios and Portduino, restores the original LoRa config, validates reconfigure, and rejects failed candidates.
- Exact radio frequency ranges now validate and clamp unsupported overrides before RadioLib. The applied wide-band decision is shared by validation, modem setup, preamble selection, and slot timing.
- Speculative beacon/WASM/probe validation is quiet. Auto-repair performs one reporting clamp rather than duplicate validation/clamp notifications.
- Reconfigure failures carry the first operation and RadioLib code to logs and the client, and RX restarts only after successful reconfiguration across LR11x0, LR20x0, SX126x, and SX128x.
- Tests cover all three turbo presets on LR1121/LR2021 profiles, frequency override rejection, dual-band wide parameters and slot timing, one-notification repair, and detailed client failure text.
- The LR2021 bandwidth comment documents that the bundled driver exposes the same LoRa bandwidth enum in both RF bands.

Validation:
- `native-macos` `test_admin_radio` + `test_radio`: 126/126 passed.
- `tlora-t3s3-v1`: build passed.
- `muzi-base`: build passed, 766744/815104 bytes, 35 KB clear of the guarded warm region.
- `rak4631`: build passed, 756120/815104 bytes, 45 KB clear of the guarded warm region.
- Exact `origin/develop` RAK4631 baseline: 755392 bytes, so this update adds 728 bytes.
- `trunk fmt`, `trunk check`, and `git diff --check`: passed.

No hardware was touched for this review update. SX128x/LR1121 hardware evidence already recorded on the PR remains valid for the earlier commit; LR2021 behavior remains build- and unit-tested only until someone with LR2021 hardware runs the live matrix.

PR #11347, opened 2026-08-03
Fix stack buffer overflow in aes_ccm_encr for partial blocks

matutetandil:
`aes_ccm_encr()` writes the full 16-byte AES keystream block straight into the output buffer and only afterwards XORs the first `last` bytes in place:

```c
if (last) {
    WPA_PUT_BE16(&a[AES_BLOCK_SIZE - 2], i);
    crypto->aesEncrypt(a, out);      // writes 16 bytes
    /* XOR zero-padded last block */
    for (i = 0; i < last; i++)
        *out++ ^= *in++;
}
```

When the final block is partial, that `aesEncrypt` writes up to 15 bytes past the length the caller asked for.

**This is latent — nothing on `develop` misbehaves today.** Every caller in the tree hands over a buffer with enough slack to absorb the overshoot, the existing tests included, so nothing trips. It is still worth fixing: the overshoot is invisible at the call sites, the decrypt path clears it by only a few bytes, and any future caller that sizes a buffer to the payload gets a silent out-of-bounds write. #9749 adds exactly such a caller, and AddressSanitizer flags it as a `stack-buffer-overflow` in `coverage:test_crypto`.

The fix encrypts into a temporary block and XORs out of it — the same pattern `aes_ccm_encr_auth()` and `aes_ccm_decr_auth()` already use a few lines below in this same file:

```c
if (last) {
    uint8_t tmp[AES_BLOCK_SIZE];
    WPA_PUT_BE16(&a[AES_BLOCK_SIZE - 2], i);
    crypto->aesEncrypt(a, tmp);
    /* XOR zero-padded last block */
    for (i = 0; i < last; i++)
        out[i] = tmp[i] ^ in[i];
}
```

The ciphertext is unchanged.

### Regression test

`test_AES_CCM_partial_block_bounds` drives `aes_ccm_ae()` / `aes_ccm_ad()` at lengths 5 (pure partial block) and 20 (one full block plus a partial one), with guard bytes laid down past the requested length in both the ciphertext and the plaintext buffer.

It asserts on those guard bytes rather than leaning on a sanitizer, so it fails identically under `native` and `coverage`, and the test itself never actually goes out of bounds. Reverting the fix turns it red with `Expected 165 Was 23`.

### Leftovers removed

`encryptCurve25519()` copied `extraNonce` into `auth + 8` both before and after the `aes_ccm_ae()` call, because the call used to trample it. It no longer reaches that far, so the copy before it is gone and the one after does the job on its own. The comment on the call warning that it "can write up to 15 bytes longer than numbytes past bytesOut" no longer describes the code, so it is gone too.

### Verification

- `test_crypto` 12/12 on `native` and on `coverage` (ASan)
- `test_pki_admin_fallback` 6/6 on `coverage`
- `pio run -e native` SUCCESS
- `clang-format` clean

Split out of #9749, where the overflow first surfaced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed encryption for partial data blocks to ensure output is written correctly and consistently.
  * Improved protection against memory boundary issues during partial-block encryption and decryption.

### DIFF
--- a/src/mesh/aes-ccm.cpp
+++ b/src/mesh/aes-ccm.cpp
@@ -111,11 +111,12 @@ static void aes_ccm_encr(size_t L, const uint8_t *in, size_t len, uint8_t *out, 
         in += AES_BLOCK_SIZE;
     }
     if (last) {
+        uint8_t tmp[AES_BLOCK_SIZE];
         WPA_PUT_BE16(&a[AES_BLOCK_SIZE - 2], i);
-        crypto->aesEncrypt(a, out);
+        crypto->aesEncrypt(a, tmp);
         /* XOR zero-padded last block */
         for (i = 0; i < last; i++)
-            *out++ ^= *in++;
+            out[i] = tmp[i] ^ in[i];
     }
 }
 static void aes_ccm_encr_auth(size_t M, const uint8_t *x, uint8_t *a, uint8_t *auth)
--- a/src/mesh/CryptoEngine.cpp
+++ b/src/mesh/CryptoEngine.cpp
@@ -230,8 +230,6 @@ bool CryptoEngine::encryptCurve25519(uint32_t toNode, uint32_t fromNode, meshtas
     if (!HardwareRNG::fill((uint8_t *)&extraNonceTmp, sizeof(extraNonceTmp)))
         CryptRNG.rand((uint8_t *)&extraNonceTmp, sizeof(extraNonceTmp));
     auth = bytesOut + numBytes;
-    memcpy((uint8_t *)(auth + 8), &extraNonceTmp,
-           sizeof(uint32_t)); // do not use dereference on potential non aligned pointers : *extraNonce = extraNonceTmp;
     LOG_DEBUG("Random nonce value: %d", extraNonceTmp);
     if (remotePublic.size == 0) {
         LOG_DEBUG("Node %d or their public_key not found", toNode);
@@ -246,8 +244,7 @@ bool CryptoEngine::encryptCurve25519(uint32_t toNode, uint32_t fromNode, meshtas
     // Calculate the shared secret with the destination node and encrypt
     printBytes("Attempt encrypt with nonce: ", nonce, 13);
     printBytes("Attempt encrypt with shared_key starting with: ", shared_key, 8);
-    aes_ccm_ae(shared_key, 32, nonce, 8, bytes, numBytes, nullptr, 0, bytesOut,
-               auth); // this can write up to 15 bytes longer than numbytes past bytesOut
+    aes_ccm_ae(shared_key, 32, nonce, 8, bytes, numBytes, nullptr, 0, bytesOut, auth);
     memcpy((uint8_t *)(auth + 8), &extraNonceTmp,
            sizeof(uint32_t)); // do not use dereference on potential non aligned pointers : *extraNonce = extraNonceTmp;
     return true;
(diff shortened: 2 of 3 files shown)


src/mesh/aes-ccm.cpp
```diff
@@ -111,11 +111,12 @@ static void aes_ccm_encr(size_t L, const uint8_t *in, size_t len, uint8_t *out,
         in += AES_BLOCK_SIZE;
     }
     if (last) {
+        uint8_t tmp[AES_BLOCK_SIZE];
         WPA_PUT_BE16(&a[AES_BLOCK_SIZE - 2], i);
-        crypto->aesEncrypt(a, out);
+        crypto->aesEncrypt(a, tmp);
         /* XOR zero-padded last block */
         for (i = 0; i < last; i++)
```
caveman99, line 118: This part is fine, it is the same thing the two functions right below already do.

caveman99 (changes requested):
Fix itself looks right. I pulled it on top of develop and ran test_crypto and test_pki_admin_fallback, both green, output unchanged.

Three things before this goes in:

- The description oversells it. Nothing actually breaks on develop as it stands today, the buffers in the tests are big enough. You only get the ASan hit once the extra tests from #9749 are in. Please reword it as a latent bug. Still worth fixing though, I checked the decrypt path and it only just barely stays inside the buffer. A byte less headroom anywhere and it would be a real one.
- Add a small test for it, otherwise someone puts this straight back in a year.
- While you are in there, clean up the leftovers in CryptoEngine.cpp. The comment on line 250 saying it can write 15 bytes past the buffer is now simply wrong. And the extraNonce memcpy on line 233 was only ever there because the old code trampled over it, the one on line 251 already does the job.

matutetandil:
Thanks for pulling it and for checking the decrypt path. All three addressed in a04a26f, on top of a rebase onto current `develop`.

**Description reworded.** You are right, I had it stated too strongly. Nothing on `develop` misbehaves as it stands — every caller hands over a buffer with enough slack, the existing tests included. The description now leads with that, and the ASan hit is framed as something #9749 provokes rather than something happening today. The commit message had the same problem and got the same treatment.

**Regression test added** — `test_AES_CCM_partial_block_bounds`. It drives `aes_ccm_ae()` / `aes_ccm_ad()` at length 5 (pure partial block) and 20 (one full block plus a partial one), with guard bytes past the requested length in both output buffers.

I went with guard bytes rather than an exactly-sized buffer plus ASan, for two reasons: it fails in `native` as well as `coverage`, and the test never actually goes out of bounds itself, so it reports a normal Unity failure instead of aborting the binary. Reverting just the `aes-ccm.cpp` hunk turns it red:

```
test/test_crypto/test_main.cpp:341: test_AES_CCM_partial_block_bounds: Expected 165 Was 23	[FAILED]
```

165 is the `0xA5` guard, 23 is raw keystream — the bytes past `last` never get XOR-ed back, so they are left as whatever `aesEncrypt` put there. Worth noting that with the fix reverted that is the *only* failure in the suite, which is the same conclusion you reached from the other direction.

**Leftovers cleaned.** Both gone:

- the `memcpy` of `extraNonce` before `aes_ccm_ae()` — the one after it is sufficient now that the call stays inside `numBytes`
- the `// this can write up to 15 bytes longer than numbytes past bytesOut` comment

I ran `test_pki_admin_fallback` as well, since that path exercises `encryptCurve25519()`.

- `test_crypto` 12/12 on `native` and on `coverage` (ASan)
- `test_pki_admin_fallback` 6/6 on `coverage`
- `pio run -e native` SUCCESS

caveman99 (approved):
All good, ignore the overzealous nitpick comment from coderabbit. the ask is covered already.

PR #11348, opened 2026-08-03
Honor an explicit -c config path when -s is given

matutetandil:
Reworked per @jp-bennett's and @caveman99's read: `-s` should keep auto-loading off, but an explicit `-c` should still be honored.

In `portduinoSetup()` the simradio flag is the first branch of the chain that also handles config loading:

```c
if (portduino_config.force_simradio == true) {
    portduino_config.lora_module = use_simradio;
} else if (configPath != nullptr) {
    // load the -c path ...
} else if (access("config.yaml", R_OK) == 0) {
    ...
```

Because `-s` short-circuits the whole chain, it swallows the `-c` branch along with the discovery ones. Skipping discovery is the intent; dropping a path the user typed by hand is not, and it happens with no warning — the file exists, the path is right, the daemon starts fine, the settings just never take effect.

This moves the `-s` check to sit *after* the `-c` branch instead of before it:

```c
if (configPath != nullptr) {
    // explicit -c: always parsed, including under -s
} else if (portduino_config.force_simradio) {
    // -s with no -c: the simulator brings its own defaults, so skip config discovery.
} else if (access("config.yaml", R_OK) == 0) {
    ...
```

so an explicit path is always parsed, and only the implicit discovery (`./config.yaml`, `/etc/meshtasticd/config.yaml`) is skipped under a bare `-s`.

The radio override then runs after every config source, because `-c` and the `ConfigDirectory` entries it can pull in are both able to set `Lora.Module`, and `-s` has to win over all of them. Putting it there also fixes `--check` and `--output-yaml`, which used to report the *configured* module rather than the simulated one, since the old override sat behind an early return.

`ConfigDirectory` needs no special case: it follows whatever config was actually loaded, so a bare `-s` leaves it empty and the loop does not run.

### Verified against the built binary

| invocation | config loaded | `Lora.Module` | `MaxNodes` |
|---|---|---|---|
| `-s` (with a `./config.yaml` present) | none | `sim` | 200 (default) |
| `-s -c custom.yaml` | `custom.yaml` | `sim` | 77 (from `-c`) |
| `-c custom.yaml` | `custom.yaml` | `sx1262` | 77 |
| *(no flags)* | `./config.yaml` | `sx1268` | 42 |

`--check` agrees: bare `-s` reports "No configuration files were found", `-s -c custom.yaml` lists `custom.yaml` and reports `Module: sim`.

`pio run -e native` SUCCESS, clang-format clean.

Found while running two simulated nodes with `EnableUDP` for multicast discovery — the setting was ignored purely because `-s` was also passed.

Split out of #9749.

## Summary by CodeRabbit

* **Bug Fixes**
  * Explicit configuration files now load correctly when simulated-radio mode is selected.
  * Using simulated-radio mode without a specified configuration file now skips automatic configuration discovery.
  * Simulated-radio mode consistently takes precedence over configured LoRa hardware across normal operation, configuration checks, and YAML output.
  * Configuration settings from all supported locations are processed consistently without unintentionally re-enabling hardware radio mode.


matutetandil:
Thanks @coderabbitai — good catch, this one was real and I introduced it. Fixed in 47972b2.

`loadConfig()` sets `portduino_config.lora_module` whenever the YAML has a `Lora.Module` key (PortduinoGlue.cpp:938), and the `ConfigDirectory` loop calls `loadConfig()` once per `.yaml` entry. With the override sitting before that loop, a directory entry could put the module back and quietly undo `-s`.

Worth noting this was not reachable before this PR: `config_directory` is itself read from the main config file, and the old `if (force_simradio) … else if (configPath …)` chain meant that with `-s` no config was ever loaded, so `config_directory` stayed empty and the loop never ran. Loading the config unconditionally is what opened the window.

The override now runs after the directory loop and before `runConfigCheck()` / `emit_yaml()`, so every config source is applied first and the flag wins over all of them. That ordering matters for `--check` and `--output-yaml` in particular: the early `return` in the `force_simradio` block further down means the radio backend was never actually mis-selected at runtime, but both of those report `lora_module` before reaching it, so they would have printed a module the daemon was not going to use.

matutetandil:
Nitpick applied in 2e4ff91 — and it was right on both counts.

`AGENTS.md:83` (and `.github/copilot-instructions.md:338`) does say *"Keep code comments minimal - one or two lines, max"*, and I had a three-line comment there. The follow-up comment I added for the ordering fix was four lines, so I trimmed that one too:

```cpp
// Load the YAML before applying -s: it also carries non-radio settings
// (EnableUDP, display, GPIO) that have to survive simulated mode.

// Applied after every config source: ConfigDirectory entries can set Lora.Module
// too, and -s must win over all of them, including in --check / --output-yaml.
```

Both are two lines and keep only the non-obvious *why*.

jp-bennett:
I meant to comment on this earlier, but this is actually intended behavior. The -s flag is for simulator use, in which case we really don't want to auto-load yaml. I do agree that the -c flag should be honored.

matutetandil:
@jp-bennett thanks — that reframes the PR, so let me check which shape you want before I push anything.

I take your point that `-s` is for simulator use and should not go hunting for YAML on its own. But I do not think the current code implements that intent so much as it happens to produce it, because `-s` is simply the first arm of an `if/else-if` chain that also owns config loading:

```cpp
if (portduino_config.force_simradio == true) {
    portduino_config.lora_module = use_simradio;
} else if (configPath != nullptr) {
    // -c handled here, never reached when -s is set
```

So `-s` swallows an explicit `-c` as well, which is exactly the case you say should be honored. Under your own reading this still needs a fix — just a much narrower one than what I wrote.

What brought me here: a simulated node with `EnableUDP` under `Config:` in the YAML never joins the multicast group, because the whole file is skipped and nothing says so. Two simulated nodes therefore cannot find each other. Passing `-c` explicitly would be a perfectly good answer to that, if `-c` worked under `-s`.

The narrow version is roughly this — explicit `-c` honored, implicit discovery still skipped under `-s`:

```cpp
if (configPath != nullptr) {
    // an explicit -c is honored even in simulated mode
    ...
} else if (portduino_config.force_simradio) {
    // -s with no -c: do not go looking for a config
} else if (access("config.yaml", R_OK) == 0) {
    ...
```

The `force_simradio` override still has to move below the `ConfigDirectory` loop, since `loadConfig()` sets `lora_module` from `Lora.Module` and a directory entry could otherwise quietly undo `-s`. That part stands either way.

Two questions:

1. Narrow version (honor `-c` only), or the current one (always load, `-s` overrides the radio)? I am happy either way — the narrow one is closer to what you describe as intended.
2. Should `ConfigDirectory` follow `-c` or follow discovery? It is read *from* the config file, so under the narrow version it only ever populates when `-c` was given, which seems right but is worth confirming.

@caveman99 — you rebased this branch onto `develop` earlier today, so flagging that the two of you read the intent differently. I would rather rewrite it once, in whichever direction you both land on, than have it merge as-is if the premise is wrong.

caveman99:
@matutetandil i fast-forwarded the underlying fork so it doesn't drift too much from develop head. That operationw as not an approval or a rejection, i stand with @jp-bennett assessment. read and parse the -c flag, then if -s is specified force the simradio on.

matutetandil:
Reworked exactly as asked, @caveman99 — thanks for the call, and for the fast-forward.

`-c` is now parsed first and always honored; `-s` moved to sit *after* that branch, so it only skips the implicit discovery and forces the simradio on:

```c
if (configPath != nullptr) {
    // explicit -c: always parsed, including under -s
} else if (portduino_config.force_simradio) {
    // -s with no -c: the simulator brings its own defaults, so skip config discovery.
} else if (access("config.yaml", R_OK) == 0) {
```

@jp-bennett's point holds in full: a bare `-s` loads no YAML at all, same as before. What changes is only that a path the user typed by hand stops being swallowed.

The radio override stays after all config sources, since `-c` and any `ConfigDirectory` it pulls in can both set `Lora.Module` and `-s` has to beat them. That also answers the `ConfigDirectory` question I raised earlier — it needs no special case, it just follows whatever config actually loaded, so a bare `-s` leaves it empty and the loop never runs.

Checked against the built binary rather than by reading:

| invocation | config loaded | `Lora.Module` | `MaxNodes` |
|---|---|---|---|
| `-s` (with a `./config.yaml` present) | none | `sim` | 200 (default) |
| `-s -c custom.yaml` | `custom.yaml` | `sim` | 77 (from `-c`) |
| `-c custom.yaml` | `custom.yaml` | `sx1262` | 77 |
| *(no flags)* | `./config.yaml` | `sx1268` | 42 |

Row 1 is the one that matters for your and @jp-bennett's position: the local `config.yaml` is sitting right there and is deliberately not read. `--check` agrees — bare `-s` reports "No configuration files were found", while `-s -c custom.yaml` lists the file and reports `Module: sim`.

Rebased onto current `develop`, still one file. `pio run -e native` SUCCESS, clang-format clean. Title and description updated, since the old ones described the wider change.

jp-bennett (approved):

matutetandil:
Thanks @jp-bennett, @thebentern.

The check-label run fired at 15:56, about four minutes before the label was added, so it's stuck on a stale payload rather than an actual missing label. A re-run should clear it — I can't trigger one from here. Everything else is green.

matutetandil:
This has been approved since Aug 9 and is showing MERGEABLE/CLEAN, so I think it just needs someone to hit merge.

The one red check is `check-label`, which ran about four minutes before the label was added and is stuck on a stale payload. A re-run would clear it, but it isn't a required check, so it shouldn't be what's blocking. Let me know if anything else is needed here.

PR #11349, opened 2026-08-03
fix(t-deck): restore CALIBRATE_TOUCH so the touch panel is calibrated

merocle:
## Problem

The T-Deck touch panel is uncalibrated on the MUI build (`t-deck-tft`): a touch registers off by the 320/240 ratio — stretched by about 1.33 on X, compressed by about 0.8 on Y — accurate only near the bottom-left corner and worse with distance from it.

## Cause

`device-ui` ships calibration parameters for this board and applies them in `include/graphics/driver/LGFXDriver.h`:

```c
uint16_t parameters[8] = {0, 2, 0, 314, 223, 5, 224, 314};   // T_DECK
...
lgfx->setTouchCalibrate(parameters);
```

The whole block is behind `#ifdef CALIBRATE_TOUCH`. That flag was active for `t-deck-tft` until 8ae05f6b ("MUI: defcon tft display size definitions", #7142, 2025-06-26) commented it out while adding `LGFX_SCREEN_WIDTH` / `LGFX_SCREEN_HEIGHT`:

```diff
-  -D CALIBRATE_TOUCH=0
+;  -D CALIBRATE_TOUCH=0
+  -D LGFX_SCREEN_WIDTH=240
+  -D LGFX_SCREEN_HEIGHT=320
```

Neither that PR's description nor its discussion mentions touch, so this looks unintended. With the flag gone no calibration is applied at all and the panel runs on raw LovyanGFX values.

## Fix

Uncomment the line. `CALIBRATE_TOUCH=0` applies the stored parameters; `=1` would run interactive calibration, which does not persist yet — there is a `FIXME` in `device-ui` about reading it back from storage.

## Testing

Built `t-deck-tft` at v2.7.26.54e0d8d with this change and flashed app0 on a T-Deck Plus (868 MHz). Touch is accurate across the whole screen afterwards; before the change it missed as described above. LoRa, keyboard and trackball unaffected.

## Not addressed here

`t-deck` is the only variant that sets `CALIBRATE_TOUCH`, so the calibration block in `device-ui` is currently dead code for every board, even though it also carries parameters for `WT32_SC01`, `T_HMI`, `ESP32_2432S022`, `ESP32_2432S028RV1`, `UNPHONE`, `NODEMCU_32S` and `SENSECAP_INDICATOR`. I do not have those boards, so this PR touches one line rather than all of them.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Other

I have a T-Deck Plus and a Heltec V4 only. The change is scoped to the `t-deck` variant's build flags, so other boards cannot be affected.


NomDeTom:
Please retarget this change to `develop`

mverch67 (approved):
Display calibration may be individual for each display, that's why there's a calibration function in _Settings->Screen Calibration_. Though I agree that some default values may be better than nothing, especially if the lovyanGFX default is so far off that the settings button cannot be reached.
Upon startup these default values are overwritten with the stored values (from the last screen calibration) in NodeDB.

caveman99 (changes requested):
Please rebase this onto develop

vidplace7:
> Display calibration may be individual for each display, that's why there's a calibration function in _Settings->Screen Calibration_. Though I agree that some default values may be better than nothing, especially if the lovyanGFX default is so far off that the settings button cannot be reached. Upon startup these default values are overwritten with the stored values (from the last screen calibration) in NodeDB.

@mverch67 Alternatively. Can we prompt for screen calibration upon firstboot? (If a previous calibration is not already stored).

mverch67:
> @mverch67 Alternatively. Can we prompt for screen calibration upon firstboot? (If a previous calibration is not already stored).

that's also possible, yes. But there's already another prompt in place to set username / region. So we'd had two prompts (and for sure could think about having even more prompts), lol.

caveman99:
@merocle last request to rebase this one liner onto develop, or i will do it myself and close this one.

caveman99:
author unresponsive, CLA not signed.

PR #11358, opened 2026-08-05
fix(BaseUI): restore node list menu swallowed by module-frame SELECT dispatch

Ixitxachitl:
Fixes a regression introduced in #11209.

## Problem

`SELECT` no longer opens the node list menu on BaseUI devices. Reproduced on current `develop`.

## Root cause

#11209 added this branch to the `SELECT` dispatch chain in `Screen::handleInputEvent()`:

```cpp
} else if (this->ui->getUiState()->currentFrame < moduleFrames.size()) {
```

`moduleFrames` is not a list of module frames. `MeshModule::GetMeshModulesWithUIFrames(startIndex)` front-pads it with `nullptr`s so it stays index-aligned with the absolute frame index — the "lock step" arrangement `drawModuleFrame()` relies on:

```cpp
modulesWithUIFrames.resize(startIndex, nullptr);   // startIndex == numframes at call time
```

So `moduleFrames.size()` is the **total frame count**, not the module count. Since `setFrames()` builds the vector after every built-in frame is registered, `currentFrame < moduleFrames.size()` is true for all of them. Being mid-chain in an `else if` ladder, the branch then swallows `SELECT` for every branch below it, and because `moduleFrames.at()` returns a padded `nullptr` for those frames, `ownsFrame(nullptr)` is false and nothing happens at all.

This does not depend on which modules are enabled — with zero modules requesting frames the padding alone makes `size() == numframes`.

## Impact

Determined purely by registration order relative to where `moduleFrames` is built:

- **Broken:** all six node list frames (nodes / location / lastheard / hopsignal / distance / bearings).
- **Also expected to be broken, untested:** the WiFi frame, which is registered before `moduleFrames` is built. Inferred from reading `setFrames()`, not observed on hardware.
- **Unaffected:** favorites (registered *after* the module frames, so their index is `>= size()`), and home / system / gps / map / clock / lora / textMessage (they appear earlier in the chain).

No crash — `ownsFrame()` is `module == this`, so the null argument is harmless. The menus just stop responding.

This is not display- or variant-specific: `moduleFrames` padding and `setFrames()` ordering are identical on every target, so any BaseUI device is affected.

## Fix

A non-null entry is what actually distinguishes a real module frame from the padding, so test that rather than only the bound:

```diff
-                } else if (this->ui->getUiState()->currentFrame < moduleFrames.size()) {
+                } else if (this->ui->getUiState()->currentFrame < moduleFrames.size() &&
+                           moduleFrames.at(this->ui->getUiState()->currentFrame) != nullptr) {
```

Padded entries fall through to the node list / WiFi handlers below; genuine module frames still reach `environmentTelemetryMenu()`. The change only narrows an over-broad condition, so it cannot remove a branch that previously matched.

## Testing

Tested on:
- **Wio Tracker L1** — node list menu broken before, working after.
- **LilyGo T-Watch Ultra** (custom branch) — node list menu broken before, working after.

Help wanted on:
- **Single-button devices.** These were reported as seemingly unaffected, but I can't find a mechanism for that in the code — a one-button long press emits `INPUT_BROKER_SELECT` (`InputBroker.cpp`) into the same chain, so they should be affected too. Confirmation either way would be useful, along with a check that this PR doesn't regress them.
- **The WiFi menu**, which I have not been able to test.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Other devices tested: Wio Tracker L1, LilyGo T-Watch Ultra (custom branch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed module-specific menu handling so built-in frames are no longer incorrectly treated as module frames.


Xaositek:
Tested on Wio Tracker L1 (previously broken) and Meshnology W12 (was okay before) - fixed L1 and no regressions in W12

Xaositek (approved):

PR #11374, opened 2026-08-08
logging: audit log strings for terseness, reclaiming ~6.8 KB of string data

jp-bennett:
Audits all 3,161 `LOG_*` format strings in `src/` and rewrites the verbose ones to terser equivalents, reducing total log-string data from 121,628 to 114,834 bytes — **6,794 bytes saved (5.6%)**, with no strings dropped. A typical firmware image reclaims roughly 3–5 KB, since the biggest trims are in code nearly every target compiles (NodeDB, Router, AdminModule, MeshService, GPS, PhoneAPI, sensors).

What was changed, in passes:

- **Mechanical pass (all of `src/`)**: strip trailing `.`/`!`/`...` and literal `\n` (the logger appends a newline itself), `Error:`/`Warning:` prefixes (the level tag already conveys it), and redundant "successfully".
- **Judgment pass (top ~40 byte-heaviest files)**: rewrite wordy prose — "Failed to initialize X" → "X init failed", "Attempting to send" → "Send", dropped articles and filler, house abbreviations (config, init, msg, temp, BT).
- **Global pass**: "Unable to / Could not / Cannot" → "Can't".
- **clang-format rewrap** of lines whose length changed, using the pinned style (`.trunk/configs/.clang-format`, clang-format 16).

Deliberately preserved: every printf format specifier and argument list byte-for-byte (verified by an automated pre/post specifier-sequence check on every call), greppable tokens (chip/sensor/function/state names, key=value field names), the `0x%08x` node/packet-ID convention, strings the external hardware-test harness greps for, the boot-logo banner's trailing newline and bare `"."` progress ticks, and identical duplicate literals (so linker string dedup still applies). Non-log strings (UI text, NMEA/UBX, MQTT topics, CSV headers) are untouched.

*(Follow-up work — the `MESHTASTIC_TRACE_LOGGING` gate, chatty-DEBUG demotions, and redundant-log removal — moved to #11391 after this PR merged.)*

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Tested via the native (portduino) target: full compile green and the native test suite **GREEN — 44/44 suites, 352/352 test cases, all CLEAN** (`./bin/run-tests.sh`). An automated check verified the format-specifier sequence of every `LOG_*` call is unchanged. No device-specific behavior is affected beyond log text.

### DIFF
--- a/src/AmbientLightingThread.h
+++ b/src/AmbientLightingThread.h
@@ -66,7 +66,7 @@ class AmbientLightingThread : public concurrency::OSThread
 #if defined(HAS_NCP5623) || defined(HAS_LP5562)
         _type = type;
         if (_type == ScanI2C::DeviceType::NONE) {
-            LOG_DEBUG("AmbientLighting Disable due to no RGB leds found on I2C bus");
+            LOG_DEBUG("AmbientLighting disabled: no RGB leds on I2C");
             disable();
             return;
         }
@@ -92,7 +92,7 @@ class AmbientLightingThread : public concurrency::OSThread
                 pixels.setBrightness(moduleConfig.ambient_lighting.current);
 #endif
                 if (!moduleConfig.ambient_lighting.led_state) {
-                    LOG_DEBUG("AmbientLighting Disable due to moduleConfig.ambient_lighting.led_state OFF");
+                    LOG_DEBUG("AmbientLighting disabled: led_state OFF");
                     disable();
                     return;
                 }
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -171,7 +171,7 @@ static bool initAdcCalibration()
     }
 #endif
 
-    LOG_INFO("ADC calibration not supported; using approximate scaling");
+    LOG_INFO("ADC calibration unsupported; use approx scaling");
     return false;
 }
 
@@ -885,8 +885,8 @@ void Power::reboot()
     HAL_NVIC_SystemReset();
 #else
     rebootAtMsec = -1;
-    LOG_WARN("FIXME implement reboot for this platform. Note that some settings "
-             "require a restart to be applied");
+    LOG_WARN("FIXME implement reboot for this platform; some settings "
+             "need restart to apply");
 #endif
 }
 
@@ -1926,7 +1926,7 @@ meshSolarBatteryLevel meshSolarLevel;
 bool Power::meshSolarInit()
 {
     bool result = meshSolarLevel.runOnce();
-    LOG_DEBUG("Power::meshSolarInit mesh solar sensor is %s", result ? "ready" : "not ready yet");
+    LOG_DEBUG("Power::meshSolarInit sensor is %s", result ? "ready" : "not ready yet");
     if (!result)
         return false;
     batteryLevel = &meshSolarLevel;
@@ -2059,7 +2059,7 @@ bool Power::serialBatteryInit()
 #endif
 
     bool result = serialBatteryLevel.runOnce();
-    LOG_DEBUG("Power::serialBatteryInit serial battery sensor is %s", result ? "ready" : "not ready yet");
+    LOG_DEBUG("Power::serialBatteryInit sensor is %s", result ? "ready" : "not ready yet");
     if (!result)
         return false;
     batteryLevel = &serialBatteryLevel;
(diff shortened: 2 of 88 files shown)


Xaositek (approved):

thebentern (approved):

thebentern:
Went through all 17 inline findings. 12 were still live against the current branch, 4 were already resolved by the two follow-up commits, and 1 I've left alone deliberately.

### Fixed

**Node/packet ID convention** (`copilot-instructions.md`: "Format node IDs and packet IDs as `0x%08x`")

- `NextHopRouter.cpp:70` — `p->to` → `0x%08x`. `p->next_hop` stays `0x%x`: it's the last-byte relay hint, not a NodeNum, which matches lines 269/284/287 in the same file.
- `NodeDB.cpp:4333` — `oldNodeNum`/`newNodeNum` `%u` → `0x%08x`.
- `AdminModule.cpp:209` and `CannedMessageModule.cpp:1114`.
- I also converted the sibling in each `if`/`else` pair (`AdminModule.cpp:212`, `CannedMessageModule.cpp:1111`). Those were pre-existing `%x` and untouched by this PR, but leaving one branch on `%x` and the other on `0x%08x` three lines apart would read as an oversight.

**Message names the wrong operation** — these were real bugs, good catches:

- `SCD4XSensor.cpp:511` and `:580` are `getAmbientPressure()` / `setAmbientPressure()` but logged "altitude". Now "ambient pressure". (Lines 487/540 genuinely are the altitude accessors and keep their wording.)
- `SCD30Sensor.cpp:227` is `getASC()` but logged "Can't send command" → "Can't get ASC status". Line 207 in `setASC()` really is a write, so it keeps the original.

**Redundant level text in `LOG_ERROR`**

- `. Error: %u` → `, rc=%u` at 15 sites, matching the existing `rc=%d` house style. All of them are confined to the two Sensirion files.
- `Error executing X()` → `X() failed`; `Error performing FRC` → `FRC failed`.
- `HostMetrics.cpp:54` — "Error decoding" → "Can't decode".

**Typos and wording**

- `MeshtasticOTA.cpp:112` — "OTA partiton.&nbsp;&nbsp;(Reason %d)" → "OTA partition (reason %d)".
- `t_deck_pro/variant.cpp:122` — "CST3530 not response ~" → "CST3530 no response".
- `Router.cpp:339` — "Packet received with to: of 0" → "with to=0".
- `NextHopRouter.cpp:506` — dropped the dangling `": "`.

**RTC** — worth a note, because the stated rationale doesn't hold but the underlying problem does. The four sites sit in mutually-exclusive `#ifdef` branches, so no build ever compiles two RTC families and the family can't actually be ambiguous. What *is* ambiguous is read vs. set: `readFromRTC()` and `perhapsSetRTC()` each emit a byte-identical literal, which the linker then dedups into one string. Split into `"RTC read: not found"` / `"RTC set: not found"`.

### Not changed

- **`main.cpp:457`** (banner `\n\n`) — deliberate. The leading blank lines separate the boot logo from preceding output, and 33de80d restored them after the terseness pass over-trimmed them. The PR description lists the banner as intentionally preserved.
- **`WebServer.cpp:167`, `StoreForwardModule.cpp:633`, `main-nrf52.cpp:537`** (`LOG_DEBUG("")`) — already fixed; 33de80d restored all three to `"."`. These findings were raised against the intermediate state.
- **Nitpick on comment length** at `main.cpp:1370-1373` / `AdminModule.cpp:1180-1182` — those comments are pre-existing and untouched by this PR; they show up only as diff context. Compressing security rationale isn't something I want to fold into a log-string audit.

### Verification

`trunk fmt` clean. An automated check over all 37 changed lines confirms each printf specifier sequence is byte-identical except for the 6 intentional `%x`/`%u` → `%08x` node-ID widenings, all on `uint32_t` arguments.

Built `t-deck-pro` (esp32-s3) locally, which compiles all 11 touched files — including the four the native target can't reach (both Sensirion sensors, the ESP32 OTA path, and the T-Deck Pro variant). SUCCESS, no new warnings. The native target can't be built on macOS (portduino needs glibc's `argp.h`), so I used the cross-build instead of `run-tests.sh`.

Issue #10040, opened 2026-03-30
[Bug]: Lilygo T-Lora T3 v1.6.1 reboots when I connect to it via HTTP


tommyblue:
### Category

WiFi

### Hardware

T-Lora v2 1.6

### Is this bug report about any UI component firmware like InkHUD or Meshtatic UI (MUI)?

- [ ] Meshtastic UI aka MUI colorTFT
- [ ] InkHUD ePaper
- [ ] OLED slide UI on any display

### Firmware Version

2.7.15.567b8ea

### Description

If I setup the wifi connection the device connects correctly, but if I try to open it via browser or with the meshtastic android app, it immediately reboots. The app can find the device in the network, the device only reboots if I try to open its settings.

I tried using different cables and power supplies (including a 60W laptop usb-c power supply), so I'm pretty sure is not something related to low power. I also tried with and without a battery plugged in. Same result.

The same issue happens with the firmware 2.7.20.6658ec2

I use Europe 868 with MEDIUM_FAST

### Relevant log output

```Shell
[ 44870][E][WiFiClient.cpp:67] fillBuffer(): Not enough memory to allocate buffer
[ 44887][E][WiFiClient.cpp:474] read(): fail on fd 52, errno: 11, "No more processes"
INFO  | 22:31:38 44 [ServerAPI] Client dropped connection, suspend API service
DEBUG | 22:31:38 44 [ServerAPI] PhoneAPI::close()
DEBUG | 22:31:38 44 [ApiServer] PhoneAPI::close()

abort() was called at PC 0x4020f20f on core 1


Backtrace: 0x40084149:0x3ffe2340 0x400976e9:0x3ffe2360 0x4009d51d:0x3ffe2380 0x4020f20f:0x3ffe2400 0x4020f256:0x3ffe2420 0x4020f59b:0x3ffe2440 0x4020f4fa:0x3ffe2460 0x40144c1a:0x3ffe2480 0x40144d82:0x3ffe24d0 0x40164ba6:0x3ffe24f0 0x40143ab9:0x3ffe2510 0x400e9ff1:0x3ffe2560
```

Anatoli4-dot:
hi) my logs:
```
[382269][E][WiFiClient.cpp:429] write(): fail on fd 52, errno: 11, "No more processes"
[383268][E][WiFiClient.cpp:429] write(): fail on fd 52, errno: 11, "No more processes"
```

mverch67:
[...]

PR #11377, opened 2026-08-09
fix: null-check NodeDB pointers and harden RX path to prevent heap-pressure crashes

clawoneloke:
## Summary

Three related fixes that address the T-Lora T3 reboot on HTTP connect reported in [meshtastic/firmware#10040](https://github.com/meshtastic/firmware/issues/10040). Root cause: heap exhaustion during config dump triggers null derefs and packet pool exhaustion.

### 1. `src/mesh/MeshService.cpp` — Null-check `getMeshNode(mp->from)` before dereferencing

Cache the NodeDB lookup in a local pointer before dereferencing `->has_user` in `handleFromRadio()`. This closes the TOCTOU window where a concurrent task could evict the node between the lookup and the dereference, and reduces two linear NodeDB scans per RX packet to one.

Before:
```cpp
} else if (... && !nodeDB->getMeshNode(mp->from)->has_user && ...)
```

After:
```cpp
meshtastic_NodeInfoLite *fromNode = nodeDB->getMeshNode(mp->from);
if (fromNode != nullptr && !fromNode->has_user) {
```

### 2. `src/mesh/Router.cpp` — Cache NodeDB lookups in PKI decode path

Cache `getMeshNode(p->from)` and `getMeshNode(p->to)` before dereferencing `->user.public_key.size`. Eliminates duplicate linear scans and closes the TOCTOU hazard.

### 3. `src/mesh/RadioLibInterface.cpp` — Four hardening fixes

- **`randomBytes()`**: Fix half-open range — `random(0, 255)` can never produce byte value 255. Changed to `random(0, 256)` per RadioLib semantics.
- **CRC mismatch log level**: Downgrade `RADIOLIB_ERR_CRC_MISMATCH` from `LOG_ERROR` to `LOG_INFO` — CRC mismatches are routine in RF-congested areas and not an error condition.
- **`rxGood` placement**: Move `rxGood++` to always count CRC-valid OTA receptions before short/from-null packet rejection, so valid airtime is not silently lost from stats.
- **Pool exhaustion guard**: Add null-check after `packetPool.allocZeroed()` to prevent NULL deref when heap pressure is high during config dumps.

## Testing

These fixes address the crash pattern observed in #10040:
```
[WiFiClient.cpp:67] fillBuffer(): Not enough memory to allocate buffer
abort() was called at PC 0x4020f20f on core 1
```

Compiles successfully on ESP32 (T-Lora T3 target). Please test on SX126x devices with active mesh and phone connections.

## Related

- Closes [meshtastic/firmware#10040](https://github.com/meshtastic/firmware/issues/10040)
- Related closed PRs: #10229, #10226, #10252 (same fixes, closed without merge)

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -94,18 +94,23 @@ int MeshService::handleFromRadio(const meshtastic_MeshPacket *mp)
         mp->decoded.portnum == meshtastic_PortNum_TELEMETRY_APP && mp->decoded.request_id > 0) {
         LOG_DEBUG("Received telemetry response. Skip sending our NodeInfo");
         //  ignore our request for its NodeInfo
-    } else if (mp->which_payload_variant == meshtastic_MeshPacket_decoded_tag && !nodeDB->getMeshNode(mp->from)->has_user &&
-               nodeInfoModule && !isPreferredRebroadcaster && !nodeDB->isFull()) {
-        if (airTime->isTxAllowedChannelUtil(true)) {
-            const int8_t hopsUsed = getHopsAway(*mp, config.lora.hop_limit);
-            if (hopsUsed > (int32_t)(config.lora.hop_limit + 2)) {
-                LOG_DEBUG("Skip send NodeInfo: %d hops away is too far away", hopsUsed);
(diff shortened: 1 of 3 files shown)


caveman99:
This has been fixed on develop independently.

PR #11381, opened 2026-08-10
Remove proprietary Bosch BSEC blob; open in-tree IAQ estimator for BME680

thebentern:
## Why

The Bosch BSEC2 blob is one of our largest dependencies: measured ~37-39 KB flash and ~4-5 KB static RAM on every image that links it (~190 of ~240 targets), paid whether or not a BME680 is attached. It's also a no-source proprietary archive statically linked into GPLv3 release binaries, and it's been the recurring culprit in our nRF52 flash-pressure incidents (#11363 left a "TEMPORARY" `lib_ignore` hack on `nrf52_promicro_diy_tcxo` with a TODO to make the roster opt-in).

The firmware only ever consumed one BSEC-exclusive output: the `iaq` value. Temperature, humidity, pressure, and gas resistance all come from the plain sensor.

## What

- **Remove BSEC2 + Bosch-BME68x-Library from all targets.** `environmental_extra_common` / `environmental_extra` / `environmental_extra_no_bsec` collapse into a single `environmental_extra` (common sensors + Adafruit_BME680 2.0.6, which vendors Bosch's open BSD-3 `bme68x` driver). The per-variant BSEC link-path hacks (ThinkNode-M3, muzi_base) and the promicro `lib_ignore` block are gone.
- **New in-tree IAQ estimator** (`BME680IaqEstimator`): humidity-compensated log-resistance tracked against a rise-fast/decay-slow rolling ceiling, scored onto the same 0-500 scale and bands the UI already uses. Pure math, no platform dependencies, fully unit-tested on native (15 tests, including a deep-sleep reboot-cycle simulation). Warm-up/burn-in progress persists to `/prefs/bme680.dat` (atomic SafeFile), so one-sample-per-wake SENSOR nodes converge across reboots; the orphaned `/prefs/bsec.dat` is deleted once.
- **BME680Sensor rewrite**: async non-blocking sampling once per minute (~0.25% heater duty vs ~5% under BSEC LP - a real battery win), a hard 2-minute publish-freshness bound (a dead sensor stops reporting instead of freezing its last reading on the wire), and suppression of bogus `gas_resistance=0` points from heater-unstable cycles.
- **UI**: `iaq` rendering now gates on `has_iaq`, so a legitimate IAQ of 0 displays instead of vanishing. Alert thresholds (banner >150, beep >200) unchanged.
- **Size budgets**: rak4631 tightened (RAM 113000→108000, flash 786000→746000) to lock in the reclaimed headroom.
- **Tuning harness**: `bin/bme680_iaq_replay.cpp` replays a CSV captured from a BSEC build through the estimator on a dev machine and reports mean absolute error + UI-band agreement, so the constants can be tuned against real BSEC traces without reflashing (capture instructions in the file header).

## Measured (develop → this branch, same commit)

| Target | Flash | Static RAM | Notes |
|---|---:|---:|---|
| `rak4631` | **-38.8 KB** | **-4.9 KB** | image ends 0xD7188, 75 KB clear of the warm-store cap; on nRF52 static RAM converts 1:1 into heap |
| `heltec-v3` | **-36.4 KB** | **-4.0 KB** | representative of the ESP32-S3/C3/C6 family |
| `nrf52_promicro_diy_tcxo` | +7.0 KB | 0 | **regains BME680 support** (absent on develop since #11363), 36 KB clear of the cap |
| `tlora-v2-1-1_6` | +1.3 KB | 0 | OG ESP32, already on the Adafruit path |

## Behavior changes (release-notes material)

- IAQ is now a relative-to-recent-baseline estimate rather than Bosch's calibrated output. Values will read differently; the 0-500 scale and qualitative bands are preserved. There is a ~33-minute burn-in on first boot (or ~33 wakes for deep-sleep nodes) before IAQ appears.
- The OG-ESP32/native "IAQ approximation" from #9663 never actually ran (inverted `isfinite` check made it dead code). Those targets now report IAQ for the first time.
- `gas_resistance` stays kOhm on the wire for fleet continuity; the proto comment claiming MOhm gets a separate docs fix in meshtastic/protobufs.

## Testing

- `test_bme680_iaq` (new, 15 cases) passes; full native suite shows no regressions vs develop (the only failure, `test_packet_signing` B11/B12, fails identically on clean develop).
- Local builds green on rak4631, nrf52_promicro_diy_tcxo, tlora-v2-1-1_6, heltec-v3; both nRF52 targets pass the warm-region post-link guard; the rak4631 budget gate passes with the tightened limits.
- Estimator numerics verified against the design targets (band mapping at 1.31x/1.7x/3x/15x baseline ratios, humidity deadband, burn-in persistence, stale/corrupt state rejection).

Hardware soak on a RAK4631+RAK1906 against a BSEC-build reference node is the remaining validation; the estimator constants are centralized and documented as tunables for exactly that.

## Summary by CodeRabbit

- **New Features**
  - Added standalone BME680/BME688 indoor air quality estimation with warm-up tracking, humidity compensation, readiness status, and IAQ scores from 0–500.
  - IAQ state persists across restarts and sleep cycles, with validation for corrupted or outdated data.
  - Improved sensor sampling, freshness handling, and IAQ availability reporting.

- **Documentation**
  - Added instructions and tooling for replaying sensor data and comparing estimated IAQ results.

- **Chores**
  - Removed reliance on proprietary BSEC components and reduced RAK4631 memory budgets.


caveman99 (approved):

thebentern:
## Hardware A/B: open estimator vs. BSEC on two RAK4631 + BME680

14-hour side-by-side soak, `2026-08-12T18:49` → `2026-08-13T08:53`, 49,635 raw trace samples. Posting the full result for posterity, including where it does *not* meet the targets I set out with.

### Method

Two RAK4631s with BME680 at 0x76:

- **A (reference)** — `origin/develop` + BSEC, plus a local-only patch logging `IAQCSV,<raw_gas>,<heat_comp_rh>,<bsec_iaq>,<bsec_accuracy>` from `getMetrics()`.
- **B (candidate)** — this branch, open estimator.

The number that matters is the **replay**, not the live A/B: device A's own logged readings are fed offline through `BME680IaqEstimator` via `bin/bme680_iaq_replay.cpp`, so both algorithms see byte-identical input and every difference is algorithmic. The live two-board comparison is a gross-breakage check only — unit-to-unit BME680 gas-resistance spread is about an order of magnitude, and it bit me during this run (device B read 107 while device A's gas fell 90%, purely because they saw different IPA concentrations).

Two resampling details that decide whether the replay is valid at all:

- The BSEC build logs at **~1 Hz**, but this branch feeds its estimator once per `SAMPLE_INTERVAL_MS = 60 s`. Replaying raw would run the adaptation ~60× fast and collapse `ALPHA_DOWN`'s ~12 h decay to ~12 min. The harness resamples to one row per wall-clock minute.
- Scoring is restricted to rows where BSEC's own `accuracy >= 2`, but the **full** trace is replayed — the estimator is a stateful baseline tracker, so filtering its input would corrupt its history.

### Size claim: independently confirmed

Same board, same toolchain, from the linker:

| | Flash | Static RAM |
| --- | --- | --- |
| develop + BSEC | 755,496 B (92.7%) | 103,012 B (41.4%) |
| this branch | 715,608 B (87.8%) | 97,996 B (39.4%) |
| **saved** | **39,888 B (38.9 KB)** | **5,016 B (4.9 KB)** |

Matches the ~37–39 KB / ~4–5 KB in the PR description. (A carries the trace patch, so the true BSEC-only flash delta is a hair under 39,888 B. Note a UF2 size delta is 2× the flash delta — UF2 carries 256 payload bytes per 512-byte block — so don't quote UF2 bytes.)

### What passed

**Responsiveness — r = 0.92** across 93 scored rows. A dry-VOC event (IPA gauze) took the estimator **13 → 430** while BSEC went 107 → 500:

```
 n    gas     rh     est   bsec
763   31541   43.25   13    107    <- baseline
764    3050   44.33  430    500    <- IPA
767    3315   44.50  412    500
```

**No false alarms in clean air** — 718 overnight rows, max IAQ **7**, zero excursions above the 150 banner or 200 buzzer thresholds.

**Baseline resists sustained pollution** — unplanned, but I left the source in place for 45 minutes: the estimator held at 242 → 255 rather than normalizing the pollution away. `ALPHA_DOWN` behaves as designed.

**Burn-in and persistence work on real hardware** — first output at 33 samples as specified, state written to `/prefs/bme680.dat` each minute thereafter.

### What did not pass, and why I don't think the numbers mean what they look like

**Mean absolute error 123.1 (target < 50). Band agreement 23.7% (target ≥ 80%).**

Two reasons to discount those specific figures:

1. **93 scored rows, and BSEC is pegged at its 500 ceiling in a large fraction of them.** You cannot measure error against a saturated reference — at est 430 vs BSEC 500, BSEC simply cannot report higher.
2. **BSEC took 13 hours to reach `accuracy >= 2` at all**, and only got there after an environmental excursion. Per Bosch's own header, calibration needs exposure to clean *and* polluted air (~30 min each); it is exposure-driven, not time-driven. In a still room it sits at accuracy 1 indefinitely — and while it did, it drifted from 51 → 99 over twelve hours *while raw gas rose 28%*, i.e. it reported worsening air as the air measurably cleaned.

The real finding is that the disagreement is a **shape mismatch, not a scale error**:

| gas | est | BSEC | est/BSEC |
| --- | --- | --- | --- |
| 31,541 | 13 | 107 | 0.12 |
| 13,136 | 152 | 440 | 0.35 |
| 8,917 | 216 | 466 | 0.46 |
| 3,050 | 430 | 500 | 0.86 |

The curves converge under heavy pollution and diverge at the clean end — which is where users live and where the 150/200 thresholds sit. **No single `LN_RANGE` closes this**: fitting BSEC at gas 7,926 implies `LN_RANGE ≈ ln(9.3)`, fitting at 15,827 implies `ln(2.4)`. Those are irreconcilable, so it's a mapping-shape question (offset-plus-gain, or a non-log-linear curve), not a retune of the existing constants.

Whether that's worth chasing is a judgement call I'd rather surface than silently optimize toward: the estimator is relative-to-baseline **by design**, and it demonstrably responds to real pollution, resists sustained pollution, and never false-alarmed in 13 hours of clean air. If the goal is "usable IAQ without the blob," matching BSEC's absolute banding may be the wrong target.

### One real bug found, with a fix

A humid transient could **ratchet the clean-air baseline upward**.

`x = logf(gasOhms) + KH * (rh - RH_REF)` is unbounded in the humidity term. Exhaled breath drops raw gas (real VOCs) while RH jumps far enough that `exp(KH*dRH)` more than cancels the drop — so the *compensated* sample lands above the ceiling and pulls it up via `ALPHA_UP`. The inflated baseline then decays only at `ALPHA_DOWN` (~12 h).

Measured, from this run — a breath event at 07:20 (RH 41.9 → 56.5, raw gas −26%):

```
 n    gas     rh     est   bsec
751   32986   41.85    0    102   <- before
752   24450   56.52    0    252   <- breath: KH cancels it (by design)
763   31541   43.25   13    107   <- after: same air, now reads 13
```

Compensated gas went *24% above* the pre-event baseline during a pollution event. Humidity compensation normalizes a reading for comparison; it should not be able to mint a new "cleanest air seen" record.

Fix — gate upward ceiling movement on the direction of the **raw** change:

```cpp
const bool rawImproved = !haveLastRaw || xRaw >= lastXRaw;
const float alpha = (x > lnCeiling && rawImproved) ? ALPHA_UP : ALPHA_DOWN;
lnCeiling = std::clamp(lnCeiling + alpha * (x - lnCeiling), LN_FLOOR, LN_CEIL_MAX);
```

Gating on *direction* rather than clamping the ceiling to the raw value matters: clamping mixes compensated and uncompensated space and would under-report by ~65 IAQ points at any steadily elevated RH. The guard state is deliberately not persisted — `BME680IaqState`'s 24-byte on-disk layout is fixed, and after a restore the first sample is simply ungated.

Replayed against the same captured trace, post-event false elevation drops **13 → 6** and the IPA response is preserved (436 → 430). It halves rather than eliminates the effect: raw gas is also *recovering* during the tail of a humid event, so some legitimate-by-this-rule rise remains. Given the residual is ~6 points against a 150 threshold, I stopped there rather than adding more state.

Two regression tests added to `test_bme680_iaq` using the measured samples — one pinning the ratchet, one pinning that a genuine improvement at steady elevated RH still re-baselines unbiased. Suite green at 17 cases.

### Incidental

`develop`'s non-BSEC fallback is dead code: `if (gasRaw > 0.0f && !isfinite(gasRaw))` can never be true, so boards on the Adafruit path have never reported IAQ at all (introduced in #9663). This PR is what makes that path work, not merely what makes it free.

### Reproducing

`bin/bme680_iaq_replay.cpp` + `docs/bme680_iaq_replay.md` are in-tree and sufficient to redo this; the only out-of-tree piece is the one-line `LOG_INFO` on the BSEC build, documented in that same file. Tuning is a rebuild-and-rerun against a captured trace with no reflashing.

Caveats worth stating plainly: single sensor pair, one room, one night, 93 scored rows, and the strongest event saturated the reference. The pass/fail items above I'd consider settled; the absolute-agreement numbers I would not, until there's a sweep that holds BSEC in the 200–350 range without pegging.

PR #11407, opened 2026-08-11
gps: fall back to generic NMEA when vendor probes go unanswered

suteny0r:
## Summary

GPS fix: when the GNSS probe fails to get a vendor-specific response but a valid NMEA stream is arriving on the GPS UART, select a new generic-NMEA driver instead of concluding "no GPS".

## Problem

Some GNSS modules auto-stream standard NMEA but never answer chip-specific probe commands. The GAT562 Mesh Tracker Pro's L76K (CASIC-family) module is one such case: it runs in its default 9600 baud multi-GNSS (GPS + BeiDou) mode and streams `$GNGGA` / `$GNRMC` / `$GPGSV`, but it does not reply to any of the vendor probes Meshtastic sends (`$PUBX`, `$PMTK605`, `$PCAS06`, `$PQTMVERNO`, `$PDTINFO`, `$PAIR021`, UBX config, ...).

`GPS::probe()` / `GPS::setup()` interpret the silence as "no GNSS present", so the module's NMEA is never fed to the TinyGPS++ parser and no positions ever reach the mesh.

## Fix

- Add `GNSS_MODEL_GENERIC_NMEA`.
- In `GPS::probe()`, at the end of each baud-rate attempt where no chip responded to any probe command, listen for a short window (500 ms) for known NMEA sentence types (`GGA`, `RMC`, `GSV`, `GSA`, `VTG`, `GLL`, `ZDA`). If a stream is detected, return `GNSS_MODEL_GENERIC_NMEA` for that baud rate instead of `GNSS_MODEL_UNKNOWN`.
- `GPS::setup()` sends no chip-specific init for the generic driver (u-blox / MTK / L76K / Airoha / ATGM / Unicore configuration is all skipped) and feeds the streaming sentences straight to the standard TinyGPS++ parser.

The check only fires when a known NMEA stream is actually present, so true "no GPS" devices are unaffected (they still fall through to the existing `No GNSS Module` path).

## Verification

Device serial log (`gat562_mesh_tracker_pro`, `2.7.23.b2bda3b`):

```
[GPS] Trying $PDTINFO (Unicore Family)...
[GPS] Trying $PCAS06,1*1A (ATGM33xx Family)...
[GPS] Trying $PAIR021*39 (Airoha Family)...
[GPS] Trying $PQTMVERNO*58 (LC86)...
[GPS] Trying $PCAS06,0*1B (L76K)...
[GPS] Trying $PMTK605*31 (MTK Family)...
[GPS] GNSS: no probe response, but NMEA stream detected at 9600 baud - using generic NMEA driver
[GPS] GNSS: using generic NMEA stream, skipping chip-specific init
[GPS] NMEA GPS time set ...
```

The NMEA stream is parsed (time source acquired from GPS); a position fix then depends on normal cold-start satellite acquisition.

## Should be pushed upstream to Meshtastic

This fix belongs in upstream Meshtastic, not just the GAT562 fork. It likely benefits other hardware:

- Any GNSS module that streams NMEA but ignores vendor probe commands.
- Boards with one-way UART wiring (MCU RX to GNSS TX only), where probe commands can never reach the module but NMEA still arrives.
- Modules running in generic NMEA mode where a vendor driver was never selected.

The fallback is conservative (only activates on a confirmed NMEA stream) and requires no config or wiring changes.

## Question for the board manufacturer (GAT / gat-iot.com)

On the GAT562 Mesh Tracker Pro, is the MCU's GPS TX (GPIO16 / P0.16) actually wired to the L76K module's UART RX input?

The probe phase transmits `$PCAS06` and every other vendor probe command, and none get a reply while NMEA continues to stream in. That is exactly the behavior expected if the MCU-to-L76K TX path is omitted on the PCB (one-way UART: the L76K talks, but the MCU cannot send to it). If the TX line is absent, this generic-NMEA fallback is required (and is the only way) to obtain positions from this board.

## Summary by CodeRabbit

* **New Features**
  * Added support for the GAT562 Mesh Tracker Pro board, including display, LoRa, GPS, battery monitoring, USB, power, and trackball features.
  * Added automatic detection for GNSS modules that provide standard NMEA data without vendor-specific responses.
  * Generic NMEA-compatible receivers can now initialize without chip-specific configuration.

### DIFF
--- a/boards/gat562_mesh_tracker_pro.json
+++ b/boards/gat562_mesh_tracker_pro.json
@@ -1,0 +1,52 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "nrf52840_s140_v6.ld"
+    },
+    "core": "nRF5",
+    "cpu": "cortex-m4",
+    "extra_flags": "-DARDUINO_NRF52840_FEATHER -DNRF52840_XXAA",
+    "f_cpu": "64000000L",
+    "hwids": [
+      ["0x239A", "0x8029"],
+      ["0x239A", "0x0029"],
+      ["0x239A", "0x002A"],
+      ["0x239A", "0x802A"]
+    ],
+    "usb_product": "GAT562 Mesh Tracker Pro",
+    "mcu": "nrf52840",
+    "variant": "gat562_mesh_tracker_pro",
+    "bsp": {
+      "name": "adafruit"
+    },
+    "softdevice": {
+      "sd_flags": "-DS140",
+      "sd_name": "s140",
+      "sd_version": "6.1.1",
+      "sd_fwid": "0x00B6"
+    },
+    "bootloader": {
+      "settings_addr": "0xFF000"
+    }
+  },
+  "connectivity": ["bluetooth"],
+  "debug": {
+    "jlink_device": "nRF52840_xxAA",
+    "svd_path": "nrf52840.svd",
+    "openocd_target": "nrf52840-mdk-rs"
+  },
+  "frameworks": ["arduino", "freertos"],
+  "name": "GAT562 Mesh Tracker Pro",
+  "upload": {
+    "maximum_ram_size": 248832,
+    "maximum_size": 815104,
+    "speed": 115200,
+    "protocol": "nrfutil",
+    "protocols": ["jlink", "nrfjprog", "nrfutil", "stlink"],
+    "use_1200bps_touch": true,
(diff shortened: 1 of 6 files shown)


suteny0r:
Closing: this PR was mis-targeted. My upstream is mtoolstec/mt-firmware, not meshtastic/firmware. Reopening against the correct repository.

PR #11409, opened 2026-08-11
Fix the all-zero MAC address for own node

jp-bennett:
Dropping the MAC address had the unintentional effect of setting the local node's MAC address to all 0s when pulling the nodedb over the API. This PR restores that mac.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed local device information so the device’s MAC address is correctly retained when node details are sent.
  * Updated the native test suite count to reflect the additional test coverage.


caveman99 (approved):

src/mesh/PhoneAPI.cpp
```diff
@@ -615,6 +615,9 @@ size_t PhoneAPI::getFromRadio(uint8_t *buf)
             auto info = TypeConversions::ConvertToNodeInfo(us);
             info.has_hops_away = false;
             info.is_favorite = true;
+            // NodeInfoLite dropped macaddr, so ConvertToUser() zero-fills it.
+            if (info.has_user && info.num == nodeDB->getNodeNum())
+                memcpy(info.user.macaddr, owner.macaddr, sizeof(info.user.macaddr));
```
thebentern, line 620: I think we always want the hardware macaddr in there even if the num / getNodeNum don't match

jp-bennett: oh, yep. Just glazed over that bit.

PR #11415, opened 2026-08-11
fix(nrf52): LTO was dropping the board variant's weak hook overrides

thebentern:
## The bug

Whole-image LTO — enabled arch-wide for `nrf52840` in #10655 — inlines the **empty weak body** of our variant hooks at the call site, because the weak default and the call site live in the **same translation unit**. The strong override in `variants/<arch>/<board>/variant.cpp` is then never linked, and the board's hardware setup silently doesn't run. The build stays green.

`nrf52_lto.py`'s `-fno-lto` variant recompile doesn't help here: the *caller* is the problem, not the variant object.

It needs both ingredients, so **this only affects 2.8**. The `earlyInitVariant()` indirection landed in #9438 and is present in `v2.7.26` too — but `v2.7.26` has no `-flto`, so the override linked normally.

## How it surfaced

The muzi R1 Neo's `earlyInitVariant()` drives `DCDC_EN_HOLD` (P0.13, the DC-DC hold after the user button) and `NRF_ON` (P0.29, *"tells IO controller device is on"*). Both vanished from the image, so the companion MCU never saw the nRF application come up and sat in its DFU indication — a purple LED — on an otherwise perfectly working node. The missing DC-DC hold is a second, quieter symptom.

Verified in the ELF rather than inferred. Before:

- no `earlyInitVariant` symbol anywhere in the binary
- `setup()` runs straight from `waitUntilPowerLevelSafe()` into the `LED_NOTIFICATION` block
- no `pinMode`/`digitalWrite` on P0.13 or P0.29 anywhere in the image

After:

```
0007fe90 <_Z16earlyInitVariantv>:
    movs r0, #13  ->  pinMode / digitalWrite   (DCDC_EN_HOLD)
    movs r0, #29  ->  pinMode / digitalWrite   (NRF_ON)
```

Confirmed on hardware: an R1 Neo flashed with this branch.

## Blast radius

Same construct, same arch-wide LTO — the R1 Neo is just the variant whose symptom is visible from across the room.

| Hook | nrf52840 variants overriding it |
|---|---|
| `earlyInitVariant` | 10 — incl. **t-echo-card**, which sequences its RT9080 3V3 rail there |
| `variant_shutdown` | 18 — t114, t-echo, ThinkNode M1–M8, meshlink, wio-tracker-L1 … (sleep pin parking, so deep-sleep leakage) |
| `variant_nrf52LoopHook` | 3 RAK variants |

`variant_shutdown` being dropped was **confirmed by build** on `heltec-mesh-node-t114`, not just by inspection.

## The fix

`__attribute__((noinline))` on both the weak declaration and definition — the same guard already carried by `loopCanSleep()`, `preFSBegin()`, PowerHAL, and `variant_enableBatteryLpcompWake()`, whose comment in `main-nrf52.cpp` already documents this exact failure mode. The variant hooks just never got it.

## Turning it into a red build

`_VARIANT_OVERRIDES` in `extra_scripts/nrf52_lto.py` goes from `_Z11initVariantv` alone to all eight hooks. **That post-link guard already had the right logic and would have caught this on every PR** — it simply wasn't listing Meshtastic's own weak variant hooks, only the core's.

With the list extended it goes red on both `r1-neo` and `heltec-mesh-node-t114` when the `noinline` is reverted:

```
*** nrf52 LTO guard: board variant override DROPPED from the image ***
  - _Z16earlyInitVariantv is strong in variant.cpp.o but weak/absent in the ELF
```

…and green with it. The failure message now names both possible causes (in-TU weak inline vs. the `-fno-lto` middleware losing its match) so the next person doesn't have to rediscover this.

## Testing

- `r1-neo`, `heltec-mesh-node-t114`, `t-echo-card` all build clean; ISR, warm-region and variant guards pass
- Guard proven to fail red on the reverted state, green on the fix
- Disassembly checked before/after on `r1-neo`
- Flashed to a physical muzi R1 Neo

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of board-specific startup, shutdown, loop, battery wake, and default configuration behavior when link-time optimization is enabled.
  * Enhanced diagnostics to identify possible override failures and provide troubleshooting guidance.
  * Board-specific hardware behavior is now less likely to be skipped or replaced during optimized builds.


Xaositek (approved):
Resolved issues with the R1 Neo! Nice work @thebentern

jp-bennett:
What a gross bug.

PR #11418, opened 2026-08-11
fix(nrf54l15): implement reboot() and loop on short fs_write()

cvaldess:
## Summary

Two bugs in the nRF54L15 port, both surfaced while chasing "settings don't survive a reboot" on the DK.

### 1. `Power::reboot()` has no `ARCH_NRF54L15` branch

It falls through to the generic `#else`, which only logs `FIXME implement reboot for this platform` and returns. So any config change that requires a restart — a LoRa region change, for instance — tears down BLE in preparation for a reboot that never happens, and the phone app is left waiting for a node that is never coming back.

Fixed with a branch calling `sys_reboot(SYS_REBOOT_COLD)`: the same recovery path this platform's crash handler already uses, and it lets Zephyr do its own pre-reset bookkeeping (reset-reason register) before the CPU resets. No extra include is needed — the symbol is already reachable there.

### 2. `File::write()` does not handle short writes

`platform/nrf54l15/InternalFileSystem.h` issues a single `fs_write()` and returns whatever it consumed. `fs_write()` may short-write, exactly like POSIX `write()`. nanopb's stream callback treats any count short of what it asked for as a fatal `io error` and aborts the encode, so a short write silently corrupts a save.

Small protobufs (a lightly populated NodeDB) always fit in one call and never expose it; larger ones (`DeviceState`, `LocalConfig`) intermittently do not. The observed symptom was `device.proto` and `config.proto` failing to save while `nodes.proto`, `module.proto` and `channels.proto` saved fine. It now loops until the whole buffer is written or a real error occurs.

## Testing

Both fixes have been running on a nRF54L15-DK (PCA10156) in a separate NCS/Zephyr build of this same source tree: config and BLE bonds survive hot resets and power cycles, and a LoRa region change now actually reboots the node instead of leaving the app hanging.

`nrf54l15dk` build with the changes: SUCCESS, no size regression.

## Summary by CodeRabbit

- **New Features**
  - Added Ethernet-based firmware OTA with authentication, validation, and recovery support.
  - Added ESP32 OTA coprocessor workflows for Wi-Fi downloads, web-based uploads, and BLE firmware transfer.
  - Added configurable persistent GPIO output control and automatic coprocessor detection.
  - Added GPS-enabled configuration support.
- **Bug Fixes**
  - Improved Ethernet reconnection, firmware writes, reboot handling, and OTA reliability.
  - Updated radio transmit calibration.
- **Documentation**
  - Added setup and operating guides for GPIO control and OTA coprocessors.
- **Chores**
  - Added automated firmware build and publishing workflows.


caveman99:
Files changed: 53. Please learn how to PR properly.

PR #11422, opened 2026-08-12
Sign the whole Data envelope, in one unambiguous layout

jp-bennett:
## What this fixes

Today's signature covers `from | id | portnum | payload`. Everything else in the `Data` envelope rides outside it, and channel crypto is AES-CTR with no MAC — so any PSK holder (public on LongFast) can rewrite those fields in flight and the signature still verifies.

Concretely, all of this works against a signed broadcast right now:

- **Re-point a reply.** The client sets `reply_id` on an outgoing text to make a tapback; firmware signs the broadcast. `reply_id` is unsigned, so a listener can re-attach a genuine signed statement under a different message. *"Confirmed"* moves from under *"Is the bridge out?"* to under *"Is the road safe?"*, and the signature makes it more credible, not less.
- **Flip a reaction.** A reaction is a `TEXT_MESSAGE_APP` packet with the emoji in the payload, `reply_id` naming the parent, and the `emoji` flag telling the client to render it as a reaction. That flag is unsigned too, so a signed reply becomes a signed reaction — or the reverse — on a message the sender never saw.
- **Grant MQTT consent.** `bitfield` bit 0 is `OK_TO_MQTT`, which `MQTT.cpp:763` reads to decide whether to uplink. Flipping it `0 → 1` publishes to a public broker a message the sender marked private, and the sender never learns. The exploitable direction is the one that leaks; the reverse fails safe.
- **Re-address a broadcast.** `to` is unsigned, so a public broadcast can be delivered as an apparent private DM from the same signer.

## What changed

One fixed layout, replacing the conditional pair:

```
version(1) | from(4) | id(4) | to(4) | portnum(4) | request_id(4) | reply_id(4)
          | emoji(4) | bitfield(4) | flags(1) | payload(N)
```

`flags` carries `want_response` and `has_bitfield`. Nothing is added to the wire — the signing buffer is local scratch.

**Why the previous conditional scheme had to go.** It selected between two layouts based on whether the identifiers were nonzero, with nothing in the bytes recording which. Base was `header ‖ arbitrary payload`, so every byte string the extended layout could emit was also a legal base payload:

```
honest:    from|id|port| "ABCDEFGHhello"                        (12 + 13)
attacker:  from|id|port| 41424344 | 45464748 | "hello"          (12 + 8 + 5)
```

Identical bytes. Same signature verifies, the message now reads `"hello"` with a fabricated reply pointer. That's a *regression* the old approach introduced — today's payload is fully covered, so truncation breaks the signature. A marker placed only in the extended layout doesn't fix it, since base can always reproduce it. A fixed-length header does: the payload boundary is `total − XEDDSA_SIGNED_HEADER_LEN` and never depends on content.

**Why the field list, and not just `reply_id`.** `emoji` has to travel with `reply_id` or the reaction attack stays half-open. `want_response` has to travel with `bitfield`, because bit 1 mirrors it and `Router.cpp:1103` merges them with `|=` — signing either alone protects neither. The whole `bitfield` `uint32` is covered rather than its two defined bits, so 2–31 are protected before anything uses them, and presence is signed separately so stripping the field differs from sending it zero.

**Left out:** `dest` and `source` (one write in the tree, no readers), `channel` (the wire carries a hash where the decoded packet carries an index), and the hop fields, which relays rewrite by design.

**Signing the encoded `Data` wholesale isn't available.** A relay holding the channel key decodes and re-encodes `Data` before forwarding (`NextHopRouter::perhapsRebroadcast` copies an already-decoded packet, and `perhapsEncode` re-encodes it), so byte fidelity is lost and unknown fields are stripped. The byte-range excision `ack_proof` uses doesn't transfer either — `Routing` survives because it rides inside the opaque `payload`, which `Data` itself does not.

## Compatibility

**This changes every signature and is not compatible with the v2.8.0 alphas.** A verify failure against an authoritative key is an unconditional drop, not a downgrade to unverified, so signed broadcasts will not cross the version boundary in either direction.

That cost is being paid deliberately and now, because `v2.8.0.47db0e3` is still flagged prerelease and the newest stable release is 2.7.26 — the population signing today is alpha testers, not the installed base. The same change after 2.8.0 goes stable would split signed traffic across the mesh.

The earlier revision of this PR tried to preserve alpha compatibility with the conditional layout. That argument was wrong twice over: it claimed the common case was unaffected because "nothing broadcast carries `request_id`", which is true of `request_id` and misses `reply_id` — the broadcast case, and the motivating example — and the compatibility it bought came at the price of the forgery above.

## Note for review

`Data.payload` is still covered, as it always was. That makes `TRACEROUTE_APP` unsignable, since `TraceRouteModule::alterReceivedProtobuf` rewrites the payload at every hop. It survives today only because traceroute is unicast and unicasts are signed only in ham mode — where a signed traceroute would be mutated at the first hop and dropped at the destination. That looks like a live bug on develop, independent of this PR, and I have not touched it here.

## Testing

`./bin/run-tests.sh` — **1494/1494 test cases pass** across 80 suites.

The run grades AMBER on one pre-existing issue unrelated to this change: `test_waypoint_expiry` writes `.portduino/default/Waypoints_default.wpts` without a `test/state-manifest.tsv` entry. That gap exists on develop.

New coverage:

- `test_XEdDSA` now flips every covered field one at a time — payload, portnum, `from`, `id`, `to`, `request_id`, `reply_id`, `emoji`, `bitfield` value, `bitfield` presence, `want_response`.
- `test_XEdDSA_layout_is_unambiguous` performs the re-split above and asserts it does **not** verify.
- `test_XEdDSA_max_payload` signs a full 233-byte payload, because a buffer overflow makes `buildSigningBuffer` return 0 and signing fail *silently* — the failure mode worth a test rather than an argument.
- `E17`/`E18`/`E19` in `test_packet_signing` cover the receive path for the emoji flag, `OK_TO_MQTT` (granted and stripped), and a broadcast re-addressed as a DM.
- `test_XEdDSA_base_layout_cross_version_compat` is removed — there is no cross-version claim left to make.

Native suite only — no hardware testing.

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described. *(native test suite; see above)*
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

No hardware was available in this environment, so the device rows are unchecked. Given this changes the signature format, on-air testing against a mixed-version mesh is worth doing before merge.

## History

This PR previously also signed explicit acks under the Strict policy. That half was **withdrawn**: the motivating claim was that forged acks make relayers `stopRetransmission()`, but `ReliableRouter::perhapsGenerateImplicitAckForOwnOverheard` clears a pending retransmission on any overheard rebroadcast of our own `(from, id)` — header-only, no key, and it fires even on a PKI DM the originator cannot decrypt. Replaying the originator's own ciphertext stops their retries more cheaply than forging an ack, so no ack authentication of any kind closes that path.

The delivery-receipt property that *is* achievable landed separately as `Routing.ack_proof` (#11877, meshtastic/protobufs#1094).

Thanks to the reviewer whose audit found the implicit-ack path, and to CodeRabbit for independently finding the layout ambiguity on the previous revision.

https://claude.ai/code/session_012hLcVif8GDEmA2k77hmFG8


jp-bennett:
## The layout ambiguity is real, and the compatibility note in the description is wrong

CodeRabbit's finding on `CryptoEngine.cpp` holds up. `test_XEdDSA_base_layout_cross_version_compat` does not refute it — that test varies only the identifiers while holding the payload fixed, which isn't the transformation in question.

**The forgery.** An honest node signs a broadcast with no reply linkage, payload `"ABCDEFGHhello"`:

```
from | id | port | "ABCDEFGHhello"                        (12 + 13)
```

An attacker re-encodes the `Data` envelope — same `from`/`id`/`portnum`, same signature — as `request_id = 0x44434241` (`"ABCD"`), `reply_id = 0x48474645` (`"EFGH"`), `payload = "hello"`. The verifier sees nonzero identifiers and builds:

```
from | id | port | 41 42 43 44 | 45 46 47 48 | "hello"    (12 + 8 + 5)
```

Same bytes, so it verifies. The message now reads `"hello"`, with a fabricated reply pointer and the original sender's valid signature. On develop the payload is fully covered and this fails — so it is a regression this PR introduces. It needs a malleable channel, which the default non-AEAD channel is.

A marker added only to the extended layout does not fix this: base is `header ‖ arbitrary bytes`, so it can always reproduce whatever extended emits.

**Correction to the description.** It argues the bulk of signatures verify in both directions because "nothing broadcast carries `request_id`". That is true of `request_id` and misses `reply_id`, which is the broadcast case and the motivating example. 2.8.0 signs every broadcast, and a reply/tapback is a broadcast carrying `reply_id`. After this PR those are signed extended; a 2.8.0 node computes the base buffer, fails, and the failure path is `return false` — dropped, not merely unverified. Tapbacks vanish between versions in both directions.

**Two ways forward.**

```
today:  from(4) | id(4) | port(4) | payload(N)
A:      conditional — insert request_id(4) | reply_id(4) when either is nonzero
B:      V(1) | from(4) | id(4) | port(4) | request_id(4) | reply_id(4) | payload(N)   always
```

B is one fixed-length layout, so there is no second interpretation to collide with. Always including the identifiers is what kills the forgery; the leading version byte additionally separates the format from 2.8.0's (without it, zero identifiers collide with an old payload that starts with eight zero bytes) and makes the next change distinguishable. Neither option costs airtime — the signing buffer is never transmitted.

B breaks *all* cross-version verification, which is a wider blast radius than A. It is only attractive because 2.8.0 is still prerelease, so that window is open now and closes when 2.8.0 goes stable. I would take it.

Accepting either format on the verifier during a transition is a trap — the attacker picks whichever one verifies. Staging has to be sender-side.

**One thing I could not check:** the firmware never originates `reply_id` on outbound packets (only `StoreForwardModule` replaying history) — the phone sets it. If clients do not populate it on broadcast replies, the extended layout is near-dead code and both the forgery and the compatibility break shrink to ham mode. Worth confirming before picking.

---

PR #11429, opened 2026-08-12
fix(NodeDB): propagate SafeFile write failures out of saveProto()

h3lix1:
> [!IMPORTANT]
> **Do not merge before #11428** (`SafeFile`: remove a stale `.tmp` before opening it for write). On nRF52/STM32WL, propagating write failures before stale `.tmp` files are cleaned converts a leftover temp file into a full `fsFormat()`. These are separate PRs by request; the ordering constraint is real.

`saveProto()` captured the result of `SafeFile::close()` into `writeSucceeded`, used it only in a `LOG_ERROR` condition, and returned `okay` — which reflects `pb_encode()` alone.

`pb_encode()` sees only the byte stream, so three failures were reported to callers as successful saves:

1. readback XOR hash mismatch — bytes read back differ from those written
2. the tmp file cannot be reopened for the readback pass
3. `renameFile(tmp, real)` fails — the atomic commit never happens

All three mean the write did not land. Open failure and disk-full **already** failed `pb_encode()` (`writecb` checks `write() == count`), so those are unchanged — the scope is narrower than "everything `close()` checks".

This re-enables the recovery in `saveToDisk()`, which was added alongside the readback verification in 66c41e683 (#4397) and has been unreachable for these cases ever since. `SafeFile.h` documents `close()` as returning false "so higher level code can handle failures", and the encrypted-storage branch of this same function already propagates it.

### Residual risk a maintainer should weigh

`saveToDisk()` formats the whole filesystem but only re-saves `saveWhat`. Most callers pass a single segment, so a single-segment save failure wipes the *other* segments' files — channels (including PSKs), moduleConfig, devicestate, nodes.proto — plus uiconfig, rtttl, canned messages and backups. Live values survive in RAM, but anything not re-saved before the next reboot returns as defaults.

This is not new behaviour introduced here; it is behaviour that becomes *reachable* for genuine flash failures. It cannot loop: `saveToDiskNoRetry` runs at most twice, one format, no recursion.

On Portduino, `RECORD_CRITICALERROR` with a `FLASH_CORRUPTION_*` code calls `exit(2)`, terminating meshtasticd.

## Validation

---

| Check | XIAO | Heltec |
|---|---|---|

**Specific to this PR:** all saves in the sequence above returned success and persisted, confirming the propagation change does not produce false failures — the main regression risk, since a false failure would now trigger `fsFormat()`. The genuine failure conditions (readback hash mismatch, tmp reopen failure, rename failure) were not artificially induced and remain verified by inspection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved persistence reliability by detecting and reporting failures when saving data, reopening temporary files, verifying saved content, or completing file replacement.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -3021,6 +3021,10 @@ bool NodeDB::saveProto(const char *filename, size_t protoSize, const pb_msgdesc_
     if (!okay || !writeSucceeded) {
         LOG_ERROR("Can't write prefs");
     }
+
+    // pb_encode() only sees the byte stream, not the readback hash check, the tmp reopen or the
+    // rename - all of which SafeFile::close() reports. Propagate it so saveToDisk() can recover.
+    okay = okay && writeSucceeded;
 #else
     LOG_ERROR("Filesystem not implemented");
 #endif


caveman99 (changes requested):
Reviewed as part of a sweep of the PRs you opened over the last 48 hours. Refusing this one in its current form.

**The change is one line; the consequence is not.**

`saveProto()` currently returns `pb_encode()`'s result. Making it also return `writeSucceeded` re-enables the `saveToDisk()` recovery path — and that path calls `fsFormat()`, which erases the **whole** filesystem while re-saving only `saveWhat`. Most callers pass a single segment. So a single-segment write failure now destroys every file not in that segment: `channels.proto` (including PSKs), `module.proto`, `device.proto`, `nodes.proto`, plus uiconfig, rtttl, canned messages and backups.

You identify this yourself in the description and hand the decision to the maintainer. That makes it an unresolved design change filed as a bug fix, not a bug fix.

On Portduino it is worse: `RECORD_CRITICALERROR` with a `FLASH_CORRUPTION_*` code calls `exit(2)`, so this converts a silent write failure into a meshtasticd crash.

**What would make this mergeable, in this order:**

1. Fix `saveToDisk()` first — either re-save all segments after a format, or don't format at all and surface the failure to the caller. That is the actual defect; the discarded return value is downstream of it.
2. Then propagate. At that point the one-liner is safe and needs no caveat section.

**Also:**

- The hard ordering dependency on #11428 is real and correctly identified — which is itself a sign these are one change split into two PRs. Cross-PR ordering constraints are a cost the reviewer pays.
- No test. `saveProto()` failure is reachable natively through Portduino's FS.

I'd rather see this closed unless you want to take (1) on. The underlying observation is sound and worth keeping — just not as a change that ships a filesystem wipe behind a one-line return.

---

h3lix1:
Closing this in favour of fixing `saveToDisk()`, which is the right call.

@caveman99's read is correct: the discarded return value is downstream of the actual defect. `saveToDisk()` calls `fsFormat()` on any failure and then re-saves only `saveWhat`, so a single-segment write failure destroys every file outside that segment. Propagating the failure without fixing that first just makes the wipe easier to reach.

Two things I confirmed while checking the review, both of which strengthen it:

- The `exit(2)` is broader than described. `NodeDB.cpp:4586` fires on `FLASH_CORRUPTION_RECOVERABLE` as well as `_UNRECOVERABLE`, so on Portduino even a **successful** format-and-recover exits meshtasticd.
- Re-saving all segments after the format would not be sufficient on its own. `config`, `moduleConfig`, `channels` and `devicestate` are in RAM and can be rewritten, but uiconfig, rtttl, canned messages and backups are not, so a format loses them whatever we re-save. That points at not formatting for a single-file write failure at all: a readback hash mismatch on one file is not evidence the filesystem is corrupt, and the response currently outruns what the failure demonstrates.

Worth noting the `saveToDisk()` defect is already live and does not need this PR to reach it, since `saveToDiskNoRetry()` can fail today on unsafe power level, encode failure, open failure or a full disk. So it is worth fixing on its own merits rather than as scaffolding for a one-liner.

The replacement will be a single PR covering the `saveToDisk()` behaviour and the propagation together, with the Portduino-reachable tests that were missing here, rather than another ordered pair.

PR #11430, opened 2026-08-12
fix(NodeDB): recover owner fields using the key-derived node number

h3lix1:
When devicestate is discarded, `loadFromDisk()` tries to recover owner fields from our own NodeDB row:

```c
installDefaultDeviceState();
const meshtastic_NodeInfoLite *us = getMeshNode(getNodeNum());
```

`installDefaultDeviceState()` calls `pickNewNodeNum()`, whose own comment reads `// based on macaddr now`. But `config` — which holds `security.public_key`, the thing that determines our real node number — is not loaded until later in the same function. So the lookup probes a MAC-derived provisional number while our row lives at `crc32(public_key)`, and normally misses. Owner/call-sign/licence state is lost to factory defaults even though the correct values are on disk.

This is a regression from key-derived node numbers: before them, the MAC-derived number *was* the self key, so this safety net worked.

### Why not "just read index 0"

Self is pinned to index 0 by `nodeDBSelfCare()` — but that runs *after* `loadFromDisk()` returns, so at recovery time index 0 is only whatever the on-disk file's first record happens to be. Usually self; **not** for a legacy-migrated DB (order copied verbatim), a cloned image, a restored backup, or a test fixture. Trusting it would paste a stranger's `long_name` and `is_licensed` call-sign flag onto our owner and persist it — worse than the current miss. There is already an in-tree comment warning about exactly this.

### Approach

Defer the recovery until after config load and match on `crc32Buffer(config.security.public_key)` — the same derivation used elsewhere for our identity. Safe by construction: a DB that does not contain us is a clean miss, not a wrong-identity restore. Keyless/pre-PKI builds keep the old probe.

### Scope note

The miss applies to the **decode-failure / absent** devicestate path. On the `version < DEVICESTATE_MIN_VER` path devicestate decoded fine, so `my_node_num` survives and the original lookup does hit. The fix is correct for both.

## Validation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when device state is missing or invalid during startup.
  * Restores the correct device identity and owner information from saved records.
  * Preserves factory ownership when no matching record is available.
  * Supports recovery for devices with or without security keys.
  * Prevents incorrect numbering and safely limits restored device names.
  * Saves repaired state for future startups.
  * Improves contact recency tracking when the device clock is unavailable.
  * Updated Bluetooth and notification defaults for supported TFT devices.

---

| Check | XIAO | Heltec |
|---|---|---|

> [!NOTE]


thebentern (approved):

h3lix1:
Thanks — fixed in the latest push.

Verified the guideline is real before acting on it (`AGENTS.md:83`, `.github/copilot-instructions.md:339`). Both comment blocks are now two lines:

- `NodeDB.cpp` (deferral site) — keeps only the config-ordering reason: our real NodeNum comes from `config.security.public_key`, still unread at that point.
- `NodeDB.cpp` (recovery site) — keeps only the key-derived NodeNum reason, plus the one genuinely non-obvious consequence: matching on identity makes a `nodes.proto` that doesn't contain us a clean miss rather than a wrong-owner restore.

That second clause is the reason this PR doesn't simply read index 0, so I kept it — dropping it would leave the next reader wondering why the obvious approach wasn't taken. Everything else moved to the commit message.

caveman99 (changes requested):
Reviewed as part of a sweep of the PRs you opened over the last 48 hours.

**The defect is real**, and the reasoning for matching on `crc32(public_key)` rather than trusting index 0 is correct — legacy-migrated, cloned, restored-backup and test-fixture databases are exactly why index 0 can't be trusted before `nodeDBSelfCare()` runs.

**Three things before this lands.**

1. **CodeRabbit already asked for the comment fix and it wasn't done.** Its review on this PR flagged `src/mesh/NodeDB.cpp` L2448-2452 and L2566-2570, quoting the project rule verbatim: *"Keep code comments minimal—one or two lines maximum."* Both blocks are still there. `AGENTS.md:83` and `.github/copilot-instructions.md:340` are the canonical statement of that rule, and they are the first files an agent working in this repo is expected to read.

2. **The reordering is unanalysed.** You move the owner recovery roughly 135 lines later in `loadFromDisk()`, from just after `installDefaultDeviceState()` to just after the config load. Anything between those two points that reads `owner` now sees factory defaults where it previously saw recovered values. The PR doesn't enumerate what that is — that needs walking, not asserting.

3. **`clampLongName()` is a new call on this path**, not part of the reported defect. It may well be correct, but it's a behaviour addition riding along inside a fix. Call it out explicitly or split it.

**No test**, and this is natively testable: `test_nodedb_blocked` already carries a `NodeDB` shim with `seedSelf()` and hot-store push helpers. A discarded-devicestate plus populated-`nodes.proto` fixture pins both the hit and, more importantly, the clean miss — the property your "safe by construction" argument rests on.

---

h3lix1:
Test coverage added in `b62aa9fbc`, three cases for the recovery path.

The block was inline and could not be driven from a test, so it is now `NodeDB::recoverOwnerFromNodeDB()`. That extraction is behaviour preserving: same lookup, same field copies, same clamp, same return points. Only the call site changed, from an `if` body to a call.

The three cases:

- `test_ownerRecovery_restoresFromKeyDerivedRow`, the hit. A self row filed under the key-derived number is found and its owner fields are restored.
- `test_ownerRecovery_missKeepsFactoryOwner`, the clean miss. No matching row leaves the factory owner untouched and returns false.
- `test_ownerRecovery_keylessUsesProvisionalNum`, the pre-PKI path, where `config.security.public_key` is not 32 bytes and the lookup falls back to `getNodeNum()`.

Verified both directions under the `coverage` environment in Docker:

- as it stands: 11 tests, 0 failures
- with `selfNum` reverted to plain `getMeshNode(getNodeNum())`:
  `test_ownerRecovery_restoresFromKeyDerivedRow:FAIL: Expected TRUE Was FALSE` and
  `test_ownerRecovery_missKeepsFactoryOwner:FAIL: Expected FALSE Was TRUE`, 11 tests, 2 failures

Worth having both directions fail there: the first is the recovery that should happen and does not, the second is a recovery that should not happen and does, which is the half that would quietly overwrite a good owner record.

One correction on the earlier review. The comment-shortening request had already been done at the commit you reviewed: `284cfd11b` is literally "docs(NodeDB): shorten deferred owner-recovery comments" and is the head of this branch. Nothing further to trim there unless you want the remaining two-line blocks cut down as well.

h3lix1:

PR #11432, opened 2026-08-12
fix(NodeDB): stop addFromContact() erasing a stored public key

h3lix1:
`addFromContact()` reaches `TypeConversions::CopyUserToNodeInfoLite()`, which assigns `public_key` unconditionally. So a `SharedContact` with `has_user` set and an **empty** `public_key` erases a peer's stored, XEdDSA-proven key — and the erasure is persisted. Subsequent DMs to that peer then fail with `PKI_SEND_FAIL_PUBLIC_KEY`.

Per the comment in the same function, clients send `add_contact` before **every** text-message DM, making this the highest-volume key-write path on the device and the only one with no protection against erasure.

Fix: refuse only the erasure. If the entry already holds a 32-byte key and the incoming contact carries an empty/short one, keep the stored key.

### Deliberately NOT changed

- **No first-key-wins pin.** `updateUser()`'s refuse-on-mismatch is *not* replicated here. `add_contact` exists to supply keys the radio never had and to update them (QR contact sharing); a blanket pin would break that documented flow. A valid 32-byte contact key still replaces a stored one exactly as today.
- **No node-number validation.** Reserved/broadcast/self node numbers are still accepted here — separate concern.
- **`CopyUserToNodeInfoLite` untouched.** It has several other callers; the guard is at the `addFromContact` call site only.

### Is an empty key ever an intentional "clear"?

No evidence for it, and the field shape argues against it: `public_key` is a **singular, non-optional** bytes field, so "absent" and "explicitly empty" are indistinguishable on the wire — a client that merely lacks a key looks identical to one requesting a clear. The one intentional key-clear that did exist on this path (the `should_ignore` branch) was **removed** in #10705 and replaced with a comment stating an ignored peer keeps a usable identity. `test_traffic_management` asserts the same merge rule elsewhere: a keyless identity must not erase a key already learned.

## Validation

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved existing contact public keys when incoming contact information is incomplete.
  * Continued to accept and update contacts when a complete public key is provided.

---

| Check | XIAO | Heltec |
|---|---|---|

> [!NOTE]

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -3482,7 +3482,16 @@ void NodeDB::addFromContact(meshtastic_SharedContact contact)
         }
     }
     info->num = contact.node_num;
+    // CopyUserToNodeInfoLite assigns public_key unconditionally, and clients send add_contact before every
+    // DM - often from an entry that carries no key at all. A contact may still supply or update a full
+    // 32-byte key (that's what add_contact is for), but it must never *erase* a key we already hold, which
+    // would be persisted below and break subsequent DMs with PKI_SEND_FAIL_PUBLIC_KEY.
+    const meshtastic_NodeInfoLite_public_key_t storedKey = info->public_key;
     TypeConversions::CopyUserToNodeInfoLite(info, contact.user);
+    if (storedKey.size == 32 && info->public_key.size != 32) {
+        LOG_INFO("Contact 0x%08x has no key, keep the stored one", contact.node_num);
+        info->public_key = storedKey;
+    }
     if (contact.should_ignore) {
         // Block the contact and drop its rich satellite data, but keep the
         // public key copied above - an ignored peer keeps a usable identity


thebentern (approved):

h3lix1:
Thanks for merging — flagging that both review items landed on the branch **after** the merge, so they are not in `develop`:

- the `LOG_INFO` → `LOG_DEBUG` change (your point: `add_contact` precedes every DM, so that line emits one INFO record per message)
- the four regression cases you asked for

Both are now in **#11443**, cherry-picked onto current `develop`, along with the comment-length trim for `AGENTS.md`. Test matrix there covers all four directions: stored key survives an empty incoming key and a short one; a full 32-byte key still replaces; and a full key is stored when none was held. Each asserts `.size` and `.bytes` independently, since `CopyUserToNodeInfoLite` memcpy's all 32 bytes before setting size — restoring size alone would leave a zeroed array that still reads as a valid key.

PR #11433, opened 2026-08-12
fix(Router): reject reserved node numbers in the first-contact bootstrap

h3lix1:
`verifyFirstContactNodeInfo()` accepts a first-contact NodeInfo when `crc32Buffer(user.public_key) == p->from` and the signature verifies, then calls `getOrCreateMeshNode(p->from)` unconditionally. Neither that site nor `getOrCreateMeshNode()` applies the reserved-value guards `pickNewNodeNum()` applies to our own number. The only ingress reserved filter is `p->from == NODENUM_BROADCAST` — **it does not cover 0..3**.

`NODENUM_BROADCAST_NO_LORA` is `1`, and `isBroadcast(1)` is true. So an admitted entry with `num == 1` behaves as a broadcast address: `wouldEncryptWithPKC()` becomes false, **PKI is skipped and the shared channel PSK is used instead of Curve25519**, ack handling is bypassed, and `RadioLibInterface::send` drops the packet before LoRa TX.

An attacker reaches this by grinding a keypair whose public key CRC32s to a reserved value — a real but non-trivial cost. Identity binding alone does not exclude it, which is why the guard belongs at ingress.

Fix rejects reserved node numbers before any NodeDB entry exists, returning the existing `NodeInfoBootstrapResult::INVALID` so the caller's existing drop path handles it.

`NUM_RESERVED` and a new `isReservedNodeNum()` predicate move to `MeshTypes.h` beside the constants they encode; `pickNewNodeNum()`'s open-coded equivalent now uses the predicate (a literal identity, zero behaviour change).

### Reviewer notes

- The guard is placed **after** the portnum check on purpose: `test_mqtt` builds `TEXT_MESSAGE_APP` packets with `p.from = 1`, and checking before the portnum test would have broken them.
- Tests A14/A15/A16 in `test_packet_signing` use non-reserved node numbers and are unaffected.
- **Deliberately left unguarded:** what to do when *our own* derived number lands on a reserved value (regenerate the keypair? refuse to boot?) is an open design question; and `getOrCreateMeshNode()` itself is untouched — ~10 callers, wide blast radius.

## Validation

## Summary by CodeRabbit

* **Bug Fixes**
  * Reserved node numbers, including broadcast and low-number ranges, are consistently blocked for real peers.
  * Initial contacts using reserved sender numbers are rejected, while valid traffic remains supported.
  * New node-number assignment avoids reserved values.
  * Contact history and node eviction remain reliable when accurate clock time is unavailable.
  * Coordinate requests and event-channel messages are handled more reliably.
  * Malformed or undecodable packets are filtered and relayed more consistently.
  * Reliable message retry behavior and acknowledgment handling are improved.

---

| Check | XIAO | Heltec |
|---|---|---|

> [!NOTE]


thebentern (approved):

h3lix1:
Tests added in `342d813f2` — three cases in `test_packet_signing` group E, which is the group that drives `checkXeddsaReceivePolicy` directly.

**E14** — reserved senders `0`, `NODENUM_BROADCAST_NO_LORA`, `NUM_RESERVED - 1`, `NODENUM_BROADCAST` on a signed first-contact NodeInfo: each dropped, and `getMeshNode(from)` stays NULL so no entry is created.

**E15** — the boundary, asserted on `isReservedNodeNum` directly: `NUM_RESERVED` itself stays usable, `NUM_RESERVED - 1` does not.

**E16** — the non-NodeInfo case. Both an unsigned and a signed `TEXT_MESSAGE_APP` packet with `from == 1` still pass, pinning `test_mqtt`'s downlink fixtures.

One honest caveat worth recording: **E14 is a behaviour pin, not a fix-discriminating test.** Grinding a keypair whose public key CRC32s to a chosen reserved value is ~2^32 work, so an otherwise-valid reserved packet cannot be constructed in a unit test — without the guard those same packets are rejected one line later by the `crc32Buffer(...) != p->from` identity binding. E14's value is pinning the invariant so a future relaxation of that binding cannot silently reopen the hole. The tests that actually discriminate the change are **E15** (the predicate the guard calls) and **E16**'s signed variant (the only construction whose result flips if the guard were hoisted above the portnum check).

caveman99 (changes requested):
Reviewed as part of a sweep of the PRs you opened over the last 48 hours. The hardening is welcome; the collision is not.

**Direct conflict with #11422** (jp-bennett, "Sign explicit acks under Strict; bind request_id/reply_id in the XEdDSA signing buffer"):

- Both modify `verifyFirstContactNodeInfo()` in `src/mesh/Router.cpp`. #11422 changes the `crypto->xeddsa_verify(...)` call; you add an early return immediately above it. Adjacent hunks — semantically compatible, textually conflicting.
- Both add tests to `test/test_packet_signing/test_main.cpp` **at the same insertion point**, immediately after `RUN_TEST(test_E13_decoded_unsigned_nodeinfo_padded_inside_payload_dropped)`.
- Both claim the **E14 and E15** slots. #11422 adds `test_E14_decoded_signed_ack_retargeted_request_id_dropped` and `test_E15_decoded_signed_reply_retargeted_reply_id_dropped`; you add `test_E14_reserved_sender_first_contact_nodeinfo_dropped`, `test_E15_reserved_nodenum_boundary_excludes_num_reserved` and `test_E16_reserved_sender_non_nodeinfo_still_accepted`.

The function names differ, so this isn't a symbol clash — but it is a guaranteed merge conflict, and it would leave two different E14s and two different E15s in one suite. #11422 is older and changes the signing buffer format, so it lands first. Rebase onto it and take E16 onward.

**Narrow the claim.** Your E14 asserts four reserved senders are dropped, but two are already dropped upstream: `RadioLibInterface::handleReceiveInterrupt()` refuses `header.from == 0`, and `Router::perhapsHandleReceived()` refuses `p->from == NODENUM_BROADCAST`. The genuinely new coverage is `from == 1` (`NODENUM_BROADCAST_NO_LORA`) and `from == 2..3`. State that, rather than implying the guard closes four holes.

**Coding guidelines** (`AGENTS.md:83`): the `Router.cpp` comment is fine at two lines. The three test headers are 3–4 lines each — trim them.

The `NUM_RESERVED` hoist into `MeshTypes.h` behind an `isReservedNodeNum()` helper is a good cleanup and I'd keep it.

---

h3lix1:
**What it is.** `NUM_RESERVED` (`NodeDB.cpp:2021`) is enforced in exactly one place, `pickNewNodeNum()` at `:2047`, which covers only our own node number. Nothing checks numbers arriving from the mesh.

**What it does.** Any packet with `from = 1` reaches `getOrCreateMeshNode()` through `updateFrom` (`NodeDB.cpp:3681`, called on every RX packet at `MeshService.cpp:93`) and creates a node numbered 1, with no crc32 binding or signature needed. Because `NODENUM_BROADCAST_NO_LORA` is `1` and `isBroadcast(1)` is true (`NodeDB.cpp:681-684`), a DM to that entry has `want_ack` cleared (`Router.cpp:513`) and is dropped before TX (`RadioLibInterface.cpp:182`), so it fails invisibly while the client shows it sent. Only `1` matters: `0` is already filtered at every remote ingress, and `2`/`3` are inert.

**Scope.** Low-severity hardening, not a vulnerability. Fake named contacts are already plantable at ordinary node numbers, no key is exposed, retries self-cancel (`ReliableRouter.cpp:50`), the entry is removable, and the MQTT confidentiality angle needs a non-default gateway config.

**Fix.** Guarding only `verifyFirstContactNodeInfo()` leaves the cheap path open. `if (n < NUM_RESERVED) return NULL;` at the top of `getOrCreateMeshNode()` closes every write path, and every caller already handles a NULL return. The "wide blast radius" note in my PR description was wrong.

*Rewritten; an earlier version overstated the severity and named the wrong entry point.*

h3lix1:
Closing - there are much more interesting bugs that need fixing.

PR #11436, opened 2026-08-12
docs(nexthop): correct stale createNewIdentity() description

h3lix1:
`docs/nexthop-routing-reliability.md` stated that `createNewIdentity()` "retires the old entry by marking it **ignored** and clearing its pubkey", citing `src/mesh/NodeDB.cpp:3123-3125`.

Both halves are stale:

1. `createNewIdentity()` now calls `removeNodeByNum()` outright. The in-code comment explains why the ignored-flag approach was dropped: it left a keyless ghost of the node's own former identity that survived cleanup and eviction, was still streamed to clients, and made any DM or admin message aimed at it fail permanently with `PKI_SEND_FAIL_PUBLIC_KEY`.
2. The cited line numbers no longer point at that code.

This matters because the doc's mitigation reasoning rested on M1's candidate gate skipping *ignored* entries — a mechanism that no longer exists.

**Removal makes the claim stronger, not weaker.** The old wording depended on the `nodeInfoLiteIsIgnored` branch firing during the resolver's scan. `removeNodeByNum()` compacts the row out of `meshNodes` entirely — plus `eraseNodeSatellites()`, `warmStore.remove()` and `purgeNode()` — so the old identity is not in the scanned array at all.

Cites the function rather than line numbers, since those rot.

### Also verified, left unchanged

- "Node numbers are now immutable / identity-bound … cannot be resolved operationally by renumbering" — still accurate. `pickNewNodeNum()` still rerolls on conflict, but it runs before `createNewIdentity()` overwrites `my_node_num` with the key-derived value, so for any key-holding node the derived number wins.
- "CRC32 … uniformly distributed over 256 values" vs "255 buckets" elsewhere is not a contradiction: 256 is the raw CRC byte, 255 is after `getLastByteOfNodeNum()`'s `0x00`→`0xFF` projection.

Docs-only; no file under `src/` or `test/` touched.

## Validation

## Summary by CodeRabbit

* **Documentation**
  * Clarified how node numbers are derived from public keys.
  * Documented that identity changes re-derive node numbers.
  * Explained that old identity entries are fully removed during key rotation, preventing stale entries from appearing in resolution or client-visible state.

---

| Check | XIAO | Heltec |
|---|---|---|

### DIFF
--- a/docs/nexthop-routing-reliability.md
+++ b/docs/nexthop-routing-reliability.md
@@ -127,7 +127,8 @@ meshes, efficiency _is_ reliability.
 ### Note: pubkey-derived node numbers (develop / 2.8) - does not change the plan
 
 develop derives the node number from the public key:
-`my_node_num = crc32Buffer(public_key)`, set in `NodeDB::createNewIdentity()`
+`my_node_num = crc32Buffer(config.security.public_key.bytes, config.security.public_key.size)`,
+set in `NodeDB::createNewIdentity()`
 (`src/mesh/NodeDB.cpp`) and re-derived there on every key change. This **reinforces** the
 plan rather than changing it:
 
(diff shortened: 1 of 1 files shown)


h3lix1:
Both points addressed in the latest push — thanks, the second one was a genuine correctness issue.

**1. Exact `crc32Buffer()` inputs.** Taken. Now reads `crc32Buffer(config.security.public_key.bytes, config.security.public_key.size)`, matching `NodeDB::createNewIdentity()`.

**2. The one-byte value does not uniquely identify an identity.** You're right, and the doc contradicted itself: the bullet directly above already states the last byte is uniformly distributed over 256 values, so "keep a learned byte bound to one identity" read as uniqueness where only *temporal stability* was meant. Reworded:

> A learned byte stays associated with the same identity over time (good for M3 freshness) - it does not identify that identity uniquely, since 256 node numbers share each byte.

Stability is the property M3 freshness actually depends on, so this states the real claim rather than a stronger one.

For transparency on scope: both flagged lines were **pre-existing** — this PR's diff only replaced the stale `createNewIdentity()` description and its rotted line-number citations. I've fixed them anyway, since the PR is explicitly about the accuracy of this passage and leaving a known-imprecise claim in place would be odd.

caveman99:
Sorry, these files don't belong there in the first place, we're discussing moving or deleting them.

PR #11438, opened 2026-08-12
fix(nrf54l15dk): declare the real cpuapp RAM and flash limits

cvaldess:
Reopens #11416 with the noise removed: that branch was accidentally cut from my
fork's `develop` instead of `upstream/develop`, so GitHub showed 52 files when
the change is a **single 2-line edit to one file**. Sorry for the review time
that wasted. This branch is cut clean off `develop` and touches
`boards/nrf54l15dk.json` only.

---

`boards/nrf54l15dk.json` advertises 262144 B of RAM and 1572864 B of flash.
Neither is reachable by the application core:

- **RAM**: Nordic's devicetree (`zephyr/dts/vendor/nordic/nrf54l15.dtsi`) splits
  the 256 KB SRAM into `cpuapp_sram` = 188 KB (192512 B) plus 68 KB that stay
  reserved for the FLPR (RISC-V) coprocessor, whether or not an application
  uses it.
- **Flash**: `cpuapp_rram` = 1428 KB (1462272 B) out of the part's 1524 KB, the
  remaining 96 KB being the FLPR code partition. The old value of 1572864 B did
  not exist even on the whole chip.

These numbers only feed PlatformIO's size report and its overflow check (the
linker takes its real bounds from the devicetree), so **no binary changes**.
They did make the report understate usage: the `nrf54l15dk` build showed 46.3%
RAM while really using 63.1%, hiding 17 points of headroom that never existed.
Worse, a firmware growing past the true 188 KB but still under the declared
256 KB would spill into the FLPR's physical region without PlatformIO ever
flagging a RAM overflow.

Cross-checked against two independent sources: Nordic's `nrf54l15.dtsi`, and the
partition map an NCS/sysbuild build generates for this same board
(`sram_primary 0x2f000`, `flash_primary 0x165000` — byte for byte the same
values).

Build after the change: SUCCESS, RAM 63.1% (121399/192512), Flash 38.4%
(562052/1462272).

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated nRF54L15-DK upload limits to accurately reflect the available RAM and flash memory.


cvaldess:

PR #11440, opened 2026-08-12
fix(nrf54l15): implement reboot() and loop on short fs_write()

cvaldess:
Reopens #11418 with the noise removed: that branch was accidentally cut from my
fork's `develop` instead of `upstream/develop`, so GitHub showed 53 files when
the change is **22 lines across 2 files**, both `nrf54l15`-gated. Sorry for the
review time that wasted. This branch is cut clean off `develop` and touches
`src/Power.cpp` and `src/platform/nrf54l15/InternalFileSystem.h` only.

---

Two bugs in the nRF54L15 port, both found while chasing "settings do not survive
a reboot" on the DK.

**1. `Power::reboot()` has no `ARCH_NRF54L15` branch**, so it falls through to
the generic `#else` that only logs `FIXME implement reboot for this platform`
and returns. Any config change that requires a restart (LoRa region, for
instance) therefore tears down BLE in preparation for a reboot that never
happens, and the phone app is left waiting for a node that is never coming back.
Added a branch calling `sys_reboot(SYS_REBOOT_COLD)` — the same recovery path
this platform's crash handler already uses, and it lets Zephyr do its own
pre-reset bookkeeping (reset-reason register) before the CPU resets. No extra
include is needed; the symbol already arrives via `platform/nrf54l15/Arduino.h`.

**2. `InternalFileSystem.h` `File::write()` issues a single `fs_write()`** and
returns whatever it consumed. `fs_write()` may short-write, exactly like POSIX
`write()`. nanopb's stream callback treats any count short of what it asked for
as a fatal "io error" and aborts the encode, so a short write silently corrupts
a save. Small protobufs (a lightly populated NodeDB) always fit in one call and
never expose it; larger ones (DeviceState, LocalConfig) intermittently do not —
the observed symptom was `device.proto` and `config.proto` failing to save while
`nodes.proto`, `module.proto` and `channels.proto` saved fine. Now loops until
the whole buffer is written or a real error occurs.

Both fixes have been running on an nRF54L15-DK (PCA10156) in a separate
NCS/Zephyr build of this same source tree: config and BLE bonds survive hot
resets and power cycles, and a LoRa region change now actually reboots the node
instead of hanging the app.

Build: `nrf54l15dk` SUCCESS on a clean `develop` base, no size regression.

## Summary by CodeRabbit

- **New Features**
  - Added cold-reboot support for devices based on the nRF54L15 platform.

- **Bug Fixes**
  - Improved file-writing reliability by ensuring larger data buffers are fully written when possible, reducing the risk of incomplete file updates or corrupted stored data.


cvaldess:

PR #11443, opened 2026-08-12
fix(NodeDB): addFromContact follow-up - DEBUG log level, shorter comment, regression tests

h3lix1:
Follow-up to #11432, which merged before these three items landed on the branch. All were raised in review of that PR.

**1. Log the keyless-contact path at DEBUG.** Clients send `add_contact` before every text-message DM, so a contact carrying no key is an expected, high-frequency case. `LOG_INFO` there emits one record per DM. (CodeRabbit flagged this on #11432.)

**2. Shorten the comment to two lines**, per `AGENTS.md`: *"Keep code comments minimal - one or two lines, max."*

**3. Add the regression coverage** requested on #11432 — four cases in `test/test_nodedb_blocked`, which already builds a real `NodeDB` via its test shim:

| Stored key | Incoming key | Asserts |
|---|---|---|
| 32 × `0xA5` | empty | stored key survives |
| 32 × `0x5A` | 4 bytes | stored key survives |
| 32 × `0x11` | 32 × `0x22` | incoming replaces stored |
| none | 32 × `0x33` | incoming stored |

Each asserts `.size` **and** `.bytes` independently. That matters: `CopyUserToNodeInfoLite` memcpy's all 32 bytes and *then* sets size, so restoring size alone would leave a zeroed byte array that still reads as a valid key. The third case is the guard against over-correcting the original fix into a blanket pin — `add_contact` must still be able to update a key.

No manifest changes needed: `test_nodedb_blocked` already declares `nodes.proto` and `warm.dat` writes in `test/state-manifest.tsv`, and suite discovery is dynamic since #11413.

Behaviour of the merged fix is unchanged — this only lowers a log level, trims a comment, and adds tests.


h3lix1:
Superseeded in #11515 by @thebentern

PR #11446, opened 2026-08-12
fix(MeshService): address the rate-limit NAK with getFrom(), not mp->from

h3lix1:
`MeshService::sendRoutingErrorResponse()` passes `mp->from` straight through as the NAK destination.
Every other ack/nak call site uses `getFrom()`, which turns the unset sentinel 0 into our own node
number.

The only caller is the 2 second `TEXT_MESSAGE_APP` rate limit at `PhoneAPI.cpp:1865`. That path
returns false before `MeshService::handleToRadio()` reaches its `p.from = 0` normalisation, so
`from` is whatever the client put in the ToRadio packet.

If the client left it unset, `to` on the NAK is 0. `isToUs(0)` is false so `Router::sendLocal()`
does not deliver it to the phone, and the client gets no `ROUTING_APP` response at all for the
message that was just rejected, only a `QueueStatus`. The NAK is instead encrypted and transmitted
over LoRa addressed to node 0. It goes out with the default `hopLimit` of 0 so nothing rebroadcasts
it, which keeps the cost to one wasted transmission plus a client left without a verdict.

Since `getFrom()` only rewrites 0, none of this shows up when the client sets `from` to its own node
number, which both first party clients normally do. On Android `CommandSenderImpl.kt` uses
`from = nodeManager.myNodeNum.value ?: 0`, so it degrades to 0 during the window before `myNodeNum`
is populated. I have not reproduced that: it needs two text sends inside the 2 second window during
startup. The convention of leaving `from` unset is real and used by first party code, for instance
`AccessoryManager+Lockdown.swift` carries the comment "meshPacket.from intentionally NOT set. Proto
default 0 means firmware...". That path does not send text so it cannot trigger this rate limit, but
it does show unset `from` is expected rather than malformed.

Fix is `getFrom(mp)`, matching the other call sites. No behaviour change when `from` is already set.

`sendRoutingErrorResponse()` had no native coverage before this change; `test_mesh_module` now
carries three cases for it. `test_routingErrorResponse_unsetFromAddressesUs` is the one that holds
the fix in place: it builds the request the rate limit rejects with `from` unset and asserts the NAK
is addressed to us rather than node 0, plus the request id and channel the client needs to match the
verdict to its message. Reverting to `mp->from` fails it with `Expected 286331153 Was 0`.

The other two state what the fix does not change, and pass either way: a sender that is already set
is passed through untouched, and a null packet generates no NAK. Suite is 27/27.

Testing: builds clean on `heltec-v4` and `seeed-xiao-s3`, targeted native suites unchanged. Not
exercised on hardware, since the CLI sets `from` and I could not construct the failing case with it.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected routing error responses so acknowledgments and negative acknowledgments reach the intended destination, including local requests.
  * Preserved sender information when it is already set.
  * Prevented responses from being generated for missing packets.
  * Ensured error responses retain the relevant error, request, and channel details.
  * Improved reliability for routing error handling across local and remote requests.


RCGV1 (commented):
I independently traced the unset-`from` rate-limit path and agree with this fix. `sendRoutingErrorResponse()` is reached before `handleToRadio()` normalizes phone-originated packets, so using `mp->from` can address the NAK to node 0. `getFrom(mp)` is the correct normalization and matches the other ACK/NAK call sites.

I also exercised a focused native regression that constructs a phone request with `from = 0`, sends `RATE_LIMIT_EXCEEDED`, and asserts the generated ACK/NAK targets the local node while preserving the request ID and channel. It passes on this PR head; the complete `test_mesh_module` suite passed 25/25.

I recommend adding that regression to this PR so the pre-normalization behavior stays covered. The essential assertion is that `ackNaks[0].to == LOCAL_NODE` after calling `sendRoutingErrorResponse(..., &request)` with `request.from = 0`.

h3lix1:
Thanks, that is a useful review, and the regression is in as `6fbd1d9b9`.

Your reading of the ordering is right, and the assertion you named is the one that discriminates. Reverting the fix to `mp->from` fails it with `Expected 286331153 Was 0`, `0x11111111` being the harness's `LOCAL_NODE`, so it does hold the fix in place rather than passing either way.

I added the id and channel assertions you mentioned for the same reason: those are what let the client match the verdict to the message it just sent, so a NAK that arrives correctly addressed but unmatchable would still leave the client without an answer.

Two more went in alongside it, to state the parts the fix does **not** change:

- `test_routingErrorResponse_setFromIsPassedThrough`, since `getFrom()` only rewrites `0` and the PR claims no behaviour change when the sender is already set. Worth pinning rather than asserting in prose.
- `test_routingErrorResponse_nullPacketGeneratesNothing`, covering the existing null guard.

Both pass with or without the fix, which is deliberate: they describe the surrounding contract, and only the unset-`from` case is evidence for the change itself.

`test_mesh_module` is 27/27 under the `coverage` environment. The count differs from your 25 because of the three added here.

One correction to the PR description, which I will update: it says there is no native coverage for this path either before or after, which was true when I wrote it and is now only true of "before".

PR #11455, opened 2026-08-12
fix(graphics): crash and leak fixes across display drivers

thebentern:
Five independent fixes in the display layer, found during a memory audit:

## TFTDisplay (portduino): uninitialized touch pointer dereference for GT911

`_touch_instance` is an uninitialized member, the guard `if (portduino_config.touchscreenModule)` accepts any configured module, but the branches only assign the pointer for `xpt2046`/`stmpe610`/`ft5x06`. The config enum also permits `gt911` (supported by the color-UI path in `tftSetup.cpp`), so `Module: GT911` with a non-color display mode called `config()` through an indeterminate pointer in the `LGFX` constructor. Initialize to `nullptr` and guard the config block — GT911 users get no touch on this path (as before at best), not a crash.

## Screen: destructor leaked the display driver and UI

`~Screen()` freed `normalFrames` but not the owned `dispdev` (display driver incl. framebuffer) or `ui`. Screen genuinely gets destroyed — it's held in `std::unique_ptr` and reset on the portduino reboot path in `Power.cpp`. `OLEDDisplay` has a virtual destructor, so `delete dispdev` through either declared type is safe.

## EInkDisplay2: nine orphaned low-level driver allocations

`GxEPD2_BW`'s constructor takes the driver **by value** and stores a copy (`GxEPD2_Type epd2` member), so every `auto lowLevel = new EINK_DISPLAY_MODEL(...)` was copied and orphaned as soon as `connect()` returned — a one-time boot leak at nine sites, and misleading ownership for readers. Pass temporaries instead, exactly like `GxEPD2Multi.h` already does.

## EInkParallelDisplay: task-exit ordering race

The async refresh task cleared `asyncFullRunning` **before** nulling `asyncTaskHandle` and self-deleting. The destructor waits on the flag, then `vTaskDelete(asyncTaskHandle)` — in the window it could delete a freed TCB. Null the handle first; `eink/Drivers/EInkParallel.cpp` carries the identical fix with the comment "Handle first: once asyncRunning reads false, no other thread may touch task state." Latent today (this destructor only runs once Screen deletes its display, per the fix above) but wrong ordering regardless.

## Panel_sdl: initFrameBuffer reported success on failed allocations

Only the first of three allocations was null-checked; on `framebuffer == NULL` the function filled `_lines_buffer` with `NULL+offset` garbage pointers and still returned `true`, converting an OOM into a wild write in the next redraw. Simulator-only code, but now it checks all three, frees partial allocations, and returns `false`.

## Testing

`trunk fmt` clean. The EInkDisplay2 change is mechanical across the nine `#ifdef` arms; CI builds the affected targets (T-Echo family, ThinkNode M1/M5/M8, MeshLink, RAK4630, Heltec e-ink family, PCA10059, M5 CoreInk, T-Deck Pro, Mesh Pocket, Wio Tracker L1).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved display initialization reliability across supported screen types.
  * Prevented crashes when touchscreen hardware is unsupported or unavailable.
  * Added safer handling for framebuffer allocation failures.
  * Fixed asynchronous display task shutdown state handling.
  * Prevented memory leaks when closing screens.

### DIFF
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -652,6 +652,10 @@ Screen::Screen(ScanI2C::DeviceAddress address, meshtastic_Config_DisplayConfig_O
 Screen::~Screen()
 {
     delete[] graphics::normalFrames;
+    // Owned by the constructor; Screen is genuinely destroyed on the portduino reboot path
+    // (screen = nullptr in Power.cpp), which previously leaked the display and UI objects.
+    delete ui;
+    delete dispdev;
 }
 
 /**
--- a/src/graphics/EInkDisplay2.cpp
+++ b/src/graphics/EInkDisplay2.cpp
@@ -161,9 +161,9 @@ bool EInkDisplay::connect()
 #if defined(TTGO_T_ECHO) || defined(ELECROW_ThinkNode_M1) || defined(T_ECHO_LITE) || defined(TTGO_T_ECHO_PLUS) ||                \
     defined(ELECROW_ThinkNode_M8)
     {
-        auto lowLevel = new EINK_DISPLAY_MODEL(PIN_EINK_CS, PIN_EINK_DC, PIN_EINK_RES, PIN_EINK_BUSY, SPI1);
-
-        adafruitDisplay = new GxEPD2_BW<EINK_DISPLAY_MODEL, EINK_DISPLAY_MODEL::HEIGHT>(*lowLevel);
+        // GxEPD2_BW stores a copy of the driver, so pass a temporary instead of leaking a heap object
+        adafruitDisplay = new GxEPD2_BW<EINK_DISPLAY_MODEL, EINK_DISPLAY_MODEL::HEIGHT>(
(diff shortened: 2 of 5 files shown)


caveman99 (approved):

src/graphics/Screen.cpp
```diff
@@ -652,6 +652,10 @@ Screen::Screen(ScanI2C::DeviceAddress address, meshtastic_Config_DisplayConfig_O
 Screen::~Screen()
 {
     delete[] graphics::normalFrames;
+    // Owned by the constructor; Screen is genuinely destroyed on the portduino reboot path
```
mverch67: On the portduino reboot path there won't be any actual leakage as the kernel releases all memory and restarts the application with a new pid and new isolated memory.

All the inserted deletes just add to the used flash memory that's why most arduino libraries do not implement destructors.

PR #11384, opened 2026-08-10
feat(stm32wl): add ST Nucleo-WL55JC variant

ndoo:
STM32 Nucleo-64 development board with STM32WL55JC MCU, SMPS, supports Arduino and ST morpho connectivity.

https://www.st.com/en/evaluation-tools/nucleo-wl55jc.html

I am actually thinking this board should be the first candidate for STM32WL to be promoted to a release-level board, but what do you think?

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [x] Other (please specify below)
    - [x] ST NUCLEO-WL55JC1 (MB1389E-02)

## Summary by CodeRabbit

- **New Features**
  - Added support for the STM32 Nucleo-WL55JC board, including connectivity, battery monitoring, clock, LED, button, and radio configuration.
  - Added RF switch modes for idle, receive, low-power transmit, and high-power transmit operation.
- **Bug Fixes**
  - Improved initialization of the optional LoRa status LED on compatible boards.
  - Prevented duplicate LoRa LED configuration during board startup.

PR #11456, opened 2026-08-12
fix(platform): OOM null-write in stm32wl File and exception leak in portduino GPIO init

thebentern:
Two small platform-layer fixes from a memory audit:

## stm32wl `File::_open_dir` — unchecked malloc + strcpy

The `_dir_path` allocation was the only unchecked `rtos_malloc` in the file (both `_file` and `_dir` are checked), and it's followed immediately by `strcpy`, so OOM became a NULL write. Worse, the object was left half-initialized — dir open, `_is_dir` already true, null path — and `openNextFile()` would later call `strlen(NULL)`. Now checked, unwinding the already-opened dir exactly like the `_dir` failure path above it; `_is_dir` is only set once the state is complete.

## portduino `initGPIOPin` — pin object lost on exception

The body is wrapped in `try/catch(...)`; if `setSilent()` or `gpioBind()` threw after `new LinuxGPIOPin(...)`, the catch block logged and returned with the object unreachable. Hold it in a `unique_ptr` and `release()` only after `gpioBind` has registered it.

(A third candidate from the audit — the dropped `new BQ25713()` pointer in `main-nrf52.cpp` — was left alone: no in-tree variant defines `BQ25703A_ADDR`, so the change couldn't be compile-verified.)

`trunk fmt` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when initializing Linux GPIO pins.
  * Added safer handling for directory path allocation failures, ensuring resources are properly released and directory state remains consistent.

### DIFF
--- a/src/platform/portduino/PortduinoGlue.cpp
+++ b/src/platform/portduino/PortduinoGlue.cpp
@@ -845,10 +845,10 @@ int initGPIOPin(int pinNum, const std::string &gpioChipName, int line)
     std::string gpio_name = "GPIO" + std::to_string(pinNum);
     std::cout << "Initializing " << gpio_name << " on chip " << gpioChipName << std::endl;
     try {
-        GPIOPin *csPin;
-        csPin = new LinuxGPIOPin(pinNum, gpioChipName.c_str(), line, gpio_name.c_str());
+        auto csPin = std::make_unique<LinuxGPIOPin>(pinNum, gpioChipName.c_str(), line, gpio_name.c_str());
         csPin->setSilent();
-        gpioBind(csPin);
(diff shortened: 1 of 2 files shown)


thebentern:
@ndoo

ndoo:
> @ndoo 

Good finding. Will test on hardware today.

ndoo:
Flashed to a Nucleo-WL55JC (cherry-pick #11384) and so far flash works (settings and nodedb loaded).

Happy to let it soak for 24h if you want, but to simulate actual fs corruption might be a bit hard (especially now with safe writes)

PR #11457, opened 2026-08-12
refactor(inkhud): make tile and framebuffer ownership explicit

thebentern:
Follow-up to the memory-audit fix series — this one is a no-behavior-change ownership cleanup in NicheGraphics/InkHUD.

## `userTiles` → `std::vector<std::unique_ptr<Tile>>`

`createUserTiles()` is a genuinely repeated runtime path (every rotation change, tile-count change, and on-screen keyboard open/close) and contained the only manual delete/new cycle in the subsystem. The pairing was correct, but only by convention — any future early return in the rebuild sequence would have leaked. Applets keep raw non-owning back-pointers, as before.

## `~Tile()` breaks the reciprocal Tile↔Applet link

`Tile::assignApplet()` creates a two-way link, but deleting a tile left the applet's `assignedTile` dangling until `findOrphanApplets()` happened to relink it. There's a latent hazard behind that: `AppSwitcherApplet` borrows a user tile while foreground, and only the Menu applet is restored after a rebuild — today unreachable (AppSwitcher locks input while shown), but the invariant was undocumented. The destructor now clears the applet's back-pointer and the static `highlightTarget` when they point at the dying tile.

## `Renderer::imageBuffer` / `UC8175::previousBuffer` → `unique_ptr<uint8_t[]>`

One-time setup allocations that live for the process lifetime — not leaks, just self-documenting ownership now.

## Testing

`trunk fmt` clean; building the t114 InkHUD target locally alongside CI.

### DIFF
--- a/src/graphics/niche/InkHUD/Renderer.cpp
+++ b/src/graphics/niche/InkHUD/Renderer.cpp
@@ -41,7 +41,7 @@ void InkHUD::Renderer::setDriver(Drivers::EInk *driver)
     imageBufferHeight = driver->height;
 
     // Allocate the image buffer
-    imageBuffer = new uint8_t[imageBufferWidth * imageBufferHeight];
+    imageBuffer = std::make_unique<uint8_t[]>(imageBufferWidth * imageBufferHeight);
 }
 
 // Set the target number of FAST display updates in a row, before a FULL update is used for display health
@@ -236,7 +236,7 @@ void InkHUD::Renderer::render(bool async)
 
         // Tell display to begin process of drawing new image
         LOG_INFO("Updating display");
-        driver->update(imageBuffer, updateType);
+        driver->update(imageBuffer.get(), updateType);
 
         // If not async, wait here until the update is complete
         if (!async)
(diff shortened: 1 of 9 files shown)


HarukiToreda (approved):

src/graphics/niche/InkHUD/Renderer.cpp
```diff
@@ -312,7 +312,7 @@ void InkHUD::Renderer::clearTile(Tile *t)
 
     // Clear the pixels
     if (xStart == 0 && xEnd == driver->width) { // full width box is easier to clear
-        memset(imageBuffer + (yStart * imageBufferWidth), 0xFF, (yEnd - yStart) * imageBufferWidth);
+        memset(imageBuffer.get() + (yStart * imageBufferWidth), 0xFF, (yEnd - yStart) * imageBufferWidth);
```
caveman99, line 315: cppcheck false positive - &imageBuffer[yStart * imageBufferWidth] instead is not flagged. Same a few lines down, the 2nd finding could be &imageBuffer[i + byteStart]

src/graphics/niche/InkHUD/Renderer.cpp
```diff
@@ -41,7 +41,7 @@ void InkHUD::Renderer::setDriver(Drivers::EInk *driver)
     imageBufferHeight = driver->height;
 
     // Allocate the image buffer
-    imageBuffer = new uint8_t[imageBufferWidth * imageBufferHeight];
+    imageBuffer = std::make_unique<uint8_t[]>(imageBufferWidth * imageBufferHeight);
```
caveman99, line 44: zeroing the buffer. the next call is clearBuffer which FF's it. Same pattern as in #11461, just not as obvious, easy save on one pass.

caveman99 (changes requested):

PR #11458, opened 2026-08-12
refactor(io): unique_ptr ownership for motion sensors and I2C keyboard

thebentern:
Second ownership-cleanup PR from the memory audit. No behavior change.

## Motion threads

`AccelerometerThread` and `MagnetometerThread` owned their `MotionSensor` via raw pointer with manual `delete`+null in `clean()`, called from the destructor, the init-failure path, and `copy()`. The member becomes `std::unique_ptr<MotionSensor>` (`MotionSensor`'s destructor is virtual, so the polymorphic delete is unchanged) and `clean()` reduces to `sensor.reset()`.

## KbI2cBase

The TCA-family keyboard was held as a **reference member bound to an anonymous `new`** in the constructor initializer list — ownership invisible, unfreeable by construction. It becomes `unique_ptr<TCA8418KeyboardBase>` with an out-of-line `= default` destructor (the base is forward-declared in the header; `TCA8418KeyboardBase` has a virtual destructor).

## GeoCoord::pointAtDistance

Returned `std::shared_ptr<GeoCoord>` — the only `shared_ptr` in the tree — with no shared ownership anywhere and currently no callers. `GeoCoord` is a small value type; return by value.

## Deliberately not converted

`DFRobotGravitySensor::gravity` was flagged in the audit, but the vendored sensor class has virtual methods without a virtual destructor, so its delete needs the existing pragma suppression; moving that into a custom deleter is more code than the destructor it would remove.

## Testing

`trunk fmt` clean; native-macos integration build covers kbI2cBase, motion threads, and GeoCoord.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard and motion-sensor resource management for more reliable startup and shutdown.
  * Enabled safe cleanup of keyboard devices through their base interface.
* **Refactor**
  * Updated internal device ownership handling to reduce manual cleanup.
  * Removed unused geographic coordinate distance and destination helpers.

### DIFF
--- a/src/gps/GeoCoord.h
+++ b/src/gps/GeoCoord.h
@@ -113,7 +113,7 @@ class GeoCoord
     static double toDegrees(double r);
 
     // Point to point conversions
-    std::shared_ptr<GeoCoord> pointAtDistance(double bearing, double range);
+    GeoCoord pointAtDistance(double bearing, double range) const;
 
     // Lat lon alt getters
     int32_t getLatitude() const { return _latitude; }
--- a/src/gps/GeoCoord.h
+++ b/src/gps/GeoCoord.h
@@ -4,7 +4,6 @@
 #include <cstdint>
 #include <cstring>
 #include <math.h>
-#include <memory>
 #include <stdexcept>
 #include <stdint.h>
 #include <string>
@@ -103,7 +102,6 @@ class GeoCoord
     static void convertWGS84ToOSGB36(const double lat, const double lon, double &osgb_Latitude, double &osgb_Longitude);
     static float latLongToMeter(double lat_a, double lng_a, double lat_b, double lng_b);
     static float bearing(double lat1, double lon1, double lat2, double lon2);
-    static float rangeMetersToRadians(double range_meters);
     static unsigned int bearingToDegrees(const char *bearing);
     static const char *degreesToBearing(unsigned int degrees);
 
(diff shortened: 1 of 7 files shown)


src/gps/GeoCoord.h
```diff
@@ -113,7 +113,7 @@ class GeoCoord
     static double toDegrees(double r);
 
     // Point to point conversions
-    std::shared_ptr<GeoCoord> pointAtDistance(double bearing, double range);
```
caveman99, line 116: should be deleted, not converted. It has zero callers in this repo or device-ui. It is also numerically wrong at both ends, so nothing could have been using it correctly:

src/gps/GeoCoord.h
caveman99: shared_ptr is gone, so <memory> can go too. If somebody downstream needs this, it can be included there.

caveman99 (changes requested):
see inline comments.

thebentern:
Deleted rather than converted — you're right that conversion was the timid option. Removed `pointAtDistance`, the `<memory>` include, and also its only helper `rangeMetersToRadians`, which was the other numerically-wrong end (it multiplies meters by 1852, i.e. treats meters as nautical miles). If anyone ever needs a destination-point function, better to reimplement it correctly than resurrect this one.

caveman99 (approved):

PR #11461, opened 2026-08-12
refactor(graphics): unique_ptr ownership for display buffers

thebentern:
Final ownership-cleanup PR from the memory audit series. No behavior change.

- **TFTDisplay**: `linePixelBuffer`/`repaintChunkBuffer` become `unique_ptr<uint16_t[]>`. The `new (std::nothrow)` allocation preserves the existing OOM log-and-fail path; the destructor's manual free block disappears (its `memaudit` bookkeeping stays).
- **BaseUIEInkDisplay**: `panelBuffer` becomes `unique_ptr<uint8_t[]>`; the user-written destructor is gone.
- **EInkParallelDisplay**: `epaper` and the ghosting `dirtyPixels` buffer become `unique_ptr` (`make_unique` zero-initializes, matching the old `calloc`); the destructor keeps only its async-task wait logic.

`trunk fmt` clean; native-macos integration build (TFTDisplay path) passes locally. E-ink targets covered by CI.

## Summary by CodeRabbit

* **Reliability**
  * Improved automatic management of display memory buffers.
  * Reduced the risk of memory leaks and invalid cleanup during display operation and shutdown.
  * Improved handling when touchscreen hardware is unavailable or unsupported.
* **Compatibility**
  * Display rendering, buffer updates, and e-paper behavior remain unchanged.

### DIFF
--- a/src/graphics/BaseUIEInkDisplay.cpp
+++ b/src/graphics/BaseUIEInkDisplay.cpp
@@ -22,7 +22,8 @@ BaseUIEInkDisplay::BaseUIEInkDisplay(Drivers::EInk *driver, uint8_t rotation) : 
     // Panel-native row-major buffer
     panelRowBytes = ((driver->width - 1) / 8) + 1;
     panelBufferSize = panelRowBytes * driver->height;
-    panelBuffer = std::make_unique<uint8_t[]>(panelBufferSize);
+    // plain new (no value-init): the memset below fills the buffer anyway
+    panelBuffer.reset(new uint8_t[panelBufferSize]);
     memset(panelBuffer.get(), 0xFF, panelBufferSize); // All white
 }
 
--- a/src/graphics/BaseUIEInkDisplay.cpp
+++ b/src/graphics/BaseUIEInkDisplay.cpp
@@ -22,13 +22,8 @@ BaseUIEInkDisplay::BaseUIEInkDisplay(Drivers::EInk *driver, uint8_t rotation) : 
     // Panel-native row-major buffer
     panelRowBytes = ((driver->width - 1) / 8) + 1;
     panelBufferSize = panelRowBytes * driver->height;
-    panelBuffer = new uint8_t[panelBufferSize];
-    memset(panelBuffer, 0xFF, panelBufferSize); // All white
-}
-
-BaseUIEInkDisplay::~BaseUIEInkDisplay()
-{
-    delete[] panelBuffer;
+    panelBuffer = std::make_unique<uint8_t[]>(panelBufferSize);
+    memset(panelBuffer.get(), 0xFF, panelBufferSize); // All white
 }
 
 bool BaseUIEInkDisplay::connect()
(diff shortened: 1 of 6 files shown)


src/graphics/EInkParallelDisplay.cpp
```diff
@@ -42,19 +42,13 @@ EInkParallelDisplay::EInkParallelDisplay(uint16_t width, uint16_t height, EpdRot
     // allocate dirty pixel buffer same size as epaper buffers (rowBytes * height)
     size_t rowBytes = (this->displayWidth + 7) / 8;
     dirtyPixelsSize = rowBytes * this->displayHeight;
-    dirtyPixels = (uint8_t *)calloc(dirtyPixelsSize, 1);
```
caveman99, line 45: calloc returns a nullptr on failure, std::make_unique throws bad_alloc. I'd rather have a node without a working display than an full stop crash. dirtyPixels.reset(new (std::nothrow) uint8_t[dirtyPixelsSize]()); should do the trick to conserve old behaviour, exactly as in the TFTDisplay Code.

src/graphics/BaseUIEInkDisplay.cpp
```diff
@@ -22,13 +22,8 @@ BaseUIEInkDisplay::BaseUIEInkDisplay(Drivers::EInk *driver, uint8_t rotation) :
...
     panelBufferSize = panelRowBytes * driver->height;
-    panelBuffer = new uint8_t[panelBufferSize];
-    memset(panelBuffer, 0xFF, panelBufferSize); // All white
-}
-
-BaseUIEInkDisplay::~BaseUIEInkDisplay()
-{
-    delete[] panelBuffer;
+    panelBuffer = std::make_unique<uint8_t[]>(panelBufferSize);
+    memset(panelBuffer.get(), 0xFF, panelBufferSize); // All white
```
caveman99, line 27: std::make_unique zeroes the buffer when the next line fills it with 0xFF. replacing the make_unique line with panelBuffer.reset(new uint8_t[panelBufferSize]); saves us a pointless pass over the buffer.

caveman99 (changes requested):

thebentern:
Good catches, both addressed in 717f8f9's follow-up:

- `dirtyPixels` now uses `new (std::nothrow) uint8_t[dirtyPixelsSize]()` — zeroed like the old calloc, nullptr on OOM instead of `bad_alloc`, and the render paths keep their null checks. Matches the TFTDisplay pattern.
- `panelBuffer` switched to plain `new` so we skip make_unique's value-init pass right before the 0xFF memset.

Agreed on headless-over-crash — that was an unintended semantic change on my part.

src/graphics/TFTDisplay.cpp
```diff
@@ -1279,7 +1271,7 @@ void TFTDisplay::display(bool fromBlank)
                 y_byteIndex = (y / 8) * displayWidth;
                 y_byteMask = (1 << (y & 7));
 
-                uint16_t *chunkRow = repaintChunkBuffer + (row * displayWidth);
+                uint16_t *chunkRow = repaintChunkBuffer.get() + (row * displayWidth);
```
caveman99, line 1278: Same false positive as #11457 -> uint16_t *chunkRow = &repaintChunkBuffer[row * displayWidth];

caveman99 (changes requested):
This needs to go in after #11455 since it fixes free-before-async-wait ordering left by thins one.

PR #11465, opened 2026-08-13
Actions: Add PIO caching to build-debian-src workflow

vidplace7:
Prevent a few more transient fails by caching `build-debian-src`

## Summary by CodeRabbit

* **Chores**
  * Improved automated build caching for more efficient build runs.
  * Build artifacts are now restored whenever available, while new cache data is saved only for eligible default-branch pushes.
  * This reduces unnecessary cache updates, avoids redundant work, and helps keep development builds consistent across automated runs.


vidplace7:
This won't work as written :/ revisiting

PR #11467, opened 2026-08-13
fix(TraceRoute): only learn a next hop from an RF-delivered reply

h3lix1:
`maybeSetNextHop()` is guarded by an anti-forgery check. The route array is unauthenticated payload,
so we only trust it when the node it names as our next hop is the one that relayed the reply to us.
The comment justifies skipping the transport question with "relay_node is 0 for MQTT-sourced
packets".

That holds for MQTT, which allocates zeroed and never copies `relay_node`. It does not hold for UDP
multicast. `UdpMulticastHandler` sets `transport_mechanism`, clears `pki_encrypted`, and clears
`rx_snr` / `rx_rssi` / `has_rx_rssi`, but leaves `relay_node` at whatever the frame carried. So a
peer on the same LAN can send a traceroute reply whose route array names any node as our next hop,
set `relay_node` to match, and pass the check with no RF evidence behind it.

The route it installs also escapes the M3 route-health machinery. `maybeSetNextHop` writes
`NodeInfoLite.next_hop` and the TMM overflow cache directly and never calls `noteRouteLearned`, so
there is no `RouteHealth` record for the staleness or failure decay in `getNextHop()` to act on. It
sits there until something else overwrites it.

The fix gates on `TRANSPORT_LORA` before the `relay_node` comparison.

The alternative would be to clear `relay_node` on UDP ingress alongside the other RF-derived fields
it already clears, which fixes every consumer rather than just this one. I did not do that here:
`relay_node` also feeds PacketHistory relayer tracking and Router's `resolveUniqueLastByte` call, and
moving those in the same patch as a security gate seemed like a bad trade. Worth doing separately if
you agree with the direction.

The second commit covers this in `test_traceroute_nexthop`. Two changes there:

`makeResponse()` was building packets with `transport_mechanism` left at its zero value, which is
`TRANSPORT_INTERNAL`. Nothing off the radio looks like that, since
`RadioInterface::deliverToReceiver()` stamps `TRANSPORT_LORA` on every reception before the router
sees the packet, so the helper now sets it and the fixture models a real RF arrival.

The new case is the one the gate exists for: a UDP multicast reply naming RELAY_B in the route with a
matching `relay_node`. Revert the source change and it fails with "Expected 0 Was 11", meaning the
forged route installed a next hop.

Testing: builds clean on `heltec-v4` and `seeed-xiao-s3`, 102/102 across `test_traceroute_nexthop`,
`test_nexthop_routing` and `test_mqtt`. Not exercised against a real UDP peer, since my bench is two
LoRa nodes over USB, so the UDP field handling itself comes from reading `UdpMulticastHandler.h`
where the contrast between what it clears and what it leaves alone is explicit.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/modules/TraceRouteModule.cpp
+++ b/src/modules/TraceRouteModule.cpp
@@ -307,10 +307,8 @@ void TraceRouteModule::updateNextHops(const meshtastic_MeshPacket &p, meshtastic
 
         // The route array is unauthenticated payload, so only learn from it when the node it names as our
         // next hop is the one that actually relayed this packet to us. Otherwise a forged response could
-        // point any node's next_hop anywhere. relay_node only corroborates an RF route when the packet
-        // actually arrived over RF: MQTT ingress zeroes it (so the NO_RELAY_NODE test below catches that),
-        // but UDP multicast ingress preserves the on-wire value while clearing pki_encrypted, rx_snr and
-        // rx_rssi, so a LAN peer's reply would otherwise satisfy this check with no RF evidence at all.
+        // point any node's next_hop anywhere. relay_node only corroborates that over RF: UDP ingress
+        // keeps the on-wire value, so a LAN peer would otherwise pass this with no RF evidence.
         if (p.transport_mechanism != meshtastic_MeshPacket_TransportMechanism_TRANSPORT_LORA) {
             LOG_DEBUG("Ignore traceroute next-hop 0x%02x: non-RF transport %d", nextHopByte, p.transport_mechanism);
             return;
--- a/src/modules/TraceRouteModule.cpp
+++ b/src/modules/TraceRouteModule.cpp
@@ -307,8 +307,14 @@ void TraceRouteModule::updateNextHops(const meshtastic_MeshPacket &p, meshtastic
 
         // The route array is unauthenticated payload, so only learn from it when the node it names as our
         // next hop is the one that actually relayed this packet to us. Otherwise a forged response could
-        // point any node's next_hop anywhere. relay_node is 0 for MQTT-sourced packets, which cannot
-        // corroborate an RF route either.
+        // point any node's next_hop anywhere. relay_node only corroborates an RF route when the packet
+        // actually arrived over RF: MQTT ingress zeroes it (so the NO_RELAY_NODE test below catches that),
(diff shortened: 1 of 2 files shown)


h3lix1:
Closing this. The premise does not hold up.

The PR assumes an asymmetry between transports: that an RF sender can only nominate itself as the relay, while a UDP peer can nominate an arbitrary byte. That is not what the code does. `relay_node` is copied verbatim out of the plaintext on-air header at `RadioLibInterface.cpp:693`, with nothing binding it to the node that actually transmitted, so an RF attacker sets `hop_start=1` and any byte it likes. The LAN capability is not a superset of the RF one, it is the same capability with a narrower attacker population, and both need the channel key to reach the module at all.

The stated mechanism is also wrong in the other direction. `Router::send` sets `relay_node` to our own byte at `:480`, before the UDP fan-out at `:560`, so an honest LAN forwarder's relay byte is truthful and the existing consistency check means as much on UDP as it does on LoRa. "UDP preserves the on-wire value" only ever described a malicious sender, which is equally true of RF.

The strongest form of the attack needs no forgery of that field at all: a node that relays honestly as itself, and lists eight victim NodeNums after itself in the route array, becomes their next hop. Pointing a victim at an innocent third party is the weaker outcome, since it just degrades to flooding.

I also checked whether there was a plain correctness bug underneath, on the theory that next_hop is an RF concept being learned from a non-RF delivery. There isn't. A next_hop learned over UDP names the LAN peer that relayed the reply, and our subsequent unicasts are mirrored onto the same multicast group, so that peer relays them and directed delivery works. Gating on TRANSPORT_LORA would suppress legitimate route learning on LAN-bridged meshes, which makes this a regression rather than a fix.

Two things worth writing down for whoever looks here later, neither of which this PR addresses:

`TraceRouteModule::updateNextHops` runs on any reply with a non-zero `request_id`, without matching it against a traceroute we actually sent, and the module is promiscuous. It also skips the `resolveUniqueLastByte` ambiguity gate that the ACK-based learner in `NextHopRouter::sniffReceived` applies. Both are transport-independent, so neither is helped by the gate proposed here.

PR #11470, opened 2026-08-13
Fix out-of-bounds baud table read when GPS is re-enabled after failed probe

hackengineer:
When the probe scan exhausts `rareSerialSpeeds` and gives up, `speedSelect`
is left equal to `array_count(rareSerialSpeeds)`. If GPS is later re-enabled
at runtime (e.g. the user toggles `position.gps_mode` from the app),
`setup()` runs again with `probeTries` already at `GPS_PROBETRIES` and
indexes `rareSerialSpeeds[speedSelect]` past the end of the array, probing
garbage baud rates read from adjacent memory. Observed on hardware:

WARN | [GPS] No GNSS Module (baudrate 484401)
WARN | [GPS] No GNSS Module (baudrate 68673)
WARN | [GPS] No GNSS Module (baudrate 460269)

Fix: reset `speedSelect` before giving up, so a later re-enable restarts
the scan from the beginning of the table.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Reproduced and verified on a Seeed XIAO nRF54L15 Sense + Wio-SX1262 with an
L76K GNSS module: before the fix, disabling and re-enabling GPS from the app
after a failed scan produced the garbage-baud probes above; after the fix
the rescan walks the normal baud table. I don't have the listed devices, but
this code path only executes after a probe scan has fully failed *and* GPS
is re-enabled in the same boot — a state no device with a working, detected
GPS ever reaches, so regression risk on those targets is minimal. Community
testing welcome.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed GPS re-enablement after rare baud-rate detection attempts are exhausted.
  * Improved runtime stability by preventing invalid speed selection during GPS setup.

### DIFF
--- a/src/gps/GPS.cpp
+++ b/src/gps/GPS.cpp
@@ -828,6 +828,9 @@ bool GPS::setup()
                 if (gnssModel != GNSS_MODEL_UNKNOWN) {
                     detectedBaud = rareSerialSpeeds[speedSelect];
                 } else if (currentStep == 0 && ++speedSelect == array_count(rareSerialSpeeds)) {
+                    // Reset so a runtime GPS re-enable does not index
+                    // rareSerialSpeeds[speedSelect] out of bounds.
+                    speedSelect = 0;
                     LOG_WARN("Give up GPS probe, set to %d", GPS_BAUDRATE);
                     return true;
                 }


hackengineer:
@caveman99  friendly bump - this is the companion to #11468  which you merged
last week: same kind of small bounds fix, 5 lines, CI green, rebased onto
current develop so it fast-forwards. Also opened #11560 11560 to track the wider nRF54L15 series (#11471, #11472)
in one place, in case that helps triage.

Simplycissmus:
Confirmed by reading the state machine rather than the symptom: `speedSelect` and `probeTries` are members (`GPS.h:204-205`), so they survive a `setup()` re-entry, and the two exhaustion branches are asymmetric. `serialSpeeds` resets `speedSelect` and bumps `probeTries` (`GPS.cpp:823-825`); `rareSerialSpeeds` only logs and returns (`835`). A later re-enable therefore lands on `probe(rareSerialSpeeds[speedSelect])` with `speedSelect == array_count(...)`. The three lines remove exactly that entry.

Two things worth a thought while the file is open, neither of which blocks this:

**The guard itself stays fragile.** `++speedSelect == array_count(...)` fires only on equality. Once `speedSelect` sits *past* the count for any reason, the give-up branch cannot fire again until the `uint8_t` wraps the whole way around - so the state being fixed here is not a single bad read but a long walk before the code recovers by itself. A `>=` would make the branch self-correcting and would have contained this without needing the extra reset. Same shape one branch up, at `823`, for `serialSpeeds`.

**Should `probeTries` reset as well?** After this change a runtime re-enable restarts the scan, but only inside `rareSerialSpeeds`: `probeTries` is still at `GPS_PROBETRIES`, so the common table is never revisited. If someone toggled `gps_mode` because they changed the module, the common speeds are the ones you would want tried first. There is a precedent for the full reset a few lines above, where a failed `verifyCachedProbePresence()` zeroes `currentStep`, `speedSelect` and `probeTries` together (`811-815`).

For what it is worth on the platform coverage: the rare-speeds branch is inside `#ifndef CONFIG_IDF_TARGET_ESP32C6`, so C6 targets never reach this path at all.

PR #11476, opened 2026-08-13
fix(MQTT): stop reporting delivery from our own gateway echo

h3lix1:
Two places in the tree disagree about what the self-gateway MQTT echo means, and the disagreement is
visible to the user.

`MQTT.cpp` synthesised a local ROUTING ACK when our own packet came back off our own gateway topic.
Its comment gives the reasoning: getting the packet back means at least one node received it, "then
we'll stop our retransmissions".

`ReliableRouter::sniffReceived()` says the opposite. It exempts MQTT-sourced implicit ACKs from
`stopRetransmission()`, with the comment "Implicit ACKs from MQTT should not stop retransmissions".
So the retransmissions the MQTT comment describes stopping are never actually stopped.

Both cannot be right. The exemption is the one that matches the facts: the echo proves the broker
received the packet and sent it back, and says nothing about whether any node on the mesh heard it.
Our own gateway is specifically the node that cannot be evidence of that.

What the user sees from the disagreement is a phone told "delivered" while the radio is still
retrying, and then told `MAX_RETRANSMIT` a full retry ladder later if the message really was
undeliverable.

This drops the ACK, and the exemption goes with it.

### Why the exemption goes too

That ACK was its only firmware producer. The other writer of `TRANSPORT_MQTT` is the third-party
downlink path a few lines below, which returns early on `isFromUs()`; LoRa, UDP and PacketAPI each
stamp their own transport; and StoreForward preserves the field through its history but never sets
`decoded.request_id` on replay, so it cannot reach the `ackId || nakId` gate.

It was not literally unreachable, and the difference is worth stating plainly. `transport_mechanism`
is `MeshPacket` field 21 and rides in on `ToRadio` like any other field. `MeshService::handleToRadio`
scrubs `from`, `next_hop` and `relay_node`, each with a "we don't let clients assign X" comment, but
not this one, so a local client could hand us a self-addressed packet carrying `TRANSPORT_MQTT` and
reach this line. The guard was never a defence there: that client got the identical effect by leaving
the field at zero, since the guard only exempted MQTT-*marked* packets.

So the third commit closes that properly. `PacketAPI`, the sibling `ToRadio` ingress, already stamps
`TRANSPORT_API` before handing the packet over; `PhoneAPI` was the path with no stamp. It now stamps
`TRANSPORT_INTERNAL`, which is what a well-behaved client sends today, so compliant clients see no
change.

That is worth more than the guard was. The field is read as evidence of how a packet reached us:
`FloodingRouter`'s dupe-cancel, `ReliableRouter`'s rebroadcast gate, `NextHopRouter`'s dupe handling
and `NodeDB`'s `hops_away` and SNR bookkeeping all key off `TRANSPORT_LORA`, and `PositionModule`
gates own-position updates on the value. Forging `TRANSPORT_LORA` was the more interesting case, and
it is closed too. Stamped in `PhoneAPI` rather than at the `MeshService::handleToRadio` choke point so
`PacketAPI` keeps its own `TRANSPORT_API`, which `PositionModule` distinguishes.

Nothing in the retry ladder changes either way, because the exemption already prevented this ACK from
affecting it. What goes away is the false delivery report.

Real end-to-end ACKs are untouched. An ACK from the recipient relayed back through someone else's
gateway carries the recipient's node number in `from`, so `isFromUs(p)` was already false and the
exemption never applied to it. It still stops retransmissions and still reaches the phone, subject as
before to `ignore_mqtt` and to ordinary dupe suppression if the ACK was also heard over LoRa.

### The other direction

Honouring the MQTT comment instead, by letting the echo stop retransmissions, is the alternative and
I do not think it is the right one. It would end the retries on evidence that does not support them:
a node whose LoRa neighbours all missed the packet would stop trying as soon as the broker echoed it
back. But it is a coherent position if the intent was that MQTT connectivity substitutes for mesh
delivery, and that is a call about intended MQTT semantics rather than something to settle from
outside the project. Hence draft.

Worth knowing if you take this direction: a node whose traffic reaches the mesh only via MQTT loses
this signal, but not every signal. Where the recipient's side has a gateway on the same broker their
real ACK still comes back and still confirms. What goes away is the case where no ACK can return,
which is the same case where the echo was reporting a delivery that had not happened. There is also
no marker distinguishing an implicit ACK from a real one on the wire, so a client cannot presently
tell them apart either way, which is a separate gap.

### Tests

`test_receiveAcksOwnSentMessages` was named and documented for the behaviour this PR removes, and
every assertion in it was commented out behind a `FIXME`, so it passed in either direction. It is
renamed and given assertions that bite. `getForPhone()` is the one that does the work: the old ACK
went through neither `MockRoutingModule::sendAckNak` (the code called `allocAckNak` directly) nor
`MockRouter::enqueueReceivedMessage`, so the two commented-out assertions would have failed on
develop as well. It did reach the phone queue via `sendLocal()`.

`test_toradio_transport_mechanism_is_not_client_settable` forges `TRANSPORT_MQTT` through the real
`StreamAPI` `ToRadio` path and asserts what reaches `Router::send`.

Both were checked against a tree without their fix: the first fails on develop at the `getForPhone`
assertion, the second fails with `Expected 0 Was 5`.

Related: I have a separate PR for `ignore_mqtt` not being honoured on this same branch of
`onReceiveProto()`. It is independent of which direction you pick here, though if this lands it
covers that case too and I would close the other one.

Merges cleanly with #11477, which adds a branch inside the same if/else a few lines below; checked
with `git merge-tree`, no conflict. Worth reviewing alongside #11478 rather than separately: each
argues its own transport check is individually dead, and after both there is no `transport_mechanism`
check left anywhere on the stop-retransmissions path.

Testing: builds clean on `heltec-v4` and `seeed-xiao-s3`, 171/171 across `test_mqtt`,
`test_nexthop_routing`, `test_packet_signing` and `test_event_channel_phone_api` on `native-macos`.
Not exercised against a live broker, since I have no MQTT setup here; the contradiction is from
reading the two call sites and the exemption condition.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

### DIFF
--- a/src/mesh/PhoneAPI.cpp
+++ b/src/mesh/PhoneAPI.cpp
@@ -1766,6 +1766,10 @@ PhoneAPI::LocalAdminGate PhoneAPI::classifyLocalAdminPacket(const meshtastic_Mes
  */
 bool PhoneAPI::handleToRadioPacket(meshtastic_MeshPacket &p)
 {
+    // Clients don't get to name the transport they arrived on, the same way they don't get to assign
+    // `from`. PacketAPI already stamps its own ingress; this is the path that did not.
+    p.transport_mechanism = meshtastic_MeshPacket_TransportMechanism_TRANSPORT_INTERNAL;
+
     printPacket("PACKET FROM PHONE", &p);
 
 #if defined(MESHTASTIC_ENCRYPTED_STORAGE) && defined(MESHTASTIC_PHONEAPI_ACCESS_CONTROL)
--- a/src/mesh/ReliableRouter.cpp
+++ b/src/mesh/ReliableRouter.cpp
@@ -160,9 +160,7 @@ void ReliableRouter::sniffReceived(const meshtastic_MeshPacket *p, const meshtas
         PacketId nakId = (c && c->error_reason != meshtastic_Routing_Error_NONE) ? p->decoded.request_id : 0;
 
         // We intentionally don't check wasSeenRecently, because it is harmless to delete non existent retransmission records
-        if ((ackId || nakId) &&
-            // Implicit ACKs from MQTT should not stop retransmissions
-            !(isFromUs(p) && p->transport_mechanism == meshtastic_MeshPacket_TransportMechanism_TRANSPORT_MQTT)) {
+        if (ackId || nakId) {
             LOG_DEBUG("Received a %s for 0x%08x, stopping retransmissions", ackId ? "ACK" : "NAK", ackId);
             if (ackId) {
                 stopRetransmission(p->to, ackId);
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -116,19 +116,8 @@ inline void onReceiveProto(char *topic, byte *payload, size_t length)
     // Generate node ID from nodenum for comparison
     std::string nodeId = nodeDB->getNodeId();
     if (strcmp(e.gateway_id, nodeId.c_str()) == 0) {
-        // Our own packet, echoed back off our own gateway topic. That round trip proves the broker
-        // received and returned it. It does not show that any node on the mesh heard it, which is
-        // what an ACK delivered to the phone says.
-        //
-        // This used to synthesise a local ROUTING ACK here, on the reasoning that getting our own
-        // packet back means at least one node received it, "then we'll stop our retransmissions".
-        // Those retransmissions are not in fact stopped: ReliableRouter::sniffReceived() exempts
-        // MQTT-sourced implicit ACKs from stopRetransmission() precisely because the echo is not
-        // delivery evidence. So the ACK reached the phone as a delivery confirmation while the radio
-        // carried on retrying, and if the message really was undeliverable the phone was told
-        // "delivered" and then, one full retry ladder later, MAX_RETRANSMIT.
-        //
-        // Drop the ACK rather than the exemption. See the PR discussion for the other direction.
+        // Our own packet back off our own gateway proves the broker returned it, not that the mesh
+        // heard it, so no ACK is minted here. See the commit message.
         if (!isFromUs(e.packet)) {
             LOG_INFO("Ignore downlink msg we sent");
         }
--- a/src/mqtt/MQTT.cpp
+++ b/src/mqtt/MQTT.cpp
@@ -116,17 +116,20 @@ inline void onReceiveProto(char *topic, byte *payload, size_t length)
     // Generate node ID from nodenum for comparison
     std::string nodeId = nodeDB->getNodeId();
     if (strcmp(e.gateway_id, nodeId.c_str()) == 0) {
-        // Generate an implicit ACK towards ourselves (handled and processed only locally!) for this message.
-        // We do this because packets are not rebroadcasted back into MQTT anymore and we assume that at least one node
-        // receives it when we get our own packet back. Then we'll stop our retransmissions.
-        if (isFromUs(e.packet)) {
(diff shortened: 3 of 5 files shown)


caveman99 (changes requested):
Reviewed together with #11469, which you opened about four hours earlier taking the opposite position on these same lines.

**The diagnosis is right, and I can confirm it from the tree.** The ACK's justifying comment claims *"Then we'll stop our retransmissions"* — that is false. `ReliableRouter::sniffReceived()` (`src/mesh/ReliableRouter.cpp:165`) explicitly exempts exactly this packet:

```c
// Implicit ACKs from MQTT should not stop retransmissions
!(isFromUs(p) && p->transport_mechanism == meshtastic_MeshPacket_TransportMechanism_TRANSPORT_MQTT)
```

So the firmware mints an ACK, stamps it `TRANSPORT_MQTT`, and the router then refuses to act on it. Its only surviving effect is to tell the phone "delivered" while the radio keeps retrying — and, if the message really was undeliverable, to follow that with `MAX_RETRANSMIT`. You've described that accurately.

**But this is a product decision, not a bug fix, and I won't merge it as one.**

Removing the ACK removes the *only* delivery signal a gateway-only node ever gets. A node with no RF neighbours — MQTT-bridged deployments, which are common — currently sees its message confirmed. After this it sees nothing until `MAX_RETRANSMIT`. Whether that is an improvement (honest silence) or a regression (losing the one signal available) is a call for the dev channel, not something to settle in a draft PR. Please raise it there first.

**The change is also incomplete as written.** If the self-gateway echo is the only producer of a self-origin `TRANSPORT_MQTT` routing packet — and it is, since the third-party-gateway branch returns on `isFromUs(e.packet)` a few lines below — then removing it makes the `ReliableRouter.cpp:165` exemption unreachable. You write "Drop the ACK rather than the exemption", then leave the exemption in place undiscussed. Either it still guards something, in which case your sole-producer reasoning is wrong, or it is now dead code and belongs in this diff.

**Coding guidelines** (`AGENTS.md:83`, `.github/copilot-instructions.md:340` — *"Keep code comments minimal - one or two lines, max"*): the replacement is a 14-line comment block standing in for 12 deleted lines. The reasoning is good and it belongs in the commit message and PR body — not the source. Two lines in the code.

---

h3lix1:
> Removing the ACK removes the only delivery signal a gateway-only node ever gets.

Not completely correct. A real ACK from the recipient comes back through MQTT.cpp:157 with `from` set to the recipient, so `isFromUs(p)` is false, the `:165` exemption doesn't apply, and it stops retransmissions and reaches the phone the normal way.

I've removed the exemption. `transport_mechanism` is field 21, and `MeshService::handleToRadio` scrubs `from`, `next_hop` and `relay_node` but leaves that field exactly as the client sent it, so a local client can reach that line with a self-addressed packet. The guard only exempted MQTT-marked packets, so we remove it here and add the missing scrub in the same PR. PacketAPI already stamps its ingress; PhoneAPI was the one that didn't.

I fixed `test_receiveAcksOwnSentMessages`, which was named for the behaviour this deletes with every assertion commented out behind a FIXME. It now fails on develop and passes here. The scrub has its own test that forges TRANSPORT_MQTT through the real ToRadio path.

h3lix1:
While I think this is good to not use MQTT as confirmation a message, but likely not a universal thought. closing for now.

PR #11480, opened 2026-08-13
fix(mesh): keep ROUTING_APP responses when toPhoneQueue is full

### LINKED ISSUE #11439: [Bug]: ROUTING_APP responses can be dropped when toPhoneQueue is full
### Category

BLE / Other

### Hardware

Heltec V3 (reported); the queue policy is cross-platform.

### Is this bug report about any UI component firmware?

No.

### Firmware Version

Current `develop` at `2f6906974e723843bbd8cd4be9d970e9b558f916`. The original field report used firmware 2.8.0 on an ESP32-S3.

### Description

A field report in meshtastic/Meshtastic-Android#6601 and follow-up analysis on meshtastic/Meshtastic-Android#6624 describes phone-originated messages remaining indefinitely without a delivery confirmation when:

- the phone is connected over BLE,
- MQTT client proxying is enabled,
- a busy public channel supplies sustained downlink traffic.

Disabling only the phone-side MQTT relay avoids the symptom, and the same send over Wi-Fi resolves.

The current source supports a concrete loss path. MQTT client-proxy downlink re-enters the normal receive pipeline. Received packets and correlated `ROUTING_APP` ACK/NAK responses then share `MeshService::toPhoneQueue`. When that queue is full, `MeshService::sendToPhone()` preserves a newly arriving text/range-test packet by evicting the oldest entry, but drops any other newly arriving packet. A decoded `ROUTING_APP` packet with a nonzero `request_id` is therefore dropped with no replacement NAK.

I have not independently reproduced the reporter's BLE/MQTT timing on physical hardware, so this report does not claim that the queue-full branch has yet been captured during the field reproduction. The source-level failure mode is deterministic, however.

A local unit regression against current `develop` fills `toPhoneQueue` and verifies three invariants:

1. a correlated decoded `ROUTING_APP` response is admitted when lower-priority packets occupy the queue;
2. a later text packet cannot evict an already queued correlated routing response;
3. when the queue contains only correlated routing responses, a new text packet is dropped rather than destroying one of them.

All three tests fail against the current policy. A narrow overflow-only change that evicts the oldest non-routing-response packet passes all three while preserving the relative order of retained packets. The protected classification is intentionally limited to decoded `ROUTING_APP` packets with a nonzero `request_id`; it does not guess the type of encrypted packets.

The overflow scan is bounded by `MAX_RX_TOPHONE` and uses the same queue-rotation assumption already used by `getNodenumFromRequestId()` and `reconcilePendingRxTimes()`.

### Relevant log output

Sanitized unit result on the unmodified policy:

```text
3 queue-policy regressions failed; 15 existing tests succeeded
```

With the narrow priority policy:

```text
test_stream_api passed
formatter and static checks passed
```

No raw device logs or user data are attached. A useful hardware confirmation would capture the existing `ToPhone queue full, drop packet` warning during the reported BLE/MQTT reproduction.

caveman99:
Fixes #11439

#2918 narrowed the `toPhoneQueue` overflow policy to evict the oldest entry only for `TEXT_MESSAGE_APP` and `RANGE_TEST_APP`, dropping every other portnum. A dropped `ROUTING_APP` response leaves the phone with no delivery confirmation for a message it sent, so it stays pending indefinitely. Sustained downlink (busy channel, MQTT client proxy re-entering the receive pipeline) is enough to keep the queue full.

Before #2918 a full queue always discarded the oldest entry, so ACK/NAK delivery was unaffected.

Adds `ROUTING_APP` to the eviction list. New `test_tophone_queue` suite covers the admitted case and the two neighbouring behaviours; the first case fails without the fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved phone-delivery queue handling when the queue is full.
  - Routing responses, text messages, and range-test packets now replace the oldest queued packet when necessary.
  - Lower-priority telemetry and unsupported packets continue to be dropped when no space is available.
  - Encrypted packets are now classified correctly without relying on aliased packet metadata.

- **Tests**
  - Added coverage for queue overflow, packet prioritization, and encrypted packet handling.

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -492,8 +492,10 @@ void MeshService::sendToPhone(meshtastic_MeshPacket *p)
 #endif
 
     if (toPhoneQueue.numFree() == 0) {
+        // ROUTING_APP is the phone's only delivery confirmation for a message it sent; dropping it
+        // leaves that message pending forever, so it displaces the oldest like text does.
         if (p->decoded.portnum == meshtastic_PortNum_TEXT_MESSAGE_APP ||
-            p->decoded.portnum == meshtastic_PortNum_RANGE_TEST_APP) {
+            p->decoded.portnum == meshtastic_PortNum_RANGE_TEST_APP || p->decoded.portnum == meshtastic_PortNum_ROUTING_APP) {
             LOG_WARN("ToPhone queue full, discard oldest");
             meshtastic_MeshPacket *d = toPhoneQueue.dequeuePtr(0);
             if (d)
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -492,10 +492,11 @@ void MeshService::sendToPhone(meshtastic_MeshPacket *p)
 #endif
 
     if (toPhoneQueue.numFree() == 0) {
-        // ROUTING_APP is the phone's only delivery confirmation for a message it sent; dropping it
-        // leaves that message pending forever, so it displaces the oldest like text does.
-        if (p->decoded.portnum == meshtastic_PortNum_TEXT_MESSAGE_APP ||
-            p->decoded.portnum == meshtastic_PortNum_RANGE_TEST_APP || p->decoded.portnum == meshtastic_PortNum_ROUTING_APP) {
+        // ROUTING_APP is the phone's only delivery confirmation, so it displaces the oldest like
+        // text does. Gate the variant: decoded.portnum aliases encrypted.size in the union.
+        if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag &&
+            (p->decoded.portnum == meshtastic_PortNum_TEXT_MESSAGE_APP ||
+             p->decoded.portnum == meshtastic_PortNum_RANGE_TEST_APP || p->decoded.portnum == meshtastic_PortNum_ROUTING_APP)) {
             LOG_WARN("ToPhone queue full, discard oldest");
             meshtastic_MeshPacket *d = toPhoneQueue.dequeuePtr(0);
             if (d)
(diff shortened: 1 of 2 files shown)


caveman99:
Addressed in 3acf29e57.

Applied:

- **Decoded-variant gate.** Valid, and it predates this PR: `decoded.portnum` aliases `encrypted.size` in the nanopb oneof union, so an encrypted packet was already classifiable by its ciphertext length (1 byte read as `TEXT_MESSAGE_APP`, 66 as `RANGE_TEST_APP`). One guard covers all three portnums. Covered by `test_encrypted_packet_is_not_classified_by_portnum`.
- **`rebroadcast_mode` restored in `tearDown`.**

Not applied:

- **`request_id != 0`.** `MeshModule::allocAckNak()` sets `decoded.request_id = idFrom` unconditionally, so every routing ACK/NAK already carries one. The condition excludes nothing that occurs.
- **Never evicting a queued routing response.** That is a new policy rather than the regression. #2918 replaced an unconditional evict-oldest with a portnum allow-list; this PR puts `ROUTING_APP` back on that list and nothing more. Protecting queued responses from eviction changes behaviour for text and range-test too and belongs in its own change.

caveman99:
`trunk check` failure was a trufflehog false positive, fixed in f36324350: `test_text_still_admitted_when_queue_full` is `test_` followed by exactly 35 characters, which matches the Lob API key shape. Renamed to `test_text_evicts_oldest_when_full`. Verified both ways locally: the old name reproduces `84:0 high Secret detected trufflehog/Lob`, the new one is clean.

On the three review items, all three were already addressed above or are declined:

- **`request_id != 0`** and **never evicting a queued routing response**: unchanged from my previous reply. `MeshModule::allocAckNak()` sets `decoded.request_id = idFrom` unconditionally, so the zero-ID routing packet the comment asks me to regression-test does not occur. The eviction-protection policy is a behaviour change beyond the regression and belongs in its own PR.
- **camelCase test names**: declined. Every Unity test function in this repo is snake_case (960 of 960 `RUN_TEST` targets), matching the skeleton in `test/README.md`. The camelCase guideline covers firmware functions, not Unity entry points; renaming here would make this the only suite that differs.

thebentern (approved):

PR #11481, opened 2026-08-13
Fix NodeDB loadFromDisk to replace store instead of appending

NomDeTom:
This pull request fixes a long-standing bug in the NodeDB persistence logic, where loading the node database from disk would append to the in-memory store rather than replacing it, causing node entries to accumulate across loads. It also updates the test infrastructure and documentation to ensure test isolation and clarify NodeDB state management. Several new and improved tests are added to verify correct behavior, and the test setup is hardened to guarantee a clean state for each test.

**Persistence and NodeDB State Management**

* Fixed a bug where `loadFromDisk()` did not clear the in-memory node database before loading from disk, resulting in duplicate entries after multiple load cycles. Now, all relevant `std::vector` members and counters in `nodeDatabase` are cleared before loading.
* Added the helper function `destinationHasVectorMembers()` to prevent `memset` on protobuf structs with `std::vector` members, avoiding undefined behavior. [[1]](diffhunk://#diff-d3ca504aa401bdfaa42fcd99349d867b504756a86721afbbb46482909e8b1001R2055-R2061) [[2]](diffhunk://#diff-d3ca504aa401bdfaa42fcd99349d867b504756a86721afbbb46482909e8b1001L2076-R2083) [[3]](diffhunk://#diff-d3ca504aa401bdfaa42fcd99349d867b504756a86721afbbb46482909e8b1001L2103-R2110)

**Test Infrastructure and Isolation**

* Updated test fixtures (`setUp` in various test suites) to always remove both `nodes.proto` and `warm.dat` before constructing a new `NodeDB`, ensuring tests do not inherit persisted state from previous runs. [[1]](diffhunk://#diff-dce01576acb7598ec197d0732dba7894a08f8656a5bdd86b78986787f35dbb60R1014-R1018) [[2]](diffhunk://#diff-57fa3d226cfe5721a368f0c9fd8119e7cdd5524465cca95d75efcbf3886ae1bcR332-R335)
* Improved documentation in `test/README.md` to explain NodeDB ownership, persistence, and best practices for test isolation, including a new section on handling save/load with mocks. [[1]](diffhunk://#diff-5de36acd90308dc62abf7855a686ee7052ffb6e762c756fd735fb0c9fbd9595dR269-R301) [[2]](diffhunk://#diff-5de36acd90308dc62abf7855a686ee7052ffb6e762c756fd735fb0c9fbd9595dL373-R425)
* Added a warning about test names that can trigger false positives in secret scanning tools.

**Testing and Verification**

* Added new tests to verify that save/load cycles do not accumulate nodes, that absent or undecodable files do not leave stale nodes in memory, and that node admission and eviction logic works correctly at capacity. [[1]](diffhunk://#diff-38094f6f0c9434ab24df8074860151c7182f0261a81fda1eabe9bb534796311fR298-R411) [[2]](diffhunk://#diff-38094f6f0c9434ab24df8074860151c7182f0261a81fda1eabe9bb534796311fL297-R431)
* Refactored test utility code, e.g., adding `armSaveGate()` to ensure the persistence gate is properly armed for each save, and exposing `runLoad()`/`runSave()` in the test shim. [[1]](diffhunk://#diff-38094f6f0c9434ab24df8074860151c7182f0261a81fda1eabe9bb534796311fR29-R32) [[2]](diffhunk://#diff-38094f6f0c9434ab24df8074860151c7182f0261a81fda1eabe9bb534796311fR81-R90)

**Test Naming and Maintenance**

* Renamed tests for clarity and to avoid secret scanning false positives. [[1]](diffhunk://#diff-38094f6f0c9434ab24df8074860151c7182f0261a81fda1eabe9bb534796311fL252-R267) [[2]](diffhunk://#diff-38094f6f0c9434ab24df8074860151c7182f0261a81fda1eabe9bb534796311fL269-R284)

**Test Suite Consistency**

* Included `FSCommon.h` where needed for file operations in tests. [[1]](diffhunk://#diff-57fa3d226cfe5721a368f0c9fd8119e7cdd5524465cca95d75efcbf3886ae1bcR6) [[2]](diffhunk://#diff-38094f6f0c9434ab24df8074860151c7182f0261a81fda1eabe9bb534796311fR16)

These changes together ensure that the NodeDB hot store is always correctly replaced on load, tests are reliably isolated, and persistence logic is robustly verified.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed repeated database loads so stale node data and version information are cleared before reloading.
  * Improved handling when optional satellite databases are unavailable.
  * Ensured fresh database states when replacing or recreating stored node data.

* **Tests**
  * Added coverage for persistence round trips, missing files, stale data, and node admission limits.
  * Added guidance for reliable persistence testing and fixture cleanup.

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -161,15 +161,6 @@ lint:
         - test/test_airtime/test_main.cpp
         - test/test_throttle/test_main.cpp
         - test/test_uptime_clock/test_main.cpp
-    # Same detector, different trigger: Lob's key shape is a short word, an
-    # underscore and exactly 35 further name characters, so any test function
-    # whose name runs 35 characters past "test_" matches. Two in this file do -
-    # naming them here would trip the detector on this file too. The reported
-    # "secret" is the function name itself, and the alternative is padding
-    # identifiers to dodge a character count.
-    - linters: [trufflehog]
-      paths:
-        - test/test_nodedb_blocked/test_main.cpp
 runtimes:
   enabled:
     - python@3.14.4
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -161,6 +161,15 @@ lint:
         - test/test_airtime/test_main.cpp
         - test/test_throttle/test_main.cpp
         - test/test_uptime_clock/test_main.cpp
+    # Same detector, different trigger: Lob's key shape is a short word, an
+    # underscore and exactly 35 further name characters, so any test function
+    # whose name runs 35 characters past "test_" matches. Two in this file do -
+    # naming them here would trip the detector on this file too. The reported
+    # "secret" is the function name itself, and the alternative is padding
+    # identifiers to dodge a character count.
+    - linters: [trufflehog]
+      paths:
+        - test/test_nodedb_blocked/test_main.cpp
 runtimes:
   enabled:
     - python@3.14.4
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -2369,6 +2369,12 @@ void NodeDB::loadFromDisk()
         ~Disarm() { self.disarmNodeDatabaseDecodeTargets(); }
     } disarm{*this};
 
+    // loadProto() skips its destination memset for this struct (std::vector members), so reset it
+    // here: the nodes callback only push_back()s and nodeDatabase is a global, so a second load
+    // appended to the first, and a stale version skipped installDefaultNodeDatabase().
+    nodeDatabase.nodes.clear();
+    numMeshNodes = 0;
+    nodeDatabase.version = 0;
     // Avoid push_back's power-of-2 capacity growth wasting RAM at small N.
     nodeDatabase.nodes.reserve(MAX_NUM_NODES);
 
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -2363,6 +2363,17 @@ void NodeDB::loadFromDisk()
     {
         concurrency::LockGuard guard(&satelliteMutex);
         armNodeDatabaseDecodeTargets();
+        // Where a satellite DB is compiled out there is no map to arm, so the decode callback falls
+        // through to the vector and push_back()s into it. Reset those the same way the save path
+        // does on its way out, so no build can carry entries from one load into the next.
+        nodeDatabase.positions.clear();
+        nodeDatabase.positions.shrink_to_fit();
+        nodeDatabase.telemetry.clear();
+        nodeDatabase.telemetry.shrink_to_fit();
+        nodeDatabase.environment.clear();
+        nodeDatabase.environment.shrink_to_fit();
+        nodeDatabase.status.clear();
+        nodeDatabase.status.shrink_to_fit();
     }
     struct Disarm {
         NodeDB &self;
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -2052,6 +2052,13 @@ void NodeDB::pickNewNodeNum()
     myNodeInfo.my_node_num = nodeNum;
 }
 
+/// Both NodeDatabase descriptors decode into structs holding std::vector members, which a memset
+/// would leave in an undefined state; their callers reset the destination themselves instead.
+static inline bool destinationHasVectorMembers(const pb_msgdesc_t *fields)
+{
+    return fields == &meshtastic_NodeDatabase_msg || fields == &meshtastic_NodeDatabase_Legacy_msg;
+}
+
 /** Load a protobuf from a file, return LoadFileResult */
 LoadFileResult NodeDB::loadProto(const char *filename, size_t protoSize, size_t objSize, const pb_msgdesc_t *fields,
                                  void *dest_struct)
@@ -2073,7 +2080,7 @@ LoadFileResult NodeDB::loadProto(const char *filename, size_t protoSize, size_t 
         if (EncryptedStorage::readAndDecrypt(filename, decBuf.get(), protoSize, decLen)) {
             LOG_INFO("Load encrypted %s", filename);
             pb_istream_t stream = pb_istream_from_buffer(decBuf.get(), decLen);
-            if (fields != &meshtastic_NodeDatabase_msg)
+            if (!destinationHasVectorMembers(fields))
                 memset(dest_struct, 0, objSize);
             if (!pb_decode(&stream, fields, dest_struct)) {
                 LOG_ERROR("Can't decode protobuf %s", PB_GET_ERROR(&stream));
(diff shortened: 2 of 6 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -2369,6 +2369,12 @@ void NodeDB::loadFromDisk()
         ~Disarm() { self.disarmNodeDatabaseDecodeTargets(); }
     } disarm{*this};
 
+    // loadProto() skips its destination memset for this struct (std::vector members), so reset it
+    // here: the nodes callback only push_back()s and nodeDatabase is a global, so a second load
+    // appended to the first, and a stale version skipped installDefaultNodeDatabase().
+    nodeDatabase.nodes.clear();
+    numMeshNodes = 0;
+    nodeDatabase.version = 0;
```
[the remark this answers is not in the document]
NomDeTom: Mechanism confirmed, though not reachable in any current build combination (satellite DBs are compiled out only on ARCH_STM32WL, while the second load needs `reloadFromDisk()` → `MESHTASTIC_ENCRYPTED_STORAGE`, which `configuration.h` defines only under ARCH_NRF52) — fixed regardless: all four vectors now `clear()` + `shrink_to_fit()` under `satelliteMutex` alongside the arm.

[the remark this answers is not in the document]
NomDeTom: Agreed on the asymmetry and fixed, but the severity is lower than "Major": the excluded-DB path is STM32WL-only and has no `reloadFromDisk()` (that needs `MESHTASTIC_ENCRYPTED_STORAGE`, nRF52-only), and `saveNodeDatabaseToDisk()` already clears these vectors unconditionally before encoding, so no stale entry was ever persisted or read.

.trunk/trunk.yaml
```diff
@@ -160,6 +160,15 @@ lint:
...
         - test/test_uptime_clock/test_main.cpp
+    # Same detector, different trigger: Lob's key shape is a short word, an
+    # underscore and exactly 35 further name characters, so any test function
+    # whose name runs 35 characters past "test_" matches. Two in this file do -
+    # naming them here would trip the detector on this file too. The reported
+    # "secret" is the function name itself, and the alternative is padding
+    # identifiers to dodge a character count.
+    - linters: [trufflehog]
+      paths:
+        - test/test_nodedb_blocked/test_main.cpp
```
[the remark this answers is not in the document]
NomDeTom: Taken — exemption dropped and both triggering functions renamed instead; worth noting one of the two (`test_removeNodeByNum_presentNodeOnFullDb`) was pre-existing on develop and only surfaced because trunk lints touched files, so renaming just the new one would not have been enough.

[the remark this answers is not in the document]
NomDeTom: Agreed and done via the rename half of your suggestion — the exact-value allowlist half is not expressible here (trunk `lint.ignore` takes only `linters` + `paths`), and `test/README.md` §5 under "Creating a New Test Suite" now documents the exact-35-character rule so the next test name does not re-trip it.

NomDeTom:
On the comment-length nitpick: keeping these as-is. The block at `test_main.cpp:298-318` records why the defect survived review (two adjacent paths — `armNodeDatabaseDecodeTargets()` and `migrateLegacyNodeDatabase()` — already reset their destinations, so the gap looked covered) and why `reloadFromDisk()` makes it a runtime bug rather than a test artifact; the shorter ones state the non-obvious preconditions each test depends on, such as the save gate needing re-arming after every load because `owner` is a reference into `devicestate`. That is the *why* the guideline asks for, and a reader who has to reconstruct it from the diff will not.

.trunk/trunk.yaml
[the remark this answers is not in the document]
NomDeTom: Check again

src/mesh/NodeDB.cpp
```diff
@@ -2363,12 +2363,29 @@ void NodeDB::loadFromDisk()
...
     struct Disarm {
         NodeDB &self;
         ~Disarm() { self.disarmNodeDatabaseDecodeTargets(); }
     } disarm{*this};
 
+    // loadProto() skips its destination memset for this struct (std::vector members), so reset it
+    // here: the nodes callback only push_back()s and nodeDatabase is a global, so a second load
+    // appended to the first, and a stale version skipped installDefaultNodeDatabase().
+    nodeDatabase.nodes.clear();
+    numMeshNodes = 0;
```
caveman99: I would put the numMeshNodes = 0; before the clear call, or racing callers may think there's still content while the clear is running and tripping over it.

test/test_admin_radio/test_main.cpp
caveman99, line 1: I think this (and also test_mesh_module) also need to clear th warm.dat file. For additional freshness.

test/test_nodedb_blocked/test_main.cpp
```diff
@@ -280,6 +295,118 @@ static void test_removeNodeByNum_presentNodeOnFullDb(void)
...
+
+    FSCom.remove(nodeDatabaseFileName);
+    db->runLoad();
+
+    TEST_ASSERT_EQUAL_INT(0, (int)db->getNumMeshNodes());
+    TEST_ASSERT_NULL(db->getMeshNode(0x0B000001));
+    TEST_ASSERT_NULL(db->getMeshNode(0x0BADF00D)); // not even self survives an empty store
+
+    // Leave the file the suite declares in state-manifest.tsv on disk for the next test.
+    armSaveGate();
```
caveman99, line 362: db->clearHot() after armSaveGate means you have a genuinely empty database and not an array of MAX_NUM_NODES empty nodes.

src/mesh/NodeDB.cpp
caveman99: what about &meshtastic_NodeDatabase_Legacy_msg here? (see line 2104/2104) - i know this is not in the diff, but it's what this PR is about.

caveman99 (changes requested):

caveman99:
Overall this is sound, my comments are compaints on a high level.

src/mesh/NodeDB.cpp
NomDeTom: I think this is addressed now.

test/test_nodedb_blocked/test_main.cpp
NomDeTom: fixed.

test/test_admin_radio/test_main.cpp
NomDeTom: fixed

src/mesh/NodeDB.cpp
NomDeTom: fixed.

PR #11482, opened 2026-08-13
fix(test): make the native-windows test suite build and run

caveman99:
`pio test -e native-windows` failed all 53 suites at the build stage. Five independent causes, all Windows-only, plus one Unity configuration gap that turned out not to be Windows-specific at all.

## Causes

**`lstat()` in `test/TestUtil.cpp`.** MinGW-w64 does not provide it. The state-checkpoint walk added in #11322 is fenced with `ARCH_PORTDUINO`, which `native-windows` also satisfies, so the file failed to compile and took every suite with it. This is the regression that broke the Windows runner; the others were masked behind it. Routed through a `stat()` shim on `_WIN32`.

**Missing `setUp`/`tearDown` in four suites.** `test_default`, `test_http_content_handler`, `test_meshpacket_serializer` and `test_serial` define neither and relied on the weak defaults PlatformIO emits in its generated `unity_config.c`. Those defaults are enabled on MinGW, but GCC lowers a weak definition on PE-COFF to a weak external:

```
0000000000000000 T .weak.setUp.unityOutputStart
                 w setUp
```

`setUp` stays undefined (`w`) with a renamed alternate, rather than ELF's weak definition (`W`), so it does not satisfy `unity.c`'s strong reference and the link fails. Defined explicitly, as the other 49 suites already do.

**`<arpa/inet.h>` in `test_mqtt`.** Absent on MinGW; included only for `htonl()`. Uses `winsock2.h` there.

**`UNITY_INCLUDE_DOUBLE` was never defined anywhere.** Unity omits double support unless asked, so `TEST_ASSERT_DOUBLE_*` compiles into an unconditional "Unity Double Precision Disabled" failure. This is not Windows-specific. Verified on Debian with gcc, against the Linux env's own Unity 2.6.1 and PlatformIO's generated `native` `unity_config`:

```
UNITY_INCLUDE_DOUBLE : NOT defined
UNITY_EXCLUDE_DOUBLE : defined
test_double_within:FAIL: Unity Double Precision Disabled
```

The macro appears nowhere in the repo, the ini files, `test_native.yml`, or PlatformIO's Unity runner, which adds only `UNITY_INCLUDE_CONFIG_H`. It is therefore defined in `portduino_base.build_flags_common`, which every native env resolves: `native`, `native-tft`, `native-fb`, `native-tft-debug`, `coverage`, `coverage-event-policy`, `native-macos`, `native-windows`, `native-wasm`.

**`test_getfiles_rejects_overlong_path`.** Excluded on `_WIN32`. Overrunning the 228-byte `file_name` needs at least 229 bytes below the portduino root, and that root is already about 34 bytes, so every qualifying path passes the 260-byte `MAX_PATH`. The nested `mkdir()` fails, the file is never created, and `getFiles()` has nothing to drop. No component layout satisfies both limits.

## Verification

Full `pio test -e native-windows` on Windows 11 / UCRT64 gcc 16.1.0, before and after:

| | Before | After |
| --- | --- | --- |
| Suites built and run | 46 of 53 | 53 of 53 |
| Test cases | 839 | 938 |
| Failed | 2, plus 7 suites that never built | 1 |

## Scope

Every guard is `_WIN32`-only except `UNITY_INCLUDE_DOUBLE`, which now applies to all native envs and **does change Linux and macOS**: `TEST_ASSERT_DOUBLE_*` becomes a real comparison instead of a stub that always fails. `test_gps_update_scheduling` is the only suite in the repo using those macros, and its arithmetic is integer-based and bit-identical across platforms, so it should pass wherever it runs.

Worth flagging for review: that suite currently reports PASSED on CI in 0.03 seconds while emitting no Unity output at all, so its assertions appear never to execute there. That is a separate issue and is not addressed here.

No CI job currently compiles the test tree for `native-windows`: `test_native.yml` runs on Ubuntu against `-e coverage`, and `build_windows_bin.yml` runs `pio run`, never `pio test`. That is why the `lstat` regression stood for a week.

## Known limitation

`test_fscommon_getfiles` still fails in a full run, for a cause outside this change: `VFSImpl::rmdir()` in framework-portduino calls `unlink()`, which cannot remove a directory, so the tree its depth-limit case creates survives `setUp()`'s `rmDir()` and makes `getFiles()` report a spurious depth truncation. Fixed by meshtastic/framework-portduino#77; once that lands and the platform pin picks it up, the suite goes green on Windows with no further change here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-platform directory traversal and network byte-order handling for Windows builds.
  * Excluded unsupported long-path tests on Windows.

* **Tests**
  * Improved test compatibility with MinGW by adding required test lifecycle hooks.
  * Enabled double-precision assertions for GPS scheduling tests.

### DIFF
--- a/test/TestUtil.cpp
+++ b/test/TestUtil.cpp
@@ -63,6 +63,20 @@ void testStateCheckpoint(const char *, const char *) {}
 namespace
 {
 
+/// MinGW-w64 has no lstat(): Windows has no POSIX symlink stat, and nothing in a test sandbox
+/// creates a symlink, so stat() sees the same thing for every entry walk() can reach.
+#ifdef _WIN32
+inline int lstatCompat(const char *path, struct stat *st)
+{
+    return stat(path, st);
+}
+#else
+inline int lstatCompat(const char *path, struct stat *st)
+{
+    return lstat(path, st);
+}
+#endif
+
 /// Content fingerprint, used only to answer "did this file change?". FNV-1a rather than a real
 /// digest because the answer is a boolean and the files are a few KB of protobuf; nothing here
 /// records a hash as an expected value, which is what would make this a snapshot test.
@@ -96,7 +110,7 @@ void walk(const std::string &root, const std::string &rel, std::map<std::string,
         const std::string childRel = rel.empty() ? std::string(e->d_name) : rel + "/" + e->d_name;
         const std::string childPath = root + "/" + childRel;
         struct stat st;
-        if (lstat(childPath.c_str(), &st) != 0)
+        if (lstatCompat(childPath.c_str(), &st) != 0)
             continue;
         if (S_ISDIR(st.st_mode))
             walk(root, childRel, out);
--- a/test/test_default/test_main.cpp
+++ b/test/test_default/test_main.cpp
@@ -277,6 +277,10 @@ void test_trafficType_overflowSaturates()
     TEST_ASSERT_EQUAL_UINT32(static_cast<uint32_t>(INT32_MAX), res);
 }
 
+// Required by Unity: PlatformIO's weak defaults do not link on MinGW (PE-COFF weak externals).
+void setUp(void) {}
+void tearDown(void) {}
+
 void setup()
 {
     // Small delay to match other test mains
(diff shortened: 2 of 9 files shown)


caveman99:
The `test_fscommon_getfiles` limitation described above is fixed by meshtastic/framework-portduino#77: `VFSImpl::rmdir()` called `unlink()`, which cannot remove a directory (EISDIR on Linux, EACCES on Windows), so `rmDir()` never cleared the directory tree its depth-limit case creates and the leftovers made a later run report a spurious depth truncation.

Once that lands and the platform pin picks it up, the suite goes green on Windows with no further change here.

thebentern (approved):

PR #11487, opened 2026-08-13
fixes #11466

### LINKED ISSUE #11466: Radio echoes a phone-originated text message back over the client API without an is-from-us marker (2.7.27 regression)
## Summary

When a phone/client app sends a text message, the radio sends a copy of that same packet **back to the originating client** over the phone API (`fromRadio`), and the client has no reliable way to tell that the packet is its own transmission echoed back. On **2.7.27** this became reliable enough that clients started surfacing the user's own message as an incoming message.

This is being reported from the Apple client, but it's a firmware/protocol concern that affects every client equally, so it should be fixed device-side rather than worked around in each app.

## Observed client-side effect

On iOS, sending a message produced a local notification for the user's **own** message, reset that message's unread state (bumping the unread badge), and cleared its delivered indicator — because the echoed packet re-ingested on top of the row the client had already stored for the sent message. Android has long guarded against the same thing (`MeshDataHandlerImpl.rememberDataPacket` dedupes on packet id and marks `read = fromLocal`), and iOS has now added an equivalent guard (meshtastic/Meshtastic-Apple#2226). But both are client-side compensation for the radio handing a client back its own packet.

## Why this should be handled in firmware

`MeshService::handleFromRadio` forwards packets to the phone without an `isFromUs` filter and relies on the duplicate cache to avoid echoing a client's own transmission. That cache can miss (and store-and-forward replay can re-deliver a client's own earlier message), so from the client's side an own-message echo is indistinguishable from a genuine incoming message except by dedupe heuristics on packet id — which every client then has to reimplement.

Two directions, either of which fixes it for all clients:

1. **Don't echo a client's own transmission back to that client** over the phone API (filter on the packet originating from the connected node before enqueuing to `fromRadio`), or
2. If the echo is intentional (delivery/ack confirmation), **mark it** so clients can trivially recognize it — e.g. a from-us / locally-originated flag on the packet delivered to the client — rather than every client inferring it from packet-id dedupe.

## Steps to reproduce

1. Connect a phone client to a node running 2.7.27.
2. Send a text message (channel broadcast or DM).
3. Observe the client receive its own message back over the phone API and treat it as incoming (notification + unread + lost delivered state on clients without a dedupe/self guard).

## Notes

- Regression: this behavior became consistently reproducible on 2.7.27; earlier firmware did not surface it this way.
- Client workarounds now exist on iOS and Android, so this is not user-blocking, but the root cause is the radio returning a client its own packet without a marker, and that belongs in firmware so new/other clients don't each rediscover it.

caveman99:
## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented locally originated packets not addressed to the device from being echoed back to the phone.
  * Continued delivering packets from other devices normally.
  * Preserved delivery of locally addressed acknowledgments and negative acknowledgments.
  * Added handling for packets with an unspecified origin.

* **Tests**
  * Added coverage for local packet suppression, remote packet delivery, and acknowledgment handling.
  * Verified packet delivery across supported routing scenarios.

### DIFF
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -114,6 +114,13 @@ int MeshService::handleFromRadio(const meshtastic_MeshPacket *mp)
         }
     }
 
+    // Our own packet heard back off the mesh, which the duplicate cache only suppresses best-effort.
+    // Clients can't tell an echo from genuine ingress, so it surfaces as an incoming message.
+    if (isFromUs(mp)) {
+        LOG_DEBUG("Skip phone echo of our own packet 0x%08x", mp->id);
+        return 0;
+    }
+
     printPacket("Forwarding to phone", mp);
     if (auto *toPhone = packetPool.allocCopy(*mp))
         sendToPhone(toPhone);
--- a/src/mesh/MeshService.cpp
+++ b/src/mesh/MeshService.cpp
@@ -115,8 +115,9 @@ int MeshService::handleFromRadio(const meshtastic_MeshPacket *mp)
     }
 
     // Our own packet heard back off the mesh, which the duplicate cache only suppresses best-effort.
-    // Clients can't tell an echo from genuine ingress, so it surfaces as an incoming message.
-    if (isFromUs(mp)) {
+    // Clients can't tell an echo from genuine ingress, so it surfaces as an incoming message. Packets
+    // addressed to us are locally-generated feedback (implicit ACK, NAK, routing error), not an echo.
+    if (isFromUs(mp) && !isToUs(mp)) {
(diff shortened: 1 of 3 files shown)


src/mesh/MeshService.cpp
```diff
@@ -114,6 +114,13 @@ int MeshService::handleFromRadio(const meshtastic_MeshPacket *mp)
         }
     }
 
+    // Our own packet heard back off the mesh, which the duplicate cache only suppresses best-effort.
+    // Clients can't tell an echo from genuine ingress, so it surfaces as an incoming message.
+    if (isFromUs(mp)) {
```
thebentern: Won't this break implicit rebroadcast ACKs and potentially locally bound position packets?

PR #11512, opened 2026-08-15
feat(nodedb): ratchet satellite caps down under megamesh pressure

NomDeTom:
In a mesh large enough to churn the hot store, the per-node satellite payloads are the wrong thing to be spending memory on. Measured against NodeInfoLite's 100 B RAM / 112 B encoded, a node carrying all four satellites costs ~452 B RAM and 336 B flash - and EnvironmentMetrics alone is 196 B / 170 B of that, 4-5x every other entry and the least useful thing to persist for a node we will not see again.

NodeDBScalingModule turns MAX_SATELLITE_NODES into a step ladder driven by HopScalingModule's existing population estimate (already sampled, scaled and 13-hour hysteretic, so no new churn counter is needed). Position and telemetry share one ladder because the on-device UI reads both; environment and status get an extra step down. Descent is one step per hourly evaluation and is gated on evidence that the squeeze is on us - a full hot store or evictions since the last check - so a node parked beside a busy repeater in a small mesh does not ratchet on population alone. Release needs six consecutive quiet evaluations per step, and only once population falls below 75% of that step's own entry threshold. Nothing is persisted: the step is re-derived after boot from the histogram's own warm-started estimate.

Builds whose UI reads the satellites keep an absolute floor of entries - 12 for InkHUD, whose map applet draws every positioned node, 8 for other screens, 4 headless. Compiled out entirely on TINY parts and wherever HAS_VARIABLE_HOPS is off, since the population estimate is the module's only input.

Separately, the "heard a new node, introduce ourselves and ask for a reply" handshake in MeshService now gates on a new NodeDB::isPassiveFillOnly() keyed to NODEDB_BASELINE_NODES rather than on isFull(). The two are identical today. The point is the seam: once the hot store is allowed to flex above its baseline, the extra capacity fills passively from observed traffic, so a larger store cannot buy more handshake airtime than a small one - which the current isFull() gate would otherwise do, since a saturated node has that handshake permanently off and an unsaturated one does not.

## Converting the freed budget into node capacity

The reclaimed `nodes.proto` budget is now spent on hot-store slots. Figures are for **nRF52840 /
generic ESP32** (satellite base 40, hot-store baseline 120, `NodeInfoLite` 112 B encoded):

| step | population | pos/tel | env/status | flash freed | +nodes | hot cap | dedup ratio |
|---|---|---|---|---|---|---|---|
| 0 | — | 40 | 40 | 0 B | +0 | **120** | 2.00x |
| 1 | ≥200 | 24 | 24 | 5,376 B | +48 | **168** | 1.43x |
| 2 | ≥400 | 16 | 16 | 8,064 B | +72 | **192** | 1.25x |
| 3 | ≥800 | 8 | 8 | 10,752 B | +96 | **216** | 1.11x |
| 4 | ≥1500 | 8 | 4 | 11,788 B | +105 | **225** | 1.07x |

Five things worth knowing before reading the `+nodes` column as a promise:

**Flash is the currency, and worst case is the right column.** `MAX_NUM_NODES` is 120 on nRF52840
because a saturated worst-case `nodes.proto` (120 × 112 B of `NodeInfoLite` plus 40 satellite slots)
is ~26.9 KB against a 28,672 B LittleFS. A cap has to hold in the worst case — the failure mode is a
failed write, not a stale map. In those terms step 3 frees 37% of the whole filesystem.

**But the freed figures assume full satellite maps.** Environment and status supply most of the
reclaimed bytes and are the sparsest entries in practice (only sensor-bearing nodes create them), so
a typical device reclaims nearer 1.2–2.7 KB. The table is the ceiling a cap is safely sized against,
not a forecast of day-to-day savings.

**Growth can be declined.** `resize()` reallocates, so old and new buffers are briefly both live.
`applyHotStoreCapacity()` demands the new allocation plus an 8 KB margin and otherwise refuses, and
the reported cap only ever reflects storage actually allocated — so nothing can append past the
buffer, including `getOrCreateMeshNode`'s grow-by-one fallback on the packet path.

**The extra capacity is passive-fill only**, per the `isPassiveFillOnly()` seam above: a larger
store cannot buy more handshake airtime than a small one.

**Dedup coverage thins** — the last column. `PACKETHISTORY_MAX` stays sized from the 120-node
baseline by design; growing it would spend the very heap the growth guard protects, and
`PacketHistory` allocates its array once in the constructor. Eviction is LRU, so this degrades
rather than fails and the loss falls on the long tail — but it is a real cost, and some of the
airtime the ratchet reclaims is handed back as re-forwarded duplicates. The measurement that would
settle it is duplicate-rebroadcast rate at step 0 versus step 3 on a saturated node.

Totals are clamped to `NODEDB_MIGRATION_LOAD_CEILING` (250) so a file we write stays inside the
decode allowance every existing build already grants. On LARGE parts the ladder scales ×6.25 but
flash is not the binding constraint, so the clamp absorbs the bonus; on STM32WL the module is
compiled out entirely.

@h3lix1 could you try this out, please?

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

## Summary by CodeRabbit

- **New Features**
  - Node database capacity now adjusts dynamically based on network conditions and available memory.
  - Additional node slots may be unlocked when storage capacity allows.
  - Satellite data limits adjust independently for position, telemetry, environment, and status information.
- **Improvements**
  - Node records are managed more efficiently under pressure, with safer trimming and eviction handling.
  - Automatic NodeInfo responses are reduced when the database is configured for passive filling.
  - Capacity changes now use safeguards to prevent abrupt oscillation or excessive growth.

### DIFF
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -28,6 +28,7 @@
 #include "mesh/generated/meshtastic/deviceonly_legacy.pb.h"
 #include "meshUtils.h"
 #include "modules/NeighborInfoModule.h"
+#include "modules/NodeDBScalingModule.h"
 #include "target_specific.h"
 #if HAS_VARIABLE_HOPS
 #include "modules/HopScalingModule.h"
@@ -728,12 +729,12 @@ template <typename Map> bool evictStalestSatellite(NodeDB &db, Map &map)
     return true;
 }
 
-// Keep `map` within MAX_SATELLITE_NODES ahead of inserting `incoming` (the
-// tier-1/tier-2 split: only the freshest MAX_SATELLITE_NODES nodes carry
-// satellite payloads). Caller holds satelliteMutex.
-template <typename Map> void evictSatelliteOverCap(NodeDB &db, Map &map, NodeNum incoming)
+// Keep `map` within `cap` ahead of inserting `incoming` (the tier-1/tier-2 split: only the
+// freshest `cap` nodes carry satellite payloads). `cap` is the effective, possibly
+// ratcheted-down cap, not MAX_SATELLITE_NODES. Caller holds satelliteMutex.
+template <typename Map> void evictSatelliteOverCap(NodeDB &db, Map &map, NodeNum incoming, uint16_t cap)
 {
-    if (map.size() < MAX_SATELLITE_NODES || map.count(incoming))
+    if (map.size() < cap || map.count(incoming))
         return;
     evictStalestSatellite(db, map);
 }
@@ -1820,7 +1821,7 @@ void NodeDB::setNodeStatus(NodeNum n, const meshtastic_StatusMessage &status)
     (void)status;
 #else
     concurrency::LockGuard guard(&satelliteMutex);
-    evictSatelliteOverCap(*this, nodeStatus, n);
+    evictSatelliteOverCap(*this, nodeStatus, n, nodeDBEnvironmentCap());
     nodeStatus[n] = status;
 #endif
 }
@@ -1832,7 +1833,7 @@ void NodeDB::touchNodePositionTime(NodeNum n, uint32_t time)
     (void)time;
 #else
     concurrency::LockGuard guard(&satelliteMutex);
-    evictSatelliteOverCap(*this, nodePositions, n);
+    evictSatelliteOverCap(*this, nodePositions, n, nodeDBSatelliteCap());
     nodePositions[n].time = time;
 #endif
 }
@@ -1861,35 +1862,41 @@ bool NodeDB::enforceSatelliteCaps()
 {
     concurrency::LockGuard guard(&satelliteMutex);
     bool trimmedAny = false;
-    auto trim = [this, &trimmedAny](auto &map, const char *name) {
+    auto trim = [this, &trimmedAny](auto &map, const char *name, uint16_t cap) {
         const size_t before = map.size();
-        while (map.size() > MAX_SATELLITE_NODES) {
+        while (map.size() > cap) {
             if (!evictStalestSatellite(*this, map))
                 break;
         }
         if (map.size() != before) {
             trimmedAny = true;
-            LOG_MIGRATION("Trimmed %s satellites %u -> %u (cap %d)", name, (unsigned)before, (unsigned)map.size(),
-                          MAX_SATELLITE_NODES);
+            LOG_MIGRATION("Trimmed %s satellites %u -> %u (cap %u)", name, (unsigned)before, (unsigned)map.size(), (unsigned)cap);
         }
     };
+    // Position and telemetry share the UI-facing cap; environment and status take the
+    // deeper one (bigger entries, no map/list reader). See NodeDBScalingModule.h.
+    const uint16_t satCap = nodeDBSatelliteCap();
+    const uint16_t envCap = nodeDBEnvironmentCap();
 #if !MESHTASTIC_EXCLUDE_POSITIONDB
-    trim(nodePositions, "position");
+    trim(nodePositions, "position", satCap);
 #endif
 
 #if !MESHTASTIC_EXCLUDE_TELEMETRYDB
-    trim(nodeTelemetry, "telemetry");
+    trim(nodeTelemetry, "telemetry", satCap);
 #endif
 
 #if !MESHTASTIC_EXCLUDE_ENVIRONMENTDB
-    trim(nodeEnvironment, "environment");
+    trim(nodeEnvironment, "environment", envCap);
 #endif
 
 #if !MESHTASTIC_EXCLUDE_STATUSDB
-    trim(nodeStatus, "status");
+    trim(nodeStatus, "status", envCap);
 #endif
 
-    (void)trim; // all four maps may be compiled out
+    // all four maps may be compiled out (STM32WL), leaving these with no reader
+    (void)trim;
+    (void)satCap;
+    (void)envCap;
 
     // Approximate satellite heap usage: each std::map entry is one rb-tree node,
     // value_type plus ~44 B of node overhead (parent/left/right pointers, color,
@@ -3389,7 +3396,7 @@ void NodeDB::updatePosition(uint32_t nodeId, const meshtastic_Position &p, RxSou
 #else
     {
         concurrency::LockGuard guard(&satelliteMutex);
-        evictSatelliteOverCap(*this, nodePositions, nodeId);
+        evictSatelliteOverCap(*this, nodePositions, nodeId, nodeDBSatelliteCap());
         meshtastic_PositionLite &slot = nodePositions[nodeId]; // creates default-zero entry if missing
 
         if (src == RX_SRC_LOCAL) {
@@ -3447,7 +3454,7 @@ void NodeDB::updateTelemetry(uint32_t nodeId, const meshtastic_Telemetry &t, RxS
         }
 #if !MESHTASTIC_EXCLUDE_TELEMETRYDB
         concurrency::LockGuard guard(&satelliteMutex);
-        evictSatelliteOverCap(*this, nodeTelemetry, nodeId);
+        evictSatelliteOverCap(*this, nodeTelemetry, nodeId, nodeDBSatelliteCap());
         nodeTelemetry[nodeId] = t.variant.device_metrics;
 #endif
 
@@ -3459,7 +3466,7 @@ void NodeDB::updateTelemetry(uint32_t nodeId, const meshtastic_Telemetry &t, RxS
         }
 #if !MESHTASTIC_EXCLUDE_ENVIRONMENTDB
         concurrency::LockGuard guard(&satelliteMutex);
-        evictSatelliteOverCap(*this, nodeEnvironment, nodeId);
+        evictSatelliteOverCap(*this, nodeEnvironment, nodeId, nodeDBEnvironmentCap());
         nodeEnvironment[nodeId] = t.variant.environment_metrics;
 #endif
 
@@ -3979,6 +3986,11 @@ bool NodeDB::isFull()
     return (numMeshNodes >= MAX_NUM_NODES) || (memGet.getFreeHeap() < MINIMUM_SAFE_FREE_HEAP);
 }
 
+bool NodeDB::isPassiveFillOnly()
+{
+    return (numMeshNodes >= NODEDB_BASELINE_NODES) || (memGet.getFreeHeap() < MINIMUM_SAFE_FREE_HEAP);
+}
+
 uint32_t NodeDB::hotNodeLastHeard(NodeNum n) const
 {
     for (int i = 0; i < numMeshNodes; i++)
@@ -4241,6 +4253,7 @@ meshtastic_NodeInfoLite *NodeDB::getOrCreateMeshNode(NodeNum n)
                     meshNodes->at(i) = meshNodes->at(i + 1);
                 }
                 (numMeshNodes)--;
+                hotEvictions++;
             }
         }
         // Don't append past the end of the vector. The protected-node cap
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -2136,7 +2136,7 @@ LoadFileResult NodeDB::loadProto(const char *filename, size_t protoSize, size_t 
 #if WARM_NODE_COUNT > 0
 void NodeDB::demoteOldestHotNodesToWarm()
 {
-    const int keep = MAX_NUM_NODES;
+    const int keep = effectiveMaxNodes();
     if (numMeshNodes <= keep)
         return;
 
@@ -2178,7 +2178,7 @@ void NodeDB::nodeDBSelfCare()
         return;
 
     const NodeNum self = getNodeNum();
-    const bool nodesOverCap = numMeshNodes > MAX_NUM_NODES;
+    const bool nodesOverCap = numMeshNodes > effectiveMaxNodes();
 
     // Confirm self is present and its key matches what we just (re)derived. A
     // non-empty DB that doesn't contain us means a foreign/over-cap or corrupt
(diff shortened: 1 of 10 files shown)


src/mesh/NodeDB.cpp
```diff
@@ -3974,9 +3985,62 @@ bool NodeDB::resolveUniqueLastByte(uint8_t lastByte, bool requireDirectNeighbor,
...
 // returns true if the maximum number of nodes is reached or we are running low on memory
+pb_size_t NodeDB::effectiveMaxNodes() const
+{
+    return hotCapacity ? hotCapacity : (pb_size_t)MAX_NUM_NODES;
+}
+
+pb_size_t NodeDB::desiredMaxNodes() const
+{
+    const uint32_t total = (uint32_t)MAX_NUM_NODES + nodeDBBonusNodes();
+    return (pb_size_t)((total > NODEDB_MIGRATION_LOAD_CEILING) ? NODEDB_MIGRATION_LOAD_CEILING : total);
```
h3lix1, line 3996: **[P1] Never clamp Portduino below its configured baseline**

MAX_NUM_NODES is runtime General.MaxNodes on Portduino, but this expression returns at most 250. The first pressured transition therefore shrinks every MaxNodes > 250; MaxNodes = 1000 removes 750 full records from the hot tier, and from 252 onward even a protected record can be demoted. A warm round trip cannot reconstruct favorite, ignored, or manually-verified flags.

A safe local shape is:

~~~cpp
const uint32_t baseline = static_cast<uint32_t>(MAX_NUM_NODES);
const uint32_t ceiling =
    std::max(baseline, static_cast<uint32_t>(NODEDB_MIGRATION_LOAD_CEILING));
return static_cast<pb_size_t>(
    std::min(baseline + nodeDBBonusNodes(), ceiling));
~~~

Disabling bonus growth when the runtime baseline already exceeds the migration ceiling is also reasonable, but the target must never fall below baseline. **If unchanged:** valid Portduino configurations are silently overridden, full records leave the hot tier, warm admission may replace/refuse identities, and local protected state can be lost.

src/mesh/NodeDB.cpp
```diff
@@ -3974,9 +3985,62 @@ bool NodeDB::resolveUniqueLastByte(uint8_t lastByte, bool requireDirectNeighbor,
...
+
+    const pb_size_t target = desiredMaxNodes();
+    const pb_size_t live = effectiveMaxNodes();
+    if (target == live && meshNodes->size() == target)
+        return;
+
+    if (target > live) {
+        // Both buffers are live across the reallocation, so demand the new one plus a margin:
+        // declining costs only capacity, getting it wrong on a 99%-heap part costs the boot.
+        const size_t needed = (size_t)target * sizeof(meshtastic_NodeInfoLite);
```
h3lix1, line 4012: **[P1] Guard the allocation the vector will actually make**

target × sizeof(...) is not the allocation cost of resize() on a full vector. With the pinned libstdc++, Heltec-v3 200→250 grows capacity to 400: this guard passes with 33,192 B free while the new contiguous block alone needs 40,000 B, before the intended 8 KiB margin. Conversely, if retained capacity already covers target, no allocation occurs but this check can still decline it.

After rebasing to current develop, the nRF ladder becomes 176/205/233/244; 233→244 grows retained capacity 240→466. That allocation is 42,872 B behind a 30,640 B guard.

Make the check conditional on target > meshNodes->capacity(), then use a predictable/failure-aware allocation plan whose actual new contiguous block plus margin is validated. Coordinate this with the pointer-lifetime fix below. **If unchanged:** the guard can pass immediately before allocation aborts/reboots the device, or reject a growth that requires no allocation.

src/mesh/NodeDB.cpp
```diff
@@ -3974,9 +3985,62 @@ bool NodeDB::resolveUniqueLastByte(uint8_t lastByte, bool requireDirectNeighbor,
...
+        if (memGet.getFreeHeap() < needed + NODEDB_GROWTH_HEAP_MARGIN) {
+            LOG_WARN("NodeDB: decline grow %d->%d, %u B free", (int)live, (int)target, (unsigned)memGet.getFreeHeap());
+            return; // hotCapacity untouched: the cap keeps matching what is actually allocated
+        }
+        meshNodes->resize(target);
```
h3lix1, line 4017: **[P1] Make runtime resize safe for escaping node pointers**

A successful resize can free this backing store while the nRF52 Bluefruit authorize path or Portduino web thread consumes a raw pointer returned by readNextMeshNode(); UI and canned-message code also cache element pointers beyond one call. The shrink path sorts/rewrites the same elements. A lock only around this line is insufficient because the pointers escape, and concurrency::Lock is currently a no-op on Portduino.

Please change the access contract to stable NodeNum handles/copy-out or visitor APIs, use storage with the required stability, or add real cross-platform synchronization covering each pointer’s complete lifetime before permitting runtime resize/sort. **If unchanged:** growth can produce a use-after-free/hard fault or corrupt node-list output, and shrink can silently retarget a cached pointer to a different identity.

src/mesh/NodeDB.cpp
```diff
@@ -3974,9 +3985,62 @@ bool NodeDB::resolveUniqueLastByte(uint8_t lastByte, bool requireDirectNeighbor,
...
+        if (numMeshNodes > target) {
+#if WARM_NODE_COUNT > 0
+            demoteOldestHotNodesToWarm();
+#endif
+            numMeshNodes = std::min(numMeshNodes, target); // whatever the warm tier could not take
+        }
+        meshNodes->resize(target);
+        LOG_INFO("NodeDB: hot store %d -> %d slots, %d held", (int)live, (int)target, numMeshNodes);
+    }
+    memaudit::set("nodedb", (size_t)target * sizeof(meshtastic_NodeInfoLite));
```
h3lix1, line 4033: **[P2] Report retained vector capacity to MemAudit**

resize(smaller) changes vector size but normally retains its allocation. On nRF52, shrinking a 240-slot vector to 120 leaves 22,080 B owned while this reports 11,040 B; after the projected post-rebase 466-slot growth, shrinking to 120 would understate ownership by 31,832 B.

Please account the physical allocation here and at the analogous self-care call:

~~~cpp
memaudit::set(
    "nodedb",
    meshNodes->capacity() * sizeof(meshtastic_NodeInfoLite));
~~~

Exposing separate logical and retained figures is also fine. **If unchanged:** field heap diagnostics materially under-report NodeDB memory and can send later OOM investigation toward the wrong subsystem.

src/modules/NodeDBScalingModule.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    step = newStep;
+    refreshCaps();
+    LOG_INFO("NodeDB scaling: step %u - sats %u, env/status %u (of %d), +%u nodes", step, satelliteCap, environmentCap,
+             MAX_SATELLITE_NODES, bonusNodes);
+    // Descending must reclaim immediately; ascending is a no-op here and refills from traffic.
+    if (nodeDB) {
+        nodeDB->enforceSatelliteCaps();
+        // Spend (or hand back) the freed budget. Growth is heap-guarded and may decline.
+        nodeDB->applyHotStoreCapacity();
```
h3lix1, line 99: **[P2] Retry or roll back a declined capacity trade**

setStep() commits the new step and trims the satellite maps before this call. If growth is declined, later evaluations at the same step do not call applyHotStoreCapacity() again; a quiet release requests a smaller target, not the rejected one, and sustained MAX_STEP pressure leaves the rejection permanent.

Have applyHotStoreCapacity() return an applied/pending result and retry with backoff while the desired target remains larger, or explicitly make trimming independent and report that no capacity was funded. **If unchanged:** satellite history can remain discarded even after heap recovers while the bonus hot capacity that justified the trade is never obtained.

test/test_nodedb_scaling/test_main.cpp
```diff
@@ -0,0 +1,455 @@
...
+    }
+}
+
+// A file we write must stay inside the decode allowance every build already grants, or a peer
+// or a downgrade could not read it back.
+void test_effectiveMaxNodesNeverExceedsDecodeCeiling()
+{
+    for (uint8_t i = 0; i < NodeDBScalingModule::STEP_COUNT; i++) {
+        mod->setStep(i);
+        TEST_ASSERT_LESS_OR_EQUAL_UINT32(NODEDB_MIGRATION_LOAD_CEILING, (uint32_t)db->effectiveMaxNodes());
```
h3lix1, line 310: **[P1] Test encoded bytes, not only node count**

This assertion does not establish the file invariant in the comment. getMaxNodesAllocatedSize() adds submessage maxima without each repeated field’s outer tag and length prefix. At this head the 250-entry allowance is 124,002 B, but a maximum-shaped database written by the current encoder can reach 125,502 B; the same 1,500 B gap remains after rebase.

Construct a database containing maximum-shaped entries for every record family, obtain its real encoded size with pb_get_encoded_size(), and assert it fits the production load allowance after that allowance includes framing. **If unchanged:** this test passes while a same-configuration file can exceed its own next-boot decode stream, causing decode failure or silent tail loss.

test/test_nodedb_scaling/test_main.cpp
```diff
@@ -0,0 +1,455 @@
...
+        TEST_ASSERT_LESS_THAN_UINT32((uint32_t)db->effectiveMaxNodes() * 2u, (uint32_t)PACKETHISTORY_MAX);
+}
+
+} // namespace
+
+// Every test starts from an unpressured ladder; setStep() also clears the hysteresis run.
+void setUp(void)
+{
+    if (mod)
+        mod->setStep(0);
```
h3lix1, line 401: **[P2] Reset physical fixture state between tests**

The suite constructs one NodeDB/module for all cases, while setUp() only calls the side-effectful setStep(0). It does not clear maps or rows, reset hotCapacity, or release vector capacity. An early ladder case grows native capacity to 400, so later capacity tests reuse that allocation and do not exercise fresh geometric growth.

Reconstruct the fixture per case, or add a deterministic reset that restores and asserts baseline logical size and physical capacity; test fresh and retained-capacity paths separately. **If unchanged:** the suite remains order-dependent and can stay green while the fresh allocation guard is wrong.

h3lix1 (changes requested):
## Request changes: runtime safety and persistence invariants

I reviewed exact head 25aa56c40740b332c241297019dbcf232f2d2bad and rechecked that it is still the remote PR head immediately before submitting. The ladder/hysteresis logic is coherent, and the added native suite passes 19/19. The remaining problems are in the runtime mutation, allocation, and persistence assumptions used to apply that ladder.

### Merge-blocking changes

1. **Give meshNodes a lifetime-safe access contract.** Runtime resize can reallocate while raw element pointers escape to nRF52 BLE, Portduino web, UI, and canned-message consumers; the shrink path also sorts/reassigns elements. Use stable handles/storage or copy/visitor APIs with real cross-platform synchronization covering the complete use lifetime. **If unchanged:** a pressure transition can cause a use-after-free/hard fault, corrupt a node-list response, or make a cached pointer silently refer to the wrong node.

2. **Preserve the configured platform baseline.** Portduino resolves MAX_NUM_NODES from General.MaxNodes at runtime, while 250 is documented as a migration allowance. desiredMaxNodes() must never return below that configured baseline. This is distinct from the resolved protected-admission thread: even with admission capped at MAX_NUM_NODES - 2, shrinking below MAX_NUM_NODES can demote protected records. **If unchanged:** every Portduino configuration above 250 is silently overridden under pressure; full records leave the hot tier and favorite/ignored/manual-verification state cannot be reconstructed after a warm round trip.

3. **Guard the allocation that will actually occur.** resize() uses geometric vector capacity, not target logical size, and total free heap does not establish a sufficiently large contiguous block. The inline arithmetic shows a current Heltec-v3 failure and a substantially larger second nRF allocation after rebasing onto current develop. **If unchanged:** the heap check can pass immediately before operator new aborts/reboots the device, while retained-capacity growth can also be rejected despite needing no allocation.

4. **Establish a real write/read file bound.** The count-only ceiling test does not account for protobuf framing. At this head the allowance is 124,002 B, while a valid maximum-shaped 250-record database can encode to 125,502 B. Fix the framed load bound and test actual encoded bytes. **If unchanged:** firmware can write a database that its next boot rejects or only partially decodes.

### Persistence policy that needs an explicit decision

The premise that reducing configured satellite caps funds the awarded headers is not a hard filesystem invariant, because the award is based on theoretical full maps rather than entries actually removed:

| Case | Exact PR head | After rebase to current develop |
|---|---:|---:|
| Unpressured maximum-shaped nRF database | 28,802 B | 29,322 B |
| Physical nRF LittleFS partition | 28,672 B | 28,672 B |
| Feasible sparse-satellite/header-heavy model, step 0 → step 4 | 17,258 → 27,746 B | 17,297 → 28,221 B |

The sparse model is not claimed to be typical; it demonstrates that the file is not guaranteed to shrink. Choose a hard encoded-size policy before saving—such as basing awarded capacity on measured removal, preflighting and trimming lower-value records, or enforcing a separate file budget—and treat an expected oversize/out-of-space result differently from filesystem corruption. Direct saveNodeDatabaseToDisk() does not format, but when its failure propagates through saveToDisk(saveWhat), the existing recovery path formats /prefs and retries the same mask. **If unchanged:** a feasible large node-database save can trigger a prefs format, and files outside that save mask can be lost.

### Additional code corrections

- Reset the new capacity state in installDefaultNodeDatabase(), immediately after reconstructing the baseline vector (hotCapacity = 0 is the local fix). **If unchanged:** after a runtime reset of a ratcheted node, effectiveMaxNodes() can still advertise the old larger cap over a baseline-sized vector, allowing packet-path or later unguarded growth before reboot.
- Check and surface the bool returned by warmStore.absorb() in the demotion path. **If unchanged:** a keyless record refused by a warm tier full of keyed entries disappears without any diagnostic, making field loss look like normal demotion.
- Update the PR description’s generated-size and ladder table after rebase. Exact-head nRF caps are 120/174/202/229/240, not 120/168/192/216/225; current develop changes them again to 120/176/205/233/244. **If unchanged:** reviewers and operators are evaluating heap and flash safety from stale numbers, and the second post-rebase allocation is hidden.

### Design decisions (no single code patch implied)

1. **Top-tier ESP32-S3 behavior.** With hot and satellite baselines both 250, the ceiling allows no bonus hot slots, while the ladder can still remove up to 90% of environment/status history. Trimming may still save SRAM and persistence work, so decide whether this is intentionally a trim-only mode or whether scaling should be disabled when no hot growth is possible; log theoretical and actually applied bonuses separately. **If unchanged:** top-tier devices discard satellite history without the advertised hot-capacity benefit, while the step log reports bonus nodes that cannot be applied.

2. **Reboot lifecycle.** The step is not persisted and self-care returns the hot store to baseline before the hourly ladder re-derives pressure. Decide whether to persist/rederive capacity before self-care or explicitly accept and test this churn. **If unchanged:** long-running busy nodes can repeatedly demote identities at reboot, then spend hours regrowing and refilling them; some warm identities may be replaced or refused.

3. **Population-zero semantics.** runOnce() uses zero for both “no rollover yet” and a later valid estimate with no recently heard nodes. Decide whether those states need to be distinguished with a validity/armed flag or bounded grace period. **If unchanged:** a node that ratcheted high and then becomes isolated in the same boot can retain its most aggressive satellite trimming indefinitely.

### Validation required before merge

Please isolate physical fixture state per test and cover a fresh allocating growth, retained-capacity growth, heap rejection plus later retry, Portduino MaxNodes above 250, maximum-shaped encode versus the production load allowance, reset/reboot capacity state, and the chosen pointer-lifetime contract. **If unchanged:** the suite remains order-dependent and can stay green while the fresh allocation, decode, and concurrent-reader failures remain present.

The line comments contain the local code recommendations where a narrow fix is safe; the lifetime and policy items deliberately avoid suggesting a misleading one-line patch.

PR #8251, opened 2025-10-07
Force coverage tests to run in simulation mode

vidplace7:
Re-submit of #8249 

Force meshtastic native tests to run with `-s`

https://docs.platformio.org/en/latest/projectconf/sections/env/options/test/test_testing_command.html

PR #11515, opened 2026-08-15
test(native): add 14 suites for routing, persistence, parsing and identity gaps

thebentern:
Adds 14 native test suites (11 new, 3 extended) covering the highest-value untested logic found in a coverage audit, plus two framing-parser fixes the audit surfaced and one small refactor a test needed. Merged with develop; the full run is **GREEN 68/68, all CLEAN, attribution OK** in the Docker `coverage` env (ASan/LSan), with no accepted-failure baseline.

Since #11517 landed, that run needs no caveat. The `test_packet_signing` B11/B12 failures this description used to carry as accepted are green: they were never caused by this PR, and not #10969 as I first wrote - [@NomDeTom](https://github.com/NomDeTom) root-caused them to `[env:coverage]` passing `-s`, which sets `force_simradio`, which `wouldEncryptWithPKC()` lists among its negative preconditions (Router.cpp:1099). #11517 clears the flag in `initializeTestEnvironment()`, and both now pass.

That revamp also changes what a green run here means. Before it, PlatformIO linked every native suite to the same program and did not relink, so a suite could execute whichever binary was linked last; each suite now builds and runs its own, and `bin/check-test-attribution.py` fails the run if a reported case's source file lies outside the suite claiming it. All 11 new suites report their own cases under that checker. Every suite is also within its `errors=` budget, so `test_nodedb_boot_recovery` needs no declaration - it sits under the 100-line default despite corrupting pref files per test.

**Measured coverage:** 27.70% -> 29.80% line coverage of `src/` from the native unit suites (+763 lines), base `51eadb77d` vs this branch, both full clean runs on the same toolchain. Concentrated where it was aimed: `NodeDBLegacyMigration.cpp` 0% -> 97.5%, `xmodem.cpp` 7.8% -> 95.8%, `ReliableRouter.cpp` 60.8% -> 92.4%, `StreamAPI.cpp` 51.2% -> 89.9%, `RTC.cpp` 55.9% -> 81.2%, `PhoneAPI.cpp` +205 lines, `NodeDB.cpp` +177.

## Source changes

- **`StreamAPI::handleRecStream()`** copied `stream->read()`'s `cInt < 0` EOF idiom into the buffer-fed path, where there is no EOF sentinel: any byte >= 0x80 aborted the parse, and START1 itself is 0x94. Replaced with a direct `uint8_t` cast. Reachability is signedness-dependent rather than universal (per [@NomDeTom](https://github.com/NomDeTom)): plain `char` is unsigned on ARM (nRF52, RP2040), so the old code was correct there, and signed on x86 and Xtensa, so it was broken on ESP32 and native. That is both how it stayed latent and why the native suite catches it.
- **Stray framing marker (`0x94 0x94 0xc3 ...`)** - a byte that failed the START2 check was discarded rather than re-tested as a possible START1, so one corrupted byte on a noisy UART silently dropped the frame behind it. Now re-tested in place, in both the buffer path and the `readStream()` serial path. Originally documented as a known gap in this PR; fixed here instead on review, since it is the same defect shape as the fix above, three functions away.
- **`shouldSkipHandleForPostDecodeHop()`** (NodeDB.h) - the post-decode pre-hop predicate extracted from `Router::dispatchReceived` so a test can drive the exact expression the router calls instead of a hand-copied mirror. No behavior change.

## New suites

**Untrusted-input parsing**
- `test_stream_framing` - ToRadio frame parser on both receive paths: resync after garbage / bogus length / stray framing markers, state persistence across partial reads, the len==512 cap against `rxBuf[516]`, zero-length payloads, back-to-back frames, >=0x80 payload bytes.
- `test_mqtt` (extended) - `shouldDropMqttDownlink` / `onReceiveProto` acceptance gates: pki_encrypted rules, self-echo drop, downlink-disabled channels, hostile topic names, ServiceEnvelope decode bounds, the both-endpoints-known AND gate.
- `test_xmodem` (extended) - `handlePacket` state machine: out-of-sequence / duplicate seq, CRC-mismatch NAK, CAN mid-transfer cleanup, EOT finalize rename, `getForPhone`/`resetForPhone` lifecycle.

**Data integrity & persistence**
- `test_nodedb_boot_recovery` - the degraded-boot identity freeze: corrupt `config.proto` → DECODE_FAILED, keygen skipped (NodeNum == crc32(pubkey), so keygen = renumber), region UNSET, on-disk file fingerprint untouched; self-heal with the original identity; absent config takes the fresh-install path; freeze is config-scoped.
- `test_nodedb_legacy_migration` - v24→v25 on-disk migration from hand-encoded v24 fixtures: field fidelity, PKI key preservation, satellite-map routing, MAX_NUM_NODES truncation, version-gate ladder, `sanitizeUtf8` of hostile v24 names.
- `test_nodedb_v25_roundtrip` - save/reboot/load fidelity: SNR q4 quantization incl. negative + no-SNR sentinel, satellite rehydration, key length, packed bitfield bools.

**Security / identity**
- `test_nodedb_identity_hygiene` - `updateUser` / `addFromContact` key pinning: a mismatched pubkey for a known node cannot silently overwrite, an empty-key update cannot erase, the manually-verified bit survives merges and a cold reboot.
- `test_channel_keys` - `getKey` default-PSK expansion, `generateHash` known answers, PSK-only and name-only differences change the hash, `fixupChannel` cache, no-primary restore.

**Routing reliability**
- `test_reliable_ack_matrix` - one test per cell of the `sniffReceived` ACK/NAK matrix + `shouldSuccessAckWithWantAck`; the #11502 implicit-ACK restore is pinned through the real `perhapsHandleReceived` OPAQUE_RELAY_ONLY ingress path (LoRa stops retries, MQTT does not, foreign traffic never ACKs).
- `test_hop_start_policy` - `classifyHopStart` / `shouldDropPacketForPreHop` / `shouldSkipHandleForPostDecodeHop` truth tables incl. boundaries. The summary case asserts the table it prints rather than narrating it, so the two cannot drift.
- `test_routing_response_hops` - `getHopLimitForResponse` clamping for known / unknown nodes and config bounds.

**Protocol / API**
- `test_phone_api_config_dump` - full WANT_CONFIG dump order ending in the nonce echo, config / moduleConfig oneof variants vs the enum, nodes-only / config-only nonces, heartbeat preempt mid-dump, disconnect/reconnect restart, drain termination.

**Core infrastructure**
- `test_observer` - `Observer.h` lifecycle under ASan: notify order, detach-during-notify, destructor auto-unregister, re-target.

**Time**
- `test_rtc` (extended) - remaining `perhapsSetRTC` arbitration cells (incl. the post-throttle NTP-replaces-GPS drift branch), tm-overload year guard, `getValidTime` gating, `gm_mktime` leap-day / century known answers.

`test/state-manifest.tsv` gains entries for the suites that construct a NodeDB. The develop merge conflicted there only, exactly as NomDeTom predicted - take-both-and-re-sort, keeping develop's `errors=300` on `test_packet_signing`.

## Not fixed, documented in tests

- `Observable::notifyObservers()` walks its `std::list` with a raw iterator, so an observer that unobserves *itself* from `onNotify` and returns 0 would advance past a freed node. No caller does: the only self-detacher, `PhoneAPI::onNotify` -> `checkConnectionTimeout()` -> `close()` -> `unobserve()`, returns -1 and aborts the chain before the increment (PhoneAPI.cpp:1910). `test_self_detach_with_abort_during_notify` pins that contract so the -1 cannot be "cleaned up" later without a test going red. Hardening the dispatch is a behavior change to a header with ~76 `observe()` call sites and belongs in its own PR with hardware validation - not in a test PR.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Native only: Docker `coverage` env full run (ASan/LSan) + `native-macos` build. **No hardware regression testing performed.**

Two of the three source changes cannot alter on-device behavior: the `handleRecStream()` signedness fix is on a buffer path with no `develop` callers, and `shouldSkipHandleForPostDecodeHop()` is a pure predicate extraction of the expression the router already evaluated.

The stray-marker fix **does** change behavior on device, in `readStream()` - the serial/BLE/TCP path every phone client uses. It is one line per copy and strictly widens what the parser accepts: a byte that fails START2 is re-tested as START1 instead of discarded, so a frame that used to be dropped after a stray 0x94 is now delivered. No frame that parsed before parses differently. Covered by three new cases in `test_stream_framing`, each driving both receive paths, but not exercised on hardware - worth a second pair of eyes given where it sits.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved packet filtering to consistently reject invalid non-local packets while preserving local traffic.
  - Fixed stream parsing for high-bit payload bytes and improved recovery when frame markers appear in malformed input.

- **Tests**
  - Expanded coverage for channel security, routing, MQTT, persistence, identity handling, phone configuration transfers, RTC behavior, stream framing, observer notifications, reliable acknowledgements, and XModem transfers.
  - Added regression coverage for database migration, reboot persistence, hop validation, and malformed or truncated network data.


src/mesh/NodeDB.h
```diff
@@ -223,6 +223,19 @@ inline bool shouldDropPacketForPreHop(const meshtastic_MeshPacket &p)
 #endif
 }
 
+/// Post-decode half of the pre-hop drop: the bitfield proving the origin populated hop_start is
+/// channel-encrypted, so MISSING_OR_UNKNOWN only becomes decidable once the packet is decoded.
+/// Local-origin packets are exempt. Router::handleReceived uses this to set skipHandle.
```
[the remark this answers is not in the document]
thebentern: Correct on both counts - the call site is `Router::dispatchReceived()`, not `handleReceived`. Applied the suggested two-line comment verbatim.

test/test_hop_start_policy/test_main.cpp
```diff
@@ -0,0 +1,322 @@
...
+    p.which_payload_variant = meshtastic_MeshPacket_decoded_tag;
+    p.decoded.has_bitfield = hasBitfield;
+    p.decoded.bitfield = hasBitfield ? 1 : 0;
+    return p;
+}
+
+static void assertClassify(const meshtastic_MeshPacket &p, HopStartStatus expected, const char *label)
+{
+    HopStartStatus got = classifyHopStart(p);
+    TEST_MSG_FMT("%-44s hop_start=%u hop_limit=%u -> %d (expect %d)", label, p.hop_start, p.hop_limit, (int)got, (int)expected);
```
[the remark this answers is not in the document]
thebentern: Done - `p.hop_start` / `p.hop_limit` are now cast to `unsigned` at the `TEST_MSG_FMT` call site, and the two comment blocks are down to two lines each.

test/test_nodedb_boot_recovery/test_main.cpp
```diff
@@ -0,0 +1,409 @@
...
+//     maybe-transient on-disk file. The 2.8 total-wipe cascade rode the unprotected path.
+//   - An ABSENT config (loadProto never returns NOT_FOUND; an unopenable file is
+//     OTHER_FAILURE) takes the normal fresh-install path: defaults + new keypair.
+//   - The freeze is config-scoped: a corrupt nodes.proto alone must not freeze identity.
+//   - An old-version devicestate recovers owner fields from our own NodeDB entry.
+//
+// The tests form a deliberate ladder (state=per-suite in test/state-manifest.tsv): each test
+// arranges the on-disk /prefs files, then "reboots" by constructing a fresh NodeDB, the same
+// pattern test_firmware_edition uses. RAM globals that a real cold boot would find zeroed
+// (nodeDatabase) are reset before each reboot so the in-process reboot matches the device.
```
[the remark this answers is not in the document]
thebentern: Trimmed the 16-line header to three lines describing the test contract only; the wipe-cascade narrative is gone. Same pass applied to the other new suites in this PR.

test/test_nodedb_legacy_migration/test_main.cpp
```diff
@@ -0,0 +1,581 @@
...
+// Each test hand-encodes a legacy-format /prefs/nodes.proto (via the same
+// meshtastic_NodeDatabase_Legacy descriptor the migration decodes with), then
+// cold-boots a real NodeDB so the full ladder runs: version gate -> legacy
+// decode -> field-by-field copy (bools packed into the bitfield, names
+// sanitized, position/device_metrics routed into the satellite maps) ->
+// deferred re-save as v25 (migrationSavePending).
+//
+// The sanitizeUtf8 assertions matter beyond cosmetics: an unsanitized v24 name
+// makes the later nanopb encode fail, and a failed save is what triggers
+// saveToDisk()'s fsFormat() wipe - this suite pins the firewall in front of it.
```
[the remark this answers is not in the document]
thebentern: Trimmed to three lines. The migration detail now lives in the test names and assertions.

test/test_observer/test_main.cpp
```diff
@@ -0,0 +1,365 @@
...
+// guarantees that erasing any node other than the one the iterator points at leaves the iterator
+// valid, so detaching an earlier, later, or even the immediately-next observer mid-notify is
+// well-defined today - those cases are asserted below, and the coverage env's ASan turns any
+// future iterator-invalidation regression (e.g. a switch to std::vector) into a clean red.
+//
+// Deliberately NOT tested: an observer calling unobserve(observable) on ITSELF from onNotify.
+// That frees the node under the iterator and the following ++iterator is use-after-free - real,
+// field-reachable UB (PhoneAPI::onNotify -> checkConnectionTimeout -> close() -> unobserve of
+// service->fromNumChanged). A test asserting the crash would be asserting UB; harden
+// notifyObservers first, then add the self-detach case here.
```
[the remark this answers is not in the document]
thebentern: Hardened, but not with the obvious fix - that one is a regression, so flagging the reasoning.

Hoisting `std::next(iterator)` before dispatch (the usual idiom for this) **inverts** the hazard rather than removing it: self-detach becomes safe, but detaching the *immediately next* observer then leaves the hoisted iterator dangling. That case is safe today and `test_detach_of_immediately_next_observer_during_notify` covers it, so the naive fix trades one use-after-free for another.

Implemented deferred removal instead: `removeObserver()` during a dispatch nulls the entry rather than erasing the node, and the outermost `notifyObservers()` sweeps the nulls when it finishes. No node is erased while an iterator is live, so self-detach, next-detach, arbitrary-detach and destruction-during-notify are all safe. Cost is two bools and a sweep only when a removal actually happened; nested notifies only sweep at the outermost level.

One correction to the premise, though: this is not currently reachable as a crash. The only self-detaching observer in the tree is `PhoneAPI::onNotify` -> `checkConnectionTimeout()` -> `close()` -> `unobserve(&service->fromNumChanged)`, and it is safe **because it returns -1**, which aborts the chain before the increment (see the comment at PhoneAPI.cpp:1910). That is an unwritten contract - any observer that self-detaches and returns 0 corrupts the dispatch - and removing it is the actual value here.

Added `test_self_detach_during_notify_still_notifies_rest` (self-detach returning 0: later observers still run, no leftover null entry on the next pass) and `test_self_detach_with_abort_during_notify` (PhoneAPI's abort pattern). Full 68-suite run green under ASan/LSan.

test/test_phone_api_config_dump/test_main.cpp
```diff
@@ -0,0 +1,578 @@
+// PhoneAPI::getFromRadio() config-dump state machine. The header comment over getFromRadio()
+// says the client apps ASSUME THIS SEQUENCE, DO NOT CHANGE IT - this suite is the assertion of
+// that sequence, on decoded FromRadio protobufs rather than internals:
+//   my_info -> deviceuiConfig -> own node_info -> metadata -> region_presets -> 8 channels ->
+//   configs -> moduleConfigs -> other node_infos -> fileInfo* -> config_complete_id (nonce echo)
+// plus the heartbeat preempt, the SPECIAL_NONCE_ONLY_* jumps, mid-dump restart, and the
+// post-complete drain reaching idle (the infinite-drain regression class).
```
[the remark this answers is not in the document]
thebentern: Trimmed to three lines - the sequence itself is asserted in `test_full_want_config_dump_emits_documented_sequence`, which is the better place for it.

thebentern:
Pushed 129806473 addressing the review.

**Observer dispatch (the one substantive change).** `Observable::notifyObservers()` erased list nodes while holding an iterator into them, so an observer that unobserves itself from `onNotify` corrupted the walk. Worth noting the suggested fix - hoisting `std::next` before dispatch - would have *inverted* the hazard rather than removed it: self-detach becomes safe, detaching the immediately-next observer becomes a use-after-free, and that case is safe today and covered by `test_detach_of_immediately_next_observer_during_notify`. Went with deferred removal instead (null the entry during dispatch, sweep at the outermost notify), which keeps every mutation case safe with no allocation.

Not currently a reachable crash: the only self-detacher in the tree is `PhoneAPI::onNotify` -> `checkConnectionTimeout()` -> `close()` -> `unobserve()`, and it survives only because it returns -1 and aborts the chain before the increment (PhoneAPI.cpp:1910). That unwritten contract is what this removes.

**Also in this push:** corrected the documented caller of `shouldSkipHandleForPostDecodeHop` (`dispatchReceived`, not `handleReceived`), unsigned casts at the `%u` call site, and the new suites' file headers trimmed to the one-or-two-line rule in AGENTS.md.

**Trunk CI fix:** trufflehog flagged 13 "secrets" that were all test function names - a Lob API key is `test_` + 35 chars, so any test named exactly that length matches. Renamed the eight offenders. Might be worth a repo-level exclusion for `test/**` eventually; anyone writing a 35-char test name will hit this again.

Full 68-suite Docker coverage run (ASan/LSan) matches the pre-change baseline - only the pre-existing `test_packet_signing` B11/B12 failures from #10969. `native-macos` builds clean.

test/test_observer/test_main.cpp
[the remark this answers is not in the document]
thebentern: Reverted in 784e539d6 - walking this back, and not doing it in this PR.

Two reasons, in order of weight.

**Scope.** This PR adds test coverage. Nothing in the audit asked for a dispatch change, no test required one, and the two regression tests I added alongside it existed only to justify the change I had just made. A behavior change to a header with ~76 `observe()` call sites, validated natively with no hardware soak, does not belong in a test PR - it needs its own review and its own risk budget.

**The premise was overstated, and the suggested fix was wrong.** "Confirmed use-after-free" is not accurate: no caller reaches it. The only observer in the tree that unobserves itself from `onNotify` is `PhoneAPI` (`onNotify` -> `checkConnectionTimeout()` -> `close()` -> `unobserve()`), and it returns -1, which aborts the chain before the iterator advances past the erased node - deliberately, per the comment at PhoneAPI.cpp:1910. And the fix as proposed (hoist `std::next` before dispatch) would have *inverted* the hazard rather than removing it: self-detach becomes safe, detaching the immediately-next observer becomes the use-after-free. That case is safe today and `test_detach_of_immediately_next_observer_during_notify` covers it, so applying the suggestion as written would have shipped a regression with a green suite.

What stays is the part that belongs in a test PR: `test_self_detach_with_abort_during_notify` passes against the unmodified dispatch and pins the fact that PhoneAPI's -1 is load-bearing, so a future cleanup that simplifies it away turns a suite red instead of corrupting a dispatch. The unsafe variant is documented in a comment rather than tested - asserting it would mean asserting UB.

If the dispatch should be hardened, deferred removal (null the entry during dispatch, sweep at the outermost notify) is the shape that works - I had it passing the full 68-suite run - but it wants its own PR and on-device validation.

thebentern:
Correcting my earlier comment: the `Observer.h` dispatch change is **reverted** as of 784e539d6. `src/Observer.h` is now byte-identical to `develop`.

It was scope creep on my part - a review bot asked for it, and I implemented it inside a PR whose job is adding test coverage. It changes dispatch in a header with ~76 `observe()` call sites, and it had native verification only. That needs its own PR, not a drive-by.

Worth recording that the finding was also overstated: it is not reachable today. The one self-detaching observer, `PhoneAPI::onNotify` -> `checkConnectionTimeout()` -> `close()` -> `unobserve()`, is safe because it returns -1 and aborts before the increment (PhoneAPI.cpp:1910). And the specific fix suggested - hoisting `std::next` before dispatch - would have inverted the hazard, breaking the immediately-next-detach case that is safe today and covered by an existing test.

`test_self_detach_with_abort_during_notify` stays, because it passes against unmodified code and pins PhoneAPI's -1 as load-bearing. The hazard itself is now listed under "Not fixed, documented in tests" in the PR description.

**Remaining source changes in this PR are the two the audit actually produced:** the 3-line `StreamAPI` buffer-path fix (latent, no `develop` callers, and its test would otherwise have to codify the bug as correct) and the `shouldSkipHandleForPostDecodeHop()` predicate extraction (no behavior change). Full 68-suite Docker run green, matching baseline.

NomDeTom:
Heads-up on one line in the description, and a small merge note — #11517 overlaps this slightly.

**B11/B12 are fixed in #11517, and the cause is not #10969.**

`test_packet_signing`'s `test_B11_normal_unicast_still_uses_pki` and `test_B12_licensed_receiver_does_not_decrypt_pki` fail because `[env:coverage]` passes `-s` to the test binary (added in 74e6723a, #8251). That sets `portduino_config.force_simradio`, and `wouldEncryptWithPKC()` lists `!force_simradio` among its preconditions, so `perhapsEncode()` takes the channel-crypto branch, returns `NONE`, and leaves `pki_encrypted` false — which is exactly what both assertions check. Same binary, same commit, pristine `$HOME`:

```
$ HOME=$(mktemp -d) ./.pio/build/coverage/meshtasticd        # 77 Tests 0 Failures
$ HOME=$(mktemp -d) ./.pio/build/coverage/meshtasticd -s     # 77 Tests 2 Failures
```

`[env:native]` passes no such flag, which is the whole of the "passes under native, fails under coverage" split. #11517 clears the flag once in `initializeTestEnvironment()`, after `portduinoSetup()` has already chosen the simulated radio, so every suite exercises the production PKI path without touching hardware. Worth dropping those two from the accepted baseline here once either lands — they should be green.

**Merge note.** Trial-merging the two branches conflicts in `test/state-manifest.tsv` only: we both insert rows just after `test_admin_session_repro`, so it is take-both-and-re-sort. `src/mesh/Router.cpp` auto-merges — your hunk is in `dispatchReceived` (~1443), ours is the admin-key fallback budget (~821).

**One thing that may surface on first run.** #11517 adds a per-suite `LOG_ERROR` budget to the same manifest, defaulting to 100 lines for undeclared suites (50 of the current 57 are already under 64). `test_nodedb_boot_recovery` deliberately corrupts pref files per test, and `NodeDB`'s decode-failure path logs at ERROR — that one is genuinely an error, so it is deliberately not demoted. If that suite runs over the default it will report AMBER and want an `errors=` declaration with a reason, same shape as the `writes=` entries you have already added.

NomDeTom (commented):
Read through this looking for the harness traps we hit in #11517, and it is clean on every one: manifest entries match exactly the suites that construct a NodeDB (`test_channel_keys` documents staying disk-write free, `test_rtc` writes nothing), no duplicate test names across suites, no `RUN_TEST` function without an assertion, and `UNITY_END()` handled correctly everywhere. Four notes below, one of which I think is worth acting on in this PR.

test/test_stream_framing/test_main.cpp
```diff
@@ -0,0 +1,359 @@
...
+
+    TEST_ASSERT_EQUAL_UINT8_ARRAY(streamApi.deliveries[0].data(), bufferApi.deliveries[0].data(), payload.size());
+}
+
+/// Document current behavior: a stray START1 immediately before a real frame loses that frame.
+/// After START1 + non-START2 the parser resets WITHOUT re-considering the failing byte as a new
+/// START1, so in 0x94 0x94 0xc3 ... the real frame's own marker was consumed by the reset. This
+/// is a known robustness gap being pinned as-is, not an endorsement; a fix would re-examine the
+/// failing byte. The parser must still recover on the next clean frame.
+void test_stray_start1_before_frame_documents_frame_loss()
```
NomDeTom: **Worth fixing here rather than pinning.** This asserts that one stray `0x94` destroys the frame that follows it, which is the same shape as the bug this PR fixes three functions away: a framing parser mishandling a byte it should have re-examined. The fix is small - on `START1 + non-START2`, re-test the failing byte as a possible new START1 instead of resetting past it - and the fixture to prove it is already right here.

The cost of leaving it is that a single corrupted byte on a noisy UART silently drops the next frame, and there is now a passing test asserting that is correct, which makes it harder to change later. If it is deliberately out of scope, a linked issue would keep it from ossifying.

src/mesh/StreamAPI.cpp
```diff
@@ -86,12 +86,9 @@ int32_t StreamAPI::handleRecStream(const char *buf, uint16_t bufLen)
...
     while (bufLen > index) { // Currently we never want to block
-        int cInt = buf[index++];
-        if (cInt < 0)
-            break; // We ran out of characters (even though available said otherwise) - this can happen on rf52 adafruit
-                   // arduino
-
-        uint8_t c = (uint8_t)cInt;
+        // Unlike stream->read(), a buffer byte has no EOF sentinel: bufLen already bounds the loop,
+        // and a signed-char comparison would treat any byte >= 0x80 (START1 included) as EOF.
+        uint8_t c = (uint8_t)buf[index++];
```
NomDeTom, line 91: Correct fix. Worth noting in the description that reachability is signedness-dependent rather than universal: `char` signedness is implementation-defined, and it is **unsigned on ARM** (nRF52, RP2040) but signed on x86 and Xtensa. So `cInt < 0` never fired on ARM - the old code was correct there and broken on ESP32 and native.

That explains how it stayed latent, and it means a backport matters for some targets and not others. It is also why the native suite catches it: x86 is a signed-char platform.

test/test_reliable_ack_matrix/test_main.cpp
```diff
@@ -0,0 +1,851 @@
...
+    TEST_ASSERT_EQUAL_INT(ERRNO_OK, reliableShim->send(allocated));
+
+    TEST_ASSERT_EQUAL_UINT32(2, reliableShim->pendingCount());
+    TEST_ASSERT_EQUAL_UINT32(aBefore + 50000, reliableShim->pendingNextTx(kLocalNode, a.id));
+
+    // B's deadline is millis-at-set + getRetransmissionMsec(B); a self-extension would push it a
+    // further 50s out, past anything the wall clock could account for.
+    uint32_t bTx = reliableShim->pendingNextTx(kLocalNode, b.id);
+    uint32_t retrans = radio->getRetransmissionMsec(reliableShim->pendingPacket(kLocalNode, b.id));
+    TEST_ASSERT_TRUE_MESSAGE(bTx - retrans <= millis(), "own record must not be extended by its own send");
```
NomDeTom: `bTx - retrans <= millis()` is the idiom the `Naive millis() Deadline Compare` job rejects. That guard only scans `src/`, so this is exempt, but the house rule reads as universal. Harmless in a seconds-long process - flagging only for consistency.

test/state-manifest.tsv
```diff
@@ -45,13 +45,20 @@ test_admin_session_repro	writes=config.proto,module.proto,device.proto,channels.
...
 test_mesh_beacon	writes=module.proto	exercises the beacon's module-config save path
 test_mesh_module	writes=config.proto,module.proto,device.proto,channels.proto,nodes.proto,warm.dat	module framework tests construct a NodeDB
 test_mqtt	writes=config.proto,module.proto,device.proto,channels.proto,nodes.proto	constructs a NodeDB for node lookups in the MQTT paths
 test_nexthop_routing	writes=config.proto,module.proto,device.proto,channels.proto,nodes.proto	next-hop selection reads and updates the node DB
 test_nodedb_blocked	state=per-suite writes=config.proto,module.proto,device.proto,channels.proto,nodes.proto,warm.dat	saturates the DB with MAX_NUM_NODES-2 favourited nodes to test the protected cap; a later test's removeNodeByNum() persists that state, and the cap test depends on the fill from the test before it
+test_nodedb_boot_recovery	state=per-suite writes=config.proto,module.proto,device.proto,channels.proto,nodes.proto	deliberate boot-recovery ladder: corrupts/deletes/restores the pref files and reboots a NodeDB per test to pin the DECODE_FAILED identity freeze, so each test observes the previous test's on-disk state
```
NomDeTom, line 63: Once #11517 lands, this suite may want an `errors=` budget alongside the `writes=`. It corrupts pref files per test on purpose, and NodeDB's decode-failure path logs at ERROR - deliberately not demoted there, since local storage corruption is a real error - so it may exceed the 100-line default and report AMBER. Same declare-and-justify shape as the `writes=` entry; happy to measure the number if useful.

Also note the two PRs conflict in this file only: we both insert rows just after `test_admin_session_repro`, so it is take-both-and-re-sort. `src/mesh/Router.cpp` auto-merges.

test/test_stream_framing/test_main.cpp
thebentern: Agreed, and fixed rather than pinned. You are right that pinning it was the wrong call - a passing test asserting the bad behavior is exactly what makes it hard to change later, and it is the same defect shape as the fix three functions away.

The fix is the one you described: on START1 followed by non-START2, re-test the failing byte as a possible new START1 instead of resetting past it. One line in each copy of the state machine:

```c
if (c != START2)
    rxPtr = (c == START1) ? 1 : 0;
```

Applied to **both** `handleRecStream()` and `readStream()`. The second one matters more than the first for your noisy-UART case - `readStream()` is the production serial path, and `handleRecStream()` still has no callers on develop - so fixing only the buffer copy would have left the real-world case broken.

The test is now `test_stray_start1_before_frame_still_delivers`, asserting delivery on both receive paths, plus two more: a run of three stray markers (`0x94 0x94 0x94` + frame), and START1 followed by an unrelated byte, which must still resync on the next real frame.

Noting for the record that the two state machines are byte-for-byte duplicates apart from the EOF check, which is how they can drift - `handleRecStream()` had the signedness bug and `readStream()` legitimately does not, since `stream->read()` really does return -1. Extracting the per-byte step into one helper would remove the duplication, but that is a refactor of the same function this PR is already changing, so I have left it; happy to do it as a follow-up if you think it is worth it.

src/mesh/StreamAPI.cpp
thebentern: Thanks - that is a better explanation than the one I shipped, and I have corrected the description. Plain `char` is unsigned on ARM (AAPCS) and signed on x86 and Xtensa, so the old code was correct on nRF52 and RP2040 and broken on ESP32 and native. That is both why it stayed latent and why the native suite is what catches it. Reworded from "latent, no callers" to state the platform split explicitly, since a backport decision depends on it.

test/test_reliable_ack_matrix/test_main.cpp
thebentern: Fixed. Now `Throttle::deadlinePassed(bTx - retrans)`, which is the same predicate, wrap-safe, and consistent with the house rule even though the guard only scans `src/`. You are right that the rule reads as universal; a test that models a deadline compare is a bad place to demonstrate the idiom the guard rejects.

test/state-manifest.tsv
thebentern: Measured it so you do not have to. From a clean full run of this branch (68 suites, coverage env), counting `LOG_ERROR` lines within each suite's section:

| suite | ERROR lines |
| --- | --- |
| `test_nodedb_boot_recovery` | 26 |
| `test_nodedb_legacy_migration` | 18 |
| `test_nodedb_v25_roundtrip` | 11 |

So the corruption-heavy suite lands at ~26 against your 100-line default - comfortably inside it, and inside the 64 you mention most suites sit under. No `errors=` declaration needed on current behavior. Caveat: that count is from the coverage env with `-s`, and I do not know whether #11517 counts the isolate wrapper's own output, so treat it as the right order of magnitude rather than the exact number your budget will see.

On the conflict: agreed it is take-both-and-re-sort in `state-manifest.tsv`, and agreed `Router.cpp` auto-merges - my hunk is the post-decode hop predicate in `dispatchReceived` (~1443), nowhere near the admin-key fallback budget. Whichever of us lands second takes the re-sort; happy for that to be me.

thebentern:
@NomDeTom thanks - that is the most useful review this PR has had. All four addressed in 3a06f8cd6, threads replied to and resolved.

**B11/B12: you are right and I was wrong.** I asserted "#10969" from a stale note and never checked it. Verified your chain in the source: `[env:coverage]` passes `-s` (variants/native/portduino/platformio.ini:139), `-s` sets `force_simradio` (PortduinoGlue.cpp:118), and `wouldEncryptWithPKC()` lists `!portduino_config.force_simradio` among its preconditions (Router.cpp:1099), while `[env:native]` passes no flag. That is the entire native-vs-coverage split. Description corrected to cite the real cause and #11517; happy to drop those two from the accepted baseline as soon as either lands.

**The stray marker is now fixed rather than pinned.** Your framing of it convinced me - a passing test asserting the bad behavior is what ossifies it, and it is the same defect shape as the signedness fix three functions away. Worth noting the fix had to go in **both** copies of the state machine: I would have shipped a half-fix if I had only patched `handleRecStream()`, since `readStream()` is the path that actually sees your noisy UART and `handleRecStream()` has no callers on develop at all.

**errors= budget: measured, no declaration needed.** From a clean 68-suite run of this branch, `test_nodedb_boot_recovery` emits **26** `LOG_ERROR` lines (legacy_migration 18, v25_roundtrip 11) - well inside your 100 default and inside the 64 most suites sit under. Caveat: counted in the coverage env with `-s`, and I do not know if #11517 counts the isolate wrapper's own output, so treat it as the right order of magnitude. If your budget sees a different number when it lands I will add the `errors=` row with a reason.

**Merge:** agreed, `state-manifest.tsv` is the only conflict and it is take-both-and-re-sort; `Router.cpp` auto-merges (my hunk is the post-decode hop predicate in `dispatchReceived`, nowhere near your admin-key fallback budget). Whichever lands second takes the re-sort - happy for that to be me, and I would rather #11517 go first so B11/B12 are green here.

Full 68-suite run green after the change, `native-macos` builds clean.

PR #11532, opened 2026-08-17
Add CrowPanel Advance 4.3/5.0/7.0 ESP32-S3 support

danadash123-glitch:
## What changed

Adds dedicated support for the Elecrow CrowPanel Advance 4.3, 5.0, and 7.0 inch ESP32-S3 panels, including the v1.4/v1.5 hardware family.

- adds STC controller support on I2C for backlight/brightness and buzzer control
- shares and locks the I2C bus between STC, GT911 touch, and INA219
- uses the native v1.2+ SX1262 header wiring with NSS/CS on GPIO 8
- adds INA219 device-battery monitoring and battery percentage
- persists display brightness, screen timeout, and touch calibration
- reports hardware model 144 (`CROWPANEL_ADV_43_50_70`)
- honors Meshtastic buzzer and External Notification settings for boot and received-message alerts
- fetches Google map tiles asynchronously over Wi-Fi without checking SD for tiles first
- includes the hardware-tested application binary and usage notes

## Why

The existing large-panel variant treated GPIO 8 as a directly driven buzzer pin and assigned SX1262 CS to GPIO 0. On v1.2+ large panels the buzzer is controlled by the onboard STC over I2C, while the radio header routes NSS/CS to GPIO 8. The missing STC path also prevented reliable brightness, touch, buzzer, and shared-I2C behavior.

The map view previously depended on SD-backed tile handling. Until the large-panel SD interface is resolved, this patch uses an asynchronous URL tile service so maps work through an active Wi-Fi connection.

## User impact

The panels can use their original GPIO 8 radio-CS routing without a GPIO 0 jumper. The radio, touch, brightness, persisted display settings, INA219 battery percentage, message buzzer, Wi-Fi, and online maps have all been exercised on physical hardware.

For a clear message alert, set Device > Buzzer Mode to **All Enabled**, enable **Alert Message Buzzer** under External Notification, leave I2S disabled, and use an output duration of about 1000 ms.

## SD-card status

SD-card initialization on the tested v1.4/v1.5 hardware is still a work in progress. SD-backed/offline map tiles are not claimed as working. Maps currently use Google tiles over Wi-Fi. The UI's Backup & Restore function also remains unavailable when the SD card does not initialize because it stores key backups on SD.

## Validation

- `platformio run -e elecrow-adv1-43-50-70-tft` completed successfully
- application image flashed to an ESP32-S3 on COM3 at `0x10000`
- esptool verified the written data hash
- SX1262 communication works with native CS on GPIO 8
- Wi-Fi Google map tiles load on the panel
- STC boot and received-message buzzer tested
- INA219 battery percentage tested
- brightness, timeout, and touch calibration persistence tested

Included binary SHA-256:

`B7EFB4D9855630F8AD0E7D0B14B3DD83AC17248093EE97EAA912FB7C713CFD0C`
## Hardware registration and device category request

We would like this large-panel family to be registered as a separate supported hardware category instead of being identified as the existing small `CROWPANEL` / CrowPanel Adv 2.4/2.8 TFT device. The proposed entry is **CrowPanel Advance 4.3/5.0/7.0 TFT** with the slug `CROWPANEL_ADV_43_50_70`.

This patch uses hardware model ID **144** so neighboring nodes and Meshtastic clients can distinguish the large panels correctly. ID 144 is not otherwise assigned in the generated enum in this checkout, but the number is proposed rather than mandatory: we are happy to update the firmware to whatever official hardware-model number the Meshtastic maintainers allocate.

We would also appreciate the corresponding supported-hardware metadata/client entry so the flasher and Meshtastic apps show a dedicated CrowPanel Advance 4.3/5.0/7.0 device card rather than the 2.4/2.8 inch CrowPanel card.

## Hardware photo

CrowPanel Advance 7-inch v1.5 running Meshtastic

The photo shows the tested CrowPanel Advance 7-inch v1.5 board, installed wireless module, and Meshtastic Device UI.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631


src/mesh/generated/meshtastic/mesh.pb.h
```diff
@@ -329,6 +329,8 @@ typedef enum _meshtastic_HardwareModel {
     meshtastic_HardwareModel_SEEED_WIO_TRACKER_L2 = 137,
     /* Elecrow CrowPanel Advance P4 models, ESP32-P4 and TFT with SX1262 radio plugin */
     meshtastic_HardwareModel_CROWPANEL_P4 = 138,
+    /* Elecrow CrowPanel Advance 4.3, 5.0 and 7.0 inch models with ESP32-S3 and TFT */
```
mverch67: Tell your AI not to modify generated code. This is not allowed and will be overwritten by code-generation resulting in a non-compilable code base.

variants/esp32s3/elecrow_panel/firmware/Meshtastic-CrowPanel-4.3-5.0-7.0-v2.8.0-STC-GoogleMaps-GPIO8-CS.bin
mverch67, line 1: Do not put in .bin files of unknown origin. According to the name of this file it is likely to violate licenses.

variants/esp32s3/elecrow_panel/patch_device_ui.py
mverch67, line 1: This is a poor way of implementing things. We don't want self-modifying code which cannot be debugged using common tools.

variants/esp32s3/elecrow_panel/platformio.ini
```diff
@@ -60,6 +59,7 @@ custom_sdkconfig =
 extends = crowpanel_base
```
mverch67, line 60: These code changes below will break existing devices. Make sure that these are still supported by any changes. E.g. create a new target for new releases if there is no way for runtime detection of the hardware version.

mverch67 (changes requested):

danadash123-glitch:
@mverch67 Thank you for the review. I have addressed the requested changes in commit `361cc30d6`:

- Reverted the manual edit to `src/mesh/generated/meshtastic/mesh.pb.h`. The new hardware-model ID is no longer introduced in generated firmware code; the target temporarily uses the existing `CROWPANEL` model (97) until an appropriate protobuf change is made separately.
- Removed the compiled `.bin` file from the repository.
- Removed `patch_device_ui.py` and the self-modifying Device UI patch hook.
- Restored the existing `elecrow-adv1-43-50-70-tft` target and its GPIO behavior to the pre-PR configuration.
- Added a separate `elecrow-adv1-43-50-70-tft-v12` target for the v1.2+ panels that require the STC controller and GPIO8 SX1262 chip select.

Both targets were built successfully:

- `elecrow-adv1-43-50-70-tft`
- `elecrow-adv1-43-50-70-tft-v12`

The v1.2+ build has also been flashed to and tested on the physical panel. Please take another look when convenient.

src/mesh/generated/meshtastic/mesh.pb.h
danadash123-glitch: Addressed in 361cc30d6. The manual generated-code changes were removed; this update does not modify src/mesh/generated/ for the CrowPanel hardware model and reuses the existing CROWPANEL model (97).

variants/esp32s3/elecrow_panel/firmware/Meshtastic-CrowPanel-4.3-5.0-7.0-v2.8.0-STC-GoogleMaps-GPIO8-CS.bin
danadash123-glitch: Addressed in 361cc30d6. The externally sourced firmware binary was removed from the PR. The branch now contains source and normal Meshtastic build outputs only.

variants/esp32s3/elecrow_panel/patch_device_ui.py
danadash123-glitch: Addressed in 361cc30d6. patch_device_ui.py and the self-modifying build approach were removed. The required behavior is implemented as normal, directly compiled C++ and variant configuration.

variants/esp32s3/elecrow_panel/platformio.ini
danadash123-glitch: Addressed in 361cc30d6. The existing elecrow_panel target is preserved. Revised v1.2+ hardware now has the separate elecrow-adv1-43-50-70-tft-v12 target. That target builds successfully with PlatformIO (RAM 19.0%, flash 53.0%).

PR #11534, opened 2026-08-18
Pass framebuffer rotation through DisplayDriverConfig

caveman99:
Follow-up to #11252 and the review of meshtastic/device-ui#355: the framebuffer rotation is handed to device-ui through `DisplayDriverConfig` instead of the `MESHTASTIC_FB_ROTATION` environment variable. `Display.OffsetRotate` is passed unmasked, device-ui maps 0/1/2/3 to the lvgl rotations and warns on unsupported values.

Requires meshtastic/device-ui#355, which adds `DisplayDriverConfig::rotation()` and `FBDriver::create(const DisplayDriverConfig &)`. The rotation stays inactive until the device-ui digest in platformio.ini is bumped past that merge.

`env:native-fb` is the only environment that compiles this path and is not part of the CI matrix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved display setup on Portduino devices by applying screen rotation directly.
  * Preserved support for custom framebuffer dimensions and display-device selection.
  * Display configuration is now more reliable and consistent during startup.
  * Removed reliance on environment-based rotation settings for supported displays.

### DIFF
--- a/src/graphics/tftSetup.cpp
+++ b/src/graphics/tftSetup.cpp
@@ -15,8 +15,6 @@
 #endif
 
 #if defined(ARCH_PORTDUINO) || !defined(HAS_FREE_RTOS)
-#include <cstdio>
-#include <cstdlib>
 #include <thread>
 #endif
 
@@ -350,16 +348,13 @@ void tftSetup(void)
         } else
 #elif defined(USE_FRAMEBUFFER)
         if (portduino_config.displayPanel == fb) {
-            // Rotation from yaml Display.OffsetRotate: 1=90, 2=180, 3=270 deg
-            char rbuf[4];
-            snprintf(rbuf, sizeof(rbuf), "%d", portduino_config.displayRotate ? (portduino_config.displayOffsetRotate & 3) : 0);
-            if (setenv("MESHTASTIC_FB_ROTATION", rbuf, 1) != 0)
-                LOG_ERROR("Failed to set MESHTASTIC_FB_ROTATION, framebuffer will use its default rotation");
             if (portduino_config.displayWidth && portduino_config.displayHeight)
                 displayConfig = DisplayDriverConfig(DisplayDriverConfig::device_t::FB, (uint16_t)portduino_config.displayWidth,
                                                     (uint16_t)portduino_config.displayHeight);
             else
                 displayConfig.device(DisplayDriverConfig::device_t::FB);
+            // Rotation from yaml Display.OffsetRotate: 1=90, 2=180, 3=270 deg
+            displayConfig.rotation(portduino_config.displayRotate ? (uint8_t)portduino_config.displayOffsetRotate : 0);
         } else
 #endif
         {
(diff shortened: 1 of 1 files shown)


src/graphics/tftSetup.cpp
```diff
@@ -350,16 +348,13 @@ void tftSetup(void)
...
-            snprintf(rbuf, sizeof(rbuf), "%d", portduino_config.displayRotate ? (portduino_config.displayOffsetRotate & 3) : 0);
-            if (setenv("MESHTASTIC_FB_ROTATION", rbuf, 1) != 0)
-                LOG_ERROR("Failed to set MESHTASTIC_FB_ROTATION, framebuffer will use its default rotation");
             if (portduino_config.displayWidth && portduino_config.displayHeight)
                 displayConfig = DisplayDriverConfig(DisplayDriverConfig::device_t::FB, (uint16_t)portduino_config.displayWidth,
                                                     (uint16_t)portduino_config.displayHeight);
             else
                 displayConfig.device(DisplayDriverConfig::device_t::FB);
+            // Rotation from yaml Display.OffsetRotate: 1=90, 2=180, 3=270 deg
+            displayConfig.rotation(portduino_config.displayRotate ? (uint8_t)portduino_config.displayOffsetRotate : 0);
```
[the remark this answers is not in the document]
caveman99: Not applicable. The device-ui dependency in platformio.ini is tracked automatically (see the renovate comment on line 139, gitBranch=master), so the digest is bumped once meshtastic/device-ui#355 is merged. Pinning it here to a commit on an unmerged branch would only be reverted by the next dependency update.

env:native-fb is also not part of the CI matrix, so nothing in this repository compiles the changed path before that bump lands.

src/graphics/tftSetup.cpp
```diff
@@ -351,15 +349,13 @@ void tftSetup(void)
...
-                                                    (uint16_t)portduino_config.displayHeight);
-            else
-                displayConfig.device(DisplayDriverConfig::device_t::FB);
+            displayConfig.device(DisplayDriverConfig::device_t::FB)
+                .panel(DisplayDriverConfig::panel_config_t{
+                    .type = panels[portduino_config.displayPanel],
+                    .panel_width = (uint16_t)(portduino_config.displayWidth ? portduino_config.displayWidth : c_default_width),
+                    .panel_height =
+                        (uint16_t)(portduino_config.displayHeight ? portduino_config.displayHeight : c_default_height),
+                    .offset_rotation = (uint8_t)(portduino_config.displayRotate ? portduino_config.displayOffsetRotate : 0)});
```
mverch67: Using `portduino_config.displayRotate` here is not correct. Also, I believe the portduino layer is initializing all portduino args with reasonable defaults, so another null check is not required.

=> `.offset_rotation = (uint8_t)portduino_config.displayOffsetRotate`

mverch67 (changes requested):

src/graphics/tftSetup.cpp
caveman99: Done in b1b9eabb, the FB branch now matches the CUSTOM_TFT one: width, height and offset_rotation are taken straight from portduino_config.

One thing to confirm: displayOffsetRotate defaults to 1 in PortduinoGlue.h, so a framebuffer setup without Display.OffsetRotate in the yaml now starts at 90 degrees, where the env var version only rotated when Display.Rotate was set. Fine by you, or should the FB path treat the default as 0?

mverch67 (approved):

PR #11542, opened 2026-08-18
feat(variants): add Seeed Wio Tracker L1 Pro 1W

caveman99:
nRF52840 + SX1262 with a 1 W external PA, L76K GNSS, SH1106 OLED. Board support developed together with Seeed.

Draft until meshtastic/protobufs#1038 merges and CI regenerates the bindings. `architecture.h` references `meshtastic_HardwareModel_SEEED_WIO_TRACKER_L1_PRO_1W`, which resolves only after that. Generated headers are untouched here.

Contents:

- `boards/` entry and `variants/nrf52840/seeed_wio_tracker_L1_Pro_1W/`, hw_model 144.
- `SX126X_NO_POWER_OPTIMIZATION_TABLE`, an opt-in for boards driving an external PA. Default behaviour is unchanged for existing SX126x boards. `init()` applies the fixed PA config after `begin()`.
- PA gain table indexed by SX1262 output power in dBm, matching `RadioInterface::limitPower()`.
- `SX126X_MAX_POWER` declared explicitly rather than inherited from the driver default.
- RadioLib bumped to 510e00cf for the current LR11x0 and LR2021 fixes, matching the pin used in #11541.

Same board as #11541, which targets master for the 2.7 line.

TX power values still need bench confirmation against measured output. Hardware has not arrived yet.

## Summary by CodeRabbit

- **New Features**
  - Added support for the Seeed Wio Tracker L1 Pro 1W.
  - Enabled the board’s display, LoRa radio, GPS, battery and charging controls, QSPI storage, buzzer, trackball, and peripheral connectivity.
  - Added board-specific pin mappings, LEDs, buttons, serial interfaces, USB identification, firmware upload, and debugging support.

- **Bug Fixes**
  - Improved fixed-configuration radio output-power handling.
  - Added an optimized transmit-power curve for more consistent radio performance.

### DIFF
--- a/boards/seeed_wio_tracker_L1_Pro_1W.json
+++ b/boards/seeed_wio_tracker_L1_Pro_1W.json
@@ -1,0 +1,57 @@
+{
+  "build": {
+    "arduino": {
+      "ldscript": "nrf52840_s140_v7.ld"
+    },
+    "core": "nRF5",
+    "cpu": "cortex-m4",
+    "extra_flags": "-DARDUINO_MDBT50Q_RX -DNRF52840_XXAA",
+    "f_cpu": "64000000L",
+    "hwids": [
+      ["0x2886", "0x1668"],
+      ["0x2886", "0x1667"]
+    ],
+    "usb_product": "TRACKER L1 Pro 1W",
+    "mcu": "nrf52840",
+    "variant": "seeed_wio_tracker_L1_Pro_1W",
+    "bsp": {
+      "name": "adafruit"
+    },
+    "softdevice": {
+      "sd_flags": "-DS140",
+      "sd_name": "s140",
+      "sd_version": "7.3.0",
+      "sd_fwid": "0x0123"
+    },
+    "bootloader": {
+      "settings_addr": "0xFF000"
+    }
+  },
+  "connectivity": ["bluetooth"],
+  "debug": {
+    "jlink_device": "nRF52840_xxAA",
+    "svd_path": "nrf52840.svd",
+    "openocd_target": "nrf52840-mdk-rs"
+  },
+  "frameworks": ["arduino"],
+  "name": "seeed_wio_tracker_L1_Pro_1W",
+  "upload": {
+    "maximum_ram_size": 248832,
+    "maximum_size": 815104,
+    "speed": 115200,
+    "protocol": "nrfutil",
+    "protocols": [
+      "jlink",
+      "nrfjprog",
+      "nrfutil",
+      "stlink",
+      "cmsis-dap",
+      "blackmagic"
+    ],
+    "use_1200bps_touch": true,
+    "require_upload_port": true,
+    "wait_for_upload_port": true
+  },
+  "url": "https://www.seeedstudio.com/Wio-Tracker-L1-Pro-p-6454.html",
+  "vendor": "Seeed Studio"
+}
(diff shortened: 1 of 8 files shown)


caveman99:
someone with the actual hardware and an RF meter needs to validate the PR curve please.

thebentern:
Bench measurement of the TX curve on hardware, covering the "TX power values still need bench confirmation against measured output" item.

### Setup

- Seeed Wio Tracker L1 Pro 1W (USB `0x2886`/`0x1668`), reporting `hw_model` 144 on this branch.
- Firmware built from this PR at `ee67cc3`. nanopb headers regenerated locally against the pinned `aca181b9` — the checked-in `src/mesh/generated/` does not yet carry `SEEED_WIO_TRACKER_L1_PRO_1W`, so the build needs that regen (or CI) to compile `architecture.h`.
- ImmersionRC RF Power Meter v2, fw 1.0.11, 900 MHz calibration curve (there is no 915 MHz point), 30 dB external pad.
- Region US, `is_licensed` false, `txPower` stepped live (2.8.0 applies it without reboot). ~180 B broadcasts for burst airtime; peak reading per step.

### Results

| req dBm | SX1262 drive | table gain | table predicts | measured peak | Δ vs predicted | implied PA gain |
|---:|---:|---:|---:|---:|---:|---:|
| 14 | 4 | 10 | 14 | +13.26 | −0.74 | 9.26 |
| 16 | 6 | 10 | 16 | +15.14 | −0.86 | 9.14 |
| 18 | 7 | 10 | 17 | +16.45 | −0.55 | 9.45 |
| 20 | 9 | 11 | 20 | +18.80 | −1.20 | 9.80 |
| 22 | 11 | 11 | 22 | +20.82 | −1.18 | 9.82 |
| 24 | 13 | 11 | 24 | +22.95 | −1.05 | 9.95 |
| 26 | 15 | 11 | 26 | +24.76 | −1.24 | 9.76 |
| 27 | 16 | 11 | 27 | +25.76 | −1.24 | 9.76 |
| 28 | 17 | 11 | 28 | +26.94 | −1.06 | 9.94 |
| 29 | 18 | 11 | 29 | +27.73 | −1.27 | 9.73 |
| 30 | 20 | 10 | 30 | +29.28 | −0.72 | 9.28 |

### Findings

1. **Tracking is 1:1.** Least-squares slope 0.9954 over requested 14→30 dBm (16 dB requested, 16.02 dB delivered). No compression at 1 W — slope across the top four points is 1.14, so the PA still has headroom at request 30.
2. **The table runs ~1 dB optimistic.** Mean Δ −1.10 dB (sd 0.24). Implied PA gain is 9.1–10.0 dB (mean 9.6) against the table's 10/11, with the same rise-in-the-middle, dip-at-the-top shape.
3. Subtracting 1 from every `TX_GAIN_LORA` entry (`10`→`9`, `11`→`10`) lands every step within ~0.5 dB.

**Caveat on the absolute figure.** This meter's no-signal floor reads +5.12 dBm on the 900 MHz curve where the instrument spec is ~−25 dBm, so absolute dBm here rests on inferring a +30.1 dB instrument offset and assuming the pad is exactly 30.0 dB. Combined uncertainty is ~±1 dB, which is the same size as the offset being measured. The 1:1 tracking result is relative and unaffected by this. Suggest treating the table's shape and magnitude as confirmed, and holding the −1 dB adjustment until it can be repeated against a calibrated reference.

### Aside

The 2.7.26 build this unit shipped with (same `pioEnv`, reporting `hw_model` 139) measured **+25.5 dBm for a requested 14 dBm** — 11.5 dB hot, consistent with the PA gain table not being applied, and reaching ~32 dBm at the top of the range, above the US 30 dBm limit. This PR corrects that: the same request now measures +13.3 dBm.

thebentern (approved):

PR #11549, opened 2026-08-19
ci: build newly added variants in the PR matrix

caveman99:
A new board declares `board_level = release`, so it is excluded from the `--level pr` subset and gets no CI build until after merge. The first build then happens on a branch where a failure is harder to attribute.

This builds the first env of each `platformio.ini` **added** by a PR, regardless of `board_level`.

- Only added files qualify. Adding an env to an existing config does not trigger it.
- If an added config declares several envs, only the first is built.
- `generate_ci_matrix.py` gains `--added-config PATH` (repeatable). Without it, output is byte-identical to before.
- The setup job derives the paths from `git diff --diff-filter=A` against the merge base, and needs `fetch-depth: 0`.
- Full-matrix runs on push, schedule and workflow_dispatch are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pull request validation now includes the first environment from newly added variant configurations.
  * New variants are tested automatically, even when standard filtering would otherwise exclude them.
  * Merge-group validation now uses accurate base comparisons.
  * Unreadable configurations and files without a valid environment are handled gracefully.
  * CI setup now correctly reports configuration-discovery failures instead of silently continuing with incomplete results.

### DIFF
--- a/.github/workflows/main_matrix.yml
+++ b/.github/workflows/main_matrix.yml
@@ -55,6 +55,9 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7
+        with:
+          # Needed to diff against the base branch for newly added variants.
+          fetch-depth: 0
       - uses: actions/setup-python@v6
         with:
           python-version: 3.x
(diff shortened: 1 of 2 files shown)


caveman99:
@vidplace7 did you have a look yet?

Issue #4206, opened 2024-06-30
[Bug]: USB Serial does not recover when waking from light sleep on ESP32-S3 devices using native ESP32 serial over usb


Cisien:
### Category

Serial

### Hardware

T-Deck, Station G2, Other

### Firmware Version

2.3.11+

### Description

When the esp32 enters light sleep mode the usb serial connection does not always recover. This often leads to the USB connection getting stuck in a bad state until the device is rebooted.

Validated the behavior by setting the `device.role` to `ROUTER` and `display.screen_on_secs` to `15`
Can reproduce on Station G2 and T-Deck
Reproduces on 2.3.11 and 2.3.14

Changing the devices role from ROUTER to CLIENT mitigates the issue by not attempting to enter light sleep

### Relevant log output

```
DEBUG | ??:??:?? 55 [RadioIf] AirTime - Packet transmitted : 641ms
INFO  | ??:??:?? 55 [PowerFSM] setup LORA_DIO1 (GPIO45) with wakeup by gpio interrupt
NFO  | ??:??:?? 56 [PowerFSM] Exit light sleep gpio: btn=0
INFO  | ??:??:?? 56 [PowerFSM] Exit state: LS
```
(not ehte missing `I` in line 3 above)

A few moments later, serial dropped long enough to timeout the meshtastic cli (`--debug --noproto`)
```
DEBUG | 01:37:32 85 [Power] Battery: usbPower=1, isCharging=1, batMv=4700, batPct=100
INFO  | 01:37:38 92 [PowerFSM] lsEnter begin, ls_secs=86400
WARNING file:stream_interface.py __reader line:187 Meshtastic serial port disconnected, disconnecting... device reports readiness to read but returned no data (device disconnected or multiple access on port?)
DEBUG file:stream_interface.py __reader line:202 reader is exiting
DEBUG file:stream_interface.py _disconnected line:80 Closing our port
```
dmesg:
```
[441723.589803] usb 1-5.2: USB disconnect, device number 97
[441753.764416] usb 1-5.2: new full-speed USB device number 98 using xhci_hcd
[441753.864861] usb 1-5.2: device descriptor read/all, error -32
[441753.944412] usb 1-5.2: new full-speed USB device number 99 using xhci_hcd
[441754.028430] usb 1-5.2: device descriptor read/64, error -32
[441754.216428] usb 1-5.2: device descriptor read/64, error -32
[441754.324839] usb 1-5-port2: attempt power cycle
[...]

PR #9122, opened 2025-12-30
ESP32: Migrate to Arduino 3.x (pioarduino)

vidplace7:
Switch all ESP32 builds to [pioarduino/platform-espressif32](https://github.com/pioarduino/platform-espressif32).

https://docs.espressif.com/projects/arduino-esp32/en/latest/migration_guides/2.x_to_3.0.html

Switching to Arduino-ESP32 3.x will be necessary in order to properly support newer Espressif MCUs (esp32h2, esp32c5, esp32c6, esp32p4, etc)

**DRAFT** for now, see below.

TODO:
- [x] Test ESP32s3
  - [x] `heltec-v3`
  - [x] `t-deck-tft`
- [x] Test ESP32 (Original)
  - [x] `tbeam`
  - [x] `tlora-v2-1-1_6`
  - [x] `station-g1`
- [x] Test ESP32c3
  - [x] `heltec-ht62-esp32c3-sx1262`
- [x] Test ESP32c6
  - [x] `tlora-c6`
  - [x] `m5stack-unitc6l`
- [x] Test ESP32p4
  - @mverch67 added [preliminary support](https://github.com/meshtastic/firmware/pull/9526) for this platform as part of this PR.
    - [x] `crowpanel p4 advanced 5"`
    - [x] `crowpanel p4 advanced 9"`
    - [x] `crowpanel p4 advanced 10.1"`

Tasks:
- [x] NimBLE Transition
  - With ESP-IDF 5.5.x NimBLE has been integrated, Arduino-ESP32 exposes this nimble via the BLE API
  - Existing NimBLE code should be transitioned to the (similar but not identical) BLE API.
  - https://github.com/espressif/arduino-esp32/tree/3.3.5/libraries/BLE
  - https://docs.espressif.com/projects/arduino-esp32/en/latest/api/ble.html
  - Huge thanks to @cpatulea for getting this working correctly.
- [x] ADC API Changes
  - Calibration changes, channel reference changes, etc
- [x] CI Work
  - https://github.com/pioarduino/platform-espressif32/blob/main/.github/workflows/examples.yml
- [x] (original) ESP32 4MB flash overruns
  - Excluded `AUDIO` Module
- [x] ESP32c6 4MB flash overruns
  - Excluded SCREEN / AUDIO

Depends on:
- Remove `syslog` lib  (replace?) 
  - #8964 
- libpax
[...]

PR #11556, opened 2026-08-20
fix(power): keep light sleep off on native USB-CDC boards running on USB power

DrDoof:
Issue #4206 described USB serial dying on ESP32-S3 boards after light sleep, with the reporter noting that switching the role from ROUTER to CLIENT worked around it. That issue was closed in October 2024, but the behaviour is still present on current firmware. I ran into it on a T-Beam S3 Core running 2.7.26 and went looking for the cause.

## What happens

Set `device.role` to ROUTER on an S3 board whose console runs on the native USB-OTG peripheral, leave it plugged into a powered hub, and the serial port disappears after roughly 30 seconds. The host then loops:

```
usb 3-1.4: device descriptor read/64, error -32
usb 3-1.4: device not accepting address 100, error -71
usb 3-1-port2: unable to enumerate USB device
```

Same signature the original reporter posted. The device itself is fine, the display keeps working, only USB is gone until the cable is physically replugged. On a desk router that means losing the serial console permanently.

## Why

Three things line up in `PowerFSM.cpp`:

1. `isPowered()` computes `isPowerSavingMode = config.power.is_power_saving || isRouter`, so role ROUTER forces power saving **regardless of whether the board is running on USB power**. `hasPower` comes out false and boot lands in `stateON` instead of `statePOWER`.
2. The guard `if ((isRouter || config.power.is_power_saving) && !isWifiAvailable() && !isTrackerOrSensor)` then adds the timed transitions into `stateLS`.
3. `lsIdle()` calls `doLightSleep()` every `SLEEP_TIME` seconds.

On ESP32-S3 with `ARDUINO_USB_MODE=0` the CDC console lives on the native USB-OTG peripheral, and light sleep powers down the USB PHY. The host cannot re-enumerate it, hence the `error -32` loop. This does not affect boards using an external UART bridge, which is why it only shows up on native USB-CDC hardware.

The comment in `isPowered()` explains the intent: routers are assumed to run all the time from a low current source such as solar. That assumption is reasonable, it just does not hold when the board is demonstrably sitting on USB power.

## The change

Skip the light sleep transitions when the board uses native USB-CDC **and** `powerStatus->getHasUSB()` reports it is actually running on USB power. There is no energy to save in that case. The check is wrapped in `#if defined(ARDUINO_USB_MODE) && (ARDUINO_USB_MODE == 0) && defined(ARDUINO_USB_CDC_ON_BOOT) && (ARDUINO_USB_CDC_ON_BOOT == 1)`, so it compiles out entirely on every other target. Battery and solar behaviour is unchanged, which is the case the original logic was written for.

15 lines, one file, 96 bytes of flash.

## Testing

LilyGO T-Beam S3 Core, firmware built from this branch at 2.7.26.

Before: role ROUTER, USB dies within about 30 seconds, `error -32` in dmesg, `meshtastic --info` times out.

After: role ROUTER confirmed via `--info`, then left running for a 20 minute soak sampling the USB device and the kernel error counter every 30 seconds. 39 of 39 samples showed the device present and zero USB errors. `rebootCount` stayed at the number of deliberate restarts, so nothing crashed and recovered quietly.

```
16:41:34 usb=OK errors=0
...
17:00:37 usb=OK errors=0
```

I have only S3 hardware here, so I could not verify the Station G2 and T-Deck cases from the original report, though both are native USB-CDC boards and should be covered by the same guard.

Refs: #4206

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USB serial reliability when native USB is connected after boot.
  * Devices now exit light sleep when USB activity is detected, preserving console access.
  * Light sleep is skipped while native USB serial is attached to prevent the console from becoming unavailable.

### DIFF
--- a/src/PowerFSM.cpp
+++ b/src/PowerFSM.cpp
@@ -58,6 +58,20 @@ static bool isPowered()
     return !isPowerSavingMode && powerStatus && (!powerStatus->getHasBattery() || powerStatus->getHasUSB());
 }
 
+/// True when the CDC console runs on the native USB-OTG peripheral (ARDUINO_USB_MODE=0) and the
+/// board is currently running on USB power. Light sleep powers that PHY down, after which the host
+/// can no longer re-enumerate the device and the console is gone until the cable is replugged.
+/// See https://github.com/meshtastic/firmware/issues/4206
+static bool nativeUsbSerialActive()
+{
+#if defined(ARDUINO_USB_MODE) && (ARDUINO_USB_MODE == 0) && defined(ARDUINO_USB_CDC_ON_BOOT) &&                     \
+    (ARDUINO_USB_CDC_ON_BOOT == 1)
+    return powerStatus && powerStatus->getHasUSB();
+#else
+    return false;
+#endif
+}
+
 static bool isBluetoothEnabledForPowerFSM()
 {
 #if HAS_BLUETOOTH && !MESHTASTIC_EXCLUDE_BLUETOOTH
@@ -144,6 +158,15 @@ static void lsIdle()
     // LOG_INFO("lsIdle begin ls_secs=%u", getPref_ls_secs());
 
 #ifdef ARCH_ESP32
+
+    // USB may have been plugged in after boot, so re-check here instead of relying on the snapshot
+    // taken in PowerFSM_setup(). Leaving light sleep costs a wake cycle; losing the console costs a
+    // physical replug, and on boards without a PMU there is no VBUS insert IRQ to recover us.
+    if (nativeUsbSerialActive()) {
+        LOG_INFO("Native USB-CDC on USB power: leaving light sleep (keeps serial alive)");
+        powerFSM.trigger(EVENT_WAKE_TIMER);
+        return;
+    }
 
     // Do we have more sleeping to do?
     if (secsSlept < config.power.ls_secs) {
@@ -439,18 +462,12 @@ void PowerFSM_setup()
                              config.device.role == meshtastic_Config_DeviceConfig_Role_TAK_TRACKER ||
                              config.device.role == meshtastic_Config_DeviceConfig_Role_SENSOR;
 
-    // On ESP32-S3 boards whose CDC console runs on the native USB-OTG peripheral
-    // (ARDUINO_USB_MODE=0), light sleep powers down the USB PHY and the host can no longer
-    // re-enumerate the device: dmesg loops "device descriptor read/64, error -32" until the
-    // cable is physically replugged. See https://github.com/meshtastic/firmware/issues/4206
-    // While actually running on USB power there is no energy to save, so skip light sleep.
-    bool nativeUsbSerialAttached = false;
-#if defined(ARDUINO_USB_MODE) && (ARDUINO_USB_MODE == 0) && defined(ARDUINO_USB_CDC_ON_BOOT) &&                     \
-    (ARDUINO_USB_CDC_ON_BOOT == 1)
-    nativeUsbSerialAttached = powerStatus && powerStatus->getHasUSB();
+    // Already on USB power at boot: skip registering the light-sleep transitions entirely, so a
+    // mains powered node never churns through wake cycles. Cables plugged in later are caught live
+    // by lsIdle(). See nativeUsbSerialActive().
+    bool nativeUsbSerialAttached = nativeUsbSerialActive();
     if (nativeUsbSerialAttached)
         LOG_INFO("Native USB-CDC on USB power: light sleep disabled (keeps serial alive)");
-#endif
 
     if ((isRouter || config.power.is_power_saving) && !isWifiAvailable() && !isTrackerOrSensor &&
         !nativeUsbSerialAttached) {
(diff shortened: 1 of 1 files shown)


DrDoof:
Thanks, the boot-order point is fair and I want to split it into the part I think is wrong and the part I think is right.

**The stated reasoning does not hold for the board I tested.** The comment says both native-USB S3 targets lack `BATTERY_PIN` and PMU detection so `getHasUSB()` stays false. `tbeam-s3-core` defines `HAS_AXP2101` in its `variant.h`, so it has a PMU and `getHasUSB()` reports real VBUS state. I also checked the built binary rather than trusting the source read: the string from the new `LOG_INFO` is present in the firmware image that ran during the soak.

```
$ strings firmware-tbeam-s3-core-2.7.26.7a7d665.bin | grep "Native USB-CDC"
Native USB-CDC on USB power: light sleep disabled (keeps serial alive)
```

That only compiles under `ARDUINO_USB_MODE=0 && ARDUINO_USB_CDC_ON_BOOT=1`, and it only prints when `getHasUSB()` returned true, so on this target both the guard and the USB detection work.

**The underlying concern is still valid, just for a different reason.** `nativeUsbSerialAttached` is read once in `PowerFSM_setup()`, so it is a boot-time snapshot. Boot on battery, plug USB later, and the transitions were already registered. What saves it in practice on a PMU board is that `Power::runOnce()` fires `EVENT_POWER_CONNECTED` on `isVbusInsertIrq()`, and there is already a `stateLS -> statePOWER` transition on that event, so the node leaves light sleep when the cable goes in. On a target without a PMU that path does not exist, and there the snapshot really can leave a node sleeping with USB attached.

So the gap is narrower than described but real. I would rather fix it properly than paper over it, and I see two options:

1. Keep the registration as is and add the check to the light-sleep entry path, so the state is evaluated live rather than at boot.
2. Leave this PR as the minimal fix for the reported regression and handle the hotplug case separately.

I lean towards the first, but it touches `lsIdle()` and I would rather not expand the diff without a maintainer nod. Happy to push the change if that is preferred.

One thing I cannot verify myself: I only have the T-Beam S3 Core here, so the Station G2 and T-Deck cases from #4206 are untested by me.

DrDoof:
Pushed the live check in 6568e0f.

The condition now lives in a `nativeUsbSerialActive()` helper called from two places:

- `lsIdle()` re-evaluates it before every sleep and leaves light sleep through the existing `EVENT_WAKE_TIMER` path when USB power is present. That closes the hotplug gap, including on targets without a PMU where no VBUS insert IRQ exists to recover the node.
- `PowerFSM_setup()` keeps its check, so a node that boots already on USB skips registering the transitions instead of entering light sleep and immediately waking again.

The helper returns false outside `ARDUINO_USB_MODE=0 && ARDUINO_USB_CDC_ON_BOOT=1`, so nothing changes on other targets.

Built for `tbeam-s3-core` and checked both strings landed in the image:

```
Native USB-CDC on USB power: leaving light sleep (keeps serial alive)
Native USB-CDC on USB power: light sleep disabled (keeps serial alive)
```

Flash grew by 128 bytes over the previous revision.

Worth flagging: the hotplug path is reasoned from the code, not reproduced on hardware. My node has an AXP2101 and boots on USB, so it exercises the setup-time branch, and I have no PMU-less native-USB board to test the case this change is actually aimed at. The soak result from the original report still stands for the reported scenario.

mverch67:
This won't work for devices which don't support Li-Ion battery but have only USB connector to connect a battery like elecrow panels or sensecap indicator.
The implementation should be more flexible to account this.

DrDoof:
Fair point, that case would break and the guard needed to be narrower.

`getHasUSB()` means two different things depending on the hardware. With a PMU it comes from VBUS measured by the chip, so it genuinely says external power is present. Without one it can just as easily be a battery wired to the USB connector, and there skipping light sleep is exactly the wrong call.

So the helper is now gated on `HAS_PMU`, which `architecture.h` defines only for AXP192 or AXP2101. Neither `elecrow_panel` nor `seeed-sensecap-indicator` declares either of those, so on those boards the whole thing compiles down to `return false` and their power saving is untouched.

Pushed as a11ef45, +4/-2. On tbeam-s3-core the guard leaves the code active: the image grows 220 bytes against a build of the same tree without the patch, which is the cheap way to confirm the condition is actually true there. I don't have either of the boards you mentioned, so that side rests on reading their variant headers rather than on a measurement.

DrDoof:
Correction to my earlier numbers, and a question about whether this is still needed on develop.

The 220 byte figure came from a v2.7.26 tree. On develop the helper compiles to nothing on tbeam-s3-core: `boards/tbeam-s3-core.json` moved to `-DARDUINO_USB_MODE=1` in #9122, and the guard requires `ARDUINO_USB_MODE == 0`. I built this branch as pushed and again with the `HAS_PMU` check replaced by `0`. Both images come out at 2370992 bytes, with PowerFSM.cpp recompiled.

I also could not find any develop target that has `ARDUINO_USB_MODE=0`, `ARDUINO_USB_CDC_ON_BOOT=1` and a PMU at the same time, so as it stands the change does nothing on develop.

What I can't answer from the code: does light sleep also lose the console on the hardware USB-Serial-JTAG (mode 1)? If yes, the guard should cover mode 1 as well. If no, #9122 already fixed this for develop and the PR only matters for 2.7.x, where tbeam-s3-core still uses mode 0. I can test mode 1 on a T-Beam S3 Core if that helps.

PR #11559, opened 2026-08-21
breaking-fix(radio): Align Center Frequency in Japan to comply with ARIB STD-T108

t-miura:
# Summary
This PR fixes misaligned center frequency with all bandwidth when region is set to Japan(JP).

# Backgrounds

Under Japanese Radio Law and standard ARIB STD-T108 ("920MHz-Band Telemeter, Telecontrol and Data Transmission Radio Equipment"):

The frequency band for specified low power radio stations is divided into 200 kHz unit channels (channels 24 through 38 spanning 920.5 MHz – 923.5 MHz).
For signals occupying more than one unit channel (e.g. 250 kHz LoRa bandwidth), two consecutive unit channels must be bonded (total allocated width = 400 kHz).
The regulatory center frequency for two bonded channels (e.g., channels 24 & 25) is 920.700 MHz (Table 3-2 and Table 3-12 of ARIB STD-T108).

refer https://www.arib.or.jp/english/html/overview/doc/5-STD-T108v1_5-E1.pdf for tables and other regulations.

# Current behavior, and why it is problematic
In `src/mesh/RadioInterface.cpp`, region JP used PROFILE_STD with padding = 0 and spacing = 0.
The old formula calculated the frequency like:
```
center_freq = 920.500MHz + 250kHz/2000kHz = 920.625MHz
```
Which is 75kHz below than the standard requires(920.7MHz),
resulting no lower guard band, and all subsequent channels drifts 200kHz(this varies on bandwidth of preset),
any frequency slot selection causing "no guard band" situation regardless of bandwidth.

# Proposing changes
To address this issue while keep any other region unaffected, also any bandwidth shall not use invalid center frequency,
I would like to introduce dynamic unit-channel padding and spacing, as follows:
- `src/mesh/MeshRadio.h` : following two will be added
  - `RegionInfo::getSpacing(float bwKHz)`
    - When region is not JP, it returns profile-specific frequency spacing(thus no effects if region is not JP)
    - If region is set to JP, it calculates the required spacing frequency following the ARIB STD-T108
  - `RegionInfo::getPadding(float bwKHz)`
    - same behavior as `getSpacing`
- `src/mesh/MeshRadio.cpp`: following be added and changed:
  - Update existing lines to use two functions introducing in `src/mesh/MeshRadio.h` 
  - Use `floor()` instad of `round()` so that frequencies do not exceed the upper band limit
    - used `+0.001f` for potential floating-foint imprecision issue
    - *Note: this will affects regardless of region, but result will be exactly same regardless of any config*
- `src/graphics/draw/MenuHandler.cpp`
  - Update existing lines to use two functions introducing in `src/mesh/MeshRadio.h` 
- (not in this PR and WIP on my side) `source/graphics/common/LoRaPresets.cpp` in device-ui
  - requires updating the regionInfo for Japan
    - actually it was older definition, just a display issue though
  - Update `getNumChannels` and `getRadioFreq` to use similar logic to calculate and display the correct value
    - this is also the display issue, but worth fixing it as currently there is no support for any spacing/padding
- `test/test_admin_radio/test_main.cpp`
  - Added test cases covering all bandwidth profiles from 62.5kHz(NARROW_SLOW) to 500kHz (SHORT_TURBO)

# Caveats
## Changes on Frequency Slots
This will changes not only center frequency, but also changes Frequency Slot, both mapping and total available slots with BW > 200kHz as follows:
- 62.5kHz : 15 Frequency Slots as-is, but on different center frequency
- 125kHz : 15 Frequency Slots as-is, but on different center frequency too
- 250kHz: *7 Frequency Slots(currently 12)*, on different center frequency
- 500kHz: *5 Frequency Slots(currently 6)*,  on different center frequency too

## Needs of Frequency Migration
Since this will changes whole frequency assignment when the node's region is set to JP,
node users are required do following upon their situations:
- If they updated the firmware
  - and Frequency Slot has been changed from 0(like 12 as we usually use in here) upon previous setup
    -  Requires changing the Frequency Slot to new one(ex: on LongFast, new default slot(0) will be Slot 6/922.7MHz)
  - but Frequency Slot has *not* been changed from 0
   - The node will run at new default frequency slot, no action required to communicate same config's node
- If they updated the firmware, but still need to communicate with nodes with older frequency/firmware
  - They have to override the frequency by hand(ex. LongFast/Slot 12=923.375MHz)
- If they haven't updated the firmware, but need to communicate with newer nodes that uses this frequency plan
  - They have to override the frequency by hand as well(ex. LongFast/Slot 6=922.7MHz)

# Validations
Confirmed that the center frequencies are now in line with all multiple unit-channel bondings written in ARIB STD-T108:

| Preset / Bandwidth | Bonded Unit Channels ($n$) | Slot Width | Symmetric Padding | Center Frequency $f_c(0)$ | ARIB STD-T108 Match | Slots in Band |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| **`LongFast`** (250 kHz) | 2 (400 kHz) | 0.400 MHz | 75 kHz | **920.700 MHz** | **100% Exact** (Table 3-2) | 7 slots |
| **`LongSlow`** (125 kHz) | 1 (200 kHz) | 0.200 MHz | 37.5 kHz | **920.600 MHz** | **100% Exact** (Table 3-1) | 15 slots |
| **`NarrowSlow`** (62.5 kHz) | 1 (200 kHz) | 0.200 MHz | 68.75 kHz | **920.600 MHz** | **100% Exact** (Table 3-1) | 15 slots |
| **`ShortTurbo`** (500 kHz) | 3 (600 kHz) | 0.600 MHz | 50 kHz | **920.800 MHz** | **100% Exact** (Table 3-3) | 5 slots |

note: LongSlow has another issue(too long transmission time for most of payload), only used for 125kHz BW example.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - [x] Heltec Mesh Node T114 / nRF52840
    - [x] Seeed Studio Wio-E5 / STM32WL
    - [x] WIZnet W5500-EVB-Pico2 & SX1262 / RP2350

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved frequency-slot calculations across regional bandwidth profiles.
  - Corrected channel counts, spacing, padding, and frequency placement for Japanese region configurations.
  - Added precision handling to prevent incorrect rounding at bandwidth boundaries.

- **Tests**
  - Added coverage for four Japanese channel configurations, including channel counts, padding, and calculated frequencies.

### DIFF
--- a/src/graphics/draw/MenuHandler.cpp
+++ b/src/graphics/draw/MenuHandler.cpp
@@ -501,7 +501,7 @@ void menuHandler::FrequencySlotPicker()
 
     meshtastic_Config_LoRaConfig &loraConfig = config.lora;
     double bw = loraConfig.use_preset ? modemPresetToBwKHz(loraConfig.modem_preset, myRegion->wideLora)
-                                      : bwCodeToKHz(loraConfig.bandwidth);
+                                      : clampBandwidthKHz(bwCodeToKHz(loraConfig.bandwidth));
 
     uint32_t numChannels = 0;
     if (myRegion) {
--- a/src/graphics/draw/MenuHandler.cpp
+++ b/src/graphics/draw/MenuHandler.cpp
@@ -505,14 +505,14 @@ void menuHandler::FrequencySlotPicker()
 
     uint32_t numChannels = 0;
     if (myRegion) {
-        // Match RadioInterface::applyModemConfig(): include padding, add spacing in numerator, and use round()
-        const double spacing = myRegion->profile->spacing;
-        const double padding = myRegion->profile->padding;
+        // Match RadioInterface::applyModemConfig(): include padding, add spacing in numerator, and use floor()
+        const double spacing = myRegion->getSpacing(bw);
+        const double padding = myRegion->getPadding(bw);
         const double channelBandwidthMHz = bw / 1000.0;
-        const double numerator = (myRegion->freqEnd - myRegion->freqStart) + spacing;
+        const double numerator = (myRegion->freqEnd - myRegion->freqStart) + spacing + 0.001;
         const double denominator = spacing + (padding * 2) + channelBandwidthMHz;
         if (denominator > 0.0) {
-            numChannels = static_cast<uint32_t>(round(numerator / denominator));
+            numChannels = static_cast<uint32_t>(floor(numerator / denominator));
         } else {
             LOG_WARN("Invalid region config: non-positive channel spacing/width");
         }
--- a/src/mesh/MeshRadio.h
+++ b/src/mesh/MeshRadio.h
@@ -5,6 +5,7 @@
 #include "PointerQueue.h"
 #include "configuration.h"
 #include "detect/LoRaRadioType.h"
+#include <math.h>
 
 // Sentinel marking the end of a modem preset array. Declared `const` rather
 // than `constexpr` because the cast from 0xFF to the enum is out-of-range and
@@ -92,6 +93,24 @@ struct RegionInfo {
         while (profile->presets[n] != MODEM_PRESET_END)
             n++;
         return n;
+    }
+    float getSpacing(float bwKHz = 0) const
+    {
+        if (code == meshtastic_Config_LoRaConfig_RegionCode_JP)
+            return 0.0f;
+        return profile ? profile->spacing : 0.0f;
+    }
+    float getPadding(float bwKHz = 0) const
+    {
+        if (code == meshtastic_Config_LoRaConfig_RegionCode_JP) {
+            // ARIB STD-T108: 200 kHz unit channel grid. Center LoRa BW symmetrically within ceil(bw/200kHz) unit channels.
+            const float unitChannelMHz = 0.200f;
+            const float bwMHz = (bwKHz > 0 ? bwKHz : 250.0f) / 1000.0f;
+            const int numUnits = (int)ceil(bwMHz / unitChannelMHz);
+            const float slotWidth = numUnits * unitChannelMHz;
+            return (slotWidth - bwMHz) / 2.0f;
+        }
+        return profile ? profile->padding : 0.0f;
     }
 };
 
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -1183,8 +1183,10 @@ bool RadioInterface::checkOrClampConfigLora(meshtastic_Config_LoRaConfig &loraCo
     // Calculate width of slots (aka channels) based on bandwidth and any spacing or padding required by the region:
     // spacing = gap between slots (0 for continuous spectrum) and at the beginning of the band
     // padding = gap at the beginning and end of the slots (0 for no padding)
-    float freqSlotWidth = newRegion->profile->spacing + (newRegion->profile->padding * 2) + (check_bw / 1000); // in MHz
-    uint32_t numFreqSlots = round((newRegion->freqEnd - newRegion->freqStart + newRegion->profile->spacing) / freqSlotWidth);
+    float spacing = newRegion->getSpacing(check_bw);
+    float padding = newRegion->getPadding(check_bw);
+    float freqSlotWidth = spacing + (padding * 2) + (check_bw / 1000.0f); // in MHz
+    uint32_t numFreqSlots = floor((newRegion->freqEnd - newRegion->freqStart + spacing + 0.001f) / freqSlotWidth);
 
     // Check if the region supports the requested bandwidth
     if ((newRegion->freqEnd - newRegion->freqStart) < freqSlotWidth) {
@@ -1199,8 +1201,10 @@ bool RadioInterface::checkOrClampConfigLora(meshtastic_Config_LoRaConfig &loraCo
             check_bw = bwCodeToKHz(loraConfig.bandwidth);
 
             // Recompute slot width and number of slots based on the new bandwidth
-            freqSlotWidth = newRegion->profile->spacing + (newRegion->profile->padding * 2) + (check_bw / 1000); // in MHz
-            numFreqSlots = round((newRegion->freqEnd - newRegion->freqStart + newRegion->profile->spacing) / freqSlotWidth);
+            spacing = newRegion->getSpacing(check_bw);
+            padding = newRegion->getPadding(check_bw);
+            freqSlotWidth = spacing + (padding * 2) + (check_bw / 1000.0f); // in MHz
+            numFreqSlots = floor((newRegion->freqEnd - newRegion->freqStart + spacing + 0.001f) / freqSlotWidth);
         } else {
             return false;
         }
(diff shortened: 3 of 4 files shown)


t-miura:
@coderabbitai review

src/graphics/draw/MenuHandler.cpp
```diff
@@ -467,14 +467,14 @@ void menuHandler::FrequencySlotPicker()
...
+        // Match RadioInterface::applyModemConfig(): include padding, add spacing in numerator, and use floor()
+        const double spacing = myRegion->getSpacing(bw);
+        const double padding = myRegion->getPadding(bw);
         const double channelBandwidthMHz = bw / 1000.0;
-        const double numerator = (myRegion->freqEnd - myRegion->freqStart) + spacing;
+        const double numerator = (myRegion->freqEnd - myRegion->freqStart) + spacing + 0.001;
         const double denominator = spacing + (padding * 2) + channelBandwidthMHz;
         if (denominator > 0.0) {
-            numChannels = static_cast<uint32_t>(round(numerator / denominator));
+            numChannels = static_cast<uint32_t>(floor(numerator / denominator));
```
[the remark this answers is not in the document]
t-miura: @coderabbitai review

caveman99:
we have a similar situation on EU_866 and the definition can be re-used here to a certain extent. No matter what we do, this is a breaking change, and needs to go into 3.0, same as EU_433 is currently broken.

t-miura:
> we have a similar situation on EU_866 and the definition can be re-used here to a certain extent. No matter what we do, this is a breaking change, and needs to go into 3.0, same as EU_433 is currently broken.

thanks for the quick reply!

yes, even there's less than 100 active nodes overall this country, breaking change is breaking change.
so i'm agree that this needs to go into 3.0, meanwhile we can assume we're using one extra unit channels,
which is a waste of frequency, but believe still legit.

src/mesh/MeshRadio.h
```diff
@@ -93,6 +94,24 @@ struct RegionInfo {
             n++;
         return n;
     }
+    float getSpacing(float bwKHz = 0) const
+    {
+        if (code == meshtastic_Config_LoRaConfig_RegionCode_JP)
+            return 0.0f;
+        return profile ? profile->spacing : 0.0f;
+    }
+    float getPadding(float bwKHz = 0) const
```
caveman99, line 103: That should be part of a region specific file in the new directory i mentioned in #11747. Even better: i'd move "float unitChannelMHz;" as a new field into the regionProfile, then getPadding() computes (ceil(bw/unit)*unit - bw)/2 with no region test at all.

It also resolves a coupling that's currently invisible: as written, getPadding() ignores profile->padding entirely for JP, so the padding value in 
#11566 PROFILE_JP is ignored.

src/mesh/MeshRadio.h
```diff
@@ -93,6 +94,24 @@ struct RegionInfo {
             n++;
         return n;
     }
+    float getSpacing(float bwKHz = 0) const
```
caveman99, line 97: getSpacing never uses its parameter and returns 0 for JP, which is what PROFILE_STD already gives. The function is a no-op in every reachable path right now.

src/mesh/MeshRadio.h
```diff
@@ -93,6 +94,24 @@ struct RegionInfo {
...
+    {
+        if (code == meshtastic_Config_LoRaConfig_RegionCode_JP)
+            return 0.0f;
+        return profile ? profile->spacing : 0.0f;
+    }
+    float getPadding(float bwKHz = 0) const
+    {
```
caveman99, line 104: getPadding's bwKHz = 0 default silently substitutes 250 kHz. All four call sites pass a real value, so the default only exists to hide a future missed argument behind a plausible-but-wrong number.

src/mesh/MeshRadio.h
caveman99, line 1: Overall comment: don't put region gates in global code. See the spots for alternate suggestions

src/mesh/MeshRadio.h
```diff
@@ -93,6 +94,24 @@ struct RegionInfo {
...
+        if (code == meshtastic_Config_LoRaConfig_RegionCode_JP) {
+            // ARIB STD-T108: 200 kHz unit channel grid. Center LoRa BW symmetrically within ceil(bw/200kHz) unit channels.
+            const float unitChannelMHz = 0.200f;
+            const float bwMHz = (bwKHz > 0 ? bwKHz : 250.0f) / 1000.0f;
+            const int numUnits = (int)ceil(bwMHz / unitChannelMHz);
```
caveman99, line 109: ceil() in a PR that adds +0.001f elsewhere for exactly that reason. numUnits = (bwKHz + 199.9f) / 200.0f in integer kHz sidesteps it, and you can los the math.h include

src/graphics/draw/MenuHandler.cpp
```diff
@@ -501,18 +501,18 @@ void menuHandler::FrequencySlotPicker()
...
     if (myRegion) {
-        // Match RadioInterface::applyModemConfig(): include padding, add spacing in numerator, and use round()
-        const double spacing = myRegion->profile->spacing;
-        const double padding = myRegion->profile->padding;
+        // Match RadioInterface::applyModemConfig(): include padding, add spacing in numerator, and use floor()
+        const double spacing = myRegion->getSpacing(bw);
+        const double padding = myRegion->getPadding(bw);
         const double channelBandwidthMHz = bw / 1000.0;
-        const double numerator = (myRegion->freqEnd - myRegion->freqStart) + spacing;
+        const double numerator = (myRegion->freqEnd - myRegion->freqStart) + spacing + 0.001;
```
caveman99, line 512: +0.001f is 1 kHz which is 12% of a 7.8 kHz slot, so an absolute epsilon is dimensionally wrong even though it changes no result today. floor(x * (1 + 1e-6f)) scales properly. That goes for all occurances this is used.

src/mesh/RadioInterface.cpp
```diff
@@ -1183,8 +1183,10 @@ bool RadioInterface::checkOrClampConfigLora(meshtastic_Config_LoRaConfig &loraCo
     // Calculate width of slots (aka channels) based on bandwidth and any spacing or padding required by the region:
     // spacing = gap between slots (0 for continuous spectrum) and at the beginning of the band
     // padding = gap at the beginning and end of the slots (0 for no padding)
-    float freqSlotWidth = newRegion->profile->spacing + (newRegion->profile->padding * 2) + (check_bw / 1000); // in MHz
-    uint32_t numFreqSlots = round((newRegion->freqEnd - newRegion->freqStart + newRegion->profile->spacing) / freqSlotWidth);
+    float spacing = newRegion->getSpacing(check_bw);
+    float padding = newRegion->getPadding(check_bw);
+    float freqSlotWidth = spacing + (padding * 2) + (check_bw / 1000.0f); // in MHz
```
caveman99, line 1188: The slot-width formula appears in four places with the identical three-line block, and MenuHandler.cpp:508 literally carries a comment whose whole job is to say "keep this in sync". One RegionInfo::getSlotPlan(bw) → {width, count, firstCenter} would collapse all four

test/test_admin_radio/test_main.cpp
caveman99, line 1: * The four new tests recompute the formula in the test body rather than calling production code. They also use (uint32_t) truncation where production uses floor()
* The invariant behind the whole floor() change is untested. Add a loop over every region × every preset in its profile asserting firstCenter + (numSlots-1)*width + bw/2 <= freqEnd

caveman99 (changes requested):

t-miura:
@caveman99
Thanks for the reviews!

I'll work on this to address issues you mentioned and others i've found.
Also make this aligns with #11747 and #11566, so that this'll be also under `src/mesh/reguratory` as much as possible.

PR #11520, opened 2026-08-16
Add telemetry history and decoupled reading intervals

oscgonfer:
Add telemetry history and decoupled reading intervals in telemetry modules (AQ for now until I get a green light for the rest, if I get it) for `SENSOR` roles only.

* Adds a reading interval for sensors in AQ telemetry module, to decouple from mesh/phone availability
* Adds logic for wake/sleep around sensor reads
* For sensor node, adds a queue that stores sensor readings, and publishes them in buffer to the mesh (MQTT coming up).
* Creates necessary template in TelemetryHistory
* Adds logic for publishTelemetry and publishBufferedTelemetry in BaseTelemetryModule for reusability in other Telemetry modules

Goes with https://github.com/meshtastic/protobufs/pull/1035 and https://github.com/meshtastic/protobufs/pull/1036. Works in the direction of: https://github.com/meshtastic/firmware/issues/10997

## 🤝 Attestations

- [X] I have tested that my proposed changes behave as described.
- [ ] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [ ] Other (please specify below)
  - [X] Xiao ESP-32-S3
  - [X] Muzi base

More thorough testing needed, and feedback expected...

```
=== BUFFER (TelemetryRecordHistory) from !19ba4462  [decrypted]  DIRECT/LOCAL MQTT (gateway=!19ba4462 == origin)  topic=msh/EU_868/2/e/FAB26/!19ba4462 ===
  interval_sec=360  readings=5
  [0] 2026-08-16 17:04:20 UTC  air_quality_metrics: pm10_standard=3, pm25_standard=4, pm100_standard=4, particles_05um=2540, particles_10um=430, particles_25um=20, particles_100um=0, pm40_standard=4, particles_40um=0, particles_tps=0.46399998664855957
  [1] 2026-08-16 17:10:52 UTC  air_quality_metrics: pm10_standard=3, pm25_standard=4, pm100_standard=4, particles_05um=2550, particles_10um=420, particles_25um=10, particles_100um=0, pm40_standard=4, particles_40um=0, particles_tps=0.3659999966621399
[...]

Issue #1035, opened 2021-12-29
Add tap back like feature to Messaging


garthvh:
Messaging clients for iOS and Android and many others all contain a tap back feature that allows users to quickly respond with preselected emoji options to a message.  

@mc-hamster and I were discussing this potential feature and it seemed worth capturing.

* Could potentially be implemented as a separate packet type with a replyTo messageId and a single unicode character in the payload.
* Could also be integrated into the text message protocol directly so If the "reply to" message id isn't set (eg: regular packets) then there's no cost over the wire. If it's set, then the payload could just be the unicode character for the tap back. We need to consider the impacts of any additions to the text message proto affecting maximum payload size
* An ASCII_BELL character could be utilized as the character used to send an "emergency button" message to all users on the main broadcast.

mc-hamster:
reply_id is the message id of the message you're replying to.
is_tapback determine what kind of reply this is. if false, it's a text reply. if true, it's a emoji reply.

if reply_id = something and is_tapback = false, then this is a reply like in ios or discord.
if reply_id = something and is_tapback = true, then the payload should only be a single character. Emoji or otherwise.
if reply_id = undefined and is_tapback = true, then the use wants to send a single large emoji or in the case of our discussion of ASCII_BELL, then the user wants to send an alert.

Change to protobuf to support this:
  https://github.com/mc-hamster/Meshtastic-protobufs/commit/67232d9c95f24cfbd5c6699701fb726eaecce216

mc-hamster:
Closing. Device support for this will be included in this week's release.

Clients will also need to adopt the feature.

mc-hamster:
This will be backward compatible for clients that don't explicitly support this as it'll just show up as normal messages.

I'll hold off on supporting this in S&F for the time being.

PR #11564, opened 2026-08-21
AirQualityTelemetry: decouple sensor reads from publishing, add a store for #11520's batch

caveman99:
Builds on @oscgonfer's work in #11520, which decouples telemetry read intervals from publishing and adds a batched offload for SENSOR nodes feeding a central processing server.

This PR takes the part of that design that needs no new protos and lands it now, shaped so the batched offload drops on top once meshtastic/protobufs#1035 is settled. The split is deliberate: #11520 owns the wire format and the server pipeline, this owns where readings live and when the sensor is read.

## What changes

`AirQualityTelemetryModule` only read its sensors from inside `sendTelemetry()`, so a reading existed only when something was about to be published, and the mesh path read again rather than reusing what the local device-to-phone loop had just measured.

Readings now go into a store, each with its capture time and a note of which consumers have taken it. The local loop keeps its own 60s cadence; the offload publishes one of those readings on `air_quality_interval` instead of reading again. One warm-up feeds the phone, the mesh, the screen and on-demand requests. Nothing changes on the wire.

Alongside:

- Wake all sleeping sensors in one pass and wait for the slowest (existing `TODO multiple sensors with different wake up times collide`).
- Stamp with `getValidTime(RTCQualityDevice)`. `getTime()` falls back to uptime-as-epoch, so RTC-less nodes were sending a bogus 1970 date.
- `allocReply()` answers from the store, not a synchronous I2C read that returns nothing while the sensor is asleep or warming up.
- Sleep eligibility measured against the local loop, not the on-air interval: a sensor whose warm-up outlasts the loop never finishes warming if slept in between.
- Dropped the awake-ahead-of-time bookkeeping, cadence runs from the read itself.

## Storage

`TelemetryStore<T>` holds the readings. `at()` copies rather than returning a reference, so a backing store need not be memory, and indices are logical and oldest-first.

- `RamTelemetryStore`: heap ring. Past 2KB that is already PSRAM on ESP32 via `heap_caps_malloc_extmem_enable()`. Capacity follows `MESHTASTIC_MEM_CLASS`, 64 / 16 / 8.
- `FileTelemetryStore`: fixed-size ring in a preallocated file, surviving reboot and deep sleep, which is what an offline node buffering until it has connectivity actually needs. Never grows, and a header records the payload size so a differing struct layout rebuilds rather than decoding nonsense. Filesystem is a template parameter defaulting to `FSCom`, so a variant can point it at SD. Off unless `AIR_QUALITY_TELEMETRY_HISTORY_PATH` is defined.

One contract test drives both, since the module should not be able to tell them apart.

This is meant to be the substrate the batch sits on: `at()` walks oldest-first and `publishedMask` records which consumers have taken each reading, so a `publishBufferedTelemetry` that collects everything with the mesh bit still clear is a small addition rather than a rework.

## Left to #11520

The batched offload, the `TelemetryRecordHistory` wire format, the `TELEMETRY_HISTORY_APP` portnum and the direct MQTT uplink. Nothing here duplicates or competes with those, and I would rather they landed there than be reinvented here.

## Notes from lining the two up

Things I ran into while making these fit together, offered as review fodder rather than objections:

- `TelemetryHistory.h` is wrapped in `#if !MESHTASTIC_EXCLUDE_ENVIRONMENTAL_SENSOR || !MESHTASTIC_EXCLUDE_AIR_QUALITY_SENSOR` while `BaseTelemetryModule.h` names the type unconditionally, which trips `rak3172`, `wio-e5`, `CDEBYTE_E77-MBL`, `milesight_gs301` and `heltec_wireless_bridge` separately from the proto sync. Taking the payload as a template parameter sidesteps needing the guard.
- `MeshService::sendToMesh` returns void, so the buffered path cannot tell whether the queue accepted the packet before marking those readings published.
- The read condition ORs in `telemetryPhoneDue && telemetryPhoneAllowed`, which holds reads at the 60s phone cadence rather than `air_quality_telemetry_read_interval`.
- `publishOwnPacket()` reaches the broker without the per-channel `uplink_enabled` and default-server checks `MQTT::onSend()` applies, so a node with uplink off would still publish its own telemetry.
- On #1035: a batch from an RTC-less sender carries no timestamp anchor, and `max_count:16` puts a `TelemetryRecordHistory` on the stack at 16 slots of the largest metrics variant. Both land on the server side rather than on any mesh node.

Happy to fold any of this into #11520 instead if that is the better home, or to rebase this under it if the batch lands first.

## Testing

New suite `test/test_air_quality_telemetry/`, 19 cases: gating as pure predicates following `PositionModule::shouldSendPositionToPhone`, the store contract against both backends, plus file reopen, rebuild-on-geometry-change and no-growth. Passing on `native-windows`.

Untested on hardware, I have no air quality sensor. Wants a check on a real PMSA003I, particularly a power-saving SENSOR that deep sleeps between offloads. The file backend is Portduino-verified only, the read-write open mode differs on the Adafruit/STM32 LittleFS wrappers, so nRF52 and STM32 want a hardware check first. Nothing enables it today.

## Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Native test suite only, no hardware regression testing, see above.

## Summary by CodeRabbit

- **New Features**
  - Added configurable air-quality reading history with persistent storage where available and automatic in-memory fallback.
  - Stored readings can be reused for mesh, phone, and request responses.

- **Improvements**
  - Independent sensor-reading and mesh-publication schedules improve telemetry timing.
  - Coordinated sensor warm-up and sleep handling helps reduce power consumption.
  - Prevents duplicate telemetry delivery and safely handles storage, sensor, and timing failures.
  - Preserves compatible telemetry history across restarts and recovers safely from incomplete data.
  - Improves timestamp accuracy and scheduling across clock rollovers.


caveman99:
Thanks, seven of the eight are fixed in 7d17e2513, plus the `unique_ptr` nitpick.

Declining one: **arm deep sleep when airtime blocks mesh transmission**. It is a real gap, but it predates this PR. On `develop` the equivalent path is

```cpp
if (telemetryDue && telemetryAllowed) {
    if (sendTelemetry()) { ... }
} else if (phoneDue && phoneAllowed) {
```

so `sendTelemetry()`, and therefore the sleep arming inside it, is already skipped when airtime says no. This PR preserves that behaviour rather than introducing it.

Worth fixing, but not here: making a power-saving SENSOR deep sleep through a blocked transmission means it sleeps the whole interval instead of retrying on the next tick, so under sustained congestion it would stop transmitting rather than wait for a gap. That trade wants its own PR and someone with the hardware to measure it.

On the two I took partially:

- **Geometry validation**: `create()` now removes the file first, and reopen rejects a file shorter than `offsetOf(slots)`, covered by `test_store_fileRebuiltWhenTruncated`.
- **Mask write failures**: reported via `LOG_WARN` rather than propagated. Returning a status would mean changing `markPublished()` on the `TelemetryStore` interface and threading it back through the send path, for a failure whose only consequence is that one reading may be offered again after a reboot. Happy to do it properly if a reviewer disagrees.

caveman99:
Both addressed in bb61e2889.

The truncation assert was a fair catch: without it a failed write leaves an empty file, which rebuilds because the magic is wrong rather than because the geometry is short, so the test passed without covering what it claimed to.

For the header commit, I went at the cause rather than the protocol. The finding is right that a full ring has no rollback: the record write already overwrote the oldest slot, so the header write was the only thing making it consistent, and generation-tagged records were the suggested fix. Taking that one step further removes the commit entirely. Records now carry the sequence, which also serves as the occupied flag, and `head`/`count`/`lastSeq` are derived by scanning the slots on open. The header holds only magic, version, record size and slot count, written once at create.

So a push is now a single record write with nothing to commit after it. There is no state that can disagree with the records, no stale header to serve, and no failure-injection case left to test, because the failing step no longer exists. It also halves the writes per reading, which is worth having on the flash-backed variants.

`test_store_fileKeepsOrderAcrossReopenWhenFull` covers the part that replaced it: fill two past capacity so `head` is off zero, reopen, check the order is intact and that a further push continues the ring instead of restarting it.

Format version went to 2. An older file fails the version check and gets rebuilt, which is the same path a payload-layout change already took.

caveman99:
Both parts fixed in 6ab484bf5, though not with a full integrity check.

`seq` being the first field was the hole: a short write lands a plausible sequence over a stale remainder, and `scanSlots()` took it. Records now repeat the sequence at the tail and are only accepted when head and tail agree, which catches a tear anywhere in between. `test_store_fileRejectsTornRecord` corrupts a record's trailing word and asserts the reopen skips that slot and keeps the intact one.

The exposed head slot is handled separately: a full ring has already overwritten its oldest record by the time the write fails, so `push()` now blanks that slot and rescans, keeping the rest of the ring, and marks the store unusable if the blanking also fails.

What I did not add is a CRC over the payload. The duplicated sequence covers torn writes, which is the failure this store can actually produce, but it does not cover bit rot in `metrics`. That felt like the right stopping point for a buffer of air-quality samples where a bad record means one wrong row in a database, not a safety or security consequence, and a CRC costs another four bytes per slot plus a table on parts where flash is already the scarce thing. Happy to add one if a reviewer wants the stronger guarantee.

Worth noting for anyone reading `scanSlots()`: rejecting a slot only stays safe because a torn record can only be the most recently written one, which is at one end of the ring, so the remaining records are still contiguous. A gap in the middle would break the index mapping. That cannot arise from this write path, but it is an assumption rather than something the code enforces.

Format version is 3, older files fail the check and rebuild.

caveman99:
Swept the nitpicks and the out-of-diff findings. Three fixed in ac092b9ba, one declined.

**Zero-capacity store** (outside diff, `FileTelemetryStore.h`): guard added. Worth saying it was not reachable: the constructor returns early on `slots == 0` leaving `usable` false, and `hdr.slots` is only trusted once it has been checked equal to a non-zero `slots`, so `push()` already returned at `!usable`. The guard keeps the modulo's precondition next to the modulo rather than a constructor trace away, and `test_store_fileWithoutCapacityStoresNothing` pins the same contract `RamTelemetryStore` already had.

**Hard-coded header length** (nitpick): fair, and I had flagged the same fragility myself when I wrote it. `headerBytes()` is now exposed and both file-layout tests use it, so a layout change breaks the build rather than quietly sending the test down the generic rebuild path.

**Pre-sleep comment** (nitpick): trimmed to two lines.

**Track publication only after enqueue success** (outside diff, `AirQualityTelemetry.cpp`): declining, with the awkwardness acknowledged, since this is the same class of thing I raised about #11520. The difference is what a fix would cost. `MeshService::sendToMesh()` and `sendToPhone()` both return `void`, so there is no status to consult; getting one means changing the `MeshService` API and every caller of it, which is well outside this PR. `develop` has the identical property today, where `sendTelemetry()` returns whether the sensor read succeeded, decided before the send.

The blast radius also differs from the batched case. Here a dropped packet costs one reading, and the local loop produces a fresh one a minute later. In a batched offload it costs everything in the batch, which is why it mattered more there.

If someone wants `MeshService` to report enqueue status, that is a worthwhile change on its own and I would be glad to help, but it should not ride in on an air-quality PR.

PR #11565, opened 2026-08-22
Allow full position precision on unencrypted ham channels

RCGV1:
Summary
- Allow configured full position precision only when Ham mode is enabled and the channel's effective key is unencrypted.
- Keep the 15-bit clamp for non-Ham cleartext traffic and for default/public PSKs, including in Ham mode.
- Add regression coverage for direct cleartext channels, default-PSK channels, and a secondary channel that inherits the primary key.

Why
- Ham-mode operation prohibits encryption; an operator using a genuinely unencrypted channel can opt into their configured full precision without weakening privacy handling for ordinary or default-key channels.

Validation
- `trunk fmt src/mesh/Channels.cpp src/mesh/Channels.h src/mesh/PositionPrecision.cpp src/mesh/PositionPrecision.h test/test_position_precision/test_main.cpp`
- `git diff --check`
- `~/.platformio/penv/bin/python -m platformio test -e native -f test_position_precision` (blocked before tests by existing macOS-native build errors: GNU-only VLAs and unavailable `linux/input.h`)
- `~/.platformio/penv/bin/python -m platformio test -e coverage -f test_position_precision` (blocked before tests: macOS Clang does not recognize `-fprofile-abs-path`)
- `~/.platformio/penv/bin/python -m platformio run -e rak4631` compiled the changed `Channels.cpp` and `PositionPrecision.cpp`; final LTO/link did not complete locally and was stopped.

## Summary by CodeRabbit

* **New Features**
  * Added channel status detection for unencrypted channels.
  * Improved position precision for licensed HAM users on eligible unencrypted public-key channels.

* **Bug Fixes**
  * Public-key position precision is now limited only when channels are publicly decryptable.

* **Tests**
  * Added coverage for position precision across encrypted, unencrypted, default-key, and inherited channel settings in licensed and unlicensed HAM modes.


RCGV1:
Addressed the inherited-key coverage gap in 09aaf3f02. The new regression case sets the primary PSK to empty, leaves the secondary PSK unset, enables Ham mode, and asserts full precision (32) through the effective-key inheritance path.

Validation: `trunk fmt test/test_position_precision/test_main.cpp` and the PR-file `git diff --check` pass. The focused native test command remains blocked before execution by existing macOS Clang errors in `NotificationRenderer.cpp` and the unavailable Linux input headers.

PR #11568, opened 2026-08-22
Report the right hardware model on four boards

thebentern:
Four variants declare a `custom_meshtastic_hw_model` that the build never actually reaches, so the device announces a different model in NodeInfo and the apps cannot match it for OTA. Same defect class as the ThinkNode M9 fix in #11567.

| env | manifest says | device actually reported |
| --- | --- | --- |
| `mini-epaper-s3` | `MINI_EPAPER_S3` (125) | `PRIVATE_HW` (255) |
| `heltec-v4-r8-oled`, `heltec-v4-r8-tft` | `HELTEC_V4_R8` (132) | `PRIVATE_HW` (255) |
| `heltec-mesh-node-t096` | `HELTEC_MESH_NODE_T096` (127) | `NRF52_UNKNOWN` (36) |
| `rak_wismesh_tap_v2`, `rak_wismesh_tap_v2-tft` | `WISMESH_TAP_V2` (116) | `RAK3312` (106) |

The first three had no arm in their platform's `HW_VENDOR` chain at all, so they fell through to `#else`.

The WisMesh Tap V2 is different: it defines **both** `RAK3312` and `RAK_WISMESH_TAP_V2`, and the generic `RAK3312` arm sat ahead of the specific one, so the generic arm always won. Reordered them, following the same rule the nRF52 chain already spells out for keeping custom RAK4630 boards ahead of the generic `RAK4630`.

## Verification

Each platform's `HW_VENDOR` chain was preprocessed with the env's full define set — build flags resolved through `extends`, the board JSON's `build.extra_flags`, and the bare `#define`s in the variant's own `variant.h` (several boards get their key symbol from there rather than from the ini, so reading the ini alone gives false negatives). All four now resolve to their declared model.

Regression-checked the reorder: `rak3312`, `heltec-v4` and `heltec-v4-tft` are unchanged. Swept the other 130 envs with a manifest and found no new mismatches. `bin/generate_ci_matrix.py` runs clean for all five platforms.

## Note

Touches the same `#elif` region of `src/platform/esp32/architecture.h` as #11567, so whichever lands second will need a trivial conflict resolution.

## 🤝 Attestations

## Summary by CodeRabbit

* **New Features**
  * Added hardware support for the Heltec V4 R8, Mini Epaper S3, and Heltec Mesh Node T096 devices.
  * Improved hardware identification for the RAK WisMesh Tap V2.
  * Added Raspberry Pi configurations for ZebraHat Duo radios.
  * Added SD card support for the M5Stack Cardputer ADV.
  * Expanded regional radio configuration and preset-based setup.

* **Bug Fixes**
  * Improved filesystem operations, power management, Bluetooth pairing, and display behavior.


rcarteraz:

rcarteraz (approved):

PR #11575, opened 2026-08-23
fix(test): remove brittle pointer comparison in test_radio

t-miura:
## Summary
Fixes CI failure in `test_radio` introduced in #11573.

## Root Cause

`test_beginSending_oversizedPayloadAbortsSafely` attempted to verify packet pool release by
allocating a second packet and asserting pointer equality (`TEST_ASSERT_EQUAL_PTR(p, reallocated)`). 

While this passed on local systems using standard glibc/macOS malloc caches (LIFO),
CI runs with AddressSanitizer (`-fsanitize=address` in `env:coverage`), which quarantines freed memory.

This caused the next `malloc` to return a newly mapped chunk instead of the freed pointer, failing the pointer check.

Below is snippet from [Example Failed Job](https://github.com/meshtastic/firmware/actions/runs/32613969701/job/97186444951#step:6:1195) :
```
  ❌ test_beginSending_oversizedPayloadAbortsSafely
  	Expected 0x0000514000000240 Was 0x0000514000000440
  ❌ coverage:test_radio
  	UnitTestError: Program received signal SIGHUP (Hangup)
```

## Changes
- Removed the brittle pointer address equality check and temporary reallocation from `test_beginSending_oversizedPayloadAbortsSafely` in `test/test_radio/test_main.cpp`.
- Verifies `beginSending()` returns `0` and `sendingPacket` remains `nullptr`.
- Packet release is implicitly and strictly verified across native tests by LeakSanitizer (LSan) on exit.

## 🤝 Attestations

- [x] I have tested that my proposed changes behave as described.
- [x] I have tested that my proposed changes do not cause any obvious regressions on the following devices:
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [ ] Seeed Studio T-1000E tracker card
  - [x] Other (please specify below)
    - [x] macOS( Tahoe 26.5.2 (25F84) arm64, with `pio test -e native-macos -f test_radio` )
    - [x] Linux (Debian 13 / Kernel 7.0.12-1 x86_64, with `pio test -e native -f test_radio` and  `pio test -e coverage`)
    - [x] Windows (On WSL2 / Ubuntu 22.04 x86_64 with `pio test -e native -f test_radio` and `pio test -e coverage`)

## Summary by CodeRabbit

* **Tests**
  * Updated oversized-payload coverage to confirm that rejected transmissions are not queued.
[...]

PR #11586, opened 2026-08-24
fix(test): assert the rejected packet is released, not that malloc reuses its address

thebentern:
`develop` has been red since #11573 merged on Aug 23. One assertion in one test fails on every CI run that reaches the native suite:

```
test/test_radio/test_main.cpp:440:test_beginSending_oversizedPayloadAbortsSafely:FAIL:
    Expected 0x0000514000000240 Was 0x0000514000000440
```

## Cause

The test allocated a probe packet after `beginSending()` took its rejection path and asserted the pool handed the same address back:

```cpp
meshtastic_MeshPacket *reallocated = packetPool.allocZeroed();
TEST_ASSERT_EQUAL_PTR(p, reallocated);
```

That holds for the fixed-slot `MemoryPool<T, N>` used on embedded targets. It does not hold where the test actually runs: under `ARCH_PORTDUINO`, `Router.cpp` binds `packetPool` to `MemoryDynamic`, which is plain `malloc`/`free`, and the `coverage` env compiles with `-fsanitize=address`. ASan's quarantine deliberately withholds a freed chunk from the next `malloc`, so the address always differs. The addresses are byte-identical across all three failing CI runs and a local repro — this is deterministic, not a flaky allocator race.

`RadioInterface::beginSending()` is correct and unchanged here. It releases the oversized packet and returns 0. The LeakSanitizer report in the failing logs points at the test's own probe allocation, orphaned because Unity longjmps out of the failed assert before `release()` runs.

The second CI error, `area misc ran suites that did not match their own test binaries (1 unsourced)`, is downstream of the same assert: the ASan abort makes PlatformIO emit a synthetic whole-suite `coverage:test_radio` case with no file attribute, which `check-test-attribution.py` rejects. Both errors clear with this change.

## Change

`packetPool` is already tagged `pktpool(live)` for memaudit, and `memaudit::snapshot()` is used this way in `test_mem_audit`. Comparing live bytes across the call proves the invariant the test is named for — the rejected packet went back to the pool — without depending on allocator behaviour.

## Verification

Run in the `coverage` env (ASan), the same configuration CI uses:

| | Result |
|---|---|
| `develop` as-is | FAILED — `Expected 0x0000514000000240 Was 0x0000514000000440`, suite ERRORED on SIGHUP after the LSan abort |
| with this change | 23/23 PASSED, no sanitizer output |
| with this change + `packetPool.release(p)` removed from `beginSending()` | FAILED — `Expected 0 Was 432` |

The third row confirms the new assertion still catches the regression the test was written for, and reports it as one leaked packet (432 bytes = `sizeof(meshtastic_MeshPacket)`) rather than two opaque addresses.

## Summary by CodeRabbit

* **Tests**
  * Strengthened validation for oversized payload handling.
  * Confirmed rejected packets do not leave live packet-pool memory allocated.
  * Improved test reliability across different memory-management and sanitizer environments.

### DIFF
--- a/test/test_radio/test_main.cpp
+++ b/test/test_radio/test_main.cpp
@@ -3,6 +3,8 @@
 #include "MeshService.h"
 #include "RadioInterface.h"
 #include "TestUtil.h"
+#include "memory/MemAudit.h"
+#include <cstring>
 #include <unity.h>
 
 #include "meshtastic/config.pb.h"
@@ -417,8 +419,21 @@ static void test_regionPresetMap_unsetCarriesUserprefsIntent()
 #endif
 }
 
+// Live bytes packetPool reports under its memaudit tag, which tracks every alloc/release.
+static int32_t livePacketPoolBytes()
+{
+    memaudit::Tag rows[memaudit::kMaxTags];
+    size_t n = memaudit::snapshot(rows, memaudit::kMaxTags);
+    for (size_t i = 0; i < n; i++)
+        if (rows[i].tag && strcmp(rows[i].tag, "pktpool(live)") == 0)
+            return rows[i].bytes;
+    return 0;
+}
+
 static void test_beginSending_oversizedPayloadAbortsSafely()
 {
+    const int32_t liveBefore = livePacketPoolBytes();
+
     meshtastic_MeshPacket *p = packetPool.allocZeroed();
     TEST_ASSERT_NOT_NULL(p);
     p->from = 0x12345678;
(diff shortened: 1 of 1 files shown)


t-miura:
@thebentern
thank you for the follow-up PR that's better than mine( #11575 ),
and apologize for the fact that i couldn't noticed defects on #11573 's test code while it's in draft and reviewing phase.

I will close #11575 as this will supersedes it with better test coverage and environment-independent results.

PR #11593, opened 2026-08-24
Meshnology W10: enable the AXP2101 power key as a second button

rcarteraz:
## What

Makes the Meshnology W10's power key (`SW3`) usable as a second button, mapping a short press to the existing `INPUT_BROKER_CANCEL` action. Variant-only change — no shared code is touched.

## Why it doesn't work today

`SW3` is wired to the AXP2101 `PWRON` pin via `R44` 510R (schematic W10-MB-V1.1 pg3), but pressing it does nothing in firmware.

`Power::runOnce()` already polls the PMU IRQ status registers over I2C and maps a PEK short press to `INPUT_BROKER_CANCEL` when `PMU_POWER_BUTTON_IS_CANCEL` is set:

```c
PMU->getIrqStatus();
...
#ifdef PMU_POWER_BUTTON_IS_CANCEL
    if (PMU->isPekeyShortPressIrq()) { ... }
#endif
```

But the matching `PMU->enableIRQ(pmuIrqMask)` sits inside `#ifdef PMU_IRQ`, and PMU init runs `disableIRQ(XPOWERS_AXP2101_ALL_IRQ)` beforehand. So without `PMU_IRQ`, the `PKEY_SHORT` status bit is never armed, the polled read is always false, and `PMU_POWER_BUTTON_IS_CANCEL` on its own is inert. I confirmed this empirically — with only the cancel define, 100s of pressing produced zero events.

## Approach

`AXP_IRQ` on this board reaches only expander `EXIO5` and is not routed to any ESP32 GPIO (confirmed on both schematic sheets), so there is no real pin to give `PMU_IRQ`. This defines it as the MCP23017 virtual pin instead, mirroring how the same limitation is already handled for `LORA_DIO1` on this variant:

```c
#define EXIO_AXP_IRQ 5
#define PMU_IRQ (MCP23017_VPIN_BASE + EXIO_AXP_IRQ) // 105
```

The `attachInterrupt()` and `gpio_wakeup_enable()` uses of `PMU_IRQ` are inert on a non-GPIO value — both are unchecked calls, so an invalid pin is ignored rather than fatal. What the define actually buys is the `enableIRQ()` that they gate.

## ⚠️ Please sanity-check this bit

Using `PMU_IRQ` as a non-GPIO sentinel is arguably leaning on a side effect rather than expressing intent, and I'd rather a maintainer decide whether that's acceptable. The cleaner alternative is to hoist the mask computation and `enableIRQ()` out of `#ifdef PMU_IRQ` in `Power.cpp`, leaving only `pinMode(PMU_IRQ, INPUT)` guarded. That would be a general fix for any AXP board whose IRQ pin isn't routed to a GPIO — and would also restore VBUS insert/remove events, which are dead on such boards for the same reason. I kept this PR variant-scoped to avoid touching shared code without a maintainer's call, but I'm happy to switch it if that's preferred.

## Known limitation

Because there's no real interrupt, the ISR's `power->setIntervalFromNow(0)` never runs, so events surface on the 20s `Power::runOnce()` cadence rather than immediately. The press isn't lost — the AXP2101 latches it — but it can take up to 20s to register. Visible in the test log below as events landing exactly 20s apart.

## Testing

Built and flashed on real Meshnology W10 hardware (firmware `2.8.0.bfd1e1a`; the two files involved are identical between that commit and current `develop`).

```
23:27:58  [UserButton] LONG PRESS RELEASE AFTER 716 MILLIS
23:28:01  [UserButton] LONG PRESS RELEASE AFTER 864 MILLIS
23:28:05  [Power] Input: Corona Button Click
23:28:18  [UserButton] LONG PRESS RELEASE AFTER 777 MILLIS
23:28:22  [UserButton] LONG PRESS RELEASE AFTER 909 MILLIS
23:28:25  [Power] Input: Corona Button Click
23:28:30  [UserButton] LONG PRESS RELEASE AFTER 724 MILLIS
23:28:45  [Power] Input: Corona Button Click
23:29:05  [Power] Input: Corona Button Click
```

- Power key (`SW3`) produces `Input: Corona Button Click`
- The existing GPIO0 user button (`SW2`) continues to work independently
- Board boots normally; the non-GPIO `PMU_IRQ` value causes no crash or boot loop

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
  - [x] Other: **Meshnology W10** — the only device tested, and the only one affected. The diff is confined to `variants/esp32s3/meshnology-w10/variant.h`, so no other target's build output changes. 

## Summary by CodeRabbit

- **New Features**
  - Added support for canceling power-button actions on the Meshtastic W10 variant.
  - Added AXP2101 power-management interrupt handling through the expander’s virtual pin for improved power-event monitoring.

- **Documentation**
  - Clarified that AXP2101 interrupt handling is polled over I2C, while GPIO interrupt and wake-up operations remain inactive for the virtual pin.

### DIFF
--- a/variants/esp32s3/meshnology-w10/variant.h
+++ b/variants/esp32s3/meshnology-w10/variant.h
@@ -12,9 +12,16 @@
 
 // ─── Power management ─────────────────────────────────────────────────────────
 // AXP2101 PMIC on I2C (pg3). AXP_IRQ → EXIO5 (expander, whose /INT is not routed to the
-// ESP32), so no PMU_IRQ is possible; battery state is polled via the AXP2101 fuel gauge.
-// There is no direct battery ADC - the PMIC is the only voltage source.
+// ESP32), so there is no GPIO-backed PMU_IRQ; battery state is polled via the AXP2101 fuel
+// gauge. There is no direct battery ADC - the PMIC is the only voltage source.
 #define HAS_AXP2101
+
+// Power key: SW3 → R44 510R → PWRON (AXP2101 pin 30, schematic W10-MB-V1.1 pg3), used as a
+// second button. PMU_IRQ is defined below as the expander vpin for AXP_IRQ: Power.cpp keeps
+// the PMU IRQ *enable* (enableIRQ of PKEY_SHORT / VBUS insert+remove) behind #ifdef PMU_IRQ,
+// and init runs disableIRQ(ALL), so without it the status bit never latches and the polled
+// read in Power::runOnce() is always false.
+#define PMU_POWER_BUTTON_IS_CANCEL
 
 // ─── RTC ──────────────────────────────────────────────────────────────────────
 // PCF85063ATL on I2C (pg3); RTC_INT → EXIO4 (unused)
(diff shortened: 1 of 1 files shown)


thebentern (approved):

thebentern:
@copilot trunk fmt

PR #11599, opened 2026-08-25
Fix trackball dropping short presses and losing the click when tilted

caveman99:
The press interrupt shared its debounce timestamp with the direction axes and runOnce() required the pin to still read low, so a tilted click or a tap released between 50 ms polls produced no event at all.

## Summary by CodeRabbit

- **New Features**
  - Improved trackball press handling with reliable tap, short-press, and long-press recognition.
  - Added repeated input events while a button remains held.
  - Improved press debouncing and handling of rapid retriggering.

- **Bug Fixes**
  - Prevented unintended direction events when a press begins.
  - Improved behavior across timer wraparound conditions.

- **Tests**
  - Added coverage for press timing, release behavior, repeats, retriggering, idle polling, and timer rollover.

### DIFF
--- a/src/input/TrackballInterruptBase.cpp
+++ b/src/input/TrackballInterruptBase.cpp
@@ -31,9 +31,10 @@ TrackballInterruptBase::PressResult TrackballInterruptBase::updatePress(bool irq
     if (!irqLatched)
         return PressResult::None;
 
-    // Already released by the time we polled: the tap only exists as the latched interrupt.
+    // Already released by the time we polled: classify from the latched interrupt time, since a
+    // delayed poll can hide a long hold behind the same latch.
     if (!pinLow)
-        return PressResult::Short;
+        return Throttle::isWithinTimespanMs(irqTimeMs, LONG_PRESS_DURATION) ? PressResult::Short : PressResult::None;
 
     pressDetected = true;
     pressStartTime = irqTimeMs;
--- a/src/input/TrackballInterruptBase.cpp
+++ b/src/input/TrackballInterruptBase.cpp
@@ -246,7 +246,8 @@ int32_t TrackballInterruptBase::runOnce()
 
 void TrackballInterruptBase::intPressHandler()
 {
-    if (Throttle::isWithinTimespanMs(lastPressInterruptTime, 10))
+    // pressIrqSeq == 0 means nothing recorded yet, so a press at clock 0 is not read as a cooldown.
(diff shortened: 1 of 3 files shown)


caveman99:
All three addressed.

Rollover sentinel: repeat emission now tracked by a longPressRepeatSent flag instead of lastLongPressEventTime == 0, with a regression test that emits a repeat exactly when the clock reads 0 and asserts the next poll still waits the full interval.

Latch race: solved with a sequence counter rather than a critical section. noInterrupts() appears nowhere in src/ and has no portduino stub, so introducing one risked the native build that links this file. The ISR increments pressIrqSeq and runOnce compares against its own copy, so an interrupt landing mid-poll is picked up on the next poll instead of being cleared unseen.

Direction repeat overwriting a press: the repeat assignment is now guarded on e.inputEvent == INPUT_BROKER_NONE.

Header comment trimmed to two lines. 9/9 native tests pass.

caveman99:
Addressed in a follow-up commit. The threshold block is now gated on both !pressLatched and e.inputEvent == INPUT_BROKER_NONE, so a short press classified on a later poll survives a direction counter reaching TB_THRESHOLD. That is the third instance of the same class in this file, after the direction repeat block and the non-threshold chain.

On the requested threshold-mode regression test: TB_THRESHOLD is a compile-time variant macro (only t-deck sets it) and the block lives in runOnce(), which reads GPIO directly, so it cannot be exercised from the native suite. The extracted updatePress() state machine is what the tests cover; these three guards are reachable only through runOnce().

thebentern (approved):

caveman99:
Fixed. The debounce is now skipped until an interrupt has actually been recorded, using the sequence counter that was already there (pressIrqSeq == 0 means nothing recorded yet) rather than adding another flag.

Same sentinel class as the lastLongPressEventTime finding earlier in this PR - I fixed that one and left the identical pattern sitting in lastPressInterruptTime two lines away.

caveman99:
Fixed. The already-released branch now classifies from irqTimeMs against LONG_PRESS_DURATION, matching the active-release path, so a delayed poll no longer turns a long hold into a tap. Added test_long_hold_released_before_first_poll_is_not_short covering exactly that: latch an interrupt, advance past LONG_PRESS_DURATION, poll with the pin already high, expect no short-press event. 10/10 native tests pass.

Note this landed after thebentern's approval, so that may need re-confirming.

PR #11600, opened 2026-08-25
Show waypoints sent with no expiry and stop expiring on an unset clock

caveman99:
shouldDraw() tested expire > getTime(), which hid every waypoint sent with expire == 0 (what the clients send when no expiry is picked, while 1 is the delete convention) and, before the RTC is set, compared against seconds since boot.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved waypoint expiration handling for greater reliability.
  * Waypoints configured with no expiration remain active indefinitely.
  * Deleted waypoints are correctly recognized as inactive.
  * Expired and currently expiring waypoints are deactivated at the correct time.
  * Expiration checks now behave safely when the device clock is unavailable or untrusted.
* **Tests**
  * Added coverage for expiration boundaries, indefinite waypoints, deletion markers, and invalid clock conditions.

### DIFF
--- a/src/meshUtils.h
+++ b/src/meshUtils.h
@@ -85,6 +85,15 @@ bool sanitizeUtf8(char *buf, size_t bufSize);
 // left at the cut.
 void clampLongName(char *longName);
 
+// Is a received Waypoint still live? The clients send expire == 0 for "never expires" and expire == 1
+// to delete; now == 0 means we have no trustworthy clock, which must not expire anything.
+static inline bool waypointIsActive(uint32_t expire, uint32_t now)
+{
+    if (expire <= 1)
+        return expire == 0;
+    return now == 0 || expire > now;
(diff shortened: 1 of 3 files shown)


caveman99:
Addressed the header-comment nitpick. Declining the other two:

snake_case test names: every existing suite in test/ uses snake_case (test_colon_separated_uppercase, test_waypoint_serialization). Matching the neighbouring files is worth more here than the generic camelCase rule.

Direct shouldDraw() coverage: it sits inside #if HAS_SCREEN and the native test env builds with HAS_SCREEN=0, so it cannot be compiled there. That constraint is why the predicate was extracted into meshUtils.h in the first place. The call site is covered by CI builds and was verified on a Wio Tracker L1.

thebentern (approved):

PR #11602, opened 2026-08-25
Render waypoint name and description through the emote renderer

caveman99:
Waypoint name and description were drawn with plain drawString, so emoji from the phone clients reached the OLED font as raw bytes; rows carrying a 16 px glyph now get the height they need, and a second tall row falls back to plain text where the panel has no room.

## Summary by CodeRabbit

* **New Features**
  * Waypoint names and descriptions now support emoji rendering.
  * Waypoint panels dynamically adjust row heights to accommodate emoji content.
  * Descriptions fall back to plain text when insufficient space is available.
  * Compass and distance information are positioned more accurately around expanded content.

### DIFF
--- a/src/modules/WaypointModule.cpp
+++ b/src/modules/WaypointModule.cpp
@@ -120,7 +120,18 @@ void WaypointModule::drawFrame(OLEDDisplay *display, OLEDDisplayUiState *state, 
 
     // A glyph row is all-or-nothing: half its height just moves the collision onto the row below. A
     // row only grows where the panel has the height for it, else that row stays plain text.
-    const int growthBudget = std::max(0, SCREEN_HEIGHT - (textPos[4] + FONT_HEIGHT_SMALL));
+    // Portrait puts the compass under the text, so its reserve comes out of the budget too;
+    // landscape keeps the compass beside the text and needs no reserve.
+    int growthBudget = std::max(0, SCREEN_HEIGHT - (textPos[4] + FONT_HEIGHT_SMALL));
+    if (SCREEN_WIDTH <= SCREEN_HEIGHT) {
+#if defined(USE_EINK)
+        const int navBar = ((graphics::currentResolution == graphics::ScreenResolution::High) ? 16 : 8) + 6;
+#else
+        const int navBar = 0;
+#endif
+        const int minCompassDiameter = 2 * 8; // matches the radius clamp applied below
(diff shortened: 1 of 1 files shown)


caveman99:
Good catch, applied. Used descTall || descCost == 0 rather than the suggested three-term condition: descCost is only ever non-zero when hasEmote is true, so !descMetrics.hasEmote is subsumed by descCost == 0 and the extra term is redundant.

caveman99:
Both addressed in a follow-up commit.

Compact panels: good catch, my "one tall row always fits" assumption was wrong. nameCost is now subject to the same budget test as descCost, so on 72x40 and 64x48 the name row stays plain text rather than pushing content past SCREEN_HEIGHT.

Emote that fits the row: used descTall || descCost == 0 rather than the three-term suggestion, since descCost is only non-zero when hasEmote is true, so !hasEmote is subsumed. Same shape now applied to the name row.

caveman99:
Fixed. The portrait branch draws the compass under the text, so its reserve (nav bar + margin + the 16px minimum diameter the radius clamps to) now comes out of growthBudget before a row is allowed to grow. Landscape is left alone, since there the compass sits beside the text and never competed for that space.

Worth noting for anyone reviewing: the hardware testing behind this PR was all on a 128x64 Wio Tracker L1, which takes the landscape branch, so the portrait path here has never been looked at on a real panel - only reasoned about.

HarukiToreda:
this is already something included on the waypoint geofence PR for both UIs
https://github.com/meshtastic/firmware/pull/10920

caveman99:
Superseded by #10920 (Add waypoint & geofence support with notifications for BaseUI and InkHUD, cd6ac90f7).

That PR rewrote WaypointModule::drawFrame into the multi-waypoint card list and already renders both fields through the emote renderer - graphics::UIRenderer::drawStringWithEmotes for the name and description, plus truncateStringWithEmotes for fitting them - which is exactly what this PR set out to fix. The single-waypoint textPos[] layout and the portrait compass block these commits modified no longer exist.

One small residue for whoever owns the new list view: rows are pitched at FONT_HEIGHT_SMALL + 1 (15) against a 16px emote bitmap, so a tall glyph still overlaps the row below by about 1px. Not worth a PR on its own, but noting it here since it is the same class of issue this branch was chasing.

PR #11612, opened 2026-08-25
feat(admin): append the optional ham long_name to the call sign

vidplace7:
`HamParameters` gained a `long_name` field in meshtastic/protobufs#941, but `AdminModule::handleSetHamMode()` never read it — a client that sent one still got a node named after the bare call sign.

## What changed

`long_name` is now joined behind the call sign with the `//` separator hams already use on the air, as the protobuf change specified:

| `call_sign` | `long_name` | resulting `owner.long_name` |
| --- | --- | --- |
| `KD2ABC` | `Attic Heltec` | `KD2ABC//Attic Heltec` |
| `KD2ABC` | *(unset or blank)* | `KD2ABC` |

`long_name` is optional in both senses: unset **and** whitespace-only fall back to the call-sign-only name. It deliberately stays out of the whitespace-only rejection that guards `call_sign` and `short_name` — the field is cosmetic, and that path returns after only a `LOG_WARN`, so a stray space would have silently cost the operator the entire licensing request with nothing visible in the app. The identity fields keep their existing strictness.

The composed name is finished with `clampLongName()` instead of a bare `sanitizeUtf8()`, matching `handleSetOwner()` and `NodeDB`. The proto caps the parts at 7 + `//` + 14 bytes, which sits inside the 24-byte `MAX_LONG_NAME_BYTES` budget with a byte spare; `clampLongName()` is the backstop (graceful truncation with UTF-8 repair) if either cap ever moves.

No wire-format or protobuf change — the field already exists in the submodule on `develop`. The on-device region picker in `MenuHandler::applyLoraRegion()` sends `_init_zero`, so its behavior is unchanged.

## Testing

Five cases added to `test/test_admin_radio` covering append, the widest pair the proto can carry (7 + 2 + 14, which must arrive whole), unset, blank-is-ignored, and a regression pin that whitespace-only `short_name` is still rejected.

`bin/run-tests.sh` is Linux-only and Docker was unavailable on this host, so the suites were run directly through PlatformIO's `native-macos` env with an isolated `$HOME`:

- `test_admin_radio` — 98 tests, 0 failures, 0 ignored
- `test_admin_session_repro` — 25 tests, 0 failures
- `test_pki_admin_fallback` — 6 tests, 0 failures

That is three suites, not the full canonical run, and `native-macos` is headless (`HAS_SCREEN=0`) so `#if HAS_SCREEN` cases were skipped — CI is the real gate. `trunk` and `clang-format` are not installed on this host either; the new lines were hand-checked against `.trunk/configs/.clang-format` (102 chars max against the 130 limit), so the formatter pass in CI is worth watching.

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

**Not tested on hardware.** Verification here is the native unit suite only. The change is confined to the admin ham-mode handler, so a check that ham mode still licenses a node — and that a `long_name` sent from a client shows up as `CALLSIGN//Name` — would be welcome from anyone with a device in front of them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ham mode requests with blank call signs are now rejected with a clear `BAD_REQUEST` response.
  * Valid ham mode updates continue to be accepted and applied successfully.

* **Tests**
  * Expanded coverage for accepted and rejected ham mode requests, including direct handling and request routing.

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -368,7 +368,10 @@ bool AdminModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshta
         break;
     case meshtastic_AdminMessage_set_ham_mode_tag:
         LOG_DEBUG("Client set ham mode");
-        handleSetHamMode(r->set_ham_mode);
+        // Without this a rejected request falls through to the generic Routing_Error_NONE ack below,
+        // so a client would report ham mode as enabled on a node that changed nothing.
+        if (!handleSetHamMode(r->set_ham_mode))
+            myReply = allocErrorResponse(meshtastic_Routing_Error_BAD_REQUEST, &mp);
         break;
     case meshtastic_AdminMessage_get_ui_config_request_tag: {
         LOG_DEBUG("Client is getting device-ui config");
@@ -1928,7 +1931,7 @@ static bool isBlankName(const char *start)
     return *start == '\0';
 }
 
-void AdminModule::handleSetHamMode(const meshtastic_HamParameters &p)
+bool AdminModule::handleSetHamMode(const meshtastic_HamParameters &p)
 {
     // Validate ham parameters before setting since this would bypass validation in the owner struct.
 
@@ -1936,7 +1939,7 @@ void AdminModule::handleSetHamMode(const meshtastic_HamParameters &p)
     // without it we would license a node that never identifies itself on the air.
     if (isBlankName(p.call_sign)) {
         LOG_WARN("Rejected ham call_sign: needs 1+ non-whitespace char");
-        return;
+        return false;
     }
 
     // Set call sign and override lora limitations for licensed use. An optional long_name rides
@@ -1987,6 +1990,7 @@ void AdminModule::handleSetHamMode(const meshtastic_HamParameters &p)
 
     service->reloadOwner(false);
     saveChanges(SEGMENT_CONFIG | SEGMENT_NODEDATABASE | SEGMENT_DEVICESTATE | SEGMENT_CHANNELS);
+    return true;
 }
 
 AdminModule::AdminModule() : ProtobufModule("Admin", meshtastic_PortNum_ADMIN_APP, &meshtastic_AdminMessage_msg)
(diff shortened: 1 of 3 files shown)


RCGV1 (changes requested):
Requesting changes for one client-visible error-handling bug.

### [P2] Return `BAD_REQUEST` when a blank callsign is rejected

`handleSetHamMode()` returns early for an empty or whitespace-only `call_sign`, but the dispatcher receives no failure signal. With `want_response=true`, the generic reply path therefore emits `Routing_Error_NONE`. A client can report that Ham mode was enabled even though the node rejected the request and made no state change.

I reproduced this at `c97e96f` on both:

- `TLORA_T3_S3` (ESP32-S3)
- `MUZI_BASE` (nRF52840)

On each target, a whitespace-only callsign left owner/licensing state unchanged but returned `ROUTING_APP / NONE` for the matching request ID.

Please return `Routing_Error_BAD_REQUEST` from this path, either by validating in the dispatcher or propagating failure from `handleSetHamMode()`. Add a protocol-level test that exercises `want_response`; the current handler-level tests correctly verify no mutation but cannot observe the generated success response.

The intended behavior otherwise passed on both targets:

- `KO6CNT` plus a long name persisted as `KO6CNT//<name>`
- blank optional `short_name` preserved the existing short name
- blank optional `long_name` produced the callsign alone
- node identity remained stable

Build and validation results:

- `pio run -e tlora-t3s3-v1` passed
- `pio run -e muzi-base` passed
- hosted native and target CI are green
- local focused native execution was blocked before tests by existing Apple Clang incompatibilities (`-fprofile-abs-path` and initialized VLAs in `NotificationRenderer.cpp`)

vidplace7:
> Requesting changes for one client-visible error-handling bug.
> 
> ### [P2] Return `BAD_REQUEST` when a blank callsign is rejected
> `handleSetHamMode()` returns early for an empty or whitespace-only `call_sign`, but the dispatcher receives no failure signal. With `want_response=true`, the generic reply path therefore emits `Routing_Error_NONE`. A client can report that Ham mode was enabled even though the node rejected the request and made no state change.

Yeah good catch.
Addressed in d9ca2bfcc02ca84fa23b1e6ae00700fc5cb32487

caveman99 (approved):
Apart from the excessive comments, this looks fine.

PR #11613, opened 2026-08-26
fix(nrf52): drive the PWM RTTTL sequencer from a FreeRTOS timer

Ixitxachitl:
The NonBlockingRTTTL sequencer is advanced from `ExternalNotificationModule::runOnce()`, which
asks for a 25 ms interval while a ringtone plays. `tone()` on nRF52 is hardware timed, so a note
*ends* on time, but the next note only starts when the module thread next runs. Every note
boundary therefore carries a gap as long as whatever else the cooperative loop was doing.

That is audible on boards whose display path blocks the loop. TFT_eSPI's generic processor
backend pushes each pixel as two single-byte `SPI.transfer()` calls, and each of those is a
separate blocking `nrfx_spim_xfer` transaction, so a repaint costs tens of milliseconds — and a
ringtone is exactly when the screen is busiest, since the notification wakes the display and
`setFastFramerate()` raises it to 30 fps. E-ink boards with a PWM buzzer have the same exposure
with a much longer stall. The result is dropped or ragged notes for the duration of the alert.

This moves note scheduling off the main loop on nRF52:

- `NRF52RtttlTicker` runs `rtttl::play()` from a 2 ms FreeRTOS software timer.
- `begin()`/`stop()` stay on the calling thread under a mutex; the timer callback takes it
  non-blocking and skips a tick rather than blocking the timer service task. The timer stops
  itself when the song ends.
- Falls back to the previous main-loop polling if the timer or mutex cannot be created.

`ExternalNotificationModule` reaches the sequencer through three thin wrappers on the `use_pwm`
path. Other architectures call `rtttl::` exactly as before. The nRF52 I2S-speaker path
(`nrf52RtttlPlayer`) is untouched.

Applies to any nRF52 board using the PWM buzzer.

## Testing

Tested on Heltec Mesh Node T1 (TFT), Seeed Wio Tracker L1 (OLED) and LilyGo T-Echo Plus (e-ink):
ringtones play cleanly while the notification banner draws and the panel refreshes, where they
previously stuttered. Both display architectures the fix targets are covered.

## Attestations

- [x] behaves as described — T1, Wio Tracker L1, T-Echo Plus
- [x] no obvious regressions — Other: T1, Wio Tracker L1, T-Echo Plus

## Summary by CodeRabbit

* **New Features**
  * Added improved RTTTL melody playback on nRF52 devices.
  * Playback now advances automatically with more consistent timing.
  * Added fallback handling to maintain playback when timer initialization is unavailable.

* **Bug Fixes**
  * Improved starting, updating, and stopping of PWM-based notification sounds.
  * Prevented regular buzzer tones from interrupting an active ringtone.
  * Improved coordination between timed playback and normal device operation.


src/platform/nrf52/NRF52RtttlTicker.cpp
```diff
@@ -0,0 +1,82 @@
...
+// True only while the timer is servicing a song; pump() polls from the main loop whenever it is not.
+bool timerRunning = false;
+
+void onTick(TimerHandle_t)
+{
+    // The main thread holds the lock only across begin()/stop(); skip this tick rather than block the timer task.
+    if (xSemaphoreTake(lock, 0) != pdTRUE)
+        return;
+    if (rtttl::isPlaying())
+        rtttl::play();
```
caveman99, line 33: The mutex covers NonBlockingRTTTL's state, not Tone.cpp's. `tone()` on this core mutates a file-static `_pwm_config` and drives PWM2 through a non-atomic stopPlayback -> initializeFromPulseCountAndTimePeriod -> applyConfiguration -> startPlayback sequence. `src/buzz/buzz.cpp:155` calls `tone()` from the main thread and can now interleave with this callback.

`InputBroker::handleInputEvent` closes the button path by calling `stopNow()` and returning early while `nagging()`, but these callers do not go through it:

- `src/gps/GPS.cpp:238`, `:243`, `:2326`, `:2337` (`playGPSEnableBeep` / `playGPSDisableBeep` on lock and sleep transitions)
- `src/graphics/niche/InkHUD/Events.cpp:61`, `:99` (`playChirp` / `playBoop`)

Failure mode: the main thread is preempted between `applyConfiguration()` and `startPlayback()`, the daemon reconfigures and starts its own note, the main thread resumes and triggers a SEQSTART computed for a config the registers no longer hold. One wrong frequency or wrong length note. Not a hang, since both callers always pass a non-zero duration, so the infinite `LOOPSDONE_SEQSTART0` shorts path is unreachable.

One line in `playTones()` covers it, and also fixes the pre-existing same-thread case where a system beep stomps a ringtone note:

```cpp
if (rtttl::isPlaying())
    return; // a notification ringtone owns the buzzer
```

src/platform/nrf52/NRF52RtttlTicker.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include "DebugConfiguration.h"
+#include "freertosinc.h"
+#include <NonBlockingRtttl.h>
+#include <timers.h>
+
+namespace NRF52RtttlTicker
+{
+namespace
+{
+constexpr uint32_t kTickMs = 2;
```
caveman99: 2 ms is 512 daemon wakeups per second. The stall being fixed is tens of milliseconds, so 5 ms is inaudible here and cuts the wake rate by 2.5x.

`pdMS_TO_TICKS(kTickMs)` must also not round to zero. At `configTICK_RATE_HZ 1024` it is 2 ticks. Below 500 Hz it becomes 0, `xTimerCreate` returns NULL, and the fix degrades to the old polling silently. Worth a comment.

src/platform/nrf52/NRF52RtttlTicker.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+bool ensureInit()
+{
+    if (timer)
+        return true;
+    if (!lock)
+        lock = xSemaphoreCreateMutex();
+    if (lock)
+        timer = xTimerCreate("rtttl", pdMS_TO_TICKS(kTickMs), pdTRUE, nullptr, onTick);
```
caveman99, line 46: `configTIMER_TASK_STACK_DEPTH` is 256 words on this core, 1 KB, shared with Bluefruit's timer callbacks. `onTick -> rtttl::play -> nextnote -> tone -> applyConfiguration` is now the deepest chain on that stack, with a `uint32_t pins[4]` and several 64-bit locals. Likely fits, but one `uxTaskGetStackHighWaterMark(xTimerGetTimerDaemonTaskHandle())` reading on the T-Echo Plus would confirm it.

src/platform/nrf52/NRF52RtttlTicker.cpp
```diff
@@ -0,0 +1,82 @@
...
+    if (!timer)
+        LOG_ERROR("RTTTL timer unavailable, falling back to main-loop playback");
```
caveman99, line 50: Once creation fails this logs on every `begin()`, and `begin()` runs again on each nag restart. One-shot flag, or move the log to the caller.

src/platform/nrf52/NRF52RtttlTicker.cpp
```diff
@@ -0,0 +1,82 @@
...
+        rtttl::play();
+}
+
+void stop()
+{
+    if (!timer) {
+        rtttl::stop();
+        return;
+    }
+    xTimerStop(timer, pdMS_TO_TICKS(10));
```
caveman99: Return ignored here while `begin()` checks the symmetric `xTimerStart`. It does self-heal: `rtttl::stop()` clears `playing` and the next `onTick` calls `xTimerStop(timer, 0)`. The asymmetry still reads as an oversight; a `(void)` cast plus a one line comment would settle it.

src/modules/ExternalNotificationModule.cpp
```diff
@@ -172,10 +206,10 @@ int32_t ExternalNotificationModule::runOnce()
...
             if (rtttl::isPlaying()) {
-                rtttl::play();
+                pwmRtttlPump();
             } else if (isNagging && !Throttle::deadlinePassed(nagCycleCutoff)) {
                 // start the song again if we have time left
-                rtttl::begin(config.device.buzzer_gpio, rtttlConfig.ringtone);
+                pwmRtttlBegin(config.device.buzzer_gpio, rtttlConfig.ringtone);
             }
             // we need fast updates to play the RTTTL
             delay = EXT_NOTIFICATION_FAST_THREAD_MS;
```
caveman99, line 218: On nRF52 the module thread no longer sequences anything on this path, it only polls for song end and nag restart, so the 25 ms rate is now dead weight. Not a bug, and fine to leave if you want the diff scoped to one concern.

caveman99 (commented):
Reviewed the premise against the toolchain, it holds. `tone()` in `framework-arduinoadafruitnrf52/cores/nRF5/Tone.cpp` configures PWM2 and returns, no `delay()`, no interrupt handler. `configUSE_TIMERS 1`, `configTIMER_TASK_PRIORITY 2` against a prio 1 loop task, so the daemon does preempt the blocking repaint.

Also verified: all four mutating `rtttl::` call sites are wrapped, the two remaining raw `isPlaying()` reads are a single bool; `onTick` uses a zero timeout take so the daemon never blocks, and every `stopNow()` caller is task context, so `portMAX_DELAY` in `begin()`/`stop()` is safe; stop and tick interleave correctly in both orders; `arduino_base` strips `-<platform/>` and `variants/nrf52840/nrf52.ini:63` adds `+<platform/nrf52/>`, so nRF54L15/Zephyr never compiles the new file and `ARCH_NRF52` is undefined there.

Six points inline. Only the first is a correctness issue.

src/platform/nrf52/NRF52RtttlTicker.cpp
Ixitxachitl: Took your one-liner (38d845bd3), scoped to the tone() loop in a72ae0bff. rtttl::isPlaying() only tracks the PWM sequencer, and runOnce() starts the I2S and PWM ringtone paths from independent conditions — only triggerBuzzerOutput() treats them as exclusive — so above the I2S branches it could silence I2S system tones on a board where PWM rtttl happened to be running.

NonBlockingRtttl.h is only a lib_dep under arduino_base, and ExternalNotificationModule.h carries a stub rtttl class for Portduino/STM32WL. Rather than pull that header into buzz.cpp I mirrored the guard, so the include and the check are both behind !ARCH_PORTDUINO && !ARCH_STM32WL.

src/platform/nrf52/NRF52RtttlTicker.cpp
Ixitxachitl: 5 ms now, with the wake-rate reasoning and the pdMS_TO_TICKS() rounding note in the comment (38d845bd3).

Added a static_assert(kTickMs * configTICK_RATE_HZ >= 1000) alongside it, so lowering the period or landing on a slower tick rate is a build error rather than a silent fall back to polling.

src/platform/nrf52/NRF52RtttlTicker.cpp
Ixitxachitl: Measured on a T-Echo Plus: uxTaskGetStackHighWaterMark(xTimerGetTimerDaemonTaskHandle()) logged once after a ringtone finishes reports 195 of 256 words free, so onTick -> rtttl::play -> nextnote -> tone -> applyConfiguration peaks at 61 words / 244 bytes — with Bluefruit's timer callbacks sharing the stack.

Dropped the probe and left the figure as a comment on onTick (6408bcd12) so anything that deepens that chain later has a baseline.

src/platform/nrf52/NRF52RtttlTicker.cpp
Ixitxachitl: One-shot flag in ensureInit() (38d845bd3). Kept it there rather than at the caller so the fallback path in begin() stays a single branch.

src/platform/nrf52/NRF52RtttlTicker.cpp
Ixitxachitl: (void) cast plus a note on why a rejected stop self-heals (38d845bd3). Chasing that further turned up a real gap: timerRunning going false doesn't prove the timer stopped, so a rejected stop followed by a rejected start left pump() polling from the main thread while onTick() was still firing — two threads in rtttl::play() and tone(). pump() now takes the same mutex non-blocking (f6a54c650).

src/modules/ExternalNotificationModule.cpp
Ixitxachitl: Left as is -- the 25 ms poll is what bounds the nag restart. At EXT_NOTIFICATION_MODULE_OUTPUT_MS (1000 ms) a nagging ringtone would get up to a second of silence between repeats. Worth revisiting if the nag path gets restructured, but it'd need its own timing rather than falling back to the slow rate.

src/buzz/buzz.cpp
```diff
@@ -150,6 +155,10 @@ void playTones(const ToneDuration *tone_durations, int size)
         config.device.buzzer_gpio = PIN_BUZZER;
 #endif
     if (config.device.buzzer_gpio) {
+#ifdef HAS_PWM_RTTTL
+        if (rtttl::isPlaying())
+            return; // a notification ringtone owns the PWM, don't reprogram it mid-note
+#endif
```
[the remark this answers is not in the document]
Ixitxachitl: Skipped, every playTones() caller runs on the main task (BuzzerFeedbackThread is an InputBroker observer; the rest are OSThreads), as does NRF52RtttlTicker::begin(), so a ringtone can't start between the check and the tone() call. onTick() only reaches tone() while isPlaying() is true, which is the case that already returned early; if the song ends mid-check the pending tick sees !isPlaying and just stops the timer. A mutex there would guard a window the callback can't be in.

Ixitxachitl:
I want to do this differently and include esp32 with it, I'm closing it until I can redo it.

PR #11621, opened 2026-08-26
Route waypoint expiry through waypointIsActive instead of a raw getTime compare

caveman99:
The #10920 rewrite reimplemented expiry inline, leaving waypointIsActive() and its 8 tests orphaned with no production caller, and falling back to getTime(), which counts from boot until the RTC is set.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved waypoint expiration handling when device time is unavailable or incomplete.
  - Prevented uptime values from being misinterpreted as absolute timestamps.
  - More accurately identifies active, expired, future-dated, and sentinel waypoints.
  - Improved waypoint availability across clock conditions and device restarts.

- **Tests**
  - Added coverage for waypoint expiration using explicit timestamps, including past, future, and sentinel values.
  - Added coverage for rejecting expired waypoints when packet receive time is unavailable.

### DIFF
--- a/src/WaypointStore.cpp
+++ b/src/WaypointStore.cpp
@@ -202,7 +202,9 @@ bool WaypointStore::addFromPacket(const meshtastic_MeshPacket &packet, bool loca
     if (stored)
         *stored = entry;
 
-    if (isExpired(entry, entry.receivedTime)) {
+    // rx_time holds uptime, not an epoch, when has_rx_time is false; pass 0 so isExpired() resolves
+    // the clock itself rather than comparing an expiry against seconds since boot.
+    if (isExpired(entry, packet.has_rx_time ? packet.rx_time : 0)) {
         // Respect the lock: only the node a waypoint is locked to may delete it on our device.
         // An unauthorized deletion attempt is ignored entirely, rather than applied locally.
         for (const auto &storedEntry : waypoints) {
--- a/src/WaypointStore.cpp
+++ b/src/WaypointStore.cpp
@@ -229,11 +229,7 @@ bool WaypointStore::addFromPacket(const meshtastic_MeshPacket &packet, bool loca
 
 bool WaypointStore::purgeExpired(uint32_t now)
 {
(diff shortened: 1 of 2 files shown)


caveman99:
Fixed. purgeExpired() no longer normalises the clock itself; it passes now straight to isExpired(), which owns the policy. That also restores the expire == 1 deletes it was skipping via the early return on a zero clock.

Swept the rest of the file rather than just the reported line: the getTime() at line 199 is a received-timestamp rather than expiry policy and stays, and all seven external isExpired callers use the default now = 0, so they already route through the shared policy. purgeExpired was the only bypass.

9/9 native tests pass.

thebentern (approved):

HarukiToreda (approved):

Xaositek (approved):

caveman99:
Fixed. addFromPacket() now passes packet.has_rx_time ? packet.rx_time : 0 to isExpired(), so the shared clock policy resolves the time instead of comparing an expiry against seconds of uptime. Verified the premise in Router::computeRxTimeStamp(): it returns Time::getUptimeSecs() when has_rx_time is false.

Added test_packet_without_rx_time_still_expires, which feeds a September 2020 expiry with has_rx_time false and rx_time set to an uptime-like 300, and asserts the waypoint is not stored. 10/10 native tests pass.

Left entry.receivedTime alone deliberately: it has the same rx_time-or-getTime() shape, but it is a stored timestamp rather than expiry policy, so it is a separate concern from this PR.

caveman99:
Declining, consistently with the same nitpick on #11600.

Test functions in this repo are snake_case by overwhelming convention: 855 of them across test/, against 7 in camelCase. All 10 functions in this file follow it. Renaming only the newest one would make the file internally inconsistent, and renaming all of them would diverge from every neighbouring suite for no functional gain.

Happy to switch if a maintainer wants the guideline applied to test code specifically, but that is a repo-wide decision rather than something to change in one new function.

PR #11630, opened 2026-08-27
fix(rp2040): feed hardware watchdog during long GPS ACK waits

Simplycissmus:
On RP2040 boards with a u-blox M10, `GPS::setup()` blocks the main loop for roughly 15 s (12 VALSET packets with 300 ms ACK waits plus 750 ms delays, and a final SAVE with a 2 s ACK wait). The RP2040 hardware watchdog (`watchdog_enable(8000)` in `platform/rp2xx0/main-rp2xx0.cpp`) is only fed from `loop()`, so the reset always lands inside the M10 config sequence and the device boot-loops before GPS setup can ever complete.

Fix: feed the watchdog inside the three `GPS::getACK()` wait loops through a small `ARCH_RP2040`-only helper that compiles to a no-op on all other platforms. Same pattern as the existing feeds in `src/mesh/NodeDB.cpp` and `src/mesh/eth/ethTlsApiServer.cpp`.

### Tested

Raspberry Pi Pico W (DIY variant) with a GEPRC GEP-M10 (u-blox M10 @ 115200). Before: "Rebooted by watchdog" loop, reset always during the M10 VALSET sequence, node never came up. After: "GNSS module config saved", stable operation with GPS lock, WiFi, and LoRa RX/TX, no further watchdog resets. No Heltec/RAK/T-Beam hardware was available; the change is compile-time inert off RP2040.

- [ ] Tested on Heltec hardware
- [ ] Tested on RAK hardware
- [x] Tested on Other hardware: Raspberry Pi Pico W (DIY) + u-blox M10

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the device watchdog from expiring during lengthy GPS setup and configuration waits on RP2040 hardware.
  * Improved GPS lock and search-failure detection by consistently tracking valid fixes throughout the search process.
  * Preserved existing behavior on other supported platforms.


jp-bennett (approved):

jp-bennett:
Looks like a good fix. The only missing thing is the CLA sign-off.

Simplycissmus:
recheck

PR #11638, opened 2026-08-27
fix(nrf54l15): give the LittleFS File shim the readBytes() Stream API

cvaldess:
`nrf54l15dk` does not build on current `develop`.

#10920 added `WaypointStore::loadFromFlash()`, which reads the store header and each record through `File::readBytes()` (`src/WaypointStore.cpp`, lines 313-314 and 326). On nRF54L15, `File` is the Zephyr/LittleFS-backed shim in `src/platform/nrf54l15/InternalFileSystem.h` — a plain class rather than an Arduino `Stream` subclass — and it never implemented that method, so the build stops with:

```
src/WaypointStore.cpp:313:19: error: 'class Adafruit_LittleFS_Namespace::File' has no member named 'readBytes'
src/WaypointStore.cpp:314:19: error: 'class Adafruit_LittleFS_Namespace::File' has no member named 'readBytes'
src/WaypointStore.cpp:326:31: error: 'class Adafruit_LittleFS_Namespace::File' has no member named 'readBytes'
```

This adds `readBytes()` on top of the shim's existing `read()`, looping until `fs_read()` stops making progress. A short read is legal there, and the caller already treats a count below the requested size as end-of-file, so returning the number of bytes actually delivered matches what the Arduino `Stream` version does.

Verified: `pio run -e nrf54l15dk` goes from FAILED to SUCCESS (RAM 46.8%, flash 36.8%). The change is confined to the nRF54L15 platform shim, so no other target is affected.

## Summary by CodeRabbit

- **New Features**
  - Added Arduino Stream-compatible `readBytes` support for files.
  - Supports reading into both character and byte buffers, up to the requested length or until no more data is available.


cvaldess:

PR #11640, opened 2026-08-28
fix(nodedb): seed and persist the USERPREFS_FIXED_GPS position

Simplycissmus:
The `USERPREFS_FIXED_GPS` block in `NodeDB.cpp` was broken four ways. It did not compile: the satellite-map insert uses `info->num` with `info` undeclared in that scope. It called `nodeDB->setLocalPosition()` from inside the NodeDB constructor, where the global `nodeDB` pointer is still null — a store through null `this`, which panics with a store-access fault on ESP32-C6 and segfaults the native build. It ran after the boot save decision, so even a successful seed was never written to flash. And it was gated on `myNodeInfo.reboot_count == 1`, an ESP32-only NVS counter that is >1 after any non-erase reflash, so the path effectively never ran on real devices.

This change moves the feature to where the other `USERPREFS_` defaults live: `installDefaultConfig()` turns `config.position.fixed_position` on when the defines are present, and the constructor's existing "restore persisted position" block seeds the compiled-in coordinates via `updatePosition()` when `fixed_position` is on but no self position is stored, ORing `SEGMENT_CONFIG | SEGMENT_NODEDATABASE` into `saveWhat` (the `nodePositions` satellite map is not covered by the nodeDatabase CRC compare). A position set or removed from the app wins on later boots.

Tests: new native suite `test_fixed_gps_userprefs`, run in two flavours. The plain `coverage` env (no defines) covers 3 cases: no position appears from nowhere, an app-set fixed position is restored after reboot, remove stays removed. The new `coverage-fixed-gps` env (defines set) covers 5 cases: first boot seeds and persists, reboot restores, app override survives reboot, a lost self position is re-seeded, remove stays removed. On unfixed code the seeded suite segfaults in the first `new NodeDB()`. CI runs it as its own step, mirroring the existing coverage steps.

Tested on hardware: Seeed XIAO ESP32-C6 DIY node. After erase+flash, the first boot seeded and persisted the position; the next boot logged "Restored fixed position", and the CLI shows `fixedPosition true` with the compiled-in coordinates. No Heltec/RAK hardware available.

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Tested on Other: Seeed XIAO ESP32-C6 DIY (LLCC68).

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a fixed GPS position at build time.
  * Fixed GPS coordinates are automatically initialized and persisted when no position is already stored.
  * Existing position updates and removals continue to override the configured default.

* **Tests**
  * Added comprehensive coverage for first-boot setup, reboot restoration, overrides, reseeding, and builds without fixed GPS preferences.

### DIFF
--- a/.github/workflows/test_native.yml
+++ b/.github/workflows/test_native.yml
@@ -472,6 +472,17 @@ jobs:
           ./bin/check-test-attribution.py --label coverage-channel-table \
             --expect "$expect" channel-table-testreport.xml
 
+      - name: Build-time fixed position tests
+        run: platformio test -e coverage-fixed-gps -v --junit-output-path fixed-gps-testreport.xml
+
+      - name: Verify the fixed-gps suite ran its own tests
+        run: |
+          set -euo pipefail
+          expect=$(python3 -c "from platformio.project.config import ProjectConfig; \
+            print(' '.join(ProjectConfig().get('env:coverage-fixed-gps', 'test_filter', [])))")
+          ./bin/check-test-attribution.py --label coverage-fixed-gps \
+            --expect "$expect" fixed-gps-testreport.xml
+
       - name: Save test results
         if: always() # run this step even if previous step failed
         uses: actions/upload-artifact@v7
--- a/src/mesh/NodeDB.cpp
+++ b/src/mesh/NodeDB.cpp
@@ -647,6 +647,27 @@ NodeDB::NodeDB()
             setLocalPosition(TypeConversions::ConvertToPosition(fixedPos));
             LOG_INFO("Restored fixed position to localPosition: lat=%d lon=%d", fixedPos.latitude_i, fixedPos.longitude_i);
         }
+#if defined(USERPREFS_FIXED_GPS) && defined(USERPREFS_FIXED_GPS_LAT) && defined(USERPREFS_FIXED_GPS_LON)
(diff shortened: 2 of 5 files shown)


Simplycissmus:
#11825 has since landed the persistence half of this, and this branch is now conflicted with develop.

What is still not upstream is the gate. `NodeDB.cpp:691` reads `if (myNodeInfo.reboot_count == 1)`, and `reboot_count` is only assigned under `#ifdef ARCH_ESP32` (`NodeDB.cpp:559-565`). On nRF52, RP2040, STM32WL and portduino it stays 0, so the whole `USERPREFS_FIXED_GPS` block never runs there — the feature is ESP32-only in practice, whatever a vendor puts in `userPrefs.jsonc`.

The other part that is not upstream is the native test and the `coverage-fixed-gps` environment, which is the `coverage-*` pattern #11825 asks for in its own reviewer notes.

As it stands this PR is five files and conflicted, which is not worth anyone's review time. I can either re-cut it onto current develop with the persistence half dropped, or close it and put the gate and the coverage environment up as a small separate PR. Whichever you prefer — happy to leave it closed too if the gate is intentional.

Issue #3535, opened 2024-04-02
[Bug]: Pi Pico W - unreliable WiFi 


madeofstown:
### Category

WiFi

### Hardware

Raspberry Pi Pico (W)

### Firmware Version

2.3.3

### Description

Wifi seems to stop working and I am unable to access the device until I reboot the Pico or connect a serial cable. Not sure what could be causing this, it seems to be irregular. Would being in range of an MQTT router just overload the Pico hardware?

GUVWAF:
Is it still responding over LoRa when Wi-Fi stops working?

Also, can you post your configuration (e.g. with the CLI `meshtastic --info` after removing personal details)? Especially regarding MQTT: are you using the default MQTT server with default channel LongFast? In that case, indeed it might get overloaded.

madeofstown:
Yeah, it appears in the node list of my other devices and I am able to successfully "Request Position" and have it update. I'm just unable to log in over the network.

GUVWAF:
Might be related to #3031 then. 
I would advise to try another channel name or a different MQTT root topic and see if the issue persists.

madeofstown:
I can confirm that changing from the default root topic has fixed the problem. I guess all those messages were too much for the pico to handle.

GUVWAF:
Thanks for confirming. I'm afraid there's not much we can do about it, so I'm closing this.

PR #11641, opened 2026-08-28
fix(rp2040): run the blocking CYW43 WiFi join in a task and retry failed joins

Simplycissmus:
On the RP2040 (Pico W), joining WiFi resets the node before the join can complete. Meshtastic builds the RP2040 target with FreeRTOS, and under FreeRTOS the arduino-pico CYW43 shim holds the caller for several seconds even in `WiFi.beginNoBlock()` (measured 4.4 s in a bare FreeRTOS sketch, more than 8 s inside Meshtastic with the radio and GPS threads running). The RP2040 hardware watchdog (`watchdog_enable(8000)` in `platform/rp2xx0/main-rp2xx0.cpp`) is only fed from `loop()`, so the reset lands about 8 s after "Reconnecting to WiFi access point ..." and the node never associates. A second, smaller defect: on RP2040 a failed first join was never retried, because `needReconnect` was only re-armed once `APStartupComplete` is set, which requires a successful connection.

Fix: on `ARCH_RP2040` the join runs in a short-lived FreeRTOS task (`startWifiJoin()`); the existing reconnect poll keeps running in the main loop and re-arms a new join if none has come up within 30 s. Other platforms are unchanged.

### Tested

Raspberry Pi Pico W (DIY variant with LLCC68), on hardware, on the branch as it stands.

The original two commits: before, a watchdog reset exactly 8 s after every "Reconnecting" line and the node never appeared on the AP. After, "Obtained IP address", mDNS, NTP and the TCP API all come up, joins survive AP roaming gaps, and there are no watchdog resets. Related report: #3535 (closed unresolved).

The four commits added after review, re-tested on the same board:

- Cold start joins and gets an IP in 129 s. Pinning the join to core 0 changes nothing there.
- 10 minutes connected: rebootCount 0, no spurious "WiFi join stuck in the driver".
- The retry is still live after a failure: rejoin 22 s after the AP let the node back in. Verified by outcome, not by measuring the 35 s cadence.
- 6.5 minutes with the node's MAC blocked at the AP: zero reboots. A normally failed join returns from beginNoBlock and releases the claim, so the wedge detector never arms - which is the failure mode worth being sure about, since a false arm would reboot the node about every 65 s.
- The wedge path cannot be provoked naturally, so it was exercised with a temporary harness (shortened timeout, claim held): exactly one error line per cycle and the scheduled reboot about 5 s later, over 6 cycles. The harness is not in this branch.

One trap worth naming for anyone reproducing this: with a serial monitor attached the node looks like it is in a boot loop. That is the RP2040 CDC config dump starving the 8 s watchdog, not this change. Check RP2040 WiFi behaviour over the network rather than over USB serial.

No Heltec/RAK/T-Beam hardware was available; the change is compile-time inert off RP2040.

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Tested on Other: Raspberry Pi Pico W (DIY) + LLCC68.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Wi‑Fi connection reliability on RP2040 devices.
  * Prevented Wi‑Fi association attempts from triggering the hardware watchdog.
  * Added more resilient retry handling when connections do not establish within the expected timeframe.
  * Improved recovery when a Wi‑Fi connection attempt becomes unresponsive, including automatic device restart after 60 seconds.
  * Other supported platforms continue using the existing Wi‑Fi connection behavior.

### DIFF
--- a/src/mesh/wifi/WiFiAPClient.cpp
+++ b/src/mesh/wifi/WiFiAPClient.cpp
@@ -29,11 +29,13 @@
 #include <esp_wifi.h>
 static void WiFiEvent(WiFiEvent_t event);
 #elif defined(ARCH_RP2040)
+#include <FreeRTOS.h>
 #include <SimpleMDNS.h>
+#include <task.h>
 #endif
 
-#ifndef DISABLE_NTP
 #include "Throttle.h"
+#ifndef DISABLE_NTP
 #include <NTPClient.h>
 #endif
 
@@ -61,6 +63,31 @@ unsigned long lastrun_ntp = 0;
 
 bool needReconnect = true;   // If we create our reconnector, run it once at the beginning
 bool isReconnecting = false; // If we are currently reconnecting
+
+#ifdef ARCH_RP2040
+// On the CYW43 under FreeRTOS even WiFi.beginNoBlock() holds the caller for several seconds (up to the WiFi timeout,
+// 15 s); from the main loop that trips the 8 s hardware watchdog. Join from a short-lived task and poll for the link.
+static TaskHandle_t wifiJoinTask = nullptr;
+
+static void wifiJoinTaskFn(void *)
+{
+    const char *psk = config.network.wifi_psk[0] ? config.network.wifi_psk : NULL;
+    WiFi.beginNoBlock(config.network.wifi_ssid, psk);
+    wifiJoinTask = nullptr;
+    vTaskDelete(NULL);
+}
+
+static bool startWifiJoin()
+{
+    if (wifiJoinTask)
+        return true; // previous join still in progress
+    if (xTaskCreate(wifiJoinTaskFn, "wifijoin", 1536, NULL, uxTaskPriorityGet(NULL), &wifiJoinTask) != pdPASS) {
+        LOG_ERROR("Could not start WiFi join task");
+        return false;
+    }
+    return true;
+}
+#endif
 #if defined(USE_WS5500) || defined(USE_CH390D)
 static volatile bool ethNetworkConnectedPending = false;
 #endif
@@ -281,7 +308,11 @@ static int32_t reconnectWiFi()
                 WiFi.useStaticBuffers(true);
                 WiFi.mode(WIFI_STA);
 #endif
+#ifdef ARCH_RP2040
+                startWifiJoin();
+#else
                 WiFi.begin(wifiName, wifiPsw);
+#endif
             }
             isReconnecting = false;
             wifiReconnectPending = false;
@@ -311,7 +342,9 @@ static int32_t reconnectWiFi()
 
     if (config.network.wifi_enabled && !WiFi.isConnected()) {
 #ifdef ARCH_RP2040 // (ESP32 handles this in WiFiEvent)
-        needReconnect = APStartupComplete;
+        // Lost the link, or a join that has not come up within 30 s: start the join over once the join task is done.
+        needReconnect = !wifiJoinTask && (APStartupComplete || (!isReconnecting && !Throttle::isWithinTimespanMs(
+                                                                                     wifiReconnectStartMillis, 30000)));
 #endif
         return 1000; // check once per second
     } else {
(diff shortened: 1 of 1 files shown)


moomoo2402 (commented):
Please fix the data race issue on lines 70 to 84 where, "wifiJoinTaskFn() clears wifiJoinTask, while startWifiJoin() and reconnectWiFi() read it. These accesses have no concurrency::Lock/LockGuard or atomic protocol, so the active-task guard has a C++ data race and cannot reliably prevent overlapping"
and please fix lines 346 and 347 by not treating 0 as a valid timestamp.

moomoo2402 (changes requested):
!wifiJoinTask &&
    (APStartupComplete ||
     (!isReconnecting &&
      wifiReconnectStartMillis != 0 &&
      Throttle::hasElapsed(wifiReconnectStartMillis, 30000)));
Lines 346-347 
I think this is just the simplest solution.

Simplycissmus:
Thanks for the review - both points are addressed, and the branch has now been tested on hardware.

**The guard.** It is no longer a task handle at all. `reconnectWiFi()` is a single `Periodic`, so it is the only writer of `true` and it claims the flag before the task exists; the task clears it as its last act, and those two writers are mutually exclusive. Nothing needs the handle any more, so the create no longer asks for one. Load and store only: on ARMv6-M an `exchange()` lowers to a `__atomic_exchange_1` call, and with plain load/store the translation unit ends up with no undefined `__atomic_*` symbol at all.

**The deadline.** Guarding the read alone would have been a trade rather than a fix. The only way to reach that line with a zero in `wifiReconnectStartMillis` is `millis()` returning exactly 0 at the arm site, once per 49.7 day wrap - and a bare `!= 0` turns that instant from one premature retry into no retry at all, because the retry is what would have re-armed the field. So the arm site is made incapable of writing the sentinel first, the way `GPS.cpp` already does for `fixHoldEnds`, and only then does the read test it.

Two things the review did not reach, both fixed here:

- The join task was the only unpinned task in the build. arduino-pico pins every task it creates - the Arduino loop and the LWIP thread to core 0, the second sketch task to core 1 - and the CYW43 shim is written for the core-0 LWIP thread. Its `get_core_num()` assertions are `NDEBUG`-only, so a join scheduled onto core 1 would not have trapped, it would have gone wrong quietly. Now `xTaskCreateAffinitySet(..., 1u << 0, ...)`.
- Moving the join off the main loop had quietly removed a recovery. `CYW43::begin()` does not always return: past its own 15 s timeout it can wait for the link in an untimed `while (true)`. Inline that starved the 8 s watchdog and the node rebooted into a working state; in a task the loop keeps feeding the watchdog, the claim is never released, and every retry in this file is gated on that claim - so the node would have sat off the network indefinitely with nothing logged. The claim is timed now, and a wedge asks for a reboot through `rebootAtMsec`.

Re-tested on a Pico W with an LLCC68: clean join (IP 129 s from cold start), 10 minutes connected with `rebootCount` 0 and no spurious "stuck" line, and the retry still alive after blocking and unblocking the node's MAC at the AP. The case I most wanted to be sure of is the inverse one: 6.5 minutes blocked produced zero reboots, so an ordinary failed join does not arm the wedge detector. The wedge path itself cannot be provoked naturally, so it was exercised with a temporary harness; that is stated as such in the description, together with a caveat about reading RP2040 behaviour over a serial monitor.

If you would rather keep this PR to the two points you raised, say so and I will split the last commit out into its own.

moomoo2402:
Dont particularly mind, this is my first time doing this!

Simplycissmus:
For a first one that was a good review - both points were real, and the atomic one changed the guard rather than just its wording. So nothing to split: the wedge commit stays, and what is on the branch now is exactly what ran on the Pico W.

If it reads right to you, would you take another look and update the review? That is the only thing sitting on the PR apart from the checks.

moomoo2402:
sure will have another look!

moomoo2402 (approved):
No obvious mistakes or errors!

Simplycissmus:
Correction to the test notes above. `rebootCount` is only populated under `ARCH_ESP32` (`NodeDB.cpp:559-565`) and stays 0 on RP2040, so the "rebootCount 0" I gave for the 10-minute connected run is not evidence of anything. The other results stand: the IP after 129 s from cold start, the rejoin 22 s after unblocking, and the wedge path, where each reboot was seen as the "Rebooting" log line. The no-reboot claims for the 10-minute run and the 6.5-minute blocked-MAC run are therefore unverified.

PR #11643, opened 2026-08-28
Update applyLoraRegion to enable TX on set

### LINKED ISSUE #11636: BaseUI region chooser does not re-enable tx_enabled, leaving recovered nodes unable to transmit
### Summary

Setting the LoRa region from the **BaseUI** on-screen region chooser does not re-enable `config.lora.tx_enabled`. On a node whose region was previously cleared to `UNSET` — which force-disables TX — picking a region on the device leaves it permanently unable to transmit, with no indication beyond the screen's "transmit disabled" text.

The phone/CLI path and InkHUD both re-enable TX here. BaseUI is the only one that doesn't.

### Where

`applyLoraRegion()` in `src/graphics/draw/MenuHandler.cpp` (line 230 on `develop`) never touches `tx_enabled`.

| Path | Re-enables TX on UNSET → region |
| --- | --- |
| Phone app / CLI — `AdminModule.cpp:1047` | Yes — *"new region is valid and we're coming from an unset region, so enable tx"* |
| InkHUD — `MenuApplet.cpp:336` | Yes (unconditional — see note below) |
| **BaseUI — `MenuHandler.cpp` `applyLoraRegion()`** | **No** |

Going *to* `UNSET` forces `tx_enabled = false` in `AdminModule.cpp:1051`, `NodeDB.cpp:2423` (lockdown) and `NodeDB.cpp:2574` (degraded boot), so the flag is genuinely off by the time the user reaches the chooser.

`tx_enabled` is a gate independent of the region gate, in `RadioLibInterface.cpp:157`:

```c
if (config.lora.region != meshtastic_Config_LoRaConfig_RegionCode_UNSET) {   // region gate
    if (disabled || !config.lora.tx_enabled) {                               // tx_enabled gate
        LOG_WARN("send - !config.lora.tx_enabled");
```

Selecting a region lifts the first gate but not the second, so the node stays silent.

### Who is affected

**Not a fresh flash.** `installDefaultConfig()` sets `tx_enabled = true`, with a comment that anticipates this: *"FIXME: maybe false in the future, and setting region to enable it. (unset region forces it off)"*.

It affects nodes that reach `UNSET` **after** having a region, and are then recovered from the device screen:

- region unset via phone app / CLI / admin message
- lockdown mode (`NodeDB.cpp:2423`)
- degraded boot after a config decode failure (`NodeDB.cpp:2574`)
- any config reinstall that leaves region `UNSET` while `tx_enabled` is already false

### Reproduction

Verified on hardware (Heltec Mesh Node T096, nRF52, `2.8.0`):

1. Start from a working node — region `US`, `tx_enabled: true`.
2. Set `lora.region = UNSET` (phone/CLI) and reboot. `tx_enabled` is now `false`.
3. On the device, pick **US** in the region chooser.
4. Read the config back:

```
region: US
tx_enabled: <absent — false>
```

Region applied, TX still disabled. The screen's "transmit disabled" indicator is correct, not a rendering artifact.

### Suggested fix

Mirror `AdminModule`'s **conditional** form in `applyLoraRegion()`, sampling the region before it is overwritten:

```c
const bool regionWasUnset = config.lora.region == meshtastic_Config_LoRaConfig_RegionCode_UNSET;
...
if (regionWasUnset)
    config.lora.tx_enabled = true;
```

It should be conditional rather than unconditional, because BaseUI has its own **Transmit Enabled** menu item (`loraMenu` → `txEnabledMenu`). An unconditional assignment would silently re-enable TX every time the user changes region, discarding a deliberate choice.

### Side note

`MenuApplet.cpp:336` (InkHUD) sets `tx_enabled = true` unconditionally on every region change, so it has that clobbering behaviour today. Probably worth tightening to the same condition, but it is a separate defect from this one.

Xaositek:
- Update applyLoraRegion to enable TX on set
- Don't enable TX when owner.is_licensed is set

Tested on Wio Tracker L1

Steps to reproduce for testing:

1. Update firmware
2. meshtastic --set lora.tx_enabled FALSE 
3. meshtastic --set lora.region UNSET --set lora.modem_preset 0
4. Pick new region and verify region is set and TX is enabled.

## New Features

* Applying a configured LoRa region now automatically enables LoRa transmission when transmission is disabled.
* A warning is displayed before the transmission setting changes.
* LoRa transmission remains unchanged for licensed users, including when applying a non-ham region.
* LoRa transmission remains unchanged when applying a region in ham mode.

### DIFF
--- a/src/graphics/draw/MenuHandler.cpp
+++ b/src/graphics/draw/MenuHandler.cpp
@@ -272,7 +272,7 @@ static void applyLoraRegion(meshtastic_Config_LoRaConfig_RegionCode region, bool
     if (gps != nullptr && !gps->isEnabled() && config.position.gps_mode == meshtastic_Config_PositionConfig_GpsMode_ENABLED)
         gps->enable();
 #endif
-    if (config.lora.region != meshtastic_Config_LoRaConfig_RegionCode_UNSET && !config.lora.tx_enabled) {
+    if (config.lora.region != meshtastic_Config_LoRaConfig_RegionCode_UNSET && !config.lora.tx_enabled && !isHam) {
         LOG_WARN("Setting config.lora.tx_enabled to true");
         config.lora.tx_enabled = true;
     }
--- a/src/graphics/draw/MenuHandler.cpp
+++ b/src/graphics/draw/MenuHandler.cpp
@@ -272,6 +272,10 @@ static void applyLoraRegion(meshtastic_Config_LoRaConfig_RegionCode region, bool
     if (gps != nullptr && !gps->isEnabled() && config.position.gps_mode == meshtastic_Config_PositionConfig_GpsMode_ENABLED)
         gps->enable();
 #endif
+    if (config.lora.region != meshtastic_Config_LoRaConfig_RegionCode_UNSET && !config.lora.tx_enabled) {
+        LOG_WARN("Setting config.lora.tx_enabled to true");
+        config.lora.tx_enabled = true;
+    }
     service->reloadConfig(changes);
 }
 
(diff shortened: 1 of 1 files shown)


thebentern (approved):

vidplace7:
This will likely result in TX_ENABLED when `N0CALL` is set for ham mode.
@Xaositek check what happens when you select an amateur-only region.

PR #11652, opened 2026-08-29
fix(transmithistory): atomic saves, and no flash writes from the RX path

thebentern:
## Why

Two defects in TransmitHistory persistence, both observed on hardware:

1. **Crash during save destroys the history file.** `saveToDisk()` removed the old file before writing the new one, so any reset in that window (watchdog, panic, power loss) lost the entire transmit history. Captured in the field: a Guru Meditation landed inside this window and the node came back with `no history file found`.
2. **The flash write runs inside packet processing.** Modules call `setLastSentToMesh()` from their send paths — i.e. while the Router is handling an RX — and the first call after boot performed a synchronous LittleFS write right there, stalling core 0 with the flash cache disabled in the middle of packet handling.

## What

- `saveToDisk()` now writes through `SafeFile` (tmp file → hash readback → atomic rename), same pattern as MessageStore/WaypointStore. A crash mid-save leaves the previous file intact. The file is ≤118 bytes, so `fullAtomic` costs nothing.
- `setLastSentToMesh()` only marks the history dirty. The new `flushIfDue()` runs from the main loop — outside packet processing — with unchanged cadence: immediate for the first change after boot (crash-reboot loops still can't dodge persistence), then at most once per `SAVE_INTERVAL_MS`. The direct pre-deep-sleep flush in `sleep.cpp` is unchanged.

## Tests

- New `test_flushIfDue_persists_pending_changes` covers the deferred path end-to-end through SafeFile persistence and reload.
- `test_transmit_history`: 15/15 pass (Docker native run).
- `tbeam` target build: SUCCESS.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transmit-history saving reliability through verified, atomic writes.
  * Prevented frequent flash writes by deferring and throttling persistence.
  * Preserved pending history changes across reloads and restarts.

* **Tests**
  * Added coverage confirming pending transmit history is saved and restored correctly.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1474,6 +1474,13 @@ void loop()
 #endif
     power->powerCommandsCheck();
 
+    // Persist transmit-history changes here, outside packet processing:
+    // setLastSentToMesh() only marks state dirty, so the flash write (which
+    // stalls core 0 with the cache disabled) never runs inside Router RX
+    // handling. Throttled internally to once per 5 minutes.
+    if (transmitHistory)
+        transmitHistory->flushIfDue();
+
     if (RadioLibInterface::instance != nullptr) {
         static uint32_t lastRadioMissedIrqPoll;
         if (!Throttle::isWithinTimespanMs(lastRadioMissedIrqPoll, 1000)) {
--- a/src/mesh/TransmitHistory.cpp
+++ b/src/mesh/TransmitHistory.cpp
@@ -1,6 +1,7 @@
 #include "TransmitHistory.h"
 #include "FSCommon.h"
 #include "SPILock.h"
+#include "SafeFile.h"
 #include "gps/RTC.h"
 #include <Throttle.h>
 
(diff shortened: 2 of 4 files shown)


Simplycissmus:
Both defects are real and the shape of the fix looks right to me. I checked the one assumption the comment rests on: `SafeFile` does take `spiLock` itself - constructor, `close()` and `testReadback()` each guard their own filesystem work - so dropping the outer `spiLock->lock()` around the write is correct rather than merely convenient. Keeping the explicit `LockGuard` for the bare `FSCom.mkdir()` is the right split.

One thing I would look at before this lands.

**A failed save now retries every loop iteration.**

`flushIfDue()` only stamps `lastDiskSave` when `saveToDisk()` returns true, and `saveToDisk()` only clears `dirty` on success. So after a failed write both guards stay open: `dirty` is still true, `lastDiskSave` is unchanged, the throttle condition is still satisfied, and the next `loop()` pass tries the whole SafeFile sequence again - open `.tmp`, write, close, hash readback, rename. At loop frequency, indefinitely, for as long as the failure persists.

The first-save-after-boot case is the sharpest one: `lastDiskSave == 0` short-circuits the throttle entirely, so if that first write fails there is no interval to fall back on at all.

Before this change the same condition sat inside `setLastSentToMesh()`, so a retry could only happen on the next broadcast - minutes apart at best. Moving it into the loop turns a rare retry into a continuous one, which is the same stall the PR is removing from the RX path, only sustained.

Stamping the attempt rather than the success would keep the back-off:

```cpp
void TransmitHistory::flushIfDue()
{
    if (!dirty)
        return;
    if (lastDiskSave == 0 || !Throttle::isWithinTimespanMs(lastDiskSave, SAVE_INTERVAL_MS)) {
        lastDiskSave = millis(); // stamp the attempt, so a failing write backs off too
        saveToDisk();
    }
}
```

Two smaller things:

- The call sits after `power->powerCommandsCheck()` in `loop()`, so on the iteration where a scheduled reboot fires the flush does not run. `sleep.cpp` covers deep sleep, but the `rebootAtMsec` path has no equivalent - and durability across a reset is what motivates the PR. Moving the call above `powerCommandsCheck()` would close that without new code.
- `clear()` sets `lastDiskSave = 0` with the comment "so the next legit broadcast persists immediately". After this change it is the next loop iteration, not the next broadcast; worth rewording while the file is open.

PR #11654, opened 2026-08-29
Meshlink v1.3 board updates and heartbeat LED fixes

macvenez:
We fixed some stuff needed for V1.3 board revision and did some other changes that may benefit older revisions too.

We still need to make changes to `StatusLEDModule.cpp` otherwise our heartbeat LED is not light up when charging and then the hardware Watchdog timer doesn't get fed, causing the board to reboot every 30s.
These changes are needed for all devices, not just our

More [here on Discord](https://discord.com/channels/867578229534359593/919642584480112750/1542815173906010122)

Tested working on v1.2 and v1.3 board revisions (apart from WD Timer when charging, which as said needs to be addressed upstream)

## Summary by CodeRabbit

- **New Features**
  - Added support for the MeshLink user button.
  - Updated GPS, serial, and heartbeat LED hardware mappings.
  - Increased the e-ink display’s fast-refresh limit to 100 consecutive refreshes.

- **Bug Fixes**
  - Corrected battery current readings from the INA219 sensor.
  - Improved startup behavior for the white status LED.
  - Updated MeshLink board identification and e-ink refresh configuration.

### DIFF
--- a/variants/nrf52840/meshlink/platformio.ini
+++ b/variants/nrf52840/meshlink/platformio.ini
@@ -1,10 +1,20 @@
 ; MeshLink board developed by LoraItalia. NRF52840, eByte E22900M22S (Will also come with other frequencies), 25w MPPT solar charger (5v,12v,18v selectable), support for gps, buzzer, oled or e-ink display, 10 gpios, hardware watchdog
-; https://www.loraitalia.it
+; https://www.loraitalia.it/meshlink
+
 ; firmware for boards with or without oled display
 [env:meshlink]
+
+custom_meshtastic_hw_model = 87
+custom_meshtastic_hw_model_slug = MESHLINK
+custom_meshtastic_architecture = nrf52840
+custom_meshtastic_actively_supported = true
+custom_meshtastic_support_level = 1
+custom_meshtastic_display_name = MeshLink
+custom_meshtastic_tags = Loraitalia
+
 extends = nrf52840_base
 board = meshlink
-board_level = extra
+board_level = pr
 build_flags = ${nrf52840_base.build_flags}
   -I variants/nrf52840/meshlink
   -D MESHLINK
@@ -20,9 +30,18 @@ debug_tool = jlink
 ;upload_protocol = jlink
 
 [env:meshlink_eink]
+
+custom_meshtastic_hw_model = 87
+custom_meshtastic_hw_model_slug = MESHLINK
+custom_meshtastic_architecture = nrf52840
+custom_meshtastic_actively_supported = true
+custom_meshtastic_support_level = 1
+custom_meshtastic_display_name = MeshLink
+custom_meshtastic_tags = Loraitalia
+
 extends = nrf52840_base
 board = meshlink
-board_level = extra
+board_level = pr
 build_flags = ${nrf52840_base.build_flags}
   -I variants/nrf52840/meshlink
   -D MESHLINK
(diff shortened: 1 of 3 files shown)


variants/nrf52840/meshlink/platformio.ini
```diff
@@ -1,10 +1,20 @@
...
+custom_meshtastic_architecture = nrf52840
+custom_meshtastic_actively_supported = true
+custom_meshtastic_support_level = 1
+custom_meshtastic_display_name = MeshLink
+custom_meshtastic_tags = Loraitalia
+
 extends = nrf52840_base
 board = meshlink
-board_level = extra
+board_level = pr
```
vidplace7:
```suggestion
board_level = extra
```
extra was correct before, we do not need to build this board upon PR

variants/nrf52840/meshlink/platformio.ini
```diff
@@ -20,9 +30,18 @@ debug_tool = jlink
...
```
vidplace7:
```suggestion
board_level = extra
```
extra was correct before, we do not need to build this board upon PR

variants/nrf52840/meshlink/platformio.ini
```diff
@@ -20,9 +30,18 @@ debug_tool = jlink
...
 [env:meshlink_eink]
+
+custom_meshtastic_hw_model = 87
+custom_meshtastic_hw_model_slug = MESHLINK
+custom_meshtastic_architecture = nrf52840
+custom_meshtastic_actively_supported = true
+custom_meshtastic_support_level = 1
+custom_meshtastic_display_name = MeshLink
+custom_meshtastic_tags = Loraitalia
+
```
vidplace7:
```suggestion
```
this should be removed

variants/nrf52840/meshlink/platformio.ini
```diff
@@ -1,10 +1,20 @@
...
 [env:meshlink]
+
+custom_meshtastic_hw_model = 87
+custom_meshtastic_hw_model_slug = MESHLINK
+custom_meshtastic_architecture = nrf52840
+custom_meshtastic_actively_supported = true
+custom_meshtastic_support_level = 1
+custom_meshtastic_display_name = MeshLink
+custom_meshtastic_tags = Loraitalia
+
```
vidplace7:
```suggestion
```
this should be removed

vidplace7 (changes requested):
This is a DIY / community board. These metadata changes cannot be merged as-is.

variants/nrf52840/meshlink/platformio.ini
macvenez: Alright, thanks!

variants/nrf52840/meshlink/platformio.ini
```diff
@@ -35,12 +36,12 @@ build_flags = ${nrf52840_base.build_flags}
   -D EINK_WIDTH=250
   -D EINK_HEIGHT=122
   -D USE_EINK_DYNAMICDISPLAY            ; Enable Dynamic EInk
-  -D EINK_LIMIT_FASTREFRESH=5           ; How many consecutive fast-refreshes are permitted
+  -D EINK_LIMIT_FASTREFRESH=100           ; How many consecutive fast-refreshes are permitted
```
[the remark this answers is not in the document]
macvenez: This was tested on the field and confirmed to be the best setting

variants/nrf52840/meshlink/variant.h
```diff
@@ -25,17 +25,20 @@ extern "C" {
...
 // LEDs
-#define PIN_LED1 (24) // Built in white led for status
-#define LED_BLUE PIN_LED1
+#define LED_HEARTBEAT (24)     // Force strictly heartbeat on Pin 24
+#define LED_STATE_ON 0         // Active LOW
 
-#define LED_STATE_ON 0 // State when LED is lit
+#define HAS_HARDWARE_WATCHDOG
+#define HARDWARE_WATCHDOG_DONE (24)
+#define HARDWARE_WATCHDOG_TIMEOUT_MS (25 * 1000) // our hardware one expires in 30s so we feed it each 25s for safety
```
caveman99: I'd not use this and the heartbeat at the same time on the same GPIO. Just use the watchdog on - say - a 10 second timer. That will both feed the watchdog and occasonally blink the LED.

macvenez: I'd agree but the watchdog module blinks for 1ms and is nearly not visible. Furthermore it will leave that LED permanently on since we use inverted logic.
Mixing the two seemed to me to retain the past functionality with still having the watchdog not reset

Issue #11656, opened 2026-08-29
tlora-v1 still builds and works fine on current master - reconsider for release matrix


Tom-Neverwinter:
Related to #5210. That board (LILYGO TTGO LoRa32 V1.0) is marked `custom_meshtastic_actively_supported = false` and excluded from the official release build matrix, so no official binary has existed past 2.4.2.

I built current master (`v2.7.26.54e0d8d`) locally via `pio run -e tlora-v1` - succeeds cleanly, no code changes needed:
- Flash: 87.4% used (2,118,521 / 2,424,832 bytes)
- RAM: 34.2%

Flashed to physical hardware and confirmed: OLED display works normally, and the current Android app completes the config handshake without issue (on 2.4.2 it was stuck at module-config, which looks like a protocol-version mismatch with the modern app, not a hardware limitation).

Given it builds clean and works on real hardware with plenty of flash headroom, could this be added back to the release matrix? Happy to share the build if useful.

caveman99:
While this board (and many others) continue to work, the hardware is not up to the task of todays mesh networking, considering ressource constraints and sensitivity of the built-in radio chip (1st generation, no TCXO). If you want to use these boards, you'll have to compile yourself. 

We are supporting way over 100 board variants now, it's not feasible tu build them all for a release. And yes, we base these decision on demand and download statistics.

Issue #5210, opened 2024-10-31
[Bug]: firmware-tlora-v1 no longer build?


edegraaff:
### Category

Hardware Compatibility

### Hardware

T-Lora v1

### Firmware Version

firmware-esp32-2.5.8.6485f03.zip

### Description

Maybe i am wrong but in my notes from march this year is written use the firmware-tlora-v1-2.4.2.5b45303.bin firmware otherwise the screen is not working anymore. I still use 2 old the https://meshtastic.org/docs/hardware/devices/lilygo/lora/?t-lora=v1

In the notes in the download page i see this:
This board is still in production but for various reasons not recommended for new purchases or for unattended installations. Firmware support is phased out. If in doubt, choose the Lora V2.1-1.6 or Lora T3S3 board.

Question:
I know the T3S3 firmware did not work with the display in march 2024. What would be the appropiate approach? Could i safely upgrade to firmware-tlora-v2-1-1_6-2.5.8.6485f03.bin

Thanks

garthvh:
The v1 tlora is no longer suppor

edegraaff:
I  was affraid of this answer, but clear, and saved me a lot of time. I will go for the newer hardware. 
Thanks for the great project, 
Regards
Eelco

C4:
Is there any chance someone else is maintaining a fork for the display to work on these? I have a bunch of these that I just found and would like to get them working on the mesh. Is it just a pin mapping for the display that's needed?

edegraaff:
I have tested in the past the other builds and they do work, but indeed, the display won't work, the why i have no clue. In fact, i understand the development goes on, on the another hand, the tlora-v1 is ok, it does the job, especcially on locations where they are used as a router. I havend any facts and figures to drop the v1 and schred it. They stay usable to show people the possibillitys of meshtastic. Even for that i want to spread the wordt by giving demos or loan people my V1 to experiment.

garthvh:
It has an old radio and esp32 chip that has been difficult to keep working, a fork is not really necessary, but it has been broken for quite a while which is why it is no longer officially supported, I don't know that the pins on the screen changed, something in the esp32 toolchain seems to have caused issues.

edegraaff:
Thanks for your response, i will step over to Heltec the upcomming weeks. I am quite curious what new features it will bring me :-)

C4:
[...]

PR #11657, opened 2026-08-29
FU firmware update: restore tlora-v1 (TTGO LoRa32 V1.0) to release builds

Tom-Neverwinter:
Closes/relates to #11656 and #5210.

`tlora-v1` (LILYGO TTGO LoRa32 V1.0) is set to `board_level = extra`, which excludes it from official release builds. It still builds cleanly against current master with zero source changes:

- Flash: 87.4% used (2,118,521 / 2,424,832 bytes)
- RAM: 34.2%

Flashed to **four separate physical boards** and confirmed working on all of them: OLED display functions normally, and BLE config completes with the current app (previously stuck on old 2.4.2 builds due to a protocol-version gap, not a hardware issue).

This change only flips `board_level` from `extra` to `release` so it's included in the normal release matrix again. No code/pin changes.

## This isn't a new claim - it's already independently confirmed multiple times in #5210

The original reason given for dropping support (@garthvh): *"it has been broken for quite a while... something in the esp32 toolchain seems to have caused issues"* - not a hardware dead-end, just bitrot at some point in the past.

Since then, multiple people have independently built `tlora-v1` from source and confirmed it works fine, across several versions:

- **@C4** (2.5.14): *"Everything worked from there... running with display on my device."*
- **@okaestne** (2.7.15): *"the display works fine"*
- **@sysmanalex** (2.7.19, with GPS): *"IMHO board hw support - should be enabled back. all works ok."*
- **@RobertSasak** opened #9973 fixing the actual build break (missing `LED_BUILTIN`) - already merged (`a632d01`), which is why this now builds clean with zero further changes.

So the toolchain issue was already fixed upstream; this PR is just the remaining step of flipping the release-matrix flag back given that fix and the repeated independent confirmations across multiple units and multiple contributors.

## Summary by CodeRabbit

* **Chores**
  * Updated the TLORA V1 ESP32 build environment to use the release board level.

### DIFF
--- a/variants/esp32/tlora_v1/platformio.ini
+++ b/variants/esp32/tlora_v1/platformio.ini
@@ -6,7 +6,7 @@ custom_meshtastic_actively_supported = false
 custom_meshtastic_display_name = LILYGO T-LoRa V1
 custom_meshtastic_tags = LilyGo
 
-board_level = extra
+board_level = release
 extends = esp32_base
 board = ttgo-lora32-v1
 build_flags = 


caveman99:
While this board (and many others) continue to work, the hardware is not up to the task of todays mesh networking, considering ressource constraints and sensitivity of the built-in radio chip (1st generation, no TCXO). If you want to use these boards, you'll have to compile yourself. 

We are supporting way over 100 board variants now, it's not feasible tu build them all for a release. And yes, we base these decision on demand and download statistics.

PR #11673, opened 2026-08-31
Load optional modules dropped into src/modules/optional/

caveman99:
Adding a module today means editing `Modules.cpp` and `platformio.ini`, which is a merge conflict on every rebase for anyone carrying a module out of tree; `bin/optional-modules.py` instead scans `src/modules/optional/` for a directory `<Name>/` holding `<Name>.h`, and generates `$BUILD_DIR/OptionalModules.h` with an include and a `setup<Name>()` call for each one it finds, which `Modules.cpp` picks up through `__has_include`. The directory does not exist in a stock checkout, so a stock build generates a header that defines nothing, `OPTIONAL_MODULES_SETUP` compiles away and no module is registered; sources under the directory are already covered by the default recursive `build_src_filter`, so dropping a module in needs no `platformio.ini` edit at all. Verified with `pio run -e native-windows` on a stock tree and with a module present, checking the generated header in both cases.

## Summary by CodeRabbit

* **New Features**
  * Optional modules are now detected and integrated automatically during the build.
  * Available optional modules are initialized without requiring manual configuration.
  * Standard builds remain unchanged when no optional modules are present.


NomDeTom:
Looks like a good idea, but it will need documentation to make sure people use it properly.

NomDeTom (approved):
It's a good idea. Per my other comment it'll need docs to be used right.

thebentern (approved):

caveman99:
I'll make a hello world module distribution as a drop in.

PR #11675, opened 2026-08-31
feat(tmm): add antispam probation greylist, gossiped rate budgets, and relay pricing

toddfrosty:
Adds anti-spam protections to the TrafficManagementModule (TMM) that target the
**rotating-node-ID spammer**: a node that mints fresh `node_num`s each time it gets
rate-limited, so per-sender budgets never engage. All new behavior is config-gated;
the only default-on machinery is the probation accounting plus the reduced hop cap
for probation IDs, which is the minimum to make the greylist meaningful. Signed
identity, network-wide reputation scoring, and proof-of-work challenges are
deliberately out of scope (see Not in scope).

### 1. Probation greylist + attestation promotion

Every first-seen node ID enters a **probation window** (default 300 s) where its
broadcasts are still relayed but at a reduced hop cap (2 instead of 3) and the
lower of the probation multiplier vs. `rate_limit_max_packets`. Promotion out of
probation requires an **ID attestation** from a long-tenured peer: a new
`ID_ATTESTATION_APP` portnum (80) carries an unsigned "I have known 0x… since
epoch-window W" payload; attestations are only accepted from an attester that is
itself at least 24 h old (`attestation_min_tenure_secs`), so a fresh spammer can't
vouch for its clones. First-seen accounting ships on by default — it only
records state, changes no behavior.

### 2. Gossiped median rate budgets

TMM now carries a **top-senders micro-field** (top-3: node, packets-this-window,
RSSI-class) in the periodic `DeviceMetrics` telemetry broadcast — piggybacked on
traffic that already floods, so marginal airtime cost is near zero. When
`budget_gossip_enabled` is set, TMM computes the **median of neighbors' observed
per-sender rates** and, if it exceeds the local `rate_limit_max_packets`, clamps
the local per-sender budget to the median (config becomes floor/ceiling instead of
absolute). The **group budget** catches the rotation case directly: when N
or more *fresh* (probation) IDs co-occur on the same channel within the same
5-minute RSSI-class window, they share one budget — each member gets
`max(1, rate_limit_max_packets / group_size)`.

### 3. Relay congestion pricing

Per-sender **relay budget**: after the local relayer has re-broadcast
`relay_budget_max_packets` of a sender's packets this window, it stops relaying
for that sender (still delivers locally) and gossips a **NO_RELAY** bit via the
same `ID_ATTESTATION_APP` portnum. **Congestion hop cap**: when
channel air-time utilization exceeds `congestion_hop_cap_pct`, probation
broadcast senders are capped at 1 hop.

## New config knobs (`module_config.traffic_management`)

| Knob | Type | Default | Off = |
|---|---|---|---|
| `probation_window_secs` (15) | u32 | **300** (5 min, shipped on) | 0 disables probation entirely |
| `attestation_min_tenure_secs` (16) | u32 | **86400** (24 h, shipped on) | 0 accepts attestations from any attester age |
| `probation_max_hop_limit` (17) | u32 | **2** (shipped on) | 0 relays probation IDs at the normal hop limit |
| `budget_gossip_enabled` (18) | u32 | **0** (off) | local `rate_limit_max_packets` stays authoritative |
| `group_budget_enabled` (19) | u32 | **0** (off) | no group sharing of budgets |
| `relay_budget_max_packets` (20) | u32 | **0** (off) | unlimited relaying |
| `congestion_hop_cap_pct` (21) | u32 | **0** (off) | no congestion hop cap |

Defaults live in `src/mesh/Default.h` (`default_traffic_mgmt_*`); per-knob
semantics are documented in the proto comments in
`protobufs/meshtastic/module_config.proto` (the config-doc location used by the
other TMM knobs).

## Protocol additions (no reordering, additions only)

- `PortNum.ID_ATTESTATION_APP = 80` + `IdAttestation` message (mesh.proto).
- `TopSender` micro-field, `DeviceMetrics.top_senders` (telemetry.proto).
- `TrafficManagement` stats fields (telemetry.options + telemetry.proto) for the
  greylist/budget/relay counters.
- Nanopb regen from protobufs `561658b1490dfd3a9361589501b1321bec2c3543` (already
  on `meshtastic/protobufs`, so the submodule gitlink resolves in CI).

## Files touched

- `protobufs` (submodule → `561658b`), generated `src/mesh/generated/meshtastic/{mesh,telemetry,module_config,portnums,deviceonly*,localonly}.pb.{h,cpp}` (regen)
- `src/modules/TrafficManagementModule.{h,cpp}` — probation, budget-gossip, and relay-pricing logic
- `src/mesh/Default.h` — `default_traffic_mgmt_*` defaults
- `src/mesh/NodeDB.cpp` — first-seen / known-since bookkeeping touchpoint
- `src/mesh/NextHopRouter.cpp` — hop-cap clamp and relay accounting at rebroadcast
- `src/modules/Telemetry/DeviceTelemetry.cpp` — top-senders emission

## Build & test evidence

Host (aarch64 DGX, native env):

```
pio build -e native
=============== [SUCCESS] Took 37.94 seconds ===============
native         SUCCESS   00:00:37.936
========================= 1 succeeded in 00:00:37.936 =========================
```

```
./bin/run-tests.sh   # canonical native suite (AGENTS.md)
...
=============== 1358 test cases: 1358 succeeded in 00:20:08.252 ===============
...
RESULT: AMBER 1 suite(s) left undeclared shared state
    test_waypoint_expiry (undeclared: .portduino/default/Waypoints_default.wpts)
```

AMBER is a **pre-existing** state-manifest artifact of `test_waypoint_expiry`
(this PR does not touch waypoints); no new RED vs. baseline — all 1358 test
cases pass, including the full `test_traffic_management` suite.

Formatting: `clang-format` 16 (trunk config `.trunk/configs/.clang-format`)
clean on all touched handwritten files; `git-clang-format` flags only the
generated `.pb.*` files, which carry `DisableFormat: true`. `bin/lint-*` clean.

## Not in scope (follow-up work)

- Signed identity (XEdDSA-signed attestations) — `IdAttestation` is unsigned here.
- Network-wide reputation (trust-weighted scoring).
- Proof-of-work challenge.
- Any reordering of existing message fields; new portnums only.

## Acceptance evidence

- [x] Diff limited to feature surface (files listed above)
- [x] `make native`/`pio build -e native` passes (tail above)
- [x] `./bin/run-tests.sh`: 1358/1358 cases pass; AMBER = pre-existing waypoint state artifact
- [x] Every new knob documented (proto comments + Default.h)
- [x] No new dependencies; no protocol-breaking changes
- [x] Bisectable commits: one for proto regen, one per mechanism

## Summary by CodeRabbit

* **New Features**
  * Added traffic-management protections to limit excessive relaying and reduce network congestion.
  * Added configurable probation, relay-budget, hop-limit, and congestion controls.
  * Added sharing of top-sender traffic observations through device telemetry.
  * Added safeguards to prevent relaying packets marked as non-relay or exceeding local limits.
* **Bug Fixes**
  * Improved antispam tracking, timing, attestation handling, and cache maintenance for more reliable traffic decisions.
* **Chores**
  * Updated the protobuf definitions reference.

### DIFF
--- a/src/mesh/Default.h
+++ b/src/mesh/Default.h
@@ -61,8 +61,11 @@ enum class TrafficType { POSITION, TELEMETRY };
 #define default_traffic_mgmt_attestation_min_tenure_secs (24 * 60 * 60) // 24 h: attesters must be this old
 #define default_traffic_mgmt_probation_max_hop_limit 2                  // probation broadcasts relay at <=2 hops (normal: 3)
 // M4/M6 knobs ship disabled (0); the gossiped budget / relay-budget /
+// congestion-cap machinery engages only when an operator sets them.
 #define default_traffic_mgmt_budget_gossip_enabled 0
 #define default_traffic_mgmt_group_budget_enabled 0
+#define default_traffic_mgmt_relay_budget_max_packets 0
+#define default_traffic_mgmt_congestion_hop_cap_pct 0
 
 // Hop scaling defaults
 #define default_hop_scaling_min_target_nodes 40          // walk threshold: first hop reaching this cumulative count
--- a/src/mesh/Default.h
+++ b/src/mesh/Default.h
@@ -52,6 +52,15 @@ enum class TrafficType { POSITION, TELEMETRY };
 // Unlike before, lost-and-found is NOT exempt from the relayed precision clamp.
 #define default_traffic_mgmt_lost_and_found_position_min_interval_secs (15 * 60) // 15 minutes
 
+// Antispam L1 (M2/M4/M6). The probation machinery ships on: first-seen
+// accounting (F2.1), the 2-hop relay cap for probation IDs (F2.3), and
+// tenured-node vouching. The M4/M6 budget knobs ship disabled (0) and engage
+// only when an operator sets them. See the TrafficManagementConfig proto
+// comments for per-knob semantics.
+#define default_traffic_mgmt_probation_window_secs (5 * 60)             // 5 minutes of probation for first-seen IDs
+#define default_traffic_mgmt_attestation_min_tenure_secs (24 * 60 * 60) // 24 h: attesters must be this old
+#define default_traffic_mgmt_probation_max_hop_limit 2                  // probation broadcasts relay at <=2 hops (normal: 3)
+
 // Hop scaling defaults
 #define default_hop_scaling_min_target_nodes 40          // walk threshold: first hop reaching this cumulative count
 #define default_hop_scaling_max_target_nodes 80          // generous extension ceiling (2 × min)
--- a/src/mesh/Default.h
+++ b/src/mesh/Default.h
@@ -60,6 +60,9 @@ enum class TrafficType { POSITION, TELEMETRY };
 #define default_traffic_mgmt_probation_window_secs (5 * 60)             // 5 minutes of probation for first-seen IDs
 #define default_traffic_mgmt_attestation_min_tenure_secs (24 * 60 * 60) // 24 h: attesters must be this old
 #define default_traffic_mgmt_probation_max_hop_limit 2                  // probation broadcasts relay at <=2 hops (normal: 3)
+// M4/M6 knobs ship disabled (0); the gossiped budget / relay-budget /
+#define default_traffic_mgmt_budget_gossip_enabled 0
+#define default_traffic_mgmt_group_budget_enabled 0
 
 // Hop scaling defaults
 #define default_hop_scaling_min_target_nodes 40          // walk threshold: first hop reaching this cumulative count
--- a/src/mesh/NextHopRouter.cpp
+++ b/src/mesh/NextHopRouter.cpp
@@ -247,10 +247,11 @@ bool NextHopRouter::perhapsRebroadcast(const meshtastic_MeshPacket *p)
 #endif
 
 #if HAS_TRAFFIC_MANAGEMENT
-    // Antispam L1 (M6): stop relaying a sender whose local relay budget is
-    // exhausted (or whose NO_RELAY was gossiped). The packet still flows to
-    // our own client - only propagation stops.
-    if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag && trafficManagementModule &&
+    // Check if traffic management wants to stop relaying a sender whose local
+    // relay budget is exhausted (or whose NO_RELAY was gossiped). Only relayed
+    // traffic counts: a packet addressed to us (or from us) must still reach
+    // our own client and its ACK/NAK path, so it is never "consumed" here.
+    if (p->which_payload_variant == meshtastic_MeshPacket_decoded_tag && trafficManagementModule && !isToUs(p) && !isFromUs(p) &&
         !trafficManagementModule->shouldRelay(*p)) {
         LOG_DEBUG("Antispam: not relaying 0x%08x (relay budget / no-relay)", getFrom(p));
         return true; // consumed: deliver locally, do not propagate
--- a/src/mesh/NextHopRouter.cpp
+++ b/src/mesh/NextHopRouter.cpp
@@ -246,6 +246,17 @@ bool NextHopRouter::perhapsRebroadcast(const meshtastic_MeshPacket *p)
     }
 #endif
 
+#if HAS_TRAFFIC_MANAGEMENT
+    // Antispam L1 (M6): stop relaying a sender whose local relay budget is
+    // exhausted (or whose NO_RELAY was gossiped). The packet still flows to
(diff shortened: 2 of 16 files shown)


toddfrosty:
All 7 CodeRabbit findings fixed. One commit each, in review order:

| # | Finding (bot comment id) | Fix | Commit |
|---|---|---|---|
| 1 | `shouldRelay` ran before the relay/destination gate in `perhapsRebroadcast()`, so packets addressed to us from a no-relay sender were consumed and lost their ACK (id 3893638765) | added the `!isToUs(p) && !isFromUs(p)` condition to the check | 47167d2 |
| 2 | `recordRelayed()` had no decoded-variant guard, so it read the portnum exemption off ciphertext (id 3893638772) | wrap the call in a `which_payload_variant == decoded` check | dc19a4a |
| 3 | vouch cadence compared `millis()` directly and the `lastVouchSentMs == 0` cold start blocked vouching for an hour of uptime (id 3893638776) | use `Throttle::hasElapsed`, and let the first vouch go out immediately | f03a812 |
| 4 | `findAntispamEntry()` allocated/evicted on every miss even for read-only callers, so remote-supplied senders could churn the table and evict real probation / no-relay state (id 3893638781) | split into `findAntispamEntry()` (lookup only, never allocates) + `findOrCreateAntispamEntry()` (allocates, oldest-first eviction); every read-only caller now uses the lookup, `noteFirstSeen` / `ingestNeighborTopSenders` the creating variant; eviction ordered by elapsed ticks (wrap-safe for the 4-bit tick) and new entries are stamped with the current tick before they can be picked as victims | fa468a6 |
| 5 | tenure floor accepted `attester_tenure_secs == 0`, which the proto defines as "no clock" — an attester with no clock could promote the subject (id 3893638785) | drop the `!= 0` term; zero is now rejected as untenured | 0acb06a |
| 6 | `maintainAntispamLocked()` dereferenced the table with no null guard after a failed alloc (id 3893638791) | `if (!antispam) return;` at the top | 9446c14 |
| 7 | `delta & 0x0F >= 1` parsed as `delta & (0x0F >= 1)` == `delta & 1` — rollover only fired on odd tick deltas (id 3893638795) | parenthesized in both the entry sweep and the group-cell sweep | 1121b4a |

Deviations from the suggested shape (finding 4):
- `recordRelayed` is one of the read-only (non-allocating) callers, per the finding's caller list. It does modify the existing entry, so the lookup there uses a non-const pointer (`findAntispamEntry` returns the entry's address directly; mutation does not need `mutable` state).
- New entries get their `firstSeenTick` stamped at creation (the creating variant), which the finding called for; this also means a just-evicted-into slot can never be re-picked as its own victim on the next insert.

Build (native, aarch64):
```
native build, aarch64 (DGX):
  clean .pio, pio run -e native:
    Linking .pio/build/native/meshtasticd
    ========================= [SUCCESS] Took 45.74 seconds =========================
  re-verified after the final recordRelayed const-correctness fix (incremental):
    ========================= [SUCCESS] Took 15.19 seconds =========================
  Environment  Status  Duration
  native       SUCCESS 00:00:15.188
```

Tests (`./bin/run-tests.sh`, coverage env):
```
./bin/run-tests.sh (coverage env), head 1121b4a:
  =============== 1358 test cases: 1358 succeeded in 00:17:49.077 ===============
  RESULT: test attribution OK [coverage]
  RESULT: AMBER 1 suite(s) left undeclared shared state
          test_waypoint_expiry (pre-existing on this PR head, unrelated to these changes)
```

Baseline at the previous PR head (88114ec): 1358/1358 test cases passed, verdict AMBER solely due to `test_waypoint_expiry` leaving undeclared shared state (pre-existing in this PR, unrelated to these changes). The run above is the same: no new RED, no new AMBER causes.

Push note: the card's prescribed push URL used the x-access-token:***@host form, which GitHub rejects for a classic ghp_ PAT (it is only valid for app/OAuth tokens). Pushed with the equivalent user:token form instead; the token was inline in the push command only and is not stored in any git config.

toddfrosty:
Rebased the feature commits to drop internal labels from commit messages and rewrote the in-code comments that referenced them; comments-only change, no behavior change; native build green, 1358/1358 tests (same pre-existing waypoint AMBER as before).

toddfrosty:
Note: this PR is currently closed (not merged) - I see it was closed Aug 31 at 18:28 UTC. The branch (antispam-l1-greylist-budgets-relay) was just updated to 9f0a5be: the feature commit messages no longer carry internal labels, and the comments that referenced them are rewritten in plain English (comments-only, no behavior change). The closed PR's head snapshot is still 1121b4a, so if this should stay open, either re-open it or open a fresh PR from the branch to pick up the updated head.

PR #11676, opened 2026-08-31
fix(radio): recover a chip that lost its state instead of assert-crashing in reconfigure()

thebentern:
## Problem

Since #9962, every LoRa config change applies live: `configChanged` observer → the radio driver's `reconfigure()`. If the chip has lost its runtime configuration by then (chip-internal reset/brownout), `reconfigure()` crashes the node on an `assert`: on every one of these chips the power-on-reset modem is FSK/GFSK, so the packet-type-guarded RadioLib setters return `RADIOLIB_ERR_WRONG_MODEM`, and on some (measured on SX1262) the first `standby()` after state loss fails outright with `SPI_CMD_TIMEOUT` (-707), tripping the assert inside `setStandby()` itself.

The assert reboots the node from inside `MeshService::reloadConfig()` **before** `saveToDisk()` runs, so the config change that triggered the reconfigure is lost. On a first-time region set the damage is worse: `ensurePkiIdentity()` has already saved nodes.proto with the old node purged, but the unsaved devicestate reverts to the old identity on reboot — the node comes back with region UNSET, its old node number, and its own entry missing from the node DB.

Observed in the field on a T1000-E (LR1110, 2.8.0.52d5214) during a phone-driven region set:

```
ERROR | 07:14:38 47 NOTE! Record critical error 7 at src/mesh/LR11x0Interface.cpp:282
ERROR | 07:14:38 47 NOTE! Record critical error 7 at src/mesh/LR11x0Interface.cpp:286
ERROR | 07:14:38 47 NOTE! Record critical error 7 at src/mesh/LR11x0Interface.cpp:290
ERROR | 07:14:38 47 assert failed src/mesh/LR11x0I
```

Reproduced on demand on a RAK4631 (SX1262) by cold-sleeping the chip (drops all config, mimicking a brownout) immediately before a live config apply: the stock firmware crash-loops on the `setStandby()` assert; with this PR the same stimulus logs the error, re-inits the chip in place, and the node keeps running:

```
DEBUG | SX126x standby RadioLib err=-707
ERROR | NOTE! Record critical error 7 at src/mesh/SX126xInterface.cpp:301
ERROR | SX126x rejected modem params, chip state lost? Full re-init
INFO  | SX126x init result 0
INFO  | SX126x recovered after re-init
```

The modem parameters themselves were valid in the field trace (SF11/BW500/CR8, provable from the logged preamble time), so this is not the old zero coding-rate/spread-factor bug; the AdminModule clamps already cover that.

## Fix (LR11x0, SX126x, SX128x, RF95, LR20x0)

- `setStandby()` is split: `trySetStandby()` does the work and returns the error; `setStandby()` keeps the assert for all other callers. `reconfigure()` uses `trySetStandby()` so a standby failure routes into recovery instead of crashing.
- Modem parameter programming moves into `programModemParams()`, which returns the first RadioLib error and logs each failure with its error code (previously critical error 7 was recorded with no code, making field logs undiagnosable).
- On failure, `reconfigure()` recovers in place via `reinitChip()` (LR20x0: the existing band-hop `fullBegin()` path, now shared): `begin()` hardware-resets the chip and reprograms it, then chip-side state that `begin()` does not restore is re-applied per family — DIO RF-switch table, DIO2-as-RF-switch, OCP limit, PA ramp, RX gain, RX-sensitivity register patch, CRC. `init()` shares the same helper instead of duplicating it.
- `reinitChip()` clamps `power` before `begin()`: `applyModemConfig()` resets it to the raw config value and the recovery path doesn't pass through the params clamp (without this, tx_power=30 on a 22 dBm part made LR11x0 recovery fail with INVALID_OUTPUT_POWER; observed as `init result -13` on the RAK4631 during bring-up of this fix).
- Only if recovery also fails does `reconfigure()` give up — with a logged error and `return false`, never a crash — so the pending config save still completes.

Recovery deliberately re-runs `begin()` directly rather than `init()`: `Observer::observe()` appends unconditionally, so a second `init()` would double-register `configChangedObserver` and run every subsequent reconfigure twice.

## Testing

- T1000-E (LR1110): flashed; live LoRa config set over serial programs cleanly and persists.
- RAK4631 (SX1262): induced chip state loss (cold sleep) right before a live config apply. Stock develop: assert crash + reboot loop, config change lost. This PR: standby -707 tolerated, full re-init, `SX126x recovered after re-init`, node keeps running and the config persists.
- Builds: `rak4631`, `tracker-t1000-e`, `native-macos` (portduino compiles all five drivers, including SX128x and LR20x0).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved modem recovery after resets, brownouts, or power interruptions.
  - Automatically reinitializes affected radios and reapplies communication settings after configuration failures.
  - Preserves the detected oscillator voltage during recovery for more reliable startup.
  - Handles standby and parameter-programming errors without triggering premature assertions.
  - Reports configuration failures accurately and stops recovery when retry attempts are unsuccessful.

### DIFF
--- a/src/mesh/LR11x0Interface.cpp
+++ b/src/mesh/LR11x0Interface.cpp
@@ -172,6 +172,8 @@ template <typename T> bool LR11x0Interface<T>::init()
         res = tryBegin(3, attemptVoltage);
     }
 
+    resolvedTcxoVoltage = attemptVoltage;
+
     // \todo Display actual typename of the adapter, not just `LR11x0`
     LOG_INFO("LR11x0 init result %d", res);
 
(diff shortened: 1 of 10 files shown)


Xaositek:
INFO  | 09:12:53 30 Set config: LoRa
WARN  | 09:12:53 30 Invalid coding_rate 0, set to 5
INFO  | 09:12:54 30 Wanted region 1, using US
INFO  | 09:12:54 30 GPS power state OFF -> ACTIVE
INFO  | 09:12:54 30 [GPS] Save changes to disk
DEBUG | 09:12:54 31 [GPS] Expand short PSK #1
INFO  | 09:12:54 31 [GPS] Wanted region 1, using US
INFO  | 09:12:54 31 [GPS] Default Coding Rate above custom setting, use 8
INFO  | 09:12:54 31 [GPS] Radio freq=908.750, config.lora.frequency_offset=0.000
INFO  | 09:12:55 31 [GPS] Set radio: region=US, name=LongTurbo, config=9, ch=13, power=30
INFO  | 09:12:55 31 [GPS] newRegion->freqStart -> newRegion->freqEnd: 902.000000 -> 928.000000 (26.000000 MHz)
INFO  | 09:12:55 31 [GPS] numFreqSlots: 52 x 500.000kHz
INFO  | 09:12:55 31 [GPS] channel_num: 14
INFO  | 09:12:55 31 [GPS] frequency: 908.750000
INFO  | 09:12:55 31 [GPS] Slot time: 17 msec, preamble time: 65 msec
INFO  | 09:12:55 32 [GPS] Final Tx power: 22 dBm
DEBUG | 09:12:55 32 [GPS] Save to disk 3

Issue #6987, opened 2025-06-08
[Feature Request]: Add public label to any channel using public PSK


willhubs:
### Platform

Cross-Platform

### Description

Whenever a channel is displayed, if it's using the public PSK of AQ== there would be a designation of public next to the channel name.

So in the "contacts" tab and when sending messages it would say LongFast(public) instead of just LongFast.

This would help clarify for new users that what they are sending is on a public channel.

fifieldt:
Good idea. Probably the best place to execute this change is in the clients: Android, IoS, Web -- would you have time to open issues in those repositories?

willhubs:
Yes, I can do that.

Issue #980, opened 2021-12-09
Remove BATTERY_PIN from lora32 v1 board


mc-hamster:
[The schematic ](https://github.com/meshtastic/Meshtastic/blob/master/website/static/documents/WIFI_LoRa_32_V1(868-915version)Schematic_diagram.PDF) shows there's no voltage divider. 

This could be the cause of our problems on that board.

PR #11681, opened 2026-08-31
feat(screen): stream the display framebuffer to local clients (DisplayFrame mirroring)

jamesarich:
Streams the device's screen to its locally connected client and completes the remote-control story started by `send_input_event` (meshtastic/protobufs#702). Prior art: meshtastic/web#224. **Draft until meshtastic/protobufs#1054 merges; the submodule pointer will move to protobufs master after.** Android client: meshtastic/Meshtastic-Android#6987.

## How it works

- `graphics::ScreenMirror` snapshots the 1bpp `OLEDDisplay` framebuffer after each committed frame (single hook at `updateUiFrame`'s exits, so the lockdown build mirrors the redacted LOCKED frame, never operator content). A `memcmp` diff gates capture: a static screen streams nothing.
- Chunks drain through `PhoneAPI`'s existing `getFromRadio` chain at **lowest priority** - mesh packets, notifications and the satellite replay always outrank pixels - with a **per-client cursor** (`frameId`, `offset`), so coexisting BLE/serial/TCP clients each receive complete frames.
- **Color**: `TFTDisplay`/`HUB75Display` hand the mirror their color-region table at paint time (immediately before `clearTFTColorRegions()`), byte-swapped to logical RGB565. Palettes stream as `display_palette` chunks keyed by the existing region-table signature and re-send only when layout/theme changes; frames carry the signature of the palette they were painted with (coherent even across a mid-drain theme change, which also counts as frame-worthy so recolors propagate without pixel churn).
- `DeviceMetadata.display` (`DisplayInfo`) reports geometry, `PanelClass` and touch during the handshake.
- Admin verbs are **local-only** (`mp.from != 0` is dropped): frames ride `FromRadio` and can never cross the mesh.

| Condition | Behavior |
| --- | --- |
| Screen unchanged, mirroring on | nothing sent (diff-gated) |
| Theme recolor, pixels unchanged | new frame + palette sent (signature change is frame-worthy) |
| Client disconnects | mirror disarmed, snapshot + palette freed (memaudit-tracked) |
| Unauthorized client under `MESHTASTIC_PHONEAPI_ACCESS_CONTROL` | never receives frames/palettes (same `getAdminAuthorized` bar as mesh packets) |
| Display-less / `MESHTASTIC_EXCLUDE_SCREEN_MIRROR` build | compiles out (`HAS_SCREEN_MIRROR`), verbs ignored |
| E-ink / HUB75 | same 1bpp path (all BaseUI targets share the buffer) |

## Cost

Zero allocation until armed: snapshot (1 KB on 128×64, 9.6 KB on 320×240 1bpp) + 576 B palette table on arm, freed on disarm/disconnect, `memaudit`-tracked. ~2 KB flash when enabled; excluded builds pay only the nanopb `FromRadio` union growth from the new oneof members (encoded `FromRadio_size` stays 510 ≤ 512).

## Known limitations (deliberate for this round)

- **Single global arm flag**: any local client's disconnect disarms the stream, and under access control an unauthorized local client can arm/disarm (control-plane only - it can never *read* frames). Per-client refcounted arming is the natural follow-up; the per-client drain cursors it needs already exist.
- MUI/LVGL (RGB565) devices aren't mirrored yet; `DisplayFrame`'s reserved rect fields are the path.
- `test_screen_mirror` native suite (chunk reassembly, mid-drain restart, palette signature changes, memaudit balance) is sketched and queued as a follow-up commit.

## 🤝 Attestations

- Devices this was regression-tested on:
  - [ ] Heltec V3 (builds; not flashed)
  - [x] T-Deck - live color mirror + remote input, plain `t-deck` (BaseUI) env
  - [ ] T-Beam
  - [x] RAK 4631 - live mono mirror + remote input (WisMesh Pocket)
  - [ ] T-1000E
  - [x] Other: `native-macos`, `heltec-mesh-node-t114` (ST7789), `heltec-v3` - build verification incl. the coloring-disabled TFT combination

---

## Status (2026-09-08)

Rebased onto `develop`. Submodule and generated nanopb headers regenerated with nanopb 0.4.9.1 for the `FromRadio.display_frame` 20 → 21 / `display_palette` 21 → 22 / `DeviceMetadata.display` 15 → 16 renumbering in meshtastic/protobufs#1054 (those tags collided with open PRs #980 and #1030). No source change was needed - every call site already goes through the generated `*_tag` macros.

**Blocked on device-ui.** The MUI/LVGL colour path needs a device-ui carrying `DisplayMirror` (device-ui#394, draft and green). That is in no released device-ui. So `HAS_MUI_MIRROR` is 0 in every configuration this branch commits, and roughly half the diff - `onMuiRect`, the rect pool, `muiInjectInputEvent`, `muiDisplayInfo`, the `DisplayMirror` wiring in `tftSetup`, the `DisplayInfo`-on-MUI branch in `main.cpp` - compiles out. Only the 1bpp mono path is reachable from these branches.

**Fixed:** `3d35b6a0d` set out to drop a spike-local device-ui override but appended a *second* archive to `[device-ui_base] lib_deps`, so every MUI board pulled two libraries both named `device-ui`. Now one pin.

**Known gaps, raised by a cross-repo audit and not yet addressed here:**

- `frameReady.notifyObservers()` runs on the LVGL/tft task on MUI builds, while `Observable`/`Observer` are backed by an unsynchronised `std::list` mutated from the main/BLE thread - and the notify chain reaches NimBLE `notify()` from that task.
- The MUI rect-pool overflow path resets the queue including `muiRectSendOffset`, abandoning a rect the client is mid-way through receiving and re-arming a full repaint that will overflow again.
- `PhoneAPI::close()` calls `setMirror(false)`, so any handshaked client tearing down disarms mirroring for every other client - and `close()` also runs from `checkConnectionTimeout()`, so an idling serial/TCP session silently kills a live stream.

---

## Feature stack

Five repos, and the release order runs top to bottom. Nothing below can ship until the piece above it lands.

| Repo | PR | Role |
| --- | --- | --- |
| meshtastic/design | [#142](https://github.com/meshtastic/design/issues/142) | Cross-platform feature spec |
| meshtastic/device-ui | [#394](https://github.com/meshtastic/device-ui/pull/394) | `DisplayMirror`: frame capture + remote input |
| meshtastic/protobufs | [#1054](https://github.com/meshtastic/protobufs/pull/1054) | `DisplayFrame` / `DisplayPalette` / `DisplayInfo` + the two admin verbs |
| meshtastic/firmware | [#11681](https://github.com/meshtastic/firmware/pull/11681) | Producer: streams the framebuffer, bridges input |
| meshtastic/Meshtastic-Android | [#6987](https://github.com/meshtastic/Meshtastic-Android/pull/6987) | Mirror tab with remote control |
| meshtastic/meshtastic-mcp | [#78](https://github.com/meshtastic/meshtastic-mcp/pull/78) | `capture_display` tooling |

device-ui#394 is the current blocker: without `DisplayMirror` the MUI/colour path compiles out of every committed firmware configuration, leaving only the 1bpp mono path.

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -645,6 +645,16 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define HAS_SCREEN_MIRROR 0
 #endif
 
+// Mirroring MUI (device-ui/LVGL) streams RGB565 dirty rects instead of the
+// 1bpp framebuffer, and needs a device-ui carrying the flush observer. Until
+// that lands upstream the whole path - queue, pool and input seam - compiles
+// out, so a color build that does not opt in carries none of it.
+#if HAS_SCREEN_MIRROR && HAS_TFT && defined(MESHTASTIC_MUI_MIRROR)
+#define HAS_MUI_MIRROR 1
+#else
+#define HAS_MUI_MIRROR 0
+#endif
+
 #ifndef USE_ETHERNET_DEFAULT
 #define USE_ETHERNET_DEFAULT 0
 #endif
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -637,6 +637,14 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #define HAS_SCREEN 0
 #endif
 
+// Display mirroring to the client (FromRadio.display_frame) rides the screen
+// and can be excluded independently with MESHTASTIC_EXCLUDE_SCREEN_MIRROR.
+#if HAS_SCREEN && !defined(MESHTASTIC_EXCLUDE_SCREEN_MIRROR)
+#define HAS_SCREEN_MIRROR 1
+#else
+#define HAS_SCREEN_MIRROR 0
+#endif
+
 #ifndef USE_ETHERNET_DEFAULT
 #define USE_ETHERNET_DEFAULT 0
 #endif
--- a/src/graphics/HUB75Display.cpp
+++ b/src/graphics/HUB75Display.cpp
@@ -3,6 +3,7 @@
 #if defined(USE_HUB75)
 
 #include "HUB75Display.h"
+#include "ScreenMirror.h"
 #include "TFTColorRegions.h"
 #include "TFTPalette.h"
 #include <ESP32-HUB75-MatrixPanel-I2S-DMA.h>
@@ -121,6 +122,9 @@ void HUB75Display::display()
     firstFrame = false;
 #if GRAPHICS_TFT_COLORING_ENABLED
     lastColorSig = colorSig;
+#if HAS_SCREEN_MIRROR
+    graphics::screenMirror.capturePalette(colorSig, onBe, offBe, graphics::colorRegions, graphics::getTFTColorRegionCount());
+#endif
     // Regions are re-registered every frame by the renderers; clear so they
     // don't accumulate across frames.
     graphics::clearTFTColorRegions();
--- a/src/graphics/Screen.cpp
+++ b/src/graphics/Screen.cpp
@@ -23,6 +23,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "Screen.h"
 #include "NodeDB.h"
 #include "PowerMon.h"
+#include "ScreenMirror.h"
 #include "Throttle.h"
 #include "configuration.h"
 #include "meshUtils.h"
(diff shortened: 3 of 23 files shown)


jamesarich:
Cross-platform feature spec (client-agnostic wire contract, reassembly rules, input mapping, UX and device requirements): meshtastic/design#142

jamesarich:
Rebased onto develop. It had gone conflicting overnight (`PhoneAPI.cpp`, plus the device-ui pin in `platformio.ini`), and a conflicting PR runs no workflows at all, silently, so nothing had been checked since.

Also follows device-ui#394's new shape. The mirror moved out of device-ui's `InputDriver` and `DisplayDriver` base classes into a `DisplayMirror` class, so the call sites here move with it: `DisplayMirror::start(deviceScreen->getDisplayDriver())` after `DeviceScreen::init()` replaces `InputDriver::enableInjection()` before it. Renames otherwise. All of it sits inside `HAS_MUI_MIRROR`, which no committed configuration defines.

PR #11688, opened 2026-09-01
Honor mute when waking the screen for a received message

### LINKED ISSUE #11674: [Feature Request]: Do not wake up screen for muted channels / direct messages
### Platform

Cross-Platform

### Current behavior
Currently, the device screen wakes up and displays the text for every received message, regardless of the channel's priority or notification settings. In busy public channels or high-traffic mesh networks, this behavior causes unnecessary battery drain.
It also negatively impacts the user experience when a user wants to remain part of a channel for logging or occasional reading but does not need real-time screen alerts for every single message.

### Proposed solution
I would like to introduce an option to suppress screen wake-up and message rendering for specific channels or peer-to-peer interactions that are marked as "muted" or "low-priority".

caveman99:
Closes #11674

## Problem

The screen wakes for every received text message regardless of mute state.

`TextMessageModule::handleReceived()` fires `powerFSM.trigger(EVENT_RECEIVED_MSG)` for every text packet, gated only by `shouldWakeOnReceivedMessage()`. That helper checks whether external notification is enabled, the device role, and the battery level. It never consults the mute flags. `EVENT_RECEIVED_MSG` transitions LS/NB/DARK to ON, so a muted channel suppressed the banner and still lit the screen.

Separately, `MessageRenderer::handleNewMessage()` computed mute only for `MessageType::BROADCAST`, so a DM from a muted node produced both a banner and a wake.

## Change

Add `isMutedForPacket()` to `Channels`. A DM addressed to us reads the sender's `NodeInfoLite` mute bit; every other packet reads the mute bit of the channel it arrived on. This is the predicate `ExternalNotificationModule` already applied to the buzzer, vibra and LED outputs, hoisted so all three call sites share one definition.

- `TextMessageModule.cpp`: the wake trigger is now gated on it.
- `MessageRenderer.cpp`: banner suppression now covers muted DM senders, not just muted broadcast channels.
- `ExternalNotificationModule.cpp`: inline logic replaced by the helper. Behavior unchanged.

Bell and alert messages still break through mute on both paths. `isAlert` bypasses the muted return in `MessageRenderer`, and that path calls `screen->setOn(true)` itself, so an alert on a muted channel still lights the screen.

No protobuf or config change. `ChannelSettings.module_settings.is_muted` and the `NodeInfoLite` mute bit already exist, and are already settable from the device menu and via `AdminMessage.toggle_muted_node`.

## Tests

New `test/test_muted_source`, 10 cases covering both branches:

- broadcast on an unmuted channel, on a muted channel, and with `has_module_settings` absent
- mute is per channel, not global
- `packet.channel == 0` resolves to the primary slot, which is not always slot 0
- DM to us from a muted sender, from an unmuted sender, and from a sender not in the DB
- a DM to us does not inherit its channel's mute state
- a DM to a third party falls back to the channel

Declared in `test/state-manifest.tsv` (constructs a NodeDB, whose constructor persists a default set in an empty prefs directory).

Verified locally: suite green (10/10), `pio run -e native-windows` exit 0, `trunk fmt` clean. `MessageRenderer.cpp` is not built by `native-windows`, which is headless, so that file is covered by CI only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Muted direct messages from muted senders are now consistently suppressed.
  * Muted channels no longer trigger message banners, notifications, or screen wake events.
  * Message mute handling correctly distinguishes direct messages from channel broadcasts.
  * Configured alert messages can still trigger notifications and screen wake events, even on muted channels.
  * Message banners no longer appear over modal screens.
* **Tests**
  * Added coverage for muted sources, alert messages, channel handling, and notification behavior.

### DIFF
--- a/src/graphics/draw/MessageRenderer.cpp
+++ b/src/graphics/draw/MessageRenderer.cpp
@@ -3,6 +3,7 @@
 #include "MessageRenderer.h"
 
 // Core includes
+#include "Channels.h"
 #include "MessageStore.h"
 #include "NodeDB.h"
 #include "UIRenderer.h"
@@ -1138,13 +1139,7 @@ void handleNewMessage(OLEDDisplay *display, const StoredMessage &sm, const mesht
         // still happens so a message can light the screen back up.
         const bool menuShowing = NotificationRenderer::isMenuShowing();
 
-        // Determine if message belongs to a muted channel
-        bool isChannelMuted = false;
-        if (sm.type == MessageType::BROADCAST) {
-            const meshtastic_Channel channel = channels.getByIndex(packet.channel ? packet.channel : channels.getPrimaryIndex());
-            if (channel.settings.has_module_settings && channel.settings.module_settings.is_muted)
-                isChannelMuted = true;
-        }
+        const bool isMuted = isMutedForPacket(packet);
 
         // Banner logic
         const meshtastic_NodeInfoLite *node = nodeDB->getMeshNode(packet.from);
@@ -1186,8 +1181,8 @@ void handleNewMessage(OLEDDisplay *display, const StoredMessage &sm, const mesht
             else
                 strcpy(banner, "Alert Received");
         } else {
-            // Skip muted channels unless it's an alert
-            if (isChannelMuted)
+            // Skip muted channels/senders unless it's an alert
+            if (isMuted)
                 return;
 
             if (truncatedLongName[0]) {
--- a/src/graphics/draw/MessageRenderer.cpp
+++ b/src/graphics/draw/MessageRenderer.cpp
@@ -4,6 +4,7 @@
 
 // Core includes
 #include "Channels.h"
+#include "MeshService.h"
 #include "MessageStore.h"
 #include "NodeDB.h"
 #include "UIRenderer.h"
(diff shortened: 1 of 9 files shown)


caveman99:
Addressed in ad6837da0.

**Alert wake exception (`TextMessageModule.cpp:39`).** Valid, fixed. In COLOR display mode `handleNewMessage()` is skipped, so `powerFSM.trigger(EVENT_RECEIVED_MSG)` is the only wake an alert gets, and gating it on mute alone dropped it. Added `MeshService::isAlertPayload()`: an ASCII BEL in the payload while at least one `alert_bell_*` output is enabled. The wake gate is now "not muted, or an alert".

`MessageRenderer` now calls the same predicate instead of its own inline bell scan, so the banner and the wake cannot disagree about what counts as an alert. That also lifts the scan's arbitrary 100 byte cap.

**Trunk check.** The three failures were trufflehog false positives, not formatting. `test_broadcast_on_muted_channel_is_muted`, `test_dm_to_us_from_muted_sender_is_muted` and `test_dm_to_third_party_uses_channel_mute` each carry exactly 35 characters after the `test_` prefix, which matches Lob's API key format. Renamed rather than adding permanent ignore comments.

**Docstring coverage (50% vs 80%).** Not addressed. The count is dominated by test functions, and `AGENTS.md` says "Keep code comments minimal - one or two lines, max." Adding docstrings to each test case would work against the repo's own convention.

Suite is now 14 cases, all passing. `trunk fmt` and `trunk check` clean across all touched files.

thebentern (approved):

PR #11698, opened 2026-09-01
fix(stm32wl): improve reboot-to-DFU reliability

ndoo:
## What this does

Makes `enter_dfu_mode_request` reliably land the STM32WL in its ROM bootloader so a subsequent flash succeeds on the first try.

On `develop` the admin handler calls `enterDfuMode()` → `HAL_NVIC_SystemReset()` inline, before the `want_response` ACK is sent and while the client still holds the console UART. The STM32WL ROM bootloader autobauds off the first byte it receives on USART1 (PB6/PB7) or USART2 (PA2/PA3).

This doesn't give enough time for both the Meshtastic device and the client to cleanly detach the port before any stray byte(s) remaining in the buffer or in the client app locks the autobauding of the bootloader at the previous baudrate - the bootloader is then stuck at 115200 or whatever baudrate the corrupted bytes got detected at.

This can only be cleared with an RST or a power cycle.

### Changes

- **Defer the jump.** `enter_dfu` now arms `enterDfuAtMsec = millis() + 5s` and returns, so the ACK goes out the normal path; `Power::powerCommandsCheck()` calls `enterDfuMode()` at the deadline. The 5 s is the client's detach window and the margin a WebSerial web-flasher needs (meshtastic/web-flasher#426). New `enterDfuAtMsec` global, `#ifdef ARCH_STM32` only; arm site remaps a wrapped-to-0 deadline to 1, matching the existing GPS `fixHoldEnds` pattern. nrf52/rp2040 keep the inline call.
- **Drain the UARTs.** `enterDfuMode()` now stops the GPS and flush/end's every configured UART (`Serial`, `Serial1`, `Serial2`) before the reset. Factored into `quiesceSerial()` and reused in `cpuDeepSleep()`.
- **Fix the early-boot redirect.** `earlyBootCheck` moves from `constructor(101)` to `.preinit_array`, so it runs ahead of the STM32 core's `premain()`/`SystemClock_Config()` regardless of link order, and resets RCC to its reset state (mirroring CMSIS `SystemInit()`) before jumping to system memory. No option-byte changes.

### Known limitation

`gps->disable()` only issues a UBX sleep command. A non-u-blox or free-running GPS with no hardware enable/standby pin keeps transmitting on its UART; if that UART is USART1 or USART2 the ROM bootloader can still autobaud onto the GPS stream.

The above is an existing limitation for all STM32WL platforms and pre-dates this change. STM32WL designs should keep GPS UARTs off those pins or provide a way to power down / hold-in-reset the GPS before DFU.

## Test plan

- Build-verified: `rak3172` (87.8% flash), `russell`.
- Native C++ test suite unaffected — every hunk is behind `#if defined(ARCH_STM32)` or lives in `main-stm32wl.cpp`.
- Hardware-verified on rak3172 (BOOT0-button bootloader entry as baseline reference): `--enter-dfu` → release port → wait untouched → STM32CubeProgrammer `br=115200 -w -v -g 0x08000000` → "Activating device: OK" on the first attempt, every cycle.

## Attestations

  - [x] Other: RAK3172 (STM32WLE5) — hardware-verified reboot-to-DFU; rak3172 + russell build-verified.

## Summary by CodeRabbit

* **New Features**
  * Added delayed DFU mode transition on STM32 devices, allowing request acknowledgments to complete before switching modes.
  * DFU entry now occurs automatically after a brief delay.

* **Bug Fixes**
  * Improved STM32 bootloader handoff reliability by resetting system state and preparing connected peripherals correctly.
  * Disabled GPS and safely closed serial connections before entering DFU mode to prevent communication interference.

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -976,6 +976,14 @@ void Power::powerCommandsCheck()
         shutdownAtMsec = 0;
         shutdown();
     }
+
+#ifdef ARCH_STM32
+    // Deferred DFU entry; the delay is armed by AdminModule's enter_dfu handler (rationale there).
+    if (enterDfuAtMsec && Throttle::deadlinePassed(enterDfuAtMsec)) {
+        enterDfuAtMsec = 0;
+        enterDfuMode(); // never returns
+    }
+#endif
 }
 
 void Power::reboot()
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1283,6 +1283,9 @@ void setup()
 uint32_t rebootAtMsec;     // If not zero we will reboot at this time (used to reboot shortly after the update completes)
 uint32_t shutdownAtMsec;   // If not zero we will shutdown at this time (used to shutdown from python or mobile client)
 bool suppressRebootBanner; // If true, suppress "Rebooting..." overlay (used for OTA handoff)
+#ifdef ARCH_STM32
+uint32_t enterDfuAtMsec; // If not zero, enter DFU mode at this millis() deadline (see main.h)
+#endif
 
 #if defined(MESHTASTIC_ENCRYPTED_STORAGE) && defined(MESHTASTIC_PHONEAPI_ACCESS_CONTROL)
 volatile bool lockdownReloadPending;  // see main.h - deferred NodeDB reload after lockdown unlock
--- a/src/main.h
+++ b/src/main.h
@@ -92,6 +92,9 @@ extern uint32_t timeLastPowered;
 extern uint32_t rebootAtMsec;
 extern uint32_t shutdownAtMsec;
 extern bool suppressRebootBanner;
+#ifdef ARCH_STM32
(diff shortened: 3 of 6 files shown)


ndoo:
Addressed the CodeRabbit nitpick — the `enterDfuAtMsec` entry in the `Throttle.h` sentinel-meaning comment is now two lines, within the one-to-two-line comment guideline.

@coderabbitai review

ndoo:
Review called too soon perhaps.

@coderabbitai review

thebentern (approved):

caveman99 (approved):
You need to cut down on the comments a LOT. Documentation does not live in code, especially if it burys the code itself.

ndoo:
> You need to cut down on the comments a LOT. Documentation does not live in code, especially if it burys the code itself.

Hmm I kept them because there's a lot of magic numbers. But I'll bear that in mind and if the same code is touched by me in future will trim it.

PR #11699, opened 2026-09-02
fix(position): exempt trackers from reply throttle

### LINKED ISSUE #11696: Allow TRACKER role to bypass the 3-minute position reply throttle
Directed position replies are currently limited to one reply every 3 minutes.

`LOST_AND_FOUND` bypasses this limit, but `TRACKER` does not.

For tracker use cases, 3 minutes is too long for interactive position requests, especially when the tracked device is moving.

### Proposal

Allow `TRACKER` to bypass the same throttle as `LOST_AND_FOUND`.

Conceptually:

```cpp
if (role != LOST_AND_FOUND &&
    role != TRACKER &&
    replyWasSentLessThan3MinutesAgo) {
    skipReply();
}
```

This keeps the existing protection for normal nodes while making directed position requests practical for tracker devices.

lnx13:
fixes #11696
## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
     RAK WisMesh Tag

## Summary by CodeRabbit

* **Bug Fixes**
  * Tracker devices now receive position replies without being blocked by the three-minute reply limit.

### DIFF
--- a/src/modules/PositionModule.cpp
+++ b/src/modules/PositionModule.cpp
@@ -282,7 +282,8 @@ meshtastic_MeshPacket *PositionModule::allocPositionPacket(uint32_t atPrecision)
 
 meshtastic_MeshPacket *PositionModule::allocReply()
 {
-    if (config.device.role != meshtastic_Config_DeviceConfig_Role_LOST_AND_FOUND && lastSentReply &&
+    if (config.device.role != meshtastic_Config_DeviceConfig_Role_LOST_AND_FOUND &&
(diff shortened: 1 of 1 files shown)


thebentern (approved):

caveman99 (changes requested):
Sorry, but this is a hard no from me. see issue reply.

Issue #11655, opened 2026-08-29
[Bug]: Heartbeat LED stops pulsing while charging (breaks external watchdog)


macvenez:
### Category

Other

### Hardware

Other

### Is this bug report about any UI (https://meshtastic.org/docs/configuration/device-uis/) component firmware?

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [ ] BaseUI

### Firmware Version

2.8.0 and earlier

### Description

As [discussed on Discord](https://discord.com/channels/867578229534359593/919642584480112750/1542815173906010122)

On our `Meshlink` board the heartbeat LED is needed to feed an external watchdog timer which resets the MCU in case no heartbeat is detected for 30s.

With the current implementation the heartbeat LED stops pulsing when the battery is charging (from USB or solar).

We believe that if a HEARTBEAT_LED is defined it should just be used for that and nothing more.
Furthermore most boards have a separate charging indicator LED.
Optionally we could think of a different LED (maybe STATUS_LED) which acts both as Heartbeat and charge indicator (maybe long pulse charging, short pulse not charging) 

This might be a possible fix: https://github.com/Ixitxachitl/meshtastic-firmware/tree/fix/heartbeat-led-while-charging
Thanks to @Ixitxachitl !

### Relevant log output

```Shell

```

caveman99:
If you define LED_Heartbeat on a separate pin you can connect another LED that exclusively has the heartbeat pulse besides the charging state. That's how every other of the 100+ boards we support does it.

macvenez:
> If you define LED_Heartbeat on a separate pin you can connect another LED that exclusively has the heartbeat pulse besides the charging state. That's how every other of the 100+ boards we support does it.

Sorry @caveman99 but I believe this is still broken, @Ixitxachitl  confirmed it [here on discord](https://discord.com/channels/867578229534359593/919642584480112750/1544264950699003934) also.

`src/modules/StatusLEDModule.cpp`
```
#if defined(LED_HEARTBEAT)
    if (power_state != charging && power_state != charged && !doing_fast_blink && !config.device.led_heartbeat_disabled) {
        if (HEARTBEAT_LED_state == LED_STATE_ON) {
            HEARTBEAT_LED_state = LED_STATE_OFF;
            my_interval = 999;
        } else {
            HEARTBEAT_LED_state = LED_STATE_ON;
            my_interval = 1;
        }
[...]

PR #11704, opened 2026-09-02
feat(status-led): add GPIO_HEARTBEAT liveness pin for external watchdogs

caveman99:
Toggles GPIO_HEARTBEAT (unset by default) every 10 s gated on nothing, so an external watchdog keeps seeing edges while charging, unlike LED_HEARTBEAT which is suppressed in the charging and charged states. Refs #11655

## Summary by CodeRabbit

- **New Features**
  - Added optional external watchdog heartbeat support.
  - When enabled, the device periodically toggles a dedicated heartbeat signal every 10 seconds to indicate that the system is running.
  - Added configuration validation to prevent the watchdog heartbeat signal from conflicting with the status LED heartbeat signal.

### DIFF
--- a/src/modules/StatusLEDModule.cpp
+++ b/src/modules/StatusLEDModule.cpp
@@ -22,6 +22,10 @@ StatusLEDModule::StatusLEDModule() : concurrency::OSThread("StatusLEDModule")
 #ifdef LED_LORA
     loraRxObserver.observe(&RadioInterface::loraRxPacketObservable);
 #endif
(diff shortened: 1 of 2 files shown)


caveman99:
Duplicate: src/watchdog/watchdogThread.cpp already provides this via HAS_HARDWARE_WATCHDOG / HARDWARE_WATCHDOG_DONE / HARDWARE_WATCHDOG_TIMEOUT_MS. No firmware change needed.

PR #11706, opened 2026-09-02
ci(test): shard the native test suite across a matrix

caveman99:
Splits the native test suite across a dynamically generated matrix and collects the results in a separate job.

## Why

`platformio-tests` was one runner walking every `test_*` suite in sequence, up to an hour of wall clock. It also compiled first and ran second: `platformio test -e coverage --without-testing` built src plus every test program, then the run relinked each suite anyway, because PlatformIO links every native test program to the same `$BUILD_DIR/$PROGNAME`. The warm-up bought a shared src build and paid for ~75 throwaway links to get it.

## Matrix

`bin/test-shards.py` derives the matrix from the tree, so adding a suite needs no CI change. It keeps the existing area rules, splits an area too big for one runner, and packs the ones too small to fill one:

```
routing-1       coverage                 7 suites
routing-2       coverage                 6
misc-1..misc-5  coverage               9,8,8,8,8
position+io     coverage                 8
admin+fuzz      coverage                 7
crypto+packets  coverage                 6
event-policy    coverage-event-policy    5
channel-table   coverage-channel-table   1
```

12 shards, largest 9. `misc` held 41 suites unsplit; `crypto` and `position` held 2 each and would have booted a runner apiece. `--max-shards` (24) bounds the row count so no branch can size the fan-out by adding directories.

Sharding is safe because isolation is already per suite, not per run: every suite executes under its own scratch `$HOME` via `bin/pio-test-isolate.sh`.

## Job graph

```
discover ─┬─> platformio-tests (matrix, fail-fast: false) ─┬─> generate-reports (verdict)
          │                                                │
simulator-tests ────────────────────────────────────────────┤
attribution-canary ─────────────────────────────────────────┘

          platformio-tests ─> Native PlatformIO Tests (required check)
```

`generate-reports` merges every shard's JUnit report, checks the union against the canonical `test/test_*` set, merges the coverage tracefiles, and states the verdict. Only the collector can see a shard that never started, which is why the whole-run attribution gate lives there.

`Native PlatformIO Tests` is kept as a job name so branch protection still has a single required check over the fan-out. The `coverage-event-policy` and `coverage-channel-table` envs and the attribution canary are now matrix rows and a parallel job instead of trailing steps.

## Toolchain

`.github/actions/setup-native-test` is a separate action, not a change to `setup-native`, which is shared with the firmware matrix. It drops the redundant second checkout, both submodules (`src/mesh/generated` is tracked, `meshtestic` is the hardware harness), `cppcheck`, and the `adafruit-nrfutil` / `poetry` / `meshtastic` pip installs, and folds in ccache and lcov. Measured saving: 65s to 43s of per-shard setup.

ccache is wired via `/usr/lib/ccache` on `$GITHUB_PATH`, with one shard flagged `cache_writer` so a single entry is saved rather than twelve racing for the key. Writes stay gated on `SAVE_CACHE`, so **no PR run can populate or benefit from it** - the first populated cache appears only after this merges to develop, and the first run to benefit is the one after that. `ccache --show-stats` prints in every shard, so the hit rate will be visible then. If it does not collapse the duplicated src compile, the ~4.5 min per-shard floor stays and `max_suites_per_shard` should go up.

## Measured

Two full runs on this branch, both with a cold ccache. Per-shard run time was flat between them (`admin+fuzz` 725s / 720s, `crypto+packets` 482s / 480s), which is the evidence that ccache contributed nothing yet.

| | run 1 (contended) | run 2 (free) |
| --- | --- | --- |
| mean shard queue | 579 s | 148 s |
| worst shard queue | 975 s | 302 s |
| shard run phase | ~13 min | ~13 min |

Queue time exceeded run time when the pool was contended. A marginal shard therefore costs a queue slot as well as ~4.5 min of duplicated setup, and the peak/off-peak spread is about 4x - larger than any shard-size tuning. `max_suites_per_shard` is a workflow input (default 10) if that needs revisiting.

Run 1 also lost a runner: one shard sat 48 minutes with its step stuck `in_progress` and its log blob never finalised. The collector correctly refused to pass on partial data. All 9 of that shard's suites pass locally and the shard is green in run 2. `timeout-minutes: 30` now caps the exposure, which would otherwise have been GitHub's 360 minute default times twelve shards.

## Hardening

Script output now feeds `strategy.matrix`, a new data path from a script into workflow control:

- The matrix row count is bounded by `--max-shards`. Without it, `--max-suites 1` produced 77 shards.
- `$GITHUB_OUTPUT` values are rejected if multi-line or empty. A newline there writes additional `key=value` entries, letting a step set outputs it never declared.
- The whole-run attribution gate fails on an empty expected set. `check-test-attribution.py --expect ""` over an empty report returns 0, which is reachable when `discover` fails and the collector still runs.
- Artifact uploads name the exact report and tracefile instead of globbing. A test binary runs arbitrary code with the workspace writable, and the collector merges whatever arrives into the report its gate reads.
- Suite names are held to `^test_[A-Za-z0-9_]+$` at the producer, including the tokens read from a fixed env's `test_filter` (PlatformIO accepts globs there). The shard word-splits with `read -ra` so a token cannot glob against the workspace either.
- `bin/test-shards.py` passes the repo path to `bin/lib/shuffle.sh` as an argument rather than interpolating it into `bash -c` source text.

Verified rather than assumed: no caller passes `secrets:` to this workflow, there is no `pull_request_target` in its trigger path, no event data is spliced into a `run:` block, and cache writes remain gated to pushes on the default branch.

## Notes

`pack()` could exceed `--max-suites`: `ceil(total / cap)` is a lower bound and whole areas do not divide, so three areas of 6 at cap 10 put 12 in one of two bins. Fixed by growing the bin count until every bin fits.

`suite_order_seed` still works but is repurposed. Shards run in parallel, so the seed now varies which suites share a shard rather than the order they run in. `pull_request` keeps the declared arrangement so a contributor's PR never goes red for a pairing they did not choose.

`bin/run-tests.sh` is untouched; local runs are unaffected.

## Attestations

CI-only change, no firmware code touched. This PR's own runs are the test: 19/19 test-native jobs green on the second run.

## Summary by CodeRabbit

- **CI/CD**
  - Native test runs now discover suites and distribute them across configurable parallel shards.
  - Added deterministic suite shuffling, per-shard attribution checks, coverage artifacts, caching, and consolidated test reports.
  - Added controls for maximum suites per shard and suite-order seeding.
  - Final CI results include shard, simulator, and attribution-canary outcomes.
  - Added streamlined setup for native test builds.

- **Documentation**
  - Updated CI and testing guidance to explain sharding, coverage reporting, and seed behavior.

### DIFF
--- a/.github/actions/setup-native-test/action.yml
+++ b/.github/actions/setup-native-test/action.yml
@@ -1,0 +1,52 @@
+name: Setup native test build
+description: >-
+  Minimal toolchain for the native test suites. Separate from setup-native, which is shared with
+  the firmware matrix and is not the place to trim things only the test path can spare.
+
+runs:
+  using: composite
+  steps:
+    # No checkout: the caller must already have one to reference this action at all.
+    - name: Setup Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: 3.x
+        cache: pip
+        cache-dependency-path: |
+          .github/actions/**
+          **.ini
+
+    - name: Install build and test dependencies
+      shell: bash
+      # C libraries are the full setup-native list: they are real link deps of the portduino HAL,
+      # the web server and the config parser. cppcheck is check_tool for `pio check`, not `pio test`.
+      run: |
+        set -euo pipefail
+        sudo apt-get -y update --fix-missing
+        sudo apt-get install -y ccache lcov \
+          libbluetooth-dev libgpiod-dev libyaml-cpp-dev libjsoncpp-dev openssl libssl-dev \
+          libulfius-dev liborcania-dev libusb-1.0-0-dev libi2c-dev libuv1-dev libcurl4-gnutls-dev
+
+    - name: Install PlatformIO
+      shell: bash
+      # No adafruit-nrfutil (nRF52 DFU), poetry, or meshtastic (the client, needed only by the
+      # simulator job), and no `pio upgrade` right after installing the current release.
+      run: |
+        set -euo pipefail
+        python -m pip install --upgrade pip
+        pip install -U --no-build-isolation --no-cache-dir "setuptools<72"
+        pip install -U platformio --no-build-isolation
+
+    - name: Configure ccache
+      shell: bash
+      # /usr/lib/ccache holds compiler-named symlinks, so PATH is the whole wiring. time_macros:
+      # ccache otherwise refuses any TU mentioning __DATE__/__TIME__, which BUILD_EPOCH disables.
+      run: |
+        set -euo pipefail
+        echo "/usr/lib/ccache" >> "$GITHUB_PATH"
+        {
+          echo "CCACHE_DIR=$HOME/.ccache"
+          echo "CCACHE_COMPRESS=1"
+          echo "CCACHE_MAXSIZE=400M"
+          echo "CCACHE_SLOPPINESS=time_macros"
+        } >> "$GITHUB_ENV"
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -632,7 +632,7 @@ The project uses GitHub Actions extensively for CI/CD. Key workflows are in `.gi
   - Includes native tests and hardware-in-the-loop testing
 
 - **`test_native.yml`** - Native platform unit tests
-  - Runs `pio test -e native`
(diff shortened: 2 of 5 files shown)


NomDeTom (commented):
Good idea in theory. Is the saving from the parallel running or the pre-building of the parts?

caveman99:
Mainly the pre-building. - Also it uses a cache once it's deployed to develop. Which will only be read on PR's - so there's no way telling if it is actually useful.

NomDeTom (approved):

PR #11713, opened 2026-09-03
Fix architecture name for Seeed Wio Tracker L2

rcarteraz:
Every other ESP32-S3 board spells it esp32-s3. Across all 109 variants declaring an architecture, this is the only one using esp32s3. The web flasher keys on the hyphenated form, so this board's build manifest carries an architecture string nothing else recognizes. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the Seeed Wio Tracker L2 build configuration to correctly identify its ESP32-S3 platform architecture.
  * Improved manifest generation so the detected device architecture takes precedence over conflicting configuration values.
  * Clarified the diagnostic message shown when the detected architecture overrides the configured value.

### DIFF
--- a/bin/platformio-custom.py
+++ b/bin/platformio-custom.py
@@ -255,8 +255,16 @@ def manifest_write(files, env, ram_bytes=None, flash_bytes=None):
         if parsed is not None and parsed != "":
             device_meta[manifest_key] = parsed
 
-    # Determine architecture once; if we can't infer it, skip manifest generation
-    board_arch = device_meta.get("architecture") or infer_architecture(env.BoardConfig())
+    # The board MCU wins over a hand-typed custom_meshtastic_architecture: only the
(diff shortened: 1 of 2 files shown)


mverch67 (approved):

mverch67:
I copied the header from another variant (tracker v2):
image

@rcarteraz I suggest to change these three as well if that's an issue in the web flasher

rcarteraz:
> I copied the header from another variant (tracker v2): image
> 
> @rcarteraz I suggest to change these three as well if that's an issue in the web flasher

Are you sure you are looking at the right branch? These are all on develop. The only one I can't confirm is the X2 since that hasn't been merged yet. 
Screenshot 2026-09-03 at 6 38 27 AM
Screenshot 2026-09-03 at 6 39 05 AM
Screenshot 2026-09-03 at 6 39 45 AM

PR #11716, opened 2026-09-03
Change default preset from LONG_FAST to LONG_TURBO

NomDeTom:
This pull request makes a change to the default US region configuration in the `RegionInfo` array. Specifically, it updates the preset used for the US region from `LONG_FAST` to `LONG_TURBO` in `RadioInterface.cpp`. This will change the default on new devices, and the default offered when a client app connects and receives a list of permitted presets.

- Changed the US region preset from `LONG_FAST` to `LONG_TURBO` in the `RegionInfo` array in `RadioInterface.cpp`

## 🤝 Attestations

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

## Summary by CodeRabbit

* **Enhancements**
  * Updated the default modem preset for the US region to use the Long Turbo configuration.

### DIFF
--- a/src/mesh/RadioInterface.cpp
+++ b/src/mesh/RadioInterface.cpp
@@ -109,7 +109,7 @@ const RegionInfo regions[] = {
         https://link.springer.com/content/pdf/bbm%3A978-1-4842-4357-2%2F1.pdf
         https://www.thethingsnetwork.org/docs/lorawan/regional-parameters/
     */
-    RDEF(US, 902.0f, 928.0f, 100, 30, false, false, PROFILE_STD, PRESET(LONG_FAST), 0),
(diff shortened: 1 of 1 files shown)


thebentern:
This will almost certainly break the upgrade path for current US LongFast users

dblaber:
> This will almost certainly break the upgrade path for current US LongFast users

In my opinion, ideally, anyone who all ready has LongFast should stay on longfast, but new devices (baseui (all ready done via #11637 ), or onboarding via api/mobile clients)  would default to longturbo. Right now its super confusing as there is a different default preset in the US if you onboard via baseui vs onboarding via android/ios app, and quite difficult to explain that to new meshtastic users.

NomDeTom:
Not my worst idea

PR #11725, opened 2026-09-04
fix(messagestore): text pool could overwrite a live record's text (InkHUD shows foreign text under your own messages)

Aubron:
Sent from Attlerock Station, Black Rock City, during Burning Man 2026, where a ThinkNode M1 on the Burning Mesh event build (2.8.0.9334cd3) started showing this on its InkHUD "Channel 1" thread: my own messages, correct sender and time, but someone else's text, usually a fragment cut mid-word ("nup in our lounge at 6:45 Esplanade").

## What's wrong

Message text lives in one fixed ring buffer (`MESSAGE_TEXT_POOL_SIZE`, 2200 bytes on nRF52840) and a `StoredMessage` only keeps an offset into it. The ring is sized for exactly `MAX_MESSAGES_SAVED` records, so as long as every allocation comes with a record it cannot be over-subscribed. Two things broke that:

- `InkHUD::Events::onReceiveTextMessage()` copied the text of **every received broadcast on every channel** into the pool for the notification cache, without adding a record. On a busy primary channel that alone laps the ring in a few dozen messages, rewriting the bytes behind whatever records a quiet secondary channel still holds.
- The wrap check was `>=`, so a full pool of maximum-length messages wrapped one message early and overwrote the oldest live record.

`storeTextInPool()` never checked what it was overwriting, so neither case produced anything but silent wrong text.

## Fix

- InkHUD's `LatestMessage` keeps its own copy of the text (`broadcastText` / `dmText`, 2 x `MAX_MESSAGE_SIZE`) instead of a pool offset. Applets read `latestMessage->textOf(isBroadcast)`. The one-time 2.7 to 2.8 message migration had the same shape (text allocated before the record was live) and is restructured the same way.
- Wrap check is `>`.
- As a guard, `MessageStore::allocText()` evicts any live record whose text range it is about to reuse, and logs it. With the two changes above this should never fire on the store's own paths; it exists so that a future recordless allocation loses a message loudly instead of corrupting one silently.

## Reproduce (before)

InkHUD device with the "Channel 1" applet enabled, joined to any busy mesh:

1. Send two or three messages from the device on a quiet secondary channel in slot 1.
2. Let 30 or so broadcasts arrive on the primary channel.
3. Open the Channel 1 applet. Your messages now carry fragments of primary-channel text.

## Validate

- `./bin/run-tests.sh -f test_message_store`. New suite; on `develop` it fails with `record N reads foreign text`, with this PR it passes.
- On a device: repeat the steps above after flashing. Records corrupted before the update stay corrupted in flash; **Menu > Wipe All Messages** once clears them.

RAM cost: 440 bytes in `Persistence` for the two text copies. `dmText` could be avoided by re-deriving the latest DM from the store at render time, at the cost of the DM applet going blank once the DM ages out of the 10-record store and of the notification popup reading a pool offset again; I kept the copy.

<details>
<summary>Notes for the next reader (or agent)</summary>

- Trace: `Router::perhapsDecode()` sets `p->channel` only from the key that decrypted the packet, so channel index and sender on the bad records are genuine. Only the text is wrong. Channel hashes on the reporting device were all distinct, so cross-channel decode was ruled out first.
- Corruption is persisted: `writeMessageRecord()` copies text out of the pool at autosave (2 h) and shutdown, so a reboot does not clean it.
- Standalone model of the allocator math, old vs new: 10 max-length inserts corrupt 3 reads; 3 records plus 100 recordless 200-byte allocations corrupt 270 reads; new allocator 0 in both. Uniform-size traffic with records almost never corrupts, which is why this shows up on quiet secondary channels next to a busy primary.
- `allocText()` only protects records already in `liveMessages`. Every caller allocates and pushes in one step; `readMessageRecord()` allocates through the instance being loaded, not the global. If the pool never allocated (`g_messagePool == nullptr`) it returns 0 before the eviction scan, otherwise every record with offset 0 would be dropped.
- `test_message_store` uses `storeText()` + `addLiveMessage()` directly (the CannedMessage path) so it needs no NodeDB and writes no files; `loadFromFlash()` in `setUp` resets pool and deque without touching flash when no store file exists. Guarded on `HAS_SCREEN || MESHTASTIC_INCLUDE_NICHE_GRAPHICS` like the module.
- Not run locally: the native suite needs a Linux host (`bin/run-tests.sh` refuses macOS, `native-macos` lacks `argp.h`). Relying on CI here. The nRF52 target `thinknode_m1-inkhud` was built from this change and flashed to the reporting device.
- A second, independent review pass (no context on how the fix was found) produced the follow-up commit: migration path, OOM-path regression, OOB read in the cache copy, per-instance flash load, eviction log, test guard.
</details>

https://claude.ai/code/session_012F3khtp9ct5VvRawcZfv1T

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved message handling when storage wraps, preventing live message text from being overwritten.
  - Ensured the latest direct and broadcast messages remain readable after message-store updates.
  - Improved loading and migration of saved messages, including safer handling of message length limits.
  - Corrected exact-fit storage behavior and removal of messages affected by reused storage space.

- **Documentation**
  - Clarified how latest message text is retained independently from shared message storage.


Xaositek (commented):
I'll leave this to @HarukiToreda , not an expert in InkHUD

HarukiToreda (approved):
Looks good to me, nice fix for the message corruption issue.

PR #11734, opened 2026-09-04
docs(agents): document the test naming rules and why camelCase does not apply

caveman99:
## Problem

Automated reviewers repeatedly flag test suite directories and `test_*` functions under `test/` as violating the `camelCase` naming convention and ask for them to be renamed. The convention in `.github/copilot-instructions.md` was written for `src/` only and never stated the test exception, so the complaint had nothing to point back at.

Acting on that request is harmful, not just noise.

## The rule

Two rules that are easy to conflate:

| Thing                           | Rule                                                                          | Examples                                                                     |
| ------------------------------- | ----------------------------------------------------------------------------- | ---------------------------------------------------------------------------- |
| Suite directory                 | Strictly `test_[a-z0-9_]+` - lowercase `snake_case`, no exceptions            | `test_gps_fix_hold/`, `test_admin_radio/`                                    |
| Test function                   | `test_` prefix, then `_`-separated segments. Case _within_ a segment is free  | `test_5byte_sequence_rejected()`, `test_getRegion_returnsCorrectRegion_US()` |
| Helpers/fixtures inside a suite | normal `src/` conventions                                                     | `makeFakePacket()`, `class FakeRadio`                                        |

For functions, only two things are forbidden: dropping the `test_` prefix, and collapsing the segments into one `camelCase` identifier. Segment case is genuinely free - the tree currently holds 743 test functions with an uppercase segment and 675 without.

## Why it is fixed

- `bin/run-tests.sh` enumerates suites with `find test -maxdepth 1 -type d -name 'test_*'`, then matches PlatformIO's per-suite verdict lines against `test_[a-z0-9_]+` (lines 204, 365, 369) - lowercase only. A suite directory containing an uppercase letter is enumerated but never matched, so the canonical cross-check reports it as missing and the run downgrades from GREEN to AMBER.
- `RUN_TEST` in `test/TestUtil.h` passes `#func` to `UnityDefaultTestRun()`, `testAssertEnvironmentIntact()` and `testStateCheckpoint()`. The identifier is the only attribution a CI log carries for a failed assertion or a dirtied sandbox, and the underscores are what keep it readable.
- It is Unity's own convention.

## Changes

- `.github/copilot-instructions.md` - new `Test naming` subsection under Naming Conventions, anchored at `#test-naming`. Marked as the single authoritative statement. The existing bullets are now explicitly scoped to `src/`.
- `AGENTS.md` - one House rules bullet, pointer only.
- `CLAUDE.md` - one quick-reference row, pointer only.
- `.coderabbit.yaml` - `test/**` path instruction. This is the one permitted copy of the rule, since a YAML instruction cannot follow a link, and it is named as such in the canonical section.

The three-file agent doc structure is unchanged: `.github/copilot-instructions.md` stays canonical, the other two stay pointers. Only the new rule is deduplicated.

Documentation-only. No firmware code touched.

## Summary by CodeRabbit

* **Documentation**
  * Clarified naming conventions for native tests.
  * Documented that test suite directories must use lowercase `test_`-prefixed names.
  * Documented that test functions must use a `test_` prefix with underscore-separated segments.
  * Clarified that source-code naming conventions do not apply to tests.
  * Added authoritative references and examples for contributors and reviewers.
  * Clarified acceptable capitalization within test-name segments and prohibited naming forms.


caveman99:
Both findings were valid. Fixed in b4fdd9168.

**"Use one definition for `snake_case`"** - correct, and the underlying error was mine. I labelled the test-function form `snake_case` while every example carried camelCase segments. Checked against the tree: 743 test functions have an uppercase segment, 675 do not, so `snake_case` was wrong for more than half of them.

Not fixed the way the comment suggested, though - rewriting the examples to full underscores would have meant renaming ~743 real test functions. The defect was the description, not the code. Split into the two rules that were being conflated:

- Suite directories: strictly `test_[a-z0-9_]+`, lowercase, no exceptions. This one really is enforced, by the verdict regex in `bin/run-tests.sh`.
- Test functions: `test_` prefix plus underscore-separated segments; case within a segment is free. Forbidden are dropping the prefix and collapsing the segments into one camelCase identifier.

**"Make the canonical-copy policy explicit"** - also correct. The policy sentence said "do not restate this anywhere else", and then the same commit restated it in `AGENTS.md` and `.coderabbit.yaml`. Resolved by naming `.coderabbit.yaml` as the single permitted copy, since a YAML instruction cannot follow a link, and trimming the `AGENTS.md` bullet to a pointer. `CLAUDE.md` was already link-only.

The `.coderabbit.yaml` entry has been updated to mirror the corrected definition, including an explicit instruction not to flag a camelCase segment inside an otherwise underscore-separated test name.

caveman99:
Valid again. Fixed in 6108d97b6.

b4fdd9168 corrected the label in the canonical section but left the identical stale one in both pointers, which is the worse half of the mistake: an agent reads the pointer first, so `CLAUDE.md` was handing out the whole-name rule that the canonical section had just rejected.

Both pointers now describe the scope rather than restating the rule:

- `CLAUDE.md`: `Test naming (the src/ rule does **not** apply)`
- `AGENTS.md`: `The src/ naming rule does not apply under test/.`

Grepped the four changed files afterwards. The only surviving `snake_case` is on the suite-directory row of the canonical table, where it is accurate: suite directories really are strictly lowercase, and there are currently zero uppercase directories under `test/`.

NomDeTom:
Can we have another rule added that tests can have more than 2 comment lines? They're usually generated in anger when something **must never happen again** and part of making sure **it never happens again** is writing a test that prevents it, and documenting **why it must never happen again**. If it's shortened to 2 lines, I'm pretty sure **it will happen again**.

PR #11743, opened 2026-09-05
feat(tmm): Antispam Trust Ladder

toddfrosty:
This change makes it more expensive to flood a public Meshtastic mesh by spinning up throwaway radio IDs. It is not a ban list, and it does not try to decide who is “allowed” on the mesh. Honest neighbors keep working as they do today. A brand-new ID is treated like a new phone number on a party line: for a short time, fewer people will repeat what it says.

## What users will notice

**On by default (new installs, and 2.8 devices whose new antispam settings were still all zero):**

- For the first **5 minutes** after a radio ID is first heard, other nodes still deliver its packets locally, but they rebroadcast them with a shorter range (`hop_limit` 2 instead of the usual 3).
- After that window, or after two nearby, long-lived neighbors vouch that they already know the ID, the hop cap is lifted.
- An unsigned vouch only ends that short “new ID” period. A **signed** vouch from a neighbor this device has actually heard for a long time can raise the ID to a higher trust level (L2). Manual key verification (L3) is still the only permanent “I checked this key out of band” step.

**Off unless an operator turns it on:**

- Per-sender **relay budget** and **NO_RELAY** gossip (`relay_budget_max_packets` stays 0). When enabled, this node will stop *repeating* a noisy sender after N rebroadcasts in a window, tell one-hop neighbors to do the same, and still show the traffic on the local device.
- Neighbor **budget gossip**, **group budget**, and **congestion hop cap**. If gossip is enabled, neighbors’ reports can only **raise** this node’s rate ceiling. They cannot tighten it.

**Also in this PR, independent of the greylist:**

- **Identity dual-read:** once a stored public key is the one that actually produces that node number, firmware refuses a replacement key that does not. That closes one “swap the key, keep the ID” hole. It is TOFU hygiene, not a second product.

## How to read the risk table

| Rating | Meaning |
| ------ | ------- |
| **High** | Default-on behavior substantially raises attacker cost on a typical public mesh |
| **Medium** | Helps in some conditions, or only if an operator enables an opt-in knob |
| **Low** | Small effect, easy to work around, or only a narrow slice of the attack |
| **None** | Not addressed here (by design or out of scope) |

These ratings are about *mesh flood / identity abuse*, not about encryption of DMs or remote-admin authorization.

## Attack vectors

| Attack | What it looks like | Mitigation | Rating |
| ------ | ------------------ | ---------- | ------ |
| **Rotating-ID flood** | Attacker changes node number every few packets so every ID looks “new,” and the mesh pays full hop-3 fan-out for each | Default 5-minute greylist + hop cap 2. Each new ID still gets through, but each copy travels one hop less. No central ban. | **Medium** |
| **Single-ID broadcast flood** | One stable ID blasts text/position/telemetry | Default hop cap only applies while that ID is still in probation. Relay budget / NO_RELAY is **off** unless set. Existing optional TMM rate limits are unchanged and still opt-in. | **Low** (default) / **Medium** (if relay budget is enabled) |
| **“Stop repeating them” censorship (NO_RELAY abuse)** | Attacker gossips `NO_RELAY` so neighbors stop relaying an innocent sender | Feature off by default. When on: ignore MQTT, ignore self-reported tenure, require this node already relayed that sender this window, require locally observed attester age, require **two** distinct claimers (or local exhaustion). Learned bits are not re-gossiped. ACKs, routing, and admin still relay. Packets addressed *to us* still ACK. | **High** against cheap fake gossip (when the knob is on); **None** when left at 0 |
| **Colluding vouches** | Several attacker radios vouch `KNOWN_SINCE` so a spam ID skips probation | Accept only if the attester was *heard on RF here* long enough (default 24 h). Need two non-co-located attesters. Unsigned vouches do **not** grant L2. Caps: 1 vouch per subject per window, 3 subjects per attester, 1 vouch packet per hour from this device. | **Medium** (raises cost; a patient, geographically spread set of old IDs can still vouch) |
| **MQTT-injected “neighborhood” claims** | Internet-side packets pretending to be RF neighbors | Attestations, vouches, and antispam accounting skip `via_mqtt` | **High** for that injection path |
| **Key swap on a key-derived ID** | Replace the stored public key with one that does not actually produce that node number | Dual-read refuses the non-deriving replacement once the stored key already derives the nodenum | **High** for that specific TOFU hole; **None** for other impersonation (cloned key, first-seen TOFU, unsigned name games) |
| **Encrypted / undecoded flood** | Ciphertext the node cannot decode, so port-based accounting never runs | Relay budget and `shouldRelay` only apply to decoded packets. Unknown-packet limiting is a separate, existing optional knob. | **None** (this PR) / **Low** if unknown-packet threshold is already enabled |
| **ACK / routing / admin spam** | Flood `want_ack`, `ROUTING_APP`, or `ADMIN_APP` | Intentionally exempt so the mesh can still ACK and route | **None** (by design) |
| **Table exhaustion** | Cycle many IDs to fill the RAM greylist and evict useful rows | nRF / small devices cap at 64 rows; others at 256. Eviction drops the oldest first-seen entry and its NO_RELAY bit. A determined rotator can churn the table. | **Low** |
| **Remote admin / PKC takeover** | Abuse the legacy admin channel or admin keys | Out of scope. This PR reverts an admin-channel policy experiment and does not change `admin_channel_enabled`. | **None** |
| **DM / PKI cryptographic spam** | Flood encrypted direct messages | Out of scope. Channel PSK and per-peer PKI are unchanged. | **None** |

## Defaults after migration

**On:** 5-minute probation, hop cap 2, vouch caps, 24 h observed-attester floor, NO_RELAY hardening knobs (only matter if relay budget is later enabled), 30-day L2 floor measured in uptime seconds.

**Off:** `relay_budget_max_packets`, `budget_gossip_enabled`, `group_budget_enabled`, `congestion_hop_cap_pct`.

Position and rate-limit settings already on a device are left alone.

## Tests

Native coverage suite (`./bin/run-tests.sh -e coverage`): 1388/1388 cases passed. New cases cover a real `NO_RELAY` send/receive round-trip, the ACK/routing allowlist, hop-cap and L2 exemption, tick-0 first-seen, MQTT skip, table eviction, hop_limit honesty, and all-zero knob migration.

## Attestations

- [x] Native C++ suite behaves as described.
- [ ] Hardware regression (Heltec V3 / T-Deck / T-Beam / RAK4631 / T-1000E): not run in this environment.

## Summary by CodeRabbit

- **New Features**
  - Added traffic-management protections, including probation controls, relay budgets, trust-based attestations, and anti-spam safeguards.
  - Improved relay decisions with hop limits, congestion controls, and protection against excessive propagation.
  - Added neighboring-node traffic observations to device telemetry.
  - Added safeguards for key-derived node identities and safer identity updates.

- **Bug Fixes**
  - Existing traffic-management configurations are preserved while previously unconfigured settings receive suitable defaults.
  - Local packet delivery and acknowledgements continue even when relaying is blocked.


caveman99:
There is an AGENTS.md file in the repository root stating a few conventions for code contributions. Your PR violates several of them.

caveman99:
Additional comment: The odds of us accepting a massive change from a first time contributor with a convolution of core behaviour changes is pretty much zero. Fan them out to separate reviewable PR's and reason about them in a few sentences, not by providing a avalance of code changes as a single PR

toddfrosty:
@caveman99 I am a contributor, this is just my personal account. I'll recreate the PR, one per level of the trust ladder and from my other account.

PR #11747, opened 2026-09-05
feat(radio): enforce ARIB STD-T108 carrier sensing, pause, and 4s max airtime for JP

t-miura:
### Summary & TL;DR
This PR implements regulatory transmission controls for the **Japan (JP) region** in accordance with **ARIB STD-T108 §3.4.1 and §3.4.2** (Ministry of Internal Affairs and Communications Notice No. 49) for 920 MHz band operation:
1. **Carrier Sensing (LBT)**: Continuous RSSI sensing for $\ge 5\text{ ms}$ before transmission with a $-80\text{ dBm}$ absolute threshold and exponential backoff (250–4000 ms) upon detecting channel activity. Carrier sense samples continuously every 250 µs (~20 samples per 5 ms window) so transmissions at SF7 symbol rates ($T_{\text{sym}} \approx 0.25\text{--}1.0\text{ ms}$) do not fall through the cracks.
2. **Inter-Transmission Pause**: Mandatory $\ge 50\text{ ms}$ quiet period enforced between consecutive bursts.
3. **4-Second Maximum Airtime Gated at Configuration Time**: Under ARIB STD-T108 §3.4.1 and Table 3-8, continuous transmission burst duration must not exceed 4000 ms:
   - **Presets (`use_preset == true`)**: Handled by companion PR [#11566](https://github.com/meshtastic/firmware/pull/11566) via `PRESETS_JP` / `PROFILE_JP`, which restricts the preset catalog to presets that never exceed 4s (excluding `LONG_SLOW` and `LONG_MODERATE`).
   - **Custom Configurations (`use_preset == false`)**: Validated in `RadioInterface::checkOrClampConfigLora()` via a static formulaic calculator `RadioInterface::calculateLoRaAirtimeMs()`. Maximum MTU (255 bytes) airtime exceeding `JapanTxHook::getMaxTxDurationMs(newRegion->code)` (4000 ms) is rejected in validation mode (`validateConfigLora()`) and auto-corrected to `LONG_FAST` with an informative client notification in clamping mode (`clampConfigLora()`).
   - **Zero Runtime Overhead**: Transmit paths and backoff retries in `JapanTxHook::beforeTransmit()` have zero serialization or static buffer overhead.
   - Only exception is airtime validation in `RadioInterface::checkOrClampConfigLora()`, while it is also skipped with near-zero cost outside JP region.
4. **Encapsulated & Isolated (`src/mesh/regulatory/`)**: Country-specific regulatory logic is housed in `src/mesh/regulatory/JapanTxHook.*`, hooking cleanly into `RadioTxHooks`.
5. **Dynamic Lifecycle & Near-zero Overhead Outside JP**: `JapanTxHook` is dynamically instantiated only when the active region is `JP`. For non-JP regions (`US`, `EU_868`, `ANZ`, etc.), `japanTxHook` is `nullptr` with 0 heap bytes allocated and 0 hook iterations in the transmit pipeline. Live region reconfiguration dynamically attaches or detaches the hook cleanly.
6. **Driver TX Lifecycle Hygiene**: Moves `postTransmit()` exclusively to the hardware `TX_DONE` interrupt path (`handleTransmitInterrupt()`), and replaces error-path `completeSending()` with direct packet release so failed SPI transmissions never falsely record airtime or increment `txGood`.
7. **Non-Breaking & 100% Backwards Compatible**: Operates seamlessly over the air with existing nodes. Wire framing, protobuf definitions, encryption, and client APIs are unchanged.

---

### Relationship & Dependency on PR #11566

This PR is closely coordinated with companion PR [#11566](https://github.com/meshtastic/firmware/pull/11566) (`fix/update-jp-lora-presets`):
* **Preset Definitions**: That PR defines `PRESETS_JP` and `PROFILE_JP`, tailoring the allowed modem presets for Japan and eliminating presets exceeding 4s (`LONG_SLOW`, `LONG_MODERATE`).
* **Preset Clamping False Alarm Fix**: Upstream `RadioInterface::checkOrClampConfigLora()` previously logged an error and executed `RECORD_CRITICALERROR(meshtastic_CriticalErrorCode_INVALID_RADIO_SETTING)` whenever a node had an unsupported preset in flash (such as migrating between regions or profiles), triggering a `"Critical fault #7"` screen latch before clamping. That PR (commit `043963a18`) addresses this by switching to `LOG_INFO`, dispatching an INFO client notification, and cleanly deferring to clamp without latching a critical error.

---

### Why is this needed? (Regulatory Context)
Under Japanese Radio Law (ARIB STD-T108), operating unlicensed 920 MHz low-power data equipment over the air mandates:
* **Carrier sense**: Verifying no RF energy $\ge -80\text{ dBm}$ exists for at least 5 ms on the unit channel prior to transmission.
* **Pause duration**: Waiting at least 50 ms after completing a burst before emitting radio waves again.
* **Transmission duration**: Continuous radio wave emission per burst must be $\le 4\text{ seconds}$ (4000 ms).

Without these controls, emitting RF in the 920 MHz band in Japan is non-compliant with local radio law. While companion work in PR #11566 restricts illegal presets for JP, this PR provides the complete carrier sensing, inter-transmission pause, and custom configuration airtime guardrails required for lawful operation in Japan.

---

### Client Behavior on Non-Legal Configurations in JP

When a client (Web, Android, iOS, CLI, or serial client) requests a LoRa configuration that is non-compliant in JP:

| Client Type | Mechanism | Clamping Behavior | User Experience |
| :--- | :--- | :--- | :--- |
| **Meshtastic Web / Android / iOS** | Local Admin (`set_config` via BLE/Serial/WiFi) | Evaluated by `clampConfigLora()`. Automatically clamped to `LONG_FAST`. | UI receives `FromRadio.clientNotification` (level INFO) explaining the adjustment. Setting fields update to `LONG_FAST`. No false critical error banner or fault screen. |
| **Python CLI (`meshtastic --set-lora ...`)** | Local Admin (Serial/TCP protobuf) | Clamped in-place to `LONG_FAST`. | Command succeeds (exit code 0). Node prints `INFO` message acknowledging the clamp. |
| **Direct Protobuf (`AdminMessage.set_config_lora`)** | Local Admin (`fromOthers = false`) | In-place clamping via `checkOrClampConfigLora(cfg, clamp=true)`. | Replaces invalid custom airtime or unsupported preset with `LONG_FAST`. |
| **Remote Admin (`fromOthers = true`)** | Remote Over-The-Air Admin packet | Evaluated by `validateConfigLora(cfg)` (`clamp=false`). Fails closed. | Request is rejected; previous valid configuration is preserved. Remote admin cannot force an illegal configuration over the air. |

---

#### 1. Hook-Based Integration (`JapanTxHook` in `src/mesh/regulatory/`)
Regional logic is isolated in `src/mesh/regulatory/JapanTxHook.*`, hooking into `RadioTxHooks`:
* `beforeTransmit()`:
  1. Evaluates the 50 ms pause deadline.
  2. Performs continuous 5 ms carrier sensing (sampling every 250 µs). Returns `PRETX_DEFER` with exponential backoff (base 250 ms, capped at 4000 ms) if busy or paused, or `PRETX_SEND` when clear.
* `postTransmit()`: Timestamps `lastTxEndTime = Time::getMillis()` upon hardware transmission completion.
* `packetReleased()`: Resets backoff state once a packet is dispatched or released.
* `getMaxTxDurationMs()`: Exposes `4000` for `JP`, `UINT32_MAX` for non-JP.
* `getTxPauseDurationMs()`: Exposes `50` for `JP`, `0` for non-JP.

#### 2. Configuration-Time Airtime Gating (`RadioInterface.cpp`)
* Added static arithmetic helper `RadioInterface::calculateLoRaAirtimeMs(bwKHz, sf, cr, payloadLen)` implementing the standard Semtech time-on-air formula (including preamble and Low Data Rate Optimization).
* In `RadioInterface::checkOrClampConfigLora()`, custom settings for region `JP` are evaluated for worst-case payload (255 bytes). If airtime $> 4000\text{ ms}$:
  - Validation mode (`clamp = false`) rejects the change.
  - Clamping mode (`clamp = true`) auto-corrects to `LONG_FAST` and emits an INFO-level client notification.
  - Remote admin (`fromOthers = true`) rejects the update and restores `oldLoraConfig`.

#### 3. Dynamic Lifecycle & Near-zero Cost Outside JP
* `initJapanTxHook()` reconciles hook presence against `isJapanRegion()`.
* Wired into both initial boot (`initLoRa()`) and runtime reconfiguration (`RadioInterface::reconfigure()`).
* When active region is non-JP, `japanTxHook` is `nullptr` and unlinked from `RadioTxHook::hookList`.

#### 4. Radio Driver TX Lifecycle Hygiene
* **`postTransmit()` on true `TX_DONE` only**: `postTransmit()` is placed strictly inside `handleTransmitInterrupt()`, ensuring transmission-complete hooks are never fired on failed transmit attempts or mid-air aborts (`setStandby()`).
* **Error-path cleanup without false accounting**: In `RadioLibInterface::startSend()`, if `startTransmit()` fails (e.g. SPI error), the packet is released cleanly via `packetPool.release()` and `RadioTxHooks::packetReleased()` without calling `completeSending()`. This eliminates spurious airtime recording and false `txGood++` increments on hardware start errors.

#### 5. Physical RF Sensitivity Realism ($[-192, -1]\text{ dBm}$)
* Valid RSSI range is established at $[-192, -1]\text{ dBm}$ (`RSSI_VALID_MIN = -192`), ensuring that quiet RF readings from real hardware are recognized as clean/free channel states:
  * Semtech **SX1261/SX1262** sensitivity reaches $-141\text{ dBm}$ (instantaneous RSSI register rounds to $-128\text{ dBm}$).
  * Semtech **SX1276/SX1278** computes instant RSSI as $\text{offset} + (\text{reg} / 2)$ with a $-164\text{ dBm}$ HF offset in RadioLib, returning down to $-164\text{ dBm}$.
* Rejects uninitialized fallback values ($0\text{ dBm}$) and positive saturation numbers ($> 0\text{ dBm}$), preventing uninstrumented hardware from locking up in false carrier sense.
* Ignores large negative driver errors ($< -192\text{ dBm}$, e.g. `RADIOLIB_ERR_SPI_CMD_INVALID = -706`).

#### 6. Rollover & Sentinel Safety
* Strictly compliant with Meshtastic's `Throttle` APIs (`Throttle::hasElapsed`, `Throttle::deadlinePassedAt`). Zero direct arithmetic comparisons on `millis()`.
* Single-snapshot clock reads (`nowAfter = Time::getMillis()`) prevent race conditions across long loops.
* Deadline zero-wraparounds (`tx_after == 0`) are remapped to `1` so deadlines on 32-bit wraps are not interpreted as "inactive".

---

#### 1. Automated Tests (`pio test -e native-macos`)
* **`test_japan_compliance`**: **36/36 PASSED** (Verified clean on native and with `-fsanitize=address` ASan/LSan active):
  * Free/busy boundary threshold test ($-81\text{ dBm}$ pass, $-80\text{ dBm}$ busy).
  * Continuous 5 ms window with mid-window spike detection (250 µs sampling resolution across 20 samples).
  * Exact 20-sample count verification over the 5 ms carrier sense window.
  * Valid RSSI bounds test across $-1$ down to $-192\text{ dBm}$ ($-128$, $-141$, $-150$, $-164$, $-192\text{ dBm}$).
  * Rejection of $0\text{ dBm}$, positive values, and driver errors.
  * Mandatory 50 ms pause hold and deferral.
  * Start transmit error isolation (failed starts do not trigger pause).
  * Static formula airtime validation:
    * `LONG_FAST` $\le 4000\text{ ms}$ verified.
    * `LONG_SLOW` $> 10000\text{ ms}$ verified.
    * Custom compliant (SF10, BW125, CR5) $\le 4000\text{ ms}$ verified.
  * Configuration validation and clamping:
    * Rejection of non-compliant custom settings (SF12, BW125, CR 4/8) in JP.
    * Acceptance of compliant custom settings (SF10, BW125, CR 4/5) in JP.
    * Auto-clamping of non-compliant custom settings to `LONG_FAST` in JP.
    * Non-JP bypass verification (US allows custom settings $> 4\text{s}$).
  * Dynamic hook attachment/detachment on region switch and runtime `RadioInterface::reconfigure()`.
  * Stack allocation safety: stack-allocated hook instances do not modify or clear the global pointer.
  * Exponential backoff computation (base 250 ms, max 4000 ms).
  * Millisecond rollover and sentinel wrap handling.
* **Regression Suites**: Verified clean across `test_default`, `test_mesh_beacon`, `test_radio`, `test_airtime`, `test_meshpacket_queue` (175/175 passing).

#### 2. Live Hardware Validation (Serial Capture on Heltec V3)
Verified on physical hardware (Heltec V3 `IGUS-MBGW1`) running firmware build `2.8.1.05030e5`:
* **Boot & Radio Initialization**:
  ```text
  INFO  | Wanted region 5, using JP
  INFO  | Start meshradio init
  INFO  | Radio freq=923.375, config.lora.frequency_offset=0.000
  INFO  | Set radio: region=JP, name=LongFast, config=0, ch=11, power=13
  INFO  | numFreqSlots: 12 x 250.000kHz, channel_num: 12
  INFO  | SX126x init result 0
  ```
* **Clean Carrier Sense (5 ms window) & On-Air Transmission**:
  Direct message transmission to node `IGPL` (`0x0a9978fc`):
  ```text
  DEBUG | JP LBT: carrier sense clear (5ms window), transmit permitted for packet 0xd68039fa
  INFO  | Started Tx (id=0xd68039fa fr=0x7a6f98a7 to=0x0a9978fc, transport = 0, WantAck=0, HopLim=3 Ch=0 encrypted len=52 rxtime=1788860244 hopStart=3 relay=0xa7 priority=70)
  ```
* **Continuous Carrier Sensing & Collision Avoidance in active mesh traffic**:
  ```text
  DEBUG | ... [RadioIf] JP LBT: carrier sensed during 5ms window (RSSI -42 dBm >= -80 dBm)
  DEBUG | ... [RadioIf] JP LBT: channel busy (attempt 2), backing off 729ms for packet 0x4319ac10
  DEBUG | ... [RadioIf] Lora RX (id=0xc7360635 ... rxRSSI=-43)
  ```
* **50 ms Pause Deferral**:
  ```text
  DEBUG | ... [RadioIf] Completed sending (id=0x7d86e641 ...
  DEBUG | ... [RadioIf] JP LBT: deferring packet 0x869b4e43 for mandatory 50ms pause (remaining 50ms)
  ```
* **Zero heap leakage** across queue drains and retransmissions.

## 🤝 Attestations

  - [x] Heltec (Lora32) V3
  - [x] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631
    - [x] Seeed Studio Xiao ESP32-S3
    - [x] Seeed Studio Xiao nRF52840
    - [x] Seeed Studio Wio Tracker L1 Pro
    - [x] Heltec Mesh Node T114
    - [x] Heltec Vision Master E213 
    - [x] LilyGo T-Echo
    - [x] LilyGo T-Watch S3 
    - [x] M5Stack Unit C6L
    - [x] M5Stack Cardputer ADV & Cap LoRa-1262 
    - [x] WIZnet W5500-EVB-Pico2 & Wio-SX1262 

### `bin/test-native-docker.sh --rebuild`

```
=============== 1417 test cases: 1417 succeeded in 00:25:29.246 ===============
```

### `bin/run-test.sh`
```
# All cases passed, while i forgot to keep the log...
```

## Summary by CodeRabbit

- **New Features**
  - Added Japan-specific radio compliance controls, including listen-before-talk carrier sensing, transmission spacing, exponential backoff, and maximum airtime limits.
  - Controls activate automatically for the Japan region and deactivate when switching regions.
  - Custom Japan LoRa configurations exceeding airtime limits are clamped or rejected during validation.
  - Added transmit lifecycle handling for completed, cancelled, deferred, and failed packets.

- **Bug Fixes**
  - Improved transmit deadline handling and simulated-radio behavior for throttling and regulatory checks.


t-miura:
@coderabbitai review

src/mesh/RadioLibInterface.cpp
```diff
@@ -587,6 +593,7 @@ void RadioLibInterface::completeSending()
         if (!isFromUs(p))
             txRelay++;
         printPacket("Completed sending", p);
+        RadioTxHooks::postTransmit(this, p);
```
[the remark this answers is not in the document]
t-miura: @coderabbitai review as i fixed the issue.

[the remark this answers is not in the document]
t-miura: @coderabbitai review

t-miura:
@coderabbitai full review

src/mesh/JapanTxHook.h
```diff
@@ -0,0 +1,66 @@
...
+class RadioInterface;
+
+/**
+ * RadioTxHook implementing ARIB STD-T108 compliance for Japan (JP) region.
+ * Enforces:
+ *  1. Continuous RSSI carrier sensing for >= 5 ms at -80 dBm threshold before transmission,
+ *     with exponential backoff (500 ms - 4000 ms) when busy.
+ *  2. Minimum 50 ms inter-transmission pause between consecutive transmissions.
+ *  3. Complete isolation for non-JP regions (zero delay, zero RSSI gating).
+ */
```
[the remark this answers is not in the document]
t-miura: @coderabbitai Since our radio laws are strict and complex(in other word, weird), the contents in this header may unclear to whom reading the code without any references(especially whom outside the Japan), so will keep as-is.

src/mesh/regulatory/JapanTxHook.cpp
caveman99, line 1: Please move the country specific files to a new subdir /src/mesh/regulatory/

src/mesh/JapanTxHook.h
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+  private:
+    uint32_t lastTxEndTime = 0;
+    uint32_t busyCount = 0;
+};
+
+extern JapanTxHook *japanTxHook;
+void initJapanTxHook();
+uint32_t getTxPauseDurationMs();
```
caveman99: What's the point of this 'free' function nobody except the test calls?

src/mesh/JapanTxHook.h
```diff
@@ -0,0 +1,66 @@
...
+  public:
+    static constexpr int16_t CARRIER_SENSE_THRESHOLD_DBM = -80;
+    static constexpr uint32_t CARRIER_SENSE_TIME_MS = 5;
+    static constexpr uint32_t INTER_TX_PAUSE_MS = 50;
+    static constexpr uint32_t BACKOFF_BASE_MS = 500;
+    static constexpr uint32_t BACKOFF_MAX_MS = 4000;
+    static constexpr int16_t RSSI_UNAVAILABLE = 0;
+    // Lower bound accommodates SX126x (-141 dBm), SX127x (-164 dBm HF offset), and LR11x0, above SPI errors (< -500).
+    static constexpr int16_t RSSI_VALID_MIN = -192;
+    static constexpr int16_t RSSI_MIN_VALID_DBM = RSSI_VALID_MIN;
```
caveman99: What is the point of RSSI_VALID_MIN and RSSI_MIN_VALID_DBM when they are aliased here?

src/mesh/JapanTxHook.cpp
```diff
@@ -0,0 +1,139 @@
...
+        return true;
+    const uint32_t start = Time::getMillis();
+    while (!Throttle::hasElapsed(start, CARRIER_SENSE_TIME_MS)) {
+        int16_t rssi = iface->getCurrentRSSI();
+        if (isValidRssi(rssi) && rssi >= CARRIER_SENSE_THRESHOLD_DBM) {
+            LOG_DEBUG("JP LBT: carrier sensed during 5ms window (RSSI %d dBm >= %d dBm)", rssi, CARRIER_SENSE_THRESHOLD_DBM);
+            return false;
+        }
+        delay(1);
+#ifdef PIO_UNIT_TESTING
```
caveman99: Please dont move test scaffolding into the code. Advance the clock from getCurrentRSSI() instead

src/mesh/RadioInterface.cpp
```diff
@@ -372,6 +373,7 @@ extern SPIClass SPI1;
 
 std::unique_ptr<RadioInterface> initLoRa()
 {
+    initJapanTxHook();
```
caveman99, line 376: You init the japan TX Hook for every region. Not a biggie, but it contradicts the claim of zero cost outside JP.

src/mesh/JapanTxHook.cpp
```diff
@@ -0,0 +1,139 @@
...
+    if (!iface)
+        return true;
+    const uint32_t start = Time::getMillis();
+    while (!Throttle::hasElapsed(start, CARRIER_SENSE_TIME_MS)) {
+        int16_t rssi = iface->getCurrentRSSI();
+        if (isValidRssi(rssi) && rssi >= CARRIER_SENSE_THRESHOLD_DBM) {
+            LOG_DEBUG("JP LBT: carrier sensed during 5ms window (RSSI %d dBm >= %d dBm)", rssi, CARRIER_SENSE_THRESHOLD_DBM);
+            return false;
+        }
+        delay(1);
```
caveman99: This literally sleeps for 1ms. So you take 5 samples in the 5ms (blocking) window. Symbol length at SF7 is roughly 1ms as well so that can fall through the craks. I would swap that for delayMicroseconds(250) to come nearer to the requirement of continuously sampling the airwave.

caveman99 (changes requested):

src/mesh/regulatory/JapanTxHook.cpp
t-miura: moved files to the new directory, it'll be handy if other region requires their own hooks, too.

we're the first region to have this, yay...(at -80dBm), haha...(at -79dBm)

src/mesh/JapanTxHook.h
t-miura: it must be used `JapanTxHook::getTxPauseDurationMs()` directly, so now does with removal of free function.

src/mesh/JapanTxHook.h
t-miura: Removed the redundant alias `RSSI_MIN_VALID_DBM` and kept only `RSSI_VALID_MIN = -192`.

src/mesh/JapanTxHook.cpp
t-miura: Thanks! i'm not still good at thinking about unit-test....and/or keep code in its style.

stripped `#ifdef PIO_UNIT_TESTING out of performCarrierSense()`,
moved the virtual clock advancement into `MockRadioInterface::getCurrentRSSI()` in `test_main.cpp`,
resulting no test scaffolding in main code.

src/mesh/RadioInterface.cpp
t-miura: Indeed, how did i missed that...i don't want to consume any extra heap...

So, to make it really zero cost, updated initJapanTxHook() to dynamically reconcile based on isJapanRegion().

This is also wired into RadioInterface::reconfigure(), so even on runtime region changes dynamically attach or detach the hook.

src/mesh/JapanTxHook.cpp
t-miura: agree with that, so now it uses `delayMicroseconds(250)`, 250us ought to be enough for anybody.

src/mesh/RadioInterface.cpp
```diff
@@ -941,6 +943,7 @@ RadioInterface::RadioInterface()
 
 bool RadioInterface::reconfigure()
 {
+    initJapanTxHook();
```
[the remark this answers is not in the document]
t-miura: @coderabbitai fixed, review.

t-miura:
Claude pointed out one "Critical" issue with this pr, which is (unfortunately) intended:

```
2. performCarrierSense() busy-waits 5 ms on the radio worker thread

JapanTxHook.cpp:73-85

The 20× delayMicroseconds(250) loop runs inside onNotify() (the NotifiedWorkerThread). This blocks ISR_RX notification processing for the entire 5 ms window. At SF7 (symbol ≈0.25 ms) the preamble fits inside one sampling interval, so the lock is not benign. A stateful carrier-sense approach across multiple PRETX_DEFER re-entries would eliminate the hold.
```

It is required to measure the RSSI "continously" for 5ms, so it HAS to be blocked...

t-miura:
@coderabbitai resume

caveman99:
> Claude pointed out one "Critical" issue with this pr, which is (unfortunately) intended:
> 
> ```
> 2. performCarrierSense() busy-waits 5 ms on the radio worker thread
> 
> JapanTxHook.cpp:73-85
> 
> The 20× delayMicroseconds(250) loop runs inside onNotify() (the NotifiedWorkerThread). This blocks ISR_RX notification processing for the entire 5 ms window. At SF7 (symbol ≈0.25 ms) the preamble fits inside one sampling interval, so the lock is not benign. A stateful carrier-sense approach across multiple PRETX_DEFER re-entries would eliminate the hold.
> ```
> 
> It is required to measure the RSSI "continously" for 5ms, so it HAS to be blocked...

It has a point there though - i didn't notice the blocking was in ISR context. I'll have a look at this.

t-miura:
> It has a point there though - i didn't notice the blocking was in ISR context. I'll have a look at this.

sorry, my comment should be confused you, ISR is not blocked even during `delayMicroseconds()`,
so HW INT like from DIO1 on SX1262 will still fire immidiately,
and FIFO buffer on radio chip ensures no RX packet drops with this 5ms measurement window.

And, Tx during active Rx can't happen as there's been `busyRx（canSendImmediately()` to block such collision,
so the context/situation Claude pointed out can't be happen.

Also when if new packets comes in during this 5ms period in Tx phase,
this LBT(when signal  > -80dBm) or existing CAD(when signal < -80dBm) will block the Tx,
and radio will do Rx instead, and back to Tx later with all the hooks again.

test/test_japan_compliance/test_main.cpp
```diff
@@ -0,0 +1,774 @@
...
+class MockRadioInterface : public RadioInterface
+{
+  public:
+    int16_t rssiToReturn = -100;
+    std::vector<int16_t> rssiSequence;
+    size_t sequenceIndex = 0;
+
+    int16_t getCurrentRSSI() override
+    {
+        Time::advanceTestMillis(1);
```
caveman99: That test steps wrong now - the actual code has 250us sampling, but this is still 1ms. Also the fixture just feeds 5 values instead of 20.

src/mesh/regulatory/JapanTxHook.cpp
```diff
@@ -0,0 +1,145 @@
...
+}
+
+RadioTxHook::PreTxAction JapanTxHook::beforeTransmit(RadioInterface *iface, meshtastic_MeshPacket *p)
+{
+    if (!isJapanRegion() || !p)
+        return PRETX_SEND;
+
+    if (iface) {
+        const uint32_t airtimeMs = iface->getPacketTime(p);
+        if (airtimeMs > MAX_TX_DURATION_MS) {
```
caveman99: shouldn't this use getMaxTxDurationMs() instead?

src/mesh/regulatory/JapanTxHook.cpp
```diff
@@ -0,0 +1,145 @@
...
+    return true;
+}
+
+RadioTxHook::PreTxAction JapanTxHook::beforeTransmit(RadioInterface *iface, meshtastic_MeshPacket *p)
+{
+    if (!isJapanRegion() || !p)
+        return PRETX_SEND;
+
+    if (iface) {
+        const uint32_t airtimeMs = iface->getPacketTime(p);
```
caveman99: getPacketTime(p) is expensive. It runs a full protobuf encode into the shared static bytes buffer on every TX attempt including every backoff retry. I'd rather gate the max packet tx time on preset change (validateConfigLora) and disallow the set, rather than check it on every TX. This way the user is notified of the constraint as well instead of dropping a packet silently.

t-miura:
these logs should verifies safe behavior.

Whenever an incoming packet's preamble/header is detected, `isActivelyReceiving()` becomes `true`.
`canSendImmediately()` immediately guards the transmit pipeline, deferring any TX attempts before `beforeTransmit()` can even be invoked:

```text
DEBUG | 03:26:03 [RadioIf] JP LBT: carrier sensed during 5ms window (RSSI -69 dBm >= -80 dBm)
DEBUG | 03:26:03 [RadioIf] JP LBT: channel busy (attempt 1), backing off 863ms for packet 0xff322c1d
WARN  | 03:26:04 [RadioIf] Can not send yet, busyRx
WARN  | 03:26:04 [RadioIf] Can not send yet, busyRx
WARN  | 03:26:04 [RadioIf] Can not send yet, busyRx
... (TX repeatedly held while 1098ms frame is on the air) ...
DEBUG | 03:26:04 [RadioIf] Lora RX (id=0x7d86e641 rxSNR=13.25 rxRSSI=-52)
DEBUG | 03:26:04 [RadioIf] Packet RX: 1098ms
DEBUG | 03:26:04 [Router]  decoded message (id=0x7d86e641)
```

also as as an insightful example counitnues, this time, the packet wanted to Tx was from MQTT,
and other node was bit eariler to relay, making this node's Tx attempt duplicated, and Tx from this node has been sucessfully cancelled.

```
DEBUG | 09:55:25 112 [Router]  enqueue for send (id=0x4e556f26 ...)
WARN  | 09:55:25 112 [RadioIf] Can not send yet, busyRx
WARN  | 09:55:26 113 [RadioIf] Can not send yet, busyRx
WARN  | 09:55:26 113 [RadioIf] Can not send yet, busyRx
WARN  | 09:55:26 113 [RadioIf] Can not send yet, busyRx
WARN  | 09:55:26 113 [RadioIf] Can not send yet, busyRx
DEBUG | 09:55:26 113 [RadioIf] Lora RX (id=0x4e556f26 ... rxSNR=11.25 rxRSSI=-49 ...)
DEBUG | 09:55:26 113 [RadioIf] Packet RX: 410ms
DEBUG | 09:55:26 113 [Router]  cancelSending id=0x4e556f26, removed=1
```

caveman99 (changes requested):
There's one more thing when BeaconMode is enabled, switching from and to JP regions disables the hooks or races. but this is a design decision that needs to be handled outside of this PR.

src/mesh/regulatory/JapanTxHook.cpp
t-miura: this sounds better, and i found no existing function to calculate the maximum airtime(data rate exists though),
so this will introduce a new function, maybe useful outside JP too(so i hope).

t-miura: for presets validation, #11566 will remove two 4sec-exceeding presets.
and its preset list will tied only to JP region w/ clamping to LongFast w/o CRITICAL_ERROR,
so the remaining case we need to validate is non-preset mode, and i'm working on that way!

p.s.: with only this PR merged, it'll shows `Critical Fault #7` on preset clamping as #11566 will fix that behavior,
so merging both at same time will be the best for JP users.

test/test_japan_compliance/test_main.cpp
t-miura: Test suite has now updated to use correct value,
alongside with new tests that verifies clear 5 ms window executes exactly 20 sample iterations before completing.

src/mesh/regulatory/JapanTxHook.cpp
t-miura: Whole flow has been refined in following way:
- On non-preset mode, `validateConfigLora` will validate, reject/gate the configuration on LoRa parameter changes.
  - needs #11566 for updated preset list w/ no false alarm on clamping, if it happens.
- No more Tx-time calculation required, as such combination of parameter is now impossible to be exist on Tx phase.
- Resulting no runtime check for packet size/airtime, removing whole block.
- This change introduces new function `RadioInterface::calculateLoRaAirtimeMs()`, as using RadioLib's function needs RadioLib driver init, resulting "chicken-egg/opener-in-can" scenario, so it needs to be there as safe choice.

src/mesh/regulatory/JapanTxHook.cpp
t-miura: since new checking flow is no longer required on run-time, this has been removed.

src/mesh/regulatory/JapanTxHook.cpp
```diff
@@ -0,0 +1,136 @@
...
+}
+
+bool JapanTxHook::performCarrierSense(RadioInterface *iface)
+{
+    if (!iface)
+        return true;
+    const uint32_t start = Time::getMillis();
+    while (!Throttle::hasElapsed(start, CARRIER_SENSE_TIME_MS)) {
+        int16_t rssi = iface->getCurrentRSSI();
+        if (isValidRssi(rssi) && rssi >= CARRIER_SENSE_THRESHOLD_DBM) {
```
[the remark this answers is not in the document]
t-miura: good catch, implemented in fail-safe way,back-off the packet if RSSI sample is not valid and/or no sample receieved.

t-miura:
ah, the new notification for client is not very user-frendly, fixing...

t-miura:
@coderabbitai

> Nitpick comments:
> In `@test/test_japan_compliance/test_main.cpp`:
> - Around line 267-268: Update the backoff assertions around beforeTransmit to
> capture Time::getMillis() immediately after each call, then assert pkt.tx_after
> minus that captured start time falls within the expected delay range. Apply the
> same rollover-safe calculation to both assertion pairs, including the later
> checks currently at lines 275–276.

Already addressed in commit 0fb1dfa0a (`test/test_japan_compliance/test_main.cpp` lines 267–270 and 277–280). Both assertion pairs snapshot `Time::getMillis()` immediately after `beforeTransmit()` returns and assert that the unsigned delta `pkt.tx_after - now` falls within the expected backoff range, making the checks rollover-safe.

t-miura:
@coderabbitai

Your finding is incorrect.
Both overloads of `getMaxTxDurationMs` are explicitly declared in `JapanTxHook.h` (lines 34–35) and defined in `JapanTxHook.cpp` (lines 48–56):
- `getMaxTxDurationMs()`: evaluates the currently active hardware/runtime region via `isJapanRegion()`.
- `getMaxTxDurationMs(meshtastic_Config_LoRaConfig_RegionCode region)`: evaluates a prospective region code directly.

In `RadioInterface::checkOrClampConfigLora()`, we are validating `loraConfig` (a prospective configuration that is not yet applied to `config.lora.region` or `myRegion`).

If a node transitions from another region (e.g., US or UNSET) to JP with custom LoRa parameters, calling the no-argument version would evaluate `isJapanRegion()` as false and return `UINT32_MAX`, allowing illegal >4s settings to bypass validation.

Passing `newRegion->code` ensures prospective region validation works reliably regardless of current runtime state. Both `test_japan_compliance` and firmware targets (`heltec-v3`) compile and pass.

t-miura:
sorry, i forgot to fix false-positive secret detecion on `trunk`.
root cause was exact same as what happened iin #11480, my new tests were named in  `test_`, followed by exactly 35 characters.

i've renamed affected four tests, and now it passes without throwing any false positives, no new issues reported from trunk as well, so I believe it's ok as of now.

before:
```
% trunk check test/test_japan_compliance/test_main.cpp

  ISSUES

test/test_japan_compliance/test_main.cpp:158:0
 158:0  high  Secret detected  trufflehog/Lob
 552:0  high  Secret detected  trufflehog/Lob
 766:0  high  Secret detected  trufflehog/Lob
 778:0  high  Secret detected  trufflehog/Lob
```

after:
```
% trunk check test/test_japan_compliance/test_main.cpp
# Checked 1 file
# ✔ No issues
```

test on native-macos:
```
=================================== SUMMARY ===================================
Environment    Test                   Status    Duration
-------------  ---------------------  --------  ------------
native-macos   test_japan_compliance  PASSED    00:02:24.188
================= 37 test cases: 37 succeeded in 00:02:24.188 =================
```

t-miura:
also found a build issue with native-wasm/portduino as it lacked new directory `src/mesh/regulatory` in its whitelist.
now it's added, and CI run on my forked repo ran with all green(except for skipped ones).

src/mesh/regulatory/JapanTxHook.h
```diff
@@ -0,0 +1,60 @@
...
+
+    static uint32_t getTxPauseDurationMs();
+    static uint32_t getTxPauseDurationMs(meshtastic_Config_LoRaConfig_RegionCode region);
+    static uint32_t getMaxTxDurationMs();
+    static uint32_t getMaxTxDurationMs(meshtastic_Config_LoRaConfig_RegionCode region);
+    static bool isJapanRegion();
+    static uint32_t computeBackoffMs(uint32_t count);
+
+    // Valid RSSI must be strictly negative (< 0 rejects 0 fallback and positive saturation) and >= -192 dBm.
+    static bool isValidRssi(int16_t rssi) { return rssi < 0 && rssi >= RSSI_VALID_MIN; }
```
t-miura: should be using `RSSI_UNAVAILABLE` for higher bound

src/mesh/regulatory/JapanTxHook.h
```diff
@@ -0,0 +1,60 @@
...
+class JapanTxHook : public RadioTxHook
+{
+  public:
+    static constexpr int16_t CARRIER_SENSE_THRESHOLD_DBM = -80;
+    static constexpr uint32_t CARRIER_SENSE_TIME_MS = 5;
+    static constexpr uint32_t INTER_TX_PAUSE_MS = 50;
+    static constexpr uint32_t MAX_TX_DURATION_MS = 4000;
+    static constexpr uint32_t BACKOFF_BASE_MS = 250;
+    static constexpr uint32_t BACKOFF_MAX_MS = 4000;
+    static constexpr int16_t RSSI_UNAVAILABLE = 0;
```
t-miura, line 20: on M5Stack C6L(which has known for having STRONG LNA enabled by default), nearby node's RSSI measured 0dBm, so this value is not right and may cause mismeasurement.

Upon SX1262/SX1276 datasheet, maximum allowed input power is 10dBm,
Also by JP radio law, maximum output power is 13dBm.
So, turning up it to either 10 or 13dBm ensures "very close and strong" signal doesn't handled as "oops, RSSI measurement failed", needs fix ASAP.

t-miura: note: current code does accept 0dBm, as it does `rssi < 0(needs replace with defined value`, but it's still not quite right.

test/test_japan_compliance/test_main.cpp
```diff
@@ -0,0 +1,807 @@
...
+    TEST_ASSERT_EQUAL_INT(RadioTxHook::PRETX_DEFER, hook.beforeTransmit(&radio, &pkt));
+    TEST_ASSERT_EQUAL_UINT32_MESSAGE(1, pkt.tx_after, "Wrapped deadline 0 must be remapped to 1");
+}
+
+void test_is_valid_rssi_helper(void)
+{
+    // Unavailable / Default
+    TEST_ASSERT_FALSE(JapanTxHook::isValidRssi(0));
+    // Positive values (garbage or error)
+    TEST_ASSERT_FALSE(JapanTxHook::isValidRssi(1));
```
t-miura, line 510: these two lines needs update when bumping up the upper bound of isValidRssi()

src/mesh/regulatory/JapanTxHook.h
t-miura: while just raising 0dBm uppoer bound may cause another issue,
as actual "rssi unavailable" does also returns 0dBi, no way to distinguish either unavailable or really high RSSI.

But, anyway we'd better defer the TX until RSSI reading become available, or situation settles down,
so i think it's not a huge problem(unlike tweaking lower bound that may have more blast radius)

test/test_japan_compliance/test_main.cpp
t-miura: maybe keep 0 as boundry-test value, but opposite assertion, 1 might be the same.
adding `TEST_ASSERT_FALSE` with higher than upper bound RSSI values are what actually I need to add.

src/mesh/regulatory/JapanTxHook.h
t-miura: how about...setting "no rssi available now" value to something very high, like 255?
255 dBm(almost 3^22W) can't be in RF frontend(even TX side, unless someone used microwave heater as LoRa radio),
so that value can't exist in the wild, while some lower/negative values are used/reserved as RadioLib's error code, so using positive number is the way i guess.

t-miura: https://github.com/meshtastic/firmware/blob/8ef996f5b474300c9f06128beff5ce36369f812e/src/mesh/RadioLibInterface.cpp#L278
changing initial value will cause conflict with these codes, so...maybe keep 0 as initial value,
and JPhook will just wait til rssi readings become available, or lower than upper bound,
and then compare with -80dBm CS threshold, maybe that's the fail-safest way.

t-miura: as for now, i'll keep 0 dBm as "invalid RSSI", even if it's actual measurement.
since it'll be back-off'ed either way, as 0 dBm is way higher than JP LBT threshold: -80dBm,
and node will attempts to remeasure the RSSI on next path anyways.

while there's slight deference in the path,but i think treating "unavailable" magic number as valid value has more potential to cause unwanted behavior, so i'll resolve this self-review for now.

test/test_japan_compliance/test_main.cpp
t-miura: won't touch for now, but if we change the upper bound for valid RSSI values, they needs to be changed as well.

src/mesh/regulatory/JapanTxHook.h
t-miura: Only this should be applied for now, working.

t-miura: done.

PR #11760, opened 2026-09-07
Revert "fix(ci): build with pioarduino core instead of upstream platformio"

vidplace7:
Reverts meshtastic/firmware#11759 , it's a no-op as is (this change should have been made in `gh-action-firmware`, working with @caveman99 on that now)

setup-action is used by things like the test suite (which should be using platformio)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ESP build reliability by using the standard PlatformIO package and upgrading it during setup.
  - Pinned the SCons build tool to a compatible version to prevent ESP builds from failing before compilation.


vidplace7:
https://github.com/meshtastic/gh-action-firmware/pull/61
Proposed fix in the build container

PR #11761, opened 2026-09-07
fix(rp2xx0): match the real library name when ignoring iLabs_Hearth

caveman99:
Follow-up to #11757, which did not work. `lib_ignore` matches the name declared in the library manifest, and iLabs_Hearth's `library.properties` says `name=iLabs Hearth` with a space. #11757 used the directory name, so the entry never matched and the Pico builds still pull the library in and still fail on its `ESP_SERIAL_PORT` #error.

Confirmed from a build after #11757 merged: `picow` compiled 653 objects including `iLabs_Hearth/MatterEndpoints/*.cpp.o` before hitting `*** [.pio/build/picow/libd9c/iLabs_Hearth/Hearth.cpp.o] Error 1`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected library exclusion configuration for RP2040 and RP2350 platforms, improving build and dependency handling.

### DIFF
--- a/variants/rp2040/rp2040.ini
+++ b/variants/rp2040/rp2040.ini
@@ -27,7 +27,7 @@ lib_ignore =
   lvgl
   ; Ships its own Preferences.h, which the LDF matches against the ARCH_ESP32-only
   ; include in NodeDB.cpp; compiling it then trips its own "not a Challenger" #error.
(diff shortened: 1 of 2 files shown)


vidplace7 (approved):
Looks like it worked this time 👍

PR #11765, opened 2026-09-07
fix(traffic): round the default position dedup window to 6h, derive the floor from it

garthvh:
Rounds the default Traffic Management position dedup window from 5 h to 6 h and derives the stationary/fixed-position broadcast floor from it instead of pairing the two by hand.

- `default_traffic_mgmt_position_min_interval_secs`: 5 h -> 6 h (21600 s).
- `default_position_stationary_broadcast_secs`: 6 h -> window + one dedup tick (6 h 06 min). Receivers measure position age in whole 6-minute ticks (`kPosTimeTickMs`), so one tick is the exact margin that guarantees a stationary node's refresh clears its neighbours' window instead of being dropped as a duplicate. The previous margin was a hand-picked hour.
- `static_assert` in `shouldDropPosition()` enforces floor >= window + tick.
- Role caps (tracker 1 h, lost-and-found 15 min) are unchanged and still shorter than the default.
- Dedup TTL sizing (4x window) is 24 h = 240 ticks, under the 255-tick cap.
- Existing saved configs are not migrated, matching the 11 h -> 5 h change in #11606; the new default applies when the TM config is first installed.

Companion protobufs comment change documenting the zero-means-disabled contract and this default: meshtastic/protobufs#1066.

## 🤝 Attestations

- [ ] I have tested that my proposed changes behave as described. The native test suite does not build on this macOS host (`argp.h` is Linux-only), so `test_native` CI is the gate for `test_traffic_management` and `test_position_module`. Macro arithmetic and the assert were verified with a standalone compile.
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Not tested on hardware; the change is limited to two default constants and a compile-time check.

## Summary by CodeRabbit

* **Behavior Changes**
  * Increased the default traffic-management position deduplication interval from 5 hours to 6 hours.
  * Adjusted stationary position broadcasts to remain just beyond the deduplication window.
  * New local identities now remain prioritized as the device’s primary identity, with node information refreshed accordingly.
* **Reliability**
  * Added validation to keep the stationary broadcast interval safely above the deduplication period.
* **Tests**
  * Updated traffic-management timing scenarios to reflect the new 6-hour default.


NomDeTom:
I set 5 hours to give a decent margin for bad RTC. Is one tick window enough? There are some pretty crappy RTCs out there...

thebentern (approved):

src/mesh/Default.h
NomDeTom, line 1: I would prefer a wider margin than 360s on a 21600s limit was provided. The point at which false early deduplication kicks in is an error of 0.8% on both ends or 1.6% on one end. It's unlikely for a calibrated crystal, but several of the sleep methods we use on esp32 will slow the accumulation of millis() and lead to an early dedupe. Add on mesh rebroadcast delays and asymmetry in routing and you have a recipe for blocked broadcasts.

PR #11771, opened 2026-09-08
Allow Spacebar to advance frames

Xaositek:
Allow Spacebar to advance frames, similar to how backspace key allows to go backwards a frame.

Verified that space bar still works inside of canned messages after a message is started.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pressing the spacebar now advances to the next screen frame when keyboard input is not being intercepted.
  * Prevented the spacebar from opening the free-text composer in inactive canned-message screens.


thebentern (approved):

HarukiToreda (approved):
Yeah good idea

PR #11774, opened 2026-09-08
fix(sx126x): let CalibrateImage settle before re-applying RX registers in resetAGC()

ip2k:
## What
`resetAGC()` (#9705) recalibrates the SX126x every interval and then re-applies the RX boosted-gain and `0x8B5` RX-sensitivity registers (#10219). This adds a 50 ms settle after `calibrateImage()` before those register writes.

## Why
On a RAK WisMesh Repeater Mini (RAK3401 nRF52840 + RAK13302 1 W SX1262) the radio periodically went deaf until reboot. Traced into `resetAGC()`: `SX126x::calibrateImage()` returns as soon as the command is accepted, and the chip does **not** hold BUSY for the rest of that calibration. The next register write (`setRxBoostedGainMode`, then the `0x8B5` patch) lands in that window, fails write-verify (`RADIOLIB_ERR_SPI_WRITE_FAILED`, -16), and leaves the radio needing a full re-init.

Related to #9933 (closed): that was a *duty-cycle wake* BUSY race; this is a sibling at a different site, the `calibrateImage()` call inside `resetAGC()`.

## How it works
`Calibrate(0x7F)` earlier in the same function is already followed by a BUSY poll; the image calibration is not. Instrumentation confirmed **BUSY never asserts after `calibrateImage()`** (0 of ~2,000 cycles), so polling cannot guard this window - a fixed settle is required. Register writes issued inside ~10 ms fail; none fail at >=20 ms; 50 ms is ~2.5x that floor. nRF52 `delay()` is 10 ms-tick quantized (measured 49-51 ms). The pass already holds the radio out of RX for the calibration, so the cost is immaterial (<=60 ms per 60 s).

## What it changes
One line in `SX126xInterface::resetAGC()`. No API, config or wire-format change. SX126x only.

## Testing
RAK3401 + RAK13302, nRF52840, live 915 MHz mesh, AGC interval shortened to 10 s to accelerate. Each `resetAGC` pass logged Calibrate / CalibrateImage / register-write status, so the settle-window failure (both calibrations return 0, the following write fails -16) is counted distinctly from unrelated chip faults.

| settle after CalibrateImage | settle-window -16 failures |
|---|---|
| 0 ms | 9 / 13 (69%) |
| 10 ms | 1 / 38 |
| 20 ms | 0 / 205 |
| 50 ms (continuous RX) | 0 / 182 |
| 50 ms (duty-cycle RX) | 0 / 216 |
| **50 ms, 16.3 h soak** | **0 / 3,085 resets, 0 reboots** |

The dose-response and soak ran on a 2.8.0-based build on this hardware; this commit is the identical one-line change on `develop`, compile-verified for `rak3401-1watt`. A `develop`-based on-air soak of this exact commit is running now; I will post the result here.

## 🤝 Attestations
  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

Tested on RAK WisBlock RAK3401 + RAK13302 (1 W, nRF52840). No other SX126x boards on hand; regression testing on other SX126x hardware is welcome.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved radio calibration reliability by allowing the operation to complete before applying subsequent settings.


ip2k:
Hi, just wanted to note here that while Claude helped me dig deep into this issue and did most of the legwork, there’s a real human who was having a real issue for a long time that was hard to pin down previously after banging my head against the wall for quite a while and putting my Meshtastic gear on a shelf because it didn’t work well. I’ll be on Discord too, please let me know what I can do better or how I can otherwise help :)

src/mesh/SX126xInterface.cpp
```diff
@@ -569,6 +569,9 @@ template <typename T> void SX126xInterface<T>::resetAGC()
     // 5. Re-calibrate image rejection for actual operating frequency
     //    Calibrate(0x7F) defaults to 902-928 MHz which is wrong for other regions.
     lora.calibrateImage(getFreq());
+    // CalibrateImage keeps working internally after it returns, and BUSY does not stay asserted for it; a
```
virtualjmills: Suggestion: Shouldn't this represent "Step 6", much like "Step 4" is the busy/wait loop for "Step 3"?

Also re-wrap to match ~85 character width of prior steps.

```suggestion
   // 6. CalibrateImage keeps working internally after it returns, and BUSY does not
    //    stay asserted for it; a register write in that window fails write-verify
    //    (RADIOLIB_ERR_SPI_WRITE_FAILED) and stalls the chip.
```

src/mesh/SX126xInterface.cpp
```diff
@@ -569,6 +569,9 @@ template <typename T> void SX126xInterface<T>::resetAGC()
...
+    // register write in that window fails write-verify (RADIOLIB_ERR_SPI_WRITE_FAILED) and stalls the chip.
```
virtualjmills: See [Suggestion] for prior line.

src/mesh/SX126xInterface.cpp
ip2k: Good call, thanks — done in 1db6707: numbered the settle as step 6 (it's the wait for step 5, the way step 4 waits for step 3), re-wrapped to the width of the other steps, and bumped "Resume receiving" to step 7 so the sequence stays consecutive. Comment-only, no functional change.

ip2k:
**On-air result of this exact commit on `develop`** (RAK3401 + RAK13302, nRF52840; AGC interval shortened to 10 s to accelerate):

- 5.2 h, 899 `resetAGC` passes: **0 settle-window failures** (`0x8B5 RX patch re-apply failed` never logged), 0 reboots. The fix holds on `develop`.
- Corrected figures for the 2.8.0-based soak in the description (that table was a snapshot at 16.3 h): full window 17.4 h / 3,302 passes / 0 settle failures / 0 re-inits / 0 reboots.

One thing to disclose plainly because it is *not* this PR's bug but appears in the same log: during the `develop` soak there were 5 chip-state-loss recoveries via the #11676/#11678 path, each triggered by `SX126x standby … err=-705` or `startReceiveDutyCycleAuto … err=-705`, each re-init succeeding in ~70 ms with the queued packet then sent. All 5 (and 4 similar `-705` transients that the 2.8.0 build merely logged) fall 0.09–0.77 s after a `resetAGC()` that ran **with a TX already queued** — the guard checks in-flight TX/RX, not the queue. It is unrelated to the CalibrateImage settle (the register writes never fail), about 6× rarer at the default 60 s interval, and correlated with a warm afternoon. I'll propose a small separate change (skip `resetAGC()` while the TX queue is non-empty) once it has soaked. #11782 (`limitPower()` re-seed) kept TX power at the configured level through each re-init.

thebentern (approved):

PR #11777, opened 2026-09-08
fix(ci): unbreak the ESP32 static analysis gate after the cppcheck 2.20 jump

thebentern:
Every PR targeting `develop` has been red since 2026-09-07 on the seven ESP32 `check` jobs (and the `ci-gate` that waits on them), while the nRF52, RP2040 and STM32 jobs pass on identical source. `check` only runs on `pull_request` / `merge_group`, so `develop` itself has looked green the whole time.

## Cause

meshtastic/gh-action-firmware#61 split `requirements_*.txt` so the `esp32*` container images install the pioarduino core instead of upstream platformio. Its esp32 platform ships its own `tool-cppcheck` and reinstalls it over anything the repo pins:

| image | core | tool-cppcheck |
| --- | --- | --- |
| `develop-nrf52840` | platformio 6.1.19 | `1.21100.230717` -> cppcheck **2.11** |
| `develop-esp32s3` | pioarduino 6.1.19 | `2.20.1` -> cppcheck **2.20** |

cppcheck 2.20 parses far more of this tree than 2.11 ever managed, so checks that have always been enabled fired for the first time: **388 defects, ESP32 only, same source**. `bin/check-all.sh` runs with `--fail-on-defect=low`, so all of it is fatal.

Pinning from this repo does not work. `pio pkg install -g --tool platformio/tool-cppcheck@1.21100.230717` installs cleanly, and then `pio check` silently restores 2.20.1 - pioarduino's esp32 `tools.json` marks it `install: "always"`. Upstream's registry also has nothing newer than 2.11, so the two sides cannot currently meet in the middle.

## Changes

Rebased on #11776, which cleared the two `unknownMacro` errors. Of the 386 left, two are worth acting on and are fixed rather than suppressed:

- **`SerialModule` dereferenced a null `Position`.** In NMEA/CALTOPO mode `decoded` stays `NULL` when `pb_decode_from_bytes()` fails, but `printWPL()` was called with `*decoded` regardless - a malformed position payload on our portnum crashes the node. The waypoint is now emitted only on a successful decode.
- **InkHUD's 12-hour clock** passed a signed `12` to a `%u` conversion.

The remaining 384 are style and performance suggestions - `functionStatic` and the const-correctness family account for 351 of them. Those check ids are suppressed so the gate means the same thing on every platform again, with the one-off sites scoped to their file so a new occurrence elsewhere still fails. Burning the rest down is worth doing deliberately, not under a CI outage.

## Verification

Run in the CI container images themselves, via `bin/check-all.sh`:

- **pass** (all seven that were failing): `heltec-v3`, `t-deck-tft`, `heltec-vision-master-e213-inkhud`, `seeed-sensecap-indicator-tft`, `thinknode_m7`, `heltec-ht62-esp32c3-sx1262`, `tlora-c6`
- **no regression under cppcheck 2.11**: `rak4631`, `tracker-t1000-e`, `t-echo-plus`

The version split itself is worth a look on the container side: meshtastic/gh-action-firmware#64.

### DIFF
--- a/suppressions.txt
+++ b/suppressions.txt
@@ -52,13 +52,8 @@ postfixOperator:*/mqtt/*
 missingOverride
 virtualCallInConstructor
 
-passedByValue:*/RedirectablePrint.h
-
 internalAstError:*/CrossPlatformCryptoEngine.cpp
 uninitMemberVar:*/AudioThread.h
-// False positive
-constVariableReference:*/Channels.cpp
-constParameterPointer:*/unishox2.c
 
 // False positive: make_zeroizing_array() returns unique_ptr<uint8_t[], ...>, so
 // .get() is uint8_t*, not void*. cppcheck can't resolve the custom-deleter alias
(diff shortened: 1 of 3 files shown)


vidplace7:
@thebentern sure you aren't over-suppressing here? 😬 

> burning them down is worth doing deliberately, not under a CI outage.

Fair enough 😅

suppressions.txt
```diff
@@ -67,4 +62,41 @@ arithOperationsOnVoidPointer:*/EncryptedStorage.cpp
...
+// resolves platformio/tool-cppcheck 1.21100.230717, i.e. cppcheck 2.11. 2.20 parses far more of
+// this tree than 2.11 managed to, so checks that have always been enabled started firing for the
+// first time - 388 defects on ESP32, none anywhere else, for identical source.
+//
+// Silence the checks that appeared with that jump so the gate means the same thing on every
+// platform again. These are style and performance suggestions, not defects; burning them down is
+// worth doing deliberately, not under a CI outage.
+functionStatic
+staticFunction
+constParameterPointer
```
vidplace7, line 80: `constParameterPointer` isn't new (there are existing suppressions for it in the code)

vidplace7 (approved):
Let's merge it (to unblock CI) and then try to clean up after ourselves (un-suppress and address)

PR #11789, opened 2026-09-09
fix(ham): synchronize license transitions

RCGV1:
## Summary

- announce a Ham/license transition once, after the radio and channel configuration commits
- bypass only the normal NodeInfo cadence for that transition announcement
- send the announcement with `want_response=false`, so peers do not fan out replies
- retain a denied announcement locally until airtime admission allows the single send
- preserve the existing reboot behavior and `CLIENT_HIDDEN` transmission policy
- cover both admin/API transitions and Device UI exit from licensed mode

## Root cause

Peers cache `User.is_licensed` in NodeDB and apply licensed-mode routing rules from that cached state. The previous admin path called `reloadOwner` before `saveChanges` applied the new radio/channel configuration. During a channel transition, that NodeInfo could therefore leave on the old channel and be missed by peers on the new one. The normal NodeInfo cadence could then suppress a timely replacement, leaving peers to reject otherwise valid signed traffic with `Packet to or from unlicensed user, ignoring packet`.

The fix defers the transition NodeInfo until configuration commit and lets this one packet bypass cadence suppression.

## Traffic and compatibility

- multiple transition requests coalesce into one pending announcement
- at most one transition broadcast is handed to the router
- the packet has `want_response=false`, so it solicits zero peer replies
- if airtime admission denies the send, a local check runs again after 30 seconds; denied checks allocate and queue no radio packet
- the pending flag clears as soon as the one packet is admitted, so there is no transmitted retry burst
- packet allocation, signing, and router behavior remain unchanged
- existing reboot semantics remain unchanged
- existing `CLIENT_HIDDEN` behavior remains unchanged

## Validation

- `native-macos:test_admin_radio`: 110/110 passed
- `native-macos:test_packet_signing`: 81/81 passed
- `pio run -e tbeam-s3-core`: passed (RAM 30.1%, flash 71.4%)
- focused tests verify one announcement with `want_response=false`, rejection-without-queue followed by one admitted send, and `CLIENT_HIDDEN` silence
- earlier Muzi Base + T-Beam S3 Core bench reproduction on `develop` confirmed stale licensed-state rejection and showed that a post-config NodeInfo converges peers
- the final sender-only reduction was validated with native tests and the ESP32 build; it has not been reflashed to hardware after removing the earlier retry/reply design

## Summary by CodeRabbit

* **New Features**
  * Owner and license-status changes now publish refreshed node information after changes are committed.
  * Mode transitions can broadcast updated owner details without waiting for the normal messaging cadence.

* **Bug Fixes**
  * Improved owner-information synchronization when switching between licensed and standard modes.
  * Prevented unnecessary transition replies and repeated synchronization attempts.

* **Tests**
  * Added coverage for deferred owner synchronization, retry handling, and client-hidden behavior.

### DIFF
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -837,6 +837,7 @@ void AdminModule::handleSetOwner(const meshtastic_User &o)
 
     if (changed) { // If nothing really changed, don't broadcast on the network or write to flash
         ownerSyncPending = ownerSyncPending || licenseChanged;
+        // A license change can also change the radio config, so announce only after that config commits.
         service->reloadOwner(!hasOpenEditTransaction && !licenseChanged);
         saveChanges(SEGMENT_DEVICESTATE | SEGMENT_NODEDATABASE | (identityUpdated ? SEGMENT_CONFIG : 0) |
                     (channelsSanitized ? SEGMENT_CHANNELS : 0));
@@ -1914,8 +1915,6 @@ void AdminModule::saveChanges(int saveWhat, bool shouldReboot)
     if (!hasOpenEditTransaction) {
         LOG_INFO("Save changes to disk");
         service->reloadConfig(saveWhat); // Calls saveToDisk among other things
-        if (ownerSyncPending)
-            shouldReboot = false;
         flushPendingOwnerSync();
     } else {
         LOG_INFO("Delay disk save until open transaction commits");
@@ -1930,7 +1929,7 @@ void AdminModule::saveChanges(int saveWhat, bool shouldReboot)
 void AdminModule::flushPendingOwnerSync()
 {
     if (ownerSyncPending && nodeInfoModule) {
-        nodeInfoModule->requestOwnerSync();
+        nodeInfoModule->sendOurNodeInfo(NODENUM_BROADCAST, false, 0, true, true);
         ownerSyncPending = false;
     }
 }
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -5,6 +5,7 @@
 #include "HardwareRNG.h"
 #include "MeshService.h"
 #include "NodeDB.h"
+#include "NodeInfoModule.h"
 #include "PositionPrecision.h"
 #include "PowerFSM.h"
 #include "SPILock.h"
@@ -786,6 +787,7 @@ void AdminModule::handleSetOwner(const meshtastic_User &o)
     int changed = 0;
     bool identityUpdated = false;
     bool channelsSanitized = false;
+    const bool licenseChanged = owner.is_licensed != o.is_licensed;
 
     if (*o.long_name) {
         // Apps built against the older 39-byte limit may send longer names; clamp
@@ -804,7 +806,7 @@ void AdminModule::handleSetOwner(const meshtastic_User &o)
         owner.short_name[sizeof(owner.short_name) - 1] = '\0';
         sanitizeUtf8(owner.short_name, sizeof(owner.short_name));
     }
-    if (owner.is_licensed != o.is_licensed) {
+    if (licenseChanged) {
         changed = 1;
 #if !(MESHTASTIC_EXCLUDE_PKI_KEYGEN || MESHTASTIC_EXCLUDE_PKI)
         const bool identityWillMigrate =
@@ -834,7 +836,8 @@ void AdminModule::handleSetOwner(const meshtastic_User &o)
     }
 
     if (changed) { // If nothing really changed, don't broadcast on the network or write to flash
-        service->reloadOwner(!hasOpenEditTransaction);
+        ownerSyncPending = ownerSyncPending || licenseChanged;
+        service->reloadOwner(!hasOpenEditTransaction && !licenseChanged);
         saveChanges(SEGMENT_DEVICESTATE | SEGMENT_NODEDATABASE | (identityUpdated ? SEGMENT_CONFIG : 0) |
                     (channelsSanitized ? SEGMENT_CHANNELS : 0));
     }
@@ -1911,6 +1914,9 @@ void AdminModule::saveChanges(int saveWhat, bool shouldReboot)
     if (!hasOpenEditTransaction) {
         LOG_INFO("Save changes to disk");
         service->reloadConfig(saveWhat); // Calls saveToDisk among other things
+        if (ownerSyncPending)
+            shouldReboot = false;
+        flushPendingOwnerSync();
     } else {
         LOG_INFO("Delay disk save until open transaction commits");
         editTransactionActivityMs = millis(); // still in use, so not the abandoned kind we time out
@@ -1918,6 +1924,14 @@ void AdminModule::saveChanges(int saveWhat, bool shouldReboot)
     }
     if (shouldReboot && !hasOpenEditTransaction) {
         reboot(DEFAULT_REBOOT_SECONDS);
+    }
+}
+
+void AdminModule::flushPendingOwnerSync()
+{
+    if (ownerSyncPending && nodeInfoModule) {
+        nodeInfoModule->requestOwnerSync();
+        ownerSyncPending = false;
     }
 }
 
@@ -1994,6 +2008,7 @@ bool AdminModule::handleSetHamMode(const meshtastic_HamParameters &p)
     }
 
     service->reloadOwner(false);
+    ownerSyncPending = true;
     saveChanges(SEGMENT_CONFIG | SEGMENT_NODEDATABASE | SEGMENT_DEVICESTATE | SEGMENT_CHANNELS);
     return true;
 }
--- a/src/modules/AdminModule.cpp
+++ b/src/modules/AdminModule.cpp
@@ -1929,7 +1929,7 @@ void AdminModule::saveChanges(int saveWhat, bool shouldReboot)
 void AdminModule::flushPendingOwnerSync()
 {
     if (ownerSyncPending && nodeInfoModule) {
-        nodeInfoModule->sendOurNodeInfo(NODENUM_BROADCAST, false, 0, true, true);
+        nodeInfoModule->requestOwnerSync();
         ownerSyncPending = false;
     }
 }
--- a/src/modules/NodeInfoModule.cpp
+++ b/src/modules/NodeInfoModule.cpp
@@ -19,6 +19,10 @@
 NodeInfoModule *nodeInfoModule;
 
 static constexpr uint32_t NodeInfoReplySuppressSeconds = USERPREFS_NODEINFO_REPLY_SUPPRESS_SECS;
+static constexpr uint32_t TransitionReplyAllowanceSeconds = 5 * 60;
+static constexpr uint32_t OwnerSyncRetryMs = 5 * 1000;
+static constexpr uint8_t TransitionReplyAttempts = 2;
+static constexpr uint8_t OwnerSyncAttempts = 3;
 
 bool NodeInfoModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, meshtastic_User *pptr)
 {
@@ -50,6 +54,7 @@ bool NodeInfoModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mes
         return true;
     }
     NodeNum sourceNum = getFrom(&mp);
+    const UserLicenseStatus previousLicense = nodeDB->getLicenseStatus(sourceNum);
     // Broadcasts only: unicast NodeInfo is unsigned off ham, so updateUser refuses the identity
     // write instead. isKnownXeddsaSigner also covers the warm tier.
     if (nodeDB->isKnownXeddsaSigner(sourceNum) && !mp.xeddsa_signed && isBroadcast(mp.to)) {
@@ -63,6 +68,11 @@ bool NodeInfoModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mes
     // updateUser() refuses the identity write for a known signer sending unsigned (all unicast
     // NodeInfo), so the exchange above still proceeds but cannot spoof the stored name.
     bool hasChanged = nodeDB->updateUser(getFrom(&mp), p, mp.channel, mp.xeddsa_signed);
+    const bool canReply = mp.decoded.want_response && !isFromUs(&mp) && (isBroadcast(mp.to) || isToUs(&mp));
+    if (canReply && hasChanged && previousLicense != UserLicenseStatus::NotKnown &&
+        (previousLicense == UserLicenseStatus::Licensed) != p.is_licensed) {
+        transitionReplyAllowances[sourceNum] = {Time::getUptimeSecs(), TransitionReplyAttempts};
+    }
 
     bool wasBroadcast = isBroadcast(mp.to);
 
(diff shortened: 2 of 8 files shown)


src/modules/NodeInfoModule.h
```diff
@@ -48,13 +51,23 @@ class NodeInfoModule : public ProtobufModule<meshtastic_User>, private concurren
     virtual int32_t runOnce() override;
 
   private:
+    struct TransitionReplyAllowance {
```
NomDeTom: The whole `transitionReplyAllowances` mechanism is ~41 lines, a third of the non-test diff, and it solves a different problem from the sender-side announce (refreshing *our* view of the peer, not theirs) - it would be far easier to reason about as its own PR.

src/modules/NodeInfoModule.cpp
```diff
@@ -63,6 +68,11 @@ bool NodeInfoModule::handleReceivedProtobuf(const meshtastic_MeshPacket &mp, mes
     // updateUser() refuses the identity write for a known signer sending unsigned (all unicast
     // NodeInfo), so the exchange above still proceeds but cannot spoof the stored name.
     bool hasChanged = nodeDB->updateUser(getFrom(&mp), p, mp.channel, mp.xeddsa_signed);
+    const bool canReply = mp.decoded.want_response && !isFromUs(&mp) && (isBroadcast(mp.to) || isToUs(&mp));
+    if (canReply && hasChanged && previousLicense != UserLicenseStatus::NotKnown &&
+        (previousLicense == UserLicenseStatus::Licensed) != p.is_licensed) {
+        transitionReplyAllowances[sourceNum] = {Time::getUptimeSecs(), TransitionReplyAttempts};
```
NomDeTom: This arms a reply off a *broadcast*, so one license flip can arm every node in range to answer - that is the amplification concern in a nutshell, and it needs justifying against a dense mesh rather than a two-node bench.

src/modules/NodeInfoModule.cpp
```diff
@@ -19,6 +19,10 @@
 NodeInfoModule *nodeInfoModule;
 
 static constexpr uint32_t NodeInfoReplySuppressSeconds = USERPREFS_NODEINFO_REPLY_SUPPRESS_SECS;
+static constexpr uint32_t TransitionReplyAllowanceSeconds = 5 * 60;
+static constexpr uint32_t OwnerSyncRetryMs = 5 * 1000;
+static constexpr uint8_t TransitionReplyAttempts = 2;
```
NomDeTom: Three announce broadcasts at 5s intervals plus up to two replies per in-range node is a burst of 3 + 2N packets from a single settings change; what is the worst case on a busy channel?

src/modules/NodeInfoModule.cpp
```diff
@@ -204,6 +259,14 @@ void NodeInfoModule::pruneLastNodeInfoCache()
         }
     }
 
+    for (auto it = transitionReplyAllowances.begin(); it != transitionReplyAllowances.end();) {
```
NomDeTom: This prune loop exists only to service the allowance map, so it goes away with the receiver-side half.

src/modules/NodeInfoModule.cpp
```diff
@@ -144,7 +170,35 @@ meshtastic_MeshPacket *NodeInfoModule::allocReply()
...
+
+    meshtastic_MeshPacket *reply = allocReplyWithOptions(bypassCadenceThrottle);
+    if (reply && allowance != transitionReplyAllowances.end() && --allowance->second.remaining == 0)
+        transitionReplyAllowances.erase(allowance);
+    return reply;
+}
+
+meshtastic_MeshPacket *NodeInfoModule::allocReplyWithOptions(bool bypassCadenceThrottle)
+{
+    const bool isReplyingToExternalRequest = currentRequest &&
```
NomDeTom, line 158: These seven lines are copied verbatim from `allocReply()` directly above; the duplication disappears if the bypass flag is simply a parameter on one function.

src/modules/NodeInfoModule.cpp
```diff
@@ -144,7 +170,35 @@ meshtastic_MeshPacket *NodeInfoModule::allocReply()
...
+        }
+    }
+
+    meshtastic_MeshPacket *reply = allocReplyWithOptions(bypassCadenceThrottle);
+    if (reply && allowance != transitionReplyAllowances.end() && --allowance->second.remaining == 0)
+        transitionReplyAllowances.erase(allowance);
+    return reply;
+}
+
+meshtastic_MeshPacket *NodeInfoModule::allocReplyWithOptions(bool bypassCadenceThrottle)
```
NomDeTom: The `allocReply` / `allocReplyWithOptions` split can collapse into a single function with a default argument, with no behaviour change.

src/modules/NodeInfoModule.cpp
```diff
@@ -96,20 +106,27 @@ void NodeInfoModule::alterReceivedProtobuf(meshtastic_MeshPacket &mp, meshtastic
 }
 
 bool NodeInfoModule::sendOurNodeInfo(NodeNum dest, bool wantReplies, uint8_t channel, bool _shorterTimeout)
+{
+    return sendOurNodeInfoWithOptions(dest, wantReplies, channel, _shorterTimeout, false);
+}
+
+bool NodeInfoModule::sendOurNodeInfoWithOptions(NodeNum dest, bool wantReplies, uint8_t channel, bool _shorterTimeout,
```
NomDeTom: Same here: a default argument on `sendOurNodeInfo()` removes the need for a five-arg twin plus a forwarder that only delegates.

src/modules/NodeInfoModule.cpp
```diff
@@ -96,20 +106,27 @@ void NodeInfoModule::alterReceivedProtobuf(meshtastic_MeshPacket &mp, meshtastic
...
 {
     // cancel any not yet sent (now stale) position packets
     if (prevPacketId) // if we wrap around to zero, we'll simply fail to cancel in that rare case (no big deal)
         service->cancelSending(prevPacketId);
     shorterTimeout = _shorterTimeout;
     DEBUG_HEAP_BEFORE;
-    meshtastic_MeshPacket *p = allocReply();
+    meshtastic_MeshPacket *p = allocReplyWithOptions(ownerSync);
     DEBUG_HEAP_AFTER("NodeInfoModule::sendOurNodeInfo", p);
+    shorterTimeout = false;
```
NomDeTom: Genuine fix - `shorterTimeout` previously leaked on the failure path - but it is unrelated to license sync and would land faster on its own.

src/modules/NodeInfoModule.cpp
```diff
@@ -121,15 +138,24 @@ bool NodeInfoModule::sendOurNodeInfo(NodeNum dest, bool wantReplies, uint8_t cha
             p->channel = channel;
         }
 
-        prevPacketId = p->id;
-
-        service->sendToMesh(p);
-        shorterTimeout = false;
-        return true;
+        const PacketId packetId = p->id;
+        const ErrorCode result = service->sendToMesh(p);
```
NomDeTom: Another real and unrelated bugfix bundled in here: `prevPacketId` was previously set even when the send failed.

src/modules/NodeInfoModule.cpp
```diff
@@ -96,20 +106,27 @@ void NodeInfoModule::alterReceivedProtobuf(meshtastic_MeshPacket &mp, meshtastic
...
+    meshtastic_MeshPacket *p = allocReplyWithOptions(ownerSync);
     DEBUG_HEAP_AFTER("NodeInfoModule::sendOurNodeInfo", p);
+    shorterTimeout = false;
 
     if (p) { // Check whether we didn't ignore it
         p->to = dest;
-        bool requestWantResponse = (config.device.role != meshtastic_Config_DeviceConfig_Role_TRACKER &&
-                                    config.device.role != meshtastic_Config_DeviceConfig_Role_SENSOR) &&
-                                   wantReplies;
+        bool requestWantResponse =
```
NomDeTom: The `requestWantResponse` restructure is a third independent change riding along in this PR.

src/modules/NodeInfoModule.cpp
```diff
@@ -227,12 +290,19 @@ NodeInfoModule::NodeInfoModule()
 
 int32_t NodeInfoModule::runOnce()
 {
-    if (airTime->isTxAllowedAirUtil() && config.device.role != meshtastic_Config_DeviceConfig_Role_CLIENT_HIDDEN) {
+    if (airTime->isTxAllowedAirUtil() &&
+        (ownerSyncPending || config.device.role != meshtastic_Config_DeviceConfig_Role_CLIENT_HIDDEN)) {
```
NomDeTom: This makes a `CLIENT_HIDDEN` node transmit; a deliberately silent role breaking its silence is a policy decision rather than plumbing, and should be called out explicitly.

src/modules/AdminModule.cpp
```diff
@@ -1911,6 +1914,9 @@ void AdminModule::saveChanges(int saveWhat, bool shouldReboot)
     if (!hasOpenEditTransaction) {
         LOG_INFO("Save changes to disk");
         service->reloadConfig(saveWhat); // Calls saveToDisk among other things
+        if (ownerSyncPending)
+            shouldReboot = false;
```
NomDeTom: Suppressing a reboot the existing code asked for needs its own justification: some of those reboots exist because the config change is not fully applied without one.

src/modules/AdminModule.cpp
```diff
@@ -834,7 +836,8 @@ void AdminModule::handleSetOwner(const meshtastic_User &o)
     }
 
     if (changed) { // If nothing really changed, don't broadcast on the network or write to flash
-        service->reloadOwner(!hasOpenEditTransaction);
+        ownerSyncPending = ownerSyncPending || licenseChanged;
```
NomDeTom, line 839: Worth a comment explaining why a license change in particular must defer `reloadOwner` until the transaction commits.

NomDeTom (commented):
This seems like a strange edge-case. Who is swapping between licensed and unlicensed mode with such frequency that this becomes necessary?

Also, the burst of packets and responses on a settings change sounds like an amplification attack, not a feature.

Finally, the code itself is adding things that could be separate bugfixes.

src/modules/NodeInfoModule.h
RCGV1: Removed in dc384a07d. The PR is sender-only now; transition reply allowances and all receiver-side transition state are gone.

src/modules/NodeInfoModule.cpp
RCGV1: Removed in dc384a07d. The transition broadcast now has want_response=false and cannot arm or solicit peer replies.

src/modules/NodeInfoModule.cpp
RCGV1: Reduced in dc384a07d to at most one sender broadcast and zero solicited replies: 1 + 0N packets. There is no retry burst.

src/modules/NodeInfoModule.cpp
RCGV1: Removed in dc384a07d with the allowance map.

src/modules/NodeInfoModule.cpp
RCGV1: Collapsed in dc384a07d. allocReply() now delegates to one allocNodeInfo(bool) implementation, so the allocation logic is no longer duplicated.

src/modules/NodeInfoModule.cpp
RCGV1: Collapsed in dc384a07d into one allocNodeInfo(bool) implementation. The virtual allocReply() remains only as the fixed-signature override and delegates to it.

src/modules/NodeInfoModule.cpp
RCGV1: Collapsed in dc384a07d into the existing sendOurNodeInfo() with one defaulted bypassCadenceThrottle argument; the forwarding twin is gone.

src/modules/NodeInfoModule.cpp
RCGV1: Reverted from this PR in dc384a07d; the shorterTimeout failure-path change is no longer in the diff.

src/modules/NodeInfoModule.cpp
RCGV1: Reverted from this PR in dc384a07d; prevPacketId behavior is unchanged from develop.

src/modules/NodeInfoModule.cpp
RCGV1: Reverted from this PR in dc384a07d; the existing want_response role logic is unchanged.

src/modules/NodeInfoModule.cpp
RCGV1: Removed in dc384a07d. runOnce() is unchanged from develop, so the existing CLIENT_HIDDEN transmission policy is preserved.

src/modules/AdminModule.cpp
RCGV1: Removed in dc384a07d. saveChanges() retains the existing reboot behavior; owner sync no longer suppresses it.

src/modules/AdminModule.cpp
RCGV1: Added the ordering comment in dc384a07d: a license change can also change radio config, so the announcement waits until that config commits.

src/modules/AdminModule.cpp
```diff
@@ -1921,6 +1926,14 @@ void AdminModule::saveChanges(int saveWhat, bool shouldReboot)
     }
 }
 
+void AdminModule::flushPendingOwnerSync()
+{
+    if (ownerSyncPending && nodeInfoModule) {
+        nodeInfoModule->sendOurNodeInfo(NODENUM_BROADCAST, false, 0, true, true);
```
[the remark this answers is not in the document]
RCGV1: Fixed in a10b18181. Admin and Device UI now hand the transition to the same NodeInfo pending path. Airtime rejection retains the pending flag and retries only the local admission check after 30 seconds; rejected checks queue no packet, and the flag clears after the first admitted packet. N13 covers reject-then-admit with exactly one queued broadcast, and N14 verifies CLIENT_HIDDEN remains silent.

RCGV1:
@coderabbitai review

NomDeTom (commented):
Rework checked against the diff rather than the replies - all thirteen points hold: the receiver-side allowance is gone, the transition broadcast can no longer solicit replies, the allocation paths are collapsed into one, and the three drive-by fixes and both policy changes are withdrawn. Non-test additions are down from 120 to 55 lines.

Two small things the rework itself introduced, both in `runOnce()`.

src/modules/NodeInfoModule.cpp
```diff
@@ -229,10 +246,15 @@ int32_t NodeInfoModule::runOnce()
 {
     if (airTime->isTxAllowedAirUtil() && config.device.role != meshtastic_Config_DeviceConfig_Role_CLIENT_HIDDEN) {
         // If we changed channels, ask everyone else for their latest info
-        bool requestReplies = currentGeneration != radioGeneration;
+        const bool isOwnerSync = ownerSyncPending;
+        bool requestReplies = !isOwnerSync && currentGeneration != radioGeneration;
```
NomDeTom, line 250: An owner sync suppresses `requestReplies`, but line 253 still consumes the pending generation, so a `radioGeneration` bump that has not yet been announced is swallowed without ever asking peers for their info. Narrow window - `radioGeneration++` only happens on a fresh install (`NodeDB.cpp:1105`) - but "flash a node, then set ham mode" reaches it. Suggest only assigning `currentGeneration` when `requestReplies` was actually sent.

src/modules/NodeInfoModule.cpp
```diff
@@ -229,10 +246,15 @@ int32_t NodeInfoModule::runOnce()
...
         LOG_INFO("Send our nodeinfo to mesh (wantReplies=%d)", requestReplies);
-        if (sendOurNodeInfo(NODENUM_BROADCAST, requestReplies))
+        if (sendOurNodeInfo(NODENUM_BROADCAST, requestReplies, 0, isOwnerSync, isOwnerSync)) {
```
NomDeTom, line 252: Passing `isOwnerSync` as `_shorterTimeout` is redundant - `bypassCadenceThrottle` (the next argument) already skips both throttle branches - and it has a side effect: it sets the `shorterTimeout` member, which is only cleared inside `if (p)`, so an allocation failure leaves it `true` and the next ordinary periodic broadcast takes the 60s window instead of the 10-minute cadence. Passing `false` here drops both the redundancy and the leak.

Issue #11797, opened 2026-09-09
[Bug]: Telemetry can exceed the 233-byte payload; oversized encodes are silently sent as empty packets


thebentern:
**Category:** Other
**Hardware:** Not Applicable, the defect is board independent. Field evidence came from a Seeed SenseCAP Mesh Tracker X1.
**Firmware Version:** Original regression present since 2.6 era, still present on develop at cccefa09a

### Summary

`meshtastic_Telemetry` can encode larger than a mesh packet payload. When it does, `pb_encode_to_bytes` returns 0, the caller stores that as the payload size, and a well formed packet carrying **nothing** is transmitted or delivered instead of the encode failing visibly.

### Original regression

The ceiling was broken by the automated protobuf pull #6836 (`5b312ab91`, 2025-05-17). To be explicit about attribution, this is the original protobuf pull, **not** the later #11531 "Update protobufs and classes".

| commit | PR | `Telemetry_size` | vs. 233 byte payload |
| --- | --- | --- | --- |
| `94de2315c` | #6266 | 120 | fits |
| `5b312ab91` | **#6836** | **272** | **overflows** |
| `fdb67309a` | #11741 | 320 | overflows |

`meshtastic_Constants_DATA_PAYLOAD_LEN` is 233. #6836 introduced `meshtastic_HostMetrics_size` at 264, which alone pushed the `Telemetry` union past the payload limit. From that commit onward, a `Telemetry` has not been guaranteed to fit in a packet, and no size guard was added anywhere.

### Live consequence on the LoRa wire

`HostMetrics` is sent to the mesh, not just to a client. `src/modules/Telemetry/HostMetrics.cpp:130`:

```cpp
meshtastic_MeshPacket *p = allocDataProtobuf(telemetry);
...
service->sendToMesh(p, RX_SRC_LOCAL, true);
```

The variant carries `char user_string[200]`, populated at `src/modules/Telemetry/HostMetrics.cpp:112` from the output of the operator's configured `HostMetrics.UserStringCommand`. A sufficiently long command result overflows the payload, and `ProtobufModule::allocDataProtobuf` stores the failure verbatim at `src/mesh/ProtobufModule.h:50`:

```cpp
p->decoded.payload.size =
    pb_encode_to_bytes(p->decoded.payload.bytes, sizeof(p->decoded.payload.bytes), fields, &payload);
```

[...]

PR #11802, opened 2026-09-09
Regenerate telemetry protobufs for the SoilWaterMetrics split

thebentern:
Regenerated telemetry protobufs for meshtastic/protobufs#1071, which moves the
15 soil and water chemistry fields out of `EnvironmentMetrics` into their own
`SoilWaterMetrics` message on `Telemetry` oneof tag 11.

meshtastic/protobufs#1071 is **merged** as `fa26b5b`, and the submodule now
points at that `master` commit. Ready for review.

### What changes

Generated output only, plus the submodule pointer. Nothing in `src/` or `test/`
referenced any of the moved fields, so there are no call sites to update.

| constant | before | after |
| --- | --- | --- |
| `meshtastic_EnvironmentMetrics_size` | 312 | 222 |
| `meshtastic_SoilWaterMetrics_size` | — | 75 |
| `meshtastic_Telemetry_size` | 320 | 272 |

Struct sizes: `EnvironmentMetrics` 424 to 304, `Telemetry` 432 to 312.

### Why it matters here

`EnvironmentMetrics` is stored per node in the NodeDB satellite map
(`src/mesh/NodeDB.h`), so the 120-byte struct reduction is 4,800 bytes at
`MAX_SATELLITE_NODES` 40 on nRF52840, and 30,000 bytes at 250 elsewhere.
`meshtastic_NodeEnvironmentEntry_size` drops 321 to 231, which also lowers the
persisted-database decode ceiling.

### What this does not fix

`Telemetry_size` lands at 272, still above the 233-byte `DATA_PAYLOAD_LEN`,
because `HostMetrics` at 264 becomes the binding variant. That is where it sat
from meshtastic/protobufs#6836 until the recent protobuf pull. The encode
plumbing that turns an oversize message into a silently empty packet is #11797
and is unaffected by this PR.

### Verification

- nanopb 0.4.9.1 reproduced the pre-change constants exactly before generating,
  so the sizes above are generator output, not arithmetic.
- Regenerated unit compiles clean.
- `grep` over `src/` and `test/` for every moved field name returns nothing.

## Summary by CodeRabbit

- **Chores**
  - Updated the underlying protocol definitions.
  - This maintenance update does not introduce changes to exported functionality or user-facing behavior.
  - No new features, fixes, or workflow changes are included in this release.

### DIFF
--- a/src/mesh/generated/meshtastic/telemetry.pb.cpp
+++ b/src/mesh/generated/meshtastic/telemetry.pb.cpp
@@ -12,6 +12,9 @@ PB_BIND(meshtastic_DeviceMetrics, meshtastic_DeviceMetrics, AUTO)
 PB_BIND(meshtastic_EnvironmentMetrics, meshtastic_EnvironmentMetrics, 2)
 
 
+PB_BIND(meshtastic_SoilWaterMetrics, meshtastic_SoilWaterMetrics, AUTO)
+
+
 PB_BIND(meshtastic_PowerMetrics, meshtastic_PowerMetrics, AUTO)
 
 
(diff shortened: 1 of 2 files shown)


oscgonfer (approved):

caveman99:
i'll pull them with the auto updater

PR #11811, opened 2026-09-10
fix(nodedb): track whether each node was heard on the current LoRa config

### LINKED ISSUE #11745: Track whether each node has been heard on the current LoRa config
Reference implementation for meshtastic/design#146. Needs meshtastic/protobufs#1060 + a submodule bump.

After a preset, region or slot change the node DB keeps every node it knew before and nothing marks them, so clients present unreachable nodes as reachable. Firmware owns this because MUI has no phone, and because a client-side heuristic misses config changes made from the device menu or CLI.

**The bit** - `NODEINFO_BITFIELD_HEARD_ON_CURRENT_LORA_SHIFT 11`. `bitfield` is a uint32 with bits 0–10 used, so this costs no NodeDB bytes. Mirror it in `TypeConversions::ConvertToNodeInfo`.

**Set** - in `NodeDB::updateFrom`, RF hears only (not `via_mqtt`), in **both** branches of the `has_rx_time` test - the `recordHeardWhileClockUntrusted` branch is a real hear. Also `stampContactHeardNow`. Leave it clear on `getOrCreateMeshNode` admissions that aren't hears (favourite, ignore, `add_contact`).

**Clear** - hook `MeshService::reloadConfig`, **not** `AdminModule::handleSetConfig`. `MenuHandler.cpp` calls `reloadConfig` from ~15 sites: the device's own screen menu never touches AdminModule, and that's the path this exists for. It also picks up scanned channel/config URLs, licensed-mode region changes, `resetRadioConfig` and factory reset. Compare a snapshot of the slot tuple - region, `use_preset` (or bw/sf/cr), `override_frequency`, `channel_num`, and the **primary channel name**, since the slot is that name's hash.

**Persist** - `reloadConfig` ends in `saveToDisk(saveWhat)` and a LoRa write passes `SEGMENT_CONFIG`, not `SEGMENT_NODEDATABASE`. Without OR-ing that in, the clear is lost across the reboot the change triggers.

**Warm tier** - carry nothing. Re-admission legitimately leaves the bit clear.

**Upgrade** - a spare bitfield bit needs no schema change, so no migration hook fires and every node reads unheard on first boot after the update. Don't bump `DEVICESTATE_CUR_VER` (it also drives the legacy-decode gate; 26 is reserved) - use the one-time watermark pattern `POSITION_TELEMETRY_OPTIN_VER` uses.

Bench checks: preset change from the **device menu** with no client attached; a scanned channel QR; a primary-channel rename; and that the cleared bits survive the reboot.

caveman99:
Sets `NODEINFO_BITFIELD_HEARD_ON_CURRENT_LORA` (bit 11) on a genuine RF hear and clears it for every node when the LoRa slot config moves, so clients can mark the nodes that went unreachable after a preset, region, slot or primary-channel-name change. Hooked into `MeshService::reloadConfig()` — the single funnel for the device menu, admin/CLI and scanned-URL paths — plus `NodeDB::restorePreferences()`, which reboots without passing through it. Verified with the new `test_nodedb_lora_slot` suite (20/20 on `native-windows`) and a clean `tbeam` build.

Fixes #11745

## Summary by CodeRabbit

- **Bug Fixes**
  - Node activity indicators now accurately reflect the LoRa slot on which a node was heard, including after configuration changes, restores, or storage unlocks.
  - Node status now distinguishes direct radio reception from information relayed through MQTT.
  - Temporary LoRa changes used for beacon transmissions no longer incorrectly alter the committed radio slot.
  - Invalid or orphaned satellite entries are now excluded from stored node data.
  - Legacy node data no longer misinterprets reserved bitfield values.

- **Reliability**
  - LoRa slot tracking now preserves node information while updating radio-hearing status consistently across configuration changes.


thebentern (approved):

jamesarich (approved):

jamesarich:
One question - does this also accommodate `discovery` when we're rolling through all the presets listening to traffic? We shouldn't be resetting the flag if we're not actually committing to being on the new config.
@garthvh called this out to me yesterday.

caveman99:
@jamesarich no it isn't - back to the drawing board :-)

caveman99:
Discovery is client-driven, so every preset hop reaches reloadConfig and the sweep cleared the marks on the way out and again on the way home, and anything heard while parked on another preset kept its mark once the radio came back. Each node now stores a 12-bit fingerprint of the slot it was heard on in spare bitfield bits, and heard_on_current_lora is derived by comparing that against the slot the radio is committed to, so hopping through presets writes nothing and returning to a slot restores its marks by itself. The suite covers both directions of that round trip along with the beacon TX switch, which is pinned so it does not read as the radio moving.

jamesarich (approved):
Thanks for the rework @caveman99 - that should do the trick 💚

PR #11817, opened 2026-09-11
feat(raksensorhub): SensorHub uplink for WisMesh Hub (RAK2560)

Ethan-chen1234-zy:
## Summary
- Add RAK SensorHub **uplink-only** driver split across `RAKSensorHub` (1-Wire bus/poll/hot-plug) and `RAKSensorHubUplink` (IPSO → EnvCache → Telemetry).
- New PlatformIO env `rak_wismesh_sensorhub` (`HAS_RAKHUB`). Stock `env:rak2560` remains `HAS_RAKPROT` / RAK9154 only.
- Wire into `EnvironmentTelemetry` only. No `AirQualityTelemetry` changes in this PR.
- Standard on-air fields: T / RH / P / lux / wind / soil_moisture / bus voltage+current. Soil/water chemistry + solar irradiance parse into cache/logs but stay **off-air** (`RAK_SENSORHUB_EXTENDED_ENV_METRICS` commented out).

## Protobuf follow-up
- protobufs#1071 moved chemistry tags out of `EnvironmentMetrics` into `SoilWaterMetrics` (`Telemetry.variant = 11`).
- This PR does **not** populate the old flat tags 42–56. A follow-up will retarget `fillExtended*` to `variant.soil_water_metrics` after firmware syncs that protobuf change.

## Test plan
- [ ] `pio run -e rak_wismesh_sensorhub` builds
- [ ] `pio run -e rak2560` still builds / behavior unchanged (RAK9154 path)
- [ ] On hardware: probe discovery + poll logs; MQTT/app shows T/RH/P (and other standard env fields when present)
- [ ] Confirm chemistry fields are **not** present in on-air environment telemetry with default flags
- [ ] `trunk check` / CI format gate green

## Summary by CodeRabbit

* **New Features**
  * Added support for the RAK WisMesh SensorHub.
  * SensorHub probes can provide environmental, air-quality, soil, water, weather, motion, lighting, and power metrics.
  * Added SensorHub-specific build support for compatible RAK2560 hardware.
  * Added battery level, charging state, voltage, and current reporting.
  * Supports automatic probe discovery, hot-plug events, and reliable periodic sensor polling.
  * Sensor readings and power metrics are reported only while sufficiently recent.


src/modules/Telemetry/EnvironmentTelemetry.cpp
```diff
@@ -423,6 +427,11 @@ int32_t EnvironmentTelemetryModule::runOnce()
             if (rak9154Sensor.hasSensor())
                 result = rak9154Sensor.runOnce();
 #endif
+#ifdef HAS_RAKHUB
+            // 1-Wire SensorHub poll (not I2C); same UART as legacy RAK9154 but different env macro
+            if (rakSensorHub.hasSensor())
+                result = rakSensorHub.runOnce();
+#endif
```
caveman99, line 434: is it technically possible to unify the 9154 and sensor hub protocol into one driver?

Ethan-chen1234-zy: Yes — same RakSNHub half-duplex UART and IPSO payloads. RAK9154 is a single-PID power slave, not a second framing.

We are keeping two drivers for now because they are two shipping configs of the same board, not two protocols. env:rak2560 is already in the field as the solar-battery WisMesh Hub: quiet one-shot poll, power IPSO only, wired into Power.cpp. This PR’s SensorHub path is a multi-PID master (discovery, hot-plug, half-duplex gating, full env IPSO) and is intentionally a separate env. Collapsing them here would change the released 9154 poll/endian/scale behavior without a solar regression.

A unified master is the right follow-up when the 9154 path can be a profile of this driver (and mixed solar+probes would actually need that). Out of scope for this PR.

Ethan-chen1234-zy:
@coderabbitai thanks — the moisture raw == 0 skip is intentional, same sentinel as the other IPSO parsers (missing/not-ready), not a dry 0% sample. Leaving the last good reading for the freshness window. Out of scope unless a soil probe actually reports dry as 0.

Ethan-chen1234-zy:
@coderabbitai  No follow-up issue needed. If a probe later reports dry as 0%, we can change the parser then. Current raw == 0 sentinel stays.

src/modules/Telemetry/Sensor/RAKSensorHubUplink.cpp
```diff
@@ -0,0 +1,1020 @@
...
+        if (len < 7)
+            return false;
+        int16_t x = (int16_t)((msg[2] << 8) + msg[1]);
+        int16_t y = (int16_t)((msg[4] << 8) + msg[3]);
+        int16_t z = (int16_t)((msg[6] << 8) + msg[5]);
+        env.accel.x = x / 1000.0f;
+        env.accel.y = y / 1000.0f;
+        env.accel.z = z / 1000.0f;
+        env.accel.valid = true;
+        env.accel.lastUpdateMs = now;
```
caveman99: trunk fmt says: lastUpdateMs is 0-means-unset - arm it with Time::timerEndsAtMillis(delay), or Time::skipZero(Time::getMillis()) for a stamp (see src/UptimeClock.h)

caveman99 (changes requested):

src/modules/Telemetry/Sensor/RAKSensorHubUplink.cpp
Ethan-chen1234-zy: Stamped lastUpdateMs (and the other 0-unset cache times) with Time::skipZero / Time::stampMillis(), not timerEndsAtMillis.

caveman99 (approved):

src/modules/Telemetry/Sensor/RAKSensorHubUplink.cpp
```diff
@@ -0,0 +1,1021 @@
...
+        measurement->variant.environment_metrics.has_soil_moisture = true;
+        measurement->variant.environment_metrics.soil_moisture =
+            (uint8_t)(env.moisture.value < 0 ? 0 : (env.moisture.value > 100 ? 100 : (uint32_t)env.moisture.value));
+        any = true;
+    }
+    if (scalarFresh(env.high_precision_humidity, now, maxAgeMs)) {
+        // IPSO 0x70 is air RH on weather probes and water content on soil probes.
+        float v = env.high_precision_humidity.value;
+        const uint8_t pid = hpHumidityPidValid ? hpHumidityPid : 0;
+        if (hpHumidityIsSoil(pid)) {
```
oscgonfer, line 965: hi @Ethan-chen1234-zy this solves the double sending of data to the phone app on two fields we discussed, right?

Ethan-chen1234-zy: @oscgonfer Yes — 0x70 now goes to only one field: weather probe → relative_humidity, soil probe → soil_moisture. The app should not show the same value on both.

Issue #11796, opened 2026-09-09
[Bug]: Display doesn't wake via button after sleep on Heltec V4 (RST/power-cycle required)


MrMythoclast:
### Category

Other

### Hardware

Other

### Is this bug report about any UI (https://meshtastic.org/docs/configuration/device-uis/) component firmware?

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [ ] BaseUI

### Firmware Version

2.7.15.567b8ea

### Description

Heltec v4

After some time running purely on USB power (no battery attached), the display turns off and can no longer be woken via the button (USER button). Only pressing RST or disconnecting/reconnecting USB brings the display back.

Reproducible across multiple firmware versions (2.8.1, 2.7.26, 2.7.15), each time after a complete erase + reflash via flasher.meshtastic.org, with no custom configuration applied.

To rule out a hardware defect: with MeshCore firmware (identical hardware, Companion USB), button-wake works fine. This points to a Meshtastic-specific bug in the display sleep/wake logic for this board, rather than defective hardware.

Steps to reproduce:

1. Fully erase the Heltec V4 and flash it with current Meshtastic firmware
2. Connect via USB (no battery)
3. Leave the device idle until the display goes off
4. Press the USER button → display stays off
5. Press RST or disconnect/reconnect USB → display works normally again

### Relevant log output

```Shell

```

caveman99:
@MrMythoclast can you check whether the linked PR resolves your issue? once it builds you can flash the device right from the PR.

PR #11821, opened 2026-09-11
fix(power): stop a battery-less board deep-sleeping itself forever

caveman99:
A battery-less board's floating divider drifts through the battery-present window, and because the low-battery counter only reset inside its `hasBattery && !hasUSB` guard it ratcheted up across the gaps until the node entered a ~24.8-day deep sleep that the button could not wake it from, since `doDeepSleep()` force-holds the very pad `ext1_wakeup_prepare()` needs to re-route. Candidate fix for #11796; the counter change is pinned by a new `test_low_battery_shutdown` suite (7/7, and it goes red when the fix is reverted), and heltec-v4 plus tbeam build clean. The ext1 half is reasoned from the pinned ESP-IDF rather than observed, so it wants a check on real hardware — a serial log showing `Power: battery hardware detected` on a battery-less V4 followed by `Low voltage counter: N/10` would confirm the diagnosis.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved low-battery shutdown reliability by requiring consistent low-voltage readings before entering deep sleep.
  * Low-voltage tracking now resets when USB power is connected, no battery is detected, or voltage returns to a safe level.
  * Battery cutoff detection now correctly accounts for multi-cell battery packs.
  * Improved deep-sleep button handling so the configured wake button remains responsive after waking.
  * Prevented wake-button pins from being held in a state that could interfere with wake-up detection.

### DIFF
--- a/src/Power.cpp
+++ b/src/Power.cpp
@@ -784,7 +784,7 @@ class ADS1115BatteryLevel : public AnalogBatteryLevel
                     sum += _ads.computeVolts(raw);
                 }
                 // Piggyback a toggle-engine watchdog on this same throttle interval.
-                // Only re-arm when VBUS is absent — calling this while attached
+                // Only re-arm when VBUS is absent - calling this while attached
                 // would restart CC toggling and could glitch an active sink attach.
                 if (_aw35615.isReady() && !_aw35615.isVbusPresent()) {
                     _aw35615.rearmToggle();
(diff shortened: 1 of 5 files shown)


MrMythoclast:
CI is green (23m54s success) but the Web Flasher comment is still stuck on "Building this pull request…" after 2+ hours. Also don't see a heltec-v4 firmware artifact in the build artifacts list — is that expected, or is the flasher automation stuck

caveman99:
The correction push was too fast. the flasher build doesn't pull from the latest build attempt, and the previous one was cut short. And yes, the V4 is not in the PR build matrix. I'll merge this change and the next nightly build will carry it.

PR #11826, opened 2026-09-11
fix(hopscale): gate hop scaling on measured channel congestion

### LINKED ISSUE #11794: [Bug]: Hop Scaling is too aggressive on faster and non-default presets
### Category

Other

### Hardware

Not Applicable

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [ ] BaseUI

### Firmware Version

2.8.1.xxxxxxx

### Description

I'm asking for the Hop Scaling feature to have increased role floors based on LoRa Preset selection. Faster and non-default presets other than LongFast/LongTurbo receive a roleFloor = 3 to bring the telemetry scaling in line with the greater throughput and lower congestion inherent to a faster non default preset.

The Hop Scaling feature with histogram is great for congested meshes. Along with rejecting packets without a hop_start value and disabling telemetry on upgrade it we have a lot of potential to reduce congestion when 2.8.x is widely adopted. 

However, for larger meshes on a non-default and faster preset, such as our MediumSlow mesh in Coastal British Columbia, the Hop Scaling is quite aggressive. Our mesh spans well over 500km end to end and router are located on remote mountains that are inaccessible a large portion of the year. We rely on telemetry received over 3,4, even 5 hops from these remote routers. 

We don't have congestion issues right now on MediumSlow and most nodes running 2.7.X but the nodes that have been upgraded to 2.8.x prerelease are all ceasing to report valuable telemetry data. So far this is mainly affecting routers as those are the few that have been upgraded for testing. 

The telemetry from Routers is very important for us to measure the health of our network. Packets take a variety of paths. Not always the shortest path. Combined with mandatory 12 hour telemetry intervals for routers this means that an administrator a few hops away can go for days without seeing any telemetry from their routers. This means the administrator has no material way of telling if the router is in fact still online and healthy without driving a great distance (hundreds of kilometres in some cases) to manually request telemetry. Not a scalable solution. 

The routers I have upgraded as test units just one town over have almost completely stopped reporting telemetry to me. These are generally 2 or 3 hops away and only a distance of 45-60km as the crow flies but they have gone silent. Channel Utilization on these units was a comfortable 10-15% before with occasional spikes into the high teens. Not really in the range where they should have telemetry hops scaled, IMHO. 

Additionally we rely on users with mobile client nodes on vehicles to send regular position data which helps us identify coverage holes by cross referencing with their actual locations. The scaling negatively affects that use case as well. 

Overall its a fantastic idea for meshes on the old and new defaults. Perfect, no notes. 

But for meshes that already moved to a faster preset to solve their congestion issues and have no congestion its taking away the core features that make Meshtastic cool and useful for a variety of tasks besides sending messages. 

Here's my clumsy bodge job in HopScalingModule.h for as an example. Router.cpp already honours the user-configured hop limit — line 536: if (variableHopLimit < p->hop_limit). The scaling only reduces the hop limit, never increases it.

    if (didHourlyUpdate) {
        uint8_t suggested = (histogramRollCount > 0 && count > 0) ? lastSuggestedHop : HOP_MAX;
        // Preset-based hop floor: on non-default presets (anything other than LONG_FAST or LONG_TURBO)
        // the mesh is assumed to be less congested, so a minimum floor of 4 is applied to ensure
        // remote nodes remain visible. On default presets, scaling runs freely.
        uint8_t roleFloor = 0;
        if (config.lora.modem_preset != meshtastic_Config_LoRaConfig_ModemPreset_LONG_FAST &&
            config.lora.modem_preset != meshtastic_Config_LoRaConfig_ModemPreset_LONG_TURBO) {
            roleFloor = 3;
        }
        lastRequiredHop = std::max(suggested, roleFloor);
    }

### Relevant log output

```Shell

```

caveman99:
Closes #11794

Hop scaling now applies only while measured channel utilization says the node is congested (engage 25%, release 15%, each confirmed over three 5-minute EMA samples), so a dense but idle mesh is no longer throttled as hard as a saturated one. When congestion clears the applied limit steps back up one hop per hourly roll instead of jumping to HOP_MAX, and ROUTER, ROUTER_LATE and REPEATER gain a 3-hop floor on their own broadcasts. The thresholds are proposed and open to adjustment; a user-facing opt-out is left as follow-up because it needs a new field in meshtastic/protobufs.

## Summary by CodeRabbit

- **New Features**
  - Added congestion-aware hop scaling based on measured channel utilization.
  - Hop adjustments respond to sustained congestion while avoiding rapid state changes.
  - Scaling gradually eases back as channel conditions improve.
  - Infrastructure roles maintain a minimum hop count during congestion.
  - Added smoothed channel-utilization tracking for more stable decisions.
  - Congestion engages at 20% utilization and releases at 12%.
  - Very high utilization applies the strictest hop-scaling behavior.

### DIFF
--- a/src/mesh/Default.h
+++ b/src/mesh/Default.h
@@ -58,12 +58,13 @@ enum class TrafficType { POSITION, TELEMETRY };
 #define default_hop_scaling_min_target_nodes_floor 5     // minimum allowed min_target_nodes
 #define default_hop_scaling_max_target_nodes_ceiling 512 // maximum allowed max_target_nodes
 
-// Congestion gate: hop scaling only applies while the channel is measurably busy. Engage matches
-// AirTime::polite_channel_util_percent, the point the radio already treats as busy for metadata.
-#define default_hop_scaling_congestion_engage_pct 25  // smoothed channel utilization that engages scaling
-#define default_hop_scaling_congestion_release_pct 15 // smoothed channel utilization that releases it
+// Congestion gate: hop scaling only applies while the channel is measurably busy. Engage sits below
+// AirTime::polite_channel_util_percent (25) so hops ease back before the polite gate withholds traffic.
+#define default_hop_scaling_congestion_engage_pct 20  // smoothed channel utilization that engages scaling
+#define default_hop_scaling_congestion_release_pct 12 // smoothed channel utilization that releases it
 #define default_hop_scaling_congestion_confirm_runs 3 // consecutive 5-min samples needed to flip either way
-#define default_hop_scaling_router_hop_floor 3        // hop floor for a router/repeater's own broadcasts
+// Hop floor for the infrastructure roles Router.cpp already groups for zero-cost hops.
+#define default_hop_scaling_infrastructure_hop_floor 3
 
 #ifdef USERPREFS_RINGTONE_NAG_SECS
 #define default_ringtone_nag_secs USERPREFS_RINGTONE_NAG_SECS
--- a/src/mesh/Default.h
+++ b/src/mesh/Default.h
@@ -58,6 +58,13 @@ enum class TrafficType { POSITION, TELEMETRY };
 #define default_hop_scaling_min_target_nodes_floor 5     // minimum allowed min_target_nodes
 #define default_hop_scaling_max_target_nodes_ceiling 512 // maximum allowed max_target_nodes
 
+// Congestion gate: hop scaling only applies while the channel is measurably busy. Engage matches
+// AirTime::polite_channel_util_percent, the point the radio already treats as busy for metadata.
+#define default_hop_scaling_congestion_engage_pct 25  // smoothed channel utilization that engages scaling
+#define default_hop_scaling_congestion_release_pct 15 // smoothed channel utilization that releases it
+#define default_hop_scaling_congestion_confirm_runs 3 // consecutive 5-min samples needed to flip either way
+#define default_hop_scaling_router_hop_floor 3        // hop floor for a router/repeater's own broadcasts
+
 #ifdef USERPREFS_RINGTONE_NAG_SECS
 #define default_ringtone_nag_secs USERPREFS_RINGTONE_NAG_SECS
 #else
--- a/src/mesh/Default.h
+++ b/src/mesh/Default.h
@@ -63,6 +63,9 @@ enum class TrafficType { POSITION, TELEMETRY };
 #define default_hop_scaling_congestion_engage_pct 20  // smoothed channel utilization that engages scaling
 #define default_hop_scaling_congestion_release_pct 12 // smoothed channel utilization that releases it
 #define default_hop_scaling_congestion_confirm_runs 3 // consecutive 5-min samples needed to flip either way
+// Above this the hop walk's one-hop extension is cut to its strictest setting: the radio is
+// already withholding metadata at the polite gate, so an extra relay is the wrong thing to spend.
+#define default_hop_scaling_congestion_strict_pct 25 // = AirTime::polite_channel_util_percent
 // Hop floor for the infrastructure roles Router.cpp already groups for zero-cost hops.
 #define default_hop_scaling_infrastructure_hop_floor 3
 
--- a/src/modules/HopScalingModule.cpp
+++ b/src/modules/HopScalingModule.cpp
@@ -67,12 +67,6 @@ void HopScalingModule::clear()
     lastPoliteNumer = POLITENESS_DEFAULT;
     lastTrendStats = {};
     memset(denominatorHistory, DENOM_MIN, sizeof(denominatorHistory));
-    // A cleared instance that kept congested=true would scale again as soon as the histogram
-    // refilled. lastRequiredHop is left alone: the released gate walks it back up per hourly roll.
-    utilizationAvg = 0.0f;
-    hasUtilizationSample = false;
-    congested = false;
-    congestionConfirmRuns = 0;
 #ifndef PIO_UNIT_TESTING
     hashSeed = static_cast<uint16_t>(random());
 #else
--- a/src/modules/HopScalingModule.cpp
+++ b/src/modules/HopScalingModule.cpp
@@ -445,7 +445,7 @@ void HopScalingModule::updateCongestion()
 
     // Separate engage/release thresholds, each confirmed over several ticks, so a mesh sitting
     // near a threshold does not flap the hop limit between rolls.
-    const bool wantsFlip = congested ? (utilizationAvg < CONGESTION_RELEASE_PCT) : (utilizationAvg >= CONGESTION_ENGAGE_PCT);
+    const bool wantsFlip = congested ? (utilizationAvg <= CONGESTION_RELEASE_PCT) : (utilizationAvg >= CONGESTION_ENGAGE_PCT);
     congestionConfirmRuns = wantsFlip ? static_cast<uint8_t>(congestionConfirmRuns + 1u) : 0u;
     if (congestionConfirmRuns >= CONGESTION_CONFIRM_RUNS) {
         congested = !congested;
--- a/src/modules/HopScalingModule.cpp
+++ b/src/modules/HopScalingModule.cpp
@@ -7,6 +7,7 @@
 #include "FSCommon.h"
 #include "NodeDB.h"
 #include "SPILock.h"
+#include "airtime.h"
 #include "concurrency/LockGuard.h"
 #include "mesh-pb-constants.h"
 #include <algorithm>
@@ -318,9 +319,10 @@ void HopScalingModule::rollHour()
         }
         const uint32_t scaledTotal = static_cast<uint32_t>(counts.total) * filteringDenominator;
         memcpy(lastScaledPerHop, scaled, sizeof(lastScaledPerHop));
-        LOG_INFO("[HOPSCALE] rollHour: entries=%u/128 samp=1/%u filt=1/%u counted=%u est=%u suggestedHop=%u polite=%u/4", count,
-                 samplingDenominator, filteringDenominator, counts.total, static_cast<unsigned>(scaledTotal), suggested,
-                 lastPoliteNumer);
+        LOG_INFO("[HOPSCALE] rollHour: entries=%u/128 samp=1/%u filt=1/%u counted=%u est=%u suggestedHop=%u polite=%u/4 "
+                 "congested=%u chanUtil=%u%%",
+                 count, samplingDenominator, filteringDenominator, counts.total, static_cast<unsigned>(scaledTotal), suggested,
+                 lastPoliteNumer, congested ? 1u : 0u, static_cast<unsigned>(utilizationAvg));
 
         const auto &ts = lastTrendStats;
         LOG_INFO("[HOPSCALE] scaledSeenPerHour (h0=now): [%u %u %u %u %u %u %u %u %u %u %u %u %u]", ts.scaledPerHour[0],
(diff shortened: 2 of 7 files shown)


src/modules/HopScalingModule.cpp
```diff
@@ -424,17 +426,47 @@ void HopScalingModule::trimIfNeeded()
...
+{
+    const float util = channelUtil();
+    // channelUtilizationPercent() covers only the 60 s before each 5-minute tick, so smooth it:
+    // one quiet or one busy minute must not move the gate on its own.
+    utilizationAvg = hasUtilizationSample ? utilizationAvg + (util - utilizationAvg) * CONGESTION_EMA_ALPHA : util;
+    hasUtilizationSample = true;
+
+    // Separate engage/release thresholds, each confirmed over several ticks, so a mesh sitting
+    // near a threshold does not flap the hop limit between rolls.
+    const bool wantsFlip = congested ? (utilizationAvg < CONGESTION_RELEASE_PCT) : (utilizationAvg >= CONGESTION_ENGAGE_PCT);
```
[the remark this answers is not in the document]
caveman99: Applied in 7568feaa7, but the stated failure mode is not fixed by it and cannot be.

The EMA is `avg += (sample - avg) * 0.25f` in float32. Starting from the engage seed of 25 and feeding a sustained 15.0, it reaches a fixed point after 54 ticks at 15.000001907348633, not 15.0: once `(15 - avg) * 0.25f` drops below half an ULP of 15 (about 4.8e-7), the addition rounds back to the same value. So `avg <= 15.0f` is false there too, and the gate stays engaged with either operator.

I took the change anyway because the asymmetry was real: engage was inclusive (`>= 25`) and release exclusive (`< 15`) for no stated reason, and `<=` matters whenever the average lands exactly on the threshold from below.

I did not add the regression test you asked for, because a sustained 15% sample does not release the gate and the test would fail. A genuine fix for the at-threshold case would have to stop comparing a noisy percentage at sub-ULP precision, for example by truncating to whole percent before the comparison. That is a behaviour change (release would then fire anywhere under 16%) and I would rather a maintainer decide it than smuggle it in here.

src/modules/HopScalingModule.cpp
```diff
@@ -449,6 +481,9 @@ int32_t HopScalingModule::runOnce()
     const bool isFirstRun = !hasCompletedInitialRun;
     bool didHourlyUpdate = false;
 
+    // Sampled every tick, not only on a roll, so the gate reacts within minutes of a change.
+    updateCongestion();
+
```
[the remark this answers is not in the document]
caveman99: Verified the timing claim: the constructor arms `INITIAL_DELAY_MS` at 30 s and `runOnce()` returns `RUN_INTERVAL_MS` at 5 min, so samples land at 30 s, 5 m 30 s and 10 m 30 s. Three confirmations by 10 m 30 s rather than 15 m is correct.

Not changing it, because it has no observable effect. `lastRequiredHop` only moves inside `if (didHourlyUpdate)`. The first run sets that flag while the gate still has one sample and is released, so it takes the ramp branch and `lastRequiredHop` is already `HOP_MAX`. The next hourly update needs `runsSinceLastHourlyUpdate >= RUNS_PER_HOUR` (12 more ticks), which is roughly 65 minutes after boot. The gate has settled long before anything is applied, so gating `updateCongestion()` on elapsed time would add a timestamp and a branch to change nothing a node can observe.

src/modules/HopScalingModule.h
```diff
@@ -309,6 +331,14 @@ class HopScalingModule : private concurrency::OSThread
     uint8_t lastRequiredHop = HOP_MAX;
     uint8_t histogramRollCount = 0;
 
+    // -----------------------------------------------------------------------
+    // Congestion state
+    // -----------------------------------------------------------------------
+    float utilizationAvg = 0.0f;
+    bool hasUtilizationSample = false;
+    bool congested = false;
+    uint8_t congestionConfirmRuns = 0;
```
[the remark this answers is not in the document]
caveman99: Not changing this one. The stated failure mode does not occur: `clear()` zeroes `count`, and `runOnce()` computes `suggested` as `(histogramRollCount > 0 && count > 0) ? lastSuggestedHop : HOP_MAX`, so a cleared instance yields `HOP_MAX` and cannot apply stale scaling.

The only caller is the constructor, where all four fields already hold their initialiser values. Beyond that, `clear()` deliberately resets histogram state and derived hourly results and leaves the applied output alone: `lastRequiredHop` and `histogramRollCount` are untouched today. Congestion is a live radio measurement in that second group, so resetting it while still leaving `lastRequiredHop` behind would make the split less consistent, not more.

Happy to reset all of it, including `lastRequiredHop`, if a maintainer wants `clear()` to mean a full reset, but that is a wider change than this PR.

caveman99:
Re the outside-diff comment on `test/test_hop_scaling/test_main.cpp:144-147` (busy-channel test starts from a reduced `lastRequiredHop`):

The premise does not hold. `lastRequiredHop` is written in exactly two places, `HopScalingModule.cpp:507-508` and `:531`, both inside `if (didHourlyUpdate)` in `runOnce()`. `injectSampleTraffic()` drives `rollHour()` directly through the test shim and never calls `runOnce()`, so it moves `lastSuggestedHop` but leaves `lastRequiredHop` at its `HOP_MAX` initialiser. The `<= 3` assertion therefore already required the gate to re-engage and an hourly update to apply the recommendation, and could not pass vacuously.

Took the hardening anyway in 8a12b27dd, since pinning the precondition is worth it regardless of whether it can currently regress:

```cpp
TEST_ASSERT_EQUAL_UINT8(HOP_MAX, shim->getLastRequiredHop());
```

placed before the 45% pump. It passes, which is the direct evidence that nothing was applied beforehand.

Not restoring `s_testChannelUtil` per test: every test that depends on it sets it explicitly, and the suite has no `setUp`/`tearDown` in this branch, so a teardown would be new machinery for no current defect. No `state-manifest.tsv` change either: the suite writes the same `/prefs` path it already did and adds no new persisted files, so the existing `test_hop_scaling` row still describes its footprint.

src/modules/HopScalingModule.cpp
```diff
@@ -466,23 +501,35 @@ int32_t HopScalingModule::runOnce()
...
+        } else {
+            uint8_t suggested = (histogramRollCount > 0 && count > 0) ? lastSuggestedHop : HOP_MAX;
+            // Role-based hop floor: TRACKER/TAK_TRACKER always reach at least 2 hops,
+            // SENSOR reaches at least 1, router-class roles reach ROUTER_HOP_FLOOR, so these
+            // reporting roles remain reachable even on a dense mesh recommending fewer hops.
+            uint8_t roleFloor = 0;
+            switch (config.device.role) {
+            case meshtastic_Config_DeviceConfig_Role_ROUTER:
+            case meshtastic_Config_DeviceConfig_Role_ROUTER_LATE:
+            case meshtastic_Config_DeviceConfig_Role_REPEATER:
```
NomDeTom: what is a repeater?

test/test_hop_scaling/test_main.cpp
```diff
@@ -494,6 +519,186 @@ void test_startup_blank_state()
...
+        shim->runOnce();
+        clientHop = shim->getLastRequiredHop();
+        hopScalingModule = nullptr;
+    }
+    TEST_MSG_FMT("CLIENT: hop=%u", clientHop);
+    TEST_ASSERT_TRUE(clientHop < HopScalingModule::ROUTER_HOP_FLOOR);
+
+    const meshtastic_Config_DeviceConfig_Role routerRoles[] = {meshtastic_Config_DeviceConfig_Role_ROUTER,
+                                                               meshtastic_Config_DeviceConfig_Role_ROUTER_LATE,
+                                                               meshtastic_Config_DeviceConfig_Role_REPEATER};
```
NomDeTom: Does this mean repeater or does it mean client_base?

test/test_hop_scaling/test_main.cpp
```diff
@@ -494,6 +519,186 @@ void test_startup_blank_state()
...
+}
+
+// The role floor is keyed on the sending node's own role, so this lets a remote router's telemetry
+// travel without loosening anything for client nodes. Operators read that telemetry to know a
+// mountain-top site is alive; issue #11794 is a report of exactly those routers going quiet.
+void test_router_role_floor_applies_when_congested()
+{
+    TEST_MESSAGE("=== Role floor: router-class roles keep a minimum hop count ===");
+    TEST_MESSAGE("Topology: 200 nodes at hop 0, so the hop walk recommends 0 for an unfloored role.");
+    TEST_MESSAGE("Expectation: CLIENT scales below the floor, ROUTER/ROUTER_LATE/REPEATER sit on it.");
```
NomDeTom: again - does this mean repeater?

src/mesh/Default.h
```diff
@@ -58,6 +58,13 @@ enum class TrafficType { POSITION, TELEMETRY };
 #define default_hop_scaling_min_target_nodes_floor 5     // minimum allowed min_target_nodes
 #define default_hop_scaling_max_target_nodes_ceiling 512 // maximum allowed max_target_nodes
 
+// Congestion gate: hop scaling only applies while the channel is measurably busy. Engage matches
```
NomDeTom: The polite gate is the point where nodes would normally stop sending packets, not the point where they would consider scaling back the hops. I think it needs to have a gentler lead-in to this point, even if the effect is overlapping.

src/mesh/Default.h
```diff
@@ -58,6 +58,13 @@ enum class TrafficType { POSITION, TELEMETRY };
...
+// AirTime::polite_channel_util_percent, the point the radio already treats as busy for metadata.
+#define default_hop_scaling_congestion_engage_pct 25  // smoothed channel utilization that engages scaling
```
NomDeTom: do we need another gate? can we replace the activity/politeness perhaps-extra-hop with this channel activity gate?

src/modules/HopScalingModule.cpp
```diff
@@ -424,17 +426,47 @@ void HopScalingModule::trimIfNeeded()
...
+    return s_testChannelUtil;
+#else
+    return airTime ? airTime->channelUtilizationPercent() : 0.0f;
+#endif
+}
+
+void HopScalingModule::updateCongestion()
+{
+    const float util = channelUtil();
+    // channelUtilizationPercent() covers only the 60 s before each 5-minute tick, so smooth it:
```
NomDeTom: can we use this for the OTA telemetry figure? It's nigh-on useless in the current snapshot format.

src/modules/HopScalingModule.cpp
caveman99: One of the original finding was only partial coverage of the roles, and treating uncovered roles unfavourable. I think this comes from simply iterating the roles enum, ignoring the deprecated.

src/mesh/Default.h
caveman99: correct, lowered engage to 20 and release to 12, to have a bit of a hystereses and avoid flapping

src/mesh/Default.h
caveman99: no, politeness drop inconming.

src/modules/HopScalingModule.cpp
caveman99: This is gated on HAS_VARIABLE_HOPS so a board that doesn't compile it would have no telemetry. The smoothing belongs in AirTime next to channelUtilizationPercent(), then both the gate and telemetry read it. Ill will do this in a separate PR.

caveman99:
@coderabbitai REPEATER has been deprecated a long time ago and is demoted to client on any config change. The other outside diff finding you retracted yourself 12h ago.

caveman99:
Re the outside-diff comment on `src/mesh/Default.h:61-68`, "Set the congestion thresholds to 25% and 15%":

Not doing this. There is no 25%/15% contract to violate. `15` does not appear anywhere in the hop-scaling block, and the only `25` in it is `default_hop_scaling_congestion_strict_pct`, annotated `= AirTime::polite_channel_util_percent` — that is the politeness band edge, not the engage threshold.

20/12 are deliberate. They were 25/15 in the first revision of this PR and were changed at @NomDeTom's request, reviewing this exact line:

> The polite gate is the point where nodes would normally stop sending packets, not the point where they would consider scaling back the hops. I think it needs to have a gentler lead-in to this point, even if the effect is overlapping.

Engaging at 25 put hop scaling on the same sample as `isTxAllowedChannelUtil(polite)`, so a node started withholding metadata and shortening hops at once. The ladder is now graduated: 20 hops ease back, 25 polite traffic withheld, 40 all traffic withheld. Reverting to 25 restores the coincidence the review asked to remove.

12 follows from that: the hysteresis gap has to sit below the engage point, and it keeps the reporter's measured 10-15% band with high-teens spikes from engaging at all, which is the failure #11794 reports.

NomDeTom (approved):

Issue #11666, opened 2026-08-30
[Bug]: ESP32-S3 PSRAM heap scan triggers WiFi config watchdog


giannoug:
### Category

WiFi

### Hardware

Other

### Is this bug report about any UI (https://meshtastic.org/docs/configuration/device-uis/) component firmware?

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [ ] BaseUI

### Firmware Version

2.8.0.7239fe8

### Description

The affected hardware is a **LilyGo T5 E-Paper S3 Pro V2 / H752-01**, built with the `t5s3_epaper_inkhud` PlatformIO environment. It is listed as “Other” because this device/target is not available in the issue form’s Hardware dropdown.

Connecting through the TCP/WiFi API and requesting config consistently reboots this PSRAM-equipped ESP32-S3:

~~~text
[ApiServer] Incoming wifi connection
[ServerAPI] Client wants config, nonce=69420
Guru Meditation Error: Core 1 panic'ed (Interrupt wdt timeout on CPU1)
~~~

Decoded against the exact ELF:

~~~text
CPU1:
tlsf_walk_pool
multi_heap_get_info
heap_caps_get_info
heap_caps_get_largest_free_block
getFiles
PhoneAPI::handleStartConfig
StreamAPI::readStream
ServerAPI<NetworkClient>::runOnce

CPU0:
multi_heap_internal_lock
multi_heap_malloc
heap_caps_malloc_default
malloc
wifi_malloc
wDev_ProcessRxSucData
~~~

`heap_caps_get_largest_free_block(MALLOC_CAP_DEFAULT)`, added in #11537, walks every TLSF block in each matching heap while holding its lock. On this board's large PSRAM heap, CPU1 holds the allocator lock long enough for WiFi RX on CPU0 to block in `wifi_malloc()` and trigger the interrupt watchdog.

### Reproduction

1. Boot `t5s3_epaper_inkhud` with WiFi enabled.
2. Connect a client through the TCP API.
3. Request configuration.
4. The node reboots immediately and reproducibly.

### A/B Result

On the same commit and hardware, replacing the full-heap largest-block query with a bounded `malloc()` probe prevents the crash. WiFi config sync then completes successfully.

The probe needs an observable access, such as a volatile byte write, because LTO otherwise removes the `malloc()`/`free()` pair.

### Expected Behavior

A WiFi config request should complete without holding the PSRAM allocator lock long enough to trigger the interrupt watchdog.

[...]

Issue #11538, opened 2026-08-18
[Bug]: HTTPS low-heap guard livelocks: skipped secureServer->loop() never reaps open TLS connections, heap never recovers


thebentern:
### Summary

Once free heap drops below `MIN_HEAP_FOR_SSL` (40 KB) while HTTPS connections are outstanding, the node's heap can never recover and every subsequent TCP-API / HTTPS connection fails until a reset. The node is alive and on WiFi the whole time.

### Mechanism

`handleWebResponse()` in `src/mesh/http/WebServer.cpp` (from #9693):

```cpp
if (freeHeap >= MIN_HEAP_FOR_SSL) {
    secureServer->loop();
} else {
    LOG_WARN("Low heap (%u bytes), skipping HTTPS processing", freeHeap);
}
```

`HTTPServer::loop()` (esp32_https_server) is the **only** place already-accepted connections are serviced and reaped: Step 1 iterates `_connections[]`, calls `->loop()` on the open ones (which is where `isClosed()` / the 20 s `HTTPS_CONNECTION_TIMEOUT` are evaluated) and `delete`s the closed ones. Skipping the whole loop under low heap therefore also skips servicing the up-to-`MAX_HTTPS_CONNECTIONS` (2) TLS connections that are already open. Their mbedTLS contexts + pbufs (~16-20 KB each) are never freed, so free heap never climbs back over 40 KB, so the loop is skipped forever. The guard's own precondition is what keeps it from clearing.

The library's per-connection timeout does not help: it is checked inside `HTTPConnection::loop()`, i.e. inside the code being skipped.

### Observed (Heltec V3, PR #11537 branch `ae5236bd3`, `Endor` AP)

Reproducer: hold a few raw TCP sockets on 80/4403 and start 2-3 TLS handshakes to 443 that are left pending, then poke `https://<ip>/` every ~6 s. Free heap goes 89.7 KB -> `Low heap (1868 bytes)` and then pins:

```
17:01:14 257 [WebServer] Low heap (1868 bytes), skipping HTTPS processing
17:02:08 311 [ApiServer] No heap for API connection (4512 bytes), dropping client
17:03:32 395 [ServerAPI] Got 1 files in manifest (limited to 64 entries/depth 3)
17:03:45 407 [WebServer] Low heap (11448 bytes), skipping HTTPS processing
17:04:45 468 [WebServer] Low heap (11448 bytes), skipping HTTPS processing
[...]

PR #10535, opened 2026-05-22
fix: mbed TLS crash in Arduino 3.x (pioarduino)

mverch67:
This PR fixes an mbedTLS crash e.g. when starting an HTTP client in a remote connection:

```
Guru Meditation Error: Core  0 panic'ed (LoadProhibited). Exception was unhandled.

Core  0 register dump:
PC      : 0x400556d5  PS      : 0x00060730  A0      : 0x821ab850  A1      : 0x3fcc31a0  
A2      : 0x00010100  A3      : 0x000100fc  A4      : 0x000000ff  A5      : 0x0000ff00  
A6      : 0x00ff0000  A7      : 0xff000000  A8      : 0x00000000  A9      : 0x0000c064  
A10     : 0x3fcc2fec  A11     : 0x00000000  A12     : 0x00000180  A13     : 0x3fcc316c  
A14     : 0xaaa43b47  A15     : 0x00000018  SAR     : 0x00000010  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00010100  LBEG    : 0x400556d5  LEND    : 0x400556e5  LCOUNT  : 0xffffffff  

0x400556d2: ?? ??:0
0x421ab84d: mbedtls_ssl_free_hostname at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/mbedtls/mbedtls/library/ssl_tls.c:2819
0x421abced: mbedtls_ssl_set_hostname at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/mbedtls/mbedtls/library/ssl_tls.c:2841
0x4201a235: start_ssl_client(sslclient_context*, IPAddress const&, unsigned long, char const*, int, char const*, bool, char const*, char const*, char const*, char const*, bool, char const**) at /home/manuel/.platformio/packages/framework-arduinoespressif32/libraries/NetworkClientSecure/src/ssl_client.cpp:296 (discriminator 8)
0x4201aaeb: NetworkClientSecure::connect(IPAddress, unsigned short, char const*, char const*, char const*, char const*) at /home/manuel/.platformio/packages/framework-arduinoespressif32/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp:148 (discriminator 1)
[...]

PR #11838, opened 2026-09-13
fix(http): give a TLS session the contiguous heap it actually needs (#6960)

### LINKED ISSUE #6960: [Bug]: HTTPS SSL_new_failed on ESP32 - both in 2.6.4 and 2.6.10

thebentern:
Closes #6960.

An HTTPS connection to a no-PSRAM ESP32-S3 fails its handshake with `MBEDTLS_ERR_SSL_ALLOC_FAILED` and the client gets a TCP reset. The reporters' original boards are no longer in scope — plain ESP32 has had `MESHTASTIC_EXCLUDE_WEBSERVER` since #10005 — but the Heltec V3 reproduction in the thread is still live on develop.

## What was wrong

`mbedtls_ssl_setup()` makes two independent `mbedtls_calloc(1, N)` calls, one inbound record buffer and one outbound (`ssl_tls.c:1410` and `:1420`). Each needs its own **contiguous** block of `CONTENT_LEN + 13 header + 320 overhead` = 16717 bytes symmetric.

The admission gate compared `ESP.getFreeHeap()` against 40000. That is a sum over every free block and says nothing about the largest one, so on a fragmented heap it waved the connection through and let the handshake be the thing that discovered there was no room. This was the open follow-up from #11538.

## What this does

**Sizes the two record buffers separately.** Inbound stays at the RFC maximum because browsers and our own MQTT/OTA clients receive full-size records; outbound only carries what this device sends. That is **12288 bytes of contiguous heap back per session**, the difference between one usable HTTPS connection and two. These are the ESP-IDF defaults — Espressif's arduino-lib-builder template is what turns the asymmetric split off, nothing in our ini did.

It replaces `CONFIG_MBEDTLS_SSL_MAX_CONTENT_LEN`, which no longer applies either way: that symbol is `depends on !MBEDTLS_ASYMMETRIC_CONTENT_LEN` in the IDF Kconfig, and mbedTLS dropped it in 3.0 (zero references in the 3.6 tree we build). The 16384 was actually coming from `ssl.h`'s own fallback.

No ABI hazard of the kind PR #10535 pinned six options against: none of the three gate a member of `mbedtls_ssl_context`, `_config` or `_session`, and `in_buf`/`out_buf` are pointers.

**Asks the allocator the question mbedTLS is about to ask.** The probe holds both record buffers at once, as `setup()` does, through the heap `CONFIG_MBEDTLS_*_MEM_ALLOC` pins mbedTLS to — a plain `malloc` would be served from PSRAM that mbedTLS never touches. Deliberately not `heap_caps_get_largest_free_block()`, which walks every block under the allocator lock and tripped the interrupt watchdog in #11666.

Closed connections are reaped before the measurement, so the library cannot free a slot behind it and accept into it unmeasured, and so the probe sees the memory a finished session just returned. The gate keys on free slots only, not on whether a client is waiting: `HTTPServer::loop()` runs its own `select()` *after* synchronously driving every open connection, so any "is someone knocking" pre-check is stale by construction. The answer is advisory — the heap can still turn between probe and `setup()` — which makes the failure rare rather than routine.

**Loops on short writes.** `mbedtls_ssl_write()` accepts at most one record per call and returns a short count for the rest. Neither the server library nor `Print::print()` loops on that, so any body larger than one record was quietly cut off — `/json/nodes` has been truncating at 16 kB on develop already. Looping on the returned count also covers a client that negotiates a smaller maximum fragment length, which a fixed-size chunker would not.

**Streams the node list.** `/json/nodes` built its whole body in one `std::string`. A full node DB is 200 entries of ~240 bytes and `std::string` grows by doubling, so it asked for a 64 kB contiguous block while still holding a 32 kB one, on a board with ~50 kB of heap, through `operator new`, which aborts rather than throws here. That is a reboot, not an error page.

**Stops every node paying for geofencing.** The crossing table reserved all 256 slots (4 kB exactly) at construction, though a node with no geofenced waypoints never tracks a crossing. It now allocates on the first one, steps capacity explicitly so there is a size to probe for, and hands the block back once the last waypoint is gone. The probe matters because the allocation has moved onto a live heap, where a `reserve()` that cannot find the block would abort; a refusal degrades into the bounded-drop path that already existed.

## Verification

| Check | Result |
| --- | --- |
| heltec-v3 clean build, IDF recompiled from this ini | pass |
| Compiled `CONFIG_MBEDTLS_SSL_OUT_CONTENT_LEN` in `sdkconfig.h` | 4096 |
| Full native suite | 1414 / 1414 |
| `test_geofence` | 26 / 26 |
| Flash vs develop (heltec-v3) | −20,656 bytes |
| Static RAM vs develop (heltec-v3) | −172 bytes |

The size delta is `.flash.rodata` −12,348 and `.flash.text` −7,796, measured against a worktree build of the merge base with its own freshly built IDF.

## Hardware validation

Bench Heltec V3 (ESP32-S3, 8 MB, no PSRAM) — the exact configuration in the report. Same board, same
network, A/B by flashing only the app partition at `0x10000` so the device's WiFi config and region
survived both legs.

| | develop `31f05ab` | this branch `b9156fd` |
|---|---|---|
| TLS handshake soak | **0 / 12** | **12 / 12** |
| Two simultaneous handshakes | 0 / 2 | 2 / 2 |
| Second handshake while one is held open | could not hold the first | OK |
| `/json/nodes` over TLS | unreachable | valid JSON, 6 nodes |
| Plain HTTP on :80 | OK | OK |

develop reproduces the issue verbatim — every handshake reset, not intermittent:

```
[HTTPS:E] mbedtls_ssl_setup res=FFFF8100 (SSL - Memory allocation failed)
[HTTPS:E] SSL setup failed. Aborting handshake. FID=52
WARN | [WebServer] Low heap (27384 bytes), not accepting HTTPS connections
```

That second line is the one @cpatulea posted on 2026-09-11. Under the same load this branch instead
shows `FID=52` held open while `FID=53` cycles four more connections through it, with no errors and
no low-heap warning.

The numbers explain why. Free heap is 84,392 B at boot but ~27 KB once WiFi is up. Two symmetric
record buffers need 33,434 contiguous bytes, so even a *single* session could not be set up on this
board; the asymmetric split brings a session to 21,146 B and two now fit.

## Not done here

- The library's own small writes (`printHeader`, the ≤1400 byte keep-alive cache drain) still do not loop, so a client negotiating a 512-byte maximum fragment length could truncate those. That needs a PR against `meshtastic/esp32_https_server` and a pin bump.
- `handleFsBrowseStatic` still buffers the whole `/static` listing. It is a few kB, so it was left alone.
- Every HTTP request takes 21-24 s to complete even though the body arrives immediately, which looks like the connection being held to the 20 s idle timeout rather than closed on `Connection: close`. It behaves identically on develop, so it is pre-existing and not touched here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when sending large or multi-record HTTPS responses.
  - Reduced HTTPS connection failures on memory-constrained devices.
  - Improved handling of multiple concurrent HTTPS connections.
  - Improved geofence operation under memory pressure and released unused memory when geofenced waypoints are removed.

- **Performance**
  - Reduced memory usage for HTTPS traffic.
  - Reduced memory overhead for geofence tracking by allocating storage as needed.
  - Improved efficiency when managing HTTPS connections.


src/mesh/http/ContentHandler.cpp
```diff
@@ -254,6 +254,21 @@ static std::string jsonNum(double v)
...
+    while (sent < body.size()) {
+        const size_t remaining = body.size() - sent;
+        const size_t written = res->write(reinterpret_cast<const uint8_t *>(body.data()) + sent, remaining);
+        // An error code arrives as a huge count, write() returning mbedtls' int through a size_t.
+        if (written == 0 || written > remaining)
+            return;
+        sent += written;
+    }
+}
+
```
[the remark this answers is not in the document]
caveman99: Fixed in d35d832b7. writeAll() now returns false on a zero or error-sized write, and handleNodes() returns at the first failure instead of writing further records or the trailer.

src/modules/GeofenceModule.cpp
```diff
@@ -73,6 +76,28 @@ GeofenceModule::Crossing GeofenceModule::classify(bool firstSighting, bool wasIn
...
+        target = GEOFENCE_MAX_CROSSING;
+
+    // Probed while the old block is still held, which is the state reserve() allocates in.
+    void *probe = malloc(target * sizeof(CrossingState));
+    if (!probe)
+        return false;
+    static_cast<volatile char *>(probe)[0] = 0; // observable, so LTO cannot delete the question
+    free(probe);
+
+    crossingInside.reserve(target);
```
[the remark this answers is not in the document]
caveman99: Fixed in 435c60299. The malloc()/free() probe and vector reserve() are replaced by realloc() into a unique_ptr with a free() deleter. realloc() returns the grown block or leaves the old one intact, so there is no window between probe and allocation, and a refusal falls into the bounded-drop path instead of aborting.

src/mesh/http/WebServer.cpp
```diff
@@ -101,16 +161,19 @@ static void handleWebResponse()
         if (isWebServerReady) {
             // Check heap before HTTPS processing - SSL requires significant memory
             if (secureServer) {
-                uint32_t freeHeap = ESP.getFreeHeap();
-                if (freeHeap >= MIN_HEAP_FOR_SSL) {
+                // Reap first so the probe sees the heap a finished session just returned.
+                secureServer->reapClosedConnections();
+                // With every slot busy loop() cannot accept, so the probe would buy nothing.
+                if (!secureServer->hasFreeConnectionSlot() || canAllocateTlsSession()) {
```
caveman99, line 167: The heap probe runs even when nobody is connecting. Whenever a slot is free it allocates and frees about 21.5 kB of internal heap. That happens every 50 ms while the web server is active and every 1 s when idle, on exactly the low-heap boards this PR targets.

src/mesh/http/WebServer.cpp
```diff
@@ -59,8 +61,44 @@ static const int32_t IDLE_INTERVAL_MS = 1000;
...
+{
+    if (ESP.getFreeHeap() < TLS_IN_BUFFER_BYTES + TLS_OUT_BUFFER_BYTES + TLS_HANDSHAKE_SLACK)
+        return false;
+
+    // Held together, as setup() does: one can fit where two do not.
+    void *in = heap_caps_malloc(TLS_IN_BUFFER_BYTES, TLS_PROBE_CAPS);
+    void *out = in ? heap_caps_malloc(TLS_OUT_BUFFER_BYTES, TLS_PROBE_CAPS) : nullptr;
+    const bool fits = in && out;
+    // Observable, so LTO cannot judge the pair dead and delete the question.
+    if (in)
```
caveman99, line 94: The volatile writes do nothing. The build uses -fno-lto, and heap_caps_malloc is not a function GCC can optimise away.

src/modules/GeofenceModule.cpp
```diff
@@ -101,8 +124,12 @@ bool GeofenceModule::shouldTrack(const meshtastic_Waypoint &wp, uint8_t notifica
 
 int GeofenceModule::onWaypointStoreChanged(const WaypointStore *store)
 {
-    (void)store;
-    crossingInside.clear();
+    crossingCount = 0;
+    // Give the block back once nothing can need it again.
+    if (store && store->getWaypoints().empty()) {
```
caveman99, line 129: The geofence block stays allocated after the last geofenced waypoint is gone

src/modules/GeofenceModule.cpp
```diff
@@ -138,13 +165,14 @@ void GeofenceModule::evaluatePosition(NodeNum node, const meshtastic_Position &p
...
             } else {
                 static bool warnedCrossingFull = false;
                 if (!warnedCrossingFull) {
-                    LOG_WARN("Geofence crossing-state full (%u); new (waypoint,node) pairs will not alert until space frees",
-                             (unsigned)GEOFENCE_MAX_CROSSING);
+                    LOG_WARN("Geofence crossing-state cannot grow (%u tracked, %u max); new (waypoint,node) pairs will "
+                             "not alert until space frees",
+                             (unsigned)crossingCount, (unsigned)GEOFENCE_MAX_CROSSING);
                     warnedCrossingFull = true;
                 }
```
caveman99, line 177: The one-time warnedCrossingFull warning can be used up by a brief realloc refusal.

src/mesh/http/WebServer.cpp
```diff
@@ -101,16 +161,19 @@ static void handleWebResponse()
...
+                if (!secureServer->hasFreeConnectionSlot() || canAllocateTlsSession()) {
                     secureServer->loop();
                 } else {
                     // Low heap: accept nothing new, but keep servicing open connections so they can time out
                     // and free their contexts - skipping them pins the heap below the threshold for good.
                     secureServer->serviceExistingConnections();
                     static uint32_t lastHeapWarning = 0;
                     if (lastHeapWarning == 0 || !Throttle::isWithinTimespanMs(lastHeapWarning, 30000)) {
-                        LOG_WARN("Low heap (%u bytes), not accepting HTTPS connections", freeHeap);
+                        LOG_WARN("No contiguous heap for a TLS session (%u free), not accepting HTTPS connections",
```
caveman99, line 175: The refusal log blames fragmentation even when the plain free-heap check failed

src/mesh/http/WebServer.cpp
```diff
@@ -69,22 +107,44 @@ class MeshHTTPSServer : public HTTPSServer
...
                 delete _connections[i];
                 _connections[i] = nullptr;
-            } else {
-                _connections[i]->loop();
             }
         }
     }
+
+    /// The first half of HTTPServer::loop(): drive and reap existing connections, accept nothing.
+    void serviceExistingConnections()
```
caveman99, line 125: serviceExistingConnections() reaps again right after its only caller already did, and the _running guard is repeated in all three helpers.

src/mesh/http/ContentHandler.cpp
```diff
@@ -725,7 +741,7 @@ void handleReport(HTTPRequest *req, HTTPResponse *res)
 
     out += "},\"status\":\"ok\"}";
 
-    res->print(out.c_str());
+    writeAll(res, out);
 }
 
 void handleNodes(HTTPRequest *req, HTTPResponse *res)
```
caveman99, line 747: handleNodes sends one TLS record per node. A 200-node DB means about 200 mbedtls_ssl_write calls. Buffering to about 2 to 4 kB before each write keeps memory bounded and cuts that work by roughly 20x.

src/mesh/http/WebServer.cpp
thebentern: Fixed in 132d3f085. The probe now runs only when a zero-timeout `select()` on the listen socket says a client is already in the backlog, so a server nobody is talking to does no allocation at all. With nothing knocking it drives its open connections and accepts nothing, so a connection is never accepted past an unmeasured gate; one that lands between that `select()` and `loop()`'s own waits a tick.

src/mesh/http/WebServer.cpp
thebentern: Fixed in a404fe44f. The probe returns a verdict and the warning names the check that failed - "Too little heap" for the free-size threshold, "No contiguous block" for the allocation.

src/modules/GeofenceModule.cpp
thebentern: Fixed in d5c215674. The full table keeps the one-shot. A refused malloc gets its own warning, throttled to once a minute, so a heap blip no longer spends the message that means the 256-entry cap was reached.

src/mesh/http/ContentHandler.cpp
thebentern: Fixed in 2fe43711b. `handleNodes` fills a 2 kB buffer and flushes that, so a 200-node DB is roughly 25 writes instead of 200. The buffer is reserved once with a node's headroom past the mark, so it never regrows.

src/mesh/http/WebServer.cpp
```diff
@@ -59,8 +61,44 @@ static const int32_t IDLE_INTERVAL_MS = 1000;
...
+#define TLS_PROBE_CAPS (MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT)
+#else
+#define TLS_PROBE_CAPS (MALLOC_CAP_8BIT)
+#endif
+
+/// Advisory: free heap is a sum and says nothing about the largest block, so a fragmented heap used
+/// to pass and fail the handshake instead (#6960). Not largest_free_block() - it trips the WDT (#11666).
+static bool canAllocateTlsSession()
+{
+    if (ESP.getFreeHeap() < TLS_IN_BUFFER_BYTES + TLS_OUT_BUFFER_BYTES + TLS_HANDSHAKE_SLACK)
```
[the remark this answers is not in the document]
thebentern: Addressed in a404fe44f. The free-size check reads `heap_caps_get_free_size()` for the heap `mbedtls_calloc()` draws from - `MALLOC_CAP_SPIRAM` under `CONFIG_MBEDTLS_EXTERNAL_MEM_ALLOC`, `MALLOC_CAP_INTERNAL` otherwise - instead of `ESP.getFreeHeap()`. The allocation probe itself is unchanged.

PR #11844, opened 2026-09-14
fix(router): relay opaque packets in CORE_PORTNUMS_ONLY

### LINKED ISSUE #11843: [Bug]: core-portnums-only option dropping/blocking packets on retransmission
### Category

Other

### Hardware

Heltec Mesh Node T114, Rak4631, Seeed Card Tracker T1000-E, DIY, Other

- [ ] Meshtastic UI aka MUI
- [ ] InkHUD
- [ ] BaseUI

### Firmware Version

2.8.1

### Description

Ok so today I was having an issue sending admin packets telemetry requests and direct messages through my tree node to a node a few km away. This lasted for over 12 hours, multiple reboots of the tree node and my admin node sending packets through the tree node.

I walked over to a location that I could hit the node i wanted to administer directly without hopping through my tree node, and it worked just fine.

I went home and set the retransmit from core-portnums-only to "ALL". Then all the problems I noted above immediately went away.

The tree node, the admin node and the administered node all are on 2.8.1

The admin node is in client mute, the tree node is in router-late and the administered node is in Router mode

TLDR: core-portnums-only setting appears to drop/block packets that it shouldn't in version 2.8.1

Meshtastic Discord chat about it here: https://discord.com/channels/867578229534359593/919642584480112750/1548793533761069177

### Relevant log output

```Shell

```

NomDeTom:
Fixes #11843. A packet a relay cannot decrypt - a PKI unicast between two other nodes, such as the remote admin request in the report, or a broadcast on a channel the relay does not hold - is relayed from its header alone, and only in the rebroadcast modes `NextHopRouter::relayOpaquePacket()` lists. `CORE_PORTNUMS_ONLY` was not among them, so a node in that mode, which is the `ROUTER` role default, dropped every such packet; switching the reporter's router to `ALL` made it work because `ALL` is listed. The portnum filter that `CORE_PORTNUMS_ONLY` exists for cannot be applied to a payload the relay cannot read, so this adds the mode to that list. One line; no other behaviour changes.

## Summary by CodeRabbit

* **New Features**
  * Added support for relaying eligible opaque packets without decoding them first.
  * Opaque packets can now be forwarded in supported rebroadcast modes while avoiding local delivery and duplicate processing.
  * Relay handling respects hop limits, routing constraints, and retransmission behavior.

* **Bug Fixes**
  * Improved route-health timing and retransmission scheduling for more consistent packet forwarding.


src/mesh/NextHopRouter.cpp
```diff
@@ -26,14 +26,30 @@ static void capEventRelayHops(meshtastic_MeshPacket *packet)
...
+{
+    switch (config.device.rebroadcast_mode) {
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_ALL:
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_ALL_SKIP_DECODING:
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_CORE_PORTNUMS_ONLY:
+        return true;
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_KNOWN_ONLY:
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_LOCAL_ONLY:
+        return p->channel == 0 && !isBroadcast(p->to) &&
+               (nodeInfoLiteHasUser(nodeDB->getMeshNode(p->from)) || nodeInfoLiteHasUser(nodeDB->getMeshNode(p->to)));
```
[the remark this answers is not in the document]
NomDeTom: Deliberately not changing the protobuf. The enum comments describe the decodable case; this PR adds no mode and no field, and the `KNOWN_ONLY` / `LOCAL_ONLY` rule for a PKI-shaped unicast is the identity check those modes already express, applied to the one identity the header carries in the clear (`to`) when the sender is inside the ciphertext. Added an "Explicitly out of scope" section to the PR body saying so.

test/support/AuthPipelineHarness.h
```diff
@@ -0,0 +1,533 @@
...
+    service = pipelineService = new MeshService();
+    mqtt = pipelineMqtt = new AuthPipelineMqtt();
+}
+
+/// Put back the AirTime and NodeStatus the harness replaced.
+static void pipelineHarnessDestroy()
+{
+    airTime = harnessSavedAirTime;
+    nodeStatus = harnessSavedNodeStatus;
+}
```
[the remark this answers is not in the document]
NomDeTom: Confirmed by CI: LSan reported a 160-byte direct leak from `pipelineHarnessCreate` line 458 (`new NodeStatus`) and errored both suites at exit. `pipelineHarnessDestroy()` now deletes the harness-owned `AirTime` and `NodeStatus` before restoring the saved pointers. Fix in the next push.

src/mesh/Router.cpp
```diff
@@ -1634,6 +1628,58 @@ void Router::dispatchReceived(meshtastic_MeshPacket *p, RxSource src)
     packetPool.release(p_encrypted); // Release the encrypted packet (release() handles nullptr)
 }
 
+/// A packet we had no way to read: PKI without the sender's key, or a channel we do not hold. A
+/// packet we matched and failed on (bad key, tampering, junk) is not one of these.
+static bool isUnreadableToUs(const meshtastic_MeshPacket *p)
+{
+    if (p->channel == 0) {
+        meshtastic_NodeInfoLite_public_key_t senderKey;
+        return !nodeDB->copyPublicKeyForDecrypt(p->from, senderKey);
```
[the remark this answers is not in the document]
NomDeTom: Agreed. `passesRoutingAuthGate()` now hands the attempt's `DecodeState` out, `handleOpaqueForUs()` takes `unreadable = (state == DECODE_OPAQUE)`, and `isUnreadableToUs()` is gone. To make `DECODE_OPAQUE` mean exactly "no key material to try", `perhapsDecode()` sets `pkiAttempted` only when a sender/pending/admin key was actually tried, and the `KNOWN_ONLY` short-circuit (which declines before any attempt) reports OPAQUE for a PKI-shaped unicast to us or an unheld hash, FAILURE for a held channel. Both of your cases now classify from the attempt: a hash-0 broadcast we cannot read is OPAQUE regardless of the sender's key, and a pending-key decrypt rejected as malformed is FAILURE (`NO_CHANNEL`, not to the phone). Fix in the next push.

src/mesh/Router.cpp
```diff
@@ -979,6 +988,7 @@ DecodeState perhapsDecode(meshtastic_MeshPacket *p)
             for (int i = 0; i < 3 && !decrypted; i++) {
                 if (config.security.admin_key[i].size != 32)
                     continue;
+                pkiAttempted = true;
```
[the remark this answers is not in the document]
NomDeTom: Fixed in d43900bb8: the admin-key loop no longer sets pkiAttempted, so only the sender's own or pending key counts as an attempt; a malformed admin decrypt still returns DECODE_FAILURE directly. C18 now configures an unrelated admin key and still expects PKI_UNKNOWN_PUBKEY and phone delivery in every mode (64618833f).

test/test_packet_signing/test_main.cpp
```diff
@@ -2110,27 +1881,122 @@ void test_E13_decoded_unsigned_nodeinfo_padded_inside_payload_dropped(void)
...
+}
+
+// C19: the same frame claiming to be from us is a forgery and gets nothing - no NAK, no phone,
+// no relay.
+void test_C19_undecryptable_frame_claiming_to_be_from_us_gets_no_reaction(void)
+{
+    const RelayIdentity target = makeIdentity(TARGET_NODE);
+    const RelayIdentity forger = makeIdentity(LOCAL_NODE);
+    meshtastic_MeshPacket p = makePkiUnicastBetween(forger, target, meshtastic_PortNum_TEXT_MESSAGE_APP, 0xADA60006, true);
+    p.to = LOCAL_NODE; // to us, "from" us
```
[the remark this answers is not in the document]
NomDeTom: Fixed in 64618833f: C19 installs our identity and DH key so the forged frame is a real decrypt attempt, and asserts passesRoutingAuthGate() == REJECT before the side-effect checks. Also added C23 (hash-0 broadcast from a keyed sender on an unheld channel is unreadable and reaches the phone) and C24 (KNOWN_ONLY withholds a stranger's frame on a held channel from the phone; the same stranger on an unheld channel is delivered).

test/test_packet_signing/test_main.cpp
```diff
@@ -2110,27 +1881,187 @@ void test_E13_decoded_unsigned_nodeinfo_padded_inside_payload_dropped(void)
...
+    }
+}
+
+// C19: the same frame claiming to be from us is a forgery and gets nothing - no NAK, no phone,
+// no relay. We hold our own identity, so the decrypt is attempted with our key and fails: the gate
+// must REJECT (from-us decode failure), not hand it to the opaque path.
+void test_C19_undecryptable_frame_claiming_to_be_from_us_gets_no_reaction(void)
+{
+    const RelayIdentity us = installOurIdentity();
+    useDHKey(us.priv);
```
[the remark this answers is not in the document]
NomDeTom: Fixed in 4f9f5d2f5, on the branch as of 43b938567.

Taken at the source rather than at the call site: `makePkiUnicastBetween()` now saves the engine's DH key on entry and restores it after the encrypt, so the caller's key survives and the ordering no longer matters for any caller. `CryptoEngine::private_key` is public under `PIO_UNIT_TESTING`, so this is a real save/restore, not a fresh key. The helper's doc comment states the contract, and the `generateKeyPair()` that existed only to leave a junk key behind is gone.

C19 also gained the assertion that was actually missing. You are right that `REJECT` alone did not pin the branch — but moving the key would not have pinned it either, since the forger's keypair is not ours and the decrypt fails on a key mismatch in both cases. So the test now builds the frame first, then installs our identity, and asserts a control: with our identity absent the same frame is `OPAQUE_RELAY_ONLY`. The `REJECT` is therefore attributable to the from-us decode-failure arm rather than to a mismatch present either way, and deleting that arm would now fail the test.

src/mesh/Router.cpp
```diff
@@ -1718,7 +1867,16 @@ void Router::perhapsHandleReceived(meshtastic_MeshPacket *p)
         // generate it here from the still-encrypted packet before opaque relay.
         if (isFromUs(p))
             perhapsGenerateImplicitAckForOwnOverheard(p);
-        relayOpaquePacket(p);
+        // One dedup for every consumer below, not just relay: three neighbours rebroadcasting a
+        // stranger's DM to us must still cost one NAK, one phone frame and one uplink.
+        const bool seen = p->id != 0 && opaqueWasSeenRecently(getFrom(p), p->id); // id 0 is the ring's empty slot
```
caveman99: `p->id != 0` short-circuits to `seen == false` for every id-0 frame, and nothing earlier on the receive path drops id 0 — I grepped `RadioLibInterface`, `RadioInterface` and the rest of `Router`, and there is no such guard.

That was harmless before this PR, because `relayOpaquePacket()` was the only consumer and it dropped `id == 0` outright. Now three more consumers act on the frame and none of them drops it: an undecryptable `want_ack` DM to us with `id == 0` gets a NAK **and** a phone frame on *every* copy, unbounded. So the "once per `(from,id)`" invariant the PR body states does not hold for `id == 0`.

The sentinel choice is fine — the cheap fix is to decline the frame at the other consumers too, since a NAK carrying `request_id 0` is unmatchable at the sender anyway:

```c++
// in handleOpaqueForUs() and uplinkOpaqueUnicast(), alongside the isFromUs / p->from == 0 guard
if (p->id == 0)
    return; // the ring cannot dedup id 0, and a NAK for it is unmatchable at the sender
```

---

src/mesh/Router.cpp
```diff
@@ -1656,6 +1660,151 @@ void Router::dispatchReceived(meshtastic_MeshPacket *p, RxSource src)
...
+    }
+}
+
+/// A PKI DM between two other nodes goes to MQTT as ciphertext when encrypted uplink is on, so a
+/// gateway carries traffic it cannot read. The uplink marks it pki_encrypted for the broker side.
+void Router::uplinkOpaqueUnicast(const meshtastic_MeshPacket *p, bool unreadable)
+{
+#if !MESHTASTIC_EXCLUDE_MQTT
+    // Only a frame we had no way to read: a failed decrypt on a channel we hold is not PKI ciphertext.
+    if (!unreadable || !mqtt || !moduleConfig.mqtt.enabled || !moduleConfig.mqtt.encryption_enabled || p->channel != 0 ||
```
caveman99, line 1802: This consumer is not gated by `rebroadcast_mode`, which contradicts the PR body: *"every rule the node applied to the packet before #10967 now applies to all four... That means duplicate suppression, `rebroadcast_mode` and the licensed-station rule; `encryption_enabled` (MQTT) and 'unreadable' (phone) stay local to their one consumer."*

`uplinkOpaqueUnicast()` applies neither `opaqueAllowedByMode()` nor the licensed-station check. The observable result: a gateway in `KNOWN_ONLY` / `LOCAL_ONLY` refuses to relay a stranger's opaque PKI DM (`relayOpaquePacket()` → `opaqueAllowedByMode()` says no) and withholds it from the phone (`handleOpaqueForUs()`'s `modeAllowsPhone`), but still republishes the ciphertext to a broker.

I checked the pre-#10967 behaviour and the code is faithful to it — the old uplink lived in `dispatchReceived()`, outside `RoutingModule::handleReceivedProtobuf()`'s mode filter — so this may well be deliberate. But then two things need correcting rather than the code:

- the PR body's "applies to all four" claim, and
- the comment in `handleOpaqueForUs()`: *"The mode rule that gates relay gates this too - LOCAL_ONLY / KNOWN_ONLY mean 'ignore what I cannot read', not just 'do not carry it'."* That reasoning applies at least as strongly to publishing the frame to a broker as it does to the phone.

Given that the value of this PR is a precise, stated rule table, worth picking one explicitly.

---

src/mesh/Router.h
```diff
@@ -173,6 +186,27 @@ class Router : protected concurrency::OSThread, protected PacketHistory
     void sendAckNak(meshtastic_Routing_Error err, NodeNum to, PacketId idFrom, ChannelIndex chIndex, uint8_t hopLimit = 0,
                     bool ackWantsAck = false, const meshtastic_MeshPacket *relaySource = nullptr);
 
+    static constexpr uint8_t OPAQUE_SEEN_MAX = 32; // opaque dedup slots (see relayOpaquePacket); ~8B/slot -> ~256B
```
caveman99, line 189: The PR notes flag that the ring now records every opaque frame the node hears, and suggest raising this if a dense deployment shows re-relays. Worth pushing on, because some of the widening buys nothing at all.

`opaqueWasSeenRecently()` is now called at the branch head, before any eligibility gate, so slots are consumed by frames no consumer can act on. The clearest case is `isFromUs(p)`: the only thing that runs for those is `perhapsGenerateImplicitAckForOwnOverheard()`, which sits *above* the ring and never consults it, so every overheard rebroadcast of our own PKI DM evicts a live entry for nothing. A frame with a foreign `next_hop` that is neither for us, nor a broadcast, nor a channel-0 unicast is the same story.

Those evictions come straight out of the anti-amplification property #11522 added the ring for. Skipping the frames that cannot matter keeps the shared dedup the four consumers need while recovering the capacity:

```c++
const bool couldMatter = !isFromUs(p) && (isToUs(p) || isBroadcast(p->to) || p->hop_limit > 0 || p->channel == 0);
const bool seen = p->id != 0 && couldMatter && opaqueWasSeenRecently(getFrom(p), p->id);
```

(`p->channel == 0` keeps a hop-exhausted unicast in the ring, since `uplinkOpaqueUnicast()` still acts on one.)

Raising the slot count would also help on its own — 32 entries at ~8B is cheap, and on a node that hears several channels it does not hold, the churn rate is high enough that three neighbours' rebroadcasts of one frame can land more than 32 opaque frames apart.

---

src/mesh/Router.cpp
```diff
@@ -1703,10 +1852,10 @@ void Router::perhapsHandleReceived(meshtastic_MeshPacket *p)
         return;
     }
 
-    // Decrypt and authenticate before Reliable/Flooding/NextHop filters can update retry
-    // timers, packet history, implicit ACK state, cancellation, or relay queues. A packet for
-    // an unknown channel passes as opaque traffic and retains the existing relay behavior.
-    const auto authVerdict = passesRoutingAuthGate(p);
+    // Decrypt and authenticate before Reliable/Flooding/NextHop filters can update retry timers,
+    // history, ACK state or relay queues. An unreadable packet touches no local state at all.
+    DecodeState gateState;
```
caveman99: Uninitialized. It is safe today only because `passesRoutingAuthGate()` assigns `*decodeState = DECODE_SUCCESS` as its first statement — a non-local guarantee in a function with five early returns, and the parameter is optional (`= nullptr`), so nothing forces that invariant to be kept.

```c++
DecodeState gateState = DecodeState::DECODE_SUCCESS;
```

Costs nothing and makes `unreadable` well-defined regardless of what happens inside the gate.

---

src/mesh/Router.cpp
```diff
@@ -1656,6 +1660,151 @@ void Router::dispatchReceived(meshtastic_MeshPacket *p, RxSource src)
...
+        return;
+    // A licensed station transmits in the clear and may not answer, or hand on, traffic to or from a
+    // node it knows to be unlicensed. Same rule RoutingModule applies to the decoded path.
+    if (owner.is_licensed && (nodeDB->getLicenseStatus(p->from) == UserLicenseStatus::NotLicensed ||
+                              nodeDB->getLicenseStatus(p->to) == UserLicenseStatus::NotLicensed))
+        return;
+    if (isToUs(p) && p->want_ack && !isBroadcast(p->to) && routingModule) {
+        // PKI_UNKNOWN_PUBKEY only if the frame could have been PKI at all: the same size test
+        // perhapsDecode() uses to pick a candidate. Anything shorter never had a key to miss.
+        const bool pkiShaped = p->channel == 0 && unreadable && p->encrypted.size > MESHTASTIC_PKC_OVERHEAD;
```
caveman99: Minor, and a degenerate config, but the reason code can be misleading in one direction the legacy path got right.

`pkiCandidate` in `perhapsDecode()` requires `ourNode->public_key.size > 0`. A node with no keypair of its own therefore never attempts PKI at all, falls through to the channel loop, and comes out `DECODE_OPAQUE` — so `pkiShaped` is true and we answer `PKI_UNKNOWN_PUBKEY`, asking the sender for a NodeInfo we still could not act on.

The arm this replaces keyed the reason off the *sender's* key instead (`ReliableRouter.cpp:144`: `nodeDB->getMeshNode(p->from) == nullptr || ...->public_key.size == 0`), so it said `NO_CHANNEL` in that case. If you want to keep the distinction, `pkiShaped` could also require that we hold a key — otherwise it is worth a line in the comment saying the reason reports "no usable key pair for this DM", either end, rather than "I don't know yours".

---

src/mesh/Router.cpp
```diff
@@ -928,6 +932,10 @@ DecodeState perhapsDecode(meshtastic_MeshPacket *p)
     if (config.device.rebroadcast_mode == meshtastic_Config_DeviceConfig_RebroadcastMode_KNOWN_ONLY &&
         !nodeInfoLiteHasUser(nodeDB->getMeshNode(p->from))) {
         LOG_DEBUG("Node 0x%08x not in nodeDB, Rebroadcast KNOWN_ONLY ignores packet", p->from);
+        // Declined before any attempt: a stranger's PKI DM has no key to try, a held channel does.
+        if (p->which_payload_variant == meshtastic_MeshPacket_encrypted_tag &&
+            ((p->channel == 0 && isToUs(p) && !isBroadcast(p->to)) || !channels.hasHash(p->channel)))
```
caveman99: The `isToUs(p)` in the first clause narrows this more than the comment suggests. A stranger's PKI DM **to a third party** is a channel-0 unicast with no key to try, but it only reaches `DECODE_OPAQUE` through the second clause, `!channels.hasHash(p->channel)` — so on a node that happens to hold a channel hashing to 0 it lands on `DECODE_FAILURE` instead.

Relay is unaffected (`opaqueAllowedByMode()` still carries it), and the phone is not involved for a frame that is not for us, so the only visible consequence is that `uplinkOpaqueUnicast()` stops uplinking it — `unreadable` is false. Narrow, but it is the same "a held hash-0 channel must not make a stranger's PKI DM look like channel traffic" hazard that C29 pins for the to-us case, just on the relay-side DM.

Dropping `isToUs(p) &&` from the first clause would cover both, and reads closer to the comment:

```c++
if (p->which_payload_variant == meshtastic_MeshPacket_encrypted_tag &&
    ((p->channel == 0 && !isBroadcast(p->to)) || !channels.hasHash(p->channel)))
    return DecodeState::DECODE_OPAQUE;
```

---

src/mesh/Router.h
```diff
@@ -264,8 +298,15 @@ enum class RoutingAuthVerdict { ACCEPT, OPAQUE_RELAY_ONLY, REJECT };
  */
 DecodeState perhapsDecode(meshtastic_MeshPacket *p);
 
-/** Apply receive authentication before routing state mutation; unknown-channel packets may remain opaque relay-only. */
-RoutingAuthVerdict passesRoutingAuthGate(meshtastic_MeshPacket *p);
+#if USERPREFS_EVENT_MODE
+/** Cap a relay copy's hop budget to the event-mode limit, keeping hop_start consistent. */
+void capEventRelayHops(meshtastic_MeshPacket *packet);
```
caveman99, line 302: Nit. This was `static` in NextHopRouter.cpp and is now a free function at global scope, so the name is visible to every TU that includes `Router.h`. No ODR problem — there is exactly one definition — but `capEventRelayHops` is generic enough to be worth scoping, e.g. a `static` member of `Router` alongside `isRebroadcaster()`, since both relay paths that call it are `Router` subclasses.

---

caveman99 (commented):
Read the full source diff against the surrounding code. This is a well-argued, correctly-scoped regression fix — the diagnosis is right, and moving the whole opaque path out of `NextHopRouter` into `Router` is the right structural call, since nothing in it was next-hop specific and the `(from,id)` ring belongs beside `PacketHistory`.

Two things I checked specifically that hold up:

- `pkiAttempted = haveRemoteKey` genuinely makes `DECODE_OPAQUE` mean "no key material was tried". The admin-key loop no longer poisons the classification, which is what the earlier review thread was after.
- Removing the `p_encrypted->pki_encrypted` marking from `dispatchReceived()` is correct dead-code removal, not a behaviour change: that block has been unreachable since #10967 put `DECODE_OPAQUE` behind an early return in `perhapsHandleReceived()`. `uplinkOpaqueUnicast()` is what actually restores the uplink. Worth stating in the PR body, because on a skim it reads like the uplink simply moved.

Tests are strong — the harness drives a real `ReliableRouter` through `perhapsHandleReceived()` rather than re-implementing the decision, no case was dropped in the refactor (78 → 91, plus the new 10-case suite), and the three renames tighten rather than weaken what they pin.

Seven comments below. None is a blocker: two are one-line fixes, one needs a decision from you rather than code.

- **`id == 0` bypasses the dedup for the three new consumers.** The invariant the PR body states does not hold for it.
- **MQTT uplink is not mode-gated**, which contradicts the "applies to all four" claim in the body. The code matches pre-#10967, so this may be a body/comment fix rather than a code fix — but the two currently disagree.
- **The ring's population widened without the ring growing**, and part of that widening (notably `isFromUs` frames) buys nothing.
- `DecodeState gateState;` uninitialized, safe only by a non-local guarantee.
- `PKI_UNKNOWN_PUBKEY` when *we* are the node without a keypair.
- A narrow gap in the `KNOWN_ONLY` short-circuit for a stranger's PKI DM to a third party.
- A naming nit on `capEventRelayHops`.

One note on verification: I could not run `test_packet_signing` / `test_rebroadcast_mode` locally — my sandbox blocks the GitHub archive downloads for the lib deps — so the green I am relying on is CI's `test-native` job on this head, not a local run.

---

src/mesh/Router.cpp
NomDeTom: Fixed in fd344649b. `handleOpaqueForUs()` and `uplinkOpaqueUnicast()` both decline `id == 0`, as `relayOpaquePacket()` already did, so the sentinel keeps its meaning in the ring.

`test_C31_an_opaque_frame_with_id_0_is_answered_by_nobody` feeds the same undecryptable `want_ack` DM three times and asserts no NAK, no phone frame, no uplink and no relay.

src/mesh/Router.cpp
NomDeTom: Picked explicitly in fd344649b, and split rather than taken whole: the licensed-station check now applies to `uplinkOpaqueUnicast()`, `rebroadcast_mode` deliberately does not. The two lines at the function say why — the mode governs what this node puts back on the air, and MQTT has its own switches for what leaves over IP, which is where the pre-#10967 uplink sat.

Both of the corrections you asked for are in: the PR body's "applies to all four" sentence is now a per-consumer statement, and the `handleOpaqueForUs()` comment no longer claims the mode rule reaches the broker.

`test_C32_uplink_follows_mqtt_policy_not_rebroadcast_mode` pins both halves: a `KNOWN_ONLY` gateway uplinks the stranger's DM it refuses to relay, and a licensed one uplinks nothing.

src/mesh/Router.h
NomDeTom: Filter taken as written in fd344649b, `couldMatter` and the channel-0 clause included.

`test_C33_frames_no_consumer_acts_on_do_not_evict_the_ring` relays a stranger's DM, feeds 40 of our own overheard rebroadcasts through the branch head, and asserts the stranger's second copy is still deduped.

Slot count left alone: `OPAQUE_SEEN_MAX` is already 32 (~256B, per the header comment), so I read the raise as the value the code has. Say if you meant a number above that and I will raise it.

src/mesh/Router.cpp
NomDeTom: Fixed in fd344649b — initialised to `DECODE_SUCCESS` at the declaration.

src/mesh/Router.cpp
NomDeTom: Taken as the comment option in fd344649b. `pkiShaped` keeps the frame-shape test only, and the comment now states that on a node holding no keypair the reason reads as "no usable key pair for this DM", either end, rather than "I do not know yours".

src/mesh/Router.cpp
NomDeTom: Fixed in fd344649b, though by a different cut than dropping `isToUs(p) &&`. The main path reaches `DECODE_FAILURE` for that same third-party DM through `matchedChannel`, so removing the clause on its own would have left the short-circuit and the main path disagreeing.

`isPkiShapedUnicast()` (channel 0, not a broadcast, longer than `MESHTASTIC_PKC_OVERHEAD`) now decides it in one place, used by the `KNOWN_ONLY` short-circuit, the provenance line — a hash-0 channel match is no evidence about such a frame — and the NAK reason. The `isToUs` asymmetry is gone with it.

One more thing fell out of it: on a node holding a hash-0 channel, our own overheard PKI DMs were classified `DECODE_FAILURE` and hit the `isFromUs` reject arm, so the implicit "Delivered to mesh" ACK never fired there. That path works now.

C30 flips as a consequence: it asserts the relayed DM is carried, and that a channel-0 unicast too short to have been PKI still counts as channel traffic we tried. The trade is in the PR body's AEAD section — happy to put the old row back if you would rather keep it.

src/mesh/Router.h
NomDeTom: Noted, and not addressed in fd344649b — `capEventRelayHops` is still a free function at global scope. I can scope it as a `static` member of `Router` alongside `isRebroadcaster()` in the next push if you want it in this PR.

NomDeTom:
@caveman99 I've expanded the flash guard by 2k to get this in. If it passes CI, can you merge it, plz?

src/mesh/Router.cpp
```diff
@@ -843,10 +847,10 @@ RoutingAuthVerdict passesRoutingAuthGate(meshtastic_MeshPacket *p)
         return RoutingAuthVerdict::REJECT;
     }
     if (state == DecodeState::DECODE_FAILURE) {
-        // One-byte hash collisions are indistinguishable from tampering, so relay opaquely
-        // instead of blackholing; isFromUs stays REJECT to keep forged senders off the ACK path.
-        if (!isToUs(p) && !isFromUs(p)) {
-            LOG_WARN("Decryptable packet failed decoding, relay opaquely");
+        // A hash collision is indistinguishable from tampering, so treat it as opaque: relayed if not for
+        // us, NAKed and shown to the phone if it is. isFromUs stays REJECT to keep forgeries off the ACK path.
+        if (!isFromUs(p)) {
```
caveman99: **S1 — this line opens the entire to-us reply surface, and the comment it replaced said so.**

```diff
-        // isFromUs stays REJECT to keep forged senders off the ACK path.
-        if (!isToUs(p) && !isFromUs(p)) {
+        if (!isFromUs(p)) {
```

`DECODE_FAILURE` is specifically the state meaning *we had the key material and it failed* — a wrong-PSK frame, or an **AEAD authentication failure**. Before this change a to-us `DECODE_FAILURE` was `REJECT`: dropped in silence. Now it reaches `handleOpaqueForUs()` and draws a `NO_CHANNEL` NAK if `want_ack` is set, which is an attacker-controlled header bit.

So the node transmits a reply to data that cryptographically failed authentication. The AEAD section of the PR body defends this as "what a tampered or forged frame on our own channel should get" — I'd argue silence is what it should get. Answering gives a presence/liveness oracle and a reflection vector to anyone who can flip bits in a captured frame, and AEAD has already told us the frame is not what it claims.

This is also the root of the whole to-us surface, not just the NAK: it is what admits to-us frames into `handleOpaqueForUs()` at all, and therefore into the phone-delivery and admin-key-budget paths below.

Suggested: revert to `!isToUs(p) && !isFromUs(p)`, or gate the to-us arm on `unreadable`. The "no key material to try" recovery path — the one this PR actually needs for `PKI_UNKNOWN_PUBKEY` → NodeInfo — arrives as `DECODE_OPAQUE` and is unaffected either way.

---

src/mesh/Router.cpp
```diff
@@ -1656,6 +1670,150 @@ void Router::dispatchReceived(meshtastic_MeshPacket *p, RxSource src)
...
+        return;
+    if (isToUs(p) && p->want_ack && !isBroadcast(p->to) && routingModule) {
+        // PKI_UNKNOWN_PUBKEY only if the frame could have been PKI at all. With no keypair of our own it
+        // reads as "no usable key pair for this DM", either end, rather than "I do not know yours".
+        const bool pkiShaped = unreadable && isPkiShapedUnicast(p);
+        const auto err = pkiShaped ? meshtastic_Routing_Error_PKI_UNKNOWN_PUBKEY : meshtastic_Routing_Error_NO_CHANNEL;
+        LOG_INFO("Cannot decrypt 0x%08x from 0x%08x, NAK %d", p->id, p->from, (int)err);
+        // A repeat is the sender's own retransmission, so it is a direct neighbour: answer at hop 0,
+        // as the decoded path re-ACKs, and stop there - the phone already has this frame.
+        sendAckNak(err, getFrom(p), p->id, channels.getPrimaryIndex(), repeat ? 0 : routingModule->getHopLimitForResponse(*p));
```
caveman99: **S2 — the response hop budget is chosen by the attacker, and nothing throttles the rate.**

Every input this NAK decision uses is unauthenticated cleartext. `PacketHeader` (`RadioInterface.h:35`) carries `to`, `from`, `id`, `channel`, and in `flags` the `want_ack` bit (0x08), `hop_limit` (0x07) and `hop_start` (0xE0).

**Hop budget.** `getHopsAway()` returns `hop_start - hop_limit` (`NodeDB.cpp:3497`), and `getHopLimitForResponse()` does:

```c++
if (hopsUsed > responseHopLimit) {
#if !USERPREFS_EVENT_MODE
    return hopsUsed;   // "If the request used more hops than the limit, use the same amount"
#endif
```

So `hop_start=7, hop_limit=0` makes a node configured for 3 hops emit a **7-hop** NAK, and test C28 confirms a frame with no hops left is still answered. One ~30-byte frame in, one routing packet flooded seven hops out, aimed at whatever `from` says. The lever exists on the decoded path too, but there an attacker must first produce a decodable packet; here they need no key at all.

**Rate.** The dedup ring keys on `(from,id)`, both attacker-chosen — vary `id` and every frame is a fresh NAK. Worth noting what used to sit in front of this: pre-#10967 the undecodable NAK lived in `ReliableRouter::sniffReceived()`, reached via `RoutingModule` inside module dispatch, behind `PacketHistory` *and* behind `TrafficManagementModule::shouldDropUnknown()`, which on exceeding `unknown_packet_threshold` does:

```c++
ignoreRequest = true;        // Suppress NAK for want_response packets
return ProcessMessage::STOP; // Consumed - will not be rebroadcast
```

TMM is constructed at `Modules.cpp:144` and `RoutingModule` at `:299`, and registration is push-back order, so TMM ran first and `STOP` broke the loop before the NAK. That comment is a purpose-built undecodable-NAK suppressor. Emitting from `perhapsHandleReceived()` puts the NAK ahead of both. (`unknown_packet_threshold` defaults to 0, so it is opt-in — but the operators who enable it are exactly the ones running exposed meshes.)

Two suggestions, both small:
1. Clamp to `Default::getConfiguredOrDefaultHopLimit(config.lora.hop_limit)` instead of `getHopLimitForResponse()`. A genuinely unreachable sender is inside the configured radius or is not coming back anyway.
2. Add a global emission floor, reusing the pattern already in the tree at `TrafficManagementModule.h:373` — *"Direct-response throttles bounding the reflector risk of spoofed replies"*, `kDirectResponseGlobalMs = 1'000UL`. A wall-clock floor does not care what `id` says, and one NAK per second is ample for real recovery.

**Separately, the reason code is a key-possession oracle.** `pkiAttempted = haveRemoteKey`, so for a PKI-shaped `want_ack` unicast to us claiming `from = X`: we hold X's key → `DECODE_FAILURE` → `NO_CHANNEL`; we don't → `DECODE_OPAQUE` → `PKI_UNKNOWN_PUBKEY`. One unauthenticated frame per query, answered over the air, for any X — that enumerates a node's NodeDB key-holdings, i.e. its social graph. The codebase states its position on this pattern at `MeshModule.cpp:219`:

> `// SECURITY NOTE! I considered sending back a different error code if we didn't find the psk (i.e. !isDecoded) but opted NOT TO. Because it is not a good idea to let remote nodes 'probe' to find out which PSKs were "good" vs bad.`

Same principle, different secret. The oracle existed pre-#10967 too (`ReliableRouter.cpp:144`), and this version is marginally sharper since `copyPublicKeyForDecrypt()` also counts pending handshake keys. I don't think the distinction can be dropped — `PKI_UNKNOWN_PUBKEY` is what triggers the NodeInfo this PR exists to restore — but it belongs in the PR body as an accepted trade rather than left implicit.

---

src/mesh/Router.cpp
```diff
@@ -1656,6 +1670,150 @@ void Router::dispatchReceived(meshtastic_MeshPacket *p, RxSource src)
...
+
+/// rebroadcast_mode for a packet we cannot read. The port list and sender identity are inside the
+/// ciphertext, so CORE_PORTNUMS_ONLY relays; KNOWN/LOCAL relay a PKI-shaped unicast with one known party
+/// (`to` is us for a DM to us).
+bool Router::opaqueAllowedByMode(const meshtastic_MeshPacket *p)
+{
+    switch (config.device.rebroadcast_mode) {
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_ALL:
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_ALL_SKIP_DECODING:
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_CORE_PORTNUMS_ONLY:
```
caveman99: **S3 — correct fix, but note what it does to the size of the relay surface.**

`installRoleDefaults()` (`NodeDB.cpp:1673`) sets `ROUTER` → `CORE_PORTNUMS_ONLY`. Adding that case here is exactly the fix #11843 needs, and I am not asking for it to change. But it moves the population that will flood unauthenticated ciphertext from "nodes explicitly configured to `ALL`" to **the whole high-power, well-sited backbone**.

`relayOpaquePacket()` has no broadcast restriction, so one injected frame with an arbitrary `channel` hash and `hop_limit=7` is flooded mesh-wide by every ROUTER in range.

On what bounds it, being precise about the dedup ring since it is easy to overclaim: varying `id` defeats the 32-slot ring *and* `PacketHistory` equally, since both key on `(from,id)` — so neither bounds injection rate. Where capacity matters is suppressing the multiplicative storm from a single frame echoing through a dense mesh, and there the opaque path runs on 32 untimed slots shared across four consumers while the decoded path (and the pre-#10967 opaque path) get `PACKETHISTORY_MAX = max(MAX_NUM_NODES*2, 100)` timestamped entries (`mesh-pb-constants.h:135`) — typically 200+.

The real backstops are `MAX_TX_QUEUE 16` (`RadioInterface.h:18`) and the duty-cycle abort in `Router::send()`, and the latter only bites in duty-cycle regions. Worth a sentence in the PR body so operators sizing dense deployments know the storm bound on this path is weaker than on the decoded one.

---

src/mesh/Router.cpp
```diff
@@ -1656,6 +1670,150 @@ void Router::dispatchReceived(meshtastic_MeshPacket *p, RxSource src)
...
+        return true;
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_KNOWN_ONLY:
+    case meshtastic_Config_DeviceConfig_RebroadcastMode_LOCAL_ONLY:
+        return p->channel == 0 && !isBroadcast(p->to) &&
+               (nodeInfoLiteHasUser(nodeDB->getMeshNode(p->from)) || nodeInfoLiteHasUser(nodeDB->getMeshNode(p->to)));
```
caveman99: **S4 — the `KNOWN_ONLY` / `LOCAL_ONLY` identity check runs on a field the stranger fills in.**

`p->to` is unauthenticated cleartext in `PacketHeader`. An attacker who wants a `KNOWN_ONLY` node to carry arbitrary ciphertext sets `to` to any node that node knows — learnable by listening, since `to` is in the clear on every frame that passes. So the "one party in NodeDB" rule is satisfiable by a one-field edit.

The `from` side is at least arguably defensible: the real sender identity is inside the ciphertext, and the PR body makes that argument. `to` has no such defence — it is definitionally cleartext, and it is the half an attacker gets to choose freely.

This is the same ground Copilot covered as a protobuf-documentation question, and the reply treated it as one. The security framing is different: these two modes exist so an operator can say *"do not carry traffic from strangers,"* and the new rule lets a stranger assert their way past it. I don't think that sinks the design — the alternative is dropping PKI traffic in those modes, which is the regression being fixed — but it should be stated in the PR body rather than left to the enum-comment debate, because an operator choosing `KNOWN_ONLY` for hardening reasons is getting less than the name suggests.

---

src/mesh/Router.cpp
```diff
@@ -1656,6 +1670,150 @@ void Router::dispatchReceived(meshtastic_MeshPacket *p, RxSource src)
...
+    if (repeat)
+        return;
+    // Straight to the phone queue: handleFromRadio() would updateFrom() NodeDB for an unverified sender.
+    // The relay mode gates this too; NONE is "do not relay", not "do not listen". MQTT gates itself.
+    const bool modeAllowsPhone =
+        config.device.rebroadcast_mode == meshtastic_Config_DeviceConfig_RebroadcastMode_NONE || opaqueAllowedByMode(p);
+    if (unreadable && modeAllowsPhone && (isToUs(p) || isBroadcast(p->to)) && service) {
+        if (meshtastic_MeshPacket *toPhone = packetPool.allocCopy(*p)) {
+            stampRxTime(toPhone);
+            service->sendToPhone(toPhone);
```
caveman99: **S5 — this halves the admin-key fallback budget, which the code already documents as attacker-facing.**

`MeshService::sendToPhone()` re-runs `perhapsDecode(p)` on arrival (`MeshService.cpp:484`), on the very copy this line just handed it. For a PKI-shaped undecryptable DM to us that re-enters the fallback:

```c++
if (!decrypted && adminKeyFallbackAllowed()) {   // token consumed HERE
    for (int i = 0; i < 3 && !decrypted; i++) {
        if (config.security.admin_key[i].size != 32)
            continue;
```

The token is spent *before* the key-configured check, so it burns even on nodes with no admin keys at all. Budget is `ADMIN_KEY_FALLBACK_BURST 8` refilling 1 per 250 ms → 4/sec sustained. Before this PR these frames returned at the gate and cost one token; now they cost two, so **~2 frames/sec drains the bucket instead of 4**. Once empty, `adminKeyFallbackAllowed()` returns false and a legitimate admin whose key the node has not yet learned can no longer reach it.

The budget's own comment is explicit about the threat model:

> `// The fallback costs three X25519 ops before the AEAD tag is checked. Budget is global because p->from is attacker-controlled`

So this is a straightforward regression against a defence the codebase already reasoned about. Cheapest fix is to not re-decode a frame the gate has already classified — either skip `perhapsDecode()` in `sendToPhone()` when the packet arrives already marked, or have `handleOpaqueForUs()` stamp the copy so the second call short-circuits.

Two things I checked here that are *not* problems, for the record: the phone queue refuses an encrypted frame rather than displacing a queued one when full (`MeshService.cpp:509` gates displacement on `decoded_tag` and three portnums), so there is no starvation primitive against the phone; and there is no crypto amplification beyond the token accounting above, since X25519 runs only when we hold the sender's key.

---

src/mesh/Router.cpp
```diff
@@ -1656,6 +1670,150 @@ void Router::dispatchReceived(meshtastic_MeshPacket *p, RxSource src)
...
+    // Only a frame we had no way to read: a failed decrypt on a channel we hold is not PKI ciphertext.
+    if (!unreadable || !mqtt || !moduleConfig.mqtt.enabled || !moduleConfig.mqtt.encryption_enabled || p->channel != 0 ||
+        p->id == 0 || isBroadcast(p->to) || isToUs(p) || isFromUs(p))
+        return;
+    if (owner.is_licensed && (nodeDB->getLicenseStatus(p->from) == UserLicenseStatus::NotLicensed ||
+                              nodeDB->getLicenseStatus(p->to) == UserLicenseStatus::NotLicensed))
+        return;
+    meshtastic_MeshPacket copy = *p;
+    copy.pki_encrypted = true;
+    mqtt->onSend(copy, copy, p->channel);
```
caveman99: **S6 — the MQTT queue evicts, so this is a way to flush a gateway's backlog.**

`MAX_MQTT_QUEUE` is 16 (`MQTT.h:24`), and when the gateway is not connected `MQTT::onSend()` discards the oldest to make room:

```c++
if (mqttQueue.numFree() == 0) {
    LOG_WARN("MQTT queue is full, discard oldest");
    entry = mqttQueue.dequeuePtr(0);
```

Unlike the phone queue — which refuses an encrypted frame rather than displacing one, and is the right behaviour — this evicts. So an attacker in radio range of a gateway that is momentarily offline can **flush its entire queued legitimate backlog with 16 injected frames**, each costing one unauthenticated transmission and no key material. Rate-limited only by the `(from,id)` ring, which they defeat by varying `id`.

Also worth stating in the body: every MQTT content policy — `DontMqttMeBro`, the public-server `RANGE_TEST`/`DETECTION_SENSOR` filter — sits inside `if (mp_decoded.which_payload_variant == decoded_tag)` in `onSend()`, so none of them applies to an opaque frame. That is inherent, since the bitfield is inside the ciphertext, but the consequence is that a node's "do not uplink me" preference is unenforceable for exactly the traffic this path carries.

I am not asking for the uplink to be mode-gated — your split between RF policy and IP policy is a reasonable line and the body now argues it properly. The eviction property is the part worth knowing about.

---

src/mesh/Router.cpp
```diff
@@ -1718,7 +1876,18 @@ void Router::perhapsHandleReceived(meshtastic_MeshPacket *p)
         // generate it here from the still-encrypted packet before opaque relay.
         if (isFromUs(p))
             perhapsGenerateImplicitAckForOwnOverheard(p);
```
caveman99: **S7 — the one consumer with no dedup at all, and it changes local delivery state.**

`ReliableRouter::perhapsGenerateImplicitAckForOwnOverheard()` gates on `p->from == getNodeNum()` — spoofable cleartext — and on a pending-id match does two things:

```c++
sendAckNak(meshtastic_Routing_Error_NONE, getFrom(p), p->id, old->packet->channel, 0, false, p);
if (p->transport_mechanism == ...TRANSPORT_LORA)
    stopRetransmission(key);
```

A node adjacent to a victim can absorb the victim's DM, replay its cleartext header with junk payload, and the victim's phone shows "Delivered to mesh" while retransmission stops and the message never arrives. No key material needed; the `id` is simply read off the air.

Being fair about provenance: this is pre-existing and inherent to a header-only implicit ACK — it is the same mechanism a legitimate relay triggers — and it is one-shot per pending packet, since `stopRetransmission()` removes the entry and later calls find nothing. So it is not an amplifier.

Two things this PR does touch, though:

- **`fd34464` widens reachability.** On a node holding a hash-0 channel these frames used to classify `DECODE_FAILURE` → the `isFromUs` REJECT arm, so the path was accidentally closed there. You correctly identified that as a bug and fixed it; I am noting the security direction of the fix, not disputing it.
- **It is now the only consumer with zero dedup**, because `couldMatter` excludes `isFromUs` frames from the ring. That exclusion was my suggestion, on the grounds that this call sits above the ring and never consults it. The reasoning holds and I would not reverse it — but I should have flagged that it leaves this path unthrottled, so recording that here.

No change requested. Worth a line in the body alongside the other opaque-path caveats, since it is the one place where unauthenticated input alters local delivery state rather than just costing airtime.

---

caveman99 (commented):
x

caveman99:
## Correction to my security review: S3–S6 are asks, not notes

In the review above I said S3–S7 belonged in the PR body "for whoever reads it next". That was wrong twice over. The maintainers reading this PR *are* the decision, not a downstream audience, and "documented" is not a mitigation for something a rogue node can do on day one. Retracting that framing and re-testing all five against the only baseline that matters for a merge decision: **surface relative to `develop` as it stands**, not relative to pre-#10967.

| | Increases surface vs `develop`? | Ask |
|---|---|---|
| **S1** authentication-failed frames get answered | Yes | Fix in this PR |
| **S3** opaque relay on every `ROUTER` | Yes — inherent to the fix | Bound it in this PR |
| **S4** `KNOWN_ONLY`/`LOCAL_ONLY` keyed on cleartext `to` | Yes, trivially exploitable | Fix in this PR |
| **S5** admin-key budget halved | Yes | Fix in this PR |
| **S6** MQTT queue eviction | Yes | Fix in this PR |
| **S7** implicit-ACK delivery suppression | **No** | Separate issue, no change here |

---

### S4 — the one I most want changed, and it is surgical

I checked #11843 against this. The reporter's setup is an admin node in `CLIENT_MUTE`, a tree node in `ROUTER_LATE` set to core-portnums-only, and a target in `ROUTER`; switching that node to `ALL` fixed it. The bug is **entirely** `CORE_PORTNUMS_ONLY`. `KNOWN_ONLY` / `LOCAL_ONLY` are not in it.

So the `KNOWN_ONLY`/`LOCAL_ONLY` arm of `opaqueAllowedByMode()` is scope the bug did not ask for, and it is the arm whose predicate a stranger can satisfy by editing one cleartext byte-range. Neither half authenticates — `to` is chosen freely, and `from` can equally be set to a node the target knows.

Surgical fix that keeps everything the PR wants:

```c++
    case meshtastic_Config_DeviceConfig_RebroadcastMode_KNOWN_ONLY:
    case meshtastic_Config_DeviceConfig_RebroadcastMode_LOCAL_ONLY:
        return isToUs(p) && p->channel == 0 && !isBroadcast(p->to);
```

Because `relayOpaquePacket()` already rejects `isToUs(p)`, this means **nothing opaque relays in those two modes** — matching `develop`, and matching what your own findings table records as `develop`'s behaviour. But `handleOpaqueForUs()`'s `modeAllowsPhone` still returns true for a PKI DM addressed to us, so phone delivery in those modes is preserved exactly as the PR intends.

Test cost is one inversion: `test_known_destination_satisfies_known_only_and_local_only` becomes "KNOWN_ONLY / LOCAL_ONLY decline a stranger's DM whoever it is addressed to". `test_pki_unicast_between_strangers_relays_unless_mode_needs_identity` already expects no relay in those modes and is unaffected, and `test_phone_delivery_of_unreadable_frames_follows_the_relay_mode_table` still passes on the "a PKI DM to us is always one known party" leg. Worth re-running C24 to confirm.

This also dissolves the protobuf objection Copilot raised, without a protobuf change: `KNOWN_ONLY` stops carrying traffic from nodenums it does not know, which is what the enum comment says it does.

### S6 — unauthenticated traffic must not evict authenticated traffic

`MAX_MQTT_QUEUE` is 16 and `onSend()` discards the oldest when full, so 16 injected frames flush a disconnected gateway's legitimate backlog. Your phone path already has the right property — `MeshService.cpp:509` refuses an encrypted frame rather than displacing a queued one — and the uplink should match it. `mqttQueue` is `protected`, so the minimal change is an accessor plus an early return in `uplinkOpaqueUnicast()`:

```c++
// MQTT.h, public:
bool queueHasRoom() { return mqttQueue.numFree() > 0; }
```

```c++
// uplinkOpaqueUnicast(), with the other gates
if (!mqtt->queueHasRoom())
    return; // never evict authenticated traffic to carry a frame we cannot read
```

`AuthPipelineMqtt` already reaches `mqttQueue` from the protected section, so C21/C32 can pin it directly.

### S3 — and a reversal of my earlier advice

The security statement here is that `CORE_PORTNUMS_ONLY` exists so a ROUTER carries only core portnums, and an opaque frame bypasses that filter *by construction* — you cannot classify what you cannot read. That is inherent to fixing #11843 and I am not asking for it back. What is not inherent is how weakly the volume is bounded.

**I was wrong to tell you to leave `OPAQUE_SEEN_MAX` at 32.** I gave that answer before working through the fact that this PR moves the relaying population from "nodes explicitly set to `ALL`" to the entire backbone. More relayers means more opaque frames in flight per node, means faster churn through 32 untimed slots shared across four consumers, means the duplicate-suppression bound degrades exactly where the storm risk is highest. Please size it the way `PACKETHISTORY_MAX` already is in `mesh-pb-constants.h:135` — small on constrained STM32, 128 or so elsewhere. It is `.bss`, so it costs RAM and not flash: rak4631 is at 98,892 of 108,000 RAM, and 128 slots is +768 bytes. That keeps it clear of the flash budget question, which is separate and still open.

**What I am deliberately not asking for:** a global opaque-relay rate cap. A busy ROUTER legitimately relays every PKI DM between every pair of nodes in range, so a cap tight enough to matter risks re-breaking #11843 under exactly the load where it was reported. Trading a known functional bug for an unknown one is a bad deal, and I would rather bound the storm case properly than throttle the steady state badly.

### S7 — not worsened here

Being straight about this one rather than padding the list: `perhapsGenerateImplicitAckForOwnOverheard()` is already reached from the opaque branch on `develop`, and `fd34464` only widens it to the ~1/256 of nodes holding a hash-0 channel, where it was previously broken by accident. The absorb-and-replay attack (false "Delivered to mesh" plus `stopRetransmission()`) works identically on `develop` today. It is a real weakness and worth its own issue, but it is not this PR's to carry and I am not going to manufacture a change request for it.

---

Net: S1, S4, S5, S6 are four small, independent changes, none of which touches the `CORE_PORTNUMS_ONLY` fix that #11843 actually needs. S3 is a constant and a sizing rule. I think the PR lands cleanly with those.

---

thebentern (approved):

NomDeTom:
Flash, develop-merged head `705e79b91` vs develop `3468af94a` (CI runner): rak4631 **+16 B**, nrf52_promicro_diy_tcxo **+16 B**; static RAM ±0.

Issue #11795, opened 2026-09-09
[Tracking]: RP2xx0 assert and watchdog


NomDeTom:
Assert and watchdog behaviour on RP2xx0 is unusual compared to other architectures, and might be a suitable subject for a deep dive into the hardware.

Possible related issues:

- https://github.com/meshtastic/firmware/pull/11632
- https://github.com/meshtastic/firmware/pull/11234
- https://github.com/meshtastic/firmware/pull/10430
- https://github.com/meshtastic/firmware/pull/11235 (DRAFT)
- https://github.com/meshtastic/firmware/issues/11220
- https://github.com/meshtastic/firmware/pull/11630
- https://github.com/meshtastic/firmware/pull/11641

Simplycissmus:
> **Correction:** the `__assert_func` cited below from `newlib_interface.c:164` is inside `#if 0` and is not built. See the hardware results further down.

Reading the tree rather than the hardware, it looks like two gaps that line up.

**RP2xx0 is the only architecture without its own `__assert_func`.** nRF52 (`main-nrf52.cpp:183`), nRF54L15 (`main-nrf54l15.cpp:72`) and STM32WL (`main-stm32wl.cpp:294`, via `--wrap`) each install one. RP2xx0 falls through to arduino-pico's `__weak` default, `cores/rp2040/sdkoverride/newlib_interface.c:164`:

```c
void __weak __assert_func(const char *file, int line, const char *func, const char *failedexpr) {
    weak_raw_printf("assertion \"%s\" failed: ...");
    _exit(1);
}
```

`weak_raw_printf` is `#define`d to plain `printf` at the top of that file, so the message never reaches `LOG_ERROR`. `_exit` is twenty lines earlier at `:40`: `while (1) { __breakpoint(); }`, unless `PICO_ENTER_USB_BOOT_ON_EXIT` is set. Same in 5.6.1 and 6.0.0. Where nRF52 logs the failed expression and calls `NVIC_SystemReset()`, RP2xx0 prints where nobody is listening and then stops. These asserts are live — `airtime.h:94` already notes that nothing in this tree defines DEBUG or NDEBUG.

[...]

PR #11853, opened 2026-09-14
fix(rp2xx0): log and reset on a failed assert instead of hanging

Simplycissmus:
On RP2xx0 a failed `assert()` neither logs nor resets. The platform has no `__assert_func` of its own, so newlib's runs: it prints to stdio, which is not the Meshtastic log, then `abort()` reaches arduino-pico's `_exit`, which is `while (1) { __breakpoint(); }`. Without a debugger that stops the chip. Once `rp2040Loop()` has armed the 8 s watchdog, the node comes back after a silent stall. An assert during `setup()`, before the watchdog exists, hangs until power is removed.

nRF52, nRF54L15 and STM32WL already install their own handler. This adds one for RP2xx0 along the lines of the nRF52 one: log the failed expression with `LOG_ERROR`, then reboot through `watchdog_reboot()`. Refs #11795.

### Tested

Raspberry Pi Pico W (DIY variant with LLCC68), measured for #11795 with a compile-time hook that forces an assert inside `rp2040Setup()`. Resets were counted from USB enumeration with the serial port closed; the port was opened only to capture the log line.

- Without this handler: no reset in 188 s. The device stays enumerated but does not respond, and only BOOTSEL recovers it.
- With this handler: 38 resets in 240 s, 5.5–6.0 s apart, and `assert failed src/platform/rp2xx0/main-rp2xx0.cpp: 140, void rp2040Setup(), test=assertTestZero` on every boot.

The handler in this PR is the one from that run. The measured build was develop as of 17 Aug, and `main-rp2xx0.cpp` has not changed on develop since then. This branch builds for `picow` (RP2040) and `pico2w` (RP2350), and on both the linker map places `.text.__assert_func` in `main-rp2xx0.cpp.o` (`0x1006347c` on picow, `0x10060b74` on pico2w), not in newlib. On RP2350 it was tested by t-miura on a W5500-EVB-Pico2, with the same result; see the comment below.

Two side effects:

- A reset through `watchdog_reboot()` is reported as "Rebooted by watchdog" on the next boot, the same as a real watchdog timeout or `picotool reboot`. The `assert failed` line is what tells them apart.
- An assert that fails on every boot now gives a reboot loop instead of a hang. The node stays reachable over USB: in the test it went into BOOTSEL via `picotool reboot -f -u` and the 1200 bps touch, without pressing the button.

No separate regression run: the handler is only reached when an assert fails. No Heltec/RAK/T-Beam hardware was available; the change only touches the rp2xx0 platform.

  - [ ] Heltec (Lora32) V3
  - [ ] LilyGo T-Deck
  - [ ] LilyGo T-Beam
  - [ ] RAK WisBlock 4631

## Summary by CodeRabbit

- **Bug Fixes**
  - RP2xxx assertion failures now produce an error log and automatically restart through the watchdog, improving recovery from startup and runtime faults.


vidplace7:
@t-miura can you check this out?

t-miura:
@vidplace7 will test on rp2350 asap!

t-miura:
@vidplace7 @Simplycissmus 

Tested and verified on my W5500-EVB-Pico2 w/ Wio-SX1262.
This board is RP2350-based with no Wi-Fi/BLE but has ethernet via W5500.
(a.k.a.: meshtupper, it's maybe a first time i kept opened its lid for this long, even an hour?)

and happy to report that issue was reproducable/observed, and the fix does fix the issue.

1. Unpatched develop baseline (with forced assert in `rp2040Setup()`):
- Symbol `__assert_func` resolved to newlib libc (`assert.c:57`).
- 0 bytes on serial, 0 resets in 20s. 1200 bps touch blocked indefinitely.
  - Exactly same as what happened on RP2040's test
- The board was permanently stuck in `_exit()`'s breakpoint loop
  - To recover, pressing the BOOTSEL button while reset(or re-plug the PoE Ethernet or USB cable w/no-PoE eth) was required.

2. With this PR (with forced assert in `rp2040Setup()`):
- Symbol `__assert_func` correctly resolved to `main-rp2xx0.cpp.o`.
- Observed exact same behavior with #11795 's RP2040
  - `ERROR | assert failed src/platform/rp2xx0/main-rp2xx0.cpp: 160, void rp2040Setup(), test=assertTestZero` on every boot
  - After logging that line, node does watchdog-trrig'd restart via `watchdog_reboot(0, 0, 10)` (observed 31 cycles in 30 s).
- 1200 bps touch during the reboot loop succeeded and dropped the board into BOOTSEL mode without needing to touch the physical button.

3. Clean build
- Clean boot, Ethernet, MQTT, sensors, and UI frames operational with zero obvious regressions.

Simplycissmus:
Thanks for running this on the Pico2, and for the baseline without the patch — that was the half I could not do on RP2350. The 31 resets in 30 s against my 38 in 240 s is the open serial port: with the port closed the loop here ran at 5.9 s, with it open at about 1 s. I've updated the description to point at your results.

PR #11858, opened 2026-09-15
Stop writing the BLE pairing passkey to the nRF52 log

jamesarich:
## Summary

On nRF52 the Bluetooth pairing passkey is written to the log, at BLE init and again on every pairing attempt. Anyone who can read the log can pair.

This matters most on lockdown builds, which are nRF52 only. `PhoneAPI` zeroes `fixed_pin` out of the Bluetooth config for a client that has not supplied the lockdown passphrase, and the log stream returns the same value to that same connection.

Removing it costs nothing on this platform, because nRF52 already sends the passkey to the screen as a `pairing_pin` notification. The log was never the only way to read it.

## What changed

The passkey value is no longer formatted into either log line. `Bluetooth pin set to '%i'` becomes `Bluetooth pairing passkey configured`, and the value is dropped from the pairing message. The two split buffers that existed only to print it are removed.

Other platforms that log the value are left alone: `NimbleBluetooth.cpp:657` on ESP32, and `NRF54L15Bluetooth.cpp:476` and `:710`. Lockdown does not exist on those targets, and on a node with no screen the console may be the only place the passkey appears, so removing it there is a product decision rather than a cleanup.

This does not gate the log stream itself on lockdown authorization, which is the other half of the problem. That gate cannot live in the logging path as things stand. It would have to consult `PhoneAPI::getAdminAuthorized()`, which takes the non-recursive `g_authSlotsMutex`, or `EncryptedStorage::isLockdownActive()`, which takes `spiLock`. Both locks are already held across `LOG_` calls, by `findOrAllocSlot_LH()` and `SafeFile::open()` respectively, so either check would deadlock the first time it fired. Gating the stream needs a lock-free view of the auth state first.

## Testing

Builds clean for `rak4631` in both stock and lockdown configurations. The linked ELF no longer contains `Bluetooth pin set to` or `with passkey`, and does contain the replacement string.

`test_stream_api` passes 18/18 under the native environment.

Not exercised on hardware: this removes log output and changes no pairing behavior, and the nRF52 screen path that carries the passkey is unchanged.


caveman99:
In Lockdown Mode the passkey is not printed. This line was useful in the past when you had a DIYY device with a screen and you removed/broke the sceen. The device assumes it's in 'generate a PIN' mode but you can see it only in the logfile.

PR #11862, opened 2026-09-15
test(harness): make run-tests.sh drivable by a caller that cannot see the terminal

NomDeTom:
## What

`bin/run-tests.sh` gains a run record, `--status` / `--wait` / `--abort`, process-group ownership of the `pio` build tree, a persisted last verdict, and three new exit codes. Plus a WSL forwarder (`bin/run-tests.cmd`) and the matching doc updates. No change to the verdict logic, the shared-state check, shards or CI.

One `src/` commit rides along (`17527b389`): `WaypointStore::clearAllWaypoints()` wrote a two-byte empty store file unconditionally, so `test_waypoint_expiry` has left `Waypoints_default.wpts` behind in every full run since it landed (#11600, 2026-08-25) and the unfiltered suite has read `AMBER` ever since. It now checks `FSCom.exists()` first - an existing file, stale or unreadable included, is still rewritten as empty, so a reset after a failed load clears flash as before; a file that is not there is left not there. The `exists()` check takes `spiLock` in its own scope and releases it before `SafeFile`, which locks it itself.

## Why

The wrapper was written for a human at a terminal. A caller that reads a captured output file and can be interrupted mid-run (a tool call, a backgrounded job, a fresh session) trips over two things:

1. **An interrupt leaves an orphan nobody can see.** The script trapped only `EXIT`, to delete its temp logs. Kill the shell and `pio → scons → cc1plus` keep running with no lock, no pid file, and the log gone. The next invocation either starts a second build into the same `.pio/build/` (and `rm -rf`s the running run's state summary), or `pgrep -f`s and matches itself.
2. **Verdict-shaped text before the verdict.** pio prints `[PASSED]` per suite and `N succeeded` per invocation long before the wrapper's `RESULT:` line exists, so a partial output file reads as a completed green run.

## How

- **Run record** `.pio/runtests/current.tsv` (pid, pgid, start, args, HEAD, log, progress) for the life of a run; removed by the one `result()` function every verdict now goes through. A second invocation prints `RESULT: BUSY`, exit **4**, and touches nothing. A record is valid while the holder pid *or* the recorded process group is alive — `flock` alone would not do, since a SIGKILLed holder releases the lock while its scons children live on; that case reports ORPHANED and only `--abort` (or the tree finishing) clears it.
- **`--status`** prints RUNNING / ORPHANED with progress, or IDLE with the last verdict. **`--wait`** attaches and exits with the verdict. **`--abort`** kills the tree by pgid from any shell.
- **`pio` runs under `setsid`** with its pgid recorded; `INT`/`TERM`/`HUP` kill the tree and record `RESULT: ABORTED` (exit **5**), log kept. Ctrl-C at a terminal now actually stops the build.
- **Last verdict on disk** in `.pio/runtests/last-result.tsv` with the kept log, plus a **tree fingerprint** (HEAD + working-tree diff + untracked files by content). `--status` recomputes it and marks the verdict **STALE** when the tree has changed — persisting a verdict without this would make "a GREEN from before the current edits" the most convincing false green there is.
- `result()` prints to the stdout the script *started* with (fd 3): a signal can arrive inside a pio call whose stdout is redirected to a log, and the verdict was disappearing into it.
- **Banner** on every run naming the final `RESULT:` line as the only verdict.
- **Non-Linux host** now prints `RESULT: UNSUPPORTED`, exit **6** — it used to exit 2, the AMBER code, which reads as "ran, passed with caveats".
- **A failed build removes `.pio/build/<env>/meshtasticd`**, which run bare would reprint the last good run. (An mtime check is not usable: an unchanged tree legitimately skips the relink.)
- FILTERED reports the not-run *count* instead of listing every suite outside the filter.
- `bin/run-tests.cmd`: `wsl.exe --cd "%~dp0.." -e ./bin/run-tests.sh %*`, exit code passed through. A forwarder, not a port — the header's argument for staying Linux-only stands, and CI never runs the native suite on Windows.

Exit codes 0–3 are unchanged. Everything new is keyed on the checkout's own `.pio/`, so separate worktrees still run independently.

## Verified

On WSL2: `--status` IDLE; a stale record (dead pid, dead pgid) self-clears; a real `-e native -f test_utf8 --quiet` shows RUNNING with the whole compiler tree in its own pgid; a concurrent invocation gets `BUSY` exit 4; `--abort` from another shell leaves no `scons`/`cc1plus` survivor and records ABORTED; a run to completion prints exactly the banner (stderr) and `RESULT: FILTERED …` (stdout), exit 3, and `--status` then reports it — marked STALE after a further edit. `trunk fmt` clean.

`test_waypoint_expiry` alone after the store fix: 10/10, no longer named in the shared-state block.

Not exercised: `drop_stale_program()` needs a failing build; `run-tests.cmd` needs a Windows shell — a checkout on the WSL filesystem (`\\wsl$\…` path through `wsl --cd`) is the case to try first.

## Summary by CodeRabbit

- **New Features**
  - Added status, wait, and abort controls for active test runs.
  - Added a Windows command that forwards test runs through WSL.
  - Retained the latest test result and log for later review.
  - Added unique run tracking so waits return the matching result.
  - Added BUSY, ABORTED, and UNSUPPORTED outcomes.

- **Bug Fixes**
  - Prevented concurrent test runs from interfering with one another.
  - Interrupted runs now report clear final outcomes.

- **Documentation**
  - Clarified commands, exit codes, platform guidance, and final-result interpretation.


NomDeTom:
Flash, develop-merged head `09de89da5` vs develop `3468af94a` (CI runner): rak4631 **±0 B**, nrf52_promicro_diy_tcxo **+8 B**; static RAM ±0.

PR #11867, opened 2026-09-16
Separate the nRF54 platform code from src/platform/nrf52

caveman99:
Adds src/platform/nrf54 with its own architecture.h and main-nrf54.cpp and removes every ARCH_NRF54L branch from src/platform/nrf52, which is back to its pre-port content apart from the SoftDevice flash event forwarding and the assert Serial.flush(). The nrf54l15 env includes only the unchanged shared files NRF52Bluetooth.cpp, Nrf52SaadcLock.cpp, alloc.cpp and hardfault.cpp from the nrf52 directory, so nrf52 builds never see nrf54 code and vice versa. Builds verified for nrf54l15dk, xiao_nrf54l15, xiao_nrf54l15_lr2021, heltec-mesh-node-t114 and wio-t1000-s; the xiao_nrf54l15_lr2021 image was run on the XIAO nRF54L15 with the LoRa Plus board (boot, LR2021 init, OLED, BLE advertising, watchdog, serial API).

## Summary by CodeRabbit

- **New Features**
  - Added support for nRF54L15-based boards.
  - Added power monitoring, Bluetooth control, deep sleep, watchdog management, device identification, and SEGGER logging.
  - Added recovery handling for corrupted local storage.
  - Enabled shared firmware capabilities including Bluetooth, display, GPS, sensors, telemetry, and radio features.

- **Bug Fixes**
  - Standardized reset, DFU, and error handling across supported Nordic platforms.
  - Improved watchdog startup validation and reliability.

Issue #13, opened 2020-03-04
Add support for 433MHz radios in the EU (current build is 865MHz only)


astro-arphid:
Hi! I'm really excited to give this a go, but just want to check something before flashing.

In the UK, the use of [433MHz is also permitted](https://www.thethingsnetwork.org/docs/lorawan/frequencies-by-country.html#u), but this isn't listed in the documentation here. Given that my hardware is compatible with this band, would Meshtastic work out of the box, or would additional changes to the code need to be made to support this?

Thanks!

astro-arphid:
Sorry - I got excited and didn't fully look over everything. Checked the firmware releases and there isn't anything for the hardware requirement at this time.

Closing for now, but will look into building a working version based on the documentation.

Thanks!

geeksville:
@astro-arphid  Sounds good.  If you wanna shoot over a PR to add a EU433 region that would be great.  Essentially you'd need to

* change bin/build-all to add EU433 
* change src/MeshRadio.h (see CH0_EU for example) 
* and the README.md to mention the new choice of 433 mhz radios for EU.

geeksville:
(the current code for EU builds configures the radios for 865MHz - which you definitely don't want if the mfg stuffed parts for 433MHz)

geeksville:
btw @astro-arphid do you think you might be able to sign up for this task?  I'm sort of able to do it, but it would be better if someone that actually has 433MHz hardware tried tested it.

astro-arphid:
@geeksville I'll certainly give it a go this weekend, if everything goes to plan!

astro-arphid:
@geeksville Looking over the [LoRA Alliance paramaters](https://lora-alliance.org/resource-hub/rp2-101-lorawanr-regional-parameters) at the moment and just want to check that I'm following this correctly - let me know if this isn't the standard being worked with. 

Based on the documentation, 433.175MHz is the minimum center frequency and 434.665MHz is the maximum, where 433.175, 433.375 and 433.575 are specified as channels 0, 1 and 2. Would it make sense to have the following, maxing out at 8 channels, where channel 8 would be 434.575MHz?

```
// src/MeshRadio.h
// EU433 channel settings
#define CH0_EU443          433.175f // MHz
#define CH_SPACING_EU433   0.2f // MHz
#define NUM_CHANNELS_EU433 8
```

geeksville:
Yep.  That looks perfect!

(Sent from a phone - please ignore typos)

[...]

PR #11872, opened 2026-09-16
fix(nrf52): stop concurrent flash writers corrupting LittleFS, and stop a failed save formatting it

thebentern:
Addresses the nRF52 reports of **critical error #13** (`FLASH_CORRUPTION_UNRECOVERABLE`), where a node
loses its region, channels and keypair after a filesystem format. Every commit here is one flash write
failing because another operation was in flight.

## The substrate

`flash_nrf5x.c` has **one** 4 KB page cache, **one** SoftDevice flash semaphore and **one** result word,
shared by every caller. With BLE up, flash operations are asynchronous: a competing operation gets
`NRF_ERROR_BUSY` and retries 20 x 1 ms against an **85 ms** page erase, then gives up — and
`flash_cache_flush()` discards the erase and program return values and invalidates the page anyway.
32 LittleFS blocks vanish (128 B block on a 4 KB page) with no error reaching the filesystem.

Worse than a lost error code: `flash_cache_write()` **repoints** that shared cache when the requested
page differs from the cached one. A second writer arriving mid-write flushes the first writer's page and
re-points the buffer, so the first writer's remaining `memcpy` lands in the **wrong page's image**. Warm
ring records inside LittleFS metadata, or the reverse.

## What each commit does

**1. Serialise the warm-node ring against LittleFS.** On nRF52840 the warm store writes its 3-page
record ring straight through `flash_nrf5x_*` holding only `spiLock`. Every LittleFS writer instead holds
Adafruit_LittleFS's own mutex — and two of them run on other tasks that never take `spiLock`: Bluefruit's
bond saves on the callback task, and phone config writes on the BLE task. The ring now takes the
filesystem mutex too, **after `spiLock` and never before**, matching the order every existing path uses.
The ring calls no LittleFS function, so the non-recursive mutex is never re-entered. Longest new hold is
a page rotation, estimated at roughly half a second and not yet measured. The 90 s watchdog is the budget
that actually applies; the 2 s supervision timeout cited in an earlier revision of this description is
not, since the SoftDevice maintains the link in interrupt context and is indifferent to a stall on an
application task.

**2. Make `saveProto` report a failed readback or rename.** It returned `pb_encode`'s result and threw
away `SafeFile::close()`, so a failed verify or rename was logged and then reported as success. For the
non-atomic saves the original file was already deleted at that point, so the node came back with a
missing `config.proto`, installed defaults and minted a **new identity** — with no error recorded. The
recovery path below never fired for the one failure it exists for.

**3. Retry a failed save before formatting, and never format on a low rail.** `saveToDisk` went straight
from one failed write to `fsFormat()`, which takes every file with it. A failed write is far more often a
busy SoftDevice or a sagging rail than a corrupt filesystem. Now: two retries 150 ms apart, the power
level re-checked before each retry **and** immediately before the format, `fsFormat()`'s result checked,
and the post-format resave uses the **full** segment mask — a nodes-only save that ended in a format used
to drop `config.proto` and re-key the node. Lockdown builds skip the plaintext resave when the format
just took the DEK. RP2040 feeds its watchdog across the delays.

**4 and 5. Quiesce flash before any software reset.** A reset mid-page-program tears the page. The nRF52
reboot, DFU entry and deep-sleep paths now drain the shared cache and hold both locks first. The library's
stock BLE DFU handler jumps to the bootloader from the callback task, so it gets a wrapper that quiesces
first — and **drops the filesystem mutex before calling the library handler**, because that handler opens
the bond file and the mutex is non-recursive; holding it would wedge the node until the 90 s watchdog.

**6. Tests.** `test_nodedb_save_retry` pins the contract: a rail dip mid-write retries and lands with
devicestate still present (proving no format fired), a rail still unsafe at retry bails without
formatting, and a failed rename returns false.

## Not in scope

This closes the **collision**, not the **loss**. A genuine busy-timeout from SoftDevice-versus-radio
contention still drops 32 blocks silently, because `flash_cache_flush` still discards its results. That
fix belongs in the pinned Adafruit fork and should land after this, not before: surfacing those errors
without the retry above would turn a transient into a format.

The stolen-flash-event half of this bug was fixed independently on develop while this branch was in
progress (forwarding `NRF_EVT_FLASH_OPERATION_*` to the driver), so that commit was dropped from here.
One gap survives there: the `default:` branch still swallows `NRF_EVT_POWER_USB_*`, which Bluefruit
routes to `tusb_hal_nrf_power_event`.

## Testing

- `rak4631` builds (91.8% flash), `xiao_nrf54l15` builds — the nRF54L target matters because the DFU
  change touches its code path. Since #11867 split the nRF54 platform out of `src/platform/nrf52`,
  `nrf52FlashQuiesce()` is defined in both platform mains; `xiao_nrf54l15` does not link without it,
  because `NRF52Bluetooth.cpp` and `Power.cpp` are shared and nRF54L still defines `ARCH_NRF52`.
- Full native suite **1461/1461** in Docker.
- Review round (@caveman99): `rak4631`, `makerpython_nrf52840_sx1280_oled` (the SX128x nRF52 target that
  exercises the new `main.h` include), `xiao_nrf54l15`, `rak11310` and `betafpv_2400_tx_micro` all build;
  suite still 1461/1461. The RP2040 watchdog feeds and the DFU opcode gate compile out on native, so both
  are compile-verified only and want a bench check: an rp2350 forced through the format path with a full
  node DB, and a rak4631 written a non-0x01 then a 0x01 byte on the DFU control point.
- Still open from that review: on a full 28 KB InternalFS the node stops persisting and only logs it. A
  capacity gate is deferred — `fsUsedBytes()` is not trustworthy across all five backends, and
  `lfs_dir_alloc()` needs two free blocks, so `used < total` misses the near-full case it would target.
  The non-atomic `nodes.proto` save (NodeDB.cpp:3278) destroys the node DB on every failed attempt on all
  platforms and is the larger of the two; both are follow-ups, not this PR.
- Review round (@NomDeTom): the retry loop lengthens a stall on the BLE task specifically. A phone config
  write reaches `saveToDisk()` synchronously there — `toRadio`'s write callback is registered with
  `useAdaCallback = false` (NRF52Bluetooth.cpp:271), so it runs inline on Bluefruit's BLE task rather than
  the deferred callback task. Three failed attempts therefore cost that task 300 ms of `vTaskDelay` plus
  two extra full-mask write attempts, each able to block further on `spiLock` behind a warm-ring rotation.
  The link is not at risk and the phone's write is acked in interrupt context; what is delayed is the
  `fromRadio` reply and any queued BLE events. Set against it, the same path previously formatted the
  filesystem on the *first* failure, so the common transient case gets shorter — only the worst case grows.
  For the bench list: measure BLE-task occupancy across the retry loop *plus* whatever rotation it waits
  behind, not the rotation alone, which is what the figure above quotes.
- **Not yet exercised on hardware.** The intended bench proof is a byte-level one: drive warm-ring
  evictions and BLE config writes concurrently, then dump `0xEA000` for `0xA000` from the UF2 bootloader
  and check for the ring magic `WRN3` inside the LittleFS region at `0xED000`. Worth knowing for anyone
  attempting it: SWD-flashing an nRF52 app looks like it works and then will not boot, because the
  bootloader validates the app against its settings page and a direct flash bypasses that update.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved storage reliability by detecting close failures and retrying unsuccessful writes.
  * Added power-safety checks to prevent unsafe storage operations and protect encrypted data during recovery.
  * Prevented unnecessary filesystem formatting when storage remains readable after a failed save.
  * Improved nRF52 and nRF54 reset, sleep, and DFU transitions by safely completing pending flash activity first.
  * Prevented conflicts between flash storage operations and filesystem access.
* **Tests**
  * Added coverage for failed saves, retry behavior, power interruptions, and unsafe power conditions.


src/mesh/NodeDB.cpp
```diff
@@ -3408,13 +3409,45 @@ bool NodeDB::saveToDisk(int saveWhat)
...
+        LOG_ERROR("Save to disk failed after retries, formatting");
+#ifdef MESHTASTIC_ENCRYPTED_STORAGE
+        // The format takes the DEK with it, and without it the resave below would land the keys in plaintext.
+        const bool lockdownWasActive = EncryptedStorage::isLockdownActive();
+#else
+        const bool lockdownWasActive = false;
+#endif
         spiLock->lock();
-        fsFormat();
+        const bool formatted = fsFormat();
```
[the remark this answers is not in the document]
thebentern: Fixed in b71989509 — you found a real hole, and it was one this PR opened.

Making `saveProto` honest (commit 2) gave the recovery path a new way in: any persistent write failure now reached `fsFormat()`, which takes every file with it. A busy or lock-protected nRF52 flash fails *every* write for as long as it lasts, so two retries cannot tell that apart from corruption, and guessing wrong costs the node its config, keys and bonds — the exact outcome this PR exists to prevent.

Rather than thread a reason code through the whole save stack, I used an asymmetry in the hardware: **reads never touch the SoftDevice write path a busy flash fails on.** So before formatting, `saveToDisk` now checks whether `/prefs` still walks and a stored proto still opens and reads. If it does, the metadata chain is intact, the failure was transient, and it returns false so the caller can retry later — no format, no critical error recorded. Genuine corruption is not silently tolerated: lfs asserts on it and the nRF52 handler reboots and formats on the way back up.

That keeps the fix in one function instead of changing every signature in the save path, which I think is the right trade here given how many callers `saveProto` has.

New test `test_saveToDisk_writeFailsButFsReadable_doesNotFormat` pins it, and I checked it is not vacuous — with the gate removed it fails (3/4).

src/platform/nrf52/main-nrf52.cpp
```diff
@@ -520,6 +522,15 @@ void nrf52Setup()
     assert(r == NRFX_OK);
 }
 
+// Waits out any flash write another task has in flight, drains the shared page cache, and keeps
+// both locks: the caller resets next, and a reset mid-program tears the page.
+void nrf52FlashQuiesce()
+{
+    spiLock->lock();
+    InternalFS._lockFS();
+    flash_nrf5x_flush();
```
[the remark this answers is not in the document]
thebentern: Correct, and deliberately out of scope for this PR — but worth stating plainly rather than leaving implied.

`flash_cache_flush()` calls `fc->erase()` and `fc->program()`, discards both results, and invalidates the page regardless; `_internal_flash_sync()` then returns 0 unconditionally. So `nrf52FlashQuiesce()` genuinely cannot know whether the flush landed, and its callers reset anyway. Those two functions live in the pinned Adafruit fork (`libraries/InternalFileSytem/src/flash/`), not in this repo, so the fix is a fork change plus a pin bump.

What the quiesce does buy in the meantime is **ordering, not verification**: it takes `spiLock` and the filesystem mutex and drains the cache, so no *other* task has a write in flight when the reset lands. That removes the multi-writer tear. A single-writer failure that the layer below swallowed is still invisible, and I would rather ship the ordering fix honestly than claim more than it does.

On sequencing, which matters here: surfacing those errors to lfs **before** the format gate in b71989509 would have made things worse, not better — every transient busy failure would have become a format. With that gate in place, propagating the failure is safe to do next. That is the follow-up, and it belongs in the fork.

src/mesh/NodeDB.cpp
```diff
@@ -3408,13 +3434,50 @@ bool NodeDB::saveToDisk(int saveWhat)
...
+    }
+
     if (!success) {
-        LOG_ERROR("Save to disk failed, retry");
+        if (!powerHAL_isPowerLevelSafe()) {
+            LOG_ERROR("saveToDisk() on unsafe device power level");
+            return false;
+        }
+        // The format below takes every file with it, so spend one read proving it is warranted.
+        if (filesystemStillReadable()) {
```
caveman99, line 3462: When every retry fails but a pref file still reads back, this returns without formatting and without recording a critical error.

That is right for a busy SoftDevice, but the same state results from `LFS_ERR_NOSPC`, or from `LFS_ERR_CORRUPT` hit during allocation: `lfs_alloc` returns those as errors, so reads of `/prefs` keep working while every write fails. Before this change, the format plus full resave from RAM brought config and identity back in that state, at the cost of bonds and the other files. With this branch the node stops persisting for good and no client is told. `nodes.proto` is saved non-atomically, so each attempt also deletes it before failing. On the 28 KB InternalFS a full filesystem is a realistic case.

The premise in the doc comment on `filesystemStillReadable()` does not hold on nRF52 either. In the pinned framework fork (0fd295f), `LFS_ASSERT` is plain `assert()`, which ends in `__assert_func` in main-nrf52.cpp and resets without setting `NRF52_MAGIC_LFS_IS_CORRUPT`. `lfs_assert()`, the path that does set it, has no caller in the fork. Only a failed mount formats.

Suggest recording a critical error on this path at minimum, and letting the format through when the failure repeats across calls or `fsUsedBytes()` reports the filesystem full.

src/mesh/NodeDB.cpp
```diff
@@ -3408,13 +3434,50 @@ bool NodeDB::saveToDisk(int saveWhat)
...
+        LOG_ERROR("Save to disk failed and the filesystem is unreadable, formatting");
+#ifdef MESHTASTIC_ENCRYPTED_STORAGE
+        // The format takes the DEK with it, and without it the resave below would land the keys in plaintext.
+        const bool lockdownWasActive = EncryptedStorage::isLockdownActive();
+#else
+        const bool lockdownWasActive = false;
+#endif
         spiLock->lock();
-        fsFormat();
+        const bool formatted = fsFormat();
```
caveman99, line 3474: On RP2040 the watchdog is only fed at the top of each retry. The last retry, `filesystemStillReadable()`, the format and the full five-segment resave then share one 8 s budget. If it expires during the resave, the filesystem is left empty and the node boots on defaults with a new identity. A `watchdog_update()` before the format, and again before the resave, avoids that.

src/Power.cpp
```diff
@@ -1013,6 +1013,7 @@ void Power::reboot()
 #if defined(ARCH_ESP32)
     ESP.restart();
 #elif defined(ARCH_NRF52)
+    nrf52FlashQuiesce();
```
caveman99, line 1016: One software reset is still not quiesced: `SX128xInterface::reinitChip()` saves the region, waits 2 s and calls `NVIC_SystemReset()` directly (src/mesh/SX128xInterface.cpp). It only runs at boot before BLE is up, so the exposure is small, but routing it through `nrf52FlashQuiesce()` would make "every software reset" in the commit title hold.

src/platform/nrf52/NRF52Bluetooth.cpp
```diff
@@ -21,7 +23,39 @@ static BLEDis bledis;             // DIS (Device Information Service) helper cla
...
+class QuiescingBLEDfu : public BLEDfu
+{
+    struct Peek : BLECharacteristic {
+        using BLECharacteristic::_wr_authorize_cb;
+    };
+    static BLECharacteristic::write_authorize_cb_t libraryCb;
+
+    static void onControlWrite(uint16_t conn_hdl, BLECharacteristic *chr, ble_gatts_evt_write_t *request)
+    {
+        nrf52FlashQuiesce();
```
caveman99, line 44: This quiesces on every control-point write, before the library handler has checked the CCCD or the opcode, and only START_DFU (`data[0] == 1`) actually resets. Every other write blocks the Bluefruit callback task on `spiLock`, forces an early page cache commit, and holds back the GATT authorize reply. Gating the quiesce and the matching unlock on `request->len && request->data[0] == 1` keeps those writes off the locks.

src/mesh/NodeDB.cpp
thebentern: Confirmed and fixed. It is worse than pre-existing too - this PR is what lengthened that window, by inserting filesystemStillReadable() ahead of the format and widening the resave from saveWhat to all five segments (NodeDB.cpp:3487-3488). On develop it was fsFormat() followed by a single-segment resave.

I put the first feed one line earlier than you suggested, before filesystemStillReadable() rather than before fsFormat(), so the probe and the format share a fresh budget, and the second right after the format's spiLock->unlock() so the whole resave gets its own. It has to go there rather than "before the resave" literally, since the resave sits in the else arm of an if/else-if chain. Both feeds are outside any lock. No new include - <hardware/watchdog.h> is already pulled in at NodeDB.cpp:74-76 - and I matched the raw #ifdef ARCH_RP2040 form you used at NodeDB.cpp:3292 rather than introducing a second spelling. The retry-loop feed at :3442 stays as is.

On whether 8 s is really tight: the tree already answers that. mesh-pb-constants.h:166-169 sets WARM_NODE_COUNT 150 for ARCH_RP2040 with the comment that it is bounded so the warm.dat write fits the 8 s watchdog, deviating from the RAM-derived class on purpose. fsFormat() itself is cheap - LittleFS::format() is unmount, lfs_format, remount, not a device erase - so the budget goes almost entirely into the resave, which is where you put it.

Residual worth naming rather than papering over: even a fresh 8 s may not cover config + moduleconfig + channels + devicestate + nodes.proto on a full 120-node DB, since #10746 showed nodes.proto plus warm.dat alone could overrun one budget, and every saveProto is a write plus a full testReadback plus a rename (SafeFile.cpp:60-110), so that is five verify-rename cycles in one window. The robust form is a single feed at the top of NodeDB::saveProto(), which covers all five segments, both retries and the ordinary multi-segment save. That changes watchdog behaviour for every RP2040 save, not just the post-format recovery, so I would rather not fold it into an nRF52 fault-13 PR - happy to open it separately and link it here if you agree. These two feeds fix what this PR lengthened; that one is a behaviour change for the platform generally.

src/Power.cpp
thebentern: Agreed, added. nrf52FlashQuiesce() now runs immediately before NVIC_SystemReset() at SX128xInterface.cpp:104, after the existing delay(2000) so spiLock is not held across the 2 s stall, plus an #include "main.h" for the declaration at main.h:117 (the file's existing includes do not reach it - NodeDB.h has no path to main.h).

I checked the deadlock case first, since nrf52FlashQuiesce() takes spiLock and never gives it back and spiLock is a plain binary semaphore taken with portMAX_DELAY (concurrency/Lock.cpp:11,22-27). It is provably unheld there: RadioInterface::init() takes no lock, RadioLibInterface has no init() override, the only radio-path acquisition is LockingArduinoHal::spiBeginTransaction/spiEndTransaction (RadioLibInterface.cpp:21-33), which RadioLib pairs strictly inside lora.begin() - and both pairs in Module.cpp are straight-line with no early return between them. The clincher is two lines above: nodeDB->saveToDisk(SEGMENT_CONFIG) takes spiLock itself, so the shipped code would already hang if it were held. The FS mutex is created in the Adafruit_LittleFS constructor (Adafruit_LittleFS.cpp:49), so it is valid pre-mount, and every _lockFS caller in src/ is spiLock-then-FS or FS-alone, so there is no inversion.

The include sits inside the file's existing `#if RADIOLIB_EXCLUDE_SX128X != 1` guard, so the variants that set RADIOLIB_EXCLUDE_SX128X=1 - rak4631, meshlink, gat562, t-watch-ultra, stm32 - pay nothing. And since src/platform/nrf54/architecture.h:4 also defines ARCH_NRF52, the one arm covers nRF54 too and resolves to main-nrf54.cpp:377; no second branch needed. ESP32 arm untouched.

One honest note on what this buys: lfs's .sync callback is _internal_flash_sync() -> flash_nrf5x_flush() (InternalFileSystem.cpp:92-95,107), so the config write on line 98 was already reaching flash on commit. What this closes is the residual in-flight-write and page-cache window, and it makes the claim in bbe866fc6 hold for the last reset path in this PR's scope. The comment in the tree says that rather than claiming it rescues the save.

I did not route it through Power::reboot(). setupModules() runs at main.cpp:1088, before initLoRa at :1165, so notifyReboot.notifyObservers() and waypointStore.saveToFlash() (Power.cpp:1009-1013) are live - that would add an observer fan-out and an extra flash write to a path whose whole job is to get to a clean reset, and it would change the ESP32 arm too.

I also left two other resets alone on purpose. __assert_func (main-nrf52.cpp:184-191) can fire from a context that already holds spiLock or the FS mutex - lfs_assert's own reset at :347 definitely does - so quiescing there would turn a fast assert-reboot into a hang until the 90 s watchdog. APProtect.cpp:89 is lock-free and would be safe, but it is a different subsystem with no save-then-reset pairing, so I would rather do it separately than widen this diff. So "every software reset" means every path where quiescing is safe, and I will say that in the commit body.

src/platform/nrf52/NRF52Bluetooth.cpp
thebentern: Gated, with one change to the predicate. I dropped the `request->len &&` term and matched the library byte for byte, because a len term here is stricter than the library's own reset condition. BLEDfu.cpp:110 is `if (request->data[0] == START_DFU)` with no len check, and Bluefruit hands us a stale byte rather than a zeroed one - bluefruit.cpp:747/761 reuse a single rtos_malloc'd event buffer for every SoftDevice event, and BLECharacteristic.cpp:445-448 copies sizeof(*wr_req) into the ada_callback payload, which includes the data[1] placeholder. So a zero-length write whose stale data[0] happens to be 0x01 still resets in the library, and with `len &&` in our gate that reset would run with no quiesce - the exact torn page this PR is fixing. Reading data[0] is always in bounds (ble_gatts.h:371, and it is inside the bytes that were copied); data[1] onward would not be, so the wrapper does not touch len at all.

The rest of your reading holds and is why the gate is worth having. The handler checks the CCCD at BLEDfu.cpp:99 and sends the SUCCESS authorize reply at :107 before it ever looks at the opcode, notifyEnabled touches no flash, and the only filesystem access in the whole handler is conn->loadBondKey() at :143 inside the START_DFU branch - so every other write was taking spiLock, taking and releasing the FS mutex and force-committing the page cache for nothing.

Two corrections to the framing, neither of which changes the outcome. The task being blocked is not the Bluefruit task in the normal case: setWriteAuthorizeCallback defaults useAdaCallback=true (BLECharacteristic.h:111), so this runs deferred on the Callback task at prio 2, not the BLE task at prio 3, and the authorize reply is already asynchronous - holding it back was never an ATT-timeout risk against the 30 s budget. What it did cost is real, and is a bit worse than you said: a spiLock stall on that task also holds back every other deferred Bluefruit callback, including the toRadio write path. And when ada_callback's allocation fails - heap exhaustion, which is the condition this PR is about - BLECharacteristic.cpp:446-449 falls back to invoking the handler inline on the BLE task at prio 3. That is why I also reworded the comment at :40-41: "this task outranks the loop" is only true on the deferred path, so the comment now claims only that spiLock fences the other writers until the jump.

After this the unsecure wrapper matches what this PR already does in BLEDfuSecure.cpp:125, where the quiesce sits inside the START_DFU branch after the CCCD check and the reply.

One thing I considered and left out: a null check on the captured library callback. BLEDfu::begin() installs it at BLEDfu.cpp:206 immediately before _chr_control.begin() at :207, and nrf52.ini:10 pins the framework to an exact SHA in our own fork, so it cannot move without a deliberate pin bump in this repo that would retest DFU anyway - and a silent fallback there would quietly disable this fix rather than fail loudly. Say the word if you want it and I will add it with a LOG_ERROR so it is not silent.

Same code path on nRF54 (variants/nrf54l15/nrf54.ini:39 compiles this file; the Bluefruit54Lib handler is identical with DFU_OTA_MAGIC 0xA8 and NVIC_SystemReset instead of the bootloader jump), so the gate covers both with no arch conditional.

src/mesh/NodeDB.cpp
thebentern: I went through this one carefully and I am deferring the capacity gate. Taking the pieces in order.

Every mechanism you describe checks out. lfs_alloc() returns LFS_ERR_NOSPC and propagates lfs_traverse()'s LFS_ERR_CORRUPT (Adafruit_LittleFS littlefs/lfs.c:287-330) while a read-only open never reaches the allocator, so filesystemStillReadable() returns true in exactly the state where every write is permanently failing. nodes.proto really is the one fullAtomic=false save (NodeDB.cpp:3278) and SafeFile.cpp:9-11 removes the live file before the first byte, so each failed attempt does destroy the old copy. 28 KB / 224 blocks is exact (InternalFileSystem.cpp:34-35,112), and mesh-pb-constants.h:98-101 already sizes MAX_NUM_NODES against that budget, so a full filesystem is a designed-for state, not a hypothetical.

Where I push back is the counterfactual. On develop the post-format resave is saveToDiskNoRetry(saveWhat) - the same mask that just failed - and the caller that actually fires here is saveToDisk(SEGMENT_NODEDATABASE) at NodeDB.cpp:3878. So develop formats everything and rewrites only nodes.proto: config.proto goes, taking config.security.private_key with it, and the node comes back with a new key and a new NodeNum once a region is set again. Pre-PR full-filesystem behaviour was a re-key, not "config and identity come back at the cost of bonds". This branch changes that twice over - the resave at NodeDB.cpp:3487 is now all five segments, which is what makes a format survivable at all, and the format no longer fires on a transient write failure. So in the state you name the node now keeps its NodeNum, channels and owner and loses only persistence, which a factory reset clears locally instead of breaking PKI for every peer that has it.

On recording a critical error: error_code is in no protobuf. Its only readers are Screen.cpp:1406 and NotificationRenderer.cpp:1153, so it is an OLED banner, not a client surface - it would not tell a phone either way, and that was equally true before this PR. On Portduino recordCriticalError() calls exit(2) for both FLASH_CORRUPTION codes (NodeDB.cpp:4866-4871), so an unconditional one on the still-readable path would abort a native node on a transient host-FS hiccup and take test_saveToDisk_writeFailsButFsReadable_doesNotFormat with it. A real client surface would be a meshtastic_ClientNotification from the failure path, which I am not bolting on here.

On letting the format through when fsUsedBytes() reports full: I implemented it and then threw it away, for two reasons. It does not catch the case you are describing - lfs_dir_alloc() needs two free blocks (lfs.c:408-415) and v1 commits relocate, so NOSPC fires while the traverse fsUsedBytes() runs still counts total-1 or total-2 in use, and `used < total` skips the format in the common near-full regime. And the helper is not trustworthy off nRF52: arduino-pico's _getUsedBlocks() returns the raw lfs error as an int into a uint32_t (LittleFS.h:296-301, :114), so a transient LFS_ERR_CORRUPT reads as ~4.29e9 and would format a healthy RP2040; nRF54 is an ARCH_NRF52 build on littlefs v2 whose lfs_fs_size() is documented best-effort and can exceed the real size (lfs.h:718-724); and on Portduino those helpers read the host volume (FSCommon.cpp:70-84) while fsFormat() is rmDir("/prefs") (FSCommon.cpp:159-161) and the critical error then exit(2)s - a full dev-machine disk would wipe the node and kill the process. Scoping it to littlefs v1 fixes the cross-platform half but still leaves the two-block case. Doing it right means a reservation margin like the one already at NodeDB.h:151-156 plus a bench run on a deliberately filled rak4631, which is its own change.

The repeat-across-calls half I am declining outright: a BLE DFU, a long xmodem transfer that slips the xModem.isBusy() guard at NodeDB.cpp:3204, or a sustained low rail can all span many saveToDisk() calls, and formatting there is the fail-deadly wipe this PR exists to stop.

What I did change is the log line, so the state is at least diagnosable from a field log: NodeDB.cpp:3459 now reads "Save to disk failed but the filesystem still reads, not formatting (full or busy?)".

Two follow-ups I will open and link here, in the order I think they matter. First, the non-atomic nodes.proto save - that destroys the node DB on every failed attempt on all platforms and all paths, independent of the format decision, and it is the larger of the two data-loss bugs you found here. Second, the capacity decision itself, with a margin constant and per-backend scoping. If you would rather the second one block this PR, say so and I will do it there instead, but I do not want to land a half-right predicate that re-arms a wipe.

---

The premise does hold on nRF52 - LFS_ASSERT is never plain assert() in this build. variants/nrf52840/nrf52.ini:22 force-includes variants/nrf52840/cpp_overrides/lfs_util.h into every TU in the env, including the framework's own lfs.c, and nrf52.ini:40 sets -DLFS_NO_ASSERT. That selects the override's #else branch at cpp_overrides/lfs_util.h:78-82, `extern void lfs_assert(const char *reason); #define LFS_ASSERT(test) if (!(test)) lfs_assert(#test)`. The `assert(test)` you quoted is line 77, the branch that is not compiled. Worth noting the fork's own header would not give you assert() either under that flag - its #else is an empty macro (libraries/Adafruit_LittleFS/src/littlefs/lfs_util.h:76-81).

Empirically, in .pio/build/rak4631/firmware-rak4631-2.8.1.9919d03.elf: nm shows `lfs_assert` at 00083194 distinct from `__assert_func` at 00083160, and objdump counts 8 `bl <lfs_assert>` call sites, resolving into lfs_cache_prog, lfs_cache_read, lfs_ctz_find, lfs_file_sync, lfs_file_write and lfs_init - framework lfs.c code, which is the proof the -include reached that compile unit. The image also carries the stringized expressions only that macro can produce (`block < lfs->cfg->block_count`, `head >= 2 && head <= lfs->cfg->block_count`). xiao_nrf54l15 has 59 call sites via variants/nrf54l15/cpp_overrides/lfs_assert.h. The symbol is ours, defined at main-nrf52.cpp:310 / main-nrf54.cpp:253, which is why grepping the package for it comes up empty.

"Only a failed mount formats" is also not right: fsInit() calls the weak preFSBegin() before FSBegin() (FSCommon.cpp:404-411), and preFSBegin() (main-nrf52.cpp:296-307) formats on RESETREAS==0 && GPREGRET==NRF52_MAGIC_LFS_IS_CORRUPT, which lfs_assert set at main-nrf52.cpp:337-342 before NVIC_SystemReset(). That format runs ahead of any mount attempt; InternalFileSystem::begin()'s mount-failure format is a separate second path. The installed package matches the pin - git rev-parse in framework-arduinoadafruitnrf52 gives 0fd295f13203e93df19d578073646ec32f2bf45a, same as nrf52.ini:10.

That said, you are right that the sentence should not be there, for a different reason: NodeDB.cpp is arch-shared and the assert/GPREGRET/format chain exists only on nRF52 and nRF54 - there is no equivalent on ESP32 or RP2040 - and even on nRF52 most real corruption surfaces as an LFS_ERR_CORRUPT return rather than an assert, plus an assert during brownout deliberately skips the magic (main-nrf52.cpp:337). So I trimmed line 3400 to `/// resolves means the write failure was transient, not a filesystem that needs formatting.`, which claims only what the function below it proves. I did not replace it with a statement about corruption in general, because the loop returns after the first existing pref file - a corrupt nodedb.proto with an intact devicestate.proto passes this probe, and I did not want to write a second claim the code does not back.

One related thing worth putting on the record while we are in here: lfs_assert's own NVIC_SystemReset() at main-nrf52.cpp:347 is deliberately left unquiesced and must stay that way. It fires from inside lfs with InternalFS._lockFS() already held, and nrf52FlashQuiesce() takes that same non-recursive mutex (Adafruit_LittleFS.cpp:49), so "completing" the quiesce coverage there would hang the corruption-recovery path until the watchdog.

src/platform/nrf52/NRF52Bluetooth.cpp
```diff
@@ -21,7 +23,44 @@ static BLEDis bledis;             // DIS (Device Information Service) helper cla
...
+{
+    struct Peek : BLECharacteristic {
+        using BLECharacteristic::_wr_authorize_cb;
+    };
+    static BLECharacteristic::write_authorize_cb_t libraryCb;
+
+    static void onControlWrite(uint16_t conn_hdl, BLECharacteristic *chr, ble_gatts_evt_write_t *request)
+    {
+        // Only START_DFU resets, and the library reads this byte without checking len, so match it exactly.
+        if (request->data[0] != 1) {
```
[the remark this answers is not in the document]
thebentern: Not taking this one. The reachability premise is right, the safety conclusion is not, and the suggested fix would introduce a worse bug than the one it reports.

Where you are right: `_chr_control.setMaxLen(23)` at BLEDfu.cpp:205 with no `setFixedLen()` leaves the attribute variable-length, so the SoftDevice will authorize a zero-length write and it does reach `onControlWrite()`. I checked that before dismissing it.

Where the read is concerned: `request->data[0]` is in bounds on both delivery paths. `ble_gatts_evt_write_t` ends in a `uint8_t data[1]` placeholder, not a C99 flexible array (s140 ble_gatts.h:371, s145 ble_gatts.h:369), so `sizeof(*wr_req)` is 14 with `data` at offset 12. On the default deferred path BLECharacteristic.cpp:447 passes that `sizeof` to `ada_callback`, which `rtos_malloc`s and `memcpy`s exactly those 14 bytes (AdaCallback.c:113-119). On the inline fallback at :449 the pointer is into `ev_buf = rtos_malloc(BLE_EVT_LEN_MAX(BLE_GATT_ATT_MTU_MAX))` (bluefruit.cpp:747), a fixed maximum-size buffer of roughly 500 bytes that `sd_ble_evt_get` only partially fills. At `len == 0` the byte's *value* is indeterminate; the read is not out of bounds, and there is no third path — `_wr_authorize_cb` is invoked only from `case BLE_GATTS_OP_WRITE_REQ` (BLECharacteristic.cpp:442), with PREP/EXEC handled in sibling arms, and an `ada_callback` allocation failure degrades to the *larger* buffer, never a shorter one.

Where a `len` guard would actively break things: the library's reset predicate is `request->data[0] == START_DFU` with no length check (BLEDfu.cpp:110, byte-identical in the nRF54 core). The wrapper forwards the same pointer it inspected, and nothing writes to that buffer in between — the deferred copy is private, and the inline path runs on `adafruit_ble_task`, which owns `ev_buf` and cannot pull another event while we are blocked. So the wrapper's gate and the library's predicate read the same byte at the same address and agree by construction. That exactness is the point. Adding `request->len &&` would desync them: a zero-length write carrying a stale `0x01` would skip our quiesce and still reset inside the library — precisely the torn page this PR exists to fix.

Where the literal suggestion goes wrong: "reject zero-length requests, then delegate non-empty requests" means not calling `libraryCb` at all for `len == 0`. The only `sd_ble_gatts_rw_authorize_reply` for a plain `WRITE_REQ` on this characteristic is issued from inside that callback (BLEDfu.cpp:102 and :107); the `BLE_GATTS_OP_WRITE_REQ` arm at BLECharacteristic.cpp:442-452 is an `if` with no `else` and no fallback reply. Dropping the call leaves the ATT transaction unanswered until the peer's 30 s timeout tears down the bearer — a DFU-control denial of service reachable by any connected client, added to fix a read that is in bounds.

Two consistency notes. The in-tree secure handler has the identical unchecked read at src/platform/nrf52/BLEDfuSecure.cpp:80 and was not flagged. And a CCCD write — literally the bytes `01 00`, the one shape that would spuriously trip a byte gate — cannot reach here at all: `wr_auth` is set only on the value attribute (BLECharacteristic.cpp:205) and `cccd_md` is memclr'd at :292, so CCCD writes arrive as `BLE_GATTS_EVT_WRITE`, never as an authorize request.

Leaving the code as is. The comment on line 40 already says why the byte is read unchecked.

caveman99 (approved):

src/mesh/NodeDB.cpp
```diff
@@ -3395,6 +3396,31 @@ bool NodeDB::saveToDiskNoRetry(int saveWhat)
...
+    dir.close();
+
+    // An existing pref proves the metadata chain resolves; a fresh device has none to check.
+    for (const char *name : {deviceStateFileName, configFileName, channelFileName}) {
+        if (!FSCom.exists(name))
+            continue;
+        auto f = FSCom.open(name, FILE_O_READ);
+        if (!f)
+            return false;
+        const bool readable = f.read() >= 0;
```
NomDeTom, line 3417: `f.read() >= 0` makes a zero-length pref read as unreadable, so an empty `devicestate.proto` takes the format branch. That is the one input where this probe answers destructively, and it is the opposite of how every other failure shape here resolves: a missing `/prefs`, a file that will not open, a file that is not there at all all end in `return false` with no format.

SafeFile's write-tmp, verify, rename makes a zero-byte live file all but unreachable in practice, so this is a note rather than a defect. But if the bias of the whole change is "do not format unless the filesystem is provably gone", a successful `open()` is already that proof and the read adds a way to get the destructive answer wrong. Letting the `!f` check carry the negative on its own would close it.

The cost of that is that a file which opens but cannot be read stops counting as corruption, which on this PR's reasoning looks like the right trade rather than a regression.

src/mesh/NodeDB.cpp
```diff
@@ -3408,13 +3434,58 @@ bool NodeDB::saveToDisk(int saveWhat)
 
     bool success = saveToDiskNoRetry(saveWhat);
 
+    // A failed write is far more often a busy SoftDevice or a sagging rail than a corrupt filesystem,
+    // and the format below takes every file with it, so retry first and never format on a low rail.
+    for (int attempt = 1; !success && attempt <= 2; attempt++) {
+        delay(150);
```
NomDeTom, line 3440: Worth adding to the bench list rather than the review list: this loop lengthens a stall on the BLE task specifically, and by a number the PR description does not yet cover.

A phone config write reaches `saveToDisk()` synchronously on that task - `toRadio.setWriteCallback(onToRadioWrite, false)` at NRF52Bluetooth.cpp:237 means no Callback-task deferral, unlike the DFU control point, and `PhoneAPI::handleToRadioPacket` dispatches admin messages on the calling task. So a save that fails twice now costs that task 300 ms of `delay()` plus two extra write attempts, and each attempt can block further on `spiLock` held by a warm-ring page rotation.

Not claiming a timeout - your reply on the NRF52Bluetooth thread already establishes that the authorize reply is asynchronous and the 30 s ATT budget is not the constraint. The point is just that the rotation figure quoted in the description is measured against the rotation alone, and the rotation plus this loop is a different figure. Both halves compile out of the native suite, so the rak4631 run is the only place it would show.

NomDeTom (commented):
Taking a break from making new things to break. Two small things from a read of the current head, neither of them blocking. A third is a reply on the trunk.yaml thread.

.trunk/trunk.yaml
```diff
@@ -149,6 +149,11 @@ lint:
     - linters: [gitleaks]
       paths:
         - test/fixtures/nodedb/seed_v25_*.jsonl
+    # trufflehog's Lob detector matches "test_" followed by alphanumerics, which is every
+    # Unity test function name. There are no credentials in these suites.
+    - linters: [trufflehog]
+      paths:
+        - test/**/test_main.cpp
```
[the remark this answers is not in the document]
NomDeTom: Reopening this one because it is the only thread on the PR that was resolved without a reply, and the concern still holds on the current head.

The rule disables the whole `trufflehog` linter for every current and future `test/**/test_main.cpp`, and those suites are where the PSKs and keypairs live - `test_packet_signing` and the PKI suites carry real key material as fixtures. The two exemptions immediately above it in the same file are both scoped to named fixture paths rather than to a glob over a whole tree, so this one is the outlier in its own neighbourhood.

The names that actually trip the Lob detector are a narrow shape: a fixed-length alphanumeric run after the `test_` prefix, which is why it fires on a handful of suites and not on the hundred-odd others. Enumerating those, or scoping the exemption to the Lob detector if trunk supports that granularity, would keep the rest of the scan on the files most likely to grow a real secret.

Entirely possible this was considered and rejected for a reason that is not in the thread - if so, the reasoning is what I am after, not the change.

